PR #7440, opened 2021-04-06
updating cue labels

happythenewsad:
## WHAT
Adds que labeling. Certain questions have been edited to use keys that are not in the current mappings. This address that issue.

### Screenshots

### Notion Card Links
https://www.notion.so/quill/Update-ELL-instructions-e9a59f9328e04288883c461cbba76664

PR Checklist | Your Answer
------------ | -------------
Have you added and/or updated tests? |  n/a - data only


anathomical (approved):
In the long run we probably want some sort of automated test or script that confirms that all of our translation files contain the same keys, but for now just updating them here seems like the right approach.

PR #7443, opened 2021-04-07
fix issue with removing student whose id is in the selected student array

emilia-friedberg:
## WHAT
Fix case where a user selects a student by checking the checkbox in their row, then removes them, resulting in a nonexistent user's in the `selectedStudentIds` array.

## WHY
This was causing the page to break.

## HOW
Fix the immediate error by nullchecking the existence of `student`, but also clear the `selectedStudentIds` array on the `onSuccess` callback from the modal actions, since it doesn't really make sense for that to persist anyway. 

### Screenshots
N/A

### Notion Card Links
https://www.notion.so/quill/Student-removal-from-classroom-does-not-render-result-correctly-6c6096bb3c014eea85a851728b8ce526

PR Checklist | Your Answer
------------ | -------------

### DIFF
--- a/services/QuillLMS/client/app/bundles/Teacher/components/classrooms/classroom_student_section.tsx
+++ b/services/QuillLMS/client/app/bundles/Teacher/components/classrooms/classroom_student_section.tsx
@@ -1,14 +1,14 @@
 import * as React from 'react'
 import moment from 'moment'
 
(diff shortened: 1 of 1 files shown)


happythenewsad (approved):
🎉

eadams17 (approved):

PR #7444, opened 2021-04-07
Comprehension/ fix prompt order

eadams17:
## WHAT
fix issue with prompts being displayed out of order in the student view

## WHY
they should be displayed in the correct order: because, but, so

## HOW
just add a sort to incoming prompt props

### Screenshots

### Notion Card Links
https://www.notion.so/quill/The-prompts-are-out-of-order-in-the-student-activity-2482931fdb9840bf85f1bf84ef565e9a

PR Checklist | Your Answer
------------ | -------------

### DIFF
--- a/services/QuillLMS/client/app/bundles/Comprehension/components/studentView/container.tsx
+++ b/services/QuillLMS/client/app/bundles/Comprehension/components/studentView/container.tsx
@@ -375,7 +375,8 @@ export class StudentViewContainer extends React.Component<StudentViewContainerPr
     const { currentActivity, } = activities
     const { submittedResponses, } = session
     if (!currentActivity) return
-    const filteredSteps = _.sortBy( currentActivity.prompts, 'id' );
+    // sort by conjunctions in alphabetical order: because, but, so
+    const filteredSteps = _.sortBy( currentActivity.prompts, 'conjunction' );
     const steps =  filteredSteps.map((prompt, i) => {
       // using i + 2 because the READ_PASSAGE_STEP is 1, so the first item in the set of prompts will always be 2
       const stepNumber = i + 2
--- a/services/QuillLMS/client/app/bundles/Comprehension/components/studentView/container.tsx
+++ b/services/QuillLMS/client/app/bundles/Comprehension/components/studentView/container.tsx
@@ -1,7 +1,6 @@
 import * as React from "react";
 import queryString from 'query-string';
 import { connect } from "react-redux";
-import * as _ from 'underscore'
 
 import PromptStep from './promptStep'
 import StepLink from './stepLink'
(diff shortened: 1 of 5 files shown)


services/QuillLMS/client/app/bundles/Comprehension/components/studentView/container.tsx
```diff
@@ -374,8 +375,8 @@ export class StudentViewContainer extends React.Component<StudentViewContainerPr
     const { currentActivity, } = activities
     const { submittedResponses, } = session
     if (!currentActivity) return
-
-    const steps =  currentActivity.prompts.map((prompt, i) => {
+    const filteredSteps = _.sortBy( currentActivity.prompts, 'id' );
```
anathomical: Do we have any guarantee that the prompts will have been created in order of ascending order?  It's possible that this is enforced on the front-end, but I don't think there's anything on the back-end that guarantees this.

anathomical (commented):
Just a quick question around our sorting methodology with worries that there are still edge cases where it might fail.

services/QuillLMS/client/app/bundles/Comprehension/components/studentView/container.tsx
anathomical: One thought: we could probably sort by `currentActivity.prompts, 'conjunction'` since, by happenstance, 'because', 'but', and 'so' are in alphabetical order.

eadams17: @anathomical I actually noticed that and made that change earlier, not sure why it didn't commit... just pushed again

In regards to the backend, do you have any proposal as to how we could enforce on the backend?

anathomical: I mean, it's possible that we could add a default scope to the model on the back-end so that it always comes out in the intended order, but as long as we have an algorithm that will always work on the front-end (which it looks like we do sorting by conjunction, at least until new conjunctions get added or something which will probably never happen), we should be save.

anathomical (approved):

services/QuillLMS/client/app/bundles/Comprehension/components/studentView/container.tsx
```diff
@@ -374,8 +375,9 @@ export class StudentViewContainer extends React.Component<StudentViewContainerPr
     const { currentActivity, } = activities
     const { submittedResponses, } = session
     if (!currentActivity) return
-
-    const steps =  currentActivity.prompts.map((prompt, i) => {
+    // sort by conjunctions in alphabetical order: because, but, so
+    const filteredSteps = _.sortBy( currentActivity.prompts, 'conjunction' );
```
emilia-friedberg: My preference is always to use native methods where possible, so I would do `currentActivity.prompts.sort((a, b) => a.conjunction.localeCompare(b.conjunction))` here instead of including `underscore`.

emilia-friedberg (approved):
Approving, with one suggested change I leave up to you.

services/QuillLMS/client/app/bundles/Comprehension/components/studentView/container.tsx
eadams17: why do you prefer them? for compatibility reasons?

cissyxyu (approved):

PR #7448, opened 2021-04-07
make updates for FeedbackHistory to use feedback_type autoML rather than semantic

eadams17:
## WHAT
fix issue with mismatch between `FeedbackHistory` feedback_type and `Comprehension::Rule` rule_type

## WHY
we want the correct feedback to be given for AutoML labels

## HOW
update references of `feedback_type` to be "autoML" rather than "semantic"

### Screenshots

### Notion Card Links
https://www.notion.so/quill/Second-layer-feedback-and-highlighting-isn-t-working-53a1f5b006b74897bd0de9ecf6fbe9f0

PR Checklist | Your Answer
------------ | -------------

### DIFF
--- a/services/QuillLMS/lib/tasks/update_feedback_history_type.rake
+++ b/services/QuillLMS/lib/tasks/update_feedback_history_type.rake
@@ -1,11 +1,8 @@
 namespace :update_feedback_history_type do
   desc 'update feedback histories with type "semantic" to "autoML" to match rule_type'
   task :run => :environment do
-    FeedbackHistory.all.each do |feedback_history|
-      if feedback_history.feedback_type == 'semantic'
-        feedback_history.feedback_type = 'autoML'
-        feedback_history.save!
-      end
+    FeedbackHistory.where(feedback_type: 'semantic').each do |feedback_history|
+      feedback_history.update(feedback_type: 'autoML')
     end
   end
 end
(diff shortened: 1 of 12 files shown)


services/QuillLMS/lib/tasks/update_feedback_history_type.rake
```diff
@@ -0,0 +1,11 @@
+namespace :update_feedback_history_type do
+  desc 'update feedback histories with type "semantic" to "autoML" to match rule_type'
+  task :run => :environment do
+    FeedbackHistory.all.each do |feedback_history|
```
anathomical: This is pretty minor, but since you're in Rails, you can actually add a `where` clause here to reduce the number of records in the loop:
```ruby
FeedbackHistory.where(feedback_type: 'semantic').each do |feedback_history|
  feedback_history.update(feedback_type: 'autoML')
end
```

anathomical (approved):
A couple of suggested tweaks around your migration, but it's purely stylistic.  The logic all makes sense to me.

services/QuillLMS/lib/tasks/update_feedback_history_type.rake
eadams17: oh cool! I'll make this change

cissyxyu (approved):

PR #7450, opened 2021-04-07
Small refactor to segment calls to fix nil errors

dandrabik:
## WHAT
We are getting a bunch of `ActiveRecord::RecordNotFound` from Segment event background jobs. These seem to be cases in which the user or unit doesn't exist.
## WHY
1. These errors are triggering error rate alerts, since they get retried many times quickly. I temporarily removed `ActiveRecord::RecordNotFound` from our error rate, but it would be nice to put that back in.
2. We should send these events along with the information we have, even if it's partial. If we find that partial events are an issue, we can resolve those.
## HOW
1. Converted a few `.find()` to `find_by_id()` which won't raise.
2. Instead of having every method search for the `user` and nil check, I slightly refactored the code so that `track(user, options)` changes back to `track(options)` and uses `options[:user_id]` to determine find the user if it's needed.
3. Changed the Preview Activity worker to also search by Activity UID.

I also changed to check for 

### Screenshots

Screen Shot 2021-04-07 at 6 03 12 PM
Screen Shot 2021-04-07 at 6 03 27 PM
Screen Shot 2021-04-07 at 6 03 46 PM

PR Checklist | Your Answer
------------ | -------------
Have you added and/or updated tests? |  YES'
Have you deployed to Staging? | Not yet - deploying soon)

### DIFF
--- a/services/QuillLMS/app/services/analytics/segment_analytics.rb
+++ b/services/QuillLMS/app/services/analytics/segment_analytics.rb
@@ -28,7 +28,7 @@ class SegmentAnalytics
   end
 
   def track_activity_assignment(teacher_id, activity_id)
-    activity = Activity.find(activity_id)
+    activity = Activity.find_by(id: activity_id) || Activity.find_by(uid: activity_id)
 
     # properties here get used by Heap
     track({
@@ -124,7 +124,7 @@ class SegmentAnalytics
   end
 
   def track_previewed_activity(user_id, activity_id)
-    activity = Activity.find_by_id(activity_id) || Activity.find_by_uid(activity_id)
+    activity = Activity.find_by(id: activity_id) || Activity.find_by(uid: activity_id)
 
     track({
       user_id: user_id,
--- a/services/QuillLMS/app/services/analytics/segment_analytics.rb
+++ b/services/QuillLMS/app/services/analytics/segment_analytics.rb
@@ -21,19 +21,17 @@ class SegmentAnalytics
     # make sure that event name is written as a string in the pattern of
     # those in app/services/analytics/segment_io.rb
     # i.e. "BUILD_YOUR_OWN_ACTIVITY_PACK"
-    user = User.find(user_id)
-    track(user, {
+    track({
        user_id: user_id,
        event: "SegmentIo::BackgroundEvents::#{event_name}".constantize
       })
   end
 
   def track_activity_assignment(teacher_id, activity_id)
-    user = User.find(teacher_id)
     activity = Activity.find(activity_id)
 
     # properties here get used by Heap
-    track(user, {
+    track({
       user_id: teacher_id,
       event: SegmentIo::BackgroundEvents::ACTIVITY_ASSIGNMENT,
       properties: activity_info_for_tracking(activity)
@@ -41,7 +39,7 @@ class SegmentAnalytics
 
     # this event is for Vitally, which does not show properties
     if Activity.diagnostic_activity_ids.include?(activity_id)
-      track(user, {
+      track({
         user_id: teacher_id,
         event: "#{SegmentIo::BackgroundEvents::DIAGNOSTIC_ASSIGNMENT} | #{activity.name}"
       })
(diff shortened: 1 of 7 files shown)


services/QuillLMS/app/services/analytics/segment_analytics.rb
```diff
@@ -21,38 +21,35 @@ def track_event_from_string(event_name, user_id)
...
-    track(user, {
+    track({
        user_id: user_id,
        event: "SegmentIo::BackgroundEvents::#{event_name}".constantize
       })
   end
 
   def track_activity_assignment(teacher_id, activity_id)
-    user = User.find(teacher_id)
     activity = Activity.find(activity_id)
```
brendanshean: Would `activity = Activity.find_by_id(activity_id) || Activity.find_by_uid(activity_id)` apply here?

brendanshean (approved):
I made one small suggestion, but overall the refactoring looks good to me.

emilia-friedberg (commented):
I guess I wonder what the point of sending just the user id, if the user doesn't seem to exist in our system at all, is. Would it make more sense to just add guard statements where you've removed the `user = User.find(teacher_id)` lines to return if there is no user? We could use `.find_by_id` instead so we don't wind up with a NewRelic error.

services/QuillLMS/app/services/analytics/segment_analytics.rb
dandrabik: Good call. Doing traces, I didn't find occasions of this method being passed `uid`, but I agree that it makes sense to make them consistent. Will update.

happythenewsad:
Similar to Emilia's thought, I'm concerned that this fix doesn't address the (potential) root cause, which is that sometimes our user_id's don't reference actual User table rows. I postulate that an upstream bug (or misunderstanding of expected behavior) is the cause.

Moreover, a grep though the monorepo yields zero invocations for `track_event_from_string`. The current NR sample of this error includes `track_classroom_creation` in the stacktrace, although the sample size is 1 so YMMV.

services/QuillLMS/app/services/analytics/segment_analytics.rb
emilia-friedberg: There is a `.find_by_id_or_uid` method on the activity model you could use!

brendanshean: Although as I'm doing the Rails 5 upgrade, I'm realizing that instead would recommend a non-dynamic finder with: : `Activity.find_by(id: activity_id) || Activity.find_by(uid: activity_id)`

dandrabik: @emilia-friedberg Good point. I was going to use that method, but `Activity.find_by_id_or_uid` raises if a record isn't found, which I was trying to avoid here, so I went with the non-raising finders.

dandrabik:
> I guess I wonder what the point of sending just the user id, if the user doesn't seem to exist in our system at all, is. Would it make more sense to just add guard statements where you've removed the `user = User.find(teacher_id)` lines to return if there is no user? We could use `.find_by_id` instead so we don't wind up with a NewRelic error.

@emilia-friedberg @happythenewsad I agree that this doesn't solve the root issue. What I'm hoping is that it helps us get more information to solve the root issue (I probably should have explained that more). Right now, the segment events don't send because they error, so trying to debug the root cause is difficult. If we fix errors here, we can then look at the Segment events that have no `user_id` as more information. That's why I avoided a guard statement for now (also, I could just put that in the track method for DRY reasons).

happythenewsad:
> > I guess I wonder what the point of sending just the user id, if the user doesn't seem to exist in our system at all, is. Would it make more sense to just add guard statements where you've removed the `user = User.find(teacher_id)` lines to return if there is no user? We could use `.find_by_id` instead so we don't wind up with a NewRelic error.
> 
> @emilia-friedberg @happythenewsad I agree that this doesn't solve the root issue. What I'm hoping is that it helps us get more information to solve the root issue (I probably should have explained that more). Right now, the segment events don't send because they error, so trying to debug the root cause is difficult. If we fix errors here, we can then look at the Segment events that have no `user_id` as more information. That's why I avoided a guard statement for now (also, I could just put that in the track method for DRY reasons).

OK, I'll approve this as an investigatory PR rather than a terminal PR

happythenewsad (approved):

PR #7451, opened 2021-04-07
Comprehension/active activity sessions

anathomical:
## WHAT
- Save data to `active_activity_session` whenever a user submits a response so that they can later come back and resume an activity that they stop in the middle of
- Load `active_activity_session` data when Comprehension first starts and attempt to restore the state of the session with entries, attempt counts, and feedback shown.
## WHY
We want to support the ability for a student to pause an activity partway through and return at a later date without having to start all over.
## HOW
We stash only a couple of values in `active_activity_session`:
- `submittedResponses`, which is a record of what responses a student sent in answer to a prompt along with what feedback they got for that response
- `activeStep`, which is just a flag for which prompt they last submitted a response for.  This allows us to restore a user to a later prompt if they temporarily skipped an earlier one.
- `completedSteps`, which is an array of `activeSteps` that are already done.  This allows us to know which prompts a student can no longer submit responses for.

Note: I'm especially uncertain about the approach that I've taken in `components/studentView/promptStep.tsx` and would love suggestions on how to clean up the spaghetti of intertwined functions that I turned that into.

### Notion Card Links
https://www.notion.so/Recording-ConceptResults-and-ActivitySessions-62c39ad1338e49d48210fb1d56201599

PR Checklist | Your Answer
------------ | -------------
Have you added and/or updated tests? |  No.  I don't believe that we're currently testing this sort of state-based restoration logic in the front-end.  It's possible that we should be, and if anyone has suggestions on that front, I'd welcome them.
Self-Review: Have you done an initial self-review of the code below on Github? | No


services/QuillLMS/client/app/bundles/Comprehension/actions/session.ts
```diff
@@ -17,12 +17,68 @@ interface GetFeedbackArguments {
...
+}
+
 export const setSessionId = (sessionID: string) => {
   return (dispatch: Function) => {
     dispatch({ type: ActionTypes.SET_ACTIVITY_SESSION_ID, sessionID });
   }
 }
 
+export const fetchActiveActivitySession = (args: FetchActiveActivitySessionArguments) => {
+  const { sessionID, activityUID, callback, } = args
```
emilia-friedberg: this is nbd but you can do `export const fetchActiveActivitySession = ({ sessionID, activityUID, callback, }: FetchActiveActivitySessionArguments) => {` and then you don't need the following line.

services/QuillLMS/client/app/bundles/Comprehension/actions/session.ts
```diff
@@ -17,12 +17,68 @@ interface GetFeedbackArguments {
...
+      if (r.statusCode < 200 || r.statusCode >= 300) return
+      const { submittedResponses, } = body
+      dispatch({ type: ActionTypes.SET_SUBMITTED_RESPONSES, submittedResponses });
+      if (callback) callback(body)
+    })
+  }
+}
+
+export const saveActiveActivitySession = (args: SaveActiveActivitySessionArguments) => {
+  const { sessionID, submittedResponses, activeStep, completedSteps, callback, } = args
```
emilia-friedberg: Same as above.

services/QuillLMS/client/app/bundles/Comprehension/components/studentView/promptStep.tsx
```diff
@@ -36,9 +36,11 @@ export default class PromptStep extends React.Component<PromptStepProps, PromptS
   constructor(props: PromptStepProps) {
     super(props)
 
+    const { submittedResponses, } = this.props
+
     this.state = {
-      html: this.formattedPrompt(),
-      numberOfSubmissions: 0,
+      html: this.formattedPrompt(submittedResponses),
+      numberOfSubmissions: submittedResponses.length || 0,
```
emilia-friedberg: if `submittedResponses` is an empty array, `submittedResponses.length` is going to evaluate to 0 anyway, and if it's null, calling `.length` will error, right? So I don't think the `|| 0` here is ever going to come into effect.

emilia-friedberg (approved):
LGTM! A couple small suggestions you can take or leave. This is very exciting!

services/QuillLMS/client/app/bundles/Comprehension/components/studentView/promptStep.tsx
```diff
@@ -59,7 +61,12 @@ export default class PromptStep extends React.Component<PromptStepProps, PromptS
 
   stripHtml = (html: string) => html.replace(/<p>|<\/p>|<u>|<\/u>|<b>|<\/b>|<br>|<br\/>/g, '').replace(/&nbsp;/g, ' ')
 
-  formattedPrompt = () => {
+  formattedPrompt = (submittedResponses?: Array<string>) => {
+    if (submittedResponses && submittedResponses.length) {
+      const lastSubmission = submittedResponses[submittedResponses.length - 1]
+      const = formattedText = this.formatHtmlForEditorContainer(lastSubmission.entry, true)
```
eadams17: think we have an extra "=" here

eadams17 (approved):

PR #7458, opened 2021-04-09
TeacherFix bug

happythenewsad:
## WHAT
removes situation where Unit creation fails silently. 

## WHY
Sentry surfaced this error 

### Screenshots

### Notion Card Links
https://www.notion.so/quill/500-error-TeacherFixController-move_student_from_one_class_to_another-14601ff38edf4e8ab17b99ac68a1c8f4

PR Checklist | Your Answer
------------ | -------------

### DIFF
--- a/services/QuillLMS/app/models/concerns/student.rb
+++ b/services/QuillLMS/app/models/concerns/student.rb
@@ -69,7 +69,7 @@ module Student
       else
 
         new_unit_name = "#{name}'s Activities from #{old_classroom.name}"
-        unit = Unit.create(user_id: new_classroom.owner.id, name: new_unit_name)
+        unit = Unit.create!(user_id: new_classroom.owner.id, name: new_unit_name)
         new_cu = ClassroomUnit.find_or_create_by(unit_id: unit.id, classroom_id: new_classroom_id, assigned_student_ids: [user_id])
 
         classroom_units.each do |cu|
--- a/services/QuillLMS/app/models/concerns/student.rb
+++ b/services/QuillLMS/app/models/concerns/student.rb
@@ -67,9 +67,8 @@ module Student
           hide_extra_activity_sessions(cu.id)
         end
       else
-
         new_unit_name = "#{name}'s Activities from #{old_classroom.name}"
-        unit = Unit.create!(user_id: new_classroom.owner.id, name: new_unit_name)
+        unit = Unit.create_with_incremented_name(user_id: new_classroom.owner.id, name: new_unit_name)
         new_cu = ClassroomUnit.find_or_create_by(unit_id: unit.id, classroom_id: new_classroom_id, assigned_student_ids: [user_id])
 
         classroom_units.each do |cu|
--- a/services/QuillLMS/app/models/unit.rb
+++ b/services/QuillLMS/app/models/unit.rb
@@ -73,6 +73,16 @@ class Unit < ActiveRecord::Base
     end
   end
 
+  def self.create_with_incremented_name(user_id:, name: )
+    unit = Unit.create(user_id: user_id, name: name)
+    return unit if unit.persisted?
+    (2..20).each do |counter|
+      unit = Unit.create(user_id: user_id, name: "#{name} #{counter}")
+      return unit if unit.persisted?
+    end
+    false
+  end
+
   private def create_any_new_classroom_unit_activity_states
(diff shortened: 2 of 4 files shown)


services/QuillLMS/app/models/concerns/student.rb
```diff
@@ -69,7 +69,7 @@ def move_activity_sessions(old_classroom, new_classroom)
       else
 
         new_unit_name = "#{name}'s Activities from #{old_classroom.name}"
-        unit = Unit.create(user_id: new_classroom.owner.id, name: new_unit_name)
+        unit = Unit.create!(user_id: new_classroom.owner.id, name: new_unit_name)
```
happythenewsad: @dandrabik The root issue was that Unit creations were failing validations silently, which meant that `unit.id` was `nil` on the next line.

The only way a Unit can fail validations is through a unique name collision. 

Would you provide some product direction on next steps? we could:
- concatenate a string onto the new name to guarantee uniqueness (opinionated, but solves the issue)
- allow the `Unit.create!` exception to be thrown, and delegate next steps to the user. 
- merge this PR as-is, and observe the names associated with the failed controller transactions to see if there's an obvious behavior pattern

anathomical: I think that I would be inclined to add a date/time stamp to the unit name.  It might be worth confirming with Dan G, but I suspect that there are valid reasons to want to move `ActivitySessions` from a classroom to somewhere else more than once.

happythenewsad: Awaiting reply from Dan G

dandrabik: @happythenewsad I would favor creating an auto-incremented retry flow for a limited number of attempts that appends a count to the end, e.g. `"Janes's Activities from Classroom"` then `"Janes's Activities from Classroom 2"` . It's opinionated, but a pattern we use in other places.

happythenewsad: From Dan G:
> Definitely—seems like a great fix, and there's no reason that activity session name change would negatively impact support/partnerships!

dandrabik: Let's add some logic to append a number then. If there's a duplicate, add a 2, else increment from there for a limited amount of numbers (let's say 20).

dandrabik (changes requested):
Per Dan. G's comment. Let's change this from a `raise` to adding a new unit name.

services/QuillLMS/app/models/concerns/student.rb
```diff
@@ -67,9 +67,8 @@ def move_activity_sessions(old_classroom, new_classroom)
           hide_extra_activity_sessions(cu.id)
         end
       else
-
         new_unit_name = "#{name}'s Activities from #{old_classroom.name}"
-        unit = Unit.create(user_id: new_classroom.owner.id, name: new_unit_name)
+        unit = Unit.create_with_incremented_name(user_id: new_classroom.owner.id, name: new_unit_name)
         new_cu = ClassroomUnit.find_or_create_by(unit_id: unit.id, classroom_id: new_classroom_id, assigned_student_ids: [user_id])
```
happythenewsad, line 72: Should all 20 similar names be already taken for some reason, this line will raise a RuntimeError

happythenewsad: No way around this edge case  - alternatively we could append SecureRandom.id to the name and remove the retry logic.

services/QuillLMS/app/models/unit.rb
```diff
@@ -73,6 +73,16 @@ def email_lesson_plan
...
   end
 
+  def self.create_with_incremented_name(user_id:, name: )
+    unit = Unit.create(user_id: user_id, name: name)
+    return unit if unit.persisted?
+    (2..20).each do |counter|
+      unit = Unit.create(user_id: user_id, name: "#{name} #{counter}")
+      return unit if unit.persisted?
+    end
+    false
```
dandrabik, line 83: Per your other comment, we could keep the retry logic (and the more user friendly names), and have a less user friendly name as the fallback, e.g.
```suggestion
    Unit.create(user_id: user_id, name: "#{name} #{SecureRandom.uuid}")
```

dandrabik (approved):
Small suggestion for the fallback case (based on your comments), but even without, LGTM.

anathomical (approved):

PR #7464, opened 2021-04-12
Lms/feedback ratings

happythenewsad:
## WHAT
This is not a discrete feature, but rather an integration of FeedbackHistoryRatings with the Rules Report query and postprocessing 

### Screenshots

### Notion Card Links
https://www.notion.so/quill/Rules-Report-V1-f07e030088ad43a79849c1b48e97a857?d=728764b3-9e57-4992-80ff-108778c0b682#948d45599ccc4a809e89728a56f0a417

PR Checklist | Your Answer
------------ | -------------

### DIFF
--- a/services/QuillLMS/app/models/feedback_history.rb
+++ b/services/QuillLMS/app/models/feedback_history.rb
@@ -47,11 +47,11 @@ class FeedbackHistory < ActiveRecord::Base
   ]
 
   before_validation :confirm_prompt_type, on: :create
-
+  has_many :feedback_history_ratings
   belongs_to :activity_session, foreign_key: :activity_session_uid, primary_key: :uid
   belongs_to :prompt, polymorphic: true
   belongs_to :concept, foreign_key: :concept_uid, primary_key: :uid
-
+  belongs_to :rule, class_name: Comprehension::Rule,  foreign_key: :rule_uid
   validates :activity_session_uid, presence: true
   validates :concept_uid, allow_blank: true, length: {is: CONCEPT_UID_LENGTH}
   validates :attempt, presence: true,
--- a/services/QuillLMS/app/models/feedback_history.rb
+++ b/services/QuillLMS/app/models/feedback_history.rb
@@ -51,7 +51,7 @@ class FeedbackHistory < ActiveRecord::Base
   belongs_to :activity_session, foreign_key: :activity_session_uid, primary_key: :uid
   belongs_to :prompt, polymorphic: true
   belongs_to :concept, foreign_key: :concept_uid, primary_key: :uid
-  belongs_to :rule, class_name: Comprehension::Rule,  foreign_key: :rule_uid
+
   validates :activity_session_uid, presence: true
   validates :concept_uid, allow_blank: true, length: {is: CONCEPT_UID_LENGTH}
   validates :attempt, presence: true,
--- a/services/QuillLMS/lib/rule_feedback_history.rb
+++ b/services/QuillLMS/lib/rule_feedback_history.rb
@@ -33,6 +33,14 @@ class RuleFeedbackHistory
                 .feedbacks
                 .min_by {|f| f.order}.text
         end
+
+        #rules = rules_sql_result.includes(:feedback_histories)
+        
+        rule_feedbacks = Comprehension::Rule.includes(:feedbacks).where(uid: rules_sql_result.map(&:rules_uid)) 
+        rules_sql_result.each do |r|
+            r.first_feedback = rule_feedbacks.find_by(uid: r.rules_uid).feedbacks.sort_by {|f| f.order}.first.text
+        end
+
     end
 
     def self.format_sql_results(relations)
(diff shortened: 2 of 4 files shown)


services/QuillLMS/lib/rule_feedback_history.rb
```diff
@@ -1,38 +1,70 @@
...
-          LEFT JOIN feedback_histories as feedbacks ON feedbacks.rule_uid = rules.uid
+          LEFT JOIN feedback_histories ON feedback_histories.rule_uid = rules.uid
           WHERE prompts.conjunction = '#{conjunction}' AND activity_id = '#{activity_id}'
           GROUP BY rules_uid 
         SQL
         rules = Comprehension::Rule.find_by_sql(sql)
     end
 
     def self.postprocessing(rules_sql_result)
+        rules_sql_result.each do |r|
```
happythenewsad, line 30: thinking about factoring the loop into a separate method for legibility

services/QuillLMS/lib/rule_feedback_history.rb
```diff
@@ -1,38 +1,70 @@
...
+                r.define_singleton_method(:pct_scored) { 0 }
+            else 
+                total_strong = feedback_histories.reduce(0) do |memo, n|
+                    memo + n.feedback_history_ratings.filter(&:rating).count
+                end
+
+                r.define_singleton_method(:pct_strong) { total_strong/total_scored.to_f }
+                # note: pct_scored may be over counting since the original spec may not
+                # account for one feedback history having more than one score.
+                r.define_singleton_method(:pct_scored) { total_scored/total_scored.to_f }
```
emilia-friedberg, line 51: Can we format these as percentages, rather than decimals, so we don't need to change the values for display on the frontend?

services/QuillLMS/lib/rule_feedback_history.rb
```diff
@@ -1,38 +1,70 @@
...
 
     def self.postprocessing(rules_sql_result)
+        rules_sql_result.each do |r| 
+            
+            feedback_histories = FeedbackHistory
+                .where(id: r.feedback_histories_id_array).includes(:feedback_history_ratings)
+
+            total_scored = feedback_histories.reduce(0) do |memo, n|
+                memo + n.feedback_history_ratings.count
+            end
```
emilia-friedberg: Can we call `n` `feedback_history` to make this more readable?

emilia-friedberg (approved):
Two small comments, but LGTM overall.

services/QuillLMS/app/models/feedback_history.rb
```diff
@@ -47,11 +47,11 @@ class FeedbackHistory < ActiveRecord::Base
   ]
 
   before_validation :confirm_prompt_type, on: :create
-
+  has_many :feedback_history_ratings
   belongs_to :activity_session, foreign_key: :activity_session_uid, primary_key: :uid
   belongs_to :prompt, polymorphic: true
   belongs_to :concept, foreign_key: :concept_uid, primary_key: :uid
-
+  belongs_to :rule, class_name: Comprehension::Rule,  foreign_key: :rule_uid
```
anathomical: Not a game-breaker, but when we originally spec'ed the `FeedbackHistory` feature, one of the goals was for it to be generic enough to work with non-Comprehension activities.  We've never actually talked about what that would mean, so a `belongs_to` here is probably fine, just wanted to call it out.

services/QuillLMS/lib/rule_feedback_history.rb
```diff
@@ -1,38 +1,70 @@
...
+            if total_scored == 0 
+                total_strong = 0
+                r.define_singleton_method(:pct_strong) { 0 }
+                r.define_singleton_method(:pct_scored) { 0 }
+            else 
+                total_strong = feedback_histories.reduce(0) do |memo, n|
+                    memo + n.feedback_history_ratings.filter(&:rating).count
+                end
+
+                r.define_singleton_method(:pct_strong) { total_strong/total_scored.to_f }
```
anathomical, line 48: I don't have alternative suggestions, but having a bunch of `define_singleton_method` calls in here makes me think there might be a cleaner approach?  But I can see the advantage of keeping the values in `rules_sql_result` as the same object type that was passed in since this is a post-processing script.  So we can go with this.

anathomical (approved):

services/QuillLMS/app/models/feedback_history.rb
happythenewsad: Actually, after writing this line I opted to use a where query, so I can remove this!

services/QuillLMS/lib/rule_feedback_history.rb
happythenewsad: Yeah, I was looking around for a way to merge an object with a struct, but didn't find a way.

services/QuillLMS/lib/rule_feedback_history.rb
happythenewsad: yes

services/QuillLMS/lib/rule_feedback_history.rb
happythenewsad: 👍

PR #7466, opened 2021-04-12
Lms/rules report view 2 rebased

happythenewsad:
## WHAT
View 2 of Rules Report

## WHY
To organize feedbacks grouped by response in the admin reporting tool 

### Notion Card Links
https://www.notion.so/quill/view-2-responses-per-rule-3dd3906be7aa45a1bbe63e610878a227

PR Checklist | Your Answer
------------ | -------------

### DIFF
--- a/services/QuillLMS/config/routes.rb
+++ b/services/QuillLMS/config/routes.rb
@@ -364,6 +364,8 @@ EmpiricalGrammar::Application.routes.draw do
     namespace :v1 do
       get 'activities/uids_and_flags' => 'activities#uids_and_flags'
       get 'rule_feedback_histories' => 'rule_feedback_histories#by_conjunction'
+      get 'rule_feedback_history/:rule_uid' => 'rule_feedback_histories#rule_detail'
+
       resources :activities,              except: [:index, :new, :edit]
       resources :activity_flags,          only: [:index]
       resources :activity_sessions,       except: [:index, :new, :edit]
--- a/services/QuillLMS/lib/rule_feedback_history.rb
+++ b/services/QuillLMS/lib/rule_feedback_history.rb
@@ -26,12 +26,12 @@ class RuleFeedbackHistory
         rules = Comprehension::Rule.find_by_sql(sql)
     end
 
-    def self.feedback_history_to_json(fh)
+    def self.feedback_history_to_json(f_h)
         {
-            response_id: fh.id,
-            datetime: fh.updated_at,
-            entry: fh.entry,
-            highlight: fh.metadata.class == Hash ? fh.metadata['highlight'] : '',
+            response_id: f_h.id,
+            datetime: f_h.updated_at,
+            entry: f_h.entry,
+            highlight: f_h.metadata.class == Hash ? f_h.metadata['highlight'] : '',
             view_session_url: 'Not yet available'
         }
     end
@@ -39,8 +39,8 @@ class RuleFeedbackHistory
     def self.generate_rulewise_report(rule_uid)
         feedback_histories = FeedbackHistory.where(rule_uid: rule_uid)
         response_jsons = []
-        feedback_histories.each do |fh|
-            response_jsons.append(feedback_history_to_json(fh))
+        feedback_histories.each do |f_h|
+            response_jsons.append(feedback_history_to_json(f_h))
         end
 
         {
--- a/services/QuillLMS/lib/rule_feedback_history.rb
+++ b/services/QuillLMS/lib/rule_feedback_history.rb
@@ -26,6 +26,30 @@ class RuleFeedbackHistory
         rules = Comprehension::Rule.find_by_sql(sql)
     end
 
+    def self.feedback_history_to_json(fh)
+        {
+            response_id: fh.id,
+            datetime: fh.updated_at,
+            entry: fh.entry,
+            highlight: fh.metadata.class == Hash ? fh.metadata['highlight'] : '',
+            view_session_url: 'Not yet available'
+        }
+    end
+
+    def self.generate_rulewise_report(rule_uid)
+        feedback_histories = FeedbackHistory.where(rule_uid: rule_uid)
+        response_jsons = []
+        feedback_histories.each do |fh|
+            response_jsons.append(feedback_history_to_json(fh))
+        end
+
+        {
+            "#{rule_uid}": {
+                responses: response_jsons
+            }
+        }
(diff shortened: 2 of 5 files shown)


services/QuillLMS/config/routes.rb
```diff
@@ -364,6 +364,8 @@
     namespace :v1 do
       get 'activities/uids_and_flags' => 'activities#uids_and_flags'
       get 'rule_feedback_histories' => 'rule_feedback_histories#by_conjunction'
+      get 'rule_feedback_history/:rule_uid' => 'rule_feedback_histories#rule_detail'
```
emilia-friedberg, line 367: Do we need to also pass the activity id here? Aren't some rules shared across activities? cc @anathomical?

services/QuillLMS/lib/rule_feedback_history.rb
```diff
@@ -26,6 +26,30 @@ def self.exec_query(conjunction:, activity_id:)
...
     end
 
+    def self.feedback_history_to_json(f_h)
+        {
+            response_id: f_h.id,
+            datetime: f_h.updated_at,
+            entry: f_h.entry,
+            highlight: f_h.metadata.class == Hash ? f_h.metadata['highlight'] : '',
+            view_session_url: 'Not yet available'
+        }
```
emilia-friedberg: I think we need a `strength` attribute here, because we need that data for determining what to show on the frontend. I'm sorry for not calling that out on your original example payload - I think we hadn't yet decided whether or not "strength" was making it into this version of the page.

emilia-friedberg (commented):
This looks good, but I think there are a couple of additional things that need to be accounted for.

services/QuillLMS/config/routes.rb
anathomical: So, technically, `Rule` records are tied to `Prompt` records via the `PromptsRule` model.  It's possible that you want to be able to both get all `FeedbackHistory` for a specific rule, AND get only feedback for that rule if it belongs to a specific `Activity`.  I don't remember precisely what the UX requires here.

happythenewsad: I just re-scanned the project spec, but don't see a clear requirement to filter on Activity here. How about we move forward and add this modification in V2 if it's ever requested?

services/QuillLMS/lib/rule_feedback_history.rb
happythenewsad: addressed!

services/QuillLMS/config/routes.rb
emilia-friedberg: This page is nested below the activity, and you can only get to it by clicking on the rule from the rules analysis page that is explicitly activity-specific. In the mockup for the interface, you can see that the base prompt for the responses is shown at the top of the table, which doesn't make sense if there are responses being shown from more than one activity. So I do think it's clear from the spec that this is activity specific. That said, I'm fine moving on for now if you want to address it later. I don't know how many rules are cross-activity anyway.

services/QuillLMS/lib/rule_feedback_history.rb
```diff
@@ -26,6 +26,31 @@ def self.exec_query(conjunction:, activity_id:)
         rules = Comprehension::Rule.find_by_sql(sql)
     end
 
+    def self.feedback_history_to_json(f_h)
```
emilia-friedberg, line 29: This is sooooo much easier to read than the previous version of this method. Nice job.

emilia-friedberg (approved):

services/QuillLMS/lib/rule_feedback_history.rb
```diff
@@ -26,6 +26,31 @@ def self.exec_query(conjunction:, activity_id:)
...
+        {
+            response_id: f_h.id,
+            datetime: f_h.updated_at,
+            entry: f_h.entry,
+            highlight: f_h.metadata.class == Hash ? f_h.metadata['highlight'] : '',
+            view_session_url: 'Not yet available',
+            strength: f_h.feedback_history_ratings.order(updated_at: :desc).first&.rating
```
emilia-friedberg, line 36: Isn't this going to need to be specific to the current user?

emilia-friedberg (commented):
I approved too fast and then realized that there's a tweak we're going to need to how we determine the `strength` variable. I do also think the activity-specific query is ultimately going to be necessary, but I'm fine moving on for now.

services/QuillLMS/lib/rule_feedback_history.rb
happythenewsad: From the spec: 
`If someone changes a response from weak to strong, it should always show the new option instead.`

I believe the current implementation respects this specification. Note: I implemented this logic expansively  - allowing `FeedbackHistory` to have multiple `FeedbackHistoryRatings` - in anticipation of future features (i.e. majority voting).

emilia-friedberg (approved):

services/QuillLMS/config/routes.rb
anathomical: Just for informational purposes: the cross-activity rules are what we've sometimes called "universal".  That's basically everything but AutoML, Regex, and Plagiarism.  So Spelling, Opinion, all the Grammar stuff, that's all cross-activity.

services/QuillLMS/lib/rule_feedback_history.rb
```diff
@@ -26,6 +26,31 @@ def self.exec_query(conjunction:, activity_id:)
...
+        }
+    end
+
+    def self.generate_rulewise_report(rule_uid)
+        feedback_histories = FeedbackHistory.where(rule_uid: rule_uid)
+        response_jsons = []
+        feedback_histories.each do |f_h|
+            response_jsons.append(feedback_history_to_json(f_h))
+        end
```
anathomical, line 45: Is this more cleanly understood as a map?
```ruby
response_jsons = feedback_histories.map { |f_h| feedback_history_to_json(f_h) }
```
Feels slightly cleaner than creating an empty array and then populating it.

anathomical (approved):
One comment for clarity, but you can take it or leave it.

services/QuillLMS/lib/rule_feedback_history.rb
happythenewsad: Absolutely; I must have been tired. I just made a patch which I will apply in version 2 later this week

PR #7470, opened 2021-04-14
teacher dashboard metrics backend

emilia-friedberg:
## WHAT
Queries for the metrics section of the new teacher homepage.

## WHY
We want to be able to show teachers this data at the top of their screen.

## HOW
Just calculate the relevant pieces of data.

### Screenshots

### Notion Card Links
This is the actual card: https://www.notion.so/quill/Backend-for-teacher-home-page-metrics-3929f49d630c411aa5ca8795b20a66ec but you can see how the data gets used at the top of the page here: https://www.notion.so/quill/Home-Post-checklist-c6fa0a95ec1b4c8c92b8eac740e2bf9b

PR Checklist | Your Answer
------------ | -------------
Have you deployed to Staging? | NO - no frontend component yet

### DIFF
--- a/services/QuillLMS/app/queries/teacher_dashboard_metrics.rb
+++ b/services/QuillLMS/app/queries/teacher_dashboard_metrics.rb
@@ -1,33 +1,54 @@
 class TeacherDashboardMetrics
 
-  def self.queries(user)
-    today = Date.today
-    days_since_last_sunday = today.wday == 0 ? 0 : (today.wday + 6) % 7 + 1
-    last_sunday = today - days_since_last_sunday
-    july_first_of_this_year = Date.parse("01-07-#{today.year}")
-    last_july_first = today.month > 7 ? july_first_of_this_year : july_first_of_this_year - 1.year
-    classroom_ids = user.classrooms_i_teach.map(&:id)
-    {
-      weekly_assigned_activities_count: count_of_assigned_activities(classroom_ids, last_sunday),
-      yearly_assigned_activities_count: count_of_assigned_activities(classroom_ids, last_july_first),
-      weekly_completed_activities_count: count_of_completed_activities(classroom_ids, last_sunday),
-      yearly_completed_activities_count: count_of_completed_activities(classroom_ids, last_july_first)
+  def initialize(user)
+    @user = user
+  end
+
+  def run
+   {
+      weekly_assigned_activities_count: count_of_assigned_activities(last_sunday),
+      yearly_assigned_activities_count: count_of_assigned_activities(last_july_first),
+      weekly_completed_activities_count: count_of_completed_activities(last_sunday),
+      yearly_completed_activities_count: count_of_completed_activities(last_july_first)
     }
   end
 
-  def self.count_of_assigned_activities(classroom_ids, start_date)
+  def count_of_assigned_activities(start_date)
     classroom_units = ClassroomUnit.where(classroom_id: classroom_ids).where("created_at >= ?", start_date)
-    total_count = 0
-    classroom_units.each do |cu|
-      number_of_students = cu.assigned_student_ids.length
-      number_of_activities = UnitActivity.where(unit_id: cu.unit_id).count
+    classroom_units.reduce(0) do |total_count, classroom_unit|
+      number_of_students = classroom_unit.assigned_student_ids.length
+      number_of_activities = UnitActivity.where(unit_id: classroom_unit.unit_id).count
       total_count += number_of_students * number_of_activities
     end
-    total_count
   end
 
-  def self.count_of_completed_activities(classroom_ids, start_date)
+  def count_of_completed_activities(start_date)
     classroom_units = ClassroomUnit.where(classroom_id: classroom_ids)
     ActivitySession.where(classroom_unit_id: classroom_units.ids).where("completed_at >= ?", start_date).count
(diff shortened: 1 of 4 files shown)


services/QuillLMS/app/queries/teacher_dashboard_metrics.rb
```diff
@@ -0,0 +1,33 @@
+class TeacherDashboardMetrics
+
+  def self.queries(user)
+    today = Date.today
+    days_since_last_sunday = today.wday == 0 ? 0 : (today.wday + 6) % 7 + 1
+    last_sunday = today - days_since_last_sunday
+    july_first_of_this_year = Date.parse("01-07-#{today.year}")
+    last_july_first = today.month > 7 ? july_first_of_this_year : july_first_of_this_year - 1.year
+    classroom_ids = user.classrooms_i_teach.map(&:id)
```
brendanshean, line 9: I would suggest `pluck(:id)` instead of map.

brendanshean: I would suggest `pluck(:id)` instead of `map(&:id)`

services/QuillLMS/app/queries/teacher_dashboard_metrics.rb
```diff
@@ -0,0 +1,33 @@
...
+      weekly_assigned_activities_count: count_of_assigned_activities(classroom_ids, last_sunday),
+      yearly_assigned_activities_count: count_of_assigned_activities(classroom_ids, last_july_first),
+      weekly_completed_activities_count: count_of_completed_activities(classroom_ids, last_sunday),
+      yearly_completed_activities_count: count_of_completed_activities(classroom_ids, last_july_first)
+    }
+  end
+
+  def self.count_of_assigned_activities(classroom_ids, start_date)
+    classroom_units = ClassroomUnit.where(classroom_id: classroom_ids).where("created_at >= ?", start_date)
+    total_count = 0
```
brendanshean: You could get rid of the temporary variable `total_count` with the enumerable method `reduce`:

```
classroom_units.reduce(0) do |total_count, classroom_unit|
  number_of_students = classroom_unit.assigned_student_ids.length
  number_of_activities = UnitActivity.where(unit_id: classroom_unit.unit_id).count
  total_count += number_of_students * number_of_activities
end
```

brendanshean (commented):
I would opt for a more object based design like 

```
class TeacherDashboardMetricsQuery
  def initialize(user)
    @user = user
  end

  def run
   {
      weekly_assigned_activities_count: count_of_assigned_activities(classroom_ids, last_sunday),
      yearly_assigned_activities_count: count_of_assigned_activities(classroom_ids, last_july_first),
      weekly_completed_activities_count: count_of_completed_activities(classroom_ids, last_sunday),
      yearly_completed_activities_count: count_of_completed_activities(classroom_ids, last_july_first)
    }
  end
....
```

Then you add a bunch of private methods that set the individual variables `today` `last_sunday`, `days_since_last_sunday`, `classroom_ids`, etc.  For example,

```
def today
  @today ||= Date.today
end
```

This would cut down on the parameter passing and keep the interface more standard.

services/QuillLMS/app/queries/teacher_dashboard_metrics.rb
emilia-friedberg: Unfortunately, `classrooms_i_teach` is an array, so `pluck` isn't available.

services/QuillLMS/app/queries/teacher_dashboard_metrics.rb
```diff
@@ -0,0 +1,54 @@
...
+    @user = user
+  end
+
+  def run
+   {
+      weekly_assigned_activities_count: count_of_assigned_activities(last_sunday),
+      yearly_assigned_activities_count: count_of_assigned_activities(last_july_first),
+      weekly_completed_activities_count: count_of_completed_activities(last_sunday),
+      yearly_completed_activities_count: count_of_completed_activities(last_july_first)
+    }
```
anathomical, line 13: Honest question here: should we be caching this value?  I'm not really sure how expensive these queries might be, but pulling data out of `ActivitySession` has often been pretty time-consuming.  It might make sense to cache this hash by wrapping the whole thing in a `Rails.cache.fetch` block with an expiration of 24 hours and basically only recalculate it once a day.  Not sure how that might impact UX, though, or teacher expectations that their numbers will go up immediately after they assign new stuff.

anathomical (approved):
This looks good to me.  One open question around caching, but nothing that means we couldn't try this as-is.

services/QuillLMS/app/queries/teacher_dashboard_metrics.rb
brendanshean: My mistake.  I missed that `to_a` at the end of the execute call.

brendanshean (approved):
Looks great!

services/QuillLMS/app/queries/teacher_dashboard_metrics.rb
emilia-friedberg: Yeah, I hear that. I think it makes sense to deploy as is and then add caching if necessary, especially because, as you point out, we might have to think about whether to invalidate caches on unit assignment or something.

PR #7473, opened 2021-04-14
CMS: Configure sidekiq from sidekiq.yml

dandrabik:
## WHAT
Tiny PR. We currently have two competing "Number of sidekiq workers" configs in the CMS: `0002_sidekiq.config` and  `sidekiq.yml` with `0002_sidekiq.config` being the one that has precedence. Let's use `sidekiq.yml` instead and eliminate the other.
## WHY
We should only have one config for a setting. The one in `sidekiq.yml` is easier to use and is universal (the other just applies to elastic beanstalk).
## HOW
Remove the `-c ${SIDEKIQ_WORKERS:-2}` from the elastic beanstalk config. 

Also note, during testing I discovered that `0002_sidekiq.config` is only updated for new servers (not restarts or deploys), so when rolling this out, we'll have to put all new servers in place.

PR Checklist | Your Answer
------------ | -------------
Have you added and/or updated tests? |  ('NO', just a config change.)


brendanshean (approved):
Looks good to me.  I was unaware of that `services/QuillCMS/.ebextensions/0002_sidekiq.config` file so thank you for the explanation.

PR #7478, opened 2021-04-15
Add an endpoint to the CMS that returns some stats for the Admin Question Dashboard

cissyxyu:
## WHAT
As part of building out a new Admin Question Dashboard, I want to add an endpoint to the CMS server that returns some calculated data about question health.

## WHY
This data is important for admin to know, and we need to put the endpoint in CMS because the LMS does not store response authors/matched/unmatched data.

## HOW
Add the endpoint in CMS and make sure it's working and the `lib` path is loaded by the server.

### Screenshots

### Notion Card Links
https://www.notion.so/quill/Change-CMS-to-provide-correct-back-end-data-f95e6cd0c4f94e4aa0f39f1b4d7f2e02

PR Checklist | Your Answer
------------ | -------------


services/QuillCMS/app/controllers/responses_controller.rb
```diff
@@ -94,6 +94,10 @@ def grade_breakdown
     render json: optimality_counts_of_question(params[:question_uid])
   end
 
+  def question_dashboard
+    render json: question_dashboard_data(params[:question_uid])
+  end
+
   def incorrect_sequences
     render json: IncorrectSequenceCalculator.incorrect_sequences_for_question(params[:question_uid])
```
happythenewsad, line 103: IncorrectSequenceCalculator, which lives in lib/modules, is being loaded here without the need of your modification to `config/application.rb`. Do you need  the modification in `config/application.rb`?

services/QuillCMS/lib/modules/response_aggregator.rb
```diff
@@ -32,6 +42,17 @@ def optimality_counts_of_question(question_uid)
...
   end
 
+  def question_dashboard_data(question_uid)
+    total_number_of_responses = Response.where(question_uid: question_uid).sum('count')
+    common_unmatched = Response.where(question_uid: question_uid, parent_id: nil, optimal: nil).where('count >= 10').sum('count')
+    algorithm_matched = Response.where(question_uid: question_uid, author: DASHBOARD_ALGOS).sum('count')
+
+    {
+      percent_common_unmatched: (common_unmatched.to_f / (total_number_of_responses.nonzero? || 1)) * 100,
+      percent_specified_algos: (algorithm_matched.to_f / (total_number_of_responses.nonzero? || 1)) * 100
```
happythenewsad: Interesting - `nonzero?` is the only Rails API I've encountered that breaks the convention that methods ending in ? return booleans.

brendanshean: `File#size?` is another gotcha.

brendanshean: I think what's most confusing is `zero?` works as expected.

services/QuillCMS/spec/controllers/responses_controller_spec.rb
```diff
@@ -215,4 +215,25 @@ def filter_responses_with_parent_id(array)
...
     end
   end
+
+  describe 'question_dashboard_data' do
+    before do
+      allow_any_instance_of(Response).to receive(:create_index_in_elastic_search)
+    end
+
+    let!(:question_uid) { SecureRandom.uuid}
+    let!(:response_1) { create(:response, question_uid: question_uid, parent_id: nil, optimal: nil, count: 20)}
```
brendanshean: I don't think it's necessary to include `parent_id: nil` as the factory doesn't mention that field.

services/QuillCMS/spec/lib/modules/response_aggregator_spec.rb
```diff
@@ -0,0 +1,29 @@
...
+RSpec.describe ResponseAggregator do
+  include ResponseAggregator
+
+  describe "#question_dashboard_data" do
+    before do
+      allow_any_instance_of(Response).to receive(:create_index_in_elastic_search)
+    end
+
+    let!(:question_uid) { SecureRandom.uuid}
+    let!(:response_1) { create(:response, question_uid: question_uid, parent_id: nil, optimal: nil, count: 20)}
```
brendanshean: I don't think it's necessary to include `parent_id: nil` as the factory doesn't mention that field.

services/QuillCMS/lib/modules/response_aggregator.rb
brendanshean: Actually, I think using `zero?` would be more clear for both `percent_common_umatched` and `percent_specified_algos`.  For example (assuming refactor from module into an object like `AdminQuestionDashboard`):

```
def percent_common_unmatched
  num_responses.zero? ? 0 : (100.0 * common_unmatched / num_of_responses)
end
```

brendanshean (commented):
I think this would be a good opportunity to move away from the `lib/modules` pattern by introducing a new object under `app/models` called `AdminQuestionDashboard` with one public method `health`.  This would do a few things:

1) Adhere to single responsibility as well as encapsulation. (adding to ResponseAggregator violates both)
2) Cut down on parameter passing
3) Give more natural specs.

I'm happy to pair on this.

services/QuillCMS/app/controllers/responses_controller.rb
cissyxyu: I changed the path that this file lives in, so removed the autoload modification line!

cissyxyu:
> I think this would be a good opportunity to move away from the `lib/modules` pattern by introducing a new object under `app/models` called `AdminQuestionDashboard` with one public method `health`. This would do a few things:
> 
> 1. Adhere to single responsibility as well as encapsulation. (adding to ResponseAggregator violates both)
> 2. Cut down on parameter passing
> 3. Give more natural specs.
> 
> I'm happy to pair on this.

Changed the module pattern to be what you said -- made a new model in app/models!

happythenewsad (approved):

brendanshean (approved):

PR #7482, opened 2021-04-16
CMS Rematch All performance adjustments

dandrabik:
## WHAT
A process called `Rematch All` is used by the Curriculum Team to update the feedback used by our tools (Connect, Grammar, etc.). It has had performance issues, so we limited it to smaller datasets. This PR aims to fix some of the performance issues so the team can update the feedback.
## WHY
It's important that our students receive quality feedback and that the Curriculum can monitor an update that feedback.
## HOW
I've added some query optimizations, but mainly this will be achieved by reducing the number of sidekiq workers.

While reviewing this PR, it's helpful to know that 
- 1  `RematchResponsesForQuestionWorker` job can create 
- 100K+ `RematchResponseWorker` jobs. 

That drives a lot of the decisions.

Changes:
1. Fix sidekiq configs (earlier PR) and reduce the number of sidekiq workers as a throttle (will be set in an `ENV` var)
2. Put rematch jobs in the LOW queue.
3. Use `.find_each` for the unmatched responses, to find them in groups of 1000
4. Use read replica for larger queries.
5. Query LMS question info once in the parent job and pass it to all child jobs

Will be testing this for a bit before turning it on fully for admins.

PR Checklist | Your Answer
------------ | -------------


services/QuillCMS/app/workers/rematch_responses_for_question_worker.rb
```diff
@@ -3,32 +3,64 @@
...
             .where(parent_id: nil)
             .where(parent_uid: nil)
             .where(optimal: nil)
+            .select(:id)
   end
 
-  def get_machine_graded_responses(question_uid)
+  def machine_graded_responses(question_uid)
     Response.where(question_uid: question_uid)
             .where("responses.parent_id IS NOT NULL OR responses.parent_uid IS NOT NULL")
```
brendanshean: The following is more just a question than a suggestion.  I don't particularly like the look of 'or' queries with ActiveRecord:

`Response.where.not(parent_id: nil).or(Response.where.not(parent_uid: nil)` but is the following query equivalent (and more importantly) more performant?

`Response.where.not(parent_id: nil, parent_uid: nil)`

brendanshean (approved):
Looks good to me.  The `find_each` batching in this PR reminds me of the setting `config.active_record.warn_on_records_fetched_greater_than = 1500` which I read in this [post](https://gist.github.com/absyah/3bf0f634b30adce8badb6426a37e5a8a#file-a-helpful-activerecord-setting-only-1-person-has-ever-used).  Not sure if that would be useful elsewhere.

services/QuillCMS/app/workers/rematch_responses_for_question_worker.rb
dandrabik: Yeah, good catch and refactor, classic [De Morgan's Law](https://en.wikipedia.org/wiki/De_Morgan%27s_laws). I'll update.

cissyxyu (approved):
Changes look good!

PR #7486, opened 2021-04-16
Comprehension internal tools/ Feedback History Views

eadams17:
## WHAT
add the views for viewing activity sessions and the feedback received

## WHY
so that Curriculum can start analyzing this data

## HOW
added several new components 
### Screenshots
Screen Shot 2021-04-16 at 5 32 53 PM
Screen Shot 2021-04-16 at 5 33 13 PM
Screen Shot 2021-04-16 at 5 33 24 PM
Screen Shot 2021-04-16 at 5 33 48 PM

### Notion Card Links
https://www.notion.so/quill/Feedback-History-Session-Views-Front-End-7660dd3f6124455eb6ecc4ad4164bde5

PR Checklist | Your Answer
------------ | -------------


services/QuillLMS/client/app/bundles/Staff/components/comprehension/activitySessions/promptTable.tsx
```diff
@@ -0,0 +1,119 @@
...
+  showHeader?: boolean;
+}
+const PromptTable = ({ activity, prompt, showHeader }: PromptTableProps) => {
+
+
+  function formatFirstTableData(prompt: any) {
+    const { attempts } = prompt;
+    const keys = attempts && Object.keys(attempts);
+    let completed: boolean;
+    if(keys.length === 5) {
```
emilia-friedberg: We should probably make the `5` here into a const so it's easier to find if we change the number of attempts at some point in the future.

emilia-friedberg: Also, even though you're nullchecking `attempts` above, if `keys` is null this line is going to error.

services/QuillLMS/client/app/bundles/Staff/components/comprehension/activitySessions/promptTable.tsx
```diff
@@ -0,0 +1,119 @@
...
+    } else {
+      const usedAttempts = [];
+      keys.forEach(key => {
+        attempts[key].forEach((attempt: any) => {
+          if(attempt.used) {
+            usedAttempts.push(attempt);
+          }
+        });
+      });
+      completed = !!usedAttempts.some((attempt) => attempt.optimal);
```
emilia-friedberg: `.some` returns a boolean, so you don't need the `!!`.

services/QuillLMS/client/app/bundles/Staff/components/comprehension/activitySessions/promptTable.tsx
```diff
@@ -0,0 +1,119 @@
...
+      attemptsValue: keys.length,
+      completedLabel: 'Completed',
+      completedValue: completed ? 'True' : 'False'
+    }
+  }
+
+  function formatFeedbackData(prompt: any) {
+    const { attempts, prompt_id } = prompt;
+    const { prompts } = activity;
+    const matchedPrompt = prompts.filter(prompt => prompt.id === prompt_id)[0];
```
emilia-friedberg: We should name the inner variable here (`prompt`) something other than the function argument `prompt`.

services/QuillLMS/client/app/bundles/Staff/components/comprehension/activitySessions/promptTable.tsx
```diff
@@ -0,0 +1,119 @@
...
+      const feedbackObject: any = {};
+      attemptObject.status = attemptLabel;
+      attemptObject.results = (<div>
+        <b>{matchedPrompt && matchedPrompt.text}</b>
+        <p className="entry">{entry}</p>
+      </div>);
+      feedbackObject.status = feedbackLabel;
+      feedbackObject.results = stripHtml(feedback_text);
+      feedbackObject.feedback = feedback_type;
+      feedbackObject.className = optimal ? 'optimal' : 'sub-optimal'
```
emilia-friedberg: I think it might be clearer to define the objects like this:

```
const attemptObject = {
    status: attemptLabel,
    results: ...
}

const feedbackObject = {
   ...
}
```

Rather than defining them as empty and then immediately adding attributes.

services/QuillLMS/client/app/bundles/Staff/components/comprehension/activitySessions/sessionOverview.tsx
```diff
@@ -0,0 +1,87 @@
...
+import { DataTable, Error, Spinner } from '../../../../Shared/index';
+import { getPromptForActivitySession } from "../../../helpers/comprehension";
+import { BECAUSE, BUT, SO } from '../../../../../constants/comprehension';
+
+const SessionOverview = ({ activity, sessionData }) => {
+
+  function sessionRows({ activitySession }) {
+    if(!activitySession) {
+      return [];
+    } else {
```
emilia-friedberg: Don't need this else since you're returning from the `if`.

services/QuillLMS/client/app/bundles/Staff/components/comprehension/activitySessions/sessionView.tsx
```diff
@@ -0,0 +1,93 @@
...
+      <div className="loading-spinner-container">
+        <Spinner />
+      </div>
+    );
+  }
+
+  if(sessionData && sessionData.error) {
+    return(
+      <div className="error-container">
+        <Error error={`${sessionData.error}`} />
```
emilia-friedberg: probably can just do `error={sessionData.error}` since there's nothing else in the string.

services/QuillLMS/client/app/bundles/Staff/components/comprehension/activitySessions/sessionsIndex.tsx
```diff
@@ -0,0 +1,140 @@
...
+
+  // cache activity data for updates
+  const { data: activityData } = useQuery({
+    queryKey: [`activity-${activityId}`, activityId],
+    queryFn: fetchActivity
+  });
+
+  // cache activity sessions data for updates
+  const { data: sessionsData } = useQuery({
+    queryKey: [`activity-${activityId}-sessions`, activityId, pageNumberForQuery],
```
emilia-friedberg, line 28: Doesn't the `pageNumberForQuery` also need to be part of the queryKey string if we want it to update when it changes?

services/QuillLMS/client/app/bundles/Staff/components/comprehension/activitySessions/sessionsIndex.tsx
```diff
@@ -0,0 +1,140 @@
...
+      formattedSession.id = session_uid;
+      formattedSession.date = date;
+      formattedSession.time = time;
+      formattedSession.because_attempts = because_attempts;
+      formattedSession.but_attempts = but_attempts;
+      formattedSession.so_attempts = so_attempts;
+      formattedSession.total_attempts = total;
+      formattedSession.view_link = <Link className="data-link" to={`/activities/${activityId}/activity-sessions/${session_uid}/overview`}>View</Link>;
+      formattedSession.completed = complete ? <img alt="quill-circle-checkmark" src={quillCheckmark} /> : "";
+      return formattedSession;
```
emilia-friedberg: This one is less confusing because you're only defining the one object, but I would probably still format this as a literal.

services/QuillLMS/client/app/bundles/Staff/helpers/comprehension.tsx
```diff
@@ -207,6 +200,23 @@ export function getActivityPromptSetter({
...
+  const { activitySession } = sessionData;
+  const { prompts } = activitySession;
+  if(prompts[conjunction]) {
+    const prompt = prompts[conjunction];
+    prompt.conjunction = conjunction;
+    return prompt;
+  }
+  return {
+    conjunction,
+    text: 'none'
```
emilia-friedberg, line 216: Why do we want to use the string `none` here? Wouldn't it work just as well to leave it undefined? I'm asking because I think it's easier to remember to check for the lack of presence of something rather than that it matches a special string that means "nothing". If we do need to use the string for some reason, we should constantize it in a shared file and import it in the places we use it.

emilia-friedberg (commented):
Looks great overall! Left a couple of stylistic comments and one or two places that I think might lead to a small bug, but I am excited to see this go out.

services/QuillLMS/client/app/bundles/Staff/components/comprehension/activitySessions/promptTable.tsx
eadams17: good point, I already have a const for this

services/QuillLMS/client/app/bundles/Staff/components/comprehension/activitySessions/promptTable.tsx
eadams17: updating!

services/QuillLMS/client/app/bundles/Staff/components/comprehension/activitySessions/sessionsIndex.tsx
eadams17: no, this query-string will be unique to the sessions data of this activity as is. If we wanted to manually refetch the query for whatever reason, we could do so like this:
```
queryClient.refetchQueries([`activity-${activityId}-sessions`, activityId,  pageNumberForQuery]) 
```

services/QuillLMS/client/app/bundles/Staff/components/comprehension/activitySessions/sessionsIndex.tsx
eadams17: yeah not sure why I did it like this, I think it's because I was checking at each step-- updating

services/QuillLMS/client/app/bundles/Staff/helpers/comprehension.tsx
eadams17: so the reason is because of the way we load the prompt in the `promptTable` component. While waiting for the data, we check for `!prompt` and render a loading spinner. There are cases where the activity session data will load, but  there may not be responses for all of the prompts, so I chose to do this to easily render in the UI like so:

```
  if(prompt && prompt.text === 'none') {
    const { conjunction, text } = prompt;
    return(
      <section className="prompt-table-container">
        <section className="attempts-section">
          <p className="attempts-label">{PROMPT_HEADER_LABELS[conjunction]}</p>
          <p className="attempts-value">{text}</p>
        </section>
      </section>
    );
  }
```

cissyxyu (approved):

emilia-friedberg (approved):
LGTM! Thanks for making those changes so fast!

PR #7492, opened 2021-04-20
comprehension/tweak-go-used-calculation

anathomical:
## WHAT
Use a deterministically determined loop order
## WHY
The return logic assumes that the loop will look over items in priority order (and thus early exit on the highest priority item that matches the return criteria), but the loop in place at the moment does NOT actually do this because `range feedbacks` provides a non-deterministic iteration order.
## HOW
Because items are placed into the map keyed on their priority, it turns out to be the case that the range of `int`s equal to the `len` of the map will give us every item in order of priority.

### Notion Card Links
https://www.notion.so/quill/Bug-Report-User-Feedback-does-not-Match-Report-Feedback-fc313a2a44ff4cb0a1adf20015077cca

PR Checklist | Your Answer
------------ | -------------
Have you added and/or updated tests? |  No behavior changes that can be tested reproducibly

### DIFF
--- a/services/comprehension/feedback-api-main/src/endpoint/endpoint.go
+++ b/services/comprehension/feedback-api-main/src/endpoint/endpoint.go
@@ -170,9 +170,10 @@ func getAPIResponse(url string, priority int, json_params [] byte, c chan Intern
 }
 
 func identifyUsedFeedbackIndex(feedbacks map[int]InternalAPIResponse) int {
-	for key, feedback := range feedbacks {
+	for i := 0; i < len(feedbacks); i++ {
+		feedback := feedbacks[i]
 		if !feedback.Error && !feedback.APIResponse.Optimal {
-			return key
+			return i
 		}
 	}
 	// If none of the feedbacks are non-optimal, check to see if automl is feedback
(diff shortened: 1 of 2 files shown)


dandrabik (changes requested):
Thanks for finding this bug. Let's add tests for this. This is a pure function, so should be easily testable (a test on this function could have caught the original bug).

happythenewsad (changes requested):
map order cannot be relied upon in golang, according to the spec: https://golang.org/ref/spec#RangeClause

happythenewsad:
> map order cannot be relied upon in golang, according to the spec: https://golang.org/ref/spec#RangeClause

Disregard, you're not relying on order

anathomical:
@dandrabik There actually was a test case for this function already, but it didn't capture the edge case that could give rise to this intermittent failure.  I've updated the case with data that should cover this failure state and run it a couple of times successfully, so hopefully it would catch intermittent errors of this class if they come up in the future.

dandrabik (approved):
:+1: LGTM. Thanks for adding/improving the test!

happythenewsad (approved):

PR #7493, opened 2021-04-20
Comprehension/save concept results

anathomical:
## WHAT
Modify Comprehension to post `activity_session` updates when the activity is completed.  These updates flag the session as finished and attach `ConceptResult` data to the session which is required for teachers to see reporting on the session.
## WHY
We want teachers to be able to see student reporting to understand how their students are performing on Comprehension activities
## HOW
Compile a payload similar to the one produced by Connect when a Comprehension activity completes, and send that payload to the `activity_session` completion API.  This API includes code that creates `concept_result` records when they are included as they are here.

NOTE: Concept results for feedback that doesn't have an attached ConceptUID will fail.  (We have in-progress plans to make sure that these get added where they're missing such as in the Opinion API.)

### Notion Card Links
https://www.notion.so/quill/Recording-ConceptResults-and-ActivitySessions-62c39ad1338e49d48210fb1d56201599

PR Checklist | Your Answer
------------ | -------------


services/QuillLMS/client/app/bundles/Comprehension/components/studentView/container.tsx
```diff
@@ -106,6 +114,59 @@ export class StudentViewContainer extends React.Component<StudentViewContainerPr
     });
   }
 
+  defaultHandleFinishActivity = () => {
+    this.specifiedActivitySessionUID();
```
emilia-friedberg: you aren't saving the output of this call in a variable so I think it can be removed

services/QuillLMS/client/app/bundles/Comprehension/components/studentView/container.tsx
```diff
@@ -106,6 +114,59 @@ export class StudentViewContainer extends React.Component<StudentViewContainerPr
...
+  }
+
+  calculatePercentage = () => {
+    const { session, } = this.props
+    const { submittedResponses, } = session
+    const attemptCounts = Object.values(submittedResponses).map((responses) => ATTEMPTS_TO_SCORE[responses.length])
+    return attemptCounts.reduce((total, value) => total + value) / attemptCounts.length
+  }
+
+  generateConceptResults = () => {
```
emilia-friedberg: I leave this up to your discretion, but I feel like these might be better as lib functions with the arguments passed in than as part of the component. That would make them easier to test, and better isolate them from the rendering logic. That also follows the pattern we have for similar functions in the other apps.

services/QuillLMS/client/app/bundles/Comprehension/components/studentView/container.tsx
```diff
@@ -106,6 +114,59 @@ export class StudentViewContainer extends React.Component<StudentViewContainerPr
...
+      responses.forEach((response, index) => {
+        const attempt = index + 1
+        conceptResults.push({
+          concept_uid: response.concept_uid,
+          question_type: 'comprehension',
+          metadata: {
+            answer: response.entry,
+            attemptNumber: attempt,
+            correct: response.optimal,
+            directions: 'Complete this sentence',
```
emilia-friedberg: We should probably constantize this.

emilia-friedberg (approved):
A couple of small suggestions, but you can choose to take or leave them. I think it might be good to get more testing in here, but I know we're on a time crunch.

eadams17 (approved):

services/QuillLMS/client/app/bundles/Comprehension/components/studentView/container.tsx
anathomical: Good catch.

services/QuillLMS/client/app/bundles/Comprehension/components/studentView/container.tsx
anathomical: Good call.  I've pulled this code into a `libs` directory and added a couple of very basic tests.

services/QuillLMS/client/app/bundles/Comprehension/components/studentView/container.tsx
anathomical: Yeah, good call.

PR #7495, opened 2021-04-21
Lms/grammar rule data migration

happythenewsad:
## WHAT
a data migration for grammar rules

## WHY
As a developer, I want grammar feedback available in the activity report, so that I can analyze
(the Rules Report tool implicitly filters feedbacks not associated with Rules, via JOIN statements)

## HOW
rake task + csv file 

### Screenshots

### Notion Card Links
https://www.notion.so/quill/Enter-Grammar-Rules-into-the-comprehension_rules-DB-653daceea79a4297925d30d54b08cd46

PR Checklist | Your Answer
------------ | -------------
Have you added and/or updated tests? |  no - simple rake task


services/QuillLMS/lib/tasks/add_grammar_api_rules_and_feedback.rake
```diff
@@ -0,0 +1,31 @@
...
+        ActiveRecord::Base.transaction do 
+            valid_grammar_rules.each do |r|
+                created_rule = Comprehension::Rule.create!(
+                    uid: r['Rule UID'],
+                    name: r['Rule'],
+                    description: r['Rule Description'],
+                    universal: true,
+                    rule_type: 'grammar',
+                    optimal: false,
+                    suborder: 1,
```
dandrabik: I believe we'd want the suborder to be the precedence from the grammar API, rather than a hard-coded value. (It's also zero-based).

services/QuillLMS/lib/tasks/add_grammar_api_rules_and_feedback.rake
```diff
@@ -0,0 +1,31 @@
...
+                    state: 'active',
+                )
+                Comprehension::Feedback.create!(
+                    rule_id: created_rule.id,
+                    text: 'lorem ipsum',
+                    order: 1
```
dandrabik: Order is zero-based, so these should all be `0`.

services/QuillLMS/lib/data/rules.csv
```diff
@@ -0,0 +1,115 @@
...
+59,Passive perfect without have,Grammar API,SpaCy Parts of Speech Rule,,,Off,,,0,No,No,,No,No,,,No,Verbs | Perfect Tense | Passive Perfect Tense,Off,,Verbs,Perfect Tense,Passive Perfect Tense,,,,,
+,Passive with simple past instead of participle,Grammar API,Grammar ML Model,,,Off,,,0,No,No,,No,No,,,No,Verbs | Active & Passive Voice | Passive Voice,Off,,Verbs,Active & Passive Voice,Passive Voice,,,,,
+,Passive with incorrect be,Grammar API,Grammar ML Model,,,Off,,,0,No,No,,No,No,,,No,Verbs | Active & Passive Voice | Passive Voice,,,Verbs,Active & Passive Voice,Passive Voice,,,,,
+,Past instead of participle,Grammar API,Grammar ML Model,,,Off,,,0,No,No,,No,No,,,No,Verbs | Perfect Tense | Perfect Tense,"This would be making an assumption that the participle was part of a perfect tense rather than an incorrectly formed passive voice verb. I think it's equally likely students will be writing passive voice as they will be perfect, so we may want to re-think this rule to be more specific. It will be difficult to write specific feedback for it as is.
+Check back in on this when Yves updates the config file.
```
dandrabik: This line seems like it needs to be removed.

services/QuillLMS/lib/tasks/add_grammar_api_rules_and_feedback.rake
anathomical: This isn't super critical, but I think we actually want to put something here if we have it.  It's not dead-on, but I suggest using `Overall Rule Priority`.  Alternatively, since it looks like the CSV is in priority order, maybe we could just use the `index + 1` of where we are in the `each` call.

services/QuillLMS/lib/tasks/add_grammar_api_rules_and_feedback.rake
```diff
@@ -0,0 +1,31 @@
...
+                    description: r['Rule Description'],
+                    universal: true,
+                    rule_type: 'grammar',
+                    optimal: false,
+                    suborder: 1,
+                    state: 'active',
+                )
+                Comprehension::Feedback.create!(
+                    rule_id: created_rule.id,
+                    text: 'lorem ipsum',
```
anathomical: It doesn't super matter, but I think we probably want to populate this with `Feedback - Revised`

services/QuillLMS/lib/tasks/add_grammar_api_rules_and_feedback.rake
```diff
@@ -0,0 +1,31 @@
+namespace :grammar_api_rules_and_feedback do
+    desc 'data migration for grammar-api specific rules and feedback'
+    task :insert => :environment do
+        rules_csv = CSV.parse(File.read('lib/data/rules.csv'), headers: true)
+        valid_grammar_rules = rules_csv.select do |r| 
+            r['Module'] == 'Grammar API' && r['Rule UID'].present? 
+        end
+
+        ActiveRecord::Base.transaction do 
+            valid_grammar_rules.each do |r|
```
anathomical: I think we also want a loop that creates the `opinion` rules as well.

anathomical (changes requested):
A couple of tweaks I think are worth making, but generally speaking this makes sense to me.

services/QuillLMS/lib/data/rules.csv
```diff
@@ -0,0 +1,115 @@
...
+24,Post-Topic Regex (Regex 1),Regex Module,,Rule,"Passage-specific rules that identify incorrect sequences in the responses. These rules may apply across all three conjunctions in a prompt, or they may apply to a single conjunction for a passage.",Live Algorithm,67e8dff7-93f6-4cf4-a601-9a8adad242a5,Individual feedback based on regex rule,1,Yes,No,No,Yes,No,Custom for Prompt,No,Yes,Individual concepts for each regex rule,,,,,,Check your work. Use a or an before singular words only.,6,,,
+25,Grammar API,Grammar API,,API,API,On Production,a55848fb-1138-4f58-8b04-36657d32b96a,,1,No,Yes,,,No,Universal: all prompts/passages,No,,,,,,,,,7,,,
```
dandrabik: This looks like a generic 'Grammar API rule', should this be here?

services/QuillLMS/lib/data/rules.csv
```diff
@@ -0,0 +1,115 @@
+﻿Overall Rule Priority,Rule,Module,Algorithm,Rule or API?,Rule Description,Status,Rule UID,Feedback - Revised,Feedback Layers,Unique feedback for each activity?,Give feedback for Turk workers?,Bolding?,Bolding: Is this what the engine returns now?,Text highlighting?,Application,Spelling corrected in background?,Reported to teachers?,Combined Concept,Notes,Notes for Feedback,Level 2 Concept,Level 1 Concept,Level 0 Concept,Feedback - Original,API Priority,Rule Priority - Opinion,Uses SpaCy?,Status Notes
```
dandrabik: Where did this CSV come from? Is this an export from Notion? There's a lot of non-grammar rule cruft in here.

services/QuillLMS/lib/tasks/add_grammar_api_rules_and_feedback.rake
dandrabik: Agreed that we'd want to populate this with the current feedback.

services/QuillLMS/lib/tasks/add_grammar_api_rules_and_feedback.rake
```diff
@@ -0,0 +1,31 @@
+namespace :grammar_api_rules_and_feedback do
```
dandrabik: I believe these files (csv and rake task) should live in the Comprehension Engine?

services/QuillLMS/lib/data/rules.csv
```diff
@@ -0,0 +1,115 @@
...
+3,"Responses that are too short ",Pre-filter,Responses that are too short,Rule,Response is fewer than 3 words,Live Algorithm,76f519bd-ad7a-4b88-8ee6-c00a877e2ccc,Revise your work. Use the text to add more information to your response.,1,No,No,No,Yes,No,Universal: all prompts/passages,No,No,Activity Guidelines | Comprehension Activity Guidelines | Response Meets Length Requirements,"Old version: Response Too Short
+Front-end only for now. Convert to back-end in the future.",,Activity Guidelines,Comprehension Activity Guidelines,Response Meets Length Requirements,"Revise your work. When writing your response, make sure to use appropriate language.",,,,
```
dandrabik: Does this line have unbalanced quotes? Is that going to cause import/parsing issues?

dandrabik (changes requested):
There are a few items that need to be cleaned up here, e.g. we should be populating the order and feedback.

Also, can you take a moment and fill out the PR description more. Please fill out the 'Why' as a communication tool to other developers. Also, the PR description is a good place to mention: Where did this `csv` come from? Why does it have so many non-grammar rules in it? Why did you decide to go with a placeholder  `1` for `suborder`? Why did you decide to use `'lorem ipsum'` instead of the feedback from the `csv`? etc.

Also, I'm a little worried about the approach here of starting out with an export file with a lot of cruft (which I think is from Notion) and then filtering it down in code. I assuming you took this approach to make it easy for the Curriculum Team. It might be worth maintain a google sheet with just the columns that are being imported. From a reviewer's perspective, it's hard to parse what will end up in the production DB. I flagged one rule which seems like a universal `Grammar API` Rule.  Is that supposed to be copied to the DB? Another line had unbalanced quotes, which also might cause parsing issues.

services/QuillLMS/lib/tasks/add_grammar_api_rules_and_feedback.rake
```diff
@@ -0,0 +1,44 @@
...
+                    description: r['Rule Description'],
+                    universal: true,
+                    rule_type: r['Module'] == 'Grammar API' ? 'grammar' : 'opinion',
+                    optimal: false,
+                    suborder: r['Overall Rule Priority'],
+                    state: 'active',
+                }
+                created_rule.save!
+
+                # rubocop:disable all
```
happythenewsad: Rubocop throws Layout/AssignmentIndentation: Indent the first line of the right-hand-side of a multi-line assignment.

even though the line seems indented correctly

happythenewsad:
FYI @anathomical 
Dan and I just synced on this. We're holding off further review of this PR until I meet with Lindsey tomorrow to establish a plan for keeping the input data clean and up to date.

services/QuillLMS/lib/tasks/add_grammar_api_rules_and_feedback.rake
happythenewsad: Hannah created a new column for this today, which I am now using.

services/QuillLMS/lib/tasks/add_grammar_api_rules_and_feedback.rake
happythenewsad: Hannah explicitly added correct values in the suborder col today.

happythenewsad:
The CSV file should be clean now. `Rule.suborder` is now extracted directly from the CSV. `Feedback.order` is set to `1` for all feedbacks created from this task. Also, the task is now idempotent.

services/QuillLMS/lib/data/rules.csv
happythenewsad: All rows were parsed correctly in the latest CSV (which is fresher than this comment)

services/QuillLMS/engines/comprehension/lib/tasks/add_grammar_api_rules_and_feedback.rake
```diff
@@ -0,0 +1,35 @@
...
+                    rule_type: r['Module'] == 'Grammar API' ? 'grammar' : 'opinion',
+                    optimal: false,
+                    suborder: r['Rule Suborder'],
+                    state: 'active',
+                }
+                created_rule.save!
+
+                feedback = Comprehension::Feedback.find_or_initialize_by(
+                    rule_id: created_rule.id,
+                    order: 1,
```
dandrabik: This should be 0, since this is zero-based.

services/QuillLMS/engines/comprehension/lib/tasks/add_grammar_api_rules_and_feedback.rake
```diff
@@ -0,0 +1,35 @@
...
+        ActiveRecord::Base.transaction do 
+            valid_rules.each do |r|
+                created_rule = Comprehension::Rule.find_or_initialize_by(uid: r['Rule UID'])
+                created_rule.attributes = {
+                    name: r['Rule'],
+                    description: r['Rule Description'],
+                    universal: true,
+                    rule_type: r['Module'] == 'Grammar API' ? 'grammar' : 'opinion',
+                    optimal: false,
+                    suborder: r['Rule Suborder'],
```
dandrabik, line 18: Flagging that the order is zero-based. I think the orders in the csv start at 1.

dandrabik (approved):
Thanks for making these changes!

So approving if a small change is made: changing the order and suborder to zero-based (instead of starting at 1).

services/QuillLMS/engines/comprehension/lib/tasks/rules.csv
```diff
@@ -0,0 +1,78 @@
+Rule Suborder,Overall Rule Priority,Rule,Module,Rule Description,Status,Rule UID,Feedback V3 [HM 4/21],Feedback - V2,Feedback Layers,Unique feedback for each activity?,Bolding?,Bolding: Is this what the engine returns now?,Text highlighting?,Application,Spelling corrected in background?,Reported to teachers?,Combined Concept,Notes,Notes for Feedback,Level 2 Concept,Level 1 Concept,Level 0 Concept,Rule Priority - Opinion
+1,6,First-Person Opinionated Phrase Keyword Check,Opinion API,Checks if a response contains a first-person opinionated phrase based on the list in the card,On Production,5ed4c961-60e6-4c29-9ace-e80c6492b377,"Revise your work. The bolded phrase expresses an opinion. Instead, focus your response on ideas your read in the text.","Revise your work. The bolded phrase expresses an opinion. Instead, focus your response on ideas your read in the text.",1,No,Bold the opinionated phrase,Yes - Yves is confirming,No,Universal: all prompts/passages,No,Yes,Academic Writing | Using Academic Tone | Avoiding First-Person to Maintain an Objective Tone,,,Academic Writing,Using Academic Tone,Avoiding First-Person to Maintain an Objective Tone,3
```
dandrabik: The suborder seems to start at 1, and should start at zero.

anathomical (approved):

PR #7500, opened 2021-04-21
Lms/flag repeated feedback history

anathomical:
## WHAT
- Add a new `FeedbackHistoryFlag` table that can be used to flag specific patterns on `FeedbackHistory` rows after they've already been saved (we don't want to modify the rows themselves after write)
- Set up a Sidekiq worker that runs after a `FeedbackHistory` record is created to calculate and create appropriate `FeedbackHistoryFlag` entries
- Add code to `FeedbackHistory` so that instances can determine if they are examples of repeated rule violations within their session
- Add new `flags` value to the SessionFeedbackHistory roll-up that is a list of each flag that any component `FeedbackHistory` record have
## WHY
We want to be able to call special attention to FeedbackSessions which have specific traits.  In this case, we know we want to pay attention to cases where a student violates the same rule twice since it suggests a case where the feedback provided to the student is not sufficient to help them figure out the next step to take with their writing.
## HOW
This basically establishes a framework in which `FeedbackHistory` records have a post-processing step after they've been created that calculates which, if any, flags should be assigned to it.  This way if we identify other features that we want to flag which require post-processing, it should be easy to add them.

### Notion Card Links
https://www.notion.so/quill/Repeated-Feedback-a5d7ded99fda4541ac7df2a695dc62a2

PR Checklist | Your Answer
------------ | -------------


brendanshean (commented):
I'm retracting my original comment on enumerated types for the time being.

services/QuillLMS/db/migrate/20210421181107_create_feedback_history_flags.rb
```diff
@@ -0,0 +1,9 @@
+class CreateFeedbackHistoryFlags < ActiveRecord::Migration
+  def change
+    create_table :feedback_history_flags do |t|
+      t.references :feedback_history, null: false
+      t.string :flag, null: false
```
happythenewsad, line 5: how about making this jsonb for extensibility?

happythenewsad (commented):
Code quality looks good, but I'm concerned that we're adding so much complexity and schema changes for what's essentially an enum. Can we store the flag in FeedbackHistory.metadata rather than a new table?

I agree that the calculation is expensive and thus needs to be async, in a worker.

Notion says that Thomas and Dan discussed various implementation options. Would you mind including your thought process here, for the rest of us?

services/QuillLMS/db/migrate/20210421181107_create_feedback_history_flags.rb
anathomical: I'm not convinced that this has immediate value, which leads me to think we should hold off on doing this unless/until we see a reason to.  Also, some of the query code currently relies on this being a string, and while it wouldn't be that hard to modify, it doesn't seem like there's a good reason to yet.

anathomical:
> Code quality looks good, but I'm concerned that we're adding so much complexity and schema changes for what's essentially an enum. Can we store the flag in FeedbackHistory.metadata rather than a new table?
> 
> I agree that the calculation is expensive and thus needs to be async, in a worker.
> 
> Notion says that Thomas and Dan discussed various implementation options. Would you mind including your thought process here, for the rest of us?

Good suggestion.  One of the big values that we are trying to preserve is the idea of `FeedbackHistory` rows as immutable.  That means that we're trying to avoid modifying the data on a row after it's been written.  Which is one reason that the "strong/weak" distinction wasn't just added to the table.  Given that we're trying to avoid modifying written `FeedbackHistory` rows, any async stuff needs to write to different tables.  That doesn't mean this is the best approach, and I'm totally open to additional feedback here, but hopefully that provides some context for why we're looking at this approach.

happythenewsad:
@anathomical thanks for the background color; that's very helpful. I'm close to approving this PR, after a second pass. I wasn't aware that we want to keep FeedbackHistory readonly. To that end, can you have the code internally document this? There are a couple techniques for earmarking AR models as read only. [Here's one](https://til.hashrocket.com/posts/7cd7e55929-read-only-modelsl). It'd be great to include this change in this PR if it's trivial. If it's not trivial, let's put it in the sprint backlog.

anathomical:
@happythenewsad Enforcing the write-once intention in code is a good call.  I've made some changes based on your suggestions and updated tests.

happythenewsad (approved):
Great, thanks for going the extra mile in removing the non-ready-only actions and specs

PR #7516, opened 2021-04-23
Rematch All Button - show for all, disable on click.

dandrabik:
## WHAT
Remove the limit on the Rematch All button, disable it on click.
## WHY
We're opening Rematch All for all questions, so removing the limit. Also, because there is now the button is very easy to double click or more accurately, click, not sure if something happened, click again (it happened a few times in testing) so disabling it on click.
## HOW
If there is a better way to do this, please let me know. It might be better to have an alert/confirmation box.

But I set a state parameter on click to disable the button.

PR Checklist | Your Answer
------------ | -------------
Have you added and/or updated tests? |  'NO', small frontend change.
Have you deployed to Staging? | deploying now!

### DIFF
--- a/services/QuillLMS/client/app/bundles/Connect/components/questions/responseComponent.jsx
+++ b/services/QuillLMS/client/app/bundles/Connect/components/questions/responseComponent.jsx
@@ -52,6 +52,7 @@ class ResponseComponent extends React.Component {
       selectedResponses: [],
       health: {},
       gradeBreakdown: {},
+      enableRematchAllButton: true,
     };
   }
 
@@ -439,11 +440,11 @@ class ResponseComponent extends React.Component {
 
   renderRematchAllButton = () => {
     const { filters } = this.props
-    let disabled = filters.numberOfResponses > 1000
+
     if (this.props.admin) {
       const text = this.state.progress ? `${this.state.progress}%` : 'Rematch Responses';
 
-      return (<button className="button is-outlined is-danger" disabled={disabled} onClick={this.rematchAllResponses} style={{ float: 'right', }} type="button">{text}</button>);
+      return (<button className="button is-outlined is-danger" disabled={!this.state.enableRematchAllButton} onClick={this.setState({enableRematchAllButton: false}) && this.rematchAllResponses} style={{ float: 'right', }} type="button">{text}</button>);
     }
   };
 
--- a/services/QuillLMS/client/app/bundles/Connect/components/questions/responseComponent.jsx
+++ b/services/QuillLMS/client/app/bundles/Connect/components/questions/responseComponent.jsx
@@ -276,6 +276,7 @@ class ResponseComponent extends React.Component {
   };
 
   rematchAllResponses = () => {
+    this.setState({enableRematchAllButton: false});
     const pageNumber = 1;
     const callback = (done) => {
       if (done) {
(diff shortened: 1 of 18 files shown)


services/QuillLMS/client/app/bundles/Connect/components/questions/responseComponent.jsx
```diff
@@ -439,11 +440,11 @@ class ResponseComponent extends React.Component {
 
   renderRematchAllButton = () => {
     const { filters } = this.props
-    let disabled = filters.numberOfResponses > 1000
+
     if (this.props.admin) {
       const text = this.state.progress ? `${this.state.progress}%` : 'Rematch Responses';
 
-      return (<button className="button is-outlined is-danger" disabled={disabled} onClick={this.rematchAllResponses} style={{ float: 'right', }} type="button">{text}</button>);
+      return (<button className="button is-outlined is-danger" disabled={!this.state.enableRematchAllButton} onClick={this.setState({enableRematchAllButton: false}) && this.rematchAllResponses} style={{ float: 'right', }} type="button">{text}</button>);
```
emilia-friedberg: I would just put `this.setState({ enableRematchAllButton: false })` as the first line in `rematchAllResponses`, since this is the only place that gets called anyway.

emilia-friedberg (approved):
One stylistic thing I think we should change (I only left the comment in one place but it applies in all three), but otherwise LGTM.

cissyxyu (commented):
should we just remove this field `disabled` all together and just have all these buttons be enabled by default without a switch? it seems like we now want the buttons to all be clickable without limits.

anathomical (approved):

services/QuillLMS/client/app/bundles/Connect/components/questions/responseComponent.jsx
dandrabik: That's fair, that'll look a little cleaner. Thanks for the suggestion, will do!

dandrabik:
> should we just remove this field `disabled` all together and just have all these buttons be enabled by default without a switch? it seems like we now want the buttons to all be clickable without limits.

The goal of the button disabling is to prevent someone from clicking it multiple times (See PR description) since there isn't feedback. The button is enabled by default, but you'd really only want to click it once.

cissyxyu:
> > should we just remove this field `disabled` all together and just have all these buttons be enabled by default without a switch? it seems like we now want the buttons to all be clickable without limits.
> 
> The goal of the button disabling is to prevent someone from clicking it multiple times (See PR description) since there isn't feedback. The button is enabled by default, but you'd really only want to click it once.

ah gotcha, I didn't understand it clearly on the first read-through. LGTM.

cissyxyu (approved):

services/QuillLMS/client/app/bundles/Connect/components/questions/responseComponent.jsx
```diff
@@ -439,11 +441,11 @@ class ResponseComponent extends React.Component {
 
   renderRematchAllButton = () => {
     const { filters } = this.props
-    let disabled = filters.numberOfResponses > 1000
+
     if (this.props.admin) {
       const text = this.state.progress ? `${this.state.progress}%` : 'Rematch Responses';
 
-      return (<button className="button is-outlined is-danger" disabled={disabled} onClick={this.rematchAllResponses} style={{ float: 'right', }} type="button">{text}</button>);
+      return (<button className="button is-outlined is-danger" disabled={!this.state.enableRematchAllButton} onClick={this.rematchAllResponses} style={{ float: 'right', }} type="button">{text}</button>);
```
eadams17: so as part of our linters, we deconstruct state and prop properties so we'd want to make this:
```
const { filters, admin } = this.props;
const { progress, enableRematchAllButton } =  this.state;
```

eadams17 (approved):

services/QuillLMS/client/app/bundles/Connect/components/questions/responseComponent.jsx
dandrabik: Ah, good call, will change. Ya'll are making this code so clean!

PR #7528, opened 2021-04-27
remove activity feed from dash

emilia-friedberg:
## WHAT
Remove activity feed from the dashboard.

## WHY
This query is running very slowly for a lot of users, and we want to stop calling it while we are working on a fix.

## HOW
Just remove the code that calls the query and that loads the feed.

### Screenshots

### Notion Card Links

PR Checklist | Your Answer
------------ | -------------

### DIFF
--- a/services/QuillLMS/client/app/bundles/Teacher/containers/dashboard.jsx
+++ b/services/QuillLMS/client/app/bundles/Teacher/containers/dashboard.jsx
@@ -41,11 +41,10 @@ const Dashboard = ({ onboardingChecklist, firstName, mustSeeModal, linkedToCleve
     getMetrics();
     getDiagnostics()
     getLessons()
-    getActivityFeed()
   }, []);
 
   React.useEffect(() => {
-    if (metrics && diagnostics && lessons && activityFeed && loading) {
+    if (metrics && diagnostics && lessons && loading) {
       setLoading(false)
     }
   }, [metrics, diagnostics, lessons, activityFeed])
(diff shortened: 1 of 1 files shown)


anathomical (approved):

eadams17 (approved):

services/QuillLMS/client/app/bundles/Teacher/containers/dashboard.jsx
```diff
@@ -41,11 +41,10 @@ const Dashboard = ({ onboardingChecklist, firstName, mustSeeModal, linkedToCleve
...
     getLessons()
-    getActivityFeed()
   }, []);
 
   React.useEffect(() => {
-    if (metrics && diagnostics && lessons && activityFeed && loading) {
+    if (metrics && diagnostics && lessons && loading) {
       setLoading(false)
     }
   }, [metrics, diagnostics, lessons, activityFeed])
```
dandrabik, line 50: Do you need to remove this `activityFeed`?. Also possibly on line 37?

dandrabik (approved):
I made a comment that is worth checking, but overall, looks good :+1:

services/QuillLMS/client/app/bundles/Teacher/containers/dashboard.jsx
emilia-friedberg: Nope! Here it's saying "call this function if the value of `activityFeed` (or `metrics` or `diagnostics` or `lessons`) changes", so it'll work fine for those other ones but never get called because the value of activityFeed will never change from how it was originally defined in line 37 (as null). If we were removing this permanently I would take this out, but leaving it in isn't doing anything right now.

PR #7532, opened 2021-04-28
LMS Rails 5.0 Prework - address view and helper spec failures

brendanshean:
## WHAT
There a couple of fixes here:
1.  There's an [issue](https://github.com/rspec/rspec-rails/issues/1076) with stubs in view and helper specs where partial double verification fails.  A [workaround](https://www.rubydoc.info/github/rspec/rspec-mocks/RSpec%2FMocks%2FExampleMethods:without_partial_double_verification) involves using an around hook that temporarily disables it for the offending specs.

2. The second [issue](https://github.com/rspec/rspec-rails/issues/1644) involves route_helpers no longer being available in view and helpers specs. 

## WHY
These particular issues are not present yet in Rails 4, but will manifest after upgrading to Rails 5. 

### Notion Card Links
https://www.notion.so/quill/Upgrade-LMS-to-Rails-5-0-b742cd5e1764427d8670944d50e2a3e3

PR Checklist | Your Answer
------------ | -------------
Have you added and/or updated tests? |  Yes.
Have you deployed to Staging? | No.  These are tiny changes only related to specs.
Self-Review: Have you done an initial self-review of the code below on Github? |. YES

### DIFF
--- a/services/QuillLMS/spec/support/shared/routing_url_helpers.rb
+++ b/services/QuillLMS/spec/support/shared/routing_url_helpers.rb
@@ -1,0 +1,7 @@
+RSpec.shared_context "routing url helpers" do
+  include Rails.application.routes.url_helpers
+
+  def default_url_options
+    { only_path: true }
+  end
+end
--- a/services/QuillLMS/spec/support/shared/routing_url_helpers.rb
+++ b/services/QuillLMS/spec/support/shared/routing_url_helpers.rb
@@ -1,4 +1,6 @@
-RSpec.shared_context "routing url helpers" do
+RSpec::ROUTING_URL_HELPERS = "routing url helpers".freeze
+
+RSpec.shared_context RSpec::ROUTING_URL_HELPERS do
   include Rails.application.routes.url_helpers
 
   def default_url_options
--- a/services/QuillLMS/spec/views/pages/careers.html.erb_spec.rb
+++ b/services/QuillLMS/spec/views/pages/careers.html.erb_spec.rb
@@ -1,11 +1,13 @@
 require 'rails_helper'
 
 describe "pages/careers.html.erb", type: :view do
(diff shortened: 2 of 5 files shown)


emilia-friedberg (approved):

services/QuillLMS/spec/support/shared/routing_url_helpers.rb
```diff
@@ -0,0 +1,7 @@
+RSpec.shared_context "routing url helpers" do
```
dandrabik: Not required, but since it's logic, you could consider using a constant string for this name that others can use to prevent future typos, e.g. `Rspec::ROUTING_URL_HELPERS`, `include_context Rspec::ROUTING_URL_HELPERS`

dandrabik (approved):
Small take or leave comment, but looks good to me!

services/QuillLMS/spec/support/shared/routing_url_helpers.rb
```diff
@@ -1,4 +1,6 @@
-RSpec.shared_context "routing url helpers" do
+RSpec::ROUTING_URL_HELPERS = "routing url helpers".freeze
```
brendanshean, line 1: @dandrabik Is this how/where you would define the constant?

dandrabik: Yeah, that works.

services/QuillLMS/spec/views/pages/careers.html.erb_spec.rb
```diff
@@ -1,11 +1,13 @@
 require 'rails_helper'
 
 describe "pages/careers.html.erb", type: :view do
+  include_context 'routing url helpers'
```
dandrabik, line 4: You'd want to use the constant here.

PR #7541, opened 2021-04-28
Change the way we calculate average attempts and attempts reached optimal

cissyxyu:
## WHAT
In my last PR, I mistakenly used the wrong attribute to calculate a student's attempt data on a question. I was using the `correct` field to see if they got a question correct, not realizing that a student can have `correct`: `true` on a record and still have gotten the question wrong (because `correct` refers to concepts, not questions). I changed it now to use the `questionScore` attribute, which records the number of attempts they made on a question.

## WHY
The way I calculated it before was giving me inaccurate numbers.

## HOW
Changing the field I look at to be the `questionScore` field, which records the number of attempts the student made on that question before they got it correct.

### Screenshots

### Notion Card Links
https://www.notion.so/quill/Prompt-Healths-and-Activity-Healths-Tables-3b1da19f1da44817a4a56c77f7cfbf9b

PR Checklist | Your Answer
------------ | -------------


services/QuillLMS/app/models/question_health_dashboard.rb
```diff
@@ -8,38 +8,26 @@ def initialize(activity_id, question_number, question_uid)
...
   end
 
-  private def total_students(hash)
-    hash&.sum { |h| h["number_of_students"].to_i }&.to_f || 0
-  end
-
-  private def attempts_times_students(hash)
-    hash&.map { |h| h["number_of_attempts"].to_i * h["number_of_students"].to_i} || 0
+  private def score_to_attempts(score)
+    5 - 4 * score
```
happythenewsad: can we document these integers so that they are no longer [magic numbers](https://en.wikipedia.org/wiki/Magic_number_(programming))?

services/QuillLMS/app/models/question_health_dashboard.rb
```diff
@@ -50,10 +38,8 @@ def cms_dashboard_stats
     @attempt_data ||= begin
```
happythenewsad: It looks like the intention here is to memoize the query result, but only the query itself is memoized

happythenewsad (changes requested):

services/QuillLMS/app/models/question_health_dashboard.rb
happythenewsad: Also, I'm not sure why the `begin` keyword is needed. You should be able to put the heredoc on the right side of the conditional assignment operator

happythenewsad: I misread the indentations. The memoization is correct, but it's hard to read.

happythenewsad: How about this?

```
  private def attempt_data
    query = <<-SQL etc etc SQL
    @attempt_data ||= ActiveRecord::Base.connection.execute(query).to_a

  end
```

happythenewsad (approved):

PR #7546, opened 2021-04-28
LMS - Use Redis to store TeacherActivityFeed

dandrabik:
## WHAT
We had some performance issue with the recently launched Activity Feed on the Teacher Dashboard. This attempts to make that feature more performant by storing the last X number of `activity_session.ids` in Redis, converting the SQL query to a faster primary_key lookup.

Note, this doesn't turn the feature on for users yet.
## WHY
This is a really cool feature that teachers really like, so we want to re-release it. Also, this adds an easy way to make the feature have real-time updates.
## HOW
Use redis arrays, push in completed activity_session ids, limit to 40. I wrote a generic base class since I could see having different Feeds on parts of the site. It was also a way to segregate the redis-specific code and the application logic for this feature.

### Notion Card Links
https://www.notion.so/quill/ea241f31a84c4b79ad56e83e99f7ee93?v=9f7aead86cc749d8a6f963070cef4f5e&p=b2ad0429bad74b6fbc063db700235a43

PR Checklist | Your Answer
------------ | -------------

TODO
- [x] Create job to Refill an individual Teacher's feed.
- [x] Determine plan to backfill active teacher data
- [x] Determine what to do with demo account, may need to backfill.
- [x] Port over helper specs.
- [x] Update spec for multiple saves.


services/QuillLMS/app/models/redis_feed.rb
```diff
@@ -0,0 +1,65 @@
...
+    $redis.ltrim(redis_key, 0, limit - 1)
+    callback_on_add(id)
+  end
+
+  # returns an array of strings
+  def ids
+    $redis.lrange(redis_key, 0, limit - 1)
+  end
+
+  # Methods defined by subclass
```
dandrabik, line 60: Note, there's not a good way to define a protocol in Ruby (that I know of, open to suggestions), so I'm raising if the subclass doesn't implement required methods.

services/QuillLMS/app/models/activity_session.rb
```diff
@@ -66,6 +66,7 @@ class ActivitySession < ActiveRecord::Base
   before_save   :set_completed_at, :set_activity_id
 
   after_save    :determine_if_final_score, :update_milestones
+  after_save :record_teacher_activity_feed, if: [:completed_at_changed?, :completed?]
```
emilia-friedberg, line 69: I've never seen the array structure for multiple conditions before. That's cool.

services/QuillLMS/spec/models/activity_session_spec.rb
```diff
@@ -955,4 +955,22 @@
...
+
+
+  describe "#teacher_activity_feed" do
+    it "should create a teacher_activity_feed item once on completed." do
+      activity_session = create(:activity_session, completed_at: nil)
+      teacher = activity_session.teachers.first
+
+      activity_session.update(completed_at: Time.now)
+      activity_session.save
+      activity_session.save
```
emilia-friedberg: Why do we have an update and then two saves here?

emilia-friedberg (approved):
This looks like a good long-term solution to me. I'm not sure it needs to be part of this PR, but it seems like maybe we need a way to handle the lack of existing data here, because otherwise the activity feed will be presenting confusing data until a teacher has had forty more activity sessions completed. Maybe it makes sense to run the larger query when that endpoint is hit if nothing is in the redis store yet, and then save those ids to the redis store so it only ever gets run once per teacher, or maybe a rake task to pre-populate the redis store for all teachers that have been active in the last six months or something. I'm thinking specifically about, like, the demo account, where students never actually "complete" activity sessions so the activity feed would always be empty even though there's data we probably want showing there.

services/QuillLMS/spec/models/teacher_activity_feed_spec.rb
```diff
@@ -0,0 +1,30 @@
+require 'rails_helper'
+
+describe TeacherActivityFeed, type: :model do
```
emilia-friedberg, line 3: There were additional tests in the test file for the helper that I wrote that we probably want to move over here.

services/QuillLMS/spec/models/activity_session_spec.rb
dandrabik: > Why do we have an update and then two saves here?

So this was an attempt to verify that multiple saves will not trigger multiple adds to the Activity Feed. I'm going to re-write this to make that clearer.

services/QuillLMS/spec/models/teacher_activity_feed_spec.rb
dandrabik: > There were additional tests in the test file for the helper that I wrote that we probably want to move over here.

Thanks for flagging, will do!

dandrabik:
> This looks like a good long-term solution to me. I'm not sure it needs to be part of this PR, but it seems like maybe we need a way to handle the lack of existing data here, because otherwise the activity feed will be presenting confusing data until a teacher has had forty more activity sessions completed. Maybe it makes sense to run the larger query when that endpoint is hit if nothing is in the redis store yet, and then save those ids to the redis store so it only ever gets run once per teacher, or maybe a rake task to pre-populate the redis store for all teachers that have been active in the last six months or something. I'm thinking specifically about, like, the demo account, where students never actually "complete" activity sessions so the activity feed would always be empty even though there's data we probably want showing there.

Thanks for raising these points. Yeah, we need a data rollout plan, likely a data backfill as you mentioned. I'll include it in this PR. I was debating just rolling this out in the background, letting the data fill for a few weeks, then turning the feature on, but in general, I don't think it's a good product experience for a teacher who has used Quill a while ago to have blank activity.  Also, thanks for flagging the demo account. We definitely will need to add code to fill that.

I've added these items as TODOs to the PR description.
Screen Shot 2021-04-29 at 10 54 27 AM

services/QuillLMS/app/models/redis_feed.rb
```diff
@@ -0,0 +1,65 @@
...
+# And those ids to be retrieved and converted to objects via a hydrate method
+# Usage:
+# Implemement subclass: SomeActivityFeed < RedisFeed
+# Add item: SomeActivityFeed.add(teacher_id, some_activity_identifier)
+# Retrieve objects: SomeActivityFeed.get(teacher_id) -> returns array of hash objects
+# Subclasses need to implement:
+# Required: key, limit, hydrate(ids:)
+# Optional: callback_on_add(id)
+
+class RedisFeed
```
brendanshean, line 12: Would `self.abstract_class = true` apply here?

dandrabik: I believe that's an [ActiveRecord feature](https://www.rubydoc.info/docs/rails/4.0.0/ActiveRecord%2FInheritance%2FClassMethods:abstract_class), not a `Ruby` one (this is Plain Old Ruby class). Unless I'm missing something.

brendanshean: You're correct.  My mistake.

services/QuillLMS/app/models/teacher_activity_feed.rb
```diff
@@ -0,0 +1,78 @@
+# Usage: TeacherActivityFeed.add(teacher_id, activity_session_id)
+# Usage: TeacherActivityFeed.get(teacher_id)
+class TeacherActivityFeed < RedisFeed
+  KEY = 'teacher-'
```
anathomical: Might we want something more descriptive/less likely to accidentally collide here?

anathomical (approved):
Can't think of a better way to define a protocol/interface other than raising NotImplemented, so that seems like a reasonable if non-optimal solution there.

Generally speaking, I like the structure of this, and appreciate you breaking out a more generalized Class.  I think you're intuition that we might want more of these feed style items is a good one.

dandrabik:
This is ready for review. A few new notes:

- There is a job to backfill an individual teacher and a job to backfill teachers whose last login falls in a given date range.
- This will launch without the activity feed visible (I will remove that code).
- I'll backfill all of the data for recently active teachers using the `TeacherActivityFeedBatchRefillWorker`.
- Once the backfill is complete, after coordinating with the proper PMs, I'll make the feature visible.
- I'll leave the Backfill update on last login (since we don't log people out often, this shouldn't create too much load).

services/QuillLMS/app/models/teacher_activity_feed.rb
```diff
@@ -13,18 +13,21 @@ def limit
   end
 
   def hydrate(ids:)
+    return [] if ids.empty?
+
     sessions = ActivitySession
-      .includes(:user, :activity, :classification, :classroom_unit)
```
emilia-friedberg: Why did you remove `:activity` here? Am I wrong in thinking that'll make the `session.activity.name` call below more expensive?

emilia-friedberg (approved):
LGTM! Excited to have this feature back.

services/QuillLMS/app/models/teacher_activity_feed.rb
```diff
@@ -0,0 +1,83 @@
...
+    # TODO: add pusher code for real time updates
+    # hydrate(ids: id_or_ids).each do |session_hash|
+    #   PusherTrigger.run(key_id, PUSHER_EVENT, session_hash)
+    # end
+  end
+
+  private def text_for_score(key, percentage)
+    return '' unless percentage
+
+    if [ActivityClassification::DIAGNOSTIC_KEY, ActivityClassification::LESSONS_KEY].include?(key)
```
anathomical: Do we want to pre-add `Comrpehension` to this list since those should also be "completed"?

services/QuillLMS/app/services/demo/report_demo_creator.rb
```diff
@@ -9,6 +9,8 @@ def self.create_demo(name)
     unit_activities = create_unit_activities(unit)
     activity_sessions = create_activity_sessions(students)
     subscription = create_subscription(teacher)
+
+    TeacherActivityFeedRefillWorker.perform_async(teacher.id)
```
anathomical, line 13: Should we be adding this to the `report_ap_demo_creater` as well?

services/QuillLMS/app/models/teacher_activity_feed.rb
dandrabik: So, I removed it because I believe it's redundant. The `:classification` relationship goes through `activity`, so having `:classification` here eager loads both `activity` and `activity_classification`. This finder performs these queries:

Screen Shot 2021-06-02 at 2 55 20 PM

services/QuillLMS/spec/workers/teacher_activity_feed_refill_worker_spec.rb
```diff
@@ -0,0 +1,56 @@
...
+  context 'teacher with classrooms and activities' do
+    let(:classroom) { create(:classroom)  }
+    let(:teacher) { classroom.owner}
+    let!(:student1) {create(:student)}
+    let!(:student2) {create(:student)}
+    let!(:classroom_unit) { create(:classroom_unit, classroom_id: classroom.id, assigned_student_ids: [student1.id, student2.id], assign_on_join: false)}
+    let!(:activity_session1) {create(:activity_session, classroom_unit_id: classroom_unit.id, user_id: student1.id, completed_at: Time.zone.now - 2.days)}
+    let!(:activity_session2) {create(:activity_session, classroom_unit_id: classroom_unit.id, user_id: student2.id, completed_at: Time.zone.now - 1.day )}
+
+    it 'should not reset and refill activity feed' do
```
brendanshean: should reset?

services/QuillLMS/spec/models/teacher_activity_feed_spec.rb
```diff
@@ -0,0 +1,167 @@
...
+    include_context "Unit Assignments Variables"
+
+    let!(:classroom_unit1) { create(:classroom_unit, classroom_id: classroom.id, assigned_student_ids: [student1.id, student2.id], assign_on_join: false)}
+    let!(:activity_session1) {create(:activity_session, classroom_unit_id: classroom_unit1.id, activity_id: activity.id, user_id: student1.id, completed_at: 1.day.ago)}
+    let!(:activity_session2) {create(:activity_session, classroom_unit_id: classroom_unit1.id, activity_id: activity.id, user_id: student2.id, completed_at: 2.days.ago)}
+
+    describe '#data_for_activity_feed' do
+      it "has all activity sessions completed for that teacher's classroom, in reverse chronological order" do
+        activity_session
+          classroom_ids = teacher.classrooms_teachers.pluck(:id)
```
brendanshean: I don't think lines 41-42 are used.

services/QuillLMS/app/workers/teacher_activity_feed_refill_worker.rb
```diff
@@ -0,0 +1,36 @@
...
+    return unless teacher
+
+    ids = ids_for_activity_feed(teacher)
+
+    return if ids.empty?
+
+    TeacherActivityFeed.reset!(user_id)
+    # Note, reversing id order since .add
+    # pops each from the front of the array
+    TeacherActivityFeed.add(user_id, ids.reverse)
```
brendanshean: I feel like a feed is more FIFO than LIFO.  Would `rpush` here:  https://github.com/empirical-org/Empirical-Core/blob/feat/redis_feed/services/QuillLMS/app/models/redis_feed.rb#L42 make more sense and also allow future self to not have to remember to reverse items?

brendanshean (approved):
Very clean code and great abstractions.  Specs were super easy to follow.

services/QuillLMS/spec/workers/teacher_activity_feed_refill_worker_spec.rb
dandrabik: Good catch, will fix!

services/QuillLMS/app/workers/teacher_activity_feed_refill_worker.rb
dandrabik: So this is suppose to be a LIFO queue, just that redis handles adding a single item to the front of the queue differently than an array. If we were to use `rpush`, you have to work around trimming to a size (there’s no `rtrim` only `ltrim`). Given that we do typically want newer items first in the list of a feed, I like the idea of adding to the beginning rather than the end, since that prevents having to do a `reverse` somewhere else. But honestly, either way would work. \

Given this, feedback, I think I might push the `reverse` complexity into my RedisFeed `.add` code so the usage will feel more obvious, so that adding an item or an array will have the same functionality.

services/QuillLMS/spec/workers/teacher_activity_feed_batch_refill_worker_spec.rb
```diff
@@ -0,0 +1,15 @@
...
+  let(:worker) { described_class.new }
+  let!(:teacher_in_window1) { create(:teacher, last_sign_in: '2020-01-05')}
+  let!(:teacher_in_window2) { create(:teacher, last_sign_in: '2020-01-06')}
+  let!(:teacher_outside) { create(:teacher, last_sign_in: '2020-02-10')}
+
+  it 'should queue jobs for teachers with last sign-in within start/end' do
+    expect(TeacherActivityFeedRefillWorker).to receive(:perform_in).with(0, teacher_in_window1.id)
+    expect(TeacherActivityFeedRefillWorker).to receive(:perform_in).with(7, teacher_in_window2.id)
+
+    worker.perform('2020-01-01', '2020-01-31', 7)
```
anathomical: It took me a few seconds figure out that the `7` in the expect was being provided as part of the `perform` call.  (I had the default of "3" in my head.). Seems like one of those places where it might be worth naming the variable.  "SPECIFIED_DELAY" or something?  Not a huge deal, just figured I'd mention it.

services/QuillLMS/spec/models/teacher_activity_feed_spec.rb
```diff
@@ -0,0 +1,166 @@
...
+    end
+
+    context 'when the timestamp was one minute ago' do
+      it 'should return "1 min ago"' do
+        expect(feed.send(:text_for_completed, 1.minute.ago)).to eq("1 min ago")
+      end
+    end
+
+    context 'when the timestamp was more than one minute but less than one hour ago' do
+      it 'should return "47 min ago"' do
```
anathomical: Not a real problem, but the "should" expectation here says "min" while the actual test is for "mins".

services/QuillLMS/spec/models/teacher_activity_feed_spec.rb
```diff
@@ -0,0 +1,166 @@
...
+      end
+    end
+
+    context 'when the timestamp was more than one week ago but this calendar year' do
+      it 'should return the written out date' do
+        date = 8.days.ago
+        if date.year == Time.now.year
+          expect(feed.send(:text_for_completed, date)).to eq(date.strftime("%b #{date.day.ordinalize}"))
+        else
+          expect(feed.text_for_completed(date)).to eq(date.strftime("%b #{date.day.ordinalize}, %Y"))
```
anathomical: I think this branch will fail without the same `.send` pattern used elsewhere in this spec due to the fact that `text_for_completed` is private.

services/QuillLMS/spec/models/teacher_activity_feed_spec.rb
```diff
@@ -0,0 +1,166 @@
...
+      it "returns #{ActivitySession::NOT_YET_PROFICIENT}" do
+        text = feed.send(:text_for_score, not_yet_proficient_activity_session.classification.key, not_yet_proficient_activity_session.percentage)
+
+        expect(text).to eq(ActivitySession::NOT_YET_PROFICIENT)
+      end
+    end
+
+  end
+
+  describe '#text_for_completed' do
```
anathomical, line 112: I love that you got exhaustive with these cases.  It makes me happy!

anathomical (approved):

services/QuillLMS/app/models/teacher_activity_feed.rb
dandrabik: Good Idea, will do.

services/QuillLMS/spec/workers/teacher_activity_feed_batch_refill_worker_spec.rb
dandrabik: Yeah, I'll update the test to make it clear. This is also a place where it would be nice if [sidekiq support keyword arguments,](https://github.com/mperham/sidekiq/issues/2372) but it does not.

services/QuillLMS/spec/models/teacher_activity_feed_spec.rb
dandrabik: These are mostly @emilia-friedberg, she did a great job here!

services/QuillLMS/spec/models/teacher_activity_feed_spec.rb
dandrabik: Yeah, breaking out this branch into its own test and converting this to a pure function for easier testing.

services/QuillLMS/spec/models/teacher_activity_feed_spec.rb
dandrabik: Will change!

services/QuillLMS/app/services/demo/report_demo_creator.rb
dandrabik: Yeah, that's seems reasonable. Will add.

services/QuillLMS/app/workers/teacher_activity_feed_refill_worker.rb
dandrabik: I've updated the interface so that an array slides onto the front of the queue maintaining its order. I agree that would be more expected than having to remember to reverse the items.

This also led to simplifying the `callback_on_add(id)` to only have to handle a single id at a time (it'll be called for each item of the array). I think this will be easier to reason around.

PR #7550, opened 2021-04-29
rename rule description to rule note

emilia-friedberg:
## WHAT
Rename the `description` attribute on the `comprehension_rules` table to `note` and update everywhere we use it in the code and in the Comprehension CMS.

## WHY
"Rule descriptions" have a meaning in Grammar that is different than what we use it to mean here, so it is better to disambiguate.

## HOW
Add a migration that renames the column and then comb through all the front- and back-end Comprehension code to update places where we use it.

### Screenshots

### Notion Card Links
https://www.notion.so/quill/Rename-Rule-Description-to-Rule-Note-c015e04753524a9197255af63d6c7711

PR Checklist | Your Answer
------------ | -------------
Design Review: If applicable, have you compared the coded design to the mockups? | YES


happythenewsad:
If you rebase against develop, the spec failure should go away

happythenewsad (approved):
Deploying this will likely cause 500s in the window between deployment and running the migration. Let's make sure there are no turking rounds going on, and in the future we should deploy backwards-compatible changes to production. `alias_method` on `note` would probably work for a two-step, backwards-compatible rollout

eadams17 (approved):

PR #7553, opened 2021-04-30
Add Controller Endpoint and JSON Hash methods to Activities and Prompt Health

cissyxyu:
## WHAT
Add an endpoint, `api/v1/activities_health.json` to the LMS. This will then return a JSON-ified object containing all activity health and prompt health data.

## WHY
This endpoint will be called by the Connect Admin Dashboard front end to get all the data at once.

## HOW
Add an endpoint and add `serializable_hash` methods to `ActivityHealth` and `PromptHealth` objects so they can be easily serialized and returned by this endpoint.

### Screenshots

### Notion Card Links
https://www.notion.so/quill/LMS-Endpoint-for-Activity-Healths-8878f90dbc284a42b68eeb356efadb1c

PR Checklist | Your Answer
------------ | -------------


services/QuillLMS/spec/factories/prompt_health.rb
```diff
@@ -0,0 +1,13 @@
+FactoryBot.define do
```
brendanshean, line 1: I would avoid magic numbers in specs and instead generate random ones.  You can use [faker](https://github.com/faker-ruby/faker/blob/master/doc/default/number.md#fakernumber) to populate various ones with conditions.

services/QuillLMS/spec/models/activity_health_spec.rb
```diff
@@ -28,4 +28,28 @@
   it { should validate_numericality_of(:recent_plays).is_greater_than_or_equal_to(0)}
   it { should validate_inclusion_of(:avg_difficulty).in_range(0..5)}
 
+  context '#serializable_hash' do
```
brendanshean, line 31: I would recommend using the rspec template of
`expect(____).to eq ____` over `assert_equals` since this communicates what is "actual" vs. "expected".  Here's a good [post](https://gist.github.com/myronmarston/5889730)

services/QuillLMS/spec/models/prompt_health_spec.rb
```diff
@@ -32,4 +32,22 @@
   it { should validate_inclusion_of(:percent_reached_optimal).in_range(0..100)}
   it { should validate_inclusion_of(:difficulty).in_range(0..5)}
 
+  context '#serializable_hash' do
+    it 'serialize into the expected shape' do
+      prompt_health = create(:prompt_health)
+      assert_equal prompt_health.serializable_hash, {
```
brendanshean: I would recommend using the rspec template of
`expect(____).to eq ____` over `assert_equals` since this communicates what is "actual" vs. "expected".  Here's a good [post](https://gist.github.com/myronmarston/5889730)

services/QuillLMS/db/structure.sql
```diff
@@ -44,6 +44,20 @@ CREATE EXTENSION IF NOT EXISTS hstore WITH SCHEMA public;
 COMMENT ON EXTENSION hstore IS 'data type for storing sets of (key, value) pairs';
```
brendanshean: It's not clear to me what this large diff has to do with the commit.  Is this just pulling in stuff that should have been pulled in other commits?

brendanshean (commented):
Just a few comments here but after that should be good to go!

services/QuillLMS/db/structure.sql
cissyxyu: This file accidentally made it into the commit. I removed it.

brendanshean (approved):

PR #7557, opened 2021-04-30
CMS - Rematch All for updated Questions in a timeframe, spread them out.

dandrabik:
## WHAT
Be able to run Rematch All for a timeframe of updated questions (so we can run this nightly). Spread those jobs out so they don't overload. Will also use this job to manually rematch all of the backlog.
## WHY
We want to keep our grading logic up to date without overloading the system.
## HOW
Add a `delay` param to the CMS `rematch_all` endpoint. Update LMS job to run for questions that have been updated in a timeframe and spread them out so they don't overload the system.

### Notion Card Links
https://www.notion.so/quill/ea241f31a84c4b79ad56e83e99f7ee93?v=9f7aead86cc749d8a6f963070cef4f5e&p=322d04e51aa4403e9e00e959548acc5d

PR Checklist | Your Answer
------------ | -------------


services/QuillCMS/spec/controllers/responses_controller_spec.rb
```diff
@@ -236,4 +236,18 @@ def filter_responses_with_parent_id(array)
...
   end
+
+  describe 'rematch_all_responses_for_question' do
+    it 'queues a job without delay as default' do
+      expect(RematchResponsesForQuestionWorker).to receive(:perform_in).with(0, '123', 'some_type')
+
+      post :rematch_all_responses_for_question, params: { uid: '123', type: 'some_type'}
+    end
+
+    it 'queues a job without delay when passed in' do
```
brendanshean: Should this be "with delay"?

services/QuillLMS/spec/workers/rematch_updated_questions_worker_spec.rb
```diff
@@ -25,16 +41,16 @@
...
+      body: {type: 'questions', uid: 'prod_question', delay: 0}.to_json,
       headers:  {'Content-Type' => 'application/json', 'Accept' => 'application/json'}
     ).once
 
     expect(HTTParty).to receive(:post).with(
       "test_url",
-      body: {type: 'questions', uid: 'beta_question'}.to_json,
+      body: {type: 'questions', uid: 'prod_question2', delay: 7}.to_json,
       headers:  {'Content-Type' => 'application/json', 'Accept' => 'application/json'}
     ).once
```
brendanshean, line 53: Is it worth checking that `uid: 'prod_question3'` is _not_ received by the endpoint?

brendanshean (approved):
A couple of very minor comments, but looks great!

cissyxyu (approved):

services/QuillLMS/spec/workers/rematch_updated_questions_worker_spec.rb
dandrabik: So this test will fail if the `HTTParty` is called without the specified arguments, so this is covered. I could explicit make a not expect, but that catchall is more thorough. Screenshot of when  `prod_question3` is called in the test.
Screen Shot 2021-05-11 at 6 09 26 PM

services/QuillCMS/spec/controllers/responses_controller_spec.rb
dandrabik: Yeah, good catch, will update.

PR #7572, opened 2021-05-03
query optimization

happythenewsad:
## WHAT
Adds materialized view for FeedbackHistory, allowing us to optimize the Rules Report query

## WHY
Rules Report times out

### Screenshots

### Notion Card Links
https://www.notion.so/quill/Product-Board-30f97e4eb01246dbb8264253fb385073?p=0c2087d11d364fbc817e65fdccb49b87

PR Checklist | Your Answer
------------ | -------------

### DIFF
--- a/services/QuillLMS/db/structure.sql
+++ b/services/QuillLMS/db/structure.sql
@@ -2,8 +2,8 @@
 -- PostgreSQL database dump
 --
 
--- Dumped from database version 10.16
--- Dumped by pg_dump version 10.16
+-- Dumped from database version 10.15
+-- Dumped by pg_dump version 10.15
 
 SET statement_timeout = 0;
 SET lock_timeout = 0;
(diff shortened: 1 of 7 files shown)


services/QuillLMS/db/structure.sql
```diff
@@ -452,6 +452,47 @@ CREATE SEQUENCE public.activity_classifications_id_seq
 ALTER SEQUENCE public.activity_classifications_id_seq OWNED BY public.activity_classifications.id;
 
 
+--
+-- Name: activity_healths; Type: TABLE; Schema: public; Owner: -
+--
+
+CREATE TABLE public.activity_healths (
```
happythenewsad, line 459: These were not related to this PR, but were not versioned before. I checked with Cissy that it's ok to include these structure.sql changes in this PR.

happythenewsad:
The build fails due to an intermittently-failing spec. Disregard

emilia-friedberg (approved):

anathomical (approved):
This looks good.  Do we need to schedule some sort of job to refresh the view?  Looks like the view will go stale pretty quickly with the current configuration.  Unless I missed something that's triggering refreshes?

happythenewsad:
> This looks good. Do we need to schedule some sort of job to refresh the view? Looks like the view will go stale pretty quickly with the current configuration. Unless I missed something that's triggering refreshes?

Good point - I'm planning to deploy this, observe the performance, and then add a chron for the refresh, with a reasonable interval. I should have mentioned this in the WHAT section. CC @emilia-friedberg as well in case she had the same question

PR #7578, opened 2021-05-04
Comprehension internal tools/ Plagiarism rules page

eadams17:
## WHAT
add a view for plagiarism rules in the Comprehension internal tools

## WHY
so Curriculum can have a more streamlined CRUD process for plagiarism rules

## HOW
added a `PlagiarismRulesIndex`, `PlagiarismRulesRouter` and shared `RuleViewForm` to handle CRUD views. This also adds validations to ensure that only one plagiarism rule is created per prompt per activity.

### Screenshots
Screen Shot 2021-05-04 at 11 50 44 AM

### Notion Card Links
https://www.notion.so/quill/New-Comprehension-Pages-Plagiarism-Rules-Page-040a827e884b4dd38b3a01f5f1539236

PR Checklist | Your Answer
------------ | -------------


services/QuillLMS/client/app/bundles/Staff/components/comprehension/plagiarismRules/plagiarismRulesIndex.tsx
```diff
@@ -0,0 +1,182 @@
...
+
+  // get cached activity data to pass to rule
+  const { data: activityData } = useQuery({
+    queryKey: [`activity-${activityId}`, activityId],
+    queryFn: fetchActivity
+  });
+
+  const { data: plagiarismRulesData } = useQuery({
+    // cache rules data for updates
+    queryKey: [`rules-${activityId}`, activityId, null, PLAGIARISM],
```
emilia-friedberg: I don't think we need to do this now, but two optional arguments feels like a code smell to me - we might want to move to passing a hash as a third object so you don't need to be passing "null".

services/QuillLMS/engines/comprehension/test/controllers/comprehension/rules_controller_test.rb
```diff
@@ -179,6 +180,16 @@ class RulesControllerTest < ActionController::TestCase
         assert_equal plagiarism_text, parsed_response['plagiarism_text']['text']
       end
 
+      should "return an error if plagiarism rule already exists for prompt" do
```
anathomical, line 192: Is it maybe worth adding a couple more tests to the model itself?
- Make sure that new items are `valid?` if they are the first Plagiarism rule for a prompt
- Make sure that new items are not `valid?` if they are not the first Plagiarism rule for a prompt
- Maybe make sure that creating back to back rules for two different prompts works?

anathomical (approved):
Looks good.  I'd like to see a couple of extra test cases around the new back-end validation added to the `Rule` model tests, but the code looks solid.

emilia-friedberg (approved):

services/QuillLMS/engines/comprehension/test/controllers/comprehension/rules_controller_test.rb
eadams17: good call, I'll add these tests

services/QuillLMS/client/app/bundles/Staff/components/comprehension/plagiarismRules/plagiarismRulesIndex.tsx
eadams17: I think that's probably wise-- could do it in one sweep for all these query calls

PR #7582, opened 2021-05-05
LMS Rails 5.0 Prework - Removing postgres_ext gem, part 1

brendanshean:
## WHAT
This is the first of two parts of removing the `postgres_ext` gem in preparation for the Rails 5 upgrade.

For this PR translates the following ActiveRecord::Relation extensions:

1.  [from_cte](https://github.com/empirical-org/Empirical-Core/blob/6bbf785dd6944fcd1455531f66be97191d2921f0/services/QuillLMS/app/queries/progress_reports/standards/student.rb#L9)
1.  [with](https://github.com/empirical-org/Empirical-Core/blob/6bbf785dd6944fcd1455531f66be97191d2921f0/services/QuillLMS/app/queries/progress_reports/concepts/concept.rb#L8)
1.  [contains](https://github.com/empirical-org/Empirical-Core/blob/6bbf785dd6944fcd1455531f66be97191d2921f0/services/QuillLMS/app/workers/archive_student_associations_for_classroom_worker.rb#L5)

Since the gem is no longer [maintained](https://github.com/DavyJonesLocker/postgres_ext#this-gem-is-no-longer-maintained) for Rails > 5, we will need to remove it and translate all of the method calls to pure ActiveRecord ones.  Gem removal will be a separate PR.

## WHY
This part of the roadmap for upgrading the LMS to Rails 5.0

## HOW
The contains query is translated via the [contains](https://github.com/DavyJonesLocker/postgres_ext/blob/master/docs/querying.md#---array-contains-operator) operator.

The bulk of this PR involves converting the common table expression functionality to pure Rails functionality.  Initially, the approach was to use raw sql containing via `WITH table AS ( [query] )` but that involved using AREL which I decided would be less readable and maintainable for our team.  I found the conversion to subqueries more natural based on the existing codebase.

I'll explain the general pattern I'm based on this [example](https://github.com/empirical-org/Empirical-Core/blob/c0b65e5a4a0050f99e210d4a1d8340cd279e922f/services/QuillLMS/app/models/classroom.rb#L96)

1.  The original assignment looked like: 
```best_activity_sessions = ProgressReports::Standards::ActivitySession.new(owner).results(filters)```

2.  Rename this variable and convert sql string: 
```best_activity_sessions_query = ProgressReports::Standards::ActivitySession.new(owner).results(filters).to_sql```

3.  Construct a subquery string:  
```best_activity_sessions = "( #{best_activity_sessions_query} ) AS best_activity_sessions"```

4.  Insert the subquery into the relevant join: 
```.joins("JOIN #{best_activity_sessions} ON activity_sessions.id = best_activity_sessions.id")```

### Notion Card Links
https://www.notion.so/quill/Remove-gem-postgres_ext-from-LMS-96bf31ec355343e4bb001d6e5c78ea01

PR Checklist | Your Answer
------------ | -------------
Have you added and/or updated tests? |  No.  This is refactoring.
Have you deployed to Staging? |. YES

### DIFF
--- a/services/QuillLMS/app/models/classroom.rb
+++ b/services/QuillLMS/app/models/classroom.rb
@@ -95,9 +95,12 @@ class Classroom < ActiveRecord::Base
 
   def unique_standard_count_array
     filters = {}
-    best_activity_sessions = ProgressReports::Standards::ActivitySession.new(owner).results(filters)
-    ActivitySession.from_cte('best_activity_sessions', best_activity_sessions)
+    best_activity_sessions_query = ProgressReports::Standards::ActivitySession.new(owner).results(filters).to_sql
+    best_activity_sessions = "( #{best_activity_sessions_query} ) AS best_activity_sessions"
+
+    ActivitySession
       .select("COUNT(DISTINCT(activities.standard_id)) as standard_count")
+      .joins("JOIN #{best_activity_sessions} ON activity_sessions.id = best_activity_sessions.id")
       .joins('JOIN activities ON activities.id = best_activity_sessions.activity_id')
       .joins('JOIN classroom_units ON classroom_units.id = best_activity_sessions.classroom_unit_id')
       .where('classroom_units.classroom_id = ?', id)
@@ -152,8 +155,8 @@ class Classroom < ActiveRecord::Base
     ActivitySession.where(classroom_unit: classroom_units).update_all(visible: false)
     classroom_units.update_all(visible: false)
     return if owner.nil?
-    
-    SetTeacherLessonCache.perform_async(owner.id) 
+
+    SetTeacherLessonCache.perform_async(owner.id)
     ids = Unit.find_by_sql("
       SELECT unit.id FROM units unit
       LEFT JOIN classroom_units as cu ON cu.unit_id = unit.id AND cu.visible = true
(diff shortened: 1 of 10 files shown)


happythenewsad (approved):

brendanshean:
NB @anathomical , @emilia-friedberg , @dandrabik I would appreciate at least two reviewers on this commit as its the most complicated of the Rails 5.0 prework.

dandrabik (approved):
This seems like a reasonable approach to me, especially given that this gem's methods aren't used that much in our codebase (so the upside of writing our own implementation is small).

We should monitor query performance. CTEs and subqueries are logically equivalent but could differ in performance. I couldn't find any clear documentation around risks of going from CTEs to subqueries performance-wise (a few anecdotal ["CTEs are more performant for me"](https://thoughtbot.com/blog/advanced-postgres-performance-tips) mentions, but nothing definitive). I mostly found the opposite: [CTEs may in some cases be less performant than subqueries](https://hakibenita.com/be-careful-with-cte-in-postgre-sql) because the query optimizer can't optimize them behind an ["optimization fence"](https://blog.crunchydata.com/blog/with-queries-present-future-common-table-expressions). So this might actually be a change for the better, performance-wise. 

Probably worth monitoring post-deploy just in case.

brendanshean:
Thank you @dandrabik .  I was surprised to find similar posts regarding the optimization fence entailed with CTEs.  I will monitor the query post deploy nonetheless.

PR #7605, opened 2021-05-07
prompt-wise admin query

happythenewsad:
## WHAT
Adds a prompt-wise report on feedback histories, for Quill admin use
Note: this PR implements the MVP version of the Notion task; not all fields are implemented. The implemented fields are designated via checkmarks in Notion.

## WHY
So that Quill admins can monitor the health of specific prompts 

### Screenshots

### Notion Card Links
https://www.notion.so/quill/Backend-New-Prompt-Health-Report-Table-c743eb47e7e547b7868d5e8ba2e350af

PR Checklist | Your Answer
------------ | -------------
Self-Review: Have you done an initial self-review of the code below on Github? | not yet


services/QuillLMS/spec/lib/rule_feedback_history_spec.rb
```diff
@@ -19,7 +19,25 @@ def rule_factory(&hash_block)
...
-  end 
+  end
+  
+  describe '#promptwise_sessions' do 
+    it 'should foo' do 
+      main_activity = create(:activity)
+      as1 = create(:activity_session, activity_id: main_activity.id)
+      as2 = create(:activity_session, activity_id: main_activity.id)
+
+      f_h1 = create(:feedback_history, activity_session_id: as1.uid)
```
happythenewsad: FeedbackHistory doesn't have an activity_session_id, so I need to change these

services/QuillLMS/app/queries/prompt_feedback_history.rb
```diff
@@ -0,0 +1,54 @@
+class PromptFeedbackHistory
+
+    def self.run(activity_id)
+        promptwise_postprocessing promptwise_sessions(activity_id)
+    end
+
+    def self.promptwise_sessions(activity_id)
+        activity_sessions = ActivitySession.where(activity_id: activity_id).includes(:feedback_sessions)
+        feedback_session_uids = activity_sessions.reduce([]) {|memo, a_s| memo.concat a_s.feedback_sessions.pluck(:uid)}
```
emilia-friedberg: I think in our Ruby style guide, we would wrap the argument to `concat` in parentheses (and that's also what I personally find easier to read): https://github.com/empirical-org/ruby#returns-val-parens.

services/QuillLMS/app/queries/prompt_feedback_history.rb
```diff
@@ -0,0 +1,54 @@
...
+    def self.promptwise_postprocessing(grouped_feedback_histories)
+        prompt_hash = {}
+
+        prompts = Comprehension::Prompt.where(id: grouped_feedback_histories.map(&:prompt_id))
+   
+        grouped_feedback_histories.each do |prompt_session|
+            prompt_id = prompt_session.prompt_id
+
+            prompt_hash[prompt_id] ||= {
+                optimal_final_attempts: 0,
```
emilia-friedberg, line 29: I think we also need a `not_optimal_final_attempts` attribute here, since it looks like the spec calls for both a percentage and an actual number.

services/QuillLMS/app/queries/prompt_feedback_history.rb
```diff
@@ -0,0 +1,54 @@
...
+            prompt_hash[prompt_id][:display_name] = prompts.find(prompt_id).text
+
+            if prompt_session.at_least_one_optimal
+              prompt_hash[prompt_id][:optimal_final_attempts] += 1
+            end
+
+            prompt_hash[prompt_id][:total_responses] += prompt_session.session_count
+            prompt_hash[prompt_id][:session_count] += 1
+
+            prompt_hash[prompt_id][:final_attempt_pct_optimal] = \
```
emilia-friedberg, line 46: What are these back slashes doing? I'm not familiar with that syntax.

services/QuillLMS/app/queries/prompt_feedback_history.rb
```diff
@@ -0,0 +1,54 @@
...
+    def self.promptwise_sessions(activity_id)
+        activity_sessions = ActivitySession.where(activity_id: activity_id).includes(:feedback_sessions)
+        feedback_session_uids = activity_sessions.reduce([]) {|memo, a_s| memo.concat a_s.feedback_sessions.pluck(:uid)}
+
+        sql = <<~SQL 
+          SELECT prompt_id, feedback_session_uid, count(*) as session_count, bool_or(optimal) as at_least_one_optimal, MAX(attempt) as attempt_cardinal
+          FROM feedback_histories
+          WHERE feedback_session_uid IN (#{feedback_session_uids.map{|e| "'#{e}'"}.join(',')})
+          GROUP BY prompt_id, feedback_session_uid
+        SQL
```
anathomical, line 16: I know that this breaks the pattern established in the last set of reports we built, but I wonder if this wouldn't be better if we did more ActiveRecord composition with this query.  I think it could let us avoid some of the awkwardness in the `WHERE` clause here.  This could be:
```ruby
FeedbackHistory.select("prompt_id, feedback_session_uid, count(*) as session_count, bool_or(optimal) as at_least_one_optimal, MAX(attempt) as attempt_cardinal")
  .where(feedback_session_uid: feedback_session_uids)
  .group(:prompt_id, :feedback_session_uid)
```
I think this will get you the same data and be a bit more ActiveRecord composable.  The big thing here is not having to do the awkward `.map.join` in the WHERE clause.

services/QuillLMS/spec/lib/rule_feedback_history_spec.rb
```diff
@@ -20,7 +20,7 @@ def rule_factory(&hash_block)
       }.merge(yield)
     )
   end
-
+
```
anathomical, line 23: Accidental white space?

anathomical (approved):
A thought on reformatting some of the query syntax, but it's just a style thing and I think this could go out as-is if you don't think it's an appropriate change to make.

emilia-friedberg (approved):
Couple of small things but nothing blocking.

services/QuillLMS/app/queries/prompt_feedback_history.rb
happythenewsad: This is a string continuation to avoid triggering Rubocop's `Layout/EmptyLinesAroundArguments: Empty line detected around arguments`

services/QuillLMS/app/queries/prompt_feedback_history.rb
happythenewsad: I like this idea, but when I tried to implement, it appears that AREL does not accept `AS` statements. Weird, I know. Going to keep this for now, because not using AS to indicate intent is going to make this code confusing.

services/QuillLMS/spec/lib/rule_feedback_history_spec.rb
happythenewsad: No; we need one there. There wasn't one previously.

services/QuillLMS/app/queries/prompt_feedback_history.rb
happythenewsad: You are correct, but as I mentioned above:

> Note: this PR implements the MVP version of the Notion task; not all fields are implemented. The implemented fields are designated via checkmarks in Notion.

services/QuillLMS/config/routes.rb
```diff
@@ -368,6 +368,7 @@
       get 'activities/activities_health' => 'activities#activities_health'
       get 'rule_feedback_histories' => 'rule_feedback_histories#by_conjunction'
       get 'rule_feedback_history/:rule_uid' => 'rule_feedback_histories#rule_detail'
+      get 'prompt_health/:activity_id' => 'prompt_feedback_histories#run'
```
emilia-friedberg: One thing I just noticed - this should probably be to `rule_feedback_histories#prompt_health`, right?

emilia-friedberg: Which makes me feel like maybe we should also have a controller test.

services/QuillLMS/app/queries/prompt_feedback_history.rb
emilia-friedberg: Yeah, I see that, but I'm not sure what the point of omitting the `not_optimal_final_attempts` is here when we have the `optimal_final_attempts` and the percentages for both. Since these are all going in the same table cell, it feels like it would make sense to also include that in the MVP.

emilia-friedberg: Which is to say, you have that box checked off in the spec but it isn't in this query.

PR #7626, opened 2021-05-11
Lms/update rules report api

anathomical:
## WHAT
- Update the way that we handle percentage values (have the back-end just return hard counts so that the front-end can calculate whatever display values are necessary)
- Add counts for the "consecutive repeated rule" and "non-consecutive repeated rule" flags so that that data can also appear in the rules report
## WHY
Discussion about additional front-end feature requirements revealed that this change in approach simplifies implementing some of those front-end changes (such as roll-ups).  The new repeated rule data is required for a couple of new columns we want to add for reporting.
## HOW
Replaced a `reduce` call with a single `each` loop which builds on multiple accumulator variables since we now count multiple items, and one loop is faster than doing multiple reducers.

### Notion Card Links
https://www.notion.so/quill/Backend-Rules-Report-Rollup-Individual-Rules-to-Summary-of-API-cc493712039c47e5954af19f32bb7d74
https://www.notion.so/quill/Rules-Report-API-Spec-d83a50cd34154aaeac6210fd543752eb

PR Checklist | Your Answer
------------ | -------------


services/QuillLMS/lib/rule_feedback_history.rb
```diff
@@ -52,26 +52,29 @@ def self.postprocessing(rules_sql_result)
...
-                total_strong = feedback_histories.reduce(0) do |memo, n|
-                    memo + n.feedback_history_ratings.filter(&:rating).count
+            total_strong = 0
+            total_weak = 0
+            repeated_consecutive = 0
+            repeated_non_consecutive = 0
+            if total_scored > 0
+                feedback_histories.each do |history|
+                  total_strong += 1 if history.feedback_history_ratings.any?(&:rating)
+                  total_weak += 1 if history.feedback_history_ratings.any? { |hr| !hr.rating }
```
emilia-friedberg: We did update `feedback_history_ratings` so ratings could have a `null` value, in an instance where someone wants to unmark a previously marked response, so I think this will have to explicitly be `{ |ht| hr.rating == false }`

emilia-friedberg (approved):
One small tweak for how we calculate weak responses but otherwise LGTM.

happythenewsad (approved):

services/QuillLMS/lib/rule_feedback_history.rb
anathomical: Good catch.  I've updated the code and added a test case to make sure we don't accidentally break this later.

PR #7635, opened 2021-05-12
Revert "Lms/update rules report api"

anathomical:
Reverts empirical-org/Empirical-Core#7626

Reverting because we merged the back-end before the front-end was ready, and don't want to deploy breaking changes.


dandrabik (approved):
:+1:

happythenewsad (approved):

PR #7640, opened 2021-05-13
add ability to mass assign ratings to feedback histories

emilia-friedberg:
## WHAT
Add ability to mass assign ratings to feedback histories.

## WHY
So the curriculum team doesn't have to manually mark each one.

## HOW
Add checkboxes that are keyboard-navigable and buttons to mass assign, as well as a route that actually does the mass assignment.

### Screenshots
Screen Shot 2021-05-13 at 3 30 44 PM

### Notion Card Links
https://www.notion.so/quill/Engineering-Spec-Mass-select-unselect-strong-weak-responses-460e2cf9ed8f4a559c471837e8139d0d

PR Checklist | Your Answer
------------ | -------------
Design Review: If applicable, have you compared the coded design to the mockups? | (YES


services/QuillLMS/app/controllers/feedback_history_ratings_controller.rb
```diff
@@ -18,6 +13,28 @@ def create_or_update
     end
   end
 
+  def mass_mark
+    params[:feedback_history_ids].each do |id|
+      rating = create_or_update_feedback_history_rating(id)
+      if rating.valid?
```
happythenewsad: This pattern allows rating #1 to be saved, and #2 to not be saved. Just want to call that out. I don't consider this high impact given the context.

happythenewsad (approved):

eadams17 (approved):

services/QuillLMS/app/controllers/feedback_history_ratings_controller.rb
anathomical: Actually, I think this particular approach could be confusing.  It saves valid rating models until the first invalid one, then stops and returns the errors for only that first invalid item.  So you could fix it and have a second invalid item that would break, while valid items after the invalid item won't be saved.

It may be cleaner to implement the same behavior we use for `batch` creating `FeedbackHistory` records: https://github.com/empirical-org/Empirical-Core/blob/e8016f231bee0ebcf950b4b48d6e3412d1dcbf61/services/QuillLMS/app/controllers/api/v1/feedback_histories_controller.rb#L27  (Basically, `create` or just `save` all of them, and if any are not valid, return an array of the various errors.)

Worth noting that this assumes we want to save what we can rather than not saving any data if any of it is invalid, which would require some different code to implement.

services/QuillLMS/client/app/bundles/Staff/utils/comprehension/feedbackHistoryRatingAPIs.ts
```diff
@@ -9,3 +9,11 @@ export const createOrUpdateFeedbackHistoryRating = async (data) => {
   });
   return { error: handleApiError('Failed to edit feedback history, please try again.', response) };
 }
+
+export const massCreateOrUpdateFeedbackHistoryRating = async (data) => {
+  const response = await mainFetch('feedback_history_rating/mass_mark', {
+    method: 'PUT',
+    body: JSON.stringify(data)
+  });
+  return { error: handleApiError('Failed to edit feedback history, please try again.', response) };
```
anathomical: This message does not fully capture current behavior (as currently we might save some values, just not all of them).  Probably not a huge deal for an internal tool.

services/QuillLMS/spec/controllers/feedback_history_ratings_controller_spec.rb
```diff
@@ -85,4 +85,30 @@
 
   end
 
+  describe "PUT #mass_mark" do
+    it "should create or update all relevant records" do
```
anathomical, line 89: Not critical, but it may be worth adding an `it` case for invalid records just to establish how we expect them to be handled (all valid get saved, but invalids return errors; nothing gets saved if any are invalid; whatever...)

anathomical (changes requested):
A couple of minor changes I think we probably want to get in here, but mostly just to clarify our intended behavior.

services/QuillLMS/spec/controllers/feedback_history_ratings_controller_spec.rb
emilia-friedberg: So, after about an hour of testing, I can't find a way to actually submit this request with data that will create an invalid record. There may be a way to stub an error for the sake of the test, but since the actual existence of an error is theoretical I think I'm going to punt on it for now if that's okay with you.

anathomical: So I think that if you passed `nil` in as the value for the rating, you'd violate a database constraint, and that could happen in the controller since we're using `permit` for `rating` instead of `require`.  That said, I think that you're point here is valid.  We don't expect invalid values here, so we don't have to create a test case to deal with them.

So let's just try to solve this by replacing the `permit` with `require` in the `feedback_history_rating_params` since that should prevent `nil` values from sneaking in there.  Also, let's replace all instances of `params` calls in the controller with `feedback_history_rating_params` since that'll run that validation for us.

anathomical (approved):
One small comment about controller params, but I think your point about validation is a good one, and I'm approving this PR based on that.

services/QuillLMS/spec/controllers/feedback_history_ratings_controller_spec.rb
emilia-friedberg: Per discussion, adding a new private method for the `mass_mark` params since they are slightly different than the `create_or_update` params.

PR #7659, opened 2021-05-18
Fix bug that was preventing new rules from being created

cissyxyu:
## WHAT
New Comprehension Rules were not being created because the validator attached to the `Rule` object was too strict. It wasn't allowing new rule creation if that rule's prompt already had a plagiarism rule. However, we should be able to create other types of rules even if that prompt already has a plagiarism rule.

## WHY
This is blocking staff from creating new AutoML rules for current live testing.

## HOW
Add an if statement to this validator to only call it if the rule being created is a plagiarism rule. Otherwise, we don't care if the prompt already has a plagiarism rule or not.

### Screenshots

### Notion Card Links
https://www.notion.so/quill/Plagiarism-errors-when-trying-to-add-a-Semantic-Label-2ef1ee46b4034e8e98e4846e2f48cf9c

PR Checklist | Your Answer
------------ | -------------


anathomical (approved):
Thanks a ton for updating the test cases.

eadams17 (approved):

PR #7676, opened 2021-05-20
Lms/prompt health updates

happythenewsad:
## WHAT
Adds full pct and num metrics in the original spec. Refactors unit tests. Please note that the strong/weak ratings metrics are on hold and not included in this PR (convo in Notion)

## WHY
It's good to have these metrics even if they're non-MVP

### Screenshots

### Notion Card Links
https://www.notion.so/quill/Backend-New-Prompt-Health-Report-Table-MVP-c743eb47e7e547b7868d5e8ba2e350af

PR Checklist | Your Answer
------------ | -------------
Self-Review: Have you done an initial self-review of the code below on Github? | not yet

### DIFF
--- a/services/QuillLMS/app/queries/prompt_feedback_history.rb
+++ b/services/QuillLMS/app/queries/prompt_feedback_history.rb
@@ -38,13 +38,18 @@ class PromptFeedbackHistory
                 avg_attempts_to_optimal: 0.0,
                 optimal_attempt_array: [],
                 display_name: '',
-                num_repeated_attempts_for_same_rule: 0
+                num_consecutive_repeated_attempts_for_same_rule: 0,
+                num_non_consecutive_repeated_attempts_for_same_rule: 0
             }
             prompt_hash[prompt_id][:display_name] = prompts.find(prompt_id).text
             if has_consecutive_repeated_rule?(prompt_session.attempts, prompt_session.rule_uids)
-              prompt_hash[prompt_id][:num_repeated_attempts_for_same_rule] += 1
+              prompt_hash[prompt_id][:num_consecutive_repeated_attempts_for_same_rule] += 1
             end 
-            
+
+            if has_non_consecutive_repeated_rule?(prompt_session.rule_uids)
+              prompt_hash[prompt_id][:num_non_consecutive_repeated_attempts_for_same_rule] += 1
+            end 
+
             if prompt_session.at_least_one_optimal
               prompt_hash[prompt_id][:optimal_final_attempts] += 1
               prompt_hash[prompt_id][:optimal_attempt_array].append prompt_session.attempt_cardinal
@@ -64,6 +69,10 @@ class PromptFeedbackHistory
           prompt_hash[k][:avg_attempts_to_optimal] = v[:optimal_attempt_array].sum / v[:optimal_final_attempts].to_f
         end
 
+        prompt_hash[k][:num_consecutive_repeated_attempts_for_same_rule] = \
+          v[:num_consecutive_repeated_attempts_for_same_rule] / v[:session_count].to_f
+        prompt_hash[k][:num_non_consecutive_repeated_attempts_for_same_rule] = \
+          v[:num_non_consecutive_repeated_attempts_for_same_rule] / v[:session_count].to_f
         prompt_hash[k][:final_attempt_pct_optimal] = \
           prompt_hash[k][:optimal_final_attempts] / prompt_hash[k][:session_count].to_f      
 
@@ -72,6 +81,10 @@ class PromptFeedbackHistory
       end
 
       prompt_hash
+    end
+
+    def self.has_non_consecutive_repeated_rule?(rule_array)
+      rule_array.uniq.count != rule_array.count
     end
 
     def self.has_consecutive_repeated_rule?(attempts_array, rule_array)
--- a/services/QuillLMS/app/queries/prompt_feedback_history.rb
+++ b/services/QuillLMS/app/queries/prompt_feedback_history.rb
@@ -11,7 +11,9 @@ class PromptFeedbackHistory
             bool_or(optimal) as at_least_one_optimal, 
             MAX(attempt) as attempt_cardinal,
             ARRAY_AGG(rule_uid) as rule_uids,
-            ARRAY_AGG(attempt) as attempts
+            -- assumes that multiple ARRAY_AGG functions process data in the same sequence
+            ARRAY_AGG(attempt) as attempts,
+            ARRAY_AGG(optimal) as optimals
           FROM feedback_histories
           JOIN comprehension_prompts
             ON feedback_histories.prompt_id = comprehension_prompts.id
@@ -39,12 +41,20 @@ class PromptFeedbackHistory
                 optimal_attempt_array: [],
                 display_name: '',
                 num_consecutive_repeated_attempts_for_same_rule: 0,
-                num_non_consecutive_repeated_attempts_for_same_rule: 0
+                num_non_consecutive_repeated_attempts_for_same_rule: 0,
+                num_first_attempt_optimal: 0
             }
             prompt_hash[prompt_id][:display_name] = prompts.find(prompt_id).text
+
+            if first_attempt_optimal?(prompt_session.attempts, prompt_session.optimals)
+              prompt_hash[prompt_id][:num_first_attempt_optimal] += 1
+            end
+
             if has_consecutive_repeated_rule?(prompt_session.attempts, prompt_session.rule_uids)
               prompt_hash[prompt_id][:num_consecutive_repeated_attempts_for_same_rule] += 1
             end 
+
+            
 
             if has_non_consecutive_repeated_rule?(prompt_session.rule_uids)
               prompt_hash[prompt_id][:num_non_consecutive_repeated_attempts_for_same_rule] += 1
@@ -62,6 +72,7 @@ class PromptFeedbackHistory
     end
 
     def self.apply_summations(prompt_hash)
+      #binding.pry
       prompt_hash.each do |k,v|
         if v[:optimal_final_attempts] == 0
           prompt_hash[k][:avg_attempts_to_optimal] = 0.0
@@ -71,16 +82,30 @@ class PromptFeedbackHistory
 
         prompt_hash[k][:num_consecutive_repeated_attempts_for_same_rule] = \
           v[:num_consecutive_repeated_attempts_for_same_rule] / v[:session_count].to_f
+
         prompt_hash[k][:num_non_consecutive_repeated_attempts_for_same_rule] = \
           v[:num_non_consecutive_repeated_attempts_for_same_rule] / v[:session_count].to_f
+
         prompt_hash[k][:final_attempt_pct_optimal] = \
           prompt_hash[k][:optimal_final_attempts] / prompt_hash[k][:session_count].to_f      
 
         prompt_hash[k][:final_attempt_pct_not_optimal] = \
           (prompt_hash[k][:session_count] - prompt_hash[k][:optimal_final_attempts]) / prompt_hash[k][:session_count].to_f 
+
+        prompt_hash[k][:pct_first_attempt_optimal] = v[:num_first_attempt_optimal] / prompt_hash[k][:session_count].to_f
+        prompt_hash[k][:pct_first_attempt_suboptimal] = (prompt_hash[k][:session_count] - v[:num_first_attempt_optimal]) / prompt_hash[k][:session_count].to_f
+        
+        v.delete_if {|k,v| k == :optimal_attempt_array}
+        v.delete_if {|k,v| k == :num_first_attempt_optimal}
+
       end
 
       prompt_hash
+    end
+
+    def self.first_attempt_optimal?(attempts_array, optimals_array)
+      attempts_rule_uids = attempts_array.zip(optimals_array).sort{|a, b| a.first <=> b.first } 
+      attempts_rule_uids.first.last 
     end
 
     def self.has_non_consecutive_repeated_rule?(rule_array)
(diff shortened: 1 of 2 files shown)


services/QuillLMS/app/queries/prompt_feedback_history.rb
```diff
@@ -18,36 +25,126 @@ def self.promptwise_sessions(activity_id)
 
     def self.promptwise_postprocessing(grouped_feedback_histories)
         prompt_hash = {}
-
+        
         prompts = Comprehension::Prompt.where(id: grouped_feedback_histories.map(&:prompt_id))
    
         grouped_feedback_histories.each do |prompt_session|
-            prompt_id = prompt_session.prompt_id
+            prompt_id = prompt_session.prompt_id.to_s.to_sym
```
happythenewsad, line 32: symbolized keys allows easier hash access later

services/QuillLMS/app/queries/prompt_feedback_history.rb
```diff
@@ -18,36 +25,126 @@ def self.promptwise_sessions(activity_id)
...
 
             prompt_hash[prompt_id] ||= {
-                optimal_final_attempts: 0, 
-                session_count: 0, 
-                total_responses: 0,
-                final_attempt_pct_optimal: 0,
-                final_attempt_pct_not_optimal: 0,
-                display_name: ''
+
+                total_responses: 0.0,
```
happythenewsad, line 36: I've standardized the order of all these metrics everywhere, for legibility

services/QuillLMS/app/queries/prompt_feedback_history.rb
```diff
@@ -18,36 +25,126 @@ def self.promptwise_sessions(activity_id)
...
+        prompt_hash[k][:pct_final_attempt_optimal] = \
+          prompt_hash[k][:num_final_attempt_optimal] / session_count      
+
+        prompt_hash[k][:pct_final_attempt_not_optimal] = \
+          prompt_hash[k][:num_final_attempt_not_optimal] / session_count
+          
+        prompt_hash[k][:pct_first_attempt_optimal] = v[:num_first_attempt_optimal] / session_count
+        prompt_hash[k][:pct_first_attempt_not_optimal] = v[:num_first_attempt_not_optimal] / session_count
+        
+        v.delete_if {|k,v| k == :optimal_attempt_array}
```
happythenewsad: we don't need to send this intermediate computations over the wire

happythenewsad:
The report logic is getting pretty gnarly. I've been down in the weeds in this - if you see some opportunities for improvement please let me know.

services/QuillLMS/app/queries/prompt_feedback_history.rb
```diff
@@ -18,36 +25,126 @@ def self.promptwise_sessions(activity_id)
 
     def self.promptwise_postprocessing(grouped_feedback_histories)
         prompt_hash = {}
-
+
```
anathomical, line 28: Extra whitespace?

services/QuillLMS/app/queries/prompt_feedback_history.rb
```diff
@@ -6,7 +6,14 @@ def self.run(activity_id)
...
         sql = <<~SQL 
-          SELECT prompt_id, feedback_session_uid, count(*) as session_count, bool_or(optimal) as at_least_one_optimal, MAX(attempt) as attempt_cardinal
+          SELECT prompt_id, feedback_session_uid, 
+            count(*) as session_count, 
+            bool_or(optimal) as at_least_one_optimal, 
+            MAX(attempt) as attempt_cardinal,
+            ARRAY_AGG(rule_uid) as rule_uids,
+            -- assumes that multiple ARRAY_AGG functions process data in the same sequence
+            ARRAY_AGG(attempt) as attempts,
+            ARRAY_AGG(optimal) as optimals
```
anathomical, line 16: I'm pretty sure that it is a safe assumption that multiple `array_agg` calls will maintain relative order..

That said, as far as I can tell, you're doing this primarily in order to calculate consecutive and non-consecutive repetitions.  I think we can actually get that by simply `LEFT JOIN feedback_history_flags` and `COUNT(flag = 'repeated-consecutive')` and `COUNT(flag = 'repeated-non-consecutive')`.  Or, better yet, find a way to reference the model constants `FeedbackHistoryFlag::FLAG_REPEATED_RULE_CONSECUTIVE` and `FeedbackHistoryFlag::FLAG_REPEATED_RULE_NON_CONSECUTIVE` in your counts.

It's an extra join, but I suspect that it's not really any more expensive to do the counts than it is to do a series of `array_agg` calls.  Thoughts?

services/QuillLMS/app/queries/prompt_feedback_history.rb
```diff
@@ -18,36 +25,126 @@ def self.promptwise_sessions(activity_id)
...
+
+                num_consecutive_repeated_attempts_for_same_rule: 0.0,
+                num_non_consecutive_repeated_attempts_for_same_rule: 0.0,
+                pct_consecutive_repeated_attempts_for_same_rule: 0.0,
+                pct_non_consecutive_repeated_attempts_for_same_rule: 0.0,
+
+                num_first_attempt_optimal: 0.0,
+                num_first_attempt_not_optimal: 0.0,
+                pct_first_attempt_optimal: 0.0,
+                pct_first_attempt_not_optimal: 0.0
```
anathomical: I think it might be worth documenting this API payload expectation in Notion in a document similar to this one: https://www.notion.so/quill/Rules-Report-API-Spec-d83a50cd34154aaeac6210fd543752eb

services/QuillLMS/app/queries/prompt_feedback_history.rb
```diff
@@ -18,36 +25,126 @@ def self.promptwise_sessions(activity_id)
...
+          v[:num_non_consecutive_repeated_attempts_for_same_rule] / session_count
+
+        prompt_hash[k][:pct_final_attempt_optimal] = \
+          prompt_hash[k][:num_final_attempt_optimal] / session_count      
+
+        prompt_hash[k][:pct_final_attempt_not_optimal] = \
+          prompt_hash[k][:num_final_attempt_not_optimal] / session_count
+          
+        prompt_hash[k][:pct_first_attempt_optimal] = v[:num_first_attempt_optimal] / session_count
+        prompt_hash[k][:pct_first_attempt_not_optimal] = v[:num_first_attempt_not_optimal] / session_count
```
anathomical: Given that these percentage calculations are trivial, and we're already passing all of the required values to calculate them on the front-end, is it worth reducing payload size and (trivially) reducing CPU cost on the server since putting the calculation in the browser is effectively free?

We've also moved away from percentage calculations in the back-end on the Rules Report API, so doing so here might be worth doing just to keep expectations aligned?  No strong feelings on this, just a thought.

services/QuillLMS/app/queries/prompt_feedback_history.rb
anathomical: I'm a big fan of this kind of standardization.  I wonder if we can come up with a good general rubric to use for this kind of thing moving forward, or if we'll just have to do it case-by-case.

anathomical (changes requested):
A couple of different considerations here.  I'm actually not that concerned about the percentage thing other than that it might make this simpler if we don't have to do it.  I do think that counting the flags directly rather than zipping attempts and rule_uids together to loop over will be way more efficient.

services/QuillLMS/app/queries/prompt_feedback_history.rb
emilia-friedberg: Agreed!

services/QuillLMS/app/queries/prompt_feedback_history.rb
emilia-friedberg: Yeah, I agree, especially because these "percentages" are actually decimals right now and have to get multiplied by *100 for display on the frontend anyway.

services/QuillLMS/app/queries/prompt_feedback_history.rb
```diff
@@ -18,36 +25,126 @@ def self.promptwise_sessions(activity_id)
...
+                num_final_attempt_optimal: 0.0,
+                num_final_attempt_not_optimal: 0.0,
+
+                avg_attempts_to_optimal: 0.0,
+
+                num_consecutive_repeated_attempts_for_same_rule: 0.0,
+                num_non_consecutive_repeated_attempts_for_same_rule: 0.0,
+                pct_consecutive_repeated_attempts_for_same_rule: 0.0,
+                pct_non_consecutive_repeated_attempts_for_same_rule: 0.0,
+
```
emilia-friedberg: These names don't seem to me like they accurately reflect the measure - they're a count/percentage of sessions, not attempts. Maybe something more like `num_sessions_with_consecutive_repeated_rule`, etc?

services/QuillLMS/app/queries/prompt_feedback_history.rb
emilia-friedberg: It also feels to me like this section would be easier to read if you just did something like:

```
aggregate_data_hash = {
  pct_consecutive_repeated_attempts_for_same_rule: v[:num_consecutive_repeated_attempts_for_same_rule] / session_count,
  pct_non_consecutive_repeated_attempts_for_same_rule: v[:num_non_consecutive_repeated_attempts_for_same_rule] / session_count,
  pct_final_attempt_optimal: num_final_attempt_optimal] / session_count,
  pct_final_attempt_not_optimal: num_final_attempt_not_optimal] / session_count,
  pct_first_attempt_optimal: v[:num_first_attempt_optimal] / session_count,
  pct_first_attempt_not_optimal: v[:num_first_attempt_not_optimal] / session_count
}
prompt_hash[k] = prompt_hash[k].merge(aggregate_data_hash)
```

services/QuillLMS/app/queries/prompt_feedback_history.rb
```diff
@@ -18,36 +25,126 @@ def self.promptwise_sessions(activity_id)
...
+          prompt_hash[k][:num_final_attempt_optimal] / session_count      
+
+        prompt_hash[k][:pct_final_attempt_not_optimal] = \
+          prompt_hash[k][:num_final_attempt_not_optimal] / session_count
+          
+        prompt_hash[k][:pct_first_attempt_optimal] = v[:num_first_attempt_optimal] / session_count
+        prompt_hash[k][:pct_first_attempt_not_optimal] = v[:num_first_attempt_not_optimal] / session_count
+        
+        v.delete_if {|key,val| k == :optimal_attempt_array}
+        v.delete_if {|key,val| k == :num_first_attempt_optimal}
```
emilia-friedberg: I know `optimal_attempt_array` doesn't need to be in the payload, but why are you deleting `num_first_attempt_optimal` here?

emilia-friedberg (changes requested):
Mostly echoing Thomas with a few smaller points.

services/QuillLMS/app/queries/prompt_feedback_history.rb
happythenewsad: No

services/QuillLMS/app/queries/prompt_feedback_history.rb
happythenewsad: Yeah, that might be more performant, or it might not. I'd like to ship as-is so Emilia can begin work though.

services/QuillLMS/app/queries/prompt_feedback_history.rb
happythenewsad: 👍

services/QuillLMS/app/queries/prompt_feedback_history.rb
happythenewsad: sure

services/QuillLMS/app/queries/prompt_feedback_history.rb
happythenewsad: 👍

services/QuillLMS/app/queries/prompt_feedback_history.rb
happythenewsad: Hopefully it now feels a lot cleaner now that the percentage computations are gone.

services/QuillLMS/app/queries/prompt_feedback_history.rb
happythenewsad: Good catch - during development I was using this as an intermediate value and forgot that it was a terminal value as well.

services/QuillLMS/app/queries/prompt_feedback_history.rb
happythenewsad: I've added the API payload expectation here: https://www.notion.so/quill/Backend-New-Prompt-Health-Report-Table-MVP-c743eb47e7e547b7868d5e8ba2e350af#c94135e5d4704996bc3056d5e7fa8ccb

emilia-friedberg: I think what @anathomical is suggesting, and what I would prefer, is that that documentation live in the documentation folder (like his example) rather than in the task card, which can be harder to find later for future reference and is also weirder to change (if you have a new task that modifies the API, for instance).

emilia-friedberg (approved):
Approving code-wise, but I do want to flag that I think we should be documenting this API in the way that @anathomical did in his example instead of in the product card.

happythenewsad:
> Approving code-wise, but I do want to flag that I think we should be documenting this API in the way that @anathomical did in his example instead of in the product card.

I copied what was in the card into the [documentation](https://www.notion.so/quill/Rules-Report-API-Spec-d83a50cd34154aaeac6210fd543752eb)

anathomical (approved):

PR #7649, opened 2021-05-17
Transitioning relevant lib/* to app/lib

happythenewsad:
## WHAT
Moves `lib/` to `app/lib` and makes necessary path adjustments for this change.

## WHY
@brendanshean 's May 13 eng. meeting discussion 
here's a good [post](https://medium.com/extreme-programming/what-goes-in-rails-lib-92c74dfd955e) for reference

Also, from the zeitwerk author: [rails/rails#37835 (comment)](https://github.com/rails/rails/issues/37835#issuecomment-757367812)
and another relevant comment: [rails/rails#13142 (comment)](https://github.com/rails/rails/issues/13142#issuecomment-275492070)
### Screenshots
(If applicable. Also, please censor any sensitive data)

### Notion Card Links
none

PR Checklist | Your Answer
------------ | -------------
Have you added and/or updated tests? |  yes
Have you deployed to Staging? | not yet
Self-Review: Have you done an initial self-review of the code below on Github? | yes
Design Review: If applicable, have you compared the coded design to the mockups? | (N/A

PR #7678, opened 2021-05-20
Move relevant files from lib/ to 'app/lib/'

brendanshean:
## WHAT
Moves lib/ to app/lib and makes necessary path adjustments for this change.

## WHY
@brendanshean 's May 13 eng. meeting discussion
here's a good post for reference

Also, from the zeitwerk author: rails/rails#37835 (comment)
and another relevant comment: rails/rails#13142 (comment)

## HOW

More detail in Peter Kong's original PR: https://github.com/empirical-org/Empirical-Core/pull/7649

### Screenshots

### Notion Card Links
https://www.notion.so/quill/LMS-Move-lib-to-app-lib-85fc6ae5895d41f587c20704328d2497

PR Checklist | Your Answer
------------ | -------------
Self-Review: Have you done an initial self-review of the code below on Github? |. YES
Design Review: If applicable, have you compared the coded design to the mockups? | (N/A or Yes). N/A


dandrabik (changes requested):
Thanks for cleaning this up! A few files that might be better to leave and/or delete (happy to discuss if there are reasons to move/keep):

- `lib/assets/.gitkeep` I think we can just delete this (to avoid another `assets` directory in `app`).
- Everything in `lib/data` seems to be used by migrations or rake tasks (you can confirm by searching by filename). I think we should leave this folder in `lib`.

brendanshean:
@dandrabik Those are sensible suggestions.  I have updated the PR.

dandrabik (approved):
Thanks for making the changes!

happythenewsad (approved):

brendanshean:
@happythenewsad @dandrabik I moved AsyncHelper into `spec/support` since it was preventing the build to complete.  (It's only used by one spec in particular).
Screen Shot 2021-05-20 at 2 48 56 PM

dandrabik (approved):
:+1: LGTM. Thanks for moving this along.

PR #7685, opened 2021-05-21
Fix anonymous url activity bug

brendanshean:
## WHAT
The table `ActivityClassification` has an attribute `module_url` where values have domains hardcoded.  
Screen Shot 2021-05-21 at 9 01 59 AM

When this attribute is called on a staging site, it causes links to redirect from a staging site to production.

## WHY
This bug will prevent integration tests from getting accurate results on staging configurations.  For example, a ghost inspector test might click on one of these links and then effectively be testing the production site.

## HOW
As a first step, keep production functionality the same (hence the guard clause), but remove domain information in other configurations.  Ideally, we would get rid of the domain information altogether but at this point, first address the change on staging to confirm that it works.  The inclusion of the `Rails.env.production?` is not ideal, but I don't see any way around that initially since this is an issue between environments rather than application code.

This fix does not address module_urls: `https://quill-writer.firebaseapp.com/` or `https://grammar.quill.org/play/sw` as those domains don't seem to have corresponding staging equivalents.

### Screenshots

### Notion Card Links
https://www.notion.so/quill/Fix-anonymous-activity-redirect-bug-68d328a5a63c4193a5861c9f8d01aa0a

PR Checklist | Your Answer
------------ | -------------

### DIFF
--- a/services/QuillLMS/app/models/activity_classification.rb
+++ b/services/QuillLMS/app/models/activity_classification.rb
@@ -60,9 +60,19 @@ class ActivityClassification < ActiveRecord::Base
     find_by_key CONNECT_KEY
   end
 
-  def module_url
-    return read_attribute(:module_url) if Rails.env.production?
+  def form_url
+    "#{ENV['DEFAULT_URL']}#{form_path}"
+  end
 
-    "#{ENV['DEFAULT_URL']}#{Addressable::URI.parse(read_attribute(:module_url)).path}"
+  def module_url
+    "#{ENV['DEFAULT_URL']}#{module_path}"
+  end
+
+  private def module_path
+    Addressable::URI.parse(read_attribute(:module_url)).path
+  end
+
+  private def form_path
+    Addressable::URI.parse(read_attribute(:form_url)).path
   end
 end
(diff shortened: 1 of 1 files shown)


dandrabik (changes requested):
So I think we'd need an equivalent method for `form_url` which is also in the DB to completely solve this issue.

Also, as you mention in the PR description, having the `if Rails.env.production?` really worries me (but I get that's the easiest fix). So I'm a bit torn between two paths:
1. Put this fix out (after adding the `form_url`) to continue testing on staging, and make a task for a more permanent solution.
2. Just put out a more permanent solution.

I'm fine with doing 1. since we want to continue testing and there are a few options of how we can fix this setup. I'm curious your thoughts.

brendanshean:
@dandrabik Great point on `form_url`.  I will address that.

I'm not crazy about the `Rails.env.production`.  I could make the coupling more explicit and change the conditional to: `ENV['default_url'] == 'https://www.quill.org'`.

brendanshean:
Then when we're happy with how staging is working, remove the  `return read_attribute(:module_url) if Rails.env.production?` altogether in a later PR.

emilia-friedberg (approved):

brendanshean:
@dandrabik I'm realizing my original reason in using the conditional was to handle the edge cases: https://quill-writer.firebaseapp.com/ or https://grammar.quill.org/play/sw.  Since those aren't relevant, I think I can dispense with conditional altogether and just offer something more permanent.

dandrabik (approved):
This looks good to me. I also don't think we need to handle those edge cases any more.

anathomical (approved):

PR #7691, opened 2021-05-21
Comprehension/hide scores from students

anathomical:
## WHAT
Treat Comprehension reporting as "ungraded" in the same way that we do Diagnostic and Lessons.  This allows us to show "Complete" instead of "Proficient" or whatever.
## WHY
The design of Comprehension expects students to have to make multiple attempts on every prompt, and in a strict grading system we expect most students to "score" less than 50%.  Since this is the expectation, we don't want to show students that they are "not proficient" if they are meeting our expectations.
## HOW
Originally as part of score reporting we returned the `ActivityClassification.id` value and checked against a hard-coded list of numbers.  However, because of the way that we added "Comprehension" as an `ActivityClassification`, it may not have the same ID in every environment.  Therefore, it made sense to me to pull the `ActivityClassification.name` in the API call, and hard code those values for comparison instead.  Same results, but slightly less brittle across environments.

One note: I left the `ActivityClassification.id` in the query in case it's used somewhere else, but I suspect that it is not.  Still, better safe than sorry.

### Notion Card Links
https://www.notion.so/quill/Do-not-show-Comprehension-activity-scores-to-students-who-completed-Comprehension-activities-369b9c6243204fd9be5f27c3025d39cb

PR Checklist | Your Answer
------------ | -------------
Have you added and/or updated tests? |  No test coverage on this

### DIFF
--- a/services/QuillLMS/app/models/unit_activity.rb
+++ b/services/QuillLMS/app/models/unit_activity.rb
@@ -106,7 +106,7 @@ class UnitActivity < ActiveRecord::Base
       activity.name,
       activity.description,
       activity.repeatable,
-      activity.activity_classification_id,
+      activity_classifications.name AS activity_classification_name,
       unit.id AS unit_id,
       ua.id AS ua_id,
       unit.created_at AS unit_created_at,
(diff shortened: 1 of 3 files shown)


emilia-friedberg (changes requested):
This is effectively the same, but I think we should probably be using the activity classification `key` here rather than the name, as we do elsewhere when we are relying on a static value.

anathomical:
@emilia-friedberg Good call.  That's less ambiguous and more succinct.  I've updated the PR.

emilia-friedberg (approved):

PR #7703, opened 2021-05-25
Implement Case Sensitive behavior for Comprehension Regex Rules

cissyxyu:
## WHAT
Due to my own oversight, I wasn't actually implementing Regex Rules as case sensitive or case insensitive even though I built the flagging ability into the model. This PR just implements that behavior so that case INSENSITIVE regex rules are applied insensitive to case.

## WHY
Curriculum wants these regex rules to work if they flag them as case insensitive

## HOW
Add branching behavior based on whether the regex rule is case sensitive or case insensitive

### Screenshots

### Notion Card Links
https://www.notion.so/quill/Comprehension-regex-case-insensitive-button-is-not-working-488edc5f24dd4788a790265b80b2119b

PR Checklist | Your Answer
------------ | -------------

### DIFF
--- a/services/QuillLMS/engines/comprehension/app/models/comprehension/rule.rb
+++ b/services/QuillLMS/engines/comprehension/app/models/comprehension/rule.rb
@@ -68,8 +68,10 @@ module Comprehension
 
     def regex_is_passing?(entry)
       regex_rules.none? do |regex_rule|
-        is_matching = match_with_case_sensitive(regex_rule.regex_text, regex_rule.case_sensitive, entry)
-        regex_rule.sequence_type == RegexRule::TYPE_INCORRECT ? is_matching : !is_matching
+        regex_match = match_with_case_sensitive(regex_rule.regex_text, regex_rule.case_sensitive, entry)
+        # for "incorrect" type regex rules, we want to "fail" if they have the regex. for "required" type regex
+        # rules, we want to "fail" when they dont have the regex.
+        regex_rule.sequence_type == RegexRule::TYPE_INCORRECT ? regex_match : !regex_match
       end
     end
 
@@ -77,7 +79,7 @@ module Comprehension
       DISPLAY_NAMES[rule_type.to_sym] || rule_type
     end
 
-    private def match_with_case_sensitive(regex, is_case_sensitive, text)
+    private def match_with_case_sensitive(regex, is_case_sensitive=false, text)
       is_case_sensitive ? Regexp.new(regex).match(text) : Regexp.new(regex, Regexp::IGNORECASE).match(text)
     end
 
--- a/services/QuillLMS/engines/comprehension/app/models/comprehension/rule.rb
+++ b/services/QuillLMS/engines/comprehension/app/models/comprehension/rule.rb
@@ -68,12 +68,17 @@ module Comprehension
 
     def regex_is_passing?(entry)
       regex_rules.none? do |regex_rule|
-        regex_rule.sequence_type == RegexRule::TYPE_INCORRECT ? Regexp.new(regex_rule.regex_text).match(entry) : !Regexp.new(regex_rule.regex_text).match(entry)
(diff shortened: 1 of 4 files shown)


services/QuillLMS/engines/comprehension/app/models/comprehension/rule.rb
```diff
@@ -68,14 +68,19 @@ def determine_feedback_from_history(feedback_history)
...
+        is_matching = match_with_case_sensitive(regex_rule.regex_text, regex_rule.case_sensitive, entry)
+        regex_rule.sequence_type == RegexRule::TYPE_INCORRECT ? is_matching : !is_matching
       end
     end
 
     def display_name
       DISPLAY_NAMES[rule_type.to_sym] || rule_type
     end
 
+    private def match_with_case_sensitive(regex, is_case_sensitive, text)
```
happythenewsad: This is a good use case for default keyword args.

If you change the method signature to 
```ruby
match_with_case_sensitive(regex, text, is_case_sensitive: true)
```

you both communicate the default value and the type of the parameter.

services/QuillLMS/engines/comprehension/app/models/comprehension/rule.rb
```diff
@@ -68,14 +68,19 @@ def determine_feedback_from_history(feedback_history)
 
     def regex_is_passing?(entry)
       regex_rules.none? do |regex_rule|
-        regex_rule.sequence_type == RegexRule::TYPE_INCORRECT ? Regexp.new(regex_rule.regex_text).match(entry) : !Regexp.new(regex_rule.regex_text).match(entry)
+        is_matching = match_with_case_sensitive(regex_rule.regex_text, regex_rule.case_sensitive, entry)
```
happythenewsad: It's hard to follow this method. `is_` by convention connotes a boolean, but in this case returns MatchData. Please rename the variable.

line 72 is also not clear to me, but perhaps fixing the variable name will help.

happythenewsad (changes requested):

happythenewsad (approved):
Approved, after linting passes.

services/QuillLMS/engines/comprehension/app/models/comprehension/rule.rb
```diff
@@ -68,14 +68,21 @@ def determine_feedback_from_history(feedback_history)
...
+        # rules, we want to "fail" when they dont have the regex.
+        regex_rule.sequence_type == RegexRule::TYPE_INCORRECT ? regex_match : !regex_match
       end
     end
 
     def display_name
       DISPLAY_NAMES[rule_type.to_sym] || rule_type
     end
 
+    private def match_with_case_sensitive(regex, is_case_sensitive=false, text)
```
anathomical: It almost feels like this function should be on the `RegexRule` model.  You could avoid passing in both `regex` and `is_case_sensitive` here and just do something like `RegexRule.check_for_match(text)`.  It's probably not a big deal, but I think that getting to avoid passing this data around is probably a plus.

services/QuillLMS/engines/comprehension/app/models/comprehension/rule.rb
```diff
@@ -68,14 +68,21 @@ def determine_feedback_from_history(feedback_history)
 
     def regex_is_passing?(entry)
       regex_rules.none? do |regex_rule|
```
anathomical, line 70: If you put the matching function on the `RegexRule` model, this function becomes a lot simpler:
```ruby
regex_rules.none? { |regex_rule| regex_rule.check_for_match(entry) }
```
Or whatever.

anathomical (approved):
The code looks fine to me.  I think I'd shift the location of your core logic to a different model, but the implementation you have in place makes sense to me.  You could also skip moving it for now, but I do think it would be cleaner.

anathomical (approved):

PR #7709, opened 2021-05-25
comprehension/null-activity-session-percentage

anathomical:
## WHAT
Tweak LMS reporting to give "null" scores
## WHY
Since we don't expect more traditional performance in Comprehension, we don't want to provide confusing numbers
## HOW
Just change the code that calculates a live percentage for code that sets the constant value `null` when saving the `ActivitySession`.

### Notion Card Links
https://www.notion.so/quill/Remove-Comprehension-Scores-b185c1265505449fbb5c85e9bfd31ac2

PR Checklist | Your Answer
------------ | -------------

### DIFF
--- a/services/QuillLMS/client/app/bundles/Comprehension/components/studentView/container.tsx
+++ b/services/QuillLMS/client/app/bundles/Comprehension/components/studentView/container.tsx
@@ -15,7 +15,7 @@ import { completeActivitySession,
          getFeedback,
          processUnfetchableSession,
          saveActiveActivitySession } from '../../actions/session'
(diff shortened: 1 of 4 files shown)


emilia-friedberg (approved):

cissyxyu (approved):

cissyxyu (commented):
I think originally this percentage was included to show the number of attempts the student used to get to the correct answer. But I think now that we have Feedback Histories we don't need to record the attempts number in this roundabout way.

eadams17 (approved):

PR #7711, opened 2021-05-26
LMS Rails 5.0 Prework - Refactor ActiveRecord::Base.connection.execute calls to use TypeMapByOID

brendanshean:
## WHAT
Refactor `ActiveRecord::Base.connection.execute` calls in Rails 4 into a new call `RawSqlRunner.execute` that fills in the appropriate TypeMap so that the resulting `PG:Result` matches the one in Rails 5.

## WHY
`ActiveRecord::Base.connection.execute` yields a `PG::Result`. Calling `to_a` on that `PG::Result` gives different values in Rails 4 and Rails 5.  This seems to be due to the [TypeMap](https://deveiate.org/code/pg/PG/TypeMap.html) associated with the `PG::Result` which is used in casting values.

Rails 4 gives `PG::TypeMapAllStrings`:
rails 4 2 11 3

whereas Rails 5 gives `PG::TypeMapByOid`
rails 5 0 7

More back story [here](https://stackoverflow.com/a/58953275)

## HOW
Find all OIDs used in Rails 5 `ActiveRecord::Base.connection.execute` calls and then use them as [coders](https://deveiate.org/code/pg/PG/Coder.html) in the monkey patch.

NB.  The Raw SQL diffs are just formatting to improve readability and adding heredoc.

### Screenshots

### Notion Card Links
https://www.notion.so/quill/Refactor-ActiveRecord-Base-connection-execute-calls-to-use-TypeMapByOID-5b1a320b11824b7088a63458667891e8

PR Checklist | Your Answer
------------ | -------------

### DIFF
--- a/services/QuillLMS/Gemfile
+++ b/services/QuillLMS/Gemfile
@@ -10,7 +10,7 @@ gem 'puma', '~> 3.12'
 gem 'dotenv-rails'
 
 # DB/MODEL
-gem 'pg', '0.19.0'
+gem 'pg', '0.18.4'
 gem 'ancestry'
 gem 'ransack'
 gem 'ranked-model'
--- a/services/QuillLMS/Gemfile
+++ b/services/QuillLMS/Gemfile
@@ -10,7 +10,7 @@ gem 'puma', '~> 3.12'
 gem 'dotenv-rails'
 
 # DB/MODEL
-gem 'pg', '0.18.4'
+gem 'pg', '0.19.0'
 gem 'ancestry'
 gem 'ransack'
 gem 'ranked-model'
--- a/services/QuillLMS/Gemfile.lock
+++ b/services/QuillLMS/Gemfile.lock
@@ -431,7 +431,7 @@ GEM
       hashery (~> 2.0)
       ruby-rc4
       ttfunk
-    pg (0.18.4)
+    pg (0.19.0)
     pointpin (1.0.1)
       faraday
       hashie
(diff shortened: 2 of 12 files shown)


brendanshean:
For future reference, I used this script to pull out type maps for different queries.  It appears as though the same coders are used for all queries regardless of the data types involved.
   ```
   results = ActiveRecord::Base.connection.execute("[some raw sql]")

    File.open('executes.csv', 'a') do |f|
      results.type_map.coders.map(&:inspect).each do |coder|
        f.write([self.class.to_s, __method__].join('#'))
        f.write(",")
        decoder, name, oid = coder.split(" ")
        f.write(decoder.split(':')[0..-2].join(':')[2..-1])
        f.write(',' + name[1..-2])
        f.puts(',' + oid[0..-2])
      end
    end
  end
```

dandrabik (approved):
:+1: LGTM. I like this approach of trying this in a few places to start, then rolling it out to more queries.

Thanks for working through this.

brendanshean:
Documentation for looking at the route suggested [here](https://stackoverflow.com/a/27196072):

Screen Shot 2021-05-28 at 7 04 40 AM

using this [method](https://github.com/ged/ruby-pg/blob/00cb2ecfaa70470234ee83efbd942b5e6ea0f4ea/spec/helpers.rb#L268)

Ultimately this path would be more work as custom ruby methods for type casting would be required.

PR #7722, opened 2021-05-27
Automl model notes

happythenewsad:
## WHAT
Add annotation feature to auto ml models so that admins can have notes

## HOW
- add a column
- add requisite controller scaffolding
- add a new component to the frontend for state management and ajax
- 
### Screenshots

### Notion Card Links
https://www.notion.so/quill/Semantic-Labels-Page-Add-Model-Notes-1cf13134be354203a170d7eab0ad8c9e

PR Checklist | Your Answer
------------ | -------------
Have you added and/or updated tests? |  no
Design Review: If applicable, have you compared the coded design to the mockups? | yes


services/QuillLMS/client/app/bundles/Staff/components/comprehension/semanticRules/modelsTable.tsx
```diff
@@ -54,7 +96,7 @@ const ModelsTable = ({ activityId, prompt }) => {
     { name: "Created At", attribute:"created_at", width: "100px" },
     { name: "Version", attribute:"version", width: "70px" },
     { name: "Model Name", attribute:"name", width: "300px" },
-    // { name: "Model Notes", attribute:"", width: "200px" },
+    { name: "Model Notes", attribute:"notes", width: "150px" },
```
happythenewsad, line 69: I added a somewhat aribtary width value that seemed reasonable.

services/QuillLMS/client/app/bundles/Staff/components/comprehension/semanticRules/modelsTable.tsx
```diff
@@ -15,6 +19,37 @@ const ModelsTable = ({ activityId, prompt }) => {
...
+      });
+  }
+
+  const NotesWrapper = (props) => {
+    const { modelId, initialNotes } = props;
+    const [noteText, setNoteText] = React.useState(initialNotes);
+    
+    const handleNoteChange = (e, modelId) => {
+      setNoteText(e.target.value)
+      updateModelNote(modelId, e.target.value )
```
happythenewsad: We have an opportunity to use useDebounce() here so that not every character is PUT'd to the API. Thoughts?

services/QuillLMS/client/app/bundles/Staff/components/comprehension/semanticRules/modelsTable.tsx
```diff
@@ -15,6 +19,37 @@ const ModelsTable = ({ activityId, prompt }) => {
     queryFn: fetchModels
   });
 
+  const updateModelNote = (id, notes)  => {
+    request.put({
+      url: `${process.env.DEFAULT_URL}/api/v1/comprehension/automl_models/${id}`,
+      json: { automl_model: {notes} }, },
+      (error, httpStatus, body) => {
+        if (error) {
```
happythenewsad: Error handling was not specified in the spec. Do we have a default input error handler that I can use?

happythenewsad:
Hey @eadams17 , please note that there are 3 places where this PR may need changes - I've added a comment to each and would like your feedback. Thanks!

services/QuillLMS/client/app/bundles/Staff/components/comprehension/semanticRules/modelsTable.tsx
eadams17: yeah-- totally fine!

services/QuillLMS/client/app/bundles/Staff/components/comprehension/semanticRules/modelsTable.tsx
eadams17: so correct me if I'm wrong... but looking at the notion doc for this, I'm not sure we have this in the right place. I think we just want to display the note data here, but edit it on the view page: https://www.quill.org/cms/comprehension#/activities/87/semantic-labels/model/8

If you look here at the [Model component](https://github.com/empirical-org/Empirical-Core/blob/develop/services/QuillLMS/client/app/bundles/Staff/components/comprehension/semanticRules/model.tsx) which is rendered for the view, you can see that there is already code for this commented out that will satisfy the spec details.

As for adding model notes on creation, in the [ModelForm component](https://github.com/empirical-org/Empirical-Core/blob/develop/services/QuillLMS/client/app/bundles/Staff/components/comprehension/semanticRules/modelForm.tsx), we would want this `TextArea` component there with the placeholder text specified in the spec.

services/QuillLMS/client/app/bundles/Staff/components/comprehension/semanticRules/modelsTable.tsx
eadams17: so we actually already have all of our model API util functions [here](https://github.com/empirical-org/Empirical-Core/blob/develop/services/QuillLMS/client/app/bundles/Staff/utils/comprehension/modelAPIs.ts). You'd want to just build an update one similar to the create. However, we are in the process of updating all these utils to have better error handling. You can see an example here with the [Rule API util functions](https://github.com/empirical-org/Empirical-Core/blob/develop/services/QuillLMS/client/app/bundles/Staff/utils/comprehension/ruleAPIs.ts#L69) and then here in a [handle rule update function](https://github.com/empirical-org/Empirical-Core/blob/develop/services/QuillLMS/client/app/bundles/Staff/components/comprehension/plagiarismRules/plagiarismRulesRouter.tsx#L49) so it would be ideal to match this error handling

services/QuillLMS/client/app/bundles/Staff/components/comprehension/semanticRules/modelsTable.tsx
happythenewsad: > When you pull up the page, you can write directly into the notes (it is a textbox in the editable state). 

I'm pretty sure we want to model notes editable inline. See the above quote and also the mockup in the notion doc.

You're right, I need to add the default text. I'm going to add that after, since I don't want the PR to be held up until I get back on Monday.

eadams17 (changes requested):
If the goal here is to have in-line editable model notes rather than the model note edits be on the view model page, then that portion looks good-- just need to address my comment about how we handle request functions. This is also still missing logic for adding model notes on creation and error handling.

services/QuillLMS/client/app/bundles/Staff/components/comprehension/semanticRules/modelsTable.tsx
happythenewsad: FYI, more discussion in the comments on inline vs. not inline. I'm going with inline for now, since that's what I've already written: https://www.notion.so/quill/Semantic-Labels-Page-Add-Model-Notes-1cf13134be354203a170d7eab0ad8c9e

services/QuillLMS/client/app/bundles/Staff/components/comprehension/semanticRules/modelsTable.tsx
happythenewsad: I used the commented out code stubs, thanks for surfacing that.

services/QuillLMS/client/app/bundles/Staff/components/comprehension/semanticRules/modelsTable.tsx
```diff
@@ -25,11 +29,13 @@ const ModelsTable = ({ activityId, prompt }) => {
         const activateLink = (
           <Link className="data-link" to={`/activities/${activityId}/semantic-labels/${prompt.id}/model/${id}/activate`}>Activate</Link>
         );
+
         return {
           id: id,
           created_at: moment(created_at).format('MM/DD/YY'),
           version: older_models + 1,
           name,
+          notes: stripHtml(model.notes),
```
anathomical: If we strip the HTML here does that reduce the value of the data in this column?  I assume people want it to be viewable here in a similar manner as it is in the editor, but maybe that's too complex to solve right now?

services/QuillLMS/engines/comprehension/app/models/comprehension/automl_model.rb
```diff
@@ -24,7 +24,7 @@ def serializable_hash(options = nil)
       options ||= {}
 
       super(options.reverse_merge(
-        only: [:id, :automl_model_id, :name, :labels, :state, :prompt_id, :created_at],
+        only: [:id, :automl_model_id, :notes, :name, :labels, :state, :prompt_id, :created_at, :text],
```
anathomical: What's this `text` attribute?

anathomical (approved):
One comment about displaying data in the list table, but this looks fine to me other than that.

services/QuillLMS/engines/comprehension/app/models/comprehension/automl_model.rb
happythenewsad: removed it, thanks

services/QuillLMS/client/app/bundles/Staff/components/comprehension/semanticRules/modelsTable.tsx
happythenewsad: I used stripHtml to avoid rendering raw HTML markup, but you're right that this removes the note's styling. @emilia-friedberg Do we have a way to render HTML within DataTable?

services/QuillLMS/client/app/bundles/Staff/components/comprehension/semanticRules/model.tsx
```diff
@@ -19,11 +19,22 @@ const Model = ({ match }) => {
     queryFn: fetchModel
   });
 
-  // function handleSetModelNotes(text: string){ setModelNotes(text) };
+  const initialNoteValue = !modelData ? '' : modelData.model.notes;
```
emilia-friedberg: We generally want to evaluate for truthiness first, so this should be `modelData ? modelData.model.notes : ''`

services/QuillLMS/client/app/bundles/Staff/components/comprehension/semanticRules/model.tsx
```diff
@@ -19,11 +19,22 @@ const Model = ({ match }) => {
...
+    request.put({
+      url: `${process.env.DEFAULT_URL}/api/v1/comprehension/automl_models/${modelId}`,
+      json: { automl_model: {notes: modelNotes} }, },
+      (error, httpStatus, body) => {
+        if (httpStatus.statusCode === 200 || httpStatus.statusCode === 204) {
+          return
+        }
+        setErrors({status: httpStatus.statusCode, message: httpStatus.statusMessage})
+      });
+  }
```
emilia-friedberg: I think Eric mentioned this, but we've been following a pattern where the request is made in a shared file.

services/QuillLMS/client/app/bundles/Staff/components/comprehension/semanticRules/modelsTable.tsx
emilia-friedberg: You can always do `<div dangerouslySetInnerHTML={{ __html: model.notes }} />` any time you want to render actual HTML. That's a general React thing, not specific to this component.

emilia-friedberg (changes requested):
Echoing Eric and answering your question.

services/QuillLMS/client/app/bundles/Staff/components/comprehension/semanticRules/model.tsx
happythenewsad: addressed

services/QuillLMS/client/app/bundles/Staff/components/comprehension/semanticRules/model.tsx
```diff
@@ -1,29 +1,44 @@
...
+  function onHandleUpdateModel() {
+    updateModel(modelId, modelNotes).then((response) => {
+      const { error } = response;
+      
+      if(error) {
+        const updatedErrors = {};
+        updatedErrors['Model Submission Error'] = error;
+        setErrors(updatedErrors);
+      } else {
+        queryCache.refetchQueries(`model-${modelId}`);
```
happythenewsad: @emilia-friedberg I patterned this code off of https://github.com/empirical-org/Empirical-Core/blob/develop/services/QuillLMS/client/app/bundles/Staff/components/comprehension/semanticRules/semanticLabelForm.tsx#L146

However, `refetchQueries` does not seem to be refetching the desired query:  once the user is navigated to the index page and clicks "View Model" again, the old note state is displayed. Only by refreshing the Model View page will the correct updated note be displayed. Any theories on why this is happening?

emilia-friedberg: I think @eadams17 had a relevant realization about `refetchQueries` recently, he can explain better than I can if you ask him!

eadams17: yeah I think you need to update this line to be `queryCache.refetchQueries([`model-${modelId}`, modelId])` and it should work

eadams17: so `refetchQueries` has been phased out. The docs show that there are two options. The first is a refetch function extracted from `useQuery`: https://react-query.tanstack.com/reference/useQuery which probably is the preferred option here since you have access to the model data query above. If you didn't, `queryCache.clear()` would be the other option: https://react-query.tanstack.com/reference/QueryCache

happythenewsad: Thanks; the second option worked.

emilia-friedberg (approved):

PR #7727, opened 2021-05-28
time tracking for Comprehension

emilia-friedberg:
## WHAT
Track how long it takes for students to read the passage and answer each individual prompt in Comprehension, saving to the `ActiveActivitySession` record on the backend after each submission. Save to `ActivitySession` when done, and also start saving anonymous `ActivitySession`s for turk sessions.

## WHY
This will be useful for analyzing student sessions. It is also required by our partners Mathematica (Gates Foundation)

## HOW
Set a start time in the state and add the increment between that time and the current time to the counter for the active step on every action. When the user goes idle for 30 seconds, we stop the timer; when the user switches tabs we also stop.

Please note: right now I am saving the `time_tracking` hash as json in the `hstore` `data` column on the `ActivitySession` table. We may adjust this if we decide to migrate that column.

### Screenshots

### Notion Card Links

PR Checklist | Your Answer
------------ | -------------

### DIFF
--- a/services/QuillLMS/app/controllers/api/v1/activity_sessions_controller.rb
+++ b/services/QuillLMS/app/controllers/api/v1/activity_sessions_controller.rb
@@ -95,8 +95,7 @@ class Api::V1::ActivitySessionsController < Api::ApiController
                   :activity_uid,
                   :activity_id,
                   :anonymous,
-                  :temporary,
-                  :data
+                  :temporary
                 )
       .merge(data: @data).reject {|k,v| v.nil? }
       .merge(timespent: @activity_session&.timespent)
--- a/services/QuillLMS/app/controllers/api/v1/activity_sessions_controller.rb
+++ b/services/QuillLMS/app/controllers/api/v1/activity_sessions_controller.rb
@@ -1,9 +1,9 @@
 class Api::V1::ActivitySessionsController < Api::ApiController
 
   before_action :doorkeeper_authorize!, only: [:destroy]
+  before_action :transform_incoming_request, only: [:update, :create]
   before_action :find_activity_session, only: [:show, :update, :destroy]
   before_action :strip_access_token_from_request
-  before_action :transform_incoming_request, only: [:update, :create]
 
   def show
     render json: @activity_session, meta: {status: 'success', message: nil, errors: nil}, serializer: ActivitySessionSerializer
@@ -78,7 +78,8 @@ class Api::V1::ActivitySessionsController < Api::ApiController
   end
 
   private def find_activity_session
-    @activity_session = ActivitySession.unscoped.find_by_uid!(params[:id])
+    @activity_session = ActivitySession.unscoped.find_by_uid(params[:id]) || ActivitySession.new(activity_session_params.except(:id, :concept_results))
+    @activity_session.uid = params[:id]
   end
 
   private def activity_session_params
@@ -94,7 +95,9 @@ class Api::V1::ActivitySessionsController < Api::ApiController
                   :activity_uid,
                   :activity_id,
                   :anonymous,
-                  :temporary)
+                  :temporary,
+                  :data
+                )
       .merge(data: @data).reject {|k,v| v.nil? }
       .merge(timespent: @activity_session&.timespent)
   end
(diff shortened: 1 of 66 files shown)


services/QuillLMS/client/app/bundles/Comprehension/components/studentView/container.tsx
```diff
@@ -92,12 +101,62 @@ export class StudentViewContainer extends React.Component<StudentViewContainerPr
...
+    const lastAttempt = responsesForPrompt[responsesForPrompt.length - 1]
+
+    return (responsesForPrompt.length === activePrompt.max_attempts) || lastAttempt.optimal
   }
 
+  resetTimers = (e=null) => {
+    this.setState((prevState, props) => {
+      const { activeStep, startTime, timeTracking, isIdle, inactivityTimer, completedSteps, } = prevState
+      if (completedSteps.includes(activeStep)) { return } // don't want to add time if a user is revisiting a previously completed step
+      if (this.outOfAttemptsForActivePrompt()) { return }// or if they are finished submitting responses for the current active step
```
eadams17, line 142: appreciate these comments!

eadams17 (approved):
LGTM!

services/QuillLMS/app/controllers/api/v1/activity_sessions_controller.rb
```diff
@@ -94,7 +95,9 @@ def destroy
                   :activity_uid,
                   :activity_id,
                   :anonymous,
-                  :temporary)
+                  :temporary,
+                  :data
```
anathomical: I' actually not sure how this would work given line 87 above.  I presume that it does since you've done testing, but maybe if we're going to access the value from `params` we should remove the `@data = params.delete(:data)` line since it provides a second path to access the data?

services/QuillLMS/client/app/bundles/Comprehension/components/studentView/container.tsx
```diff
@@ -92,12 +101,62 @@ export class StudentViewContainer extends React.Component<StudentViewContainerPr
...
+      if (inactivityTimer) { clearTimeout(inactivityTimer) }
+
+      let elapsedTime = _.round((Date.now() - startTime) / 1000)
+      if (isIdle) {
+        elapsedTime = 0
+      }
+      const newTimeTracking = {...timeTracking, [activeStep]: timeTracking[activeStep] + elapsedTime}
+      const newInactivityTimer = setTimeout(this.setIdle, 30000);  // time is in milliseconds (1000 is 1 second)
+
+      return { timeTracking: newTimeTracking, isIdle: false, inactivityTimer: newInactivityTimer, startTime: new Date(), }
```
anathomical: This is minor enough to not actually matter for practical purposes, but I figured I'd call it out for completeness sake: we use `Date.now()` for the elapsedTime calculation, and then we use it again here as a new call for `startTime`.  Technically, some very small number of ms (1 or 2, maybe?) will often pass between these two calls, meaning that we "lose" the time between the two invocations.  1 or 2 ms, even if we were tracking across thousands of events, probably doesn't matter to us, but it might be worth refactoring to call `Date.now()` once at the top of this function and use the result obtained there for both the `elapsedTime` calculation and the new value of `startTime`.

If you feel like it.

services/QuillLMS/client/app/bundles/Comprehension/components/studentView/container.tsx
```diff
@@ -92,12 +101,62 @@ export class StudentViewContainer extends React.Component<StudentViewContainerPr
     }
 
     window.addEventListener('keydown', this.handleKeyDown)
+    window.addEventListener('mousemove', this.resetTimers)
+    window.addEventListener('mousedown', this.resetTimers)
+    window.addEventListener('click', this.resetTimers)
+    window.addEventListener('keypress', this.resetTimers)
+    window.addEventListener('visibilitychange', this.setIdle)
```
anathomical, line 108: I super-love this approach of just attaching to standard event listeners for tracking and just writing a couple of functions.  Good call.

anathomical (approved):
A couple of thoughts for minor changes, but generally speaking I'm really impressed with how compact this is.  Very nice job!

services/QuillLMS/app/controllers/api/v1/activity_sessions_controller.rb
emilia-friedberg: You're right, I think this addition actually isn't doing anything - data comes back in the `.merge` below but is never present at this point. I'll remove it.

services/QuillLMS/client/app/bundles/Comprehension/components/studentView/container.tsx
emilia-friedberg: Good point. I'll update.

PR #7728, opened 2021-05-28
LMS Rails 5.0 Prework - Refactor ActiveRecord::Base.connection.execute calls to use TypeMapByOID, Part 2

brendanshean:
## WHAT
This is part 2 of https://github.com/empirical-org/Empirical-Core/pull/7711

There are 100+ occurrences of `ActiveRecord::Base.connection.execute` and this is another batch of replacements.
Refactor `ActiveRecord::Base.connection.execute` calls in Rails 4 into a new call RawSqlRunner.execute that fills in the appropriate TypeMap so that the resulting PG:Result matches the one in Rails 5.

## WHY
ActiveRecord::Base.connection.execute yields a PG::Result. Calling to_a on that PG::Result gives different values in Rails 4 and Rails 5. This seems to be due to the TypeMap associated with the PG::Result which is used in casting values.

## HOW

Large diff in `services/QuillLMS/app/controllers/teachers/classroom_manager_controller.rb` is due to some dead code.

### Screenshots

### Notion Card Links
https://www.notion.so/quill/Refactor-ActiveRecord-Base-connection-execute-calls-to-use-TypeMapByOID-5b1a320b11824b7088a63458667891e8

PR Checklist | Your Answer
------------ | -------------

### DIFF
--- a/services/QuillLMS/app/controllers/teachers/progress_reports/activity_sessions_controller.rb
+++ b/services/QuillLMS/app/controllers/teachers/progress_reports/activity_sessions_controller.rb
@@ -137,21 +137,31 @@ class Teachers::ProgressReports::ActivitySessionsController < Teachers::Progress
         }
       end
     else
-      csv_string = CSV.generate do |csv|
-        csv << ['Student', 'Date', 'Activity', 'Score', 'Standard', 'Tool']
-        activity_sessions.map do |session|
-          score = (session['percentage'] == -1.0 ? 'Completed' : "#{session['percentage'] * 100}%")
-          csv << [
-            session['student_name'],
-            session['visual_date'],
-            session['activity_name'],
-            score,
-            session['standard'],
-            session['activity_classification_name']
-          ]
-        end
+      render text: csv_string(activity_sessions)
+    end
+  end
+
+  private def score(percentage)
+    case percentage
+    when nil then ''
+    when -1 then 'Completed'
+    else "#{percentage * 100}%"
+    end
+  end
+
+  private def csv_string(activity_sessions)
+    CSV.generate do |csv|
+      csv << ['Student', 'Date', 'Activity', 'Score', 'Standard', 'Tool']
+      activity_sessions.map do |session|
+        csv << [
+          session['student_name'],
+          session['visual_date'],
+          session['activity_name'],
+          score(session['percentage']),
+          session['standard'],
+          session['activity_classification_name']
+        ]
       end
-      render text: csv_string
     end
   end
 end
--- a/services/QuillLMS/app/controllers/teachers/progress_reports/activity_sessions_controller.rb
+++ b/services/QuillLMS/app/controllers/teachers/progress_reports/activity_sessions_controller.rb
@@ -143,7 +143,7 @@ class Teachers::ProgressReports::ActivitySessionsController < Teachers::Progress
 
   private def score(percentage)
     case percentage
-    when nil then ''
+    when nil then 0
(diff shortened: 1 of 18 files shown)


services/QuillLMS/app/controllers/teachers/progress_reports/activity_sessions_controller.rb
```diff
@@ -134,7 +140,7 @@ def index
       csv_string = CSV.generate do |csv|
         csv << ['Student', 'Date', 'Activity', 'Score', 'Standard', 'Tool']
         activity_sessions.map do |session|
-          score = (session['percentage'].to_f == -1.0 ? 'Completed' : "#{session['percentage'].to_f*100}%")
+          score = (session['percentage'] == -1.0 ? 'Completed' : "#{session['percentage'] * 100}%")
```
dandrabik: Looking at this query, `(CASE WHEN activity_classifications.scored THEN activity_sessions.percentage ELSE -1 END) AS percentage`, bad data could leave `percentage` as `nil`, so that `to_f` might also be a `nil` cover. Might be worth leaving here or confirming that doesn't happen in the data.

services/QuillLMS/app/controllers/teachers/progress_reports/activity_sessions_controller.rb
```diff
@@ -45,76 +45,82 @@ def index
...
+            ON activity_sessions.classroom_unit_id = classroom_units.id
+            AND activity_sessions.state = 'finished'
+            AND activity_sessions.visible = true
+            AND activity_sessions.user_id = users.id
+            #{student_filter}
+          WHERE classrooms_teachers.user_id = #{current_user.id}
+        SQL
+      ).to_a[0]['count']
 
+      page_count = (count.to_f / PAGE_SIZE).ceil
```
dandrabik, line 123: I like how you made this a separate line, much easer to parse!

dandrabik (approved):
Looks like there are failing tests and I made one small comment about a possible `nil` check, but outside of those two things, looks good to me!

services/QuillLMS/app/controllers/teachers/progress_reports/activity_sessions_controller.rb
brendanshean: Good point.  I'm working on a fixing a potentially similar thing with FRL here:
Screen Shot 2021-05-30 at 7 05 33 AM

brendanshean: I see the issue.  We're basically coercing nils to 0%.  Wouldn't it be more appropriate to leave [blank](https://github.com/empirical-org/Empirical-Core/pull/7728/commits/4c0f59ae074129ae41d837f2622643d6bc617a00#diff-65c6f4ff88cd4b2fdb828b1f6a58e98d0361c8b2afab05e1840065a3348508c8R144)?

dandrabik: It might be more appropriate, depends on the report usage and tool (that's more of a product decision). To keep this PR scoped, it might be better to keep the functionality as it is, and to make a new task card to determine if that product change (0% -> blank) should be made (You can add it to 'Tasks').

brendanshean: I'll leave the existing functionality then.

dandrabik (approved):
This LGTM :+1:

PR #7734, opened 2021-06-01
convert activity session data column to jsonb

emilia-friedberg:
## WHAT
Convert the `data` column on `ActivitySession` from hstore to jsonb, remove the JSONification of the nested `time_tracking` hash since we no longer need that hack, and add a rake task to update any `ActivitySession` records that have been made with the JSON-ed `time_tracking` because that will wind up being doubly-JSONed and thus unqueryable once the migration is run.

## WHY
Storing a json string as the value for a nested hash in hstore is hacky, and having this column be jsonb will make it much more flexible and functional for future uses.

## HOW
Add the migration and the back-migration (tested; both work). Please note that the migration takes about 40 minutes to run. Once it has completed, we will need to run the rake task to make sure the SQL queries for time tracking can still function.

### Screenshots
N/A

### Notion Card Links
https://www.notion.so/quill/Migrate-activity-sessions-data-column-to-jsonb-instead-of-hstore-2eb27e98397841a59e3e86b55cb9701f

PR Checklist | Your Answer
------------ | -------------


services/QuillLMS/lib/tasks/convert_activity_session_timetracking_data.rake
```diff
@@ -0,0 +1,12 @@
+namespace :activity_session_timetracking_data do
+  desc 'updates data that had formerly been saved as json in hstore to be parsable in jsonb'
+  task :convert => :environment do
+    ActivitySession.where("activity_sessions.data->'time_tracking' IS NOT NULL").each do |as|
+      if as.data['time_tracking'].instance_of? String
```
brendanshean, line 5: I would double-check the type of `time_tracking` when accessing `data['time_tracking']`; namely string vs symbol:

Screen Shot 2021-06-02 at 7 38 54 AM

Screen Shot 2021-06-02 at 7 35 37 AM

services/QuillLMS/db/migrate/20210528142650_convert_activity_session_data_column_from_hstore_to_jsonb.rb
```diff
@@ -0,0 +1,22 @@
+class ConvertActivitySessionDataColumnFromHstoreToJsonb < ActiveRecord::Migration
+  def up
+    change_column :activity_sessions, :data, :jsonb, using: 'data::jsonb'
+  end
+
+  def down
+    # custom function from here: https://stackoverflow.com/questions/24259531/migrate-json-column-type-to-hstore-column-type, amended for jsonb
```
brendanshean, line 7: Good find here.

brendanshean (approved):
Minor comment on rake task for double checking something.  Otherwise good to go!  I think this approach looks great.

services/QuillLMS/lib/tasks/convert_activity_session_timetracking_data.rake
emilia-friedberg: Thanks for asking! I did run this locally, with real data from having played through a bunch of Comprehension sessions, and it did work so I feel confident about it.

emilia-friedberg: I think the difference is that when you copy and paste the rails output into your console, it's just a hash; in this script it's a `ActiveRecord::ConnectionAdapters::PostgreSQL::OID::Jsonb`, which has string keys.

dandrabik (approved):
:+1: LGTM, thanks for putting this in place!

services/QuillLMS/lib/tasks/convert_activity_session_timetracking_data.rake
brendanshean: Interesting.  Thank you for resolving confusing.

services/QuillLMS/lib/tasks/convert_activity_session_timetracking_data.rake
```diff
@@ -0,0 +1,12 @@
...
+        time_tracking = JSON.parse(as.data['time_tracking'])
+        as.data = { time_tracking: time_tracking }
```
anathomical, line 7: If I remember correctly, only Comprehension activities should have `time_tracking` as a key in `data`, and they have nothing else in `data`, so it's safe to just overwrite the values in `data` for these records completely.  Just want to confirm that that's true just in case what we'd rather do is something like `as.data['time_tracking'] = time_tracking` or something.

anathomical (approved):

PR #7741, opened 2021-06-02
Comprehension/make go deployable

anathomical:
## WHAT
Refactor the Go endpoint so that it is deployable to Heroku.  Also add a bit of scaffolding to make it easy to run the Go endpoint locally.
NOTE: I'm pretty sure that this will make this code undeployable for the purposes of Google Cloud Functions, so this could be considered a breaking change while the architecture is in transition.
## WHY
We would like more devops control over our critical infrastructure, and being able to deploy and intentionally scale the Go endpoint improves our control in this space.
Additionally, we would like to be able to better test tweaks to the Go endpoint locally, and refactoring it to run in Heroku also makes it very easy to run on a local development machine.
## HOW
Basically this is just a refactor that routes the existing `Endpoint` function through `gin` which is a lightweight Go web framework.  There were a couple of changes made to the `Endpoint` code since `gin` endpoints have different interfaces than Google Cloud Function endpoints, but no behavior changes exist.  Additionally, the test suite needed a couple of minor tweaks to trigger the new `Endpoint` signature, but those are straightforward.

### Notion Card Links
https://www.notion.so/quill/Go-Endpoint-hosting-65579883fbfb458293ed1ae875263093
https://www.notion.so/quill/Add-Go-Endpoint-setup-to-developers-local-environment-14d69ad62668468ab2fb912de6320c6f

PR Checklist | Your Answer
------------ | -------------


services/comprehension/feedback-api-main/src/endpoint/endpoint_test.go
```diff
@@ -21,7 +22,10 @@ func TestPublishMessage(t *testing.T) {
 
 	rr := httptest.NewRecorder()
 	req := httptest.NewRequest("POST", "/", json_string)
-	Endpoint(rr, req)
+
+	servrer := gin.Default()
```
dandrabik: typo? `server`

services/comprehension/feedback-api-main/src/endpoint/endpoint.go
```diff
@@ -77,43 +77,33 @@ func AssembleUrls() ([8]string) {
 	return urls
 }
 
-func Endpoint(responseWriter http.ResponseWriter, request *http.Request) {
+func Endpoint(c *gin.Context) {
```
dandrabik: Seeing this code again, I wonder if we should decide to avoid 1-letter variables. l think it's more common in go conventions (I also used one letter variables here). Curious your thoughts.

services/comprehension/feedback-api-main/src/endpoint/main.go
```diff
@@ -0,0 +1,20 @@
+package main
+
+import (
+	"net/http"
+	"github.com/gin-gonic/gin"
+)
+
+func main() {
+	router := gin.Default()
+	router.OPTIONS("/", func(c *gin.Context) {
```
dandrabik, line 27: Oh, it's nice we can set this outside of the endpoint method.

dandrabik (approved):
Small comments, but this looks good to me. It's unfortunate that it's a breaking change for Cloud Endpoints, but being able to run this locally will be really helpful. Thanks for putting this together.

services/comprehension/feedback-api-main/src/endpoint/endpoint.go
anathomical: I was just following the `gin` examples, but now that you call this out, I strongly agree.  I'll tweak this to call it `context` since that's what it is.  I strongly prefer readability over concision here, so good call.

dandrabik (approved):
:+1:

happythenewsad (approved):

happythenewsad:
Please socialize the breaking change to the team. We can push a tag for the last known working Google Cloud commit, and declare that the cloud function is in maintenance mode after that.

PR #7742, opened 2021-06-02
comprehension/stop-sending-unused-feedback-histories

anathomical:
## WHAT
Replace batch feedback storage with only saving used feedback
## WHY
Since we never actually do reports on unused feedback, we don't need to store information about it
## HOW
- Refactor the code that posts FeedbackHistory records in the Go endpoint to post single entries to the base `create` route in LMS instead
- Remove the code in Go that handled batching FeedbackHistory records since we no longer need to do that
- Make a small tweak to the LMS permit params so that batch and single create both have the same params

### Notion Card Links
https://www.notion.so/quill/Turn-Off-Recording-Unused-Feedback-History-1fd6aa1cc0264310bddd7deff3c5959d

PR Checklist | Your Answer
------------ | -------------
Have you added and/or updated tests? |  Yes, mostly in the form of removal

### DIFF
--- a/services/QuillLMS/app/controllers/api/v1/feedback_histories_controller.rb
+++ b/services/QuillLMS/app/controllers/api/v1/feedback_histories_controller.rb
@@ -51,7 +51,7 @@ class Api::V1::FeedbackHistoriesController < Api::ApiController
       :used,
       :time,
       :rule_uid,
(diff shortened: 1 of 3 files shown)


dandrabik (approved):
:+1: LGTM.

PR #7743, opened 2021-06-02
Feature/port minitest to rspec

happythenewsad:
## WHAT
Ports minitest tests and configs to rspec for Comprehension

## WHY
So we only need to manage one test suite across our apps 

## HOW
Used `mt2rspec` gem to get 80% of the code ported. This converter isn't perfect, so some custom re-writing was required.
`mt2rspec`'s rspec style differs slightly from ours:
- it uses parens in it statements: `it('should foo...')`
- it does not add a line of whitespace between rspec blocks
- it uses instance variables in before blocks instead of 'let'

I advocate to merge this PR soon and discuss style fixes afterward (or just leave it). The longer this PR is open, the more we risk accidentally deleting recently-commited minitests (and will need to port each one by hand to rspec).

### Screenshots

### Notion Card Links
https://www.notion.so/quill/Port-Comprehension-Engine-tests-to-Rspec-ef99720812714b15877afa5b8c2289c2

PR Checklist | Your Answer
------------ | -------------
Have you deployed to Staging? | n/a


services/QuillLMS/engines/comprehension/spec/controllers/comprehension/activities_controller_spec.rb
```diff
@@ -0,0 +1,192 @@
+require("rails_helper")
+
+module Comprehension
+  RSpec.describe(ActivitiesController, :type => :controller) do
+    before { @routes = Engine.routes }
+    context 'foo' do
```
happythenewsad: runs correctly via `bundle exec rspec spec/controllers/comprehension/activities_controller_spec.rb:6 `

happythenewsad:
Note: I squashed many commits to make a gnarly rebase less gnarly. I have saved an archive branch locally in case any wants to inspect it later.

happythenewsad:
~I believe the failing specs after the rebase are caused by the rails 5.1 upgrade. UPDATE: I've had to reconcile merge conflicts twice due to the rails 5.1 commit reverting and unreverting, so I'm going to wait for the final reconciliation until after 5.1 is merged into develop permanently. This PR can still be reviewed now; all specs were passing before the most recent reconciliation.~

Update: everything is working now that I've rebased against develop, which now includes rails 5.1

@dandrabik @anathomical

services/QuillLMS/engines/comprehension/spec/rails_helper.rb
```diff
@@ -0,0 +1,66 @@
...
+  config.include FactoryBot::Syntax::Methods
+  config.infer_spec_type_from_file_location!
+  config.mock_with :rspec do |mocks|
+    mocks.verify_partial_doubles = true
+  end
+end
+
+factory_files = Dir.glob('spec/factories/comprehension/*')
+
+factory_files.each {|f| require File.expand_path(File.join(Rails.root, '..', '..', f))}
```
brendanshean, line 59: Another approach for settting factories:
```
FactoryBot.definition_file_paths << File.join(File.dirname(__FILE__), 'factories')
FactoryBot.find_definitions
```

brendanshean (approved):
This commit makes me super happy.  Great job porting this over.

I agree with getting this merged in sooner rather than later.  I'm happy to write a regex to fix some of those stylistic things later.

services/QuillLMS/engines/comprehension/spec/controllers/comprehension/activities_controller_spec.rb
```diff
@@ -0,0 +1,186 @@
+require("rails_helper")
+module Comprehension
+  RSpec.describe(ActivitiesController, :type => :controller) do
+    before { @routes = Engine.routes }
+    context("index") do
+      it("return successfully - no activities") do
+        get(:index)
+        parsed_response = JSON.parse(response.body)
+        assert_response(:success)
```
dandrabik: There's still some `assert_responses` in this file.

dandrabik (approved):
Thanks for porting these over. We definitely need to clean up the syntax here (the items you mentioned). I don't think that would take very long (maybe an hour) since the spacing and parentheses changes can be done by regex replace.

So you can either:
1. Merge this and quickly follow with a syntax cleanup PR today (by this afternoon).
2. Make the syntax changes now and merge one PR.

Given this PR has been worked on for about 3 weeks, I don't know if an extra hour is that big of a deal in either direction, so those choices are basically equivalent to me (personally, 2 seems easier to me since it will be harder to miss things since the full change diff is here). Note, I just notified the quill-developer channel to hold off on writing new tests in the engine until this is merged to allow you either one of those options.

So I'll defer to you on which path to proceed, but either way, the end result should be rspec files, with proper syntax by the EOD. 🎉

services/QuillLMS/engines/comprehension/spec/dummy/app/models/activity_classification.rb
```diff
@@ -1,4 +1,4 @@
-class ActivityClassification < ApplicationRecord
+class ActivityClassification < ActiveRecord::Base
```
anathomical: I thought that `ApplicationRecord` was the new Rails 5 approach.  Do we want to go back to the old way here?

services/QuillLMS/engines/comprehension/spec/controllers/comprehension/activities_controller_spec.rb
anathomical: Looks like we're still using `assert_response` in most of the controller specs, and `assert_operator` in the `activity` model spec.

anathomical (approved):
A couple of comments, but this looks good to me.  Really excited to see this just about ready.

services/QuillLMS/engines/comprehension/spec/dummy/app/models/activity_classification.rb
happythenewsad: Good catch - this PR was created before Rails 5.1 was merged, and the rebase did not clobber the old because I used `cp -R` to create `spec/` from `test/`

happythenewsad:
Ok, the style issues are addressed.

PR #7735, opened 2021-06-01
security patches for LMS and CMS

happythenewsad:
## WHAT
Adds high-severity security patches to QuillLMS and QuillCMS. The most impactful change is bumping CMS puma from 3.x to 4.x

## WHY
We don't want security vulns. 

### Screenshots
(If applicable. Also, please censor any sensitive data)

### Notion Card Links
https://www.notion.so/quill/dependabot-patches-high-priority-part-1-2-7d89ef703c8a43778328def889d7c9dc

PR Checklist | Your Answer
------------ | -------------
Have you added and/or updated tests? |  no - no code change
Have you deployed to Staging? | yes
Self-Review: Have you done an initial self-review of the code below on Github? | yes
Design Review: If applicable, have you compared the coded design to the mockups? | n/a

PR #7770, opened 2021-06-07
Secpatch/high1 2nd attempt

happythenewsad:
## WHAT
This is an extension of #7735, which was reverted due to a prod issue. The only addition in this PR is tweaking `sidekiq` and `sidekiq-pro` versions, so that we don't upgrade to an extend that it breaks our production sidekiq workers. Because `sidekiq` and `sidekiq-pro` versions are not changed in this PR, we can be reasonably confident that it will not cause a production issue like the last, reverted PR caused.

The other gem and npm specifications have already been tested.

### Screenshots

### Notion Card Links

PR Checklist | Your Answer
------------ | -------------
Have you added and/or updated tests? |  no


services/QuillCMS/Gemfile.lock
```diff
@@ -140,37 +148,37 @@ GEM
...
-      slop (~> 3.4)
+    pry (0.14.1)
+      coderay (~> 1.1)
+      method_source (~> 1.0)
     public_suffix (4.0.6)
-    puma (3.12.6)
-    pusher (1.4.3)
+    puma (4.3.8)
+      nio4r (~> 2.0)
+    pusher (2.0.1)
```
dandrabik: There's a major version bump of Pusher (1.4.3 -> 2.0.1). Is this necessary? Has this been tested?

https://github.com/pusher/pusher-http-ruby/blob/master/CHANGELOG.md

dandrabik (changes requested):
Can we remove the Pusher update from this version bump? If not, can we confirm that this has been tested.

services/QuillCMS/Gemfile.lock
happythenewsad: Looks like newrelic_rpm and method_source were bumped significantly as well. I'll attempt to pin them to their previous versions, but will fall back to manual testing if I can't.

happythenewsad:
I was able to pin all three gems so that they didn't change versions.

services/QuillCMS/Gemfile
```diff
@@ -57,7 +58,7 @@ end
 group :development, :test do
   gem 'byebug', platforms: %i[mri mingw x64_mingw]
   gem 'factory_bot_rails', '~> 4.8.2'
-  gem 'pry'
+  gem 'pry', '~> 0.10.4'
```
dandrabik, line 61: This is a `development` only gem, so we don't need to pin it.

dandrabik (approved):
Small comment about unpinning a development gem, but :+1:

services/QuillCMS/Gemfile
happythenewsad: Actually, unpinning will cause bundler to install a high version of `method-source`, and `method-source` is used by production code via the railties gem.

dandrabik: Fine to pin then. :ship:

anathomical (approved):

PR #7779, opened 2021-06-09
Back end logic for Comprehension Change Log

cissyxyu:
## WHAT
Curriculum requested a change log feature where they can see update, create, and deletion action for around 25 different possible actions performed on Comprehension Activities, Prompts, and Rules. This is my attempt to write those actions into the back end.

## WHY
They want to be able to track staff changes to Comprehension models.

## HOW
In Comprehension's `ActivitiesController` and `RuleController`, I added `before_action` methods to parse the incoming params and save those params that are about to be changed (because we can't see their old values after updating). 

After saving those params, the CRUD action is called. If successful, the controller then calls `log_update` or `log_create` or `log_deletion` on all the instances that just got updated, including the nested instances (e.g. `RegexRule` which is nested under `Rule`).

### Screenshots

### Notion Card Links
https://www.notion.so/quill/Create-an-activity-change-log-2f7563fbef0d42be84ad733676689e08

PR Checklist | Your Answer
------------ | -------------


services/QuillLMS/app/models/change_log.rb
```diff
@@ -67,13 +67,39 @@ class ChangeLog < ActiveRecord::Base
...
+    delete_semantic: 'Semantic Label - deleted',
+    update_semantic: 'Semantic Label - updated',
+    update_feedback_1: 'Semantic Label First Layer Feedback - updated',
+    update_highlight_1: 'Semantic Label First Layer Feedback Highlight - updated',
+    update_feedback_2: 'Semantic Label Second Layer Feedback - updated',
+    update_highlight_2: 'Semantic Label Second Layer Feedback Highlight - updated',
+    create_plagiarism: 'Plagiarism - created',
+    update_plagiarism: 'Plagiarism - updated',
+    update_universal: 'Universal Rule - updated',
+    create_universal: 'Universal Rule - created'
```
anathomical, line 91: I know that this is pretty explicitly what is being asked for, but this list is so extensive and detailed that it feels out of place in a "generic" logging tool.

services/QuillLMS/engines/comprehension/app/controllers/comprehension/activities_controller.rb
```diff
@@ -4,6 +4,7 @@ module Comprehension
   class ActivitiesController < ApplicationController
     skip_before_action :verify_authenticity_token
     before_action :set_activity, only: [:show, :update, :destroy, :rules]
+    before_action :save_nested_vars_for_log, only: [:update, :create]
```
anathomical, line 7: Not a huge deal, but because you're explicitly calling `log_nested_changes` inside both `create` and `update`, and given that the values set by `save_nested_vars_for_log` seem to only be used in `log_nested_changes`, it seems like making this a `before_action` is unnecessary.  You could just call `save_nested_vars_for_log` at the top of `log_nested_changes`.

It's a pretty minor detail, but it seems lightly less complicated to avoid the `before_action` since you have to explicitly call functions in the actions anyway.

services/QuillLMS/engines/comprehension/app/controllers/comprehension/rules_controller.rb
```diff
@@ -2,6 +2,7 @@ module Comprehension
   class RulesController < ApplicationController
     skip_before_action :verify_authenticity_token
     before_action :set_rule, only: [:show, :update, :destroy]
+    before_action :save_nested_vars_for_log, only: [:create, :update]
```
anathomical, line 5: Same note here as in the `activities_controller` file.  This probably doesn't need to be a `before_action`.

services/QuillLMS/engines/comprehension/app/controllers/comprehension/rules_controller.rb
```diff
@@ -21,8 +22,9 @@ def show
     # POST /rules.json
     def create
       @rule = Comprehension::Rule.new(rule_params)
-
       if @rule.save
+        @rule.log_creation(lms_user_id)
+        @rule.label.log_creation(lms_user_id) if @rule.label.present?
```
anathomical, line 27: I actually like your syntactic approach on line 49 for this more.  `@rule.label&.log_creation(lms_user_id)` reads better than using the `if` clause.

services/QuillLMS/engines/comprehension/app/controllers/comprehension/rules_controller.rb
```diff
@@ -59,5 +65,96 @@ def destroy
...
+      @feedback_vals = []
+      @highlights_vals = []
+      @plagiarism_text_vals = []
+      @regex_rules_vals = []
+      @rule_vals = []
+    end
+
+    private def save_label_vars_for_log
+      label = Comprehension::Rule.find_by_id(params[:id])&.label
+      if label.present?
```
anathomical: This seems like a good place for an early return.  If you `return unless label.present?` here you don't need to wrap the rest of the code in an `if` block.

services/QuillLMS/engines/comprehension/app/controllers/comprehension/rules_controller.rb
```diff
@@ -59,5 +65,96 @@ def destroy
...
+
+    private def save_regex_rules_vars_for_log
+      rule_params[:regex_rules_attributes]&.each do |rr|
+        old_regex = Comprehension::RegexRule.find_by_id(rr[:id])&.regex_text
+        @regex_rules_vals.push({id: rr[:id], regex_text: old_regex || rr[:regex_text]}) if rr[:regex_text] && rr[:regex_text] != old_regex
+      end
+    end
+
+    private def save_plagiarism_text_vars_for_log
+      if rule_params[:plagiarism_text_attributes]&.key?(:text)
```
anathomical, line 108: Again, this could be a nice early return.

services/QuillLMS/engines/comprehension/app/controllers/comprehension/rules_controller.rb
```diff
@@ -59,5 +65,96 @@ def destroy
...
+        pt = rule_params[:plagiarism_text_attributes]
+        old_text = Comprehension::PlagiarismText.find_by_id(pt[:id])&.text
+        @plagiarism_text_vals.push({id: pt[:id], text: old_text || pt[:text]}) if pt[:text] && pt[:text] != old_text
+      end
+    end
+
+    private def save_rule_vars_for_log
+      rule = Comprehension::Rule.find_by_id(params[:id])
+      if rule.present?
```
anathomical, line 117: Early return again.

services/QuillLMS/engines/comprehension/app/controllers/comprehension/rules_controller.rb
```diff
@@ -59,5 +65,96 @@ def destroy
...
+        end
+      end
+    end
+
+    private def log_nested_changes
+      @feedback_vals.each do |f|
+        if f[:id]
+          Comprehension::Feedback.find(f[:id]).log_update(lms_user_id, "#{f[:label_string]}\n#{f[:text]}")
+        else
+          Comprehension::Feedback.find_by(text: f[:text])&.log_update(lms_user_id, f[:label_string])
```
anathomical, line 133: I would be concerned that this could get multiple results in cases where we have, say, a Regex Rule that is used in multiple activities.  `find_by` will take the first item it finds with no sense of ordering, so you could easily be logging "updates" to the wrong Feedback record.

At least I think so.

services/QuillLMS/engines/comprehension/app/controllers/comprehension/rules_controller.rb
```diff
@@ -59,5 +65,96 @@ def destroy
...
+          Comprehension::Highlight.find(h[:id]).log_update(lms_user_id, "#{h[:label_string]}\n#{h[:text]}")
+        else
+          Comprehension::Highlight.find_by(text: h[:text])&.log_update(lms_user_id, h[:label_string])
+        end
+      end
+      @regex_rules_vals.each do |r|
+        if r[:id]
+          Comprehension::RegexRule.find(r[:id]).log_update(lms_user_id, r[:regex_text])
+        else
+          Comprehension::RegexRule.find_by(regex_text: r[:regex_text])&.log_update(lms_user_id, nil)
```
anathomical, line 147: Again, I'd be worried about duplicate Regex rules that are used in multiple Activities.

services/QuillLMS/engines/comprehension/app/controllers/comprehension/rules_controller.rb
```diff
@@ -59,5 +65,96 @@ def destroy
...
+        end
+      end
+      @plagiarism_text_vals.each do |pt|
+        if pt[:id]
+          Comprehension::PlagiarismText.find(pt[:id]).log_update(lms_user_id, pt[:value])
+        else
+          Comprehension::PlagiarismText.find_by(text: pt[:text])&.log_update(lms_user_id, nil)
```
anathomical, line 154: In theory this should be safe to use, but I know that there's been some discussion by Curriculum about "duplicating" activities to track performance separately.  In cases of duplicates, we'll likely end up with some records that have identical text values, and this lookup could easily pull and update the wrong one.

services/QuillLMS/engines/comprehension/app/controllers/comprehension/rules_controller.rb
```diff
@@ -59,5 +65,96 @@ def destroy
...
+          Comprehension::Feedback.find(f[:id]).log_update(lms_user_id, "#{f[:label_string]}\n#{f[:text]}")
+        else
+          Comprehension::Feedback.find_by(text: f[:text])&.log_update(lms_user_id, f[:label_string])
+        end
+      end
+      @highlights_vals.each do |h|
+        if h[:id]
+          Comprehension::Highlight.find(h[:id]).log_update(lms_user_id, "#{h[:label_string]}\n#{h[:text]}")
+        else
+          Comprehension::Highlight.find_by(text: h[:text])&.log_update(lms_user_id, h[:label_string])
```
anathomical, line 140: Same note as Plagiarism below.  In theory this should be fine, but if Curriculum duplicates activities, we should expect multiple items with the same values.

services/QuillLMS/engines/comprehension/app/models/comprehension/activity.rb
```diff
@@ -51,5 +55,13 @@ def serializable_hash(options = nil)
     private def expire_turking_rounds
       turking_rounds.each(&:expire!)
     end
+
+    def log_creation(user_id)
+      log_change(user_id, :create_activity, self, nil, nil, nil, "Comprehension Activity #{id} - active")
```
anathomical: I'm not sure that there's a cleaner way to do this, but the use of `:create_activity` here is risky.  While Ruby symbols may feel more "static" than a simple string, they're just as prone to typos.  It'd be better if we could somehow make this a constant.  (That might mean making the individual values in the `ChangeLog` model into constants...)
```ruby
COMPREHENSION_ACTIONS = [
  COMPREHENSION_CREATE_ACTIVITY = 'Comprehension Activity - created',
  ...
]
```
You can reference something like this via `ChangeLog::COMPREHENSION_CREATE_ACTIVITY`.

services/QuillLMS/engines/comprehension/app/models/comprehension/activity.rb
```diff
@@ -51,5 +55,13 @@ def serializable_hash(options = nil)
...
+    end
+
+    def log_deletion(user_id)
+      log_change(user_id, :delete_activity, self, nil, nil, "Comprehension Activity #{id} - active", "Comprehension Activity #{id} - deleted")
```
anathomical: Same as above.

services/QuillLMS/engines/comprehension/app/models/comprehension/automl_model.rb
```diff
@@ -113,5 +114,13 @@ def older_models
       model = automl_client.get_model(name: automl_model_full_id)
       model.display_name
     end
+
+    def log_creation(user_id)
+      log_change(user_id, :create_automl, prompt, nil, nil, nil, automl_model_id)
```
anathomical: Same note as in `Activity` about symbols vs constants.

services/QuillLMS/engines/comprehension/app/models/comprehension/feedback.rb
```diff
@@ -20,5 +22,31 @@ def serializable_hash(options = nil)
         include: [:highlights]
       ))
     end
+
+    private def semantic_rule
```
anathomical, line 26: Not that important, but it's Rails convention to name functions that provide boolean responses with `?` at the end.  So this, and the next two functions, should probably be `def semantic_rule?`, `def first_order?`, and `def second_order?`.

services/QuillLMS/engines/comprehension/app/models/comprehension/feedback.rb
```diff
@@ -20,5 +22,31 @@ def serializable_hash(options = nil)
...
+    end
+
+    private def second_order
+      order == 1
+    end
+
+    def log_update(user_id, prev_value)
+      if semantic_rule && first_order
+        rule&.prompts&.each do |prompt|
+          log_change(user_id, :update_feedback_1, prompt, nil, nil, prev_value, "#{rule.label.name} | #{rule.name}\n#{text}")
```
anathomical: This may be something that's actually intended, but won't this cause massive duplication if used for a universal rule?  It already causes a lot of duplication for just a Regex rule that applies to all three prompts, where things get logged three times, but if a universal rule gets this update, it'll get dozens of duplicates.

services/QuillLMS/engines/comprehension/app/models/comprehension/highlight.rb
```diff
@@ -21,5 +23,29 @@ def serializable_hash(options = nil)
         only: [:id, :feedback_id, :text, :highlight_type, :starting_index]
       ))
     end
+
+    private def semantic_rule
```
anathomical, line 27: Same notes as in `Feedback` above.

services/QuillLMS/engines/comprehension/app/models/comprehension/highlight.rb
```diff
@@ -21,5 +23,29 @@ def serializable_hash(options = nil)
...
+
+    private def first_order
+      feedback.order == 0
+    end
+
+    private def second_order
+      feedback.order == 1
+    end
+
+    def log_update(user_id, prev_value)
```
anathomical, line 39: Again, possible risks around universal rules and duplication.

services/QuillLMS/engines/comprehension/app/models/comprehension/label.rb
```diff
@@ -26,5 +28,17 @@ def serializable_hash(options = nil)
         errors.add(:name, "can't be the same as any other labels related to the same prompt")
       end
     end
+
+    def log_creation(user_id)
```
anathomical, line 44: Again, possible risks around duplication.  (Again, maybe this is something we actually want, but worth calling out.)

services/QuillLMS/engines/comprehension/app/models/comprehension/prompt.rb
```diff
@@ -11,6 +13,7 @@ class Prompt < ActiveRecord::Base
     has_many :automl_models, inverse_of: :prompt
     has_many :prompts_rules
     has_many :rules, through: :prompts_rules, inverse_of: :prompts
+    has_many :change_logs
```
anathomical, line 16: Does this actually work?  `ChangeLog` is a model that isn't inside the Comprehension Engine, which leads me to suspect that this automatic relationship short-cut in Rails wouldn't work.

services/QuillLMS/engines/comprehension/lib/comprehension/comprehension/change_log.rb
```diff
@@ -0,0 +1,17 @@
...
+        user_id: user_id,
+        action: Comprehension.change_log_class::COMPREHENSION_ACTIONS[action],
+        changed_record_type: changed_record.class.name,
+        changed_record_id: changed_record.id,
+        explanation: explanation,
+        changed_attribute: changed_attribute,
+        previous_value: previous_value,
+        new_value: new_value
+      }
+      Comprehension.change_log_class.create(change_log)
```
anathomical, line 14: Not the end of the world, but there's no error handling on this.  If something does go wrong, or something is invalid, this makes it look like the error will get swallowed.  Maybe that's fine for now since we don't really know what we want to do to handle errors, but something worth thinking about, maybe.

services/QuillLMS/engines/comprehension/app/models/comprehension/rule.rb
```diff
@@ -34,6 +37,7 @@ class Rule < ActiveRecord::Base
     has_many :prompts_rules, inverse_of: :rule
     has_many :prompts, through: :prompts_rules, inverse_of: :rules
     has_many :regex_rules, inverse_of: :rule, dependent: :destroy
+    has_many :change_logs
```
anathomical, line 40: Again, does this work?  Additionally, is it necessary?

services/QuillLMS/engines/comprehension/app/models/comprehension/rule.rb
```diff
@@ -98,5 +114,64 @@ def display_name
         errors.add(:prompts, "prompt #{prompt.id} already has a plagiarism rule") if prompt.rules.where(rule_type: TYPE_PLAGIARISM).first&.id
       end
     end
+
+    def log_creation(user_id)
+      if regex?
```
anathomical, line 119: Not sure if this is actually be better, but this could be a set of early returns:
```ruby
return send_change_log(...) if ...
return send_change_log(...) if ...
```
And so on.

services/QuillLMS/engines/comprehension/app/models/comprehension/rule.rb
```diff
@@ -98,5 +114,64 @@ def display_name
...
+      elsif plagiarism?
+        send_change_log(user_id, :create_plagiarism, change_text_plagiarism)
+      elsif universal?
+        send_change_log(user_id, :create_universal, change_text_universal)
+      end
+    end
+
+    def log_deletion(user_id)
+      if regex?
+        send_change_log(user_id, :delete_regex, nil, change_text_regex)
```
anathomical, line 130: Since this is a one liner, I'd do it with the `if` at the end of the `send_change_log` line.

services/QuillLMS/engines/comprehension/app/models/comprehension/rule.rb
```diff
@@ -98,5 +114,64 @@ def display_name
...
+        prev_values = change_text_plagiarism(old_values_arr)
+        send_change_log(user_id, :update_plagiarism, change_text_plagiarism(new_values(new_values_arr || old_values_arr)), prev_values)
+      elsif universal?
+        prev_values = change_text_universal(old_values_arr)
+        send_change_log(user_id, :update_universal, change_text_universal(new_values(new_values_arr || old_values_arr)), prev_values)
+      end
+    end
+
+    private def send_change_log(user_id, action_type, new_value, prev_value=nil)
+      prompts&.each do |prompt|
```
anathomical: Again, I'm a bit worried about the potential duplication on Universal rules if that's not something we actively want.

anathomical (changes requested):
A number of minor concerns about specific implementation, but the overall approach seems solid to me.  Would like to see a couple of items cleaned up, but this is probably close to ready to go.

services/QuillLMS/app/models/change_log.rb
happythenewsad: I agree. Have you considered using an existing gem to abstract away some of this feature's complexity? Example: https://github.com/collectiveidea/audited

PR #7785, opened 2021-06-10
Lms/persist fh highlight

happythenewsad:
## WHAT
FeedbackHistory highlights were not being persisted due to a bug in the params.permit controller logic. Also, the client side was expecting the shape of 'highlight' to be a string, which does not match the [spec payload](https://www.notion.so/quill/Comprehension-API-v1-spec-2b3c21accf2d401699118af3399f0deb). 

## HOW
I added a defensively-coded highlight extractor on the client side that is compatible with both the old, malformed FeedbackHistory.metadata and the new, correctly formed FeedbackHistory.metadata.

### Screenshots

### Notion Card Links
https://www.notion.so/quill/Product-Board-30f97e4eb01246dbb8264253fb385073?p=392272bd51c44f5db89b149b03a99437

PR Checklist | Your Answer
------------ | -------------
Design Review: If applicable, have you compared the coded design to the mockups? | (N/A )

### DIFF
--- a/services/QuillLMS/client/app/bundles/Staff/components/comprehension/rulesAnalysis/ruleAnalysis.tsx
+++ b/services/QuillLMS/client/app/bundles/Staff/components/comprehension/rulesAnalysis/ruleAnalysis.tsx
@@ -184,11 +184,16 @@ const RuleAnalysis = ({ history, match }) => {
     { name: "", attribute:"value", width: "750px" }
   ];
 
+  const extractHighlight = (highlightObject) => {
+    if (!highlightObject || !highlightObject.length || !highlightObject[0].text) return '';
+    return highlightObject[0].text;
+  }
+
   const responseRows = () => {
     if (!activityData || !responses) { return [] }
     return responses.filter(filterResponsesByScored).filter(filterResponsesBySearch).map(r => {
-      const formattedResponse = {...r}
-      const highlightedEntry = r.entry.replace(r.highlight, `<strong>${r.highlight}</strong>`)
+      const formattedResponse = {...r,  ...{highlight: extractHighlight(r.highlight)}}
+      const highlightedEntry = r.entry.replace(formattedResponse.highlight, `<strong>${formattedResponse.highlight}</strong>`)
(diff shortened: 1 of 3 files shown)


services/QuillLMS/client/app/bundles/Staff/components/comprehension/rulesAnalysis/ruleAnalysis.tsx
```diff
@@ -184,11 +184,16 @@ const RuleAnalysis = ({ history, match }) => {
     { name: "", attribute:"value", width: "750px" }
   ];
 
+  const extractHighlight = (highlightObject) => {
+    if (!highlightObject || !highlightObject.length || highlightObject.length < 1 || !highlightObject[0].text) return '';
```
emilia-friedberg: 0 is falsy in Javascript, so you don't need the `highlightObject.length < 1` clause here.

emilia-friedberg (approved):
This looks like a good solution. I want to point out that the frontend was expecting the highlight in the same shape as the spec documented here: https://www.notion.so/quill/Rules-Report-V1-f07e030088ad43a79849c1b48e97a857#bca06b30c2c34c9ca63f234fb08eb44f.

services/QuillLMS/client/app/bundles/Staff/components/comprehension/rulesAnalysis/ruleAnalysis.tsx
```diff
@@ -184,11 +184,16 @@ const RuleAnalysis = ({ history, match }) => {
...
+    return highlightObject[0].text;
+  }
```
emilia-friedberg: Since this function doesn't use any props or state, we can define it outside of the component function, which will make it easier to pull into a shared file later if we want to use it elsewhere.

happythenewsad:
> This looks like a good solution. I want to point out that the frontend was expecting the highlight in the same shape as the spec documented here: https://www.notion.so/quill/Rules-Report-V1-f07e030088ad43a79849c1b48e97a857#bca06b30c2c34c9ca63f234fb08eb44f.

Good point - I just updated the example json payload to reflect the new reality.

services/QuillLMS/client/app/bundles/Staff/components/comprehension/rulesAnalysis/ruleAnalysis.tsx
happythenewsad: addressed

PR #7786, opened 2021-06-10
Comprehension internal tools/filter sessions by time

eadams17:
## WHAT
add frontend time filtering to comprehension internal tools for feedback sessions, rules analysis and rule analysis

## WHY
Curriculum would like to filter through this data with time constraints

## HOW
update various APIs to accept time params, add time widgets to select date range saved to storage session to persist beyond page refreshes; update table sorting logic to handle time constraints-- this also fixes a query refetch in `RuleAnalysis` to remove the need for keeping track of response strength updates in local state

### Screenshots
Screen Shot 2021-06-10 at 4 14 15 PM
Screen Shot 2021-06-10 at 4 14 06 PM
Screen Shot 2021-06-10 at 4 15 00 PM

### Notion Card Links
https://www.notion.so/quill/Filter-View-All-Sessions-by-time-frame-c36f841cec5845639d42c3a6af836e28

PR Checklist | Your Answer
------------ | -------------


services/QuillLMS/client/package.json
```diff
@@ -76,6 +76,7 @@
     "react-css-transition-replace": "^4.0.2",
     "react-csv": "^1.0.8",
     "react-dates": "21.8.0",
+    "react-datetime-picker": "^3.3.0",
```
emilia-friedberg, line 79: I feel like we shouldn't be adding another datepicker library when we already have `react-dates` in here. Is there a reason `react-dates` doesn't work for what we need?

services/QuillLMS/client/app/bundles/Staff/styles/rule_analysis.scss
```diff
@@ -12,6 +12,34 @@
...
+    .end-date-picker {
+      margin-right: 16px;
+    }
+    .react-datetime-picker__inputGroup {
+      padding-top: 2px;
+    }
+    .react-datetime-picker__inputGroup__input {
+      height: auto;
+    }
+  }
```
emilia-friedberg: I think there is some repeated code between this and the other places you're updating the datepicker CSS - it would be nice to put that in a shared file.

services/QuillLMS/client/app/bundles/Staff/components/comprehension/rulesAnalysis/ruleAnalysis.tsx
```diff
@@ -184,7 +215,7 @@ const RuleAnalysis = ({ history, match }) => {
     { name: "", attribute:"value", width: "750px" }
   ];
 
-  const responseRows = () => {
+  const responseRows = (responses) => {
```
emilia-friedberg: Why do we need to pass responses when it's in the local state?

services/QuillLMS/client/app/bundles/Staff/components/comprehension/rulesAnalysis/ruleAnalysis.tsx
```diff
@@ -90,35 +115,41 @@ const RuleAnalysis = ({ history, match }) => {
...
-  function updateResponseStrengthInState(responseIds, rating) {
-    const newResponses = [...responses]
-    responseIds.forEach(id => {
-      const indexOfResponseToChange = responses.findIndex(r => r.response_id === id)
-      const responseToChange = responses[indexOfResponseToChange]
-      responseToChange.strength = rating
-      newResponses[indexOfResponseToChange] = responseToChange
-    })
-    setResponses(newResponses)
-  }
```
emilia-friedberg, line 114: So the reason we were preserving this in the state wasn't because the refetching didn't work, but because that was too slow - it was specifically requested by the curriculum team that we show the value as soon as it is clicked. I think if we remove this code, it's going to get slow again, but if that doesn't bear out in testing then that's fine with me.

emilia-friedberg (changes requested):
Your general approach here looks good, but there are a couple of things I think we could clean up/double check in the implementation. My biggest concern is about the additional library - I think every time we add a new library we are introducing potential security and maintenance issues, so we should be using what we already have where possible.

services/QuillLMS/client/package.json
eadams17: @emilia-friedberg yeah so in the spec, it's specified that they want to be able to select both the date and time. `react-dates` is just a calendar widget and only let's you specify the date

services/QuillLMS/client/app/bundles/Staff/styles/rule_analysis.scss
eadams17: good point, I'll combine them into a shared file

services/QuillLMS/client/app/bundles/Staff/components/comprehension/rulesAnalysis/ruleAnalysis.tsx
eadams17: because in the hooks at the top, we're passing `ruleFeedbackHistoryData.responses` when there is data and manually setting `setResponses([]);` to initially render an empty table when a time hasn't been selected

services/QuillLMS/client/app/bundles/Staff/components/comprehension/rulesAnalysis/ruleAnalysis.tsx
eadams17: so I actually learned recently that for the query refetch to work, you have to pass the full array of args, ie:

```
queryCache.refetchQueries([`rule-feedback-histories-by-rule-${ruleId}-${promptId}`, ruleId, promptId, startDateForQuery, endDateForQuery]);
```

so I don't think this was fetching anything, I think the reason it was slow is because that initial useQuery call checks for new data every like 15 seconds or so. Testing this on staging, the update happens in less than a second

anathomical (approved):
Looks okay to me, though I share Emilia's concern about adding a new library.  How much work would it take to swap out the old library for the new one everywhere it's used?

services/QuillLMS/client/app/bundles/Staff/components/comprehension/rulesAnalysis/ruleAnalysis.tsx
emilia-friedberg: Gotcha. In that case I think we should change the name of this parameter to avoid conflict.

emilia-friedberg (approved):
Approving with a name change for that one parameter!

PR #7790, opened 2021-06-11
Lms/comprehension prompt health report performance

anathomical:
## WHAT
Refactor the Prompt Health query to be pure SQL rather than doing Ruby-based post-processing on a simpler query
## WHY
The loops in the post-processing were causing performance issues, so we're moving all the logic in to SQL in order to improve performance
## HOW
Write a relatively complex aggregate SQL query to do all of the calculations that we care about in SQL

### Notion Card Links
https://www.notion.so/quill/Comprehension-Prompt-Health-Endpoint-Performance-cefc25b427494bc18a87084909acc794

PR Checklist | Your Answer
------------ | -------------


services/QuillLMS/app/queries/prompt_feedback_history.rb
```diff
@@ -1,128 +1,43 @@
 class PromptFeedbackHistory
 
     def self.run(activity_id)
-        promptwise_postprocessing promptwise_sessions(activity_id)
+        #promptwise_postprocessing promptwise_sessions(activity_id)
```
happythenewsad: Missed a line here

happythenewsad (changes requested):
This looks promising - nice work. Just a trivial change request. Also, it would be a good idea to run EXPLAIN ANALYZE on this PR and compare with `develop`.

services/QuillLMS/app/queries/prompt_feedback_history.rb
anathomical: Good catch.  Fixed.

anathomical:
@happythenewsad We could run `EXPLAIN ANALYZE`, but in this case I'm not sure what it would demonstrate.  NewRelic tracing suggests that the time-spent in the current endpoint is mostly in Ruby, not in SQL, so we wouldn't actually expect this query to be more performant than the one we've been using on its face.  We just expect the whole call to be more performant.

happythenewsad:
> We just expect the whole call to be more performant.

Yes, we expect the whole call to be more performant. I inquired because I want to avoid the SQL call getting too expensive. That said, I'm fine with trying it out in prod.

happythenewsad (approved):

PR #7797, opened 2021-06-15
add timetracking to all tools

emilia-friedberg:
## WHAT
Add timetracking to Connect, Diagnostic, Grammar, and Lessons. Also update "timespent" function to use the values of the tracked actions instead of just a `completed_at` - `started_at` computation.

## WHY
So we can start accessing this data for both internal and external use.

## HOW
For most of the apps, I followed the pattern I started with Comprehension, where the `timeTracking` data is stored in the local state, saved to the `active_activity_session` on each submission, and then finally saved to the `activity_session` when the activity is completed. I did slightly amend the way that we are saving data, in that I realized that saving a rounded second count after each action was resulting in lost time when too little time occurred between actions, so I'm tracking time in milliseconds for active sessions and then saving in seconds at the end. Deviations from this pattern and other changes include:

- Proofreader, where I'm saving the time in Redux rather than in the local state, because the Save and Exit button actually has to do something
- I updated how Grammar handles Proofreader sessions slightly, because the Grammar follow-up was never getting resumed correctly and that was making the timetracking weird (plus presumably annoying for students)
- I removed the `nextQuestionWithoutSaving` function from Diagnostic, because it was resulting in the title cards not being incremented correctly, and instead follow the pattern we have in Connect where title cards are simply filtered out before the concept results are calculated
- There is no individual activity session save in Lessons, so I followed the pattern for the other mass-change functions that occur when a Lesson is completed and just copied the time tracking data over directly from `ActiveActivitySession`.

### Screenshots

### Notion Card Links
https://www.notion.so/quill/Time-Spent-All-Tools-Time-Tracking-Connect-Proofreader-Grammar-Diagnostic-Lessons-184ccae898464f4a8cf62800656b17f0

PR Checklist | Your Answer
------------ | -------------


services/QuillLMS/client/app/bundles/Connect/components/studentLessons/lesson.jsx
```diff
@@ -110,6 +125,62 @@ export class Lesson extends React.Component {
...
   }
 
+  componentWillUnmount() {
+    window.removeEventListener('keydown', this.resetTimers)
+    window.removeEventListener('mousemove', this.resetTimers)
+    window.removeEventListener('mousedown', this.resetTimers)
+    window.removeEventListener('click', this.resetTimers)
+    window.removeEventListener('keypress', this.resetTimers)
+    window.removeEventListener('scroll', this.resetTimers)
+    window.removeEventListener('visibilitychange', this.setIdle)
```
eadams17: I think we want to constantize these event listener keys in a shared file since we use them in several places

services/QuillLMS/client/app/bundles/Lessons/components/classroomLessons/play/container.tsx
```diff
@@ -74,33 +79,60 @@ class PlayClassroomLessonContainer extends React.Component<any, any> {
...
-    const { classroomLesson } = this.props
+    const { timeTracking, } = this.state
+    const { classroomLesson, classroomSessions, } = this.props
     const { hasreceiveddata } = classroomLesson
     if (classroomLesson.hasreceiveddata != prevProps.hasreceiveddata && hasreceiveddata) {
       document.title = `Quill.org | ${classroomLesson.data.title}`
     }
+
+    if (classroomSessions.data.current_slide !== prevProps.classroomSessions.data.current_slide) {
+      const student = getParameterByName('student');
```
eadams17, line 130: I think we could also constantize "student" as well

eadams17 (approved):
Looks great! just some comments around where we could constantize some variables. My only concern is with the removal of `nextQuestionWithoutSaving` in Diagnostic. You probably already checked how it affects the teacher preview mode, but I just wanted to flag that in case you missed it.

services/QuillLMS/app/models/activity_session.rb
```diff
@@ -376,6 +376,20 @@ def self.delete_activity_sessions_with_no_concept_results(classroom_unit_id, act
     ActivitySession.where(id: incomplete_activity_session_ids).destroy_all
   end
 
+  def self.save_timetracking_data_from_active_activity_session(classroom_unit_id, activity_id)
```
anathomical, line 381: As I read it, this function is intended for Lessons, but not other activities.  If that's true, we should probably try to find some way to avoid this being accidentally called on another activity type which I could imagine squashing data for finished sessions that get caught up in the `where` clause below.  I don't think we need to do that protection via code, but a comment here, or more explicit function name would probably solve the problem without doing more.

anathomical (approved):
One comment, but it's about style/safety.  Everything here looks good to me.

PR #7798, opened 2021-06-15
Featureflags/scaffold

happythenewsad:
## WHAT
Core of Feature Flags project: data models, migrations, logic for checking if an AppSetting is enabled for a user, and controller endpoint for accessing a user's app settings clientside (implemented separately)

## WHY
Necessary for completion of project.

### Screenshots

### Notion Card Links
https://www.notion.so/quill/Product-Board-30f97e4eb01246dbb8264253fb385073?p=bc277b68f0b744448ec61a59df327988

PR Checklist | Your Answer
------------ | -------------
Have you added and/or updated tests? |  (The answer should mostly be 'YES'. If you answer 'NO', please justify.) yes
Self-Review: Have you done an initial self-review of the code below on Github? | no

### DIFF
--- a/services/QuillLMS/app/controllers/api/v1/app_settings_controller.rb
+++ b/services/QuillLMS/app/controllers/api/v1/app_settings_controller.rb
@@ -1,0 +1,22 @@
+class API::V1::AppSettingsController < ApplicationController
+  def index
+    user = User.find(app_setting_index_params)
+    render(json: AppSetting.all_enabled_for_user(user))
+  end
+
+  def show
+    user = User.find(app_setting_show_params[:user_id])
+    name =  app_setting_show_params[:name] 
+    render(json: {
+      name => AppSetting.enabled?(name: name, user: user)
+    })
+  end
+
+  private def app_setting_index_params
+    params.require(:user_id)
+  end
+
+  private def app_setting_show_params
+    params.permit(:user_id, :name)
+  end
+end
--- a/services/QuillLMS/app/controllers/api/v1/app_settings_controller.rb
+++ b/services/QuillLMS/app/controllers/api/v1/app_settings_controller.rb
@@ -1,4 +1,4 @@
-class API::V1::AppSettingsController < ApplicationController
+class Api::V1::AppSettingsController < ApplicationController
   def index
     user = User.find(app_setting_index_params)
     render(json: AppSetting.all_enabled_for_user(user))
--- a/services/QuillLMS/app/models/app_setting.rb
+++ b/services/QuillLMS/app/models/app_setting.rb
@@ -7,7 +7,7 @@
 #  enabled_for_admins  :boolean          default(FALSE), not null
 #  name                :string           not null
 #  percent_active      :integer          default(0), not null
-#  user_ids_allow_list :string           default([]), not null, is an Array
+#  user_ids_allow_list :integer          default([]), not null, is an Array
 #  created_at          :datetime         not null
 #  updated_at          :datetime         not null
 #
@@ -49,7 +49,7 @@ class AppSetting < ActiveRecord::Base
 
     return true if user_in_rollout_bucket?(user.id)
     
-    return true if user_ids_allow_list.include?(user.id.to_s)
+    return true if user_ids_allow_list.include?(user.id)
 
     false
   end
--- a/services/QuillLMS/app/models/app_setting.rb
+++ b/services/QuillLMS/app/models/app_setting.rb
@@ -1,0 +1,61 @@
+# == Schema Information
+#
+# Table name: app_settings
+#
+#  id                  :integer          not null, primary key
+#  enabled             :boolean          default(FALSE), not null
+#  enabled_for_admins  :boolean          default(FALSE), not null
+#  name                :string           not null
+#  percent_active      :integer          default(0), not null
+#  user_ids_allow_list :string           default([]), not null, is an Array
+#  created_at          :datetime         not null
+#  updated_at          :datetime         not null
+#
+# Indexes
+#
+#  index_app_settings_on_name  (name) UNIQUE
+#
+require 'zlib'
+
+class AppSetting < ActiveRecord::Base
+  validates :percent_active, numericality: { 
+    only_integer: true, 
+    greater_than_or_equal_to: 0, 
+    less_than_or_equal_to: 100
+  }
+
+  validates :name, uniqueness: true
+
+  scope :enabled, -> { where(enabled: true) }
+
+  def self.enabled?(name:, user:)
+    app_setting = AppSetting.find_by_name(name)
+    return false unless app_setting
+    app_setting.enabled_for_user?(user)
+  end
+
+  def self.all_enabled_for_user(user)
+    app_settings = AppSetting.enabled
+    app_settings.each_with_object({}) do |app_setting, memo| 
+      memo[app_setting.name] = app_setting.enabled_for_user?(user) 
+      memo
+    end
+  end
+
+  def enabled_for_user?(user)
+    return false unless enabled
+
+    return true if enabled_for_admins && user.admin?
+
+    return true if user_in_rollout_bucket?(user.id)
+    
+    return true if user_ids_allow_list.include?(user.id.to_s)
+
+    false
+  end
+
+  def user_in_rollout_bucket?(user_id) 
+    Zlib.crc32(name + user_id.to_s).digits.take(2).join.to_i < percent_active
+  end
+
+end
--- a/services/QuillLMS/config/routes.rb
+++ b/services/QuillLMS/config/routes.rb
@@ -369,6 +369,7 @@ EmpiricalGrammar::Application.routes.draw do
       get 'rule_feedback_history/:rule_uid' => 'rule_feedback_histories#rule_detail'
       get 'prompt_health/:activity_id' => 'rule_feedback_histories#prompt_health'
 
+      
       resources :activities,              except: [:index, :new, :edit]
       resources :activity_flags,          only: [:index]
       resources :activity_sessions,       except: [:index, :new, :edit]
@@ -385,7 +386,12 @@ EmpiricalGrammar::Application.routes.draw do
           get 'level_zero_concepts_with_lineage'
         end
       end
-      resources :users,                   only: [:index]
+      
+      resources :users, only: [:index] do
+        resources :app_settings, only: [:index] 
+        get 'app_setting(/:name)' => 'app_settings#show'
+      end
+
       resources :classroom_units,         only: [] do
         collection do
           get 'student_names'
--- a/services/QuillLMS/config/routes.rb
+++ b/services/QuillLMS/config/routes.rb
@@ -388,8 +388,7 @@ EmpiricalGrammar::Application.routes.draw do
       end
       
       resources :users, only: [:index] do
-        resources :app_settings, only: [:index] 
-        get 'app_setting(/:name)' => 'app_settings#show'
+        resources :app_settings, only: [:index, :show], param: :name 
       end
 
       resources :classroom_units,         only: [] do
--- a/services/QuillLMS/db/migrate/20210614205654_create_app_settings.rb
+++ b/services/QuillLMS/db/migrate/20210614205654_create_app_settings.rb
@@ -2,7 +2,7 @@ class CreateAppSettings < ActiveRecord::Migration
   def change
     create_table :app_settings do |t|
       t.string :name, null: false
-      t.string :user_ids_allow_list, null: false, default: [], array: true
+      t.integer :user_ids_allow_list, null: false, default: [], array: true
       t.boolean :enabled_for_admins, null: false, default: false
       t.boolean :enabled, null: false, default: false
       t.integer :percent_active, null: false, default: 0
--- a/services/QuillLMS/db/migrate/20210614205654_create_app_settings.rb
+++ b/services/QuillLMS/db/migrate/20210614205654_create_app_settings.rb
@@ -1,0 +1,14 @@
+class CreateAppSettings < ActiveRecord::Migration
+  def change
+    create_table :app_settings do |t|
+      t.string :name, null: false
+      t.string :user_ids_allow_list, null: false, default: [], array: true
+      t.boolean :enabled_for_admins, null: false, default: false
+      t.boolean :enabled, null: false, default: false
+      t.integer :percent_active, null: false, default: 0
+
+      t.timestamps null: false
+    end
+    add_index :app_settings, :name, unique: true
+  end
+end
--- a/services/QuillLMS/db/structure.sql
+++ b/services/QuillLMS/db/structure.sql
@@ -77,6 +77,18 @@ CREATE FUNCTION public.blog_posts_search_trigger() RETURNS trigger
 
 
 --
+-- Name: my_jsonb_to_hstore(jsonb); Type: FUNCTION; Schema: public; Owner: -
+--
+
+CREATE FUNCTION public.my_jsonb_to_hstore(jsonb) RETURNS public.hstore
+    LANGUAGE sql IMMUTABLE STRICT
+    AS $_$
+            SELECT hstore(array_agg(key), array_agg(value))
+            FROM   jsonb_each_text($1)
+          $_$;
+
+
+--
 -- Name: old_timespent_teacher(integer); Type: FUNCTION; Schema: public; Owner: -
 --
 
@@ -702,6 +714,54 @@ CREATE SEQUENCE public.announcements_id_seq
 --
 
 ALTER SEQUENCE public.announcements_id_seq OWNED BY public.announcements.id;
+
+
+--
+-- Name: app_settings; Type: TABLE; Schema: public; Owner: -
+--
+
+CREATE TABLE public.app_settings (
+    id integer NOT NULL,
+    name character varying NOT NULL,
+    user_ids_allow_list character varying[] DEFAULT '{}'::character varying[] NOT NULL,
+    enabled_for_admins boolean DEFAULT false NOT NULL,
+    enabled boolean DEFAULT false NOT NULL,
+    percent_active integer DEFAULT 0 NOT NULL,
+    created_at timestamp without time zone NOT NULL,
+    updated_at timestamp without time zone NOT NULL
+);
+
+
+--
+-- Name: app_settings_id_seq; Type: SEQUENCE; Schema: public; Owner: -
+--
+
+CREATE SEQUENCE public.app_settings_id_seq
+    AS integer
+    START WITH 1
+    INCREMENT BY 1
+    NO MINVALUE
+    NO MAXVALUE
+    CACHE 1;
+
+
+--
+-- Name: app_settings_id_seq; Type: SEQUENCE OWNED BY; Schema: public; Owner: -
+--
+
+ALTER SEQUENCE public.app_settings_id_seq OWNED BY public.app_settings.id;
+
+
+--
+-- Name: ar_internal_metadata; Type: TABLE; Schema: public; Owner: -
+--
+
+CREATE TABLE public.ar_internal_metadata (
+    key character varying NOT NULL,
+    value character varying,
+    created_at timestamp without time zone NOT NULL,
+    updated_at timestamp without time zone NOT NULL
+);
 
 
 --
@@ -3819,6 +3879,13 @@ ALTER TABLE ONLY public.announcements ALTER COLUMN id SET DEFAULT nextval('publi
 
 
 --
+-- Name: app_settings id; Type: DEFAULT; Schema: public; Owner: -
+--
+
+ALTER TABLE ONLY public.app_settings ALTER COLUMN id SET DEFAULT nextval('public.app_settings_id_seq'::regclass);
+
+
+--
 -- Name: auth_credentials id; Type: DEFAULT; Schema: public; Owner: -
 --
 
@@ -4508,6 +4575,22 @@ ALTER TABLE ONLY public.admin_accounts_teachers
 
 ALTER TABLE ONLY public.announcements
     ADD CONSTRAINT announcements_pkey PRIMARY KEY (id);
+
+
+--
+-- Name: app_settings app_settings_pkey; Type: CONSTRAINT; Schema: public; Owner: -
+--
+
+ALTER TABLE ONLY public.app_settings
+    ADD CONSTRAINT app_settings_pkey PRIMARY KEY (id);
+
+
+--
+-- Name: ar_internal_metadata ar_internal_metadata_pkey; Type: CONSTRAINT; Schema: public; Owner: -
+--
+
+ALTER TABLE ONLY public.ar_internal_metadata
+    ADD CONSTRAINT ar_internal_metadata_pkey PRIMARY KEY (key);
 
 
 --
@@ -5383,6 +5466,13 @@ CREATE INDEX index_admin_accounts_teachers_on_teacher_id ON public.admin_account
 --
 
 CREATE INDEX index_announcements_on_start_and_end ON public.announcements USING btree (start, "end" DESC);
+
+
+--
+-- Name: index_app_settings_on_name; Type: INDEX; Schema: public; Owner: -
+--
+
+CREATE UNIQUE INDEX index_app_settings_on_name ON public.app_settings USING btree (name);
 
 
 --
@@ -7637,4 +7727,6 @@ INSERT INTO schema_migrations (version) VALUES ('20210528142650');
 
 INSERT INTO schema_migrations (version) VALUES ('20210614190031');
 INSERT INTO schema_migrations (version) VALUES ('20210525200201');
+
+INSERT INTO schema_migrations (version) VALUES ('20210614205654');
 
--- a/services/QuillLMS/db/structure.sql
+++ b/services/QuillLMS/db/structure.sql
@@ -723,7 +723,7 @@ ALTER SEQUENCE public.announcements_id_seq OWNED BY public.announcements.id;
 CREATE TABLE public.app_settings (
     id integer NOT NULL,
     name character varying NOT NULL,
-    user_ids_allow_list character varying[] DEFAULT '{}'::character varying[] NOT NULL,
+    user_ids_allow_list integer[] DEFAULT '{}'::integer[] NOT NULL,
     enabled_for_admins boolean DEFAULT false NOT NULL,
     enabled boolean DEFAULT false NOT NULL,
     percent_active integer DEFAULT 0 NOT NULL,
@@ -750,6 +750,19 @@ CREATE SEQUENCE public.app_settings_id_seq
 --
 
 ALTER SEQUENCE public.app_settings_id_seq OWNED BY public.app_settings.id;
+
+
+--
+-- Name: ar_internal_metadata; Type: TABLE; Schema: public; Owner: -
+--
+
+CREATE TABLE public.ar_internal_metadata (
+    key character varying NOT NULL,
+    value character varying,
+    created_at timestamp without time zone NOT NULL,
+    updated_at timestamp without time zone NOT NULL
+);
+
 
 --
 -- Name: auth_credentials; Type: TABLE; Schema: public; Owner: -
@@ -1243,7 +1256,8 @@ CREATE TABLE public.comprehension_automl_models (
     prompt_id integer,
     state character varying NOT NULL,
     created_at timestamp without time zone NOT NULL,
-    updated_at timestamp without time zone NOT NULL
+    updated_at timestamp without time zone NOT NULL,
+    notes text DEFAULT ''::text
 );
 
 
@@ -1555,10 +1569,9 @@ CREATE TABLE public.comprehension_rules (
     rule_type character varying NOT NULL,
     optimal boolean NOT NULL,
     suborder integer,
-    concept_uid character varying NOT NULL,
+    concept_uid character varying,
     created_at timestamp without time zone NOT NULL,
     updated_at timestamp without time zone NOT NULL,
-    sequence_type character varying,
     state character varying NOT NULL
 );
 
@@ -4570,6 +4583,15 @@ ALTER TABLE ONLY public.announcements
 ALTER TABLE ONLY public.app_settings
     ADD CONSTRAINT app_settings_pkey PRIMARY KEY (id);
 
+
+--
+-- Name: ar_internal_metadata ar_internal_metadata_pkey; Type: CONSTRAINT; Schema: public; Owner: -
+--
+
+ALTER TABLE ONLY public.ar_internal_metadata
+    ADD CONSTRAINT ar_internal_metadata_pkey PRIMARY KEY (key);
+
+
 --
 -- Name: auth_credentials auth_credentials_pkey; Type: CONSTRAINT; Schema: public; Owner: -
 --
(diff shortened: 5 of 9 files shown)


services/QuillLMS/app/models/activity_session.rb
```diff
@@ -4,7 +4,7 @@
 #
 #  id                    :integer          not null, primary key
 #  completed_at          :datetime
-#  data                  :hstore
+#  data                  :jsonb
```
happythenewsad: This should have been checked in in an unrelated commit.

services/QuillLMS/app/models/app_setting.rb
```diff
@@ -0,0 +1,59 @@
...
+    validates :name, uniqueness: true
+
+    def self.enabled?(name:, user:)
+        as = AppSetting.find_by_name name
+        return false unless as 
+        return false unless user.respond_to? :id 
+        as._enabled? user.id
+    end
+
+    def _enabled? user
```
happythenewsad: Normal development should only use `enabled?`, because it's safe to invoke with bad input (non existent user or AppSetting records). The reason why I'm not making it a private method is because the controller needs to invoke `_enabled?` for performance reasons. Alternatively, we could pass an AppSetting object into `enabled?`, then the controller would not need to use the internal `_enabled?`

services/QuillLMS/lib/tasks/app_settings.rake
```diff
@@ -0,0 +1,15 @@
+namespace :app_settings do
+  desc 'Creates a new AppSetting.'
+  # Example usage: rake 'app_settings:create[theName,true,true,50,[]]'
```
happythenewsad, line 3: There's a gotcha with zsh that requires rake tasks to be quoted like above when using params

services/QuillLMS/app/controllers/api/v1/app_settings_controller.rb
```diff
@@ -0,0 +1,17 @@
+class API::V1::AppSettingsController < ApplicationController
+  def index
```
happythenewsad, line 2: On the fence regarding authentication in this controller - share your thoughts.

services/QuillLMS/app/controllers/api/v1/app_settings_controller.rb
```diff
@@ -0,0 +1,17 @@
...
+    app_settings = AppSetting.all
```
happythenewsad: I opted not to go with a UserAppSetting join table, since it's unnecessary overhead if user whitelists are reasonably short.

happythenewsad:
FYI I'm designating @anathomical  and @brendanshean as reviewers, and @dandrabik as an auditor (I won't hold off merging due to his review). 

Thanks folks!

services/QuillLMS/app/models/app_setting.rb
```diff
@@ -0,0 +1,65 @@
...
+#  index_app_settings_on_name  (name)
+#
+class AppSetting < ActiveRecord::Base
+    PCT_ACTIVE_VALUES = [100, 90, 80, 70, 60, 50, 40, 30, 20, 10, 0] 
+
+    validates :pct_active, inclusion: { in: PCT_ACTIVE_VALUES }
+    validates :name, uniqueness: true
+
+    def self.enabled?(name:, user:)
+        as = AppSetting.find_by_name name
```
brendanshean: I think it's a good practice to avoid abbreviations for variable names.  I would use `app_setting`, especially in this case since `as` looks like a keyword.

services/QuillLMS/app/models/app_setting.rb
```diff
@@ -0,0 +1,65 @@
...
+        return false unless as 
+        return false unless user.respond_to? :id 
+        as._enabled? user.id
+    end
+
+    def _enabled? user
+        return false unless enabled
+
+        if enabled_for_admins && user.admin?
```
brendanshean: Rewriting all of these conditionals as guard clauses would improve readability.

services/QuillLMS/app/models/app_setting.rb
brendanshean: What about renaming `_enabled?` to `user_enabled?`.  This would be more transparent and would also eliminate any connection being made between `enabled?` and `_enabled?` (which have different contracts).

services/QuillLMS/db/migrate/20210614205654_create_app_settings.rb
```diff
@@ -0,0 +1,14 @@
...
+    create_table :app_settings do |t|
+      t.string :name, null: false
+      t.string :user_whitelist, null: false, default: [], array: true
+      t.boolean :enabled_for_admins, null: false, default: false
+      t.boolean :enabled, null: false, default: false
+      t.integer :pct_active, null: false, default: 0
+
+      t.timestamps null: false
+    end
+    add_index :app_settings, :name
```
brendanshean: Should there be a `unique: true` on for the `name` index?

services/QuillLMS/spec/controllers/api/v1/app_settings_controller_spec.rb
```diff
@@ -0,0 +1,24 @@
...
+  let(:user) { create(:user) }
+
+  describe "GET #index" do
+    it "returns a success response" do
+      
+      3.times { create(:app_setting) }
+      
+      get(:index, user_id: user.id) 
+
+      expected = {
```
brendanshean: Is this variable needed?

services/QuillLMS/app/models/app_setting.rb
```diff
@@ -0,0 +1,65 @@
...
+        enabled_buckets.include? user_id.digits.first
+    end
+
+    # The range of possible buckets is 0-9 inclusive, so
+    # we transform the user-facing pct_active value by taking
+    # the first radix integer and subtracting 1
+    private def enabled_buckets
+        @enabled_buckets ||=    if pct_active == 0 
+                                    [] 
+                                elsif pct_active == 100
```
brendanshean: If order doesn't matter, you could use a ranges:
`(0..9).to_a` and  `(0..(pct_active.digits.last - 1)).to_a`

services/QuillLMS/app/controllers/api/v1/app_settings_controller.rb
brendanshean: I feel like authentication is something that needs to be addressed, but probably out of scope for this feature.

services/QuillLMS/app/models/app_setting.rb
```diff
@@ -0,0 +1,65 @@
...
+#  user_whitelist     :string           default([]), not null, is an Array
+#  created_at         :datetime         not null
+#  updated_at         :datetime         not null
+#
+# Indexes
+#
+#  index_app_settings_on_name  (name)
+#
+class AppSetting < ActiveRecord::Base
+    PCT_ACTIVE_VALUES = [100, 90, 80, 70, 60, 50, 40, 30, 20, 10, 0]
```
brendanshean: I've noticed some of the ruby code has four spaces indentation instead of the standard two.  Is this intentional?

I think consistency helps with readability as four spaces usually think Python when I see that spacing.

brendanshean (approved):
Some minor comments but looks good.

services/QuillLMS/app/controllers/api/v1/app_settings_controller.rb
anathomical: Since we want this to work for non-logged-in users, this probably needs to be unauthenticated.

services/QuillLMS/app/models/app_setting.rb
dandrabik: This is in our Ruby Style Guide. Let's convert this to 2 spaces for indents: https://github.com/empirical-org/ruby#indentation

services/QuillLMS/app/models/app_setting.rb
dandrabik: I agree with this. Let's use full words for `app_setting`. I think this is also worth adding to our style guide.

services/QuillLMS/app/controllers/api/v1/app_settings_controller.rb
```diff
@@ -0,0 +1,17 @@
...
+    user = User.find app_setting_index_params
+
+    jayson = app_settings.each_with_object({}) do |app_setting, memo|
```
anathomical: Honest question here: is it better to make this endpoint ask questions about individual features rather than all flags?  We could make feature name a query param (since the front-end will have to know the name of the feature it's checking for anyway).  That would prevent us from running into issues around performance of lookups or payload size if we have hundreds of features in the system.

services/QuillLMS/app/models/app_setting.rb
```diff
@@ -0,0 +1,65 @@
...
+    validates :name, uniqueness: true
+
+    def self.enabled?(name:, user:)
+        as = AppSetting.find_by_name name
+        return false unless as 
+        return false unless user.respond_to? :id 
+        as._enabled? user.id
+    end
+
+    def _enabled? user
```
dandrabik: Let's use parentheses for method definitions that have params, per style guide: https://github.com/empirical-org/ruby#method-definitions

services/QuillLMS/db/migrate/20210614205654_create_app_settings.rb
```diff
@@ -0,0 +1,14 @@
+class CreateAppSettings < ActiveRecord::Migration
+  def change
+    create_table :app_settings do |t|
+      t.string :name, null: false
+      t.string :user_whitelist, null: false, default: [], array: true
+      t.boolean :enabled_for_admins, null: false, default: false
+      t.boolean :enabled, null: false, default: false
+      t.integer :pct_active, null: false, default: 0
```
dandrabik: Let's use full words for column names, either `percent` or `percent_active`.

services/QuillLMS/db/migrate/20210614205654_create_app_settings.rb
```diff
@@ -0,0 +1,14 @@
+class CreateAppSettings < ActiveRecord::Migration
+  def change
+    create_table :app_settings do |t|
+      t.string :name, null: false
+      t.string :user_whitelist, null: false, default: [], array: true
+      t.boolean :enabled_for_admins, null: false, default: false
```
dandrabik, line 6: `admin_enabled` is shorter and is closer to the naming pattern for the other fields.

services/QuillLMS/db/migrate/20210614205654_create_app_settings.rb
```diff
@@ -0,0 +1,14 @@
+class CreateAppSettings < ActiveRecord::Migration
+  def change
+    create_table :app_settings do |t|
+      t.string :name, null: false
+      t.string :user_whitelist, null: false, default: [], array: true
```
dandrabik: It might be better to make this clear these are ids, e.g. `user_ids` or `user_ids_allowed`. Let's avoid the term whitelist.

services/QuillLMS/app/models/app_setting.rb
anathomical: Agreed.  In fact, we should maybe see if there's a Rubocop rule we can turn on to enforce this.

services/QuillLMS/app/models/app_setting.rb
anathomical: I like a rename here.  `enabled_for_user?` seems reasonable to me?  Or maybe we rename `enabled?` to `feature_enabled?` or something?

dandrabik: Let's avoid `_` starting methods, since that's an uncommon pattern in Ruby. I agree that this should be named something along the lines of `user_enabled?(user)` or `enabled_for_user?(user)`

services/QuillLMS/app/models/app_setting.rb
```diff
@@ -0,0 +1,65 @@
...
+    PCT_ACTIVE_VALUES = [100, 90, 80, 70, 60, 50, 40, 30, 20, 10, 0] 
+
+    validates :pct_active, inclusion: { in: PCT_ACTIVE_VALUES }
+    validates :name, uniqueness: true
+
+    def self.enabled?(name:, user:)
+        as = AppSetting.find_by_name name
+        return false unless as 
+        return false unless user.respond_to? :id 
+        as._enabled? user.id
```
anathomical: Won't this break for many cases?  It looks like `_enabled?` takes a `User` object (and references its attributes), but we're just passing the `id` here.

services/QuillLMS/app/models/app_setting.rb
anathomical: Agreed.  I know it's stylistic, but I definitely find `return true if enabled_for_admins && user.admin?` to be more readable in Ruby.

services/QuillLMS/app/models/app_setting.rb
```diff
@@ -0,0 +1,65 @@
...
+        
+        if user_whitelist.include? user.id.to_s
+            return true 
+        end
+
+        false
+    end
+
+    def user_in_rollout_bucket? user_id 
+        enabled_buckets.include? user_id.digits.first
```
anathomical: I am deeply uncomfortable using a raw value from `user_id` to determine bucket eligibility.  The biggest problem I see here is that a given user will ALWAYS be in the same bucket.  So if we roll twenty features out to 10% of users at once, the same 10% will get all 20 in concert.  We won't be able to build cohorts of "Got A and B, but not C".  At the very least I think you need some sort of system that incorporates the `AppSetting.name` into the calculation.

I'd be tempted to use something like `enabled_buckets.include? ("#{name}#{user_id}".bytes.sum % 10)`

Also, if you tweak this to something that you can do a raw `%` with, you could even forgo the `enabled_buckets` function entirely and just use `user_id_based_number % pct_active == 0` or something.

services/QuillLMS/app/models/app_setting.rb
dandrabik: I don't think this method works. You are passing an `id` to `._enabled?` but it expects a `user`. (Stubbing in tests may have prevented this from being tested).

services/QuillLMS/app/models/app_setting.rb
```diff
@@ -0,0 +1,65 @@
...
+    end
+
+    # The range of possible buckets is 0-9 inclusive, so
+    # we transform the user-facing pct_active value by taking
+    # the first radix integer and subtracting 1
+    private def enabled_buckets
+        @enabled_buckets ||=    if pct_active == 0
```
dandrabik: Let's not space this indents this way. Let's keep them left aligned.

Also, I don't think this logic works for percentages 0-100 that aren't divisible by 10, e.g. 25%, 75%, etc.

services/QuillLMS/app/serializers/app_setting_serializer.rb
```diff
@@ -0,0 +1,20 @@
...
+#  pct_active         :integer          default(0), not null
+#  user_whitelist     :string           default([]), not null, is an Array
+#  created_at         :datetime         not null
+#  updated_at         :datetime         not null
+#
+# Indexes
+#
+#  index_app_settings_on_name  (name)
+#
+class AppSettingSerializer < ActiveModel::Serializer
```
dandrabik: I don't think you are using this serializer. If not, let's remove it.

services/QuillLMS/app/models/app_setting.rb
```diff
@@ -0,0 +1,65 @@
...
+class AppSetting < ActiveRecord::Base
+    PCT_ACTIVE_VALUES = [100, 90, 80, 70, 60, 50, 40, 30, 20, 10, 0] 
+
+    validates :pct_active, inclusion: { in: PCT_ACTIVE_VALUES }
+    validates :name, uniqueness: true
+
+    def self.enabled?(name:, user:)
+        as = AppSetting.find_by_name name
+        return false unless as 
+        return false unless user.respond_to? :id
```
dandrabik: I think it's ok to assume this is a user object without checking that it responds to `:id`

services/QuillLMS/app/controllers/api/v1/app_settings_controller.rb
```diff
@@ -0,0 +1,17 @@
+class API::V1::AppSettingsController < ApplicationController
+  def index
+    app_settings = AppSetting.all
+    user = User.find app_setting_index_params
+
+    jayson = app_settings.each_with_object({}) do |app_setting, memo| 
+      memo[app_setting.id] = app_setting._enabled?(user)
```
dandrabik: Shouldn't this use the setting `name` as the key?

services/QuillLMS/app/models/app_setting.rb
```diff
@@ -0,0 +1,65 @@
...
+#
+#  index_app_settings_on_name  (name)
+#
+class AppSetting < ActiveRecord::Base
+    PCT_ACTIVE_VALUES = [100, 90, 80, 70, 60, 50, 40, 30, 20, 10, 0] 
+
+    validates :pct_active, inclusion: { in: PCT_ACTIVE_VALUES }
+    validates :name, uniqueness: true
+
+    def self.enabled?(name:, user:)
```
dandrabik: I have open questions about this method. It's not used in the code (and I think it has a bug in it anyway). There is a comment that this method isn't being used in the controller for performance reasons, but if this isn't performant for our use case, maybe there should be a class method that is, e.g. `self.all_enabled_for_user?(user)`?

services/QuillLMS/config/routes.rb
```diff
@@ -369,6 +369,8 @@
       get 'rule_feedback_history/:rule_uid' => 'rule_feedback_histories#rule_detail'
       get 'prompt_health/:activity_id' => 'rule_feedback_histories#prompt_health'
 
+      get 'app_settings/:user_id' => 'app_settings#index'
```
dandrabik: I think we'd want this is a nested resource route added to the users resource below, e.g.
```ruby
resources :users, only: [:index] do
  resources :app_settings, only: [:index]
end

**Review by dandrabik — CHANGES_REQUESTED** [review]
I left a good amount of comments and requested changes here. I think some of these will be easier to talk through, so I'll schedule some time this afternoon to chat through them. Please take some time to read through my comments before then.

File: services/QuillLMS/app/serializers/app_setting_serializer.rb (hunk 20 above)
Comment (anathomical) [inline services/QuillLMS/app/serializers/app_setting_serializer.rb]: I do not object to using a serializer, but we've started moving toward defining `serializable_hash` on the models rather than building a separate serializer class.  Not strongly committed to that, but I do like it as an approach (keeping things all in one place).

File: services/QuillLMS/config/routes.rb (hunk 24 above)
Comment (anathomical) [inline services/QuillLMS/config/routes.rb]: I think it might be cleaner to just do `resources :app_settings, only[:show]`, and then renaming `index` to `show` in the controller.  Then you can access the param `id` rather than `user_id`.  Not sure this is the right approach, but letting Rails build the routing through `resources` is often cleaner/more consistent.

File: services/QuillLMS/db/migrate/20210614205654_create_app_settings.rb (hunk 8 above)
Comment (anathomical) [reply services/QuillLMS/db/migrate/20210614205654_create_app_settings.rb]: Strong agree here.  I know we're validating it as unique, but we should probably enforce it at the db level, too.

File: services/QuillLMS/spec/models/app_setting_spec.rb [hunk 25]
```diff
@@ -0,0 +1,109 @@
...
+
+  describe '#_enabled?' do 
+    
+    context 'override is false' do 
+      let(:as1) { create(:app_setting, pct_active: 40, enabled: false) }
+
+      it 'should return false' do 
+        u1 = create(:user)
+        expect(as1._enabled?(u1)).to be false
+      end
```
anathomical: Isn't it possible that this will intermittently fail if the `id` of `u1` ends up falling into `pct_active` of 40?

services/QuillLMS/spec/models/app_setting_spec.rb
```diff
@@ -0,0 +1,109 @@
...
+      it 'should return true when user is admin' do 
+        allow_any_instance_of(User).to receive(:admin?) { true }
+        u1 = create(:user, id: 9)
+        expect(as1._enabled?(u1)).to be true
+      end
+    
+      it 'should return true for non-admin user when user is in whitelist' do 
+        allow_any_instance_of(User).to receive(:admin?) { false } 
+        as1.user_whitelist = [9]
+        u1 = create(:user, id: 9)
```
anathomical: I think that it might be possible that somehow a `User` with id 9 could exist in the db based on earlier tests.  Probably safer to do.
```ruby
u1 = create(:user)
as1.user_whitelist[u1.id]
```

services/QuillLMS/spec/models/app_setting_spec.rb
```diff
@@ -0,0 +1,109 @@
...
+        expect(as1._enabled?(u1)).to be true
+      end    
+
+      it 'should return false when user does not meet enabled criteria' do 
+        u1 = create(:user, id: 9)
```
anathomical: Do we need to specify a specific ID here?

services/QuillLMS/db/migrate/20210614205654_create_app_settings.rb
happythenewsad: Yes? Having duplicately-named settings would be very confusing. In addition, it would force access by id rather than name, which is less human friendly.

services/QuillLMS/spec/models/app_setting_spec.rb
```diff
@@ -0,0 +1,109 @@
...
+
+  end
+
+  describe '#user_in_rollout_bucket?' do 
+    context 'user in bucket' do 
+      it 'should place users in buckets according to pct_active' do 
+        as1 = create(:app_setting, pct_active: 40, enabled: true) 
+        u1 = create(:user, id: 2)
+        u2 = create(:user, id: 43)
+        u3 = create(:user, id: 15)
```
anathomical: Again, I'm slightly worried about specifying `id` values, but it's probably safe?

anathomical (changes requested):
Got a couple of suggestions here and there, but definitely want to see some sort of different implementation around buckets that ensures that we don't always end up with the same users "in" every feature.

services/QuillLMS/app/controllers/api/v1/app_settings_controller.rb
happythenewsad: Per above conversations, this API is subject to change. But currently, `_enabled?` does not require a name, because the AppSetting instance already has access to itself.

services/QuillLMS/spec/models/app_setting_spec.rb
happythenewsad: The bucket logic requires this. We only need to hardcode ids in specs for testing.

services/QuillLMS/spec/controllers/api/v1/app_settings_controller_spec.rb
happythenewsad: No; removing

services/QuillLMS/app/controllers/api/v1/app_settings_controller.rb
dandrabik: As discussed - Will use name here.

services/QuillLMS/app/models/app_setting.rb
dandrabik: As discussed, we will create another class method to return all of the features that are enabled for the user. This will be used by the index controller endpoint.

services/QuillLMS/db/migrate/20210614205654_create_app_settings.rb
dandrabik: Resolved - user_ids_allow_list

services/QuillLMS/spec/models/app_setting_spec.rb
happythenewsad: n/a - implementation has changed.

services/QuillLMS/app/controllers/api/v1/app_settings_controller.rb
happythenewsad: I'm implementing both for now.

services/QuillLMS/config/routes.rb
```diff
@@ -385,7 +386,12 @@
           get 'level_zero_concepts_with_lineage'
         end
       end
-      resources :users,                   only: [:index]
+      
+      resources :users, only: [:index] do
+        resources :app_settings, only: [:index] 
+        get 'app_setting(/:name)' => 'app_settings#show'
```
happythenewsad: This line cannot be completely "rails-y" since it is not a RESTful route; we want the parameter to be name rather than id, so that the client does not need to care about AppSetting ids.

services/QuillLMS/lib/tasks/app_settings.rake
```diff
@@ -0,0 +1,23 @@
...
+      name: args[:name],
+      enabled: args[:enabled],
+      enabled_for_admins: args[:enabled_for_admins],
+      percent_active: args[:percent_active],
+      user_ids_allow_list: args[:user_ids_allow_list]
+    )
+    puts "AppSetting with name #{app_setting.name} created."
+  end
+
+  desc 'Updates user allow list for existing AppSetting.'
```
happythenewsad, line 15: Adding the user list rake task to this PR

services/QuillLMS/config/routes.rb
anathomical: You could route this the rails-y way, but the param in the controller would be `:id` instead of `:name`.  That might be confusing enough that it's worth making the router a bit more complicated to name the param the right thing, so I'm okay with this.

services/QuillLMS/db/migrate/20210614205654_create_app_settings.rb
```diff
@@ -0,0 +1,14 @@
+class CreateAppSettings < ActiveRecord::Migration
+  def change
+    create_table :app_settings do |t|
+      t.string :name, null: false
+      t.string :user_ids_allow_list, null: false, default: [], array: true
```
anathomical: Since these are IDs, should we be using an integer for this column?  It probably doesn't really matter.

services/QuillLMS/spec/models/app_setting_spec.rb
```diff
@@ -0,0 +1,117 @@
...
+        'enabled2' => true
+      }
+
+      expect(AppSetting.all_enabled_for_user(user)).to eq expected
+    end
+  end
+
+  describe '#enabled_for_user?' do 
+    context 'override is false' do 
+      let(:as1) { create(:app_setting, percent_active: 40, enabled: false) }
```
anathomical: The use of `40` here could imaginably cause `enabled_for_user?` to return inconsistent values based on the arbitrary name for the `AppSetting` created by the factory.  Presumably some combination of the name sequence and the user.id would trigger as enabled or disabled for `percent_active` of 40.  I think that to capture the intent of this test (which seems to be checking the impact of the `enabled` flag, you should set `percent_active` to 100.

services/QuillLMS/app/controllers/api/v1/app_settings_controller.rb
```diff
@@ -0,0 +1,22 @@
...
+    })
+  end
+
+  private def app_setting_index_params
+    params.require(:user_id)
+  end
+
+  private def app_setting_show_params
+    params.permit(:user_id, :name)
+  end
```
anathomical, line 21: Should we be getting the user ID from `current_user` instead of expecting it to be passed in?  I can imagine plenty of front-end environments that don't actually know the details about the current user, and thus might not have a known ID to pass in.

anathomical (changes requested):
I think that we want to use `current_user` to identify users rather than expecting the front-end to pass an ID into the controller.  I could be talked out of it, but that's my current inclination.

services/QuillLMS/config/routes.rb
dandrabik: You can change the name of the `param` for a resource: https://guides.rubyonrails.org/routing.html#overriding-named-route-parameters

This seems like a reasonable setup since `name` will likely be used in place of `id` for any route we add.

```ruby
resources :app_settings, only: [:index, :show], param: :name

File: services/QuillLMS/lib/tasks/app_settings.rake [hunk 34]
```diff
@@ -0,0 +1,23 @@
...
+  # Example usage: rake 'app_settings:updateUserIdsAllowList[theName, 4 5]'
+  task :updateUserIdsAllowList, [:name, :user_ids_allow_list] => :environment do |t, args|
```
dandrabik: Let's use snake case for this task name https://github.com/empirical-org/ruby#naming

services/QuillLMS/spec/models/app_setting_spec.rb
```diff
@@ -0,0 +1,117 @@
...
+      it 'should return false when user does not meet enabled criteria' do 
+        u1 = create(:user)
+        expect(as1.enabled_for_user?(u1)).to be false
+      end
+    end
+
+  end
+
+  describe '#user_in_rollout_bucket?' do 
+    let(:random_numbers) { (1..1000).map(&:to_i) }
```
dandrabik: Note, these aren't random numbers. Also `(1..1000).to_a` is equivalent.

I'm only mentioning the naming thing because it's good to avoid randomness in tests. You don't have randomness (which is a good thing), but the naming making it seem like you do.

services/QuillLMS/db/migrate/20210614205654_create_app_settings.rb
dandrabik: I think it makes sense to store these as integers since they are integers.

services/QuillLMS/spec/models/app_setting_spec.rb
```diff
@@ -0,0 +1,117 @@
...
+        expect(result.filter{|r| r}.count).to eq 482
+      end
+
+      it 'when percent_active is low, few users should be in bucket' do
+        app_setting = create(:app_setting, name: 'lorem2', enabled: true, percent_active: 10)
+        result = random_numbers.map {|r| app_setting.user_in_rollout_bucket?(r) }
+        expect(result.filter{|r| r}.count).to eq 97
+      end
+
+      it 'when percent_active is high, many users should be in bucket' do
```
dandrabik, line 105: A few thoughts on these tests. 

1. How fast are these tests? We could make this `1000` number lower if they are slow.
1. I think they are fine as a sanity test, but are a little odd how we're testing how 50 percent must equal exactly 48.2 percent. We might consider testing in a range (+/- 2% say) so that if we changed implementation, the specs would still be valid. I don't have strong feelings about this, since if we changed implementation we might add more tests anyway.
2. Probably worth adding a spec for `percent_active: 100`. Looking at the logic for this implementation it will be all 1000 (I believe), but a good gut check since that's an important use case.

services/QuillLMS/spec/models/app_setting_spec.rb
```diff
@@ -0,0 +1,117 @@
...
+
+  end
+
+  describe '#user_in_rollout_bucket?' do 
+    let(:random_numbers) { (1..1000).map(&:to_i) }
+
+    context 'statistical sanity test for hashing algorithm' do 
+      it 'when percent_active is 50, users should be evenishly distributed' do 
+        app_setting = create(:app_setting, name: 'lorem1', enabled: true, percent_active: 50)
+        result = random_numbers.map {|r| app_setting.user_in_rollout_bucket?(r) }
```
dandrabik: You can pass a block that returns truthy for `count` to do this `map/filter/count` in one step, e.g.
```ruby
count = numbers.count {|i| app_setting.user_in_rollout_bucket?(i) }

File: services/QuillLMS/spec/models/app_setting_spec.rb [hunk 38]
```diff
@@ -0,0 +1,117 @@
...
+
+      it 'should return false' do 
+        u1 = create(:user)
+        expect(as1.enabled_for_user?(u1)).to be false
+      end      
+
+    end
+
+    context 'override is true, admin is true' do 
+      let(:as1) { create(:app_setting, percent_active: 0, enabled: true, enabled_for_admins: true) }
```
dandrabik: Let's avoid two letter variables (there are a few here).

Also, you are using the generic `create(:user)`  in many tests. You could add a `let(:user) { create(:user) }` to the top of this spec, below `RSpec.describe AppSetting` and just call `user` in the tests (the user is only created if you call it).

dandrabik (changes requested):
This is getting really close and am excited that this will go out soon. 

Mostly small re-naming, style guide, minor refactor things. Marking as a request changes since I think we should migrate the data type for `user_ids` to `integer`.

Also, when this is ready for another review, just Slack me and I'll review right away.

services/QuillLMS/spec/models/app_setting_spec.rb
happythenewsad: the overall file takes .812 s with the 1000-count sanity checks. I've reduced the count to 100, and now the overall file takes .629 s.  I started with count=1000 because I wanted to prove to myself that the algorithm worked. Now that we know it works, a smaller counter (+bigger skew) is acceptable.

happythenewsad:
> I think that we want to use `current_user` to identify users rather than expecting the front-end to pass an ID into the controller. I could be talked out of it, but that's my current inclination.

Yeah, I'm leaning towards that too, now that you mention it. However, I'm developing the client-side portion for feature flags in my next PR, so I'd like to potentially address this in that PR, after I get a feel for the client-side portion.

services/QuillLMS/db/structure.sql
```diff
@@ -6815,826 +6904,419 @@ ALTER TABLE ONLY public.auth_credentials
...
-INSERT INTO schema_migrations (version) VALUES ('20210511161300');
-
-INSERT INTO schema_migrations (version) VALUES ('20210518151248');
-
-INSERT INTO schema_migrations (version) VALUES ('20210518162719');
-
-INSERT INTO schema_migrations (version) VALUES ('20210521152206');
-
-INSERT INTO schema_migrations (version) VALUES ('20210528142650');
+INSERT INTO "schema_migrations" (version) VALUES
```
happythenewsad, line 6907: This is the new rails 5 syntax

services/QuillLMS/spec/models/app_setting_spec.rb
```diff
@@ -0,0 +1,119 @@
...
+        expect(result).to eq 8
+      end
+
+      it 'when percent_active is high, many users should be in bucket' do
+        app_setting = create(:app_setting, name: 'lorem3', enabled: true, percent_active: 95)
+        result = pseudo_random_numbers.count {|r| app_setting.user_in_rollout_bucket?(r) }
+        expect(result).to eq 99
+      end
+
+      it 'when percent_active is 0, all users should be in bucket' do
```
dandrabik: typo - I think this should read `when percent_active is 100`

services/QuillLMS/spec/models/app_setting_spec.rb
```diff
@@ -0,0 +1,119 @@
...
+
+      it 'should return false when user does not meet enabled criteria' do 
+        expect(app_setting_1.enabled_for_user?(user)).to be false
+      end
+    end
+
+  end
+
+  describe '#user_in_rollout_bucket?' do 
+    let(:pseudo_random_numbers) { (1..100).to_a }
```
dandrabik: Let's just remove random. Let's call this `one_to_a_hundred`.

dandrabik (approved):
Some small comments on minor changes, but once those are made, this looks good to go from my end.

For the outstanding comment from @anathomical, I would ping him to talk through the `params[:user_id]` vs `QuillAuthentication/current_user` approach and come to a decision. Both are very similar and it would be easy to change if needed.

happythenewsad:
> Some small comments on minor changes, but once those are made, this looks good to go from my end.
> 
> For the outstanding comment from @anathomical, I would ping him to talk through the `params[:user_id]` vs `QuillAuthentication/current_user` approach and come to a decision. Both are very similar and it would be easy to change if needed.

Thomas and I discussed; I'm going to explore the `current_user` option in the front-end PR

PR #7799, opened 2021-06-15
comprehension/display-rule-names-in-reports

anathomical:
## WHAT
Display the rule name in Session reports
## WHY
The current report only displays the rule UID which makes it hard to know if it's actually the right rule or not
## HOW
It's a bit hacky, but we now load all Rules for the activity on this page, and then look them up by Rule UID when reporting on a specific session.  It's relatively non-performant (adding 10 seconds in testing), but it works and we should be able to optimize that query at a later date.

### Notion Card Links
https://www.notion.so/quill/Add-the-rule-name-to-the-feedback-sessions-page-cf8a6788ba794b6889f4fba1cebf8163

PR Checklist | Your Answer
------------ | -------------
Have you added and/or updated tests? |  No tests around this content


emilia-friedberg (approved):

services/QuillLMS/client/app/bundles/Staff/components/comprehension/activitySessions/sessionView.tsx
```diff
@@ -20,12 +21,34 @@ const SessionView = ({ match }) => {
...
+    queryFn: fetchRules
+  });
+
   // cache session data for updates
   const { data: sessionData } = useQuery({
     queryKey: [`activity-${activityId}-session-${sessionId}`, sessionId],
     queryFn: fetchActivitySession
   });
 
+  if(!rulesData) {
```
eadams17: instead of doing this, you can just do `if(!rulesData && !sessionData)` on line 52 to consolidate

services/QuillLMS/client/app/bundles/Staff/components/comprehension/activitySessions/sessionView.tsx
```diff
@@ -78,13 +102,13 @@ const SessionView = ({ match }) => {
       <Switch>
         <Redirect exact from='/activities/:activityId/activity-sessions/:sessionId' to='/activities/:activityId/activity-sessions/:sessionId/overview' />
         {/* eslint-disable-next-line react/jsx-no-bind */}
-        <Route component={() => <SessionOverview activity={activity} sessionData={sessionData} />} path='/activities/:activityId/activity-sessions/:sessionId/overview' />
+        <Route component={() => <SessionOverview activity={activity} rules={rules} sessionData={sessionData} />} path='/activities/:activityId/activity-sessions/:sessionId/overview' />
```
eadams17, line 95: I don't think we need to be passing a rules prop here?

eadams17 (approved):
Looks good! just two small comments

services/QuillLMS/client/app/bundles/Staff/components/comprehension/activitySessions/sessionView.tsx
anathomical: The overview page actually shows all three of the sub-items, but as far as I can tell it does need this data passed through.

services/QuillLMS/client/app/bundles/Staff/components/comprehension/activitySessions/sessionView.tsx
anathomical: Good call, will update.

PR #7800, opened 2021-06-15
Catch all the nil errors in ActivityHealth and QuestionHealth worker logic

cissyxyu:
## WHAT
We have some very old activities and questions in the database that have oddly formatted JSON data that doesn't match our current schema. This PR adds some more rigorous nil and error catching to the weekly ActivityHealth job's logic, to avoid erroring out on too many of these bad activities.

## WHY
While we shouldn't delete these archived records in case Curriculum wants to refer back to old activities, we do want to do nil catching in our current jobs so these old activities don't trigger too many errors and warnings. We'll also set retry to 1 so that if the health worker fails on one of these activities, it won't keep trying.

## HOW
Nil error catching and exception catching.

### Screenshots

### Notion Card Links
https://www.notion.so/quill/Product-Board-30f97e4eb01246dbb8264253fb385073?p=0e51f92ea9144f7d850c52649e40eb0c

PR Checklist | Your Answer
------------ | -------------

### DIFF
--- a/services/QuillLMS/app/models/question_health_dashboard.rb
+++ b/services/QuillLMS/app/models/question_health_dashboard.rb
@@ -28,7 +28,7 @@ class QuestionHealthDashboard
     begin
       JSON.parse(cms_data.body)
     rescue JSON::ParserError
-      {"percent_common_unmatched": 0, "percent_specified_algos": 0}
+      {}
     end
   end
 
--- a/services/QuillLMS/app/models/question_health_dashboard.rb
+++ b/services/QuillLMS/app/models/question_health_dashboard.rb
@@ -25,7 +25,11 @@ class QuestionHealthDashboard
   end
 
   def cms_dashboard_stats
-    JSON.parse(cms_data.body)
+    begin
+      JSON.parse(cms_data.body)
+    rescue JSON::ParserError
+      {"percent_common_unmatched": 0, "percent_specified_algos": 0}
+    end
   end
 
   private def score_to_attempts(score)
--- a/services/QuillLMS/app/models/question_health_dashboard.rb
+++ b/services/QuillLMS/app/models/question_health_dashboard.rb
@@ -25,10 +25,9 @@ class QuestionHealthDashboard
   end
 
   def cms_dashboard_stats
-      JSON.parse(cms_data.body)
-    rescue JSON::ParserError
-      {}
-    end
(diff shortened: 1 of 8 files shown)


services/QuillLMS/app/models/question_health_obj.rb
```diff
@@ -17,15 +17,15 @@ def run
     health_dashboard = QuestionHealthDashboard.new(@activity.id, @question_number, @question.uid)
     data = @question.data
     {
-      url: ENV['DEFAULT_URL'].to_s + "/" + @tool + "/#/admin/" + question_url + "/" + @question.uid + "/responses",
+      url: [ENV['DEFAULT_URL']&.to_s, "/", @tool, "/#/admin/", question_url, "/", @question&.uid, "/responses"].join(""),
```
brendanshean: nit:  you can just use `join` with no parameters.

happythenewsad (approved):

services/QuillLMS/app/models/question_health_dashboard.rb
```diff
@@ -25,7 +25,11 @@ def percent_reached_optimal_for_question
   end
 
   def cms_dashboard_stats
-    JSON.parse(cms_data.body)
+    begin
+      JSON.parse(cms_data.body)
+    rescue JSON::ParserError
+      {}
+    end
```
anathomical: Cool syntax trick that you can use here if you want: for very simple functions like this one, you can `rescue` the whole thing without a `begin` block:
```ruby
def cms_dashboard_stats
  JSON.parse(cms_data.body)
rescue JSON::ParserError
  {}
end
```

services/QuillLMS/app/models/question_health_obj.rb
anathomical: Or, alternative, you could use string interpolation which I think is more readable:
```ruby
url: "#{ENV['DEFAULT_URL']}/#{@tool}/#/admin/#{question_url}/#{@question&.uid}/responses
```
Also worth noting: this URL will be malformed if `@question` doesn't exist.  The `@question&.uid` will make the end of the url `/nil/responses`.  Which might be better than erroring?

anathomical (approved):
Love the updated test cases.  A couple of style suggestions, but all  of the logic here makes sense to me and could be merged without update.

PR #7806, opened 2021-06-16
update missing and extra word algorithms to handle two-word changes

emilia-friedberg:
## WHAT
Update the missing and extra word algorithms to handle two-word changes.

## WHY
Devin noticed that this algorithm wasn't catching two-word changes and she and Peter Gault agreed that it probably made sense to update it to do that.

## HOW
Just increase the max change length from 2 to 3, and the total number of changes from 1 to 2 so that you can add or remove up to two words, sequentially or otherwise, and still have the algorithm match.

### Screenshots

### Notion Card Links
https://www.notion.so/quill/Connect-Algorithm-Issues-Extra-word-and-missing-word-algorithms-not-working-if-more-than-1-error-is--6f154a793d1b487390ade4f049eb397a

PR Checklist | Your Answer
------------ | -------------

### DIFF
--- a/packages/quill-marking-logic/src/libs/matchers/change_object_match.ts
+++ b/packages/quill-marking-logic/src/libs/matchers/change_object_match.ts
@@ -10,6 +10,8 @@ import {conceptResultTemplate} from '../helpers/concept_result_template'
 import {getFeedbackForMissingWord} from '../helpers/joining_words_feedback'
 import {sortByLevenshteinAndOptimal} from '../responseTools'
 
+const CHANGED_WORD_MAX = 3
+
 export interface TextChangesObject {
   missingText: string|null,
   extraneousText: string|null,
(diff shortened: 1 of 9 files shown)


packages/quill-marking-logic/src/libs/matchers/change_object_match.ts
```diff
@@ -10,6 +10,8 @@ import {conceptResultTemplate} from '../helpers/concept_result_template'
 import {getFeedbackForMissingWord} from '../helpers/joining_words_feedback'
 import {sortByLevenshteinAndOptimal} from '../responseTools'
 
+const CHANGED_WORD_MAX = 3
```
dandrabik, line 13: This constant makes the code much more readable, thanks for adding! You could consider making this `2` and changing the logic to `<=` since 2 words is really the max.

anathomical (approved):
I love the detailed extra tests covering cases of in order and out of order.

happythenewsad (approved):

dandrabik (approved):
Looks good. One small comment about changing a var name, but looks good to me.

I also wonder if this would require us to run rematch all for questions, since it would change some core logic.

PR #7807, opened 2021-06-16
Fix null activity sessions params object

brendanshean:
## WHAT
This PR has two commits.  One is for an already approved Rails 5.0.7 upgrade.  The other is for a bug fix on that upgrade.

## WHY
In Rails 5+, the params hash is treated as an object, specifically ActionController::Parameters.  This object utilizes Strong Parameters which leads to different functionality when passed as a method parameter as opposed to a plain hash. ([more detail here](https://eileencodes.com/posts/actioncontroller-parameters-now-returns-an-object-instead-of-a-hash/)).  In short, any unpermitted parameters are discarded when  passed methods like `create` or `update` on ActiveRecord objects.   For this particular bug, unpermitted parameters were being passed to a [ConceptResult.bulk_insert](https://github.com/empirical-org/Empirical-Core/pull/7807/commits/237673fbc3673ff48e7fe2f2a65a0c78857e7b7e#diff-fb0cb613d7c02517cbea8eab235eb3f3effa4c0340ce0d8929139e8afdb231b7R72) which triggered an error since the method doesn't handle empty parameters.

Also, this [spec](https://github.com/empirical-org/Empirical-Core/pull/7807/commits/237673fbc3673ff48e7fe2f2a65a0c78857e7b7e#diff-e2f9583c4bd3d0b8edc7a57af23d278eab9e54b64fd28be3d3d7f8dae23f43dbR69) needed fixing since it wasn't actually sending any `concept_uid` as a parameter which in turn meant that this [code](https://github.com/empirical-org/Empirical-Core/blob/237673fbc3673ff48e7fe2f2a65a0c78857e7b7e/services/QuillLMS/app/controllers/api/v1/activity_sessions_controller.rb#L79) was never being tested.

## HOW
To fix this, we do two things:

1) Specify the [permitted attributes](https://github.com/empirical-org/Empirical-Core/blob/237673fbc3673ff48e7fe2f2a65a0c78857e7b7e/services/QuillLMS/app/controllers/api/v1/activity_sessions_controller.rb#L115) 

2) Convert each concept_result in the [@concept_results](https://github.com/empirical-org/Empirical-Core/blob/237673fbc3673ff48e7fe2f2a65a0c78857e7b7e/services/QuillLMS/app/controllers/api/v1/activity_sessions_controller.rb#L109) to a hash.

NB.  I had to update magic numbers on two of the specs with new values based on what I think is desired functionality, but please confirm that they make sense:

1)  [First spec](https://github.com/empirical-org/Empirical-Core/pull/7807/commits/237673fbc3673ff48e7fe2f2a65a0c78857e7b7e#diff-e2f9583c4bd3d0b8edc7a57af23d278eab9e54b64fd28be3d3d7f8dae23f43dbR91)
2)  [Second spec](https://github.com/empirical-org/Empirical-Core/pull/7807/commits/237673fbc3673ff48e7fe2f2a65a0c78857e7b7e#diff-e2f9583c4bd3d0b8edc7a57af23d278eab9e54b64fd28be3d3d7f8dae23f43dbR100)

PR Checklist | Your Answer
------------ | -------------


services/QuillLMS/engines/comprehension/Gemfile
```diff
@@ -9,7 +9,6 @@ gemspec
 # your gemspec. These might include edge Rails or gems from your path or
 # Git. Remember to move these dependencies to your gemspec before releasing
 # your gem to rubygems.org.
-gem 'google-cloud-automl'
```
dandrabik, line 12: Nice cleanup here, thanks!

services/QuillLMS/config/application.rb
```diff
@@ -31,6 +31,7 @@ class Application < Rails::Application
     config.autoload_paths += %W(
       #{config.root}/app/controllers/concerns
       #{config.root}/lib
+      #{config.root}/app/lib
```
dandrabik: Wasn't this already deployed? (It's not in `develop` although the actual directory `app/lib` is.). Not a big deal, since it does seem to be working in prod, just flagging in case it indicates a large commit/merge issue.

dandrabik (approved):
Looks good to me. Small comment around something I thought was already in prod, but looks good to ship :+1:

Given that we've already deployed this once and only had a few issues, I think we can deploy this at any time of day, i.e. treat it like a regular code deploy.

services/QuillLMS/app/controllers/api/v1/activity_sessions_controller.rb
```diff
@@ -104,12 +106,30 @@ def destroy
...
       params.delete(:concept_results)
     end
   end
 
+  private def concept_results_permitted_params
+    {
+      concept_results: [
+        :concept_uid,
+        :question_type,
+        metadata: [
```
dandrabik: So I think we need to double check this `metadata` params list for the different activity types. Here's one from Proofreader that includes  `unchanged` and `instructions` keys. Let's look at all of the different question_types.
```
{"answer":"on","prompt":"He put his camera lens \u003cstrong\u003earound/on\u003c/strong\u003e the glass and took one of the most famous photos in history.","correct":1,"unchanged":false,"instructions":"\u003cp\u003eRead the passage. At each underlined part of the passage, choose the correct option and delete the incorrect option and the slash (/).\u003c/p\u003e","questionNumber":6}

**Review by dandrabik — CHANGES_REQUESTED** [review]
Ok, found a small issue around the params for concept results. Let's do another sweep to ensure that we have the full list.

File: services/QuillLMS/config/application.rb (hunk 2 above)
Comment (brendanshean) [reply services/QuillLMS/config/application.rb]: Good question.  We moved various `lib/*.rb` files to app/lib in this commit `https://github.com/empirical-org/Empirical-Core/pull/7678` but I must have left this configuration out which would suggest this configuration is unnecessary.  I will remove it.

File: services/QuillLMS/app/controllers/api/v1/activity_sessions_controller.rb (hunk 3 above)
Comment (dandrabik) [reply services/QuillLMS/app/controllers/api/v1/activity_sessions_controller.rb]: Hmm, so looking at how many possible params there are, I wonder if it would be better to just allow any keys since that is feature-parity to the old code (note, I'm not entirely sure if this is even possibe, i.e. if you can do `metadata: [{}]`). 

It's less safe than named params (although it is the same as the current code), but I'm slightly worried we might be missing params here. Curious your thoughts (I could be convinced either way).

https://api.rubyonrails.org/classes/ActionController/Parameters.html#method-i-permit
Screen Shot 2021-06-21 at 2 27 43 PM

**Review by dandrabik — APPROVED** [review]
I left a comment about params that you might consider, but approving as good to go.

Leaving the decision around params up to you.

**Review by anathomical — APPROVED** [review]
Looks good to me!

File: services/QuillLMS/app/controllers/api/v1/activity_sessions_controller.rb [hunk 4]
```diff
@@ -104,12 +106,27 @@ def destroy
 
   private def transform_incoming_request
     if params[:concept_results].present?
-      @concept_results = params.delete(:concept_results)
+      @concept_results = params.delete(:concept_results).map do |concept_result|
+        concept_result
+          .permit(concept_results_permitted_params)
+          .merge(metadata: concept_result[:metadata].permit!)
```
dandrabik, line 112: This seems like a reasonable solution for 5.0.

dandrabik (approved):
:+1: 🚢

PR #7817, opened 2021-06-21
Stop filling in "directions" with last feedback received in Concept Result records

cissyxyu:
## WHAT
For a long time now, we've filled in the **directions** field of concept results with either the Question Instructions for a question (if 1st attempt) or the last feedback a student received (if it's their 2nd, 3rd, 4th, or 5th attempt). Curriculum has asked us to stop filling in last feedback on >2nd attempts and just fill it in with the question instructions instead.

## WHY
They believe that seeing the last feedback received under the "Instructions" field in the Question Reports view confuses teacher and doesn't really provide helpful information.

## HOW
Remove the clauses where we're adding in the feedback if attempts > 1.

### Screenshots
Screen Shot 2021-06-21 at 11 01 35 AM

### Notion Card Links
https://www.notion.so/quill/Question-Reports-displaying-incorrect-information-a86ff1504f8a4f368132fb1801822b2d

PR Checklist | Your Answer
------------ | -------------
Have you added and/or updated tests? |  NO, tested manually


emilia-friedberg (changes requested):
So my concern about this approach is that we do want the data for the feedback that the student received, and we do use it in other reports in a way that I think makes more sense than what we're doing in the Question Report. I think we're going to cause more issues if we change the data itself. I think a better solution to this issue would be to add a new attribute here, unless we already have the original instructions saved to the concept results as well, and then use that new attribute in the Question Report view.

anathomical:
I think that Emilia is spot on here.  While it's kind of wonky, it's probably better to just change the way that the Question Report pulls data.  Clearly we have access to the original `instructions` from the first attempt in that report, we should just use it everywhere.  We definitely use this data in the Session reports in which teachers definitely want to know what feedback the student got between attempts.

cissyxyu:
> So my concern about this approach is that we do want the data for the feedback that the student received, and we do use it in other reports in a way that I think makes more sense than what we're doing in the Question Report. I think we're going to cause more issues if we change the data itself. I think a better solution to this issue would be to add a new attribute here, unless we already have the original instructions saved to the concept results as well, and then use that new attribute in the Question Report view.

Yeah I have gone in and added the new attribute `lastFeedback` to the metadata object in ConceptResults. I kept the old behavior as a fallback to prevent errors on old data without `lastFeedback`. I've been testing and will test more on staging before I deploy to make sure this is working on reports for all types of activities.

cissyxyu:
> I think that Emilia is spot on here. While it's kind of wonky, it's probably better to just change the way that the Question Report pulls data. Clearly we have access to the original `instructions` from the first attempt in that report, we should just use it everywhere. We definitely use this data in the Session reports in which teachers definitely want to know what feedback the student got between attempts.

See my comment above on Emilia's comment. I added the new attribute `lastFeedback` so we can now keep that in a more accurately named field. The fallback behavior still uses `directions` so we don't error out on old data.

emilia-friedberg (approved):
So, I'm not sure whether it's worth changing at this point, but I actually don't think `directions` is an inaccurate name for what we've been using that field for - once a student has submitted their initial response, the feedback they get for that _does_ serve as their directions for revisions (ex: "Check your punctuation."). Evidently it can be confusing, since it did confuse you, but I'm not sure if that balances out having to add code that accounts for the same data in two different fields now, which feels inelegant to me. I think what I would have done is either add an `instructions` field that has the data you're trying to show for the Question Report, or just use the `directions` for the first attempt on the Questions Report and not mess with the metadata at all, which I think is what @anathomical was suggesting. That all being said, I think your solution works and since you've already done all the work of writing and testing it I'm going to approve it - I just wanted to flag this for future reference.

PR #7824, opened 2021-06-22
Fix logic for finding Feedback in Rule Feedback History report

anathomical:
## WHAT
Fix the Rules report to actually pull Feedback objects
## WHY
It was supposed to be part of the repot already, but the tests didn't cover confirming that the code pulling it worked
## HOW
We expected `.includes(:feedbacks)` to make this work (doing an auto-magic Rails join to the `Feedback` table for these records), but it turns out that secretly `includes` only works when you have the `id` that's used for the join relationship in the `select`.  Since `Feedback` relies on `rule_id` for its relationships, we need to ensure that the `rules.id` column is part of the `select`, which is what this change does.  While I was at it, I ensured that we ordered `Feedback` records to get the lowest `:order` value.

### Notion Card Links
https://www.notion.so/quill/Rules-Analysis-Fix-rule-note-feedback-bug-and-add-second-layer-feedback-to-toggle-62f1ee623349422989f5ea4981edfa29#423daa25cbc54dc98f60b941024de58a

PR Checklist | Your Answer
------------ | -------------

### DIFF
--- a/services/QuillLMS/app/lib/rule_feedback_history.rb
+++ b/services/QuillLMS/app/lib/rule_feedback_history.rb
@@ -6,6 +6,7 @@ class RuleFeedbackHistory
 
     def self.exec_query(conjunction:, activity_id:, start_date:, end_date:)
         query = Comprehension::Rule.select(<<~SELECT
+          comprehension_rules.id,
           comprehension_rules.uid AS rules_uid,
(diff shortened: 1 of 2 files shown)


eadams17 (approved):
thanks for fixing this!

happythenewsad (approved):

PR #7832, opened 2021-06-23
comprehension/correct-rule-analysis-counts

anathomical:
## WHAT
Correct Rule Analysis reporting counts
## WHY
We were over-zealous in one of our joins resulting in over-counting how many users got a specific feedback for a specific prompt when the rule being triggered was universal
## HOW
I tightened up the `ON` clause of the `JOIN` to ensure that we actually only count the `FeedbackHistory` records if they violated the specified Rule and also were tied to the specified Prompt (that latter part wasn't happening before).

### Notion Card Links
https://www.notion.so/quill/Rules-Analysis-Display-the-number-of-regex-and-universal-rule-responses-in-a-given-activity-prompt-d7c2d2c858c34ce394e21229bb550991

PR Checklist | Your Answer
------------ | -------------


brendanshean (approved):
Sorry for the delay on this.  Very interesting pattern of `yield` with the factories.  I'm guessing these "in-file" factories are declared so as not to pollute the global factories?

happythenewsad (approved):

PR #7834, opened 2021-06-23
Fix string coerce bug with concept results metadata

brendanshean:
## WHAT
Fix a bug involving concept results metadata field being stored as a string instead of json

## WHY
In Rails 5, JSON strings are being stored as a string rather than a Hash or Array.  [more info here](https://github.com/rails/rails/issues/28292)
The presence of a string instead of an array or hash creates a host of problems downstream.

## HOW
Remove `to_json` calls on model objects before they are saved to the database.

### Notion Card Links
https://www.notion.so/quill/Sentry-Teachers-ProgressReports-DiagnosticReportsController-question_view-04f499a8fb1a4070aad378fa418be449
https://www.notion.so/quill/Sentry-Teachers-ProgressReports-DiagnosticReportsController-students_by_classroom-0bcebfe4a0b54ef59e43ead39578cbb0

PR Checklist | Your Answer
------------ | -------------

### DIFF
--- a/services/QuillLMS/app/models/activity_session.rb
+++ b/services/QuillLMS/app/models/activity_session.rb
@@ -345,7 +345,7 @@ class ActivitySession < ActiveRecord::Base
         end
       end&.id
       concept = Concept.find_by_id_or_uid(concept_result[:concept_id])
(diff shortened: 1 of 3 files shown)


dandrabik (approved):
:+:

PR #7836, opened 2021-06-23
Add rake task to fix corrupted concept results metadata stored as string

brendanshean:
## WHAT
Fix concept results metadata that has been stored as a string

## WHY
The metadata is unusable in string format

## HOW
Only recent records have been affected from this corruption.

### Screenshots

### Notion Card Links
https://www.notion.so/quill/Sentry-Teachers-ProgressReports-DiagnosticReportsController-question_view-04f499a8fb1a4070aad378fa418be449
https://www.notion.so/quill/Sentry-Teachers-ProgressReports-DiagnosticReportsController-students_by_classroom-0bcebfe4a0b54ef59e43ead39578cbb0

PR Checklist | Your Answer
------------ | -------------
Have you added and/or updated tests? |  I tested locally and will also test on staging.

### DIFF
--- a/services/QuillLMS/lib/tasks/temporary/concept_results.rake
+++ b/services/QuillLMS/lib/tasks/temporary/concept_results.rake
@@ -1,0 +1,22 @@
+namespace :concept_results do
+  task update_metadata_strings_to_json: :environment do
+    concept_results = ConceptResult.where("id >= 1065661250 AND question_type = 'lessons-slide'")
+
+    puts "Going to check #{concept_results.count} concept results"
+    num_updates = 0
+
+    ActiveRecord::Base.transaction do
+      concept_results.each do |concept_result|
+        metadata = concept_result.metadata
+
+        next unless metadata.is_a?(String)
+
+        concept_result.update(metadata: JSON.parse(metadata))
+        num_updates += 1
+        print '.'
+      end
+    end
+
+    puts "\nUpdated #{num_updates} concept results' metadata"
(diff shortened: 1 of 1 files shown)


dandrabik (approved):

anathomical (changes requested):
The logic here is fine, but this table is almost certainly too big to do this kind of table scan on.  There are literally over a billion rows in the table, so I can't imagine that this runs in any sort of reasonable time-frame.

Additionally, it doesn't seem like it solves whatever was causing the bad data to get in in the first place.  We need to fix whatever's causing the bad data (assuming it's ongoing) in addition to fixing the existing bad data.

brendanshean:
> The logic here is fine, but this table is almost certainly too big to do this kind of table scan on. There are literally over a billion rows in the table, so I can't imagine that this runs in any sort of reasonable time-frame.
> 
> Additionally, it doesn't seem like it solves whatever was causing the bad data to get in in the first place. We need to fix whatever's causing the bad data (assuming it's ongoing) in addition to fixing the existing bad data.

@anathomical The solution for the underlying cause was in a previous PR:  https://github.com/empirical-org/Empirical-Core/pull/7834 

As far as the query goes, I'm only going to be pulling the last 1000 records or so.  If it were the whole table, then I would definitely agree.

anathomical (approved):
Upon further discussion in Slack, I have been convinced that the `WHERE` clause used here is performant enough.

PR #7837, opened 2021-06-23
end comprehension session after last feedback

emilia-friedberg:
## WHAT
Have the Comprehension session save to the LMS after the last attempt for the last prompt has feedback, rather than when students click the "Done" button.

## WHY
To reduce friction.

## HOW
Add `componentDidUpdate` function that checks to see whether this is the case and saves it to the LMS if it is.

### Screenshots

### Notion Card Links
https://www.notion.so/quill/Maxing-out-all-three-attempts-registers-an-activity-complete-6ac1d68c4bfc472f90c1ee64596309b8

PR Checklist | Your Answer
------------ | -------------


anathomical (approved):
Seems like a reasonable solution to me.

PR #7838, opened 2021-06-23
New endpoint in CMS for uploading existing optimal responses

cissyxyu:
## WHAT
We currently have a `create_or_increment` endpoint in the CMS for creating responses if they don't exist or incrementing their count if they do. Curriculum requests a new behavior where you can `create_or_update` a response -- create if it doesn't exist, update and save all the existing attributes if it does exist.

## WHY
This will allow staff to upload optimal responses from a spreadsheet and see the values of that response change to reflect the new attributes in the upload. 

## HOW
Add a `create_or_update` endpoint in the CMS that updates all the attributes in an existing Response record if the record already exists.

Point the `uploadOptimalResponses` functions from the front end to POST to this new endpoint.

### Screenshots

### Notion Card Links
https://www.notion.so/quill/Optimal-Response-Uploads-Override-Existing-Responses-0fb69d042a4b41b1ad8c3156b63ac55e
https://www.notion.so/quill/Refactor-Connect-Upload-Optimal-Responses-workflow-to-update-the-optimal-flag-on-responses-that-al-77ed3ef9fc3a47bf863f0521cca680dd

PR Checklist | Your Answer
------------ | -------------

### DIFF
--- a/services/QuillCMS/app/controllers/responses_controller.rb
+++ b/services/QuillCMS/app/controllers/responses_controller.rb
@@ -43,16 +43,9 @@ class ResponsesController < ApplicationController
   # POST /responses/create_or_update
   def create_or_update
     symbolized_vals = transformed_new_vals(params_for_create).to_h.symbolize_keys
-    response = Response.find_by(text: symbolized_vals[:text], question_uid: symbolized_vals[:question_uid])
-    if !response.present?
-      response = Response.new(symbolized_vals)
-      if !response.text.blank? && response.save
-        AdminUpdates.run(response.question_uid)
-      end
-    else
-      response.update(symbolized_vals)
-    end
-    if response.valid?
+    response = Response.find_or_initialize_by(text: symbolized_vals[:text], question_uid: symbolized_vals[:question_uid])
+    if response.update(symbolized_vals)
+      AdminUpdates.run(response.question_uid) if !response.text.blank?
       render json: response
     else
       render json: response.errors, status: :unprocessable_entity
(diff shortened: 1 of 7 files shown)


services/QuillCMS/app/controllers/responses_controller.rb
```diff
@@ -40,6 +40,25 @@ def create_or_increment
     render json: {}
   end
 
+  # POST /responses/create_or_update
+  def create_or_update
+    symbolized_vals = transformed_new_vals(params_for_create).to_h.symbolize_keys
+    response = Response.find_by(text: symbolized_vals[:text], question_uid: symbolized_vals[:question_uid])
+    if !response.present?
```
brendanshean: It feels like there's some duplication with `create` and `update`.  What about refactoring lines 47-59 with a rough idea of:

`response.present? ? update : create`

anathomical (approved):
I think that Brendan has a solid suggestion about simplifying the controller action, but this approach seems reasonable to me.

services/QuillCMS/app/controllers/responses_controller.rb
cissyxyu: Great suggestion. I was able to go back and refactor it a lot more cleanly with `find_or_initialize_by`

brendanshean: That looks even better and is more Rails-y.  Great job!

brendanshean (approved):
Nice work Cissy!

PR #7851, opened 2021-06-28
Students and teachers are forced to re-authorize Quill

brendanshean:
## WHAT
Configure Google authorization/authentication so that users are only required to authorize when necessary.

## WHY
Currently, all users are forced to reauthorize every time they login.  While authorization gives us a refresh token needed to perform offline background jobs, we are already storing it so it would make sense to only ask for re-authorization when the refresh token expires.

## HOW
Modeled after this [post](https://github.com/zquestz/omniauth-google-oauth2/issues/9#issuecomment-26525644), the fix is to add a new authentication-only route: `/google_oauth2_authentication_only` which will become the new route when a user clicks 'Log In'.  All other flows (e.g. Sign Up) will continue to use our original authentication-and-authorization route `/google_oauth2`.  

When a user clicks log-in, they will be directed to authentication only. The key point here is the `before_action :verify_credentials` which checks to see if the current user is "google authorized".  If they are, there is no need to re-authorize and the flow will proceed to the user profile page.  Otherwise, users will be asked to redirected to the old `/google_oauth2` route to re-authorize.
 
****

### Screenshots

### Notion Card Links
https://www.notion.so/quill/Google-Classroom-Students-and-teachers-are-forced-to-re-authorize-Quill-ffa87e0eb2014f0e907e9374dbb935bc

PR Checklist | Your Answer
------------ | -------------


services/QuillLMS/app/controllers/auth/google_controller.rb
```diff
@@ -1,28 +1,43 @@
...
+      redirect_to url
+    else
+      redirect_to profile_path
     end
+  end
 
-    redirect_to profile_path
+  private def verify_credentials
+    user = User.where('google_id = ? OR email = ?', @profile.google_id&.to_s, @profile.email&.downcase).first
+    return if user.nil?
```
happythenewsad, line 38: Do we need this return if we have the redirect on line 35?

services/QuillLMS/app/services/google_integration/user.rb
```diff
@@ -28,25 +23,29 @@ def update_or_initialize
...
+      params[:google_id] = profile.google_id if profile.google_id.present?
+      params[:role]      = profile.role      if profile.role.present? && user.new_record?
       params[:clever_id] = nil
 
       params
     end
   end
 
-  private def auth_credential_attributes
+  private def find_user_by_google_id_or_email
```
happythenewsad, line 39: Optional: you could make this a class method that accepts params, then reuse it in `google_controller#verify_credentials`

services/QuillLMS/config/routes.rb
```diff
@@ -475,7 +476,8 @@
   get '/sign-up/add-non-k12', to: 'accounts#new'
 
   namespace :auth do
-    get "/google_oauth2/callback" => 'google#google'
+    get '/google_oauth2/callback' => 'google#authorized_and_authenticated'
```
happythenewsad: Note: omniauth 2.x, which we want to upgrade to soon, only accepts POSTs for oauth callbacks: https://github.com/omniauth/omniauth/commit/b72a8db667960ae1a930d0df4f5e5917dcd0ad7a

services/QuillLMS/app/controllers/auth/google_controller.rb
brendanshean: Since the redirect is an `unless` we actually want to exit at this point.

services/QuillLMS/app/services/google_integration/user.rb
brendanshean: I actually toyed with extraction this morning, but opted not to...for now.

emilia-friedberg (approved):

brendanshean:
@dandrabik @anathomical @happythenewsad I'd like to get at least one more approval before pushing up as it is a major change to authentication / authorization flow.

happythenewsad:
> @dandrabik @anathomical @happythenewsad I'd like to get at least one more approval before pushing up as it is a major change to authentication / authorization flow.

Taking a look now @brendanshean

services/QuillLMS/app/controllers/auth/google_controller.rb
```diff
@@ -1,28 +1,43 @@
...
-  before_action :set_user,                        only: :google
-  before_action :save_teacher_from_google_signup, only: :google
-  before_action :save_student_from_google_signup, only: :google
-  before_action :follow_google_redirect,          only: :google
-
-  def google
-    if @user.teacher?
-      GoogleStudentImporterWorker.perform_async(@user.id, 'Auth::GoogleController')
-    end
+  before_action :set_profile, only: [:authorization_and_authentication, :authentication]
```
happythenewsad, line 2: before_action can [take an array of callbacks](https://api.rubyonrails.org/v6.1.4/classes/AbstractController/Callbacks/ClassMethods.html#method-i-before_action). Let's use this style here, to increase readability / reduce LoC

brendanshean: Good idea

services/QuillLMS/spec/models/auth_credential_spec.rb
```diff
@@ -0,0 +1,37 @@
+require 'rails_helper'
+
+describe AuthCredential, type: :model do
+  it { should belong_to(:user) }
+
+  let(:auth_credential) { create(:auth_credential, provider: 'google') }
+
+  context '#google_authorized?' do
+    context 'non google provider' do
+      before { auth_credential.update(provider: 'hooli') }
```
happythenewsad, line 10: 💯

services/QuillLMS/app/controllers/teachers/classroom_units_controller.rb
```diff
@@ -30,9 +30,6 @@ def launch_lesson
       if cuas && cuas.update(locked: false, pinned: true)
         find_or_create_lesson_activity_sessions_for_classroom
         PusherLessonLaunched.run(@classroom_unit.classroom)
-        # if @classroom_unit.is_valid_for_google_announcement_with_specific_user?(current_user)
```
anathomical, line 33: Yay removing dead code!

services/QuillLMS/client/app/bundles/Teacher/components/accounts/login/login_form.jsx
```diff
@@ -92,26 +83,26 @@ class LoginFormApp extends React.Component {
...
-        window.location = body.redirect;
-      } else {
-        let state;
-        if (body.type && body.message) {
-          const errors = {};
-          errors[body.type] = body.message;
-          state = { lastUpdate: new Date(), errors, timesSubmitted: timesSubmitted + 1, };
+      (err, httpResponse, body) => {
+        if (httpResponse.statusCode === 200 && body.redirect) {
+          // console.log(body);
```
happythenewsad: Let's remove the commented out console log.

happythenewsad (approved):
One trivial change request, but looks good

anathomical (approved):
Seems workable to me.

services/QuillLMS/app/controllers/auth/google_controller.rb
```diff
@@ -1,28 +1,48 @@
...
+  before_action :verify_credentials, only: :authentication
```
dandrabik: These routes have the same actions, but logically, shouldn't the `verify_credentials` be on the `authorization_and_authentication` route?

services/QuillLMS/app/controllers/auth/google_controller.rb
```diff
@@ -1,28 +1,48 @@
...
+    only: [:authorization_and_authentication, :authentication]
 
-    if @user.student?
-      GoogleStudentClassroomWorker.perform_async(@user.id)
-    end
+  # The reason for the separate 'authentication' and`authorization_and_authentication` methods lies in the
+  # before_action hook :verify_credentials.  This is only called with the authentication flow and if it fails
+  # will redirect to `google_oauth2` for reauthorization.
 
+  def authorization_and_authentication
```
dandrabik, line 14: Outside the scope of this PR, but separating the "create account" and the "login" even more, might be helpful in future refactors / cleaning up the code, e.g. there's a lot of code in these `before_action`s that branches on `@user.new_record?`

services/QuillLMS/app/models/auth_credential.rb
```diff
@@ -24,4 +24,18 @@
 #
 class AuthCredential < ActiveRecord::Base
   belongs_to :user
+
+  def google_authorized?
+    provider == 'google' && refresh_token_valid?
```
dandrabik: Let's make this `'google'` string a constant somewhere and start replacing occurrences of this string.

services/QuillLMS/app/models/auth_credential.rb
```diff
@@ -24,4 +24,18 @@
...
+  end
+
+  def refresh_token_valid?
+    return false if expires_at.nil? || refresh_token.nil?
+
+    Time.now < refresh_token_expires_at
+  end
+
+  def refresh_token_expires_at
+    expires_at + 6.months
```
dandrabik: `6.months` is used in 2 logical places, let's make it a constant and reference that.

services/QuillLMS/app/models/auth_credential.rb
```diff
@@ -24,4 +24,18 @@
...
   belongs_to :user
+
+  def google_authorized?
+    provider == 'google' && refresh_token_valid?
+  end
+
+  def refresh_token_valid?
+    return false if expires_at.nil? || refresh_token.nil?
+
+    Time.now < refresh_token_expires_at
```
dandrabik, line 38: This logic is written elsewhere slightly differently (same logic). It might be better to make both the same
```ruby
Time.now - 6.months >= current_credentials.expires_at

File: services/QuillLMS/spec/controllers/auth/google_controller_spec.rb [hunk 13]
```diff
@@ -3,7 +3,11 @@
 describe Auth::GoogleController, type: :controller do
 
   it 'shows error message for nonexistent accounts with no role' do
+<<<<<<< HEAD
```
dandrabik: merge leftover.

dandrabik (approved):
Overall, this looks good. Thanks for taking this on. I left some small items to address, but none of them are major.

💯

services/QuillLMS/app/controllers/auth/google_controller.rb
brendanshean: When a user arrives on the `authorization_and_authentication` route, they have already been authorized (i.e. they have a valid refresh_token). When arriving via the `authentication` route, we don't actually know if their refresh_token exists or is valid, hence the callback only used for this route.

Maybe renaming it `verify_refresh_token` would be more clear as that's what the point of this check is ultimately for.

services/QuillLMS/app/controllers/auth/google_controller.rb
brendanshean: I agree.  I think all the branching begs for this to be broken out into two parts.  However, I was reluctant to do that given the lack of test coverage.  I think once we get some integration testing in place, I would feel more comfortable cleaning this up further.

services/QuillLMS/app/controllers/auth/google_controller.rb
brendanshean: Actually, I'm going to call it `verify_authorization_for_offline_access` since that's the purpose of that check. We need authorization in order to perform the background jobs.

services/QuillLMS/app/models/auth_credential.rb
brendanshean: Good call.  I will update

services/QuillLMS/app/controllers/auth/google_controller.rb
dandrabik: Thanks for the explanation and I like the rename,

dandrabik (approved):
:+1: LGTM 🚢

PR #7852, opened 2021-06-28
Featureflags/clientside

happythenewsad:
## WHAT
enables the fetching and usage of AppSettings on the client. 

### Screenshots
Screen Shot 2021-07-02 at 11 54 22 AM

### Notion Card Links
https://www.notion.so/quill/hydrate-client-with-app-settings-175b50b0bb674de8a10a399464c65123

PR Checklist | Your Answer
------------ | -------------


services/QuillLMS/client/app/bundles/Staff/components/comprehension/activities.tsx
```diff
@@ -1,19 +1,33 @@
...
+    queryKey: [appSettingQueryKey, appSettingName],
+    queryFn: fetchAppSetting
+  });
+
+
+  let appSettingValue = false;
+  if (appSettingData && appSettingData.appSetting && appSettingData.appSetting[appSettingName]) {
+    appSettingValue = appSettingData.appSetting[appSettingName]
+  } 
+  console.log("appSettingValue: ", appSettingValue)
```
happythenewsad: This file's changes are not required for the AppSetting infrastructure. They illustrate what an integration might look like in a generic React component. I believe we can't abstract this boilerplate into a class due to the use of the UseQuery hook, but I'm open to suggestions. (It would be nice to have a one-liner for an AppSetting check, if that's possible).

services/QuillLMS/client/app/bundles/Shared/utils/appSettingAPIs.ts
```diff
@@ -0,0 +1,11 @@
+import { mainApiFetch, handleApiError } from './../../Staff/helpers/comprehension';
+
+export const fetchAppSetting = async (key: string, appSettingName: string) => {
+  const response = await mainApiFetch(`app_settings/${appSettingName}`);
+
+  const appSetting = await response.json();
+  return {
+    appSetting,
+    error: handleApiError('Failed to fetch activities, please refresh the page.', appSetting)
```
eadams17: so this is actually an outdated error handling setup. Can you model the error handling like it's done here?
- [error helper functions](https://github.com/empirical-org/Empirical-Core/blob/develop/services/QuillLMS/client/app/bundles/Staff/helpers/comprehension.tsx#L337): `requestFailed` & `handleRequestErrors`
- [API util function](https://github.com/empirical-org/Empirical-Core/blob/develop/services/QuillLMS/client/app/bundles/Staff/utils/comprehension/ruleAPIs.ts#L69)

services/QuillLMS/client/app/bundles/Staff/components/comprehension/activities.tsx
eadams17: We'll probably end up pulling it out into a custom hook like this:

```
const [appSetting, setAppSetting] = React.useState<string>(null);

handleAppSettingUpdate({ appSetting, appSettingName, setAppSetting, appSettingData });
```

and in the custom hook file:
``` 
function handleAppSettingUpdate({ appSetting, appSettingName, setAppSetting, appSettingData }) {
  return React.useEffect(() => {
    if (!appSetting && appSettingData && appSettingData.appSetting && appSettingData.appSetting[appSettingName]) {
      const appSettingValue = appSettingData.appSetting[appSettingName];
      setAppSetting(appSettingValue);
    } 
  }, [appSettingData]);
}
```

eadams17 (changes requested):
Looks good! Just requested a change about how the error handling is done

anathomical (approved):
This seems reasonable to me.  Eric's suggestion seems reasonable, but I'll leave that up to you two to hash out.

services/QuillLMS/client/app/bundles/Staff/components/comprehension/activities.tsx
dandrabik: So, as you mention, it would be really great if we could abstract more of this into the parent class so that we could could just have `hasFeatureX` be set boilerplate code.. It might be worth working with @eadams17 or @emilia-friedberg a bit to think of a way to do this.

```javascript
// Similar to Eric's but returning a boolean and abstracting more
const [hasAppSetting, setHasAppSetting] = React.useState<boolean>(false);
handleHasAppSetting({hasAppSetting, setHasAppSetting, 'feature_key'})

Comment (happythenewsad) [reply services/QuillLMS/client/app/bundles/Staff/components/comprehension/activities.tsx]: Thanks for the idea - the integration API is down to 2 lines now.

**Comment by happythenewsad** [comment]
I'd like to check in the example integration. It only applies to admins on a specific page, so it's not going to affect performance.

File: services/QuillLMS/client/app/bundles/Shared/utils/appSettingAPIs.ts [hunk 3]
```diff
@@ -0,0 +1,22 @@
...
+  if(requestFailed(status)) {
+    const errors = await response.json();
+    const returnedErrors = await handleRequestErrors(errors);
+    return { errors: returnedErrors };
+  }
+
+  const appSetting = await response.json();
+
+  if (appSetting && appSetting[key]) {
+    setHasAppSetting(true)
```
happythenewsad, line 17: just seeing this bug - modifying in a sec

happythenewsad:
This is currently demoable here: https://staging2.quill.org/cms/comprehension#/activities

services/QuillLMS/client/app/bundles/Staff/components/comprehension/activities.tsx
```diff
@@ -2,18 +2,22 @@ import * as React from "react";
...
 import { ActivityInterface } from '../../interfaces/comprehensionInterfaces';
 import { fetchActivities } from '../../utils/comprehension/activityAPIs';
 import { DataTable, Error, Spinner } from '../../../Shared/index';
+import { handleHasAppSetting } from "../../../Shared/utils/appSettingAPIs";
 
 const Activities = ({ location, match }) => {
 
   // cache activity data for updates
-  const { data } = useQuery("activities", fetchActivities);
+  const { data: activitiesData } = useQuery("activities", fetchActivities);
```
happythenewsad, line 16: This change is no longer necessary after I removed the useQuery call for appSettings, but let's just leave it as it's more precise than what was previously there.

dandrabik (approved):
I'd like one of the frontend developers to approve this before it goes out, but overall, this looks good to me.

services/QuillLMS/client/app/bundles/Shared/utils/appSettingAPIs.ts
```diff
@@ -0,0 +1,22 @@
+import { mainApiFetch, requestFailed, handleRequestErrors } from './../../Staff/helpers/comprehension';
+
+export const handleHasAppSetting = async (hasAppSetting, setHasAppSetting, key: string) => {
```
eadams17: You'll want to be passing in a `setErrors` function here from the `Activities` component, otherwise these errors are just getting swallowed and won't be displayed anywhere if they exist

eadams17 (changes requested):

services/QuillLMS/client/app/bundles/Shared/utils/appSettingAPIs.ts
```diff
@@ -0,0 +1,28 @@
...
+  key: string,
+}
+
+export const handleHasAppSetting = async (args: HandleHasAppSettingArgs) => {
+  const response = await mainApiFetch(`app_settings/${args.key}`);
+
+  const { status } = response;
+
+  if(requestFailed(status)) {
+    const returnedErrors = [`AppSetting HTTP status: ${status}`]
```
happythenewsad, line 15: Updated this logic so that we don't attempt to parse JSON when the status is FAIL (this will lead to a parsing error)

services/QuillLMS/client/app/bundles/Staff/components/comprehension/activities.tsx
```diff
@@ -47,6 +52,8 @@ const Activities = ({ location, match }) => {
 
   return(<React.Fragment>
     <Navigation location={location} match={match} />
+    {hasAppSetting && <div>App setting is enabled.</div>}
+    {errors && renderErrorsContainer(false, errors)}
```
happythenewsad, line 56: This is not necessary for the demo, and there will be times when we want an app setting to fail silently (as long as the client API returns a value of false), but Eric suggested explicit error message displaying for this component.

eadams17 (approved):

PR #7857, opened 2021-06-29
Make Comprehension Regex Rules work for more than one required sequence

cissyxyu:
## WHAT
Regex Rules with more than one required sequence are currently failing because the back-end tries to ensure that the student entry matches **all** the required sequences. Actually, we should only ensure that the entry matches one or more of the required sequences.

## WHY
Required sequence criteria for passing differs from Incorrect Sequence criteria for passing.

## HOW
Add branching logic to apply the right criteria for required sequences. Use the array method `any?` instead of the array method `none?` to match for true on any required sequence if the Rule contains required sequences.

### Screenshots

### Notion Card Links
None, reported bug in #comprehension-bugs Slack

PR Checklist | Your Answer
------------ | -------------

### DIFF
--- a/services/QuillLMS/engines/comprehension/app/models/comprehension/rule.rb
+++ b/services/QuillLMS/engines/comprehension/app/models/comprehension/rule.rb
@@ -67,8 +67,15 @@ module Comprehension
     end
 
     def regex_is_passing?(entry)
-      regex_rules.none? do |regex_rule|
-        regex_rule.entry_failing?(entry)
+      return true if regex_rules.empty?
+      if regex_rules.first.is_incorrect_sequence?
+        regex_rules.none? do |regex_rule|
+          regex_rule.entry_failing?(entry)
+        end
+      else
+        regex_rules.any? do |regex_rule|
+          !regex_rule.entry_failing?(entry)
+        end
       end
     end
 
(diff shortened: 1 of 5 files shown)


services/QuillLMS/engines/comprehension/app/models/comprehension/rule.rb
```diff
@@ -67,8 +67,15 @@ def determine_feedback_from_history(feedback_history)
     end
 
     def regex_is_passing?(entry)
-      regex_rules.none? do |regex_rule|
-        regex_rule.entry_failing?(entry)
+      return true if regex_rules.empty?
+      if regex_rules.first.incorrect_sequence?
+        regex_rules.none? do |regex_rule|
```
happythenewsad: `any?` and `none?` relate inversely, so you only need to iterate over `regex_rules` once, then memoize the result and evaluate it in both branches of the conditional.

happythenewsad (changes requested):

services/QuillLMS/engines/comprehension/app/models/comprehension/rule.rb
cissyxyu: Pushed an update based on our conversation! (Instead of memoizing the result of one loop, kept both loops but defined them under more clear and reasonable names).

happythenewsad (approved):

PR #7859, opened 2021-06-29
Validate focus points before saving

cissyxyu:
## WHAT
We sometimes have malformed focus points and incorrect sequences in our `Question` records because we save these as JSON objects with no field validations. This PR adds in some validations so we don't error out on bad data.

## WHY
We sometimes see bugs where questions can't be played because of bad focus point or incorrect sequence data.

## HOW
Add validation behavior to validate that each focus point and incorrect sequence:
1. has a text field
2. has a feedback field
3. has valid regex saved in the text field

Add in controller behavior that catches when a focus point or incorrect sequence is invalid and sends back an error response.

### Screenshots

### Notion Card Links
https://www.notion.so/quill/validate-focus-points-e93590d213774a7db997839fa5cbcd83

PR Checklist | Your Answer
------------ | -------------

### DIFF
--- a/services/QuillLMS/app/controllers/api/v1/focus_points_controller.rb
+++ b/services/QuillLMS/app/controllers/api/v1/focus_points_controller.rb
@@ -16,12 +16,15 @@ class Api::V1::FocusPointsController < Api::ApiController
 
   def update
     return not_found unless @question.data.dig("focusPoints", params[:id])
-    @question.set_focus_point(params[:id], valid_params)
-    render_focus_point
+    if @question.set_focus_point(params[:id], valid_params)
+      render_focus_point
+    else
+      render json: @question.errors, status: 422
+    end
   end
 
   def update_all
-    @question.update_focus_points(valid_params)
+    @question.update_focus_points(params_as_hash)
     render_all_focus_points
   end
 
@@ -34,8 +37,13 @@ class Api::V1::FocusPointsController < Api::ApiController
     @question = Question.find_by!(uid: params[:question_id])
   end
 
+  private def params_as_hash
+    valid_params.is_a?(Array) ? valid_params.map {|x| x.to_h } : valid_params.to_h
+  end
+
   private def valid_params
-    params.require(:focus_point).except(:uid)
+    filtered_params = params.require(:focus_point)
+    filtered_params.is_a?(Array) ? filtered_params.map {|x| x.except(:uid).permit! } : filtered_params.except(:uid).permit!
   end
 
   private def render_focus_point
--- a/services/QuillLMS/app/controllers/api/v1/focus_points_controller.rb
+++ b/services/QuillLMS/app/controllers/api/v1/focus_points_controller.rb
@@ -24,7 +24,7 @@ class Api::V1::FocusPointsController < Api::ApiController
   end
 
   def update_all
-    @question.update_focus_points(params_as_hash)
+    @question.update_focus_points(valid_params)
     render_all_focus_points
   end
 
@@ -37,13 +37,13 @@ class Api::V1::FocusPointsController < Api::ApiController
     @question = Question.find_by!(uid: params[:question_id])
   end
 
-  private def params_as_hash
-    valid_params.is_a?(Array) ? valid_params.map {|x| x.to_h } : valid_params.to_h
-  end
-
   private def valid_params
     filtered_params = params.require(:focus_point)
-    filtered_params.is_a?(Array) ? filtered_params.map {|x| x.except(:uid).permit! } : filtered_params.except(:uid).permit!
+    if filtered_params.is_a?(Array)
+      filtered_params.map {|x| x.except(:uid).permit!.to_h }
+    else
+      filtered_params.except(:uid).permit!.to_h
+    end
   end
 
   private def render_focus_point
--- a/services/QuillLMS/app/models/question.rb
+++ b/services/QuillLMS/app/models/question.rb
@@ -2,17 +2,17 @@
 #
 # Table name: questions
 #
-#  id            :integer          not null, primary key
-#  data          :jsonb            not null
-#  question_type :string           not null
-#  uid           :string           not null
-#  created_at    :datetime         not null
-#  updated_at    :datetime         not null
+#  id            :integer
+#  data          :jsonb
+#  question_type :string
+#  uid           :string
+#  created_at    :datetime
+#  updated_at    :datetime
+#  temp_id       :integer          not null, primary key
 #
 # Indexes
 #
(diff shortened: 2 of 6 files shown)


services/QuillLMS/app/models/question.rb
```diff
@@ -59,14 +60,18 @@ def as_json(options=nil)
   end
 
   def add_focus_point(new_data)
-    set_focus_point(new_uuid, new_data)
+    new_uid = new_uuid
+    if set_focus_point(new_uid, new_data)
```
brendanshean: What about: `return new_id if set_focus_point(new_uid, new_data)`

services/QuillLMS/spec/controllers/api/v1/focus_points_controller_spec.rb
```diff
@@ -3,6 +3,8 @@
 
 describe Api::V1::FocusPointsController, type: :controller do
   let!(:question) { create(:question) }
+  let!(:new_q) {create(:question, data: {"focusPoints"=>{"0"=>{"text"=>"text","feedback"=>"fff"}},"incorrectSequences"=>[{"text"=>"foo","feedback"=>"bar"}]}) }
+
```
brendanshean, line 18: I would use a multi-line declaration for this so it's easier to parse.
```
  let!(:new_q) do
    create(:question, 
      data: { 
        "focusPoints" => {
          "0" => { "text" => "text", "feedback"=>"fff" }
        },
        "incorrectSequences"=> [
          { "text"=>"foo", "feedback"=>"bar" }
        ]
      }
    ) 
  end
```

services/QuillLMS/app/models/question.rb
```diff
@@ -101,15 +106,15 @@ def add_incorrect_sequence(new_data)
     else
       new_id = new_uuid
     end
-    set_incorrect_sequence(new_id, new_data)
+    return new_id if set_incorrect_sequence(new_id, new_data)
+    nil
```
brendanshean: I don't think the `nil` here is necessary.

services/QuillLMS/app/models/question.rb
```diff
@@ -149,4 +154,37 @@ def rematch_type
   private def stored_as_array(key)
     data[key].class == Array
   end
+
+  private def validate_sequences
+    return if !data.present? || !data.is_a?(Hash)
```
brendanshean: I would use `data.blank?` over `!data.present?`  It's easier to reason through conditionals in the positive than negated ones.

services/QuillLMS/app/models/question.rb
```diff
@@ -149,4 +154,37 @@ def rematch_type
...
+    parse_and_validate(data['incorrectSequences'])
+    parse_and_validate(data['focusPoints'])
+  end
+
+  private def parse_and_validate(sequences)
+    return if !sequences.present?
```
brendanshean: `sequences.blank?`.  Also, it's good to add a line after a line [break](https://www.rubydoc.info/gems/rubocop/RuboCop/Cop/Layout/EmptyLineAfterGuardClause)) after guard clauses to signal control flow

services/QuillLMS/app/models/question.rb
```diff
@@ -149,4 +154,37 @@ def rematch_type
...
+    if sequences.is_a?(Hash)
+      sequences.each do |key, value|
```
brendanshean: A one-liner block is fine when you only have a short statement:  
`sequences.each { |key, value| validate_text_and_feedback(value) }`

services/QuillLMS/app/models/question.rb
```diff
@@ -149,4 +154,37 @@ def rematch_type
...
+        validate_text_and_feedback(value)
+      end
+    elsif sequences.is_a?(Array)
+      sequences.each do |value|
```
brendanshean: Single line block:  `sequences.each { |value| validate_text_and_feedback(value) }`

services/QuillLMS/app/models/question.rb
```diff
@@ -149,4 +154,37 @@ def rematch_type
...
+        validate_text_and_feedback(value)
+      end
+    end
+  end
+
+  private def validate_text_and_feedback(value)
+    errors.add(:data, "Focus Points and Incorrect Sequences must have text and feedback.") unless value['text'].present? && value['feedback'].present?
```
brendanshean: [Opt](https://www.rubydoc.info/gems/rubocop-rails/RuboCop/Cop/Rails/Blank) for `blank?` over `!present?`.  Single-line block here again.
```
  if value['text'].blank? || value['feedback'].blank?
    errors.add(:data, "Focus Points and Incorrect Sequences must have text and feedback.")
    return
  end

  value['text'].split('|||').each { |regex| validate_regex(regex) }
```

services/QuillLMS/app/controllers/api/v1/focus_points_controller.rb
```diff
@@ -34,8 +37,13 @@ def destroy
...
   end
 
+  private def params_as_hash
+    valid_params.is_a?(Array) ? valid_params.map {|x| x.to_h } : valid_params.to_h
+  end
+
   private def valid_params
-    params.require(:focus_point).except(:uid)
+    filtered_params = params.require(:focus_point)
+    filtered_params.is_a?(Array) ? filtered_params.map {|x| x.except(:uid).permit! } : filtered_params.except(:uid).permit!
```
brendanshean: For consistency across methods, I would remove the method: `params_as_hash` and do the hash conversion in `valid_params`. 

```
private def valid_params
    filtered_params = params.require(:focus_point)

    if filtered_params.is_a?(Array)
      filtered_params.map { |x| x.except(:uid).permit!.to_h }
    else
      filtered_params.except(:uid).permit!.to_h
    end    
end
```

services/QuillLMS/app/models/question.rb
```diff
@@ -48,6 +48,7 @@ class Question < ActiveRecord::Base
   validates :question_type, presence: true, inclusion: {in: TYPES}
   validates :uid, presence: true, uniqueness: true
   validate :data_must_be_hash
+  validate :validate_sequences
```
brendanshean, line 52: In going through this commit I noticed a related method in question.rb  `def stored_as_array`.  I think it would be worth renaming to `stored_as_array?`.

brendanshean (approved):
Some refactoring suggestions but overall code is good.  Nice job on working through the Rails 5 params stuff.  

FYI  This is a good [post](https://eileencodes.com/posts/actioncontroller-parameters-now-returns-an-object-instead-of-a-hash/) for more explanation on what the params change entails.

happythenewsad (approved):

PR #7861, opened 2021-06-29
Opinion bolding

happythenewsad:
## WHAT
- highlights multiple instances of a given word
- highlights all of a given word if the given word is a substring of a a word in the entry. Example: "should" should match "shouldn't" as well as "should".
-
### Screenshots

### Notion Card Links
https://www.notion.so/quill/Product-Board-30f97e4eb01246dbb8264253fb385073?p=e0c9347454674d3eb698b236ebe672cf

PR Checklist | Your Answer
------------ | -------------

### DIFF
--- a/services/QuillLMS/client/setEnvVars.js
+++ b/services/QuillLMS/client/setEnvVars.js
@@ -1,1 +1,1 @@
-process.env.GOLANG_FANOUT_URL = 'https://us-central1-comprehension-247816.cloudfunctions.net/comprehension-fanout-staging';
+process.env.GOLANG_FANOUT_URL = 'https://staging.comprehension-feedback.quill.org/';
--- a/services/QuillLMS/client/app/bundles/Comprehension/components/studentView/promptStep.tsx
+++ b/services/QuillLMS/client/app/bundles/Comprehension/components/studentView/promptStep.tsx
@@ -6,6 +6,7 @@ import Feedback from './feedback'
 import EditCaretPositioning from '../../helpers/EditCaretPositioning'
 import ButtonLoadingSpinner from '../shared/buttonLoadingSpinner'
 import preFilters from '../../modules/prefilters'
+import { highlightSpellingGrammar } from '../../libs/stringFormatting'
 
 interface PromptStepProps {
   active: Boolean;
@@ -105,7 +106,7 @@ export class PromptStep extends React.Component<PromptStepProps, PromptStepState
     if (lastSubmittedResponse.feedback_type == 'plagiarism') {
       return this.formatPlagiarismHighlight(str, wordsToFormat)
     } else {
-      return this.formatSpellingGrammarHighlight(str, wordsToFormat)
+      return highlightSpellingGrammar(str, wordsToFormat)
     }
   }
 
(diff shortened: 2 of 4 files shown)


services/QuillLMS/client/setEnvVars.js
```diff
@@ -1 +1 @@
-process.env.GOLANG_FANOUT_URL = 'https://us-central1-comprehension-247816.cloudfunctions.net/comprehension-fanout-staging';
+process.env.GOLANG_FANOUT_URL = 'https://staging.comprehension-feedback.quill.org/';
```
happythenewsad, line 1: This change is not directly related to the feature, but is required in order to run the feature locally and should be in place for everyone.

services/QuillLMS/client/app/bundles/Shared/utils/stringFormatting.ts
```diff
@@ -0,0 +1,15 @@
+
+class StringFormatting {
```
happythenewsad: I'd like to move toward expressing static utility functions as class methods. It allows use to put functions in logical groups, and also adds namespacing and signals intent.

services/QuillLMS/client/app/bundles/Shared/utils/stringFormatting.test.ts
```diff
@@ -0,0 +1,21 @@
+import StringFormatting from "./stringFormatting"
+
+
+describe('stringFormatting', () => {
```
brendanshean: For completeness, should there be a test for `wordsToFormat: string[]`?

services/QuillLMS/client/app/bundles/Shared/utils/stringFormatting.ts
brendanshean: I like the idea of this extraction.  Just a thought, but what about going a step further and renaming the function name the function name to  `StringFormatting.highlightSpellingGrammar(...)`  since the class is already expressing that this is formatting?

brendanshean (approved):
Minor comments, but looks good to me.

services/QuillLMS/client/app/bundles/Shared/utils/stringFormatting.ts
emilia-friedberg: I'm not strictly against it in this context, but I do think that in general, we are moving towards a functional rather than OO approach in our Javascript and it probably doesn't make sense to backtrack on that here. I think file structure can group things logically without the use of classes.

services/QuillLMS/client/app/bundles/Shared/utils/stringFormatting.ts
```diff
@@ -0,0 +1,15 @@
...
+  static highlightSpellingGrammar = (str: string, wordsToFormat: string | string[]) => {
+    let wordArray = [].concat(wordsToFormat)
+    let newString = str
+    wordArray.forEach((word) => {
+      var matched = newString.match(`${word}[\\w\']*`)
+      var augmentedRegex = new RegExp(`${matched[0]}`, 'g')
```
emilia-friedberg: Use `const` instead of `var` here: https://airbnb.io/javascript/#references--prefer-const

services/QuillLMS/client/app/bundles/Shared/utils/stringFormatting.ts
emilia-friedberg: Also, I think that since this function is used in such a specific capacity, it's better to keep it closer in the file structure to where it is actually employed (in only one place, right now) than all the way out in `Shared/utils`. Curious if @eadams17 has thoughts about either of these points.

emilia-friedberg (changes requested):
Small change requested around a style guide issue, two more points for your consideration.

services/QuillLMS/client/app/bundles/Shared/utils/stringFormatting.ts
eadams17: I agree with @emilia-friedberg. In general, React is moving away from the use of classes and is encouraging functional development with its newer features so I think we want to continue on that course. I also agree on the second point; there is already a `/libs` directory within the Comprehension app where I think makes more sense to house this since it's specific to Comprehension

emilia-friedberg (approved):
LGTM!

PR #7865, opened 2021-07-01
Add rake task for updating user email domain

brendanshean:
## WHAT
Add a rake task that allows updating the domain for particular email addresses.

## WHY
A school district is changing its domain so we need to update the corresponding email addresses in our database to main consistency with clever integration.

## HOW
Add a rake task that locates the desired domain and update it with the new one. 

### Screenshots
Screen Shot 2021-07-01 at 10 06 37 AM

### Notion Card Links
https://www.notion.so/quill/Update-districts-email-domain-for-all-users-70ac57ce7631451890be383f2ad4a031

PR Checklist | Your Answer
------------ | -------------
Have you added and/or updated tests? |  Manual check on staging; see screenshots

### DIFF
--- a/services/QuillLMS/lib/tasks/temporary/users.rake
+++ b/services/QuillLMS/lib/tasks/temporary/users.rake
@@ -1,13 +1,16 @@
 namespace :users do
   task :update_email_domain, [:old_domain, :new_domain] => [:environment] do |task, args|
-    users = User.where("email ILIKE ?", "%@#{args[:old_domain]}")
+    old_domain = "@#{args[:old_domain]}"
+    new_domain = "@#{args[:new_domain]}"
+
+    users = User.where("email ILIKE ?", "%#{old_domain}")
 
     puts "Going to update #{users.count} email addresses"
     num_updates = 0
 
     ActiveRecord::Base.transaction do
       users.find_each do |user|
-        user.update(email: user.email.sub(args[:old_domain], args[:new_domain]))
+        user.update(email: user.email.sub(old_domain, new_domain))
         num_updates += 1
         print '.'
       end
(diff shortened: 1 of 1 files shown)


services/QuillLMS/lib/tasks/temporary/users.rake
```diff
@@ -0,0 +1,18 @@
+namespace :users do
+  task :update_email_domain, [:old_domain, :new_domain] => [:environment] do |task, args|
+    users = User.where("email ILIKE ?", "%@#{args[:old_domain]}")
+
+    puts "Going to update #{users.count} email addresses"
+    num_updates = 0
+
+    ActiveRecord::Base.transaction do
+      users.find_each do |user|
+        user.update(email: user.email.sub(args[:old_domain], args[:new_domain]))
```
anathomical: This seems unlikely to be necessary, but is it worth protecting against accidental modification of the user side of the email address in weird cases where it overlaps with the domain?  Like, what if you have `penny.eduperson@ny.edu`?

Maybe use `sub("@#{args[:old_domain]}", "@#{args[:new_domain]}")`?

Also, do we need to worry about casing conflict in the `sub`?

anathomical (approved):
A couple of questions around the specifics of the `sub` call, but the general approach here seems sound.

services/QuillLMS/lib/tasks/temporary/users.rake
brendanshean: That's a great point on the '@'.  I will update.

I'm not sure I understand the question on the casing conflict.

anathomical: Is it possible for `sub` to fail to replace `Thomas.Robertson@Quill.org` if `old_domain` is `quill.org` because there's a case mismatch between the on-record email address and the specified arguments for the rake task?

brendanshean: That's another good point.  While, I think all persisted user emails are downcased, the rake task wouldn't work if `new_domain` were capitalized.  I'll just downcase both for clarity.

PR #7866, opened 2021-07-01
Add rake task for clearing data from deleted users

brendanshean:
## WHAT
Add a rake task that calls `clear_data` on all deleted users.

## WHY
Over time the `clear_data` method on `User` has been updated resulting in inconsistencies in not-cleared-out-data between older and newer deleted users.

## HOW
First run a query to find deleted users, then call `clear_data` on each of them.

### Notion Card Links
https://www.notion.so/quill/Run-Clear-data-on-all-deleted-users-221b227c3b284fc0b3eda6f95f70c160

PR Checklist | Your Answer
------------ | -------------

### DIFF
--- a/services/QuillLMS/app/models/user.rb
+++ b/services/QuillLMS/app/models/user.rb
@@ -164,6 +164,16 @@ class User < ActiveRecord::Base
 
   attr_accessor :newsletter
 
+  def self.deleted_users
+    where(
+      <<-SQL
+        name LIKE 'Deleted_User_%'
+          AND email LIKE 'deleted_user_%'
+          AND username LIKE 'deleted_user_%'
+      SQL
+    )
+  end
+
   def testing_flag
     role == STAFF ? ARCHIVED : flags.detect{|f| TESTING_FLAGS.include?(f)}
   end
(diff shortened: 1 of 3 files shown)


services/QuillLMS/app/models/user.rb
```diff
@@ -164,6 +164,16 @@ class User < ActiveRecord::Base
...
 
+  def self.deleted_users
+    where(
+      <<-SQL
+        name LIKE 'Deleted_User_%'
+          AND email LIKE 'deleted_user_%'
+          AND username LIKE 'deleted_user_%'
+      SQL
+    )
+  end
```
anathomical, line 175: This has got to be a pretty expensive query given multiple `like`s.  Not suggesting a change here, but I am wondering if we should be thinking about a better solution in the future.

anathomical (approved):

services/QuillLMS/app/models/user.rb
brendanshean: I didn't notice any performance issues while running the rake task on staging.  My understanding is that since the query is a conjunction on the same table, the result set gets smaller and smaller after each `like`, therefore we're not doing a full scan of the table each time.

That being said, I think it would be a good idea to redesign the deleted user representation in the database as hard-coded deleting_user on to attribute values is definitely a code smell.

PR #7868, opened 2021-07-01
Lms/tweak comprehension reporting

anathomical:
## WHAT
This is a bunch of semi-related tweaks around the way Comprehension activities get reported to teachers in the LMS.
## WHY
There are a number of expectations that the reporting structure makes (primarily around implied meanings for `percentage = nil` and hard-coded lists of which activities don't count as "scored") which didn't account for how we wanted to handle Comprehension.  This primarily consists of the tweaks necessary to make Comprehension display as desired in a number of different scenarios.
## HOW
- Tweaks to back-end code to use `ActivitySession.state = 'finished'` as an alternative to `ActivityState.percentage != nil` when determining if an `ActivitySession` should be reported (Comprehension activities explicitly set their `percentage` to `nil`).
- Attach the `ActivityClassification.key` value to the Question view in reporting so that we can do special display stuff when `ActivityClassification.key = 'comprehension'`
- Tweak the logic in `ActivitySession.determine_if_final_score` so that new Comprehension play-throughs are always treated as "final".  The logic for other activity types only replaces a previous play-through if the new one has a better score than the previous "final" one.
- Added an explicit `finished` flag to Activity lists rather than relying on a proxy of `max_score` since Comprehension should always have a score of `nil`, but we still want to communicate that the activity is finished.  Update front-ends to use this new attribute.
- Tweak the way Comprehension composes `ConceptResults` on completion to explicitly pull the instructions shown to the student as the initial `instructions` value, allowing us to guarantee that the reports show what the students actually saw.
- Use `ActivitySession.state = 'finished'` to calculate the number of `completed_attempts` shown in reports instead of counting `ActivitySession.percentage IS NOT NULL` since Comprehension expects completed `ActivitySession`s to have `percentage = NULL`
- Update front-ends to rely on `completed_attempts` being greater than 0 rather than `max_percentage` being greater than 0 when determining if an activity has been completed at least once.
- Consolidate a number of places where the front-end checks to see if `ActivityClassification.id` is equal to hard-coded `4` or had-coded `6` into a single importable array of `ActivityClassification.id` values to treat as non-graded, and add `9`, the ID for Comprehension, to this array.
- Write a custom tool-tip explaining that Comprehension isn't graded to the hover-state for Comprehension activity icons
- Tweak a number of reporting displays to show a score of "N/A" for Comprehension, either by checking the `ActivityClassification.key` directly, or via checking of the `ActivitySession` is finished and has a score of `null`.

### Notion Card Links
https://www.notion.so/quill/QA-Remove-Comprehension-Scores-6d3131b8d22347b0aaba848499e5f5bf#2a40d8b80701485f9f324122ba851b21

PR Checklist | Your Answer
------------ | -------------


services/QuillLMS/app/models/concerns/public_progress_reports.rb
```diff
@@ -54,14 +55,14 @@ def results_by_question(params)
         curr_quest[:question_number] ||= answer["question_number"]
         if answer["attemptNumber"] == 1 || !curr_quest[:instructions]
           direct = answer["directions"] || answer["instructions"] || ""
-          curr_quest[:instructions] ||= direct.gsub(/(<([^>]+)>)/i, "").gsub("()", "").gsub("&nbsp;", "")
+          curr_quest[:instructions] = direct.gsub(/(<([^>]+)>)/i, "").gsub("()", "").gsub("&nbsp;", "")
```
emilia-friedberg, line 58: Why are we changing from `||=` to just `=` here?

services/QuillLMS/app/models/concerns/public_progress_reports.rb
```diff
@@ -165,6 +166,8 @@ def formatted_score_obj(final_activity_session, activity, student)
       activity_classification_key = ActivityClassification.find(activity.activity_classification_id).key
       if ['lessons', 'diagnostic'].include?(activity_classification_key)
         score = get_average_score(formatted_concept_results)
+      elsif ['comprehension'].include?(activity_classification_key)
```
emilia-friedberg: I know we aren't doing it with the rest of these, but this should probably be `[ActivityClassification::COMPREHENSION_KEY].include?(activity_classification_key)`

services/QuillLMS/app/models/concerns/student.rb
```diff
@@ -34,6 +34,7 @@ def student_average_score
             ON activity_sessions.activity_id = activities.id
           WHERE activities.activity_classification_id != 6
             AND activities.activity_classification_id != 4
+            AND activities.activity_classification_id != 9
```
emilia-friedberg: Is this a good time to refactor these to use the keys instead of ids?

services/QuillLMS/app/models/unit_activity.rb
```diff
@@ -124,6 +124,7 @@ def self.get_classroom_user_profile(classroom_id, user_id)
           COALESCE(cuas.locked, false) AS locked,
           COALESCE(cuas.pinned, false) AS pinned,
           MAX(acts.percentage) AS max_percentage,
+          SUM(CASE WHEN acts.state = 'finished' THEN 1 ELSE 0 END) > 0 AS finished,
```
emilia-friedberg: Not urgent but would probably be good to use a constant here (we may already have one?)

services/QuillLMS/app/queries/scorebook/query.rb
```diff
@@ -19,7 +19,7 @@ def self.run(classroom_id, current_page=1, unit_id=nil, begin_date=nil, end_date
           MAX(acts.updated_at) AS updated_at,
           MIN(acts.started_at) AS started_at,
           MAX(acts.percentage) AS percentage,
-          SUM(CASE WHEN acts.percentage IS NOT NULL THEN 1 ELSE 0 END) AS completed_attempts,
+          SUM(CASE WHEN acts.state = 'finished' THEN 1 ELSE 0 END) AS completed_attempts,
```
emilia-friedberg: Another place a constant might be useful.

services/QuillLMS/client/app/bundles/Teacher/components/progress_reports/diagnostic_reports/question_report.jsx
```diff
@@ -10,13 +10,14 @@ export default class QuestionReport extends React.Component {
...
 
-  columnDefinitions = () => {
+  columnDefinitions = (activityClassification) => {
     return [
       {
         name: 'Score',
         field: 'score',
         sortByField: 'score',
         customCell: function(row) {
+          if (activityClassification === 'comprehension') return 'N/A'
```
emilia-friedberg: Inclined to say it probably makes sense to put `'comprehension'` in a constant in a shared file here too (maybe in the `progress_reports` subfolder?) since it is getting used for business logic throughout these files.

emilia-friedberg (approved):
LGTM overall, just some comments about constantization you can take or leave.

services/QuillLMS/app/models/concerns/public_progress_reports.rb
anathomical: Good question.  So what's happening here is that `all_answers`, which is what we're iterating over here, doesn't seem to come back in a sorted order.  So you may get the third attempt first, then the second, then the first.  In the old logic, this resulted in the `:instructions` key being set by attempt 3 which, as far as I can tell, is supposed to be an emergency fallback for cases where we don't have an `attemptNumber == 1` condition.  When this happened, the `instructions` value was set for by a later attempt, which means that it was being set to the value of some attempt's feedback.

This change should result in, at most, one extra set of this value because of the condition on line 56 which will only try to set `:instructions` if it is not set OR this is the first attempt that we're looking at.  So the logic around `||=` is kind of already in the conditional.

anathomical:
@emilia-friedberg Generally speaking, I agree with all your comments on constants and moving away from ID hard coding, so I'll make those changes.

PR #7870, opened 2021-07-01
Comprehension internal tools/ add second layer feedback

eadams17:
## WHAT
add second layer feedback to the Rules Analysis page

## WHY
it was never being supplied to the frontend and Curriculum needs to be able to see the value

## HOW
just add a second property `second_feedback` to the backend payload and add it to the  `RulesAnalysis` table

### Screenshots

### Notion Card Links
https://www.notion.so/quill/Rules-Analysis-Fix-rule-note-feedback-bug-and-add-second-layer-feedback-to-toggle-62f1ee623349422989f5ea4981edfa29

PR Checklist | Your Answer
------------ | -------------

### DIFF
--- a/services/QuillLMS/app/lib/rule_feedback_history.rb
+++ b/services/QuillLMS/app/lib/rule_feedback_history.rb
@@ -68,7 +68,7 @@ class RuleFeedbackHistory
                 api_name: r.rule_type,
                 rule_order: r.rule_suborder,
                 first_feedback: r.feedbacks.order(:order).first&.text || '',
-                second_feedback: r.feedbacks.order(:order).last&.text || '',
+                second_feedback: r.feedbacks.order(:order).second&.text || '',
                 rule_note: r.rule_note,
                 rule_name: r.rule_name,
                 total_responses: r.total_responses,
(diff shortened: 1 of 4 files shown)


services/QuillLMS/app/lib/rule_feedback_history.rb
```diff
@@ -68,6 +68,7 @@ def self.format_sql_results(relations)
                 api_name: r.rule_type,
                 rule_order: r.rule_suborder,
                 first_feedback: r.feedbacks.order(:order).first&.text || '',
+                second_feedback: r.feedbacks.order(:order).last&.text || '',
```
brendanshean: What about `r.feedbacks.order(:order).second&.text || ''`, otherwise if `r.feedbacks` has only one element, `first` and `last` will return the same element.

brendanshean (changes requested):
One very minor request, otherwise, looks good.

services/QuillLMS/app/lib/rule_feedback_history.rb
anathomical: I like `second` here, too.  Also, while no one has ever done it, I'm pretty sure someone could create a "third" layer, and using `last` would pull that one.

anathomical (approved):
Agree with Brendan that we should use `second` rather than `last`, since Ruby provides that as a keyword.  Otherwise, this looks good to me.

services/QuillLMS/app/lib/rule_feedback_history.rb
eadams17: ahh ok I didn't know you could use `second`, thanks for pointing that out @brendanshean !

brendanshean (approved):

PR #7874, opened 2021-07-02
CMS - Update New Relic, ignore certain errors

dandrabik:
## WHAT
New Relic has a new version of their gem in pre-release, with which, we can ignore errors from our error rate and alerts. Integrating this and ignoring some errors, with some app changes to make that more explicit.
## WHY
Cut down on spurious alerts. We want our alerts and error rate to indicate site issues. Some errors, like retryable ones in background jobs or an incorrect route, are better to remove from these calculations to keep our alerts accurate.
## HOW
Upgrade gem. They released 7.2.0 officially.
~~Upgrade to pre-release `7.2.0`. Note, I had to vendor this to make deploying work(they haven't pushed this to RubyGems yet) and am still testing this (If I run into issues, we can just wait for them to publish).~~

I made some errors more specific so that we can feel confident ignoring them, e.g. `RematchResponseWorker::LambdaError`

PR Checklist | Your Answer
------------ | -------------
Have you added and/or updated tests? |  Mostly no, since this is a monitoring change.
Have you deployed to Staging? | YES, still testing.

### DIFF
--- a/services/QuillCMS/app/workers/rematch_response_worker.rb
+++ b/services/QuillCMS/app/workers/rematch_response_worker.rb
@@ -6,7 +6,7 @@ class RematchResponseWorker
 
   sidekiq_options retry: 1, queue: SidekiqQueue::DEFAULT
 
-  # retry 10 seconds later to spread out attempts
+  # spread out retries
   sidekiq_retry_in { 5 }
 
   DEFAULT_PARAMS_HASH = {
--- a/services/QuillCMS/app/workers/rematch_response_worker.rb
+++ b/services/QuillCMS/app/workers/rematch_response_worker.rb
@@ -19,7 +19,7 @@ class RematchResponseWorker
     'spelling_error' => false
   }.freeze
 
-  class LambdaError < StandardError; end
+  class LambdaHTTPError < StandardError; end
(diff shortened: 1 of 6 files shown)


anathomical (approved):

anathomical:
Excited about this change.  Hopefully there are no issues and we can just run this in prod to get our error reporting under control.

services/QuillCMS/app/workers/rematch_response_worker.rb
```diff
@@ -16,6 +19,8 @@ class RematchResponseWorker
     'spelling_error' => false
   }.freeze
 
+  class LambdaError < StandardError; end
```
happythenewsad: How about `Not200Error`? `Lambda` is a bit cryptic.

happythenewsad (approved):

services/QuillCMS/app/workers/rematch_response_worker.rb
dandrabik: Good point. I renamed this to `LamdaHTTPError` since that is more clear (and better matches the `Net::HTTPError` it replaces).

PR #7882, opened 2021-07-07
Fix time tracking bug involving strong parameters

brendanshean:
## WHAT
Fix a bug involving time tracking data not being saved on activity types

## WHY
We'd like to store time tracking information for various reasons.

## HOW
When storing the data value, we have to call `permit!` on this hash:
`data = params.delete(:data).permit!`

Otherwise, when calling `@activity_session.update(activity_session_params)` the value for `data` is not permitted:
Screen Shot 2021-07-07 at 7 33 56 AM

### Screenshots

### Notion Card Links
https://www.notion.so/quill/Timetracking-data-not-saving-for-any-activity-types-except-Lessons-af281f1cfdb34deb900ff7f7ba3132c5

PR Checklist | Your Answer
------------ | -------------

### DIFF
--- a/services/QuillLMS/app/controllers/api/v1/activity_sessions_controller.rb
+++ b/services/QuillLMS/app/controllers/api/v1/activity_sessions_controller.rb
@@ -38,7 +38,6 @@ class Api::V1::ActivitySessionsController < Api::ApiController
   def create
     @activity_session = ActivitySession.new(activity_session_params)
     @activity_session.user = current_user if current_user
-    @activity_session.concept_results = []
 
     if @activity_session.save
       handle_concept_results if @concept_results
--- a/services/QuillLMS/app/controllers/api/v1/activity_sessions_controller.rb
+++ b/services/QuillLMS/app/controllers/api/v1/activity_sessions_controller.rb
@@ -37,18 +37,13 @@ class Api::V1::ActivitySessionsController < Api::ApiController
 
   def create
     @activity_session = ActivitySession.new(activity_session_params)
-    crs = @activity_session.concept_results
     @activity_session.user = current_user if current_user
     @activity_session.concept_results = []
-    # activity_session.owner=(current_user) if activity_session.ownable?
-    # activity_session.data = @data # FIXME: may no longer be necessary?
-    if @activity_session.save
-      if @activity_session.update(activity_session_params)
-        handle_concept_results if @concept_results
 
-        @status = :success
-        @message = "Activity Session Created"
-      end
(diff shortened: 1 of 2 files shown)


services/QuillLMS/app/controllers/api/v1/activity_sessions_controller.rb
```diff
@@ -85,22 +105,15 @@ def destroy
...
-                  :classroom_unit_id,
-                  :activity_uid,
-                  :activity_id,
-                  :anonymous,
-                  :temporary
-                )
-      .merge(data: @data).reject {|k,v| v.nil? }
-      .merge(timespent: @activity_session&.timespent || ActivitySession.calculate_timespent(@time_tracking))
+    data = params.delete(:data)&.permit!
+    time_tracking = data && data['time_tracking']
```
emilia-friedberg, line 103: Why don't we want the `||=` here?

emilia-friedberg (approved):
Thank you for getting to this so quickly!

services/QuillLMS/app/controllers/api/v1/activity_sessions_controller.rb
brendanshean: Unless I'm missing something, `||=` isn't necessary since `data` and `time_tracking` shouldn't have values before reaching 108-109.

brendanshean: Actually, you're right.  I was only looking at the `update` action but it's called twice in `create`.

emilia-friedberg: The solution you have here is fine with me, but I guess it still seems to me like we might as well use `||=` for the sake of idempotency. Is there a reason to avoid it?

emilia-friedberg (approved):

services/QuillLMS/app/controllers/api/v1/activity_sessions_controller.rb
brendanshean: While the `||=` was definitely needed earlier when the `create` method had a path where `activity_session_params` was called twice, it is only called [once](https://github.com/empirical-org/Empirical-Core/pull/7882/commits/06db1c2a45f16ca87a5b5b2a7f59c6c352e392a2#diff-fb0cb613d7c02517cbea8eab235eb3f3effa4c0340ce0d8929139e8afdb231b7L46) now, and I think that leaving the `||=` would obscure future reading of the code.

PR #7887, opened 2021-07-07
LMS - Better Redirects for Blog post url errors

dandrabik:
## WHAT
We get a good amount of errors that originate from bad links to our blog posts or areas within the Teacher/Student Centers.
This PR attempts to eliminate those errors and give the users a better experience by:
1. Auto-correcting  links that have extra characters (e.g. ending parentheses or punctuation from bad markdown, `/blog-slug)`, `/blog-slug.`,`/blog-slug)!`)
2. For uncorrectable urls, redirect to the homepage of `Teacher Center` or `Student Center` rather than a generic error page.
## WHY
1. To eliminate error reports. 
2. To give users with a slightly-off link a better chance of finding what they are looking for. Landing on a generic error page with no direction where to go isn't a great experience. On the Teacher Center, they can browse and search for what they were looking for.
## HOW
I was trying to be as light touch as possible, but I ended up refactoring a bunch of the view and controller (there's honestly more to refactor, but I accomplished my goal, and decided to step away - I debated separating the student and teacher logic into separate controllers, since there is still some weirdness on these pages).

### Notion Card Links
[Sample error in NR](https://one.newrelic.com/launcher/nr1-core.explorer?platform[timeRange][duration]=86400000&pane=eyJwYWdlIjoic2hvdyIsInRyYWNlSWQiOiIyNjExMjdhMi1kZTliLTExZWItODk3NS0wMjQyYWMxMTAwMDlfMF82NjgyIiwibmVyZGxldElkIjoiZXJyb3JzLXVpLm92ZXJ2aWV3IiwiZW50aXR5R3VpZCI6Ik1qWXpPVEV4TTN4QlVFMThRVkJRVEVsRFFWUkpUMDU4TlRRNE9EVTJPRGMxIn0=&sidebars[0]=eyJuZXJkbGV0SWQiOiJucjEtY29yZS5hY3Rpb25zIiwiZW50aXR5R3VpZCI6Ik1qWXpPVEV4TTN4QlVFMThRVkJRVEVsRFFWUkpUMDU4TlRRNE9EVTJPRGMxIiwic2VsZWN0ZWROZXJkbGV0Ijp7Im5lcmRsZXRJZCI6ImVycm9ycy11aS5vdmVydmlldyJ9fQ==&state=f180344b-1b37-0580-2d72-debbbb619f55)

PR Checklist | Your Answer
------------ | -------------

### DIFF
--- a/services/QuillLMS/app/controllers/blog_posts_controller.rb
+++ b/services/QuillLMS/app/controllers/blog_posts_controller.rb
@@ -6,6 +6,7 @@ class BlogPostsController < ApplicationController
   def index
     topic_names = BlogPost::TEACHER_TOPICS
     @topics = []
+    @topic = BlogPost::ALL_RESOURCES
     topic_names.each do |name|
       @topics.push({ name: name, slug: CGI::escape(name.downcase.gsub(' ','-'))})
     end
@@ -29,7 +30,7 @@ class BlogPostsController < ApplicationController
     draft_status = (@role == 'staff') ? [true, false] : false
 
     @blog_post = BlogPost.find_by!(slug: params[:slug], draft: draft_status)
-    # TODO remove SQL write from get endpoint
+    # TODO remove SQL write from GET endpoint
     @blog_post.increment_read_count
     @most_recent_posts = BlogPost.most_recent.where.not(id: @blog_post.id)
 
@@ -84,13 +85,14 @@ class BlogPostsController < ApplicationController
         redirect_to "/teacher-center/topic/#{new_topic}"
       end
     else
-      topic = CGI::unescape(params[:topic]).gsub('-', ' ').capitalize
-      if !BlogPost::TOPICS.include?(topic) && !BlogPost::STUDENT_TOPICS.include?(topic)
+      @topic = CGI::unescape(params[:topic]).gsub('-', ' ').capitalize
+      if !BlogPost::TOPICS.include?(@topic) && !BlogPost::STUDENT_TOPICS.include?(@topic)
         raise ActionController::RoutingError, 'Topic Not Found'
       end
-      @blog_posts = BlogPost.for_topics(topic)
+      @blog_posts = BlogPost.for_topics(@topic)
       # hide student part of topic name for display
-      @title = @role == 'student' ? topic.gsub('Student ', '').capitalize : topic
+      @title = @role == 'student' ? topic.gsub('Student ', '').capitalize : @topic
+
       render 'index'
     end
   end
--- a/services/QuillLMS/app/controllers/blog_posts_controller.rb
+++ b/services/QuillLMS/app/controllers/blog_posts_controller.rb
@@ -1,12 +1,16 @@
 class BlogPostsController < ApplicationController
-  before_action :set_announcement, only: [:index, :show, :show_topic]
   before_action :set_role
+  before_action :redirect_legacy_topic_urls, only: [:show_topic]
+  before_action :redirect_invalid_topics, only: [:show_topic]
+
+  before_action :set_announcement, only: [:index, :show, :show_topic]
   before_action :set_root_url
+
+
 
   def index
     topic_names = BlogPost::TEACHER_TOPICS
     @topics = []
-    @topic = BlogPost::ALL_RESOURCES
     topic_names.each do |name|
       @topics.push({ name: name, slug: CGI::escape(name.downcase.gsub(' ','-'))})
     end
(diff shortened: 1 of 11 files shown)


services/QuillLMS/app/controllers/blog_posts_controller.rb
```diff
@@ -19,22 +24,35 @@ def student_center_index
...
-    @blog_post = BlogPost.find_by!(find_by_hash)
-    @topic = @blog_post.topic
-    @display_paywall = true unless @blog_post.can_be_accessed_by(current_user)
-    @blog_post.increment_read_count
-    @author = @blog_post.author
-    @most_recent_posts = BlogPost.where("draft = false AND id != #{@blog_post.id}").order('updated_at DESC').limit(3)
-    @title = @blog_post.title
-    @description = @blog_post.subtitle || @title
-    @image_link = @blog_post.image_link
+    draft_statuses = @role == 'staff' ? [true, false] : false
```
emilia-friedberg: I know this wasn't in the older version of this method, but it probably makes sense to use `User::STAFF` here.

services/QuillLMS/app/controllers/blog_posts_controller.rb
```diff
@@ -61,24 +79,14 @@ def search
...
-      @blog_posts = BlogPost.where(draft: false, topic: topic).order('order_number')
-      # hide student part of topic name for display
-      @title = @role == 'student' ? topic.gsub('Student ', '').capitalize : topic
-      render 'index'
-    end
+    topic = CGI::unescape(params[:topic]).gsub('-', ' ').capitalize
+
+    @blog_posts = BlogPost.for_topics(topic)
+    # hide student part of topic name for display
+    @topic = @role == 'student' ? topic.gsub('Student ', '').capitalize : topic
```
emilia-friedberg: `User::STUDENT`

services/QuillLMS/app/controllers/blog_posts_controller.rb
```diff
@@ -92,4 +100,25 @@ def show_topic
...
     @root_url = root_url
   end
+
+  private def redirect_invalid_topics
+    topic = CGI::unescape(params[:topic]).gsub('-', ' ').capitalize
+    return if BlogPost::TOPICS.include?(topic)
+    return if BlogPost::STUDENT_TOPICS.include?(topic)
+
+    flash[:error] = "Oops! We can't seem to find that topic!"
+    center_home_url = @role == 'student' ? '/student-center' : '/teacher-center'
```
emilia-friedberg: `User::STUDENT`

services/QuillLMS/app/controllers/blog_posts_controller.rb
```diff
@@ -92,4 +100,25 @@ def show_topic
...
+
+    redirect_to center_home_url and return
+  end
+
+  # handling links that were possibly broken by changing slug function for topic names
+  private def redirect_legacy_topic_urls
+    return unless params[:topic].include?('_')
+
+    new_topic = params[:topic].gsub('_', '-')
+    base_url = @role == 'student' ? 'student-center' : 'teacher-center'
```
emilia-friedberg: `User::STUDENT`

services/QuillLMS/app/views/pages/shared/_teacher_center_navbar.html.erb
```diff
@@ -1,75 +1,23 @@
-<% @all = 'All' %>
-<% @research = 'Research' %>
-<% @faq = 'FAQ' %>
-<% @premium = 'Premium' %>
+<%-# locals: active_tab %>
+<%-# TODO - this feels like a bug, but everything is labeled mobile = false %>
+<% @mobile = false %>
```
emilia-friedberg: I don't think this is getting used at all anymore. It looks like `generic_tabs` takes @mobile as an argument, so I think it was previously getting passed in to be explicit that the mobile-only part wasn't supposed to show. I think just not passing it accomplishes the same thing, but then this line is unnecessary.

services/QuillLMS/app/views/pages/shared/_teacher_center_navbar.html.erb
```diff
@@ -1,75 +1,23 @@
-<% @all = 'All' %>
-<% @research = 'Research' %>
-<% @faq = 'FAQ' %>
-<% @premium = 'Premium' %>
+<%-# locals: active_tab %>
```
emilia-friedberg: Why is this here?

emilia-friedberg (approved):
Looks good overall. Some comments around constants that you can take or leave.

services/QuillLMS/app/views/pages/shared/_teacher_center_navbar.html.erb
dandrabik: Yeah, that's fair. This is equivalent code to what was there before, i.e. `@mobile = false` is set for all pages in this view. I'll remove this, since I don't see any code in the repo where `@mobile = true` is ever set (I was hesitant to remove in case the global var was set somewhere else, e.g. a layout file, meta code, etc. That's the problem with global settings).

services/QuillLMS/app/views/pages/shared/_teacher_center_navbar.html.erb
dandrabik: It's a comment about local vars for this partial. I'm going to reword the comment to make that clearer. It's trying to treat the partial like a function and list its params.

services/QuillLMS/app/controllers/blog_posts_controller.rb
dandrabik: Yeah, good call. I'll refactor this to be better.

PR #7890, opened 2021-07-08
allow curriculum team members to reorder regexes within a layer

emilia-friedberg:
## WHAT
Allow curriculum team members to reorder regex rules within a given layer.

## WHY
So that they don't have to delete and recreate the rules when they want to reorder them.

## HOW
Update the frontend to allow drag-and-drop reordering, and add a backend route to connect to that actually saves the new order.

### Screenshots
Screen Shot 2021-07-08 at 10 53 40 AM

### Notion Card Links
https://www.notion.so/quill/Allow-curriculum-team-to-change-the-order-of-the-regex-rules-7d39bbd4c4a943878e6e0f1fd160a416

PR Checklist | Your Answer
------------ | -------------


services/QuillLMS/client/app/bundles/Staff/components/comprehension/regexRules/regexRulesIndex.tsx
```diff
@@ -49,6 +49,15 @@ const RegexRulesIndex: React.FC<RouteComponentProps<ActivityRouteProps>> = ({ ma
     queryFn: fetchRules
   });
 
+  function handleReorder(sortInfo) {
+    const idsInOrder = sortInfo.map(item => item.key)
+    updateRuleOrders(idsInOrder).then((response) => {
+      queryCache.refetchQueries([`rules-${activityId}-${RULES_BASED_1}`])
+      queryCache.refetchQueries([`rules-${activityId}-${RULES_BASED_2}`])
+      queryCache.refetchQueries([`rules-${activityId}-${RULES_BASED_3}`])
```
eadams17, line 57: I'm not sure if this will work anymore? Either way, you can just put `queryCache.clear()` here

eadams17 (approved):
Looks great! Just a comment about the query cache refetch

services/QuillLMS/client/app/bundles/Staff/components/comprehension/regexRules/regexRulesIndex.tsx
emilia-friedberg: It did, I tested it! Won't `queryCache.clear()` wipe the whole cache?

services/QuillLMS/engines/comprehension/app/controllers/comprehension/rules_controller.rb
```diff
@@ -45,6 +45,23 @@ def destroy
       head :no_content
     end
 
+    def update_rule_order
+      ordered_rules = ordered_rules_params[:ordered_rule_ids].map.with_index do |id, index|
+        rule = Comprehension::Rule.find_by_id(id)
```
brendanshean, line 50: _**If**_ you wanted to include errors for the nil rules you could utilize the [Null object pattern](https://thoughtbot.com/blog/rails-refactoring-example-introduce-null-object).

```
module Comprehension
  class NullRule
    attr_reader :errors
    attr_accessor :suborder

    def initialize(errors: [])
      @errors = errors
    end

    def valid?
      false
    end
  end
end
```

Which would allow for the pretzel colon shorthand:

```
def update_rule_order
  ordered_rules = ordered_rules_params[:ordered_rule_ids].map.with_index do |id, index|
    rule = Comprehension::Rule.find_by_id(id) || Comprehension::NullRule.new(errors: "rule #{id} was not found")
    rule.suborder = index
    rule
  end

  if ordered_rules.all?(&:valid?)
    ordered_rules.each(&:save!)
    render json: { status: 200 }
  else
    render json: { error_messages: ordered_rules.map(&:errors).join('; ') }, status: :unprocessable_entity
  end
end
```

services/QuillLMS/engines/comprehension/app/controllers/comprehension/rules_controller.rb
```diff
@@ -45,6 +45,23 @@ def destroy
...
+      end
+
+      if ordered_rules.all? { |r| r&.valid? }
+        ordered_rules.each { |r| r.save! }
+      else
+        render json: {error_messages: ordered_rules.map { |r| r&.errors }.join('; ')}, status: :unprocessable_entity
+        return
+      end
+
+      render(json: {status: 200})
```
brendanshean: What about moving the render on line 62 inside the conditional after line 56 and then remove `return` on line 59?

brendanshean (approved):
Just a general comment on error reporting that you can take or leave and then a minor nit.

Otherwise looks good.  Very cool functionality.

services/QuillLMS/client/app/bundles/Staff/components/comprehension/regexRules/regexRulesIndex.tsx
eadams17: oh, awesome! I wonder why it stopped working in some places... Yes, it'll clear the whole cache.  When we eventually update the package, I think the docs suggest that that is the preferred way to update since `refetchQueries` is already phased out in newer versions

PR #7905, opened 2021-07-12
fixing failing spec

happythenewsad:
## WHAT
Fixes intermittent test failure 

## HOW
I postulate that the deletion of elements from the array - which was unnecessary for the test - caused a race condition.

### Screenshots

### Notion Card Links
https://www.notion.so/quill/Product-Board-30f97e4eb01246dbb8264253fb385073?p=01a7908c44714ab4b4e62c9a0b331dbd

PR Checklist | Your Answer
------------ | -------------

### DIFF
--- a/services/QuillLMS/spec/models/concerns/teacher_spec.rb
+++ b/services/QuillLMS/spec/models/concerns/teacher_spec.rb
@@ -542,13 +542,23 @@ describe User, type: :model do
     end
   end
 
-  describe '#number_of_assigned_students_per_activity_assigned' do
+  describe '#assigned_students_per_activity_assigned' do
     context 'assigned students' do
       let!(:teacher) { create(:teacher, :with_classrooms_students_and_activities) }
       let!(:assigned_units) { teacher.assigned_students_per_activity_assigned }
 
-      it 'should return the correct students assigned' do
-        assigned_units_arr = assigned_units.to_a
+      it 'should return only the correct students assigned' do
+        student_ids = assigned_units.pluck(:assigned_student_ids).each_with_object([]) {|n, memo| memo.concat(n)}
+
+        other_teacher = create(:teacher, :with_classrooms_students_and_activities) 
+        other_assigned_activities_ids = other_teacher.assigned_students_per_activity_assigned.pluck(:id)
+        other_student_ids = other_teacher.assigned_students_per_activity_assigned
+          .pluck(:assigned_student_ids)
+          .each_with_object([]) {|n, memo| memo.concat(n)}
+
+        expect(other_assigned_activities_ids & assigned_units.pluck(:id)).to be_empty
+        expect(other_student_ids & student_ids).to be_empty
+
         teacher.classroom_units.each do |cu|
           cu.unit.unit_activities.each do |ua|
             matching_element = assigned_units.detect {|u| u.id == ua.activity_id && u.assigned_student_ids == cu.assigned_student_ids}
--- a/services/QuillLMS/spec/models/concerns/teacher_spec.rb
+++ b/services/QuillLMS/spec/models/concerns/teacher_spec.rb
@@ -548,13 +548,13 @@ describe User, type: :model do
       let!(:assigned_units) { teacher.assigned_students_per_activity_assigned }
 
       it 'should return only the correct students assigned' do
-        student_ids = assigned_units.pluck(:assigned_student_ids).each_with_object([]) {|n, memo| memo.concat(n)}
+        student_ids = assigned_units.pluck(:assigned_student_ids).flatten
 
         other_teacher = create(:teacher, :with_classrooms_students_and_activities) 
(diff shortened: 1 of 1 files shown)


services/QuillLMS/spec/models/concerns/teacher_spec.rb
```diff
@@ -547,19 +547,14 @@
...
+        teacher.classroom_units.each do |cu|
+          cu.unit.unit_activities.each do |ua|
             matching_element = assigned_units.detect {|u| u.id == ua.activity_id && u.assigned_student_ids == cu.assigned_student_ids}
-            expect(matching_element).to_not be_nil
             expect(matching_element.assigned_student_ids).to eq(cu.assigned_student_ids)
-            expect(matching_element.created_at).to eq(ua.created_at)
-            assigned_units_arr.delete_at(assigned_units_arr.index(matching_element))
-          }
-        }
-        expect(assigned_units_arr).to be_empty
```
happythenewsad, line 562: This expectation was testing the spec itself, not the intended underlying method

brendanshean: This [line](https://github.com/empirical-org/Empirical-Core/blob/f44da4c9b8f0a6e49946366ec2269a6bd0fcf3ff/services/QuillLMS/spec/models/concerns/teacher_spec.rb#L545) should be renamed `describe '#assigned_students_per_activity_assigned'`

services/QuillLMS/spec/models/concerns/teacher_spec.rb
```diff
@@ -547,19 +547,14 @@
       let!(:teacher) { create(:teacher, :with_classrooms_students_and_activities) }
       let!(:assigned_units) { teacher.assigned_students_per_activity_assigned }
 
-      it 'should return the correct number of students assigned' do
-
+      it 'should return the correct students assigned' do
         assigned_units_arr = assigned_units.to_a
```
brendanshean: Probably don't need this line anymore

services/QuillLMS/spec/models/concerns/teacher_spec.rb
brendanshean: While I agree there might be a race condition as you described, I think we still need something related to the `delete_at` and subsequent `be_empty` test since that ensures assigned_units doesn't contain any other values outside of those in the loop.

In other words,  if we had something like `assigned_units += [some new unit]` this test should still pass.  What about just adding a counter?

```
matching_elements_count = 0

teacher.classroom_units.each do |cu|
  cu.unit.unit_activities.each do |ua|
    matching_element = assigned_units.detect {|u| u.id == ua.activity_id && u.assigned_student_ids == cu.assigned_student_ids}
    expect(matching_element.assigned_student_ids).to eq(cu.assigned_student_ids)
    matching_elements_count += 1
  end
end

expect(matching_elements_count).to eq assigned_units.count
```

brendanshean (changes requested):

services/QuillLMS/spec/models/concerns/teacher_spec.rb
happythenewsad: Ok, I incorporated the spirit of your feedback, with a different implementation

services/QuillLMS/spec/models/concerns/teacher_spec.rb
```diff
@@ -542,24 +542,29 @@
...
 
-  describe '#number_of_assigned_students_per_activity_assigned' do
+  describe '#assigned_students_per_activity_assigned' do
     context 'assigned students' do
       let!(:teacher) { create(:teacher, :with_classrooms_students_and_activities) }
       let!(:assigned_units) { teacher.assigned_students_per_activity_assigned }
 
-      it 'should return the correct number of students assigned' do
+      it 'should return only the correct students assigned' do
+        student_ids = assigned_units.pluck(:assigned_student_ids).each_with_object([]) {|n, memo| memo.concat(n)}
```
brendanshean: Would `assigned_units.pluck(:assigned_student_ids).flatten` work here?

services/QuillLMS/spec/models/concerns/teacher_spec.rb
```diff
@@ -542,24 +542,29 @@
...
 
-        assigned_units_arr = assigned_units.to_a
-        teacher.classroom_units.each { |cu|
-          cu.unit.unit_activities.each { |ua|
+        other_teacher = create(:teacher, :with_classrooms_students_and_activities) 
+        other_assigned_activities_ids = other_teacher.assigned_students_per_activity_assigned.pluck(:id)
+        other_student_ids = other_teacher.assigned_students_per_activity_assigned
+          .pluck(:assigned_student_ids)
+          .each_with_object([]) {|n, memo| memo.concat(n)}
```
brendanshean: what about `.flatten`?

brendanshean (approved):
Much better and more clear as to what we should be checking.  

One minor nit for readability.

anathomical (approved):

PR #7907, opened 2021-07-12
add comprehension to activity library and around the site

emilia-friedberg:
## WHAT
Add a Comprehension activity classification filter to the activity library, dependent on whether or not a user has that `AppSetting`, and also add icons for it around the site where we're currently missing them. We will also need to run `rake 'app_settings:create[comprehension,true,true,0,[]]'` when we deploy this in order to create that app setting.

## WHY
So that we can start rolling out Comprehension activities to teachers in a more integrated way.

## HOW
Mostly just updating frontend code, though in some cases the backend had to be adjusted as well. I tried, where it was pretty straightforward, to update places where we determine classification by id to use key instead.

### Screenshots
Screen Shot 2021-07-09 at 1 49 42 PM
Screen Shot 2021-07-09 at 1 27 25 PM
Screen Shot 2021-07-09 at 1 22 35 PM
Screen Shot 2021-07-09 at 1 20 23 PM
Screen Shot 2021-07-09 at 1 16 23 PM

### Notion Card Links
https://www.notion.so/quill/Add-Comprehension-to-Activity-Library-Page-e44de3ba6df04972a02e01382911347e

PR Checklist | Your Answer
------------ | -------------
Design Review: If applicable, have you compared the coded design to the mockups? | YES


eadams17 (approved):

cissyxyu (approved):

services/QuillLMS/client/app/bundles/Teacher/components/assignment_flow/create_unit/custom_activity_pack/index.tsx
```diff
@@ -114,7 +117,7 @@ const CustomActivityPack = ({
     number += savedActivityFilters.length ? 1 : 0
     number += flagFilters.length
 
-    activityClassificationGroupings.forEach((g) => {
+    activityClassificationGroupings(true).forEach((g) => {
```
anathomical, line 120: I'm assuming that since this is used for calculating a property, it really always should be `true` rather than being dependent on the state of `showComrpehension`.

anathomical (approved):
One point that I wanted to double-check, but this is otherwise really good.  I like the shift toward activity classification keys over IDs, smart to do that now while you're in there rather than wait.

PR #7916, opened 2021-07-13
Comprehension/close go channel

anathomical:
## WHAT
Try explicitly closing the Go `channel` object when we finish with it so that goroutines get freed up
## WHY
We think that failing to explicitly close the channel may be contributing to the non-terminating goroutines that are causing the production system to crash out
## HOW
We're explicitly closing the channel object, and modifying the code that sends messages to the channel to make sure it's not closed before we send (since sending things to closed channels crashes the code).

### Notion Card Links
https://www.notion.so/quill/Stop-the-Comprehension-Go-Endpoint-in-Heroku-from-running-so-many-Goroutines-and-runs-out-of-memory-b5bb5d3a07614e27ad488929f88edbb3

PR Checklist | Your Answer
------------ | -------------
Have you added and/or updated tests? |  No behavior changes here

### DIFF
--- a/services/comprehension/feedback-api-main/src/endpoint/endpoint.go
+++ b/services/comprehension/feedback-api-main/src/endpoint/endpoint.go
@@ -172,6 +172,16 @@ func processResults(results map[int]InternalAPIResponse, length int) (int, bool)
 	return automl_index, all_correct
 }
 
+func safeSendToChannel(c chan InternalAPIResponse, response InternalAPIResponse) {
+	select {
+	case <-c:
+		return
+	default:
+	}
+
+	c <- response
+}
+
 func getAPIResponse(url string, priority int, json_params [] byte, c chan InternalAPIResponse) {
 	// response_json, err := http.Post(url, "application/json", bytes.NewReader(json_params))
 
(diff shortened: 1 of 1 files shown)


services/comprehension/feedback-api-main/src/endpoint/endpoint.go
```diff
@@ -171,6 +172,16 @@ func processResults(results map[int]InternalAPIResponse, length int) (int, bool)
 	return automl_index, all_correct
 }
 
+func safeSendToChannel(c chan InternalAPIResponse, response InternalAPIResponse) {
```
dandrabik, line 175: So based on what I've read, we can't use this approach because the channel could close right after this `select` check, then you'd get a panic for sending to a closed channel. You can only close a channel if you are assured that you won't send anything else to it, e.g. a synchronous loop that adds things to a channel, then a close.

> Note: Only the sender should close a channel, never the receiver. Sending on a closed channel will cause a panic.
https://tour.golang.org/concurrency/4

I'm trying some other approaches.

dandrabik (changes requested):
Marking this a 'request changes' since this we can't do this approach. But I'm currently working on some other potential approaches.

anathomical:
Also it looks like this solution doesn't do anything anyway, so I'm just going to close and discard this branch.

PR #7917, opened 2021-07-13
Change Log for Comprehension (back-end, attempt #2)

cissyxyu:
## WHAT
Implementing Change Log logging behavior for 30 or so different possible actions on Comprehension records. 

## WHY
Curriculum wants to see a record of all changes made to an activity and related records.

## HOW
Essentially just putting lots of callbacks on `save` and `update` actions on these records. The callback then creates a `ChangeLog` record that records that action as well as relevent data (using [ActiveModel::Dirty](https://api.rubyonrails.org/classes/ActiveModel/Dirty.html)).

I also created an endpoint `activities/change_logs.json` which fetches all of the change log records for an activity (all change logs associated with all its nested records as well), and appends some additional data before returning a list of all change logs.

This is an attempt #2 following an [earlier attempt here](https://github.com/empirical-org/Empirical-Core/pull/7779) which was scrapped because my approach was too complicated.

### Screenshots

### Notion Card Links
https://www.notion.so/quill/Create-an-activity-change-log-2f7563fbef0d42be84ad733676689e08

PR Checklist | Your Answer
------------ | -------------


services/QuillLMS/engines/comprehension/app/models/comprehension/activity.rb
```diff
@@ -48,8 +67,26 @@ def serializable_hash(options = nil)
...
+    def url
+      "comprehension/#/activities/#{id}/settings"
+    end
+
     private def expire_turking_rounds
       turking_rounds.each(&:expire!)
     end
+
+    private def log_creation
+      return unless @lms_user_id.present?
```
anathomical: The guard statements here don't actually work, I don't think.  If, somehow, `@lms_user_id` is not set, and thus is `nil`, you'll get an error because you can't call `.present?` on `nil`.  Probably want a nil-value guard here: `unless @lms_user_id&.present?`

Alternatively, you could remove the early returns and just log with no user if it's somehow missing.

anathomical (approved):
One note on some possibly not-working guard statements, but this looks like about as clean a solution as we're going to get to a very messy spec.  I'm still pretty uncomfortable with marking changes to `Feedback` models as changes to `Activities` and such, but I know that's what's being asked for here.  Nicely done.

services/QuillLMS/engines/comprehension/lib/comprehension/comprehension/change_log.rb
```diff
@@ -0,0 +1,17 @@
+module Comprehension
+  module ChangeLog
```
happythenewsad, line 15: I suspect you can move `has_many :change_logs` and the `after_save` hook into this module, via an `included?` callback. This greatly reduces the touchpoints for AR models that include this module.

Docs: https://ruby-doc.org/core-2.6.3/Module.html#method-i-included

This would require standardizing the change log behavior (that is, all AR models would have the same AR life cycle hooks for changelogs, which is currently not the case in this PR). This may break the spec, but it's worth discussing since it'll save time and bugs in the future.

happythenewsad: (we zoomed to discuss) Cissy plans to discuss with Lindsey, the specification writer, on whether we can change the spec to make the AR callbacks more standardized across classes, thus reducing the complexity of each integration (an integration is an AR module that includes the ChangeLog module)

services/QuillLMS/engines/comprehension/app/controllers/comprehension/application_controller.rb
```diff
@@ -18,5 +18,9 @@ def render(options, *args)
       options.merge(json: json.to_json)
     end
 
+    private def lms_user_id
```
dandrabik, line 23: So, I think we'd want to modify the Engine to gain access to current_user (so we're not duplicating logic). I'm fine with leaving that as an explicit `TODO` here. Here's a reference link on one way to set that up (I think we'd want to do this in a separate PR, to be assigned later): https://www.hocnest.com/blog/configure-rails-engine/

services/QuillLMS/engines/comprehension/app/models/comprehension/activity.rb
```diff
@@ -48,8 +67,26 @@ def serializable_hash(options = nil)
       ))
     end
 
+    def change_logs
+      change_logs_for_activity(self)
+    end
+
+    def url
```
dandrabik, line 57: We shouldn't store urls in the database, since that can be a bit fragile (if we change url paths, we'd need a data migration also). 

Instead, if we need to display the url to the admin user, the url should be put together when we return the `json` for the admin panel view.

services/QuillLMS/engines/comprehension/app/models/comprehension/activity.rb
```diff
@@ -48,8 +67,26 @@ def serializable_hash(options = nil)
       ))
     end
 
+    def change_logs
```
dandrabik: So this method would overwrite `has_many :change_logs`. I think we'd want to avoid that collision.

services/QuillLMS/engines/comprehension/app/models/comprehension/activity.rb
```diff
@@ -48,8 +67,26 @@ def serializable_hash(options = nil)
...
       turking_rounds.each(&:expire!)
     end
+
+    private def log_creation
+      return unless @lms_user_id.present?
+      log_change(@lms_user_id, :create_activity, self, {url: url}.to_json, nil, nil, nil)
+    end
+
+    private def log_deletion
+      return unless @lms_user_id.present?
```
dandrabik: Wouldn't we want to log the deletion/update/creation even if a user_id wasn't preset?

services/QuillLMS/engines/comprehension/app/controllers/comprehension/automl_models_controller.rb
```diff
@@ -41,7 +41,7 @@ def update
 
     # PATCH/PUT /automl_models/1.json
     def activate
-      if @automl_model.activate
+      if @automl_model.activate_with_session_user(lms_user_id)
```
dandrabik: So, since you created an `attr_accessor` on the model, it might be cleaner to assign the lms_user in the controller, and then you can use the regular `save` and `activate` methods. Also, since it's an `attr_accessor`, that means you can use it like a regular attribute, e.g. non-instance variable

So in the controller:
```ruby
@automl_model.lms_user_id = params[:lms_user_id]
@automl_model.save
# or
before_action :set_lms_user, only: [:create, :update]
... 
```
In the model:
```ruby
def some_method
  return if lms_user_id.nil?
  ...
end

File: services/QuillLMS/engines/comprehension/app/models/comprehension/feedback.rb [hunk 8]
```diff
@@ -20,5 +25,35 @@ def serializable_hash(options = nil)
...
+
+    private def first_order
+      order == 0
+    end
+
+    private def second_order
+      order == 1
+    end
+
+    private def log_update
```
dandrabik: So I don't think we need to calculate this on storage, but rather on display to admins. It should be sufficient to store that feedback changed in the db. Then when pulling the change_log, pull the necessary information (type of rule, order) to show the correct info.

services/QuillLMS/engines/comprehension/app/models/comprehension/passage.rb
```diff
@@ -13,5 +18,11 @@ def serializable_hash(options = nil)
         only: [:id, :text, :image_link, :image_alt_text]
       ))
     end
+
+    private def log_update
+      if text_changed?
```
dandrabik: Could be a `return unless text_changed?`

services/QuillLMS/engines/comprehension/app/models/comprehension/prompt.rb
```diff
@@ -58,5 +62,11 @@ def serializable_hash(options = nil)
         end
       end
     end
+
+    private def log_update
+      if text_changed?
+        log_change(1, :update_prompt, self, {url: activity.url, conjunction: conjunction}.to_json, "text", text_was, text)
```
dandrabik: Is this `1` supposed to be here? This has been a `user_id` in other places.

services/QuillLMS/engines/comprehension/app/models/comprehension/rule.rb
```diff
@@ -113,5 +141,66 @@ def display_name
         errors.add(:prompts, "prompt #{prompt.id} already has a plagiarism rule") if prompt.rules.where(rule_type: TYPE_PLAGIARISM).first&.id
       end
     end
+
+    private def log_creation
+      if regex?
+        log_change(nil, :create_regex, self, {url: url}.to_json, nil, nil, nil)
```
dandrabik: Same comment here. I don't think we need to store different change types for different rule types. Instead we should record that a rule changed, and when we show the rule type in the admin panel.

services/QuillLMS/engines/comprehension/lib/comprehension/comprehension/change_log.rb
```diff
@@ -0,0 +1,17 @@
...
+    def log_change(user_id, action, changed_record, explanation = nil, changed_attribute = nil, previous_value = nil, new_value = nil)
+      change_log = {
+        user_id: user_id,
+        action: Comprehension.change_log_class::COMPREHENSION_ACTIONS[action],
```
dandrabik, line 49: I think we could simplify this, but having generic `action`s like `'created'`, `'updated'`, `'deleted'`. Similar to other comments, for the admin display, we would then pull all of the necessary info based on the changelog.

services/QuillLMS/app/models/change_log.rb
```diff
@@ -67,13 +67,52 @@ class ChangeLog < ActiveRecord::Base
     CREATED,
     RENAMED
   ]
+  COMPREHENSION_ACTIONS = {
```
dandrabik, line 70: I mentioned this below, but I think we can remove this entirely.

services/QuillLMS/engines/comprehension/app/models/comprehension/activity.rb
```diff
@@ -1,17 +1,26 @@
...
+
     before_destroy :expire_turking_rounds
     before_validation :set_parent_activity, on: :create
+    after_save :log_creation
+    after_destroy :log_deletion
 
     has_many :passages, inverse_of: :activity, dependent: :destroy
     has_many :prompts, inverse_of: :activity, dependent: :destroy
     has_many :turking_rounds, inverse_of: :activity
+    has_many :change_logs
```
dandrabik: I think to make this association work, you need to make this `has_many :change_logs, as: :changed_record`

services/QuillLMS/engines/comprehension/lib/comprehension/comprehension/change_log.rb
```diff
@@ -0,0 +1,17 @@
...
+        changed_record_type: changed_record.class.name,
+        changed_record_id: changed_record.id,
+        explanation: explanation,
```
dandrabik: I think we should remove this explanation field. It looks like you are storing a json string here, which is not it's intended use.

services/QuillLMS/engines/comprehension/lib/comprehension/comprehension/fetch_change_logs.rb
```diff
@@ -0,0 +1,77 @@
...
+  module FetchChangeLogs
+    include Comprehension::ChangeLog
+    UNIVERSAL_RULE_ACTIONS = ['Universal Rule - updated', 'Universal Rule - created']
+
+    def change_logs_for_activity(activity)
+      @activity = activity
+      @change_log = Comprehension.change_log_class
+      change_logs = activity_change_logs + passages_change_logs + prompts_change_logs + universal_change_logs
+      change_logs.map {|cl|
+        cl.merge(
```
dandrabik: These seem like attributes that should be edited in the json setup in the `ChangeLog` model

services/QuillLMS/engines/comprehension/lib/comprehension/comprehension/fetch_change_logs.rb
```diff
@@ -0,0 +1,77 @@
...
+          {
+          "user": Comprehension.user_class.find_by(id: cl["user_id"])&.name,
+          "updated_local_time": cl["updated_at"].localtime.to_s
+          })
+      }
+    end
+
+    def universal_change_logs
+      @change_log.where(action: UNIVERSAL_RULE_ACTIONS).map(&:attributes) || []
```
dandrabik: Using an instance variable here is a fragile, i.e. this method will error unless `change_logs_for_activity(activity)` is called before it to set `@change_log`. We should either pass in a variable or call this on `Comprehension.change_log_class`

services/QuillLMS/engines/comprehension/lib/comprehension/comprehension/fetch_change_logs.rb
```diff
@@ -0,0 +1,77 @@
+module Comprehension
+  module FetchChangeLogs
+    include Comprehension::ChangeLog
+    UNIVERSAL_RULE_ACTIONS = ['Universal Rule - updated', 'Universal Rule - created']
+
+    def change_logs_for_activity(activity)
+      @activity = activity
```
dandrabik: So, if we set up the Rails associations correctly, we can simplify this a lot, e.g.
```ruby
passages_change_logs = activity.passages.map(&:change_logs).flatten
```
It could also be sped up by eager loading, e.g.
```ruby 
Activity.where(id: activity_id)
  .includes(passages: [:change_logs], prompts: [:change_logs])

File: services/QuillLMS/engines/comprehension/app/models/comprehension/activity.rb (hunk 1 above)
Comment (cissyxyu) [reply services/QuillLMS/engines/comprehension/app/models/comprehension/activity.rb]: Pretty sure you can call `.present?` on `nil`. The source code for `present?` just calls `!blank?` which does return for values like `nil`.

https://apidock.com/rails/Object/present%3F

**Review by dandrabik — CHANGES_REQUESTED** [review]
So this is great progress! I'm requesting changes around a few keys areas:
1. We discussed this a bit, but we should store more generic information around the changes in the DB, and generate the material for the admin views, e.g. don't store urls in the DB.
2. We should use rails associations more to simplify the data pulls.
3. Using model's `attr_accessor` as a standard attribute and use the standard `save` and `activate` methods.

It might be useful to hop on a quick call to talk through some of this. Ping me when you have a chance to read through this.

File: services/QuillLMS/engines/comprehension/lib/comprehension/comprehension/change_log.rb (hunk 2 above)
Comment (cissyxyu) [reply services/QuillLMS/engines/comprehension/lib/comprehension/comprehension/change_log.rb:15]: I've added in the ChangeLog module includes feature as discussed!

**Comment by cissyxyu** [comment]
> So this is great progress! I'm requesting changes around a few keys areas:
> 
> 1. We discussed this a bit, but we should store more generic information around the changes in the DB, and generate the material for the admin views, e.g. don't store urls in the DB.
> 2. We should use rails associations more to simplify the data pulls.
> 3. Using model's `attr_accessor` as a standard attribute and use the standard `save` and `activate` methods.
> 
> It might be useful to hop on a quick call to talk through some of this. Ping me when you have a chance to read through this.

I've added in the changes we discussed!

File: services/.rubocop_todo.yml [hunk 19]
```diff
@@ -405,6 +405,8 @@ Metrics/BlockLength:
 # Configuration parameters: CountComments.
 Metrics/ClassLength:
   Max: 700
+  Exclude:
+    - 'QuillLMS/engines/comprehension/test/controllers/comprehension/rules_controller_test.rb'
```
happythenewsad: This file no longer exists, so we can remove this change.

services/QuillLMS/engines/comprehension/lib/comprehension/comprehension/change_log.rb
```diff
@@ -0,0 +1,129 @@
...
+      }.flatten
+    end
+
+    def automl_change_logs(prompt)
+      prompt.automl_models.map(&:change_logs).flatten || []
+    end
+
+    def rules_change_logs(prompt)
+      prompt.rules.map { |rule|
+        next [] if rule.universal_rule_type?
```
happythenewsad: optional: `filter` would be easier to read, rather than `next [] if...`

services/QuillLMS/engines/comprehension/spec/lib/change_log_module_test.rb
```diff
@@ -0,0 +1,53 @@
+require 'test_helper'
```
happythenewsad: This should be a spec file.

happythenewsad (changes requested):
Nice work - almost there

happythenewsad (approved):

services/QuillLMS/app/models/change_log.rb
```diff
@@ -110,4 +142,38 @@ def record_is_not_being_created_from_cms?
...
+    options ||= {}
+
+    super(options.reverse_merge(
+      only: [:id, :action, :changed_attribute, :changed_record_type, :changed_record_id,
+             :explanation, :new_value, :previous_value, :created_at, :updated_at, :user_id],
+      methods: [:comprehension_action, :comprehension_url, :comprehension_name, :conjunctions, :user, :updated_local_time]
+    ))
+  end
+
+  def comprehension_action
```
dandrabik: Let's make these methods more generic so this pattern can be used by non-comprehension projects, e.g. name this something like `action_text` or `action_expanded` or something.

services/QuillLMS/app/models/change_log.rb
```diff
@@ -110,4 +142,38 @@ def record_is_not_being_created_from_cms?
...
+    "#{changed_record&.change_log_name} - #{action}" if changed_record&.respond_to? :change_log_name
```
dandrabik: Per style guide, let's use parentheses when the method returns a value. https://github.com/empirical-org/ruby#method-calls

A few lines in this file should use this change.
```suggestion
    "#{changed_record&.change_log_name} - #{action}" if changed_record&.respond_to?(:change_log_name)
```

services/QuillLMS/app/models/change_log.rb
```diff
@@ -110,4 +142,38 @@ def record_is_not_being_created_from_cms?
...
+  end
+
+  def comprehension_url
```
dandrabik: Let's use a generic `changed_record_url` name instead so this can be reused.

services/QuillLMS/app/models/change_log.rb
```diff
@@ -110,4 +142,38 @@ def record_is_not_being_created_from_cms?
...
+    changed_record&.url if changed_record&.respond_to? :url
+  end
+
+  def comprehension_name
```
dandrabik: I don't have a good suggestion for a name, but again, let's name it something with `comprehension`.

services/QuillLMS/engines/comprehension/app/models/comprehension/rule.rb
```diff
@@ -80,6 +82,63 @@ def display_name
...
+
+    def universal_rule_type?
+      rule_type == TYPE_SPELLING || rule_type == TYPE_GRAMMAR
+    end
+
+    def universal?
+      universal
+    end
+
+    def change_log_name
```
dandrabik, line 105: This you can take or leave since your approach works, but similar to the banner project and the use of `DISPLAY_NAMES` above, you could create a map of `rule_type` to `name` and just do a hash lookup here, e.g.

```ruby
NAMES_FOR_TYPES = {
  TYPE_AUTOML => "Semantic Label",
  TYPE_PLAGIARISM  => "Plagiarism Rule"
...
}

def change_log_name
  NAMES_FOR_TYPES[rule_type] || ""
end

File: services/QuillLMS/engines/comprehension/lib/comprehension/comprehension/change_log.rb [hunk 27]
```diff
@@ -0,0 +1,128 @@
...
+# For classes with the attribute 'text' defined, the module will only log changes on the text field.
+# For classes without 'text' defined, the module will log all changes.
+
+# If you need to customize logging behavior, define log_creation or log_update in the class itself
+# to override default module behavior.
+
+module Comprehension
+  module ChangeLog
+    extend ActiveSupport::Concern
+    UNIVERSAL_RULE_ACTIONS = ['Universal Rule - updated', 'Universal Rule - created']
```
dandrabik: Is this used anymore?

services/QuillLMS/engines/comprehension/spec/controllers/comprehension/activities_controller_spec.rb
```diff
@@ -70,6 +71,18 @@ module Comprehension
         expect(Activity.count).to(eq(1))
       end
 
+      it "should make a change log record after creating the Activity record" do
+        post :create, params: { activity: { parent_activity_id: activity.parent_activity_id, scored_level: activity.scored_level, target_level: activity.target_level, title: activity.title, notes: activity.notes }}
+
+        new_activity = Activity.last
+        change_log = Comprehension.change_log_class.last
+        assert_equal change_log.serializable_hash["comprehension_action"], "Comprehension Activity - created"
```
dandrabik: Since we just converted to `rspec`, we should convert these `minitest` assertions to rspec, e.g. `expect(thing).to_be "this thing"`. Ping @happythenewsad if you'd like guidance here.

dandrabik (approved):
Overall this looks great! Great iterations on this! 

I have a few items that I think we should update, but otherwise this looks good to go. So I'm going to approve this (I don't need to re-review) with the expectation that these two items will be updated:
1. Convert the `change_log` methods to generic names that can be reused
2. Convert the tests to `rspec` methods (since we just ported this).

The rest of my comments you can take or leave.

Nice work on this! 💯 💯 💯

PR #7929, opened 2021-07-14
Teacher Activity Feed - don’t show deleted users.

dandrabik:
## WHAT
Tiny `nil` error fix. If a student is deleted, don't attempt to show their sessions in the Teacher Activity Feed. 

## WHY
If a student in your recent activity was deleted, the feed would error out. That's a bad experience. Also, to respect account deletion, we shouldn't show the record at all (even without the name). 
## HOW
Write a failing test, fix the test. Scope the session query to remove blank users.

### Notion Card Links
Sentry Error
https://sentry.io/organizations/quillorg-5s/issues/2512449125/?environment=production&project=11238&referrer=alert_email

PR Checklist | Your Answer
------------ | -------------

### DIFF
--- a/services/QuillLMS/spec/models/teacher_activity_feed_spec.rb
+++ b/services/QuillLMS/spec/models/teacher_activity_feed_spec.rb
@@ -27,22 +27,6 @@ describe TeacherActivityFeed, type: :model do
       end
     end
 
-    context "activity_session without user" do
-      it 'shouldnt show in activity feed' do
-        TeacherActivityFeed.new(1).send(:delete_all)
-        activity_session_with_user = create(:activity_session, user: create(:user))
-        activity_session_without_user = create(:activity_session, user_id: nil)
-
-        TeacherActivityFeed.add(1, activity_session_with_user.id)
-        TeacherActivityFeed.add(1, activity_session_without_user.id)
-
-        feed = TeacherActivityFeed.get(1)
-
-        expect(feed.size).to eq(1)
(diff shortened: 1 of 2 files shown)


services/QuillLMS/spec/models/teacher_activity_feed_spec.rb
```diff
@@ -27,7 +27,23 @@
       end
     end
 
-    context "activity_session's user gets after storage" do
+    context "activity_session without user" do
+      it 'shouldnt show in activity feed' do
+        TeacherActivityFeed.new(1).send(:delete_all)
+        activity_session_with_user = create(:activity_session, user: create(:user))
+        activity_session_without_user = create(:activity_session, user_id: nil)
```
brendanshean: It looks like the activity_session factory has user assignment done indirectly via a [callback](https://github.com/empirical-org/Empirical-Core/blob/develop/services/QuillLMS/spec/factories/activity_session.rb#L27). A couple options are assign nil user after creation:

```
activity_session_without_user = create(:activity_session)
activity_session_without_user.update(user: nil)
```

or update the factory.

dandrabik: Ah, yeah, somehow missed that. I'm going to delete this test, since I already have a test editing the session after the fact. I think that gives this feature enough coverage.

brendanshean (approved):
Looks good to me!

emilia-friedberg (approved):

PR #7947, opened 2021-07-20
Fix bug with unpermitted attributes for active activity sessions update

brendanshean:
## WHAT
Fix a bug involving nested parameters in an `update` action for active activity sessions.

## WHY
This bug will materialize when we upgrade to Rails 5.1, so it's better to deal with it now.

## HOW
Add `permit!` call on the params hash.

### Screenshots

### Notion Card Links
https://www.notion.so/quill/SentryError-Api-V1-ActiveActivitySessionsController-update-ActionController-UnfilteredParamet-d4e5b636581e43d9a1c5d11fab66db42

PR Checklist | Your Answer
------------ | -------------

### DIFF
--- a/services/QuillLMS/app/controllers/api/v1/active_activity_sessions_controller.rb
+++ b/services/QuillLMS/app/controllers/api/v1/active_activity_sessions_controller.rb
@@ -33,7 +33,7 @@ class Api::V1::ActiveActivitySessionsController < Api::ApiController
   end
 
   private def valid_params
-    params.require(:active_activity_session).except(:uid)
+    params.require(:active_activity_session).except(:uid).permit!
   end
 
   private def activity_session_by_uid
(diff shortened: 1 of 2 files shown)


dandrabik (approved):
Thanks for fixing this and thanks for backfilling specs for this controller. 💯

brendanshean:
> Thanks for fixing this and thanks for backfilling specs for this controller. 💯

You're welcome.  To be fair, I only added [one test](https://github.com/empirical-org/Empirical-Core/pull/7947/files#diff-5db7248663d966b06cb1a05f339a2966e93a71b9a64f05dc619c3d213e2624dbR28).  The rest were already written [here](https://github.com/empirical-org/Empirical-Core/blob/develop/services/QuillLMS/spec/controllers/api/v1/active_activity_session_controller.rb) but that file lacked the suffix: `_spec`

PR #7959, opened 2021-07-21
Upgrade Rails to 5.1.7, attempt #3

brendanshean:
## WHAT
Upgrade LMS from 5.0.7 to 5.1.7
There have been two previous failures with this commit.  The [second](https://dashboard.heroku.com/apps/empirical-grammar/activity/builds/f150dd0f-7c02-4b31-bb02-2b6c9b1988cc) failed deploy yielded some clues about a potential culprit:

```
       stderr:
       One CLI for webpack must be installed. These are recommended choices, delivered as separate packages:
        - webpack-cli (https://github.com/webpack/webpack-cli)
          The original webpack full-featured CLI.
       We will use "npm" to install the CLI via "npm install -D".
       Do you want to install 'webpack-cli' (yes/no): 
```

`webpack-cli` is listed as a devDependency in this [file](https://github.com/empirical-org/Empirical-Core/blob/develop/services/QuillLMS/client/package.json#L159), but during the build process, heroku prunes devDependencies (including I suspect `webpack-cli`):

```
-----> Pruning devDependencies
       removed 2 packages and audited 134 packages in 1.456s
       
       3 packages are looking for funding
         run `npm fund` for details
```

This has not been happening on staging apps since the ENV variable `NODE_ENV` is set to staging:

```
       
-----> Pruning devDependencies
       Skipping because NODE_ENV is not 'production'
```

Pruning can be [skipped](https://devcenter.heroku.com/articles/nodejs-support#skip-pruning) by setting the `NPM_CONFIG_PRODUCTION` flag to false.

```
-----> Pruning devDependencies
       Skipping because NPM_CONFIG_PRODUCTION is 'false'
```

## WHY
Library updates are important for new functionality, security, as well as compatibility with other libraries.

## HOW
Update Gemfile and comprehension.gemspec

### Screenshots

### Notion Card Links
https://www.notion.so/quill/LMS-Rails-5-1-Upgrade-2b84f704d83848e29a7854a49c2abbfb

PR Checklist | Your Answer
------------ | -------------
Have you added and/or updated tests? |  No.  This is an upgrade.  No new functionality


anathomical (approved):

dandrabik (changes requested):
Nice work diagnosing this! 

So, if I'm understanding correctly, to get this to work, you'd like to compile `devDependencies` for the `production` environment by setting `NPM_CONFIG_PRODUCTION='false'`? I think that might work, but then we are including a lot of non-production libraries in our production app (testing libraries, dev servers, devtools etc.). 

I think we should try another approach first, which is moving that single library `webpack-cli`, from `devDependencies` to `dependencies`.

What I'd also propose is instead of making `production` better match `staging`, we make `staging` better match `production`, and get this working there, i.e. prune `devDependencies` on staging as well.

brendanshean:
> Nice work diagnosing this!
> 
> So, if I'm understanding correctly, to get this to work, you'd like to compile `devDependencies` for the `production` environment by setting `NPM_CONFIG_PRODUCTION='false'`? I think that might work, but then we are including a lot of non-production libraries in our production app (testing libraries, dev servers, devtools etc.).
> 
> I think we should try another approach first, which is moving that single library `webpack-cli`, from `devDependencies` to `dependencies`.
> 
> What I'd also propose is instead of making `production` better match `staging`, we make `staging` better match `production`, and get this working there, i.e. prune `devDependencies` on staging as well.

That was my original plan, but I got stuck with that approach in a previous [build](https://dashboard.heroku.com/apps/empirical-grammar-staging/activity/builds/04e1a32c-afae-43e4-b4d9-d72d342f281f)  .  I will give it another shot since I agree that having `staging` match `production` makes more sense.

brendanshean:
@dandrabik I've removed the `NPM_CONFIG_PRODUCTION` setting like we discussed and moved all the essential dependencies over from devDependencies and the build is now working on `empirical-grammar-staging`.  Staging should now be more like production.

dandrabik (approved):
Nice! Glad we were able to make `staging` closer to `production` in setup, that's a nice side effect of this project.

Looks good to me! 🚢   💯  💯  💯

happythenewsad (approved):

PR #7961, opened 2021-07-21
Comprehension/integration tests

anathomical:
## WHAT
This is an integration test framework that is designed to validate that the Go Feedback endpoint.  Its primary intent is to let us store a baseline set of Comprehension entries and their expected feedback data, and then confirm that that feedback data is still coming back for those entries.
## WHY
It is intended to help us catch instances where either a bug is causing feedback to come back in an unexpected way, or grading has been changed with potentially unexpected consequences (that we need to check in about with Curriculum).
## HOW
- Added a new spec file that reads data from a local CSV and runs a test case on each row by calling the production Feedback endpoint and checking the live response against recorded expectations in the CSV.
- Populated an initial version of the CSV with data from what Curriculum considers the most stable Comprehension Activity in terms of grading (Surge Barriers).
- Tweaked our spec running code so that those integration tests do not run during regular CI workflows, and run only on demand instead (since they're synchronous calls to external APIs, and thus slow; and because resolving errors in these tests often requires discussion with Curriculum about grading expectations so they can not be solved purely via engineering effort).

### Notion Card Links
https://www.notion.so/quill/Integration-Test-Plan-for-Comprehension-a46fc86923e64907b48a10c9c7583c98

PR Checklist | Your Answer
------------ | -------------
Have you added and/or updated tests? |  Yes, this is a test
Have you deployed to Staging? | No, this code is only tests


services/QuillLMS/spec/integration/comprehension_integration_spec.rb
```diff
@@ -0,0 +1,50 @@
+require 'csv'
+require 'httparty'
+
+describe 'ComprehensionIntegration' do
+  FEEDBACK_URL = 'https://comprehension-feedback.quill.org'
+  SESSION_ID = 'INTEGRATION_TEST'
+
+  # CSV generated via the following query
+  # SELECT prompt_id, c_p.text AS stem, entry, f_h.optimal, c_r.rule_type, rule_uid, f_h.concept_uid
```
dandrabik: Maybe move this query to Metabase (and link to it), both to clear up this file and to make csv generation easy.

services/QuillLMS/.rspec
```diff
@@ -1,4 +1,5 @@
 --colour
 --color
 --format progress
---format ParallelTests::RSpec::SummaryLogger --out tmp/spec_summary.log
\ No newline at end of file
+--format ParallelTests::RSpec::SummaryLogger --out tmp/spec_summary.log
+--exclude-pattern "spec/integration/*"
```
dandrabik: We might want to rename this directory to something slightly different, since `integration` can be used by `rspec` for `request` specs. https://relishapp.com/rspec/rspec-rails/docs/directory-structure
> Request specs reside in the spec/requests directory. The directory can also be named integration or api.

dandrabik (approved):
Small comments, but this looks like a light-weight way to run some production tests as we approach launch.

services/QuillLMS/.rspec
brendanshean: I would second this recommendation.  This would also keep consistency with the [CMS](https://github.com/empirical-org/Empirical-Core/tree/develop/services/QuillCMS/spec/requests)

brendanshean: Another approach might be using [tags](https://relishapp.com/rspec/rspec-core/v/2-5/docs/command-line/tag-option) on specific files/tests rather than whole directories.  For example,

`describe 'ComprehensionIntegration', external_apis: :true do`

Then you can specify which tags to exclude from normal circleci runs.

brendanshean (approved):
Looks good.  Thanks for getting this off the ground.  A minor comment on rspec that you can take or leave.

happythenewsad (commented):
Looks good. Couple notes:
- how long does this suite take?
- let's update the documentation (Notion + README) so that it's understood that this should be run before deploying
- this test suite doesn't need rails. How much time is it costing us to use `rspec-rails` here?

anathomical:
> Looks good. Couple notes:
> 
> * how long does this suite take?
> * let's update the documentation (Notion + README) so that it's understood that this should be run before deploying
> * this test suite doesn't need rails. How much time is it costing us to use `rspec-rails` here?

Solid questions.
- The suite currently takes between 60 and 90 seconds on my local machine
- Actually, this is NOT intended to be run before deploying.  This test explicitly runs against production, so a pre-deploy run wouldn't really catch any new issues with the code.  Additionally, the intent is really to have this available as an on-demand thing and not as part of our regular development process.
- Agreed that it doesn't require Rails, and that `bundle` is probably costing 15-30 seconds of run time, but Dan and I talked about it and decided that since this is technically a test it made sense to keep this code close to that since that's least surprising.  Though an argument could be made that putting it in `spec/` actually might surprise people when they find that it doesn't run regularly?

anathomical:
@happythenewsad 
> Looks good. Couple notes:
> 
> * how long does this suite take?
> * let's update the documentation (Notion + README) so that it's understood that this should be run before deploying
> * this test suite doesn't need rails. How much time is it costing us to use `rspec-rails` here?

Solid questions.
- The suite currently takes between 60 and 90 seconds on my local machine
- Actually, this is NOT intended to be run before deploying.  This test explicitly runs against production, so a pre-deploy run wouldn't really catch any new issues with the code.  Additionally, the intent is really to have this available as an on-demand thing and not as part of our regular development process.
- Agreed that it doesn't require Rails, and that `bundle` is probably costing 15-30 seconds of run time, but Dan and I talked about it and decided that since this is technically a test it made sense to keep this code close to that since that's least surprising.  Though an argument could be made that putting it in `spec/` actually might surprise people when they find that it doesn't run regularly?

services/QuillLMS/spec/integration/comprehension_integration_spec.rb
anathomical: Good call.  Updated.

services/QuillLMS/.rspec
anathomical: I've changed the directory name to `skip_ci` for now.  We can change it again later if we want.

PR #7974, opened 2021-07-23
LMS - add Apple icon to prevent missing asset errors

dandrabik:
## WHAT
In Rails 5.1, a missing asset raises an error. The spike in errors and alerts from the last day are from mostly from bookmarks on iPads and iPhones that are looking for a `apple-touch-icon.png`, which is what [Apple and safari looks for](https://developer.apple.com/library/archive/documentation/AppleApplications/Reference/SafariWebContent/ConfiguringWebApplications/ConfiguringWebApplications.html) to use for an icon. I'm putting a temporary icon to stop.

Note, [I tried turning this type of error off](https://github.com/empirical-org/Empirical-Core/pull/7974/commits/ccee463e2000f96bba16f839ec847159200336eb), but it didn't have an effect. So this is a temporary mitigation.

Note, there are other missing assets (like font-awesome icons), but doing this at least to reduce the errors.
## WHY
This is causing alarms and high 500s for assets we don't currently host. 
## HOW
Add two assets. Also, I made a separate tasks ([design](https://www.notion.so/quill/ea241f31a84c4b79ad56e83e99f7ee93?v=9f7aead86cc749d8a6f963070cef4f5e&p=1101df70cf3141c89da8f40a9e553341), [code](https://www.notion.so/quill/ea241f31a84c4b79ad56e83e99f7ee93?v=9f7aead86cc749d8a6f963070cef4f5e&p=b461983e9466451ba0d8861b10181d3e)) to put an official apple icon in place, since we should have that.
### Apple specification
Screen Shot 2021-07-23 at 2 50 45 PM
### Error
Screen Shot 2021-07-23 at 2 53 49 PM

PR Checklist | Your Answer
------------ | -------------
Have you added and/or updated tests? |  No, we don't really have tests for assets existing


happythenewsad (approved):
Approving, since this PR has a measurable benefit. However, we should also look into disabling 'error on missing asset' in a more universal way.

eadams17 (approved):

dandrabik:
> Approving, since this PR has a measurable benefit. However, we should also look into disabling 'error on missing asset' in a more universal way.

Agreed! [This was the initial attempt](https://github.com/empirical-org/Empirical-Core/pull/7974/commits/ccee463e2000f96bba16f839ec847159200336eb), that according to the docs should work, but didn't have any effect. This PR is a mitigation in the meantime, given the approaching weekend.

PR #7978, opened 2021-07-26
Comprehension/fix spelling feedback

anathomical:
## WHAT
Fix a `nil` error that came up in Sentry caused when the Bing spelling API errors

At the same time, add a small tweak so that we don't swallow Bing API rate limit errors since those are things we'll want to know about
## WHY
We want to make sure that the our API doesn't error just because Bing's might.  However, we do want to record errors in cases where we're running into Bing rate-limiting
## HOW
The initial problem is solved by making sure we provide an empty array for the rest of our code to work with when we don't get the expected response payload shape back from Bing.  Then we check for the explicit HTTP status code that Bing documentation says they'll use to indicate rate limit violations and raise a custom exception that should report to Sentry and other systems.

### Screenshots

### Notion Card Links
https://www.notion.so/quill/Sentry-Error-NoMethodError-Comprehension-FeedbackController-spelling-664d69d67f1a43fd9e33a5680469dabd

PR Checklist | Your Answer
------------ | -------------

### DIFF
--- a/services/QuillLMS/engines/comprehension/lib/comprehension/comprehension/spelling_check.rb
+++ b/services/QuillLMS/engines/comprehension/lib/comprehension/comprehension/spelling_check.rb
@@ -49,7 +49,7 @@ module Comprehension
     end
 
     private def misspelled
-      bing_response['flaggedTokens']
+      bing_response['flaggedTokens'] || []
     end
 
     private def bing_response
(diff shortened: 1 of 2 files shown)


cissyxyu (approved):

brendanshean (approved):
Looks good to me!  I appreciate the clear testing and documentation.

PR #7988, opened 2021-07-28
Lms/security headers

happythenewsad:
## WHAT
- includes `secure_headers` gem in LMS
- enables HttpOnly and Secure flags for LMS-managed cookies
- I audited all instances of client-side scripts reading cookies, and all of them are set by client-side scripts, so HttpOnly will not break existing features.

## DISCUSSION
- I notice that we use two third-party non-Secure cookies: Google Analytics and Amplitude. I confirmed they do not contain sensitive information (they are counters/hashed data).
- We should minimize attack vectors in general, apart from the details of the CB review. We should consider implementing a security checklist whenever a new tracker is added to the app.

## WHY
- college board requirement 
- reduces possibility of leaking sensitive cookie data

### Screenshots

### Notion Card Links
https://www.notion.so/quill/HttpOnly-Cookie-Attribute-Not-Set-0659fd9e4e8a446fac89cfc40f3dc4ce
https://www.notion.so/quill/Product-Board-30f97e4eb01246dbb8264253fb385073?p=d1aa65095bd24bf58e9fe455c94754b7

PR Checklist | Your Answer
------------ | -------------

### DIFF
--- a/services/QuillLMS/config/initializers/secure_headers.rb
+++ b/services/QuillLMS/config/initializers/secure_headers.rb
@@ -1,0 +1,9 @@
+SecureHeaders::Configuration.default do |config|
+  config.cookies = {
+    secure: true, # mark all cookies as "Secure"
+    httponly: true, # mark all cookies as "HttpOnly"
+    samesite: {
+      lax: true # mark all cookies as SameSite=lax
+    }
+  }
+end
--- a/services/QuillLMS/config/initializers/secure_headers.rb
+++ b/services/QuillLMS/config/initializers/secure_headers.rb
@@ -1,9 +1,10 @@
 SecureHeaders::Configuration.default do |config|
+  config.csp = SecureHeaders::OPT_OUT
   config.cookies = {
-    secure: true, # mark all cookies as "Secure"
-    httponly: true, # mark all cookies as "HttpOnly"
+    secure: true, 
+    httponly: true, 
     samesite: {
-      lax: true # mark all cookies as SameSite=lax
+      lax: true 
     }
   }
 end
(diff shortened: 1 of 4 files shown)


services/QuillLMS/config/initializers/secure_headers.rb
```diff
@@ -0,0 +1,10 @@
+SecureHeaders::Configuration.default do |config|
+  config.csp = SecureHeaders::OPT_OUT
+  config.cookies = {
```
brendanshean, line 3: The cookie configuration for `SameSite=lax`,`secure`, and `httponly` might already be set out of the box: https://github.com/github/secure_headers/blob/main/docs/upgrading-to-5-0.md

services/QuillLMS/spec/requests/secure_headers_spec.rb
```diff
@@ -0,0 +1,25 @@
+require 'rails_helper'
+
+class DummyController < ApplicationController
```
brendanshean, line 3: I really like this pattern with dummy controller and routes.

brendanshean (approved):
Nice work here!  Minor comment about possibly removing some of the configuration.

happythenewsad:
@brendanshean For some reason I can't quote-reply in Github today. Yes, you are right that the SecureHeaders cookie configs reflect the defaults, and thus are not necessary. I'm keeping them here so that other developers know they are set (otherwise, another developer would have to have read the SecureHeaders documentation to know what the config is doing).

dandrabik (approved):
This looks good to me, thanks for working through this!

PR #7991, opened 2021-07-28
Lms/count user activity completions

anathomical:
## WHAT
Start counting the each time a user completes an activity by classification.  This will also allow us to pass that value on to activities so that individual clients can change behavior based on it if they want to.
## WHY
Ultimately we want to start using this value to decide how much of a tutorial to show to a user, but for now we're just collecting and passing this data on without actually using it.
## HOW
Hook into the `ActivitySession` completion workflow to increment counts in `UserActivityClassification` whenever a user completes a session.

### Notion Card Links
https://www.notion.so/quill/Quill-LMS-shares-number-of-previously-completed-activities-per-tool-at-the-start-of-the-activity-de4db8fbe1c34664a4de189fca935eb4

PR Checklist | Your Answer
------------ | -------------

### DIFF
--- a/services/QuillLMS/app/controllers/api/v1/activity_sessions_controller.rb
+++ b/services/QuillLMS/app/controllers/api/v1/activity_sessions_controller.rb
@@ -102,7 +102,7 @@ class Api::V1::ActivitySessionsController < Api::ApiController
   private def count_completed_activity
     return unless @activity_session.finished?
 
-    counter = UserActivityClassificationCount.find_or_create_by(user: @activity_session.user, activity_classification: @activity_session.classification)
+    counter = UserActivityClassification.find_or_create_by(user: @activity_session.user, activity_classification: @activity_session.classification)
     counter.increment_count
   end
 
--- a/services/QuillLMS/app/controllers/api/v1/activity_sessions_controller.rb
+++ b/services/QuillLMS/app/controllers/api/v1/activity_sessions_controller.rb
@@ -20,6 +20,8 @@ class Api::V1::ActivitySessionsController < Api::ApiController
       message = "Activity Session Updated"
       NotifyOfCompletedActivity.new(@activity_session).call if @activity_session.classroom_unit_id
       handle_concept_results
+      # TODO: turn this next line back on once we've backfilled the data
+      # count_completed_activity
     else
       status = :unprocessable_entity
       message = "Activity Session Update Failed"
@@ -95,6 +97,13 @@ class Api::V1::ActivitySessionsController < Api::ApiController
 
   private def find_activity_session
     @activity_session = ActivitySession.unscoped.find_by_uid!(params[:id])
+  end
+
+  private def count_completed_activity
+    return unless @activity_session.finished?
+
+    counter = UserActivityClassificationCount.find_or_create_by(user: @activity_session.user, activity_classification: @activity_session.classification)
+    counter.increment_count
   end
 
   private def activity_session_params
--- a/services/QuillLMS/app/models/activity.rb
+++ b/services/QuillLMS/app/models/activity.rb
@@ -136,7 +136,7 @@ class Activity < ApplicationRecord
 
   def module_url(activity_session)
     @activity_session = activity_session
-    classification_count = @activity_session&.classification&.user_activity_classifications&.where(user: @activity_session.user_id)&.first&.count
+    classification_count = @activity_session&.user_activity_classifications&.find_by(user_id: @activity_session.user_id)&.count
     initial_params = {student: activity_session.uid}
     initial_params[:activities] = classification_count if classification_count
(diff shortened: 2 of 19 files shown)


services/QuillLMS/spec/models/user_activity_classification_spec.rb
```diff
@@ -0,0 +1,57 @@
...
+    it { should validate_presence_of(:user) }
+    it { should validate_presence_of(:activity_classification) }
+    it { should validate_presence_of(:count) }
+
+    it { should validate_numericality_of(:count).only_integer.is_greater_than_or_equal_to(0) }
+  end
+
+  context '#new' do
+    it 'should default count to 0' do
+      new = UserActivityClassification.new()
```
brendanshean: The `new` temp variable name is a bit confusing.  What about `user_activity_classification`?

services/QuillLMS/spec/models/user_activity_classification_spec.rb
```diff
@@ -0,0 +1,57 @@
...
+      expect(new.count).to be(0)
+    end
+  end
+
+  context '#increment_count' do
+    START_COUNT = 10
```
brendanshean: Would a temp variable assignment here would make more sense:  `let(:start_count) { 10 }`.

services/QuillLMS/spec/models/activity_spec.rb
```diff
@@ -164,6 +164,27 @@
       expect(activity.module_url(student.activity_sessions.build()).to_s).to include "student"
     end
 
+    it "must add 'activities' param if it the student has completed previous sessions of this activity classification" do
```
brendanshean: "if ~~it~~ the student has..."

services/QuillLMS/spec/models/activity_spec.rb
```diff
@@ -164,6 +164,27 @@
...
+      classification = create(:activity_classification, key: 'connect')
+      classified_activity = create(:activity, classification: classification)
+      student = create(:student)
+      uac = create(:user_activity_classification, user: student, activity_classification: classification, count: 4)
+      activity_session = create(:activity_session, activity: classified_activity, user: student)
+
+      classification.reload
+
+      expect(activity_session.classification.user_activity_classifications.where(id: student.id).first.count).to eq(4)
```
brendanshean: Some suggestions for refactoring this expectation:

1.  Change finder attribute from `id` to `user_id`
2.  Add an association `has_many :user_activity_classifications, through: :classification` to activity_session.rb 
3.  Use `find_by` instead of [where(...).first](https://github.com/rubocop/rails-style-guide#find_by)
4.  Since the expectation is pulling a record, check equality with the record rather than the attribute.

 `expect(activity_session.user_activity_classifications.find_by(user_id: student.id)).to eq(uac)`

services/QuillLMS/app/controllers/api/v1/activity_sessions_controller.rb
```diff
@@ -97,6 +99,13 @@ def destroy
     @activity_session = ActivitySession.unscoped.find_by_uid!(params[:id])
   end
 
+  private def count_completed_activity
+    return unless @activity_session.finished?
+
+    counter = UserActivityClassificationCount.find_or_create_by(user: @activity_session.user, activity_classification: @activity_session.classification)
```
brendanshean: `UserActivityClassification` typo?

services/QuillLMS/app/models/activity.rb
```diff
@@ -136,7 +136,9 @@ def form_url
 
   def module_url(activity_session)
     @activity_session = activity_session
+    classification_count = @activity_session&.classification&.user_activity_classifications&.where(user: @activity_session.user_id)&.first&.count
```
brendanshean: I would use `find_by` instead of where(...).first

brendanshean (approved):
Some refactoring comments but looks good overall.  NB the [comment](https://github.com/empirical-org/Empirical-Core/pull/7991#discussion_r679083535) about possibly renaming the table.

brendanshean:
Renaming the table `UserActivityClassificationCounter` might be more informative.  Also, we might want to reserve `UserActivityClassification` model for a more generic usage down the road.

services/QuillLMS/app/controllers/api/v1/activity_sessions_controller.rb
brendanshean: This typo reminded me that I was actually thinking about suggesting renaming the table `UserActivityClassificationCounter`, `UserActivityClassificationProgress`,  or something to that effect which would express more clearly what the model is doing while also reserving the standard join table name `UserActivityClassification` for a more generic usage down the road.

happythenewsad (approved):

PR #7996, opened 2021-07-29
Unlink Google if someone signs up with Clever account with same email

brendanshean:
## WHAT
Unlink Google account if a teacher signs up with Clever Account using the same email.
Track event that attempts to link Google and Clever accounts with same email.

## WHY
We have the protection in place already for the opposite flow (i.e. User signs up with clever account and then signs up with Google account using same email), so we should also have it for the this path.

## HOW
Add check if teacher has existing google_id and if they do set to nil. 

### Screenshots

### Notion Card Links
https://www.notion.so/quill/Google-Classroom-Automatically-unlink-Google-if-someone-signs-up-with-a-Clever-account-that-has-the-4fe7bca9a3a542acbc1fe15ed6b449d9

PR Checklist | Your Answer
------------ | -------------


services/QuillLMS/spec/services/clever_integration/creators/teacher_spec.rb
```diff
@@ -0,0 +1,42 @@
...
+
+      it 'sets google_id to nil and triggers a segment event for dual clever and google linking' do
+        expect(analyzer).to receive(:track_event_from_string).with(segment_event, teacher.id)
+        CleverIntegration::Creators::Teacher.run(clever_teacher_data)
+
+        expect(teacher.reload.google_id).to be_nil
+      end
+    end
+  end
+end
```
cissyxyu: add a new line at end of every file

cissyxyu (approved):

PR #7997, opened 2021-07-29
Stop highlighting misspelled words when the student has a Missing Word Hint

cissyxyu:
## WHAT
As of now, we automatically highlight misspelled words on all algorithmic feedback. However, this is misleading for the Missing Word Hint feedback because it makes the student think the highlighted misspelled word has something to do with the missing word (which is not the case).

So Curriculum has asked us to stop highlighting misspellings on this algorithm, specifically.

## WHY
This will make feedback easier to understand.

## HOW
We currently add the spelling concept results and misspelled words array after we process the response algorithms. So, at this last step, just add an edge case to catch when the author is "Misspelled Words Hint" and omit the misspelled_words array if this is the case.

### Screenshots

### Notion Card Links
https://www.notion.so/quill/Update-the-required-words-algorithm-to-no-longer-use-bolding-3defd347fa5f474da8b728b05362d98f

PR Checklist | Your Answer
------------ | -------------

### DIFF
--- a/packages/quill-marking-logic/src/libs/graders/sentence_combining_integration.spec.ts
+++ b/packages/quill-marking-logic/src/libs/graders/sentence_combining_integration.spec.ts
@@ -92,14 +92,14 @@ describe('The checking a sentence combining question', () => {
     });
 
     it('should add spelling concept result to concept results', () => {
-      const questionString = "Batss wings, so they can fly."
+      const questionString = "Batss have wi ngs, so they can fly."
       const matchedResponse = checkSentenceCombining(responses[0].question_uid, questionString, responses, focusPoints, incorrectSequences, responses[0].question_uid);
       assert.equal(matchedResponse.spelling_error, true);
       assert.include(matchedResponse.concept_results, conceptResultTemplate('H-2lrblngQAQ8_s-ctye4g'));
     });
 
     it('should add spelling concept result to concept results if there are two or more spelling errors', () => {
-      const questionString = "Batss wingss, so they can fly."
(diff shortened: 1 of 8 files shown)


packages/quill-marking-logic/src/libs/graders/sentence_combining_integration.spec.ts
```diff
@@ -163,7 +163,7 @@ describe('The checking a sentence combining question', () => {
       const matchedResponse = checkSentenceCombining(responses[0].question_uid, questionString, responses, focusPoints, incorrectSequences, responses[0].question_uid);
       assert.equal(matchedResponse.feedback, spellingFeedbackStrings[matchedResponse.author]);
       assert.equal(matchedResponse.spelling_error, true);
-      assert.equal(matchedResponse.misspelled_words[0], 'zings')
+      assert.equal(matchedResponse.misspelled_words, undefined)
```
brendanshean, line 166: I'm having trouble making sense of this test compared with the new test on line [177](https://github.com/empirical-org/Empirical-Core/pull/7997/files#diff-4c8ca9f2ab436d457486284f86c65fd106024218321f110b11af265cf0cd9337R177).  If this is undefined now, how is it different from the new test.

cissyxyu: You're right -- it's a redundant test. I've removed the one on line 177 and kept this one in.

brendanshean (approved):
Looks good to me!

PR #7999, opened 2021-07-29
lms/count-activity-completeions-on-session-finish

anathomical:
## WHAT
Turn on updated tracking of activity classification on session update
## WHY
We want to track this data as it updates once we've backfilled it
## HOW
Hook into the `ActivitySession` completion flow to increment this counter whenever a user completes a sesison

### Notion Card Links
https://www.notion.so/quill/Quill-LMS-shares-number-of-previously-completed-activities-per-tool-at-the-start-of-the-activity-de4db8fbe1c34664a4de189fca935eb4

PR Checklist | Your Answer
------------ | -------------

### DIFF
--- a/services/QuillLMS/spec/controllers/api/v1/activity_sessions_controller_spec.rb
+++ b/services/QuillLMS/spec/controllers/api/v1/activity_sessions_controller_spec.rb
@@ -4,7 +4,9 @@ describe Api::V1::ActivitySessionsController, type: :controller do
   describe '#update' do
     let(:token) { double :acceptable? => true, resource_owner_id: user.id }
     let(:user) { create(:student) }
-    let!(:activity_session) { create(:activity_session, state: 'started', user: user, completed_at: nil) }
+    let(:activity_classification) { create(:activity_classification) }
+    let(:activity) { create(:activity, classification: activity_classification) }
+    let!(:activity_session) { create(:activity_session, state: 'started', user: user, completed_at: nil, activity: activity) }
 
     before { allow(controller).to receive(:doorkeeper_token) { token } }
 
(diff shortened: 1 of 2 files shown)


services/QuillLMS/spec/controllers/api/v1/activity_sessions_controller_spec.rb
```diff
@@ -39,6 +41,27 @@
...
+        expect(UserActivityClassification.find_by(user: user, activity_classification: activity_classification).count).to eq(start_count + 1)
+      end
+
+      it 'should not touch user_activity_classification counts if the state of the session is not "finished"' do
+        records = UserActivityClassification.count
+        put :update, params: { id: activity_session.uid }, as: :json
+        expect(UserActivityClassification.count).to eq(records)
+      end
+    end
+
```
brendanshean, line 68: rspec has a [built-in change matcher](https://relishapp.com/rspec/rspec-expectations/docs/built-in-matchers/change-matcher#expect-change) that you can use to track these types of specs.  Basically, it's:

```
expect do
  [...block of code that changes object...]
end.to change { object.method }.by(value)
```
or the one line form:  `expect { block of code }.to change { object.method }.by(value)`

Initially, I found this pattern hard to grok but I now prefer it gives a nice scaffold for this types of tests, especially when ones involving counts on database objects.

A rewrite would look something like this:

```
      it 'should create a new user_activity_classification row if necessary' do
        expect do
          put :update, params: { id: activity_session.uid, state: 'finished' }, as: :json
        end.to change { UserActivityClassification.count }.by(1)
      end

      it 'should increment an existing user_activity_classification row if one exists' do
        start_count = 10
        uac = UserActivityClassification.create(user: user, activity_classification: activity_classification, count: start_count)

        expect do
          put :update, params: { id: activity_session.uid, state: 'finished' }, as: :json
          uac.reload
        end.to change { uac.count }.from(start_count).to(start_count + 1)
      end

      it 'should not touch user_activity_classification counts if the state of the session is not "finished"' do
        expect do
          put :update, params: { id: activity_session.uid }, as: :json
        end.to_not change { UserActivityClassification.count }
      end
```

There's also an alternative parameterized form for `change` where instead of a block you pass parameters change(UserActivityClassification, :count).

brendanshean (approved):
Looks good.  Just a stylistic note about specs that you can take or leave.

services/QuillLMS/spec/controllers/api/v1/activity_sessions_controller_spec.rb
anathomical: Seems reasonable.  Updated.

brendanshean: @anathomical I just realized that rubocop doesn't like the block form of change.  (I hit the same issue [here](https://app.circleci.com/pipelines/github/empirical-org/Empirical-Core/10239/workflows/d8523f85-b444-4863-a45f-c581ffa32ee0/jobs/174896))  Good news though is the parameterized version should work fine:

```
change (UserActivityClassification, :count)
change (uac, :count)
change (UserActivityClassification.count)
```

PR #8005, opened 2021-07-30
comprehension/strip-model-id-whitespace

anathomical:
## WHAT
Strip white space from automl_model_id

Both in validation and when populating data from Google (which happens pre-validation)
## WHY
Sometimes when copying and pasting from other places, a user might accidentally provide an AutoML Model ID with extraneous spaces around it.  We want to make sure that we don't store or use the extra spaces.
## HOW
Add a `before_validation` call to strip whitespace from the value, and then make sure to explicitly also call `strip` when sending the value to Google because we do that before validation during initial creation.

### Notion Card Links
https://www.notion.so/quill/Sentry-Error-ArgumentError-Comprehension-AutomlModelsController-create-80c9a456e3d0409695186bad8d44803d

PR Checklist | Your Answer
------------ | -------------

### DIFF
--- a/services/QuillLMS/engines/comprehension/app/models/comprehension/automl_model.rb
+++ b/services/QuillLMS/engines/comprehension/app/models/comprehension/automl_model.rb
@@ -21,6 +21,8 @@ module Comprehension
     validates :name, presence: true
     validates :state, inclusion: {in: ['active', 'inactive']}
 
+    before_validation :strip_whitespace
+
     def serializable_hash(options = nil)
       options ||= {}
 
@@ -83,6 +85,10 @@ module Comprehension
       name
     end
 
+    private def strip_whitespace
+      self.automl_model_id = automl_model_id.strip unless automl_model_id.nil?
+    end
+
     private def prompt_automl_rules
       prompt.rules.where(rule_type: Rule::TYPE_AUTOML)
     end
@@ -115,7 +121,7 @@ module Comprehension
     end
 
     private def automl_model_full_id
-      @model_full_id ||= automl_client.model_path(project: ENV['AUTOML_GOOGLE_PROJECT_ID'], location: ENV['AUTOML_GOOGLE_LOCATION'], model: automl_model_id)
+      @model_full_id ||= automl_client.model_path(project: ENV['AUTOML_GOOGLE_PROJECT_ID'], location: ENV['AUTOML_GOOGLE_LOCATION'], model: automl_model_id.strip)
     end
 
     private def automl_labels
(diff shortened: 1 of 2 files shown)


services/QuillLMS/engines/comprehension/app/models/comprehension/automl_model.rb
```diff
@@ -115,7 +121,7 @@ def comprehension_name
     end
 
     private def automl_model_full_id
-      @model_full_id ||= automl_client.model_path(project: ENV['AUTOML_GOOGLE_PROJECT_ID'], location: ENV['AUTOML_GOOGLE_LOCATION'], model: automl_model_id)
+      @model_full_id ||= automl_client.model_path(project: ENV['AUTOML_GOOGLE_PROJECT_ID'], location: ENV['AUTOML_GOOGLE_LOCATION'], model: automl_model_id.strip)
```
brendanshean, line 124: I might be confused here but it seems like this `strip` call is redundant given the `strip_whitespace` validation.

happythenewsad: Is this a convenience hack to avoid backfilling? I'm cool with that

services/QuillLMS/engines/comprehension/spec/models/comprehension/automl_model_spec.rb
```diff
@@ -42,6 +42,14 @@ module Comprehension
         end
       end
 
+      context '#before_validation' do
+        it 'should strip whitespace from "automl_model_id"' do
+          automl_model = build(:comprehension_automl_model, automl_model_id: '  HAS_SPACES   ')
+          automl_model.valid?
+          expect(automl_model.automl_model_id).to eq('HAS_SPACES')
```
happythenewsad: Can we rename the dummy string literal? `'HAS_SPACES'` does not in fact have spaces.

happythenewsad (changes requested):

cissyxyu (approved):

services/QuillLMS/engines/comprehension/app/models/comprehension/automl_model.rb
anathomical: @brendanshean @happythenewsad In practice `automl_model_full_id` will almost always be called before `save`, and thus it will be called before `valid?`.  That means that the `before_validation` hooks won't have triggered before this function gets called in a number of scenarios.  So, yeah, kind of redundant, but I couldn't think of a cleaner way to handle this.

brendanshean: Would `after_initialize` apply here? (specifically in reference to `automl_model_id`)

services/QuillLMS/engines/comprehension/spec/models/comprehension/automl_model_spec.rb
anathomical: Fixed

happythenewsad (approved):

PR #8006, opened 2021-07-30
Allow students to be imported if they were previously imported and their email changed

brendanshean:
## WHAT
If a student's email address is changed but their google_id is not, our system will not import that student into newer classrooms since there will be a uniqueness violation with the google_id.

## WHY
We'd like to be able to import students into classrooms even when their email domain changes.

## HOW
The [edge case](https://github.com/empirical-org/Empirical-Core/blob/bs-google-classroom-student-email-change/services/QuillLMS/app/services/google_integration/classroom/creators/students.rb#L78) that needs to be handled before proceeding with pre-existing code is if a user exists in the database with the the given google_id given and a different email than the given email address, then update the email address with the given one.

### Screenshots

### Notion Card Links
https://www.notion.so/quill/Google-Classroom-Previously-imported-students-can-t-be-added-to-a-new-classroom-if-their-email-was--74b24a813ed5477cbaf32387ac4a3cc6

PR Checklist | Your Answer
------------ | -------------

### DIFF
--- a/services/QuillLMS/app/services/google_integration/classroom/creators/students.rb
+++ b/services/QuillLMS/app/services/google_integration/classroom/creators/students.rb
@@ -34,49 +34,71 @@ module GoogleIntegration::Classroom::Creators::Students
   end
 
   def self.create_students(student_data)
-    students = student_data.map do |k, v|
-      create_student(student_data[k])
-    end
-    students.compact
+    student_data
+      .map { |k, v| create_student(student_data[k]) }
+      .compact
   end
 
   def self.create_student(data, counter=0)
-    if counter > 2
-      return nil
-    end
-    if data[:email]
-      student = User.find_or_initialize_by(email: data[:email].downcase)
-      if student.new_record?
-        classroom = Classroom.unscoped.find(data[:classrooms].first)
-        username = GenerateUsername.new(
-          data[:first_name],
-          data[:last_name],
-          classroom.code
-        ).call
-        student.update(name: data[:name],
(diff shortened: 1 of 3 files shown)


services/QuillLMS/app/services/google_integration/classroom/creators/students.rb
```diff
@@ -34,49 +34,71 @@ def self.get_student_data_for_all_classrooms(classrooms, students_requester_and_
...
-        student = create_student(data, counter += 1)
-      else
-        data[:classrooms].each do |id|
-          classroom = Classroom.find(id)
-          Associators::StudentsToClassrooms.run(student, classroom)
-        end
-      end
-      student
+    if student.errors.any?
+      student = create_student(data, counter += 1)
```
anathomical, line 66: We don't have to change this since it's just a duplicate of what the code was doing before, but this doesn't need to use `+=` to increment `counter` since it's passing the value, and the locally scoped value isn't referenced again.  It's a little more legible to just pass `counter + 1` instead, but up to you if you want to bother.

emilia-friedberg (approved):

anathomical (approved):

services/QuillLMS/app/services/google_integration/classroom/creators/students.rb
brendanshean: That's a good point.  I'll probably just leave as if for now and rewrite during a later refactor.

PR #8008, opened 2021-07-30
Warn admin if they are about to break a translation mapping

cissyxyu:
## WHAT
We [had a bug](https://www.notion.so/quill/ELL-Diagnostic-1-timeout-on-question-7-d8f8b3819f254b2089dc77a6148dc608) where admin changes to a Title Card title broke an activity. This is the long-term fix -- warn admin before they change a diagnostic activity's title. Direct them to work with Support Dev to update the translation mappings before they make a breaking change.

## WHY
To avoid this bug happening again.

## HOW
Before admin change the title of a Diagnostic activity, show them a warning that tells them this change will break the translation mapping. Direct them to engineering first to change the translation mappings.

### Screenshots
Screen Shot 2021-07-30 at 10 53 13 AM

### Notion Card Links
https://www.notion.so/quill/ELL-title-card-errors-New-Error-Message-to-Prevent-Activity-Title-Card-Errors-37790c0d807d48b3a31d9e9bb9f7c7a0

PR Checklist | Your Answer
------------ | -------------
Have you added and/or updated tests? |  NO, small change.

### DIFF
--- a/services/QuillLMS/client/app/bundles/Connect/components/titleCards/titleCardForm.tsx
+++ b/services/QuillLMS/client/app/bundles/Connect/components/titleCards/titleCardForm.tsx
@@ -77,7 +77,10 @@ class TitleCardForm extends React.Component<TitleCardFormProps, TitleCardFormSta
   }
 
   warnTitleChange = (titleChanged) => {
-    if (titleChanged) {
+    const { match } = this.props
+    const { params } = match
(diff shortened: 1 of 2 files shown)


happythenewsad (approved):
Thanks for doing this, Cissy. 

Optional nice-to-have enhancement: convert this class into a react function with hooks (example: useState)

PR #8012, opened 2021-07-30
LMS - Fix 500s that should be 404s

dandrabik:
## WHAT
Since we upgraded to Rails 5.1, we've been returning a `500` for files that aren't found when it should be a `404`. This is a fix to return the correct status.
## WHY
These additional `500`s are throwing off our error reporting and causing spurious alarms. Using the proper status code of `404` will stop that.
## HOW
Rails deprecated `render nothing: true`, changing to `head: :not_found, body: nil`
```
DEPRECATION WARNING: :nothing option is deprecated and will be removed in Rails 5.1. Use `head` method to respond with empty response body.
```

PR Checklist | Your Answer
------------ | -------------

### DIFF
--- a/services/QuillLMS/app/controllers/application_controller.rb
+++ b/services/QuillLMS/app/controllers/application_controller.rb
@@ -53,8 +53,6 @@ class ApplicationController < ActionController::Base
   end
 
   def routing_error(error = 'Routing error', status = :not_found, exception=nil)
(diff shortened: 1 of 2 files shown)


anathomical (approved):

brendanshean:
@dandrabik Great sleuthing on this one!

PR #8018, opened 2021-08-02
setting more secure HTTP cache headers

happythenewsad:
## WHAT
Aligns LMS with CB's caching recommendation. Because our assets are served by a webserver and not Rails, our asset routes implicitly avoid these new cache controls, which is the desired behavior.

## WHY
So that private information does not get stored in user browsers (an attack vector)

## HOW
Set stricter cache headers for all non-asset requests. 

### Screenshots

### Notion Card Links
https://www.notion.so/quill/Cacheable-HTTPS-Content-e35a42c3ffef4e0994be33399ab014de

PR Checklist | Your Answer
------------ | -------------

### DIFF
--- a/services/QuillLMS/app/controllers/application_controller.rb
+++ b/services/QuillLMS/app/controllers/application_controller.rb
@@ -47,7 +47,7 @@ class ApplicationController < ActionController::Base
   end
 
   def show_errors
-    status = env["PATH_INFO"][1..-1]
+    status = env['PATH_INFO'][1..-1]
     render_error(status)
   end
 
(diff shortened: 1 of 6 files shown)


services/QuillLMS/app/controllers/application_controller.rb
```diff
@@ -69,13 +68,6 @@ def render_error(status)
     end
   end
 
-  def setup_visitor
```
happythenewsad, line 72: This code has been un-executable for 6 years. Removing.

services/QuillLMS/spec/requests/secure_headers_spec.rb
```diff
@@ -0,0 +1,26 @@
...
+  end 
+
+  after do 
+    Rails.application.reload_routes!
+  end
+
+  it 'should flag all cookies as HttpOnly' do 
+    get '/dummy'
+    expect(response.header['Cache-Control']).to match('no-cache, no-store, max-age=0, must-revalidate')
+    expect(response.header['Pragma']).to match('no-cache')
```
happythenewsad, line 29: Note: I'm not testing the Expires header here because it does not do anything. I'm including it in the code for CB compliance only.

From Mozilla:
"If there is a Cache-Control header with the max-age or s-maxage directive in the response, the Expires header is ignored."

brendanshean (approved):
Looks good to me!  Nice work on this one.

dandrabik (approved):
Looks good!

It will be worth monitoring our response times (both the API and browser load times which are both in New Relic) to make sure this doesn't have a negative impact on performance that we are missing, e.g. for some reason, some assets are being served by Rails and using this header, etc.

PR #8027, opened 2021-08-03
Add synced_name attr to classroom table

brendanshean:
## WHAT
Add a new string attribute `synced_name` to the `classrooms` table.

## WHY
Teachers have the ability to rename a course on Quill but sometimes we'd like to keep track of the original name of the classroom if it's synced with Google Classroom or Clever.  This field will store the original 'synced_name' of the classroom.

## HOW
Add a migration file.

### Screenshots

### Notion Card Links
Preparation for: https://www.notion.so/quill/Google-Classroom-and-Clever-Standardize-and-improve-how-we-handle-renaming-of-classes-that-are-sync-fb81d68ded994f4cb76db9758f379704

PR Checklist | Your Answer
------------ | -------------
Have you added and/or updated tests? |  No.  Just a new column in database.

### DIFF
--- a/services/QuillLMS/db/structure.sql
+++ b/services/QuillLMS/db/structure.sql
@@ -1131,7 +1131,7 @@ CREATE TABLE public.classrooms (
     visible boolean DEFAULT true NOT NULL,
     google_classroom_id bigint,
     grade_level integer,
-    source character varying
+    synced_name character varying
 );
 
 
--- a/services/QuillLMS/db/structure.sql
+++ b/services/QuillLMS/db/structure.sql
@@ -938,7 +938,7 @@ CREATE TABLE public.change_logs (
     action character varying NOT NULL,
     changed_record_id integer,
     changed_record_type character varying NOT NULL,
-    user_id integer NOT NULL,
+    user_id integer,
     created_at timestamp without time zone,
     updated_at timestamp without time zone,
     changed_attribute character varying,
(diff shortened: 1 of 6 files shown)


happythenewsad (commented):
How about something like `internal_name`? `source` can mean a bunch of things - the source school/district, etc.

brendanshean:
> How about something like `internal_name`? `source` can mean a bunch of things - the source school/district, etc.

While the spec calls for the word source in the UI, I agree that it's not the most informative term for an attribute.   What about `synced_name`?

services/QuillLMS/db/structure.sql
```diff
@@ -938,7 +938,7 @@ CREATE TABLE public.change_logs (
     action character varying NOT NULL,
```
cissyxyu, line 938: There are some changes that are not included in your PR. Can we ensure that the schema file is consistent?

brendanshean: @cissyxyu Good question.  The [diff](https://github.com/empirical-org/Empirical-Core/pull/8027/files#diff-f4525a8efa84d2b12114079828637b4260222ff089ba8488d02cf1890fe332c3R7391) shows [two](https://github.com/empirical-org/Empirical-Core/blob/develop/services/QuillLMS/db/migrate/20210712213724_allow_null_on_change_log_user.rb) [migrations](https://github.com/empirical-org/Empirical-Core/blob/develop/services/QuillLMS/db/migrate/20210722144950_add_highlight_prompt_to_passages.comprehension.rb) that were not checked into `db/structure.sql`.  This PR is just pulling those previous migrations in.

happythenewsad (approved):

PR #8031, opened 2021-08-03
ci/public-progress-report-intermittent-failure

anathomical:
## WHAT
Update test conditions to not assume order of responses
## WHY
Because the underlying function can be arbitrary in its return order
## HOW
Test to ensure that the `classroom_id` values we care about are in the response payload, and then confirm that they are associated with the expected `classroom_unit_id`s values

### Notion Card Links
https://www.notion.so/quill/Fix-intermittent-CircleCI-failure-bb314a2c8f734bc28ea3612d72a18609

PR Checklist | Your Answer
------------ | -------------
Have you deployed to Staging? | No.  CI fix only


brendanshean (approved):
Thank you for fixing this one!

cissyxyu (approved):
LGTM!

PR #8044, opened 2021-08-05
Persist Clever classroom name

brendanshean:
## WHAT
Store the classroom name found on the Clever server in the corresponding classroom attr `synced_name` so that we can differentiate between it and the `name` on the Quill server.

## WHY
Some classrooms are synced with classroom data from the external service Clever.   Currently, the name of the classroom on Clever overrides any changes to the name made on the Quill server.  We'd like teachers to be able to name classrooms custom names that are not overridden.

## HOW
The existing infrastructure uses the [method](https://github.com/empirical-org/Empirical-Core/pull/8044/files#diff-fb6700011735c0c55dad0b4c5b7c7eee6129930db44a90b40b6514b62bc04b8eL3): `CleverIntegration::Creators::Classrooms.run` to import Clever classroom information.  Within that [module](https://github.com/empirical-org/Empirical-Core/pull/8044/files#diff-fb6700011735c0c55dad0b4c5b7c7eee6129930db44a90b40b6514b62bc04b8eL11) is a `find_or_initialize_by` which collapses two paths into one.  However, the specs for this story involve some involved logic for updating, so I've refactored the module `.run` pattern, to a few classes:  `Importer`, `Creator`, `Updater`.  The `Updater` requires some complex logic whereas the `Creator` is a simple creation process.

NB:  While auditing this integration, I removed dead code related to `setup_from_clever` found in these files:
`services/QuillLMS/spec/models/user_spec.rb`
`services/QuillLMS/app/models/user.rb`
`services/QuillLMS/app/models/classroom.rb`

### Screenshots

### Notion Card Links
https://www.notion.so/quill/Google-Classroom-and-Clever-Persist-Clever-Classroom-name-746b7eca2b444c9ab2eeb5b7b8301285

PR Checklist | Your Answer
------------ | -------------


happythenewsad (approved):
You really left the code better than you found it. Here's a taco 🌮

cissyxyu (approved):
Wow! Looks really good!

services/QuillLMS/spec/services/clever_integration/classrooms/importer_spec.rb
```diff
@@ -0,0 +1,31 @@
...
+
+  subject { described_class.new(data) }
+
+  context 'classroom exists with clever_id' do
+    let(:clever_id) { '123_abc' }
+
+    before { create(:classroom, clever_id: clever_id) }
+
+    it 'runs classroom updater' do
+      expect { subject.run }.to_not change(Classroom, :count)
```
anathomical: Philosophical question: do we also want to set expectations around calls?  We could actually `expect` a call to Updater.run or Creator.run depending on which case we're looking at, but is that too focused on implementation rather than outcome (which the current tests are checking for already)?

Either way, we should almost certainly test something other than count because this particular case passes if nothing happens.  So we aren't actually confirming that update runs, just that creator doesn't.

services/QuillLMS/spec/services/clever_integration/classrooms/updater_spec.rb
```diff
@@ -0,0 +1,78 @@
...
+      name: name,
+      synced_name: synced_name
+    )
+  end
+
+  let(:data) do
+    {
+      clever_id: clever_id,
+      name: data_name,
+      grade: '100'
```
anathomical, line 18: I'm amused by the idea that `grade` doesn't have some sort of validation to make this impossible to use.

anathomical (changes requested):
So this looks good to me with a note about test cases in the `Importer`.  I think this could be merged as-is, but I wanted to raise a bit of discussion.  So feel free to just convince me if you disagree.

services/QuillLMS/spec/services/clever_integration/classrooms/importer_spec.rb
brendanshean: That's a really good point.  I will restructure the expectations.

services/QuillLMS/spec/services/clever_integration/classrooms/updater_spec.rb
brendanshean: I was surprised too given https://github.com/empirical-org/Empirical-Core/blob/develop/services/QuillLMS/app/models/classroom.rb#L28.

anathomical (approved):

PR #8054, opened 2021-08-05
Persist GoogleClassroom classroom name

brendanshean:
## WHAT
Store the classroom name found on the Clever server in the corresponding classroom attr `synced_name` so that we can differentiate between it and the `name` on the Quill server.

## WHY
Some classrooms are synced with classroom data from the external service Google Classroom.   Currently, the name of the classroom on Google Classroom is pulled only on initial import.  Subsequent changes made to the classroom name on the Google Classroom servers are not propagated to the LMS database; we'd like those changes persisted.

## HOW
The current teacher flow for importing google classrooms is via the ImportGoogleClassroomsModal.  This modal provides a listing of GoogleClassrooms associated with a given teacher via a call to [retrieve_google_classrooms](https://github.com/empirical-org/Empirical-Core/blob/develop/services/QuillLMS/client/app/bundles/Teacher/components/classrooms/active_classrooms.tsx#L110).  That particular method pulls from a cached value.  The problem is that these values are only pulled up until the first import.  

Before the modal loads, teacher google classroom data is pulled in via: `services/QuillLMS/app/controllers/teachers/classroom_manager_controller.rb#retrieve_google_classrooms`:
which retrieves/repopulates a cached serialized classrooms data for a particular user.  However, this data pull for import is only done for classrooms that have not already been imported.  To remedy this, a [background job](https://github.com/empirical-org/Empirical-Core/pull/8054/files#diff-b4fc947f4990b4690c30af105832cbefe2059ea85f188ba38b88ed5cee850f37R45) is now called once after google login that updates the already imported classrooms name data when necessary.

Similar to the corresponding Clever [PR](https://github.com/empirical-org/Empirical-Core/pull/8044) there's quite a bit of refactoring here, moving towards a more object-oriented design.

1) `GoogleIntegration::Users::ClassroomsCache` this is cache which is used for serialized google classrooms.
2) `GoogleIntegration::Users::ClassroomsData` is an [enumerable](https://blog.appsignal.com/2018/05/29/ruby-magic-enumerable-and-enumerator.html) intermediate object between the google classroom data pull and the eventual persistence into the LMS database.  There's also a `GoogleIntegration::Users::ClassroomDataAdapter` used to convert to the data pull to `ClassroomsData`.
3) `GoogleIntegration::Users:::ClassroomsRetriever` extracts functionality from `RetrieveGoogleClassroomsWorker` into a service object.
4) `GoogleIntegration::Classrooms:{Importer, Creator, Updater}` uses a similar pattern to the one found in the analogous Clever [PR](https://github.com/empirical-org/Empirical-Core/pull/8044)

NB: Dead code `services/QuillLMS/app/services/google_integration/classroom/teacher.rb`

### Screenshots

### Notion Card Links
https://www.notion.so/quill/Google-Classroom-and-Clever-Persist-Google-Classroom-name-82df023d569343a4afe73261b871e252

PR Checklist | Your Answer
------------ | -------------


happythenewsad (approved):
Nice work

services/QuillLMS/app/services/google_integration/users/classrooms_cache.rb
```diff
@@ -0,0 +1,28 @@
+module GoogleIntegration
+  module Users
+    class ClassroomsCache
+      SERIALIZED_GOOGLE_CLASSROOMS_FOR_ = 'SERIALIZED_GOOGLE_CLASSROOMS_FOR_'.freeze
+      SERIALIZED_GOOGLE_CLASSROOMS_CACHE_LIFE = 60
```
dandrabik: Will there be issues if this cache expires mid-job? Some of the current import jobs for edge-case large classrooms take over 60 seconds. If that is a concern, we might consider making this larger just to be safe, e.g. 5 minutes.

[NR Transaction link](https://one.newrelic.com/launcher/nr1-core.explorer?platform[timeRange][duration]=259200000&pane=eyJ0cmFuc2FjdGlvbklkIjoiNWIyMjRmNzQ2ODY1NzI1NDcyNjE2ZTczNjE2Mzc0Njk2ZjZlMmY1MzY5NjQ2NTZiNjk3MTRhNmY2MjJmNTI2NTc0NzI2OTY1NzY2NTQ3NmY2ZjY3NmM2NTQzNmM2MTczNzM3MjZmNmY2ZDczNTc2ZjcyNmI2NTcyMmY3MDY1NzI2NjZmNzI2ZDIyMmMyMjIyNWQiLCJ0eXBlIjoiU2lkZWtpcUpvYiIsIm5lcmRsZXRJZCI6ImFwbS1uZXJkbGV0cy50cmFuc2FjdGlvbnMiLCJlbnRpdHlHdWlkIjoiTWpZek9URXhNM3hCVUUxOFFWQlFURWxEUVZSSlQwNThOVFE0T0RVMk9EYzEifQ==&sidebars[0]=eyJuZXJkbGV0SWQiOiJucjEtY29yZS5hY3Rpb25zIiwiZW50aXR5R3VpZCI6Ik1qWXpPVEV4TTN4QlVFMThRVkJRVEVsRFFWUkpUMDU4TlRRNE9EVTJPRGMxIiwic2VsZWN0ZWROZXJkbGV0Ijp7Im5lcmRsZXRJZCI6ImFwbS1uZXJkbGV0cy50cmFuc2FjdGlvbnMifX0=&state=6544edd5-75fa-76ae-4273-46730283fdbb) for reference.

services/QuillLMS/app/services/google_integration/classrooms/updater.rb
```diff
@@ -0,0 +1,46 @@
...
+      private def classroom
+        @classroom ||= ::Classroom.unscoped.find_by!(google_classroom_id: google_classroom_id, teacher_id: teacher_id)
+      end
+
+      private def custom_name?
+        classroom.name != classroom.synced_name
+      end
+
+      private def grade
+        data.key?(:grade) ? data[:grade] : classroom.grade
```
dandrabik: Consider using fetch here (you can add a default value):
```suggestion
        data.fetch(:grade, classroom.grade)
```

services/QuillLMS/app/services/google_integration/classrooms/creator.rb
```diff
@@ -0,0 +1,51 @@
...
+        @teacher_id = data[:teacher_id]
+      end
+
+      def run
+        create_classrooms_teacher
+        classroom
+      end
+
+      private def classroom
+        @classroom ||= ::Classroom.create!(
```
dandrabik, line 17: This may be in a later PR, but we might consider creating the `Classroom` and `ClassroomTeacher` at the same time as nested_attributes. This would prevent classrooms from being created when the classroom teacher fails.

Issue documented in [this ticket](https://www.notion.so/quill/ea241f31a84c4b79ad56e83e99f7ee93?v=9f7aead86cc749d8a6f963070cef4f5e&p=1be9032fb94e4af48a1afa3e8156804d).

services/QuillLMS/app/services/google_integration/users/classroom_data_adapter.rb
```diff
@@ -0,0 +1,36 @@
+module GoogleIntegration
+  module Users
+    class ClassroomDataAdapter
+      attr_accessor :user, :data
```
dandrabik: I think these could both be `attr_reader` given the way they are used here (the hash updates don't use `data=`)?

services/QuillLMS/app/services/google_integration/classrooms/creator.rb
dandrabik: Also, should these be `find_or_create`s to make these more idempotent, e.g. if the background job fails at some point, will these create duplicate data?

services/QuillLMS/app/services/google_integration/users/classroom_data_adapter.rb
```diff
@@ -0,0 +1,36 @@
...
+
+      def initialize(user, data)
+        @user = user
+        @data = data
+      end
+
+      def run
```
dandrabik, line 11: You might consider removing the mutation here, i.e. the sole use of `run` here is to mutate the `data` attribute that you are passing in the initializer. So `data` will be very different depending on whether this has been `run` or not, which may not be obvious or known to a person using this object. 

You could have a memoizable attribute called `cleaned_data` or `parsed_data` (or something, even `run`). This would remove the mutation.

```ruby
def parsed_data
  @parsed_data ||= do
     ...generate the correct data hash
   end
end

File: services/QuillLMS/app/services/google_integration/users/classrooms_retriever.rb [hunk 6]
```diff
@@ -0,0 +1,50 @@
...
+        NewRelic::Agent.notice_error(e)
+        e.message
+      end
+
+      private def cache_classrooms_data
+        GoogleIntegration::Users::ClassroomsCache.set(user_id, data.to_json)
+      end
+
+      private def data
+        google_response == 'UNAUTHENTICATED' ? { errors: google_response } : { classrooms: google_response }
```
dandrabik: Maybe make this a constant in case we reuse?

dandrabik (approved):
Nice work on this. :+1:

I left a bunch of suggestions, but they are all of the form "Have you considered...". I'd read through them and make any changes that you think are appropriate (and ignore what you do not).

services/QuillLMS/app/services/google_integration/users/classrooms_cache.rb
brendanshean: This is a good point.  I will update.

services/QuillLMS/app/services/google_integration/classrooms/updater.rb
brendanshean: I was unaware of the default parameter.  That's much cleaner.

services/QuillLMS/app/services/google_integration/users/classroom_data_adapter.rb
brendanshean: I will try to make this more transparent.

services/QuillLMS/app/services/google_integration/classrooms/creator.rb
brendanshean: I've updated the PR to use nested_attributes for the Classroom and ClassroomTeacher creation.

services/QuillLMS/app/services/google_integration/users/classroom_data_adapter.rb
brendanshean: @dandrabik Does this work

dandrabik: Yeah, this works. Was there anything else that was supposed to carry through from `data`, i.e. this starts with a blank hash instead of the data hash (is this supposed to be `data.tap`)?

brendanshean: I was debating on using `data.tap` but opted for a clean hash so we are more transparent about the adapter.

brendanshean: On second thought.  I think I see what you mean.

anathomical (approved):

PR #8058, opened 2021-08-06
Tweak Increment Responses on CMS to update timestamp on increment

cissyxyu:
## WHAT
We had a bug where Responses were being incremented in the database directly without changing their `updated_at` timestamp. Therefore the nightly `UpdateIndividualResponse` worker that reindexes all of that day's updated responses wasn't catching these responses, so they weren't being re-indexed.

## WHY
When responses are not reindexed, they get out of date in ElasticSearch and they are returned in the wrong order on the admin dashboard.

## HOW
Send a `touch: true` attribute to the `increment!` call, which tells SQL to update the `updated_at` column along with the increment update.

### Screenshots

### Notion Card Links
https://www.notion.so/quill/Can-t-sort-responses-by-submissions-in-Connect-or-Diagnostic-ae550c8f61824001917072d400854d91

PR Checklist | Your Answer
------------ | -------------

### DIFF
--- a/services/QuillCMS/app/workers/create_or_increment_response_worker.rb
+++ b/services/QuillCMS/app/workers/create_or_increment_response_worker.rb
@@ -28,13 +28,13 @@ class CreateOrIncrementResponseWorker
   end
 
   def increment_counts(response, symbolized_vals)
-    response.increment!(:count)
+    response.increment!(:count, 1, touch: true)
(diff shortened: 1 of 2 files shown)


anathomical (approved):
I like the solution here.  Minimal changes required.  We still need a way to backfill the things that have fallen out of date, but that can be a different PR.

happythenewsad (approved):

PR #8061, opened 2021-08-09
CMS - Add GradedResponses

dandrabik:
## WHAT
In an effort to improve CMS performance, this PR creates a new model called `GradedResponses` which is backed by a materialized view (not a table) of graded responses set up using the `scenic` gem  that will be refreshed once a day. This will power the `responses_for_question` and `multiple_choice_options` endpoints

Note, since there needs to be testing of this set-up with `production` data, this PR doesn't make any of the endpoints 'live'. If testing goes well, that would be the next step.
## WHY
Right now, these are slow endpoints that need to be cached heavily in order to function. I originally was going to move these to the LMS, but based on feedback from the [RFC](https://www.notion.so/quill/RFC-Move-CMS-activity-endpoints-to-the-LMS-733daebd1dae4a51a3263bba1f900079), some research, and self-reflection, I think it's worth trying a more lightweight version of this in the CMS for now.
## HOW
Uses a materialized view generated using the [scenic gem](https://github.com/scenic-views/scenic). The view acts exactly like a regular active_record table. Both endpoints are powered by the `GradedResponse` model for now.

### Notion Card Links
https://www.notion.so/quill/RFC-Move-CMS-activity-endpoints-to-the-LMS-733daebd1dae4a51a3263bba1f900079

PR Checklist | Your Answer
------------ | -------------

### DIFF
--- a/services/QuillCMS/Gemfile
+++ b/services/QuillCMS/Gemfile
@@ -18,6 +18,7 @@ gem 'dotenv-rails', '~> 2.7'
 gem 'activerecord-import'
 gem 'bulk_insert'
 gem 'pg', '~> 1.2'
+gem 'scenic'
 
 # QUEUE/CACHING
 gem 'connection_pool'
--- a/services/QuillCMS/Gemfile
+++ b/services/QuillCMS/Gemfile
@@ -18,7 +18,7 @@ gem 'dotenv-rails', '~> 2.7'
 gem 'activerecord-import'
 gem 'bulk_insert'
 gem 'pg', '~> 1.2'
-gem 'scenic'
+gem 'scenic', '~> 1.5.4'
 
 # QUEUE/CACHING
 gem 'connection_pool'
--- a/services/QuillCMS/db/views/graded_responses_v01.sql
+++ b/services/QuillCMS/db/views/graded_responses_v01.sql
@@ -1,0 +1,4 @@
+-- Note, if you add or remove columns for the tables, you'll need a migration
+SELECT *
+FROM responses
+WHERE optimal IS NOT NULL;
--- a/services/QuillCMS/db/views/graded_responses_v01.sql
+++ b/services/QuillCMS/db/views/graded_responses_v01.sql
@@ -1,4 +1,6 @@
 -- Note, if you add or remove columns for the tables, you'll need a migration
+-- To update this view use `rails generate scenic:view graded_responses`
+-- See documentation here: https://github.com/scenic-views/scenic#cool-but-what-if-i-need-to-change-that-view
 SELECT *
 FROM responses
 WHERE optimal IS NOT NULL;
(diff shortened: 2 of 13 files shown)


services/QuillCMS/Gemfile
```diff
@@ -18,6 +18,7 @@ gem 'dotenv-rails', '~> 2.7'
 gem 'activerecord-import'
 gem 'bulk_insert'
 gem 'pg', '~> 1.2'
+gem 'scenic'
```
happythenewsad: Can we pin this to `~> 1.5.4`? This will make re-bundling easier in the future.

services/QuillCMS/db/views/graded_responses_v01.sql
```diff
@@ -0,0 +1,4 @@
+-- Note, if you add or remove columns for the tables, you'll need a migration
```
happythenewsad, line 1: Scenic has [very specific commands ](https://github.com/scenic-views/scenic)for updating existing views, especially long running ones. How about linking to them here, for future readers?

happythenewsad (changes requested):
Change requests are minor and/or informational. Looks good

services/QuillCMS/Gemfile
happythenewsad: Another comment that was not added to my review: there is one GradedResponse spec that is failing

anathomical (approved):

services/QuillCMS/spec/controllers/questions_controller_spec.rb
```diff
@@ -0,0 +1,104 @@
...
+      let!(:response_nonoptimal1) {create(:response, question_uid: '123', optimal: false)}
+      let!(:response_nonoptimal2) {create(:response, question_uid: '123', optimal: false)}
+      let!(:response_nonoptimal3) {create(:response, question_uid: '123', optimal: false)}
+      let!(:response_ungraded) {create(:response, question_uid: '123', optimal: nil)}
+
+      before(:each) do
+        GradedResponse.refresh
+      end
+
+      it 'should return graded responses, 2 optimal, 2 nonoptimal' do
```
brendanshean, line 62: It might be worth noting that `2`  is coupled to  `QuestionsController::MULTIPLE_CHOICE_LIMIT`

brendanshean (approved):
Looks great.  Specs were extremely readable and thorough.

services/QuillCMS/spec/controllers/questions_controller_spec.rb
dandrabik: Yeah, good idea. Putting as a comment for now, since I think leaving the numbers makes the test easier to parse.

services/QuillCMS/db/views/graded_responses_v01.sql
dandrabik: Good point! I've added comments.

services/QuillCMS/Gemfile
dandrabik: Good suggestion, pinned!

happythenewsad (approved):
👍

PR #8064, opened 2021-08-10
content security policy

happythenewsad:
## WHAT
Adds a Content-Security-Policy header to LMS requests, a way of mitigating XSS vulnerabilities. 

## WHY
This is the suggested remediation to a medium severity security concern raised by CB.

## HOW
Configured via the `secure_headers` initializer. 

## DISCUSSION
- CSP adds extra maintenance and brittleness to our app, but will most likely reach wide adoption and become an expected practice. The likelihood of a bug caused by CSP rules reaching production is reasonably low, since they should be caught locally (Chrome and Brave have easy CSP rule violation highlighting within the console).
- The CSP spec is a work in progress. CSP Level 2 is implemented by most browsers, while [Level 3 is not](https://developers.google.com/web/fundamentals/security/csp). 
- Testing this change is difficult, because it touches every part of the code and only applies after DOM load. Here's my current testing plan (on staging). 
1. Visit homepage
2. Assign an activity as a demo teacher.
3. Play through 1 Connect question as a demo student.
4. Load Comprehension activities editor as a staffer. 
5. Login with Google
6. Login with Clever
7. A Lessons activity
8. For each item above, inspect console and network browser tab and ensure that no resources are forbidden by the Content Security Policy.

## Known requests checklist
- [x] Intercom (support chat window)
- [x] Coview (Intercom plugin to share browser info)
- [x] Sentry (frontend error reporting)
- [x] Clever Login (third-party login)
- [x] Google Login (third-party login - may already be covered in the google urls)

## Follow-on work
- CB suggests injecting nonces into all served scripts. These nonces are generated fresh by every request, and if there is a mismatch between the server's nonce and the script's nonce, the script will not be loaded. Rails 5.2 supports CSP nonces, while Rails 5.1 does not. I recommend waiting until 5.2, rather than rolling our own nonce solution (with/without `secure_headers`).
- this configuration is relaxed with respect to allowing the use of `eval()`. Work to remove `eval()` from the codebase is tracked [here](https://www.notion.so/quill/Product-Board-30f97e4eb01246dbb8264253fb385073?p=9e8c0edce35940949e15aec5b9810bd6)

## Helpful CSP references
- "official" CSP spec: https://www.w3.org/TR/CSP2/
- more readable: https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Content-Security-Policy
- https://csp.withgoogle.com/docs/adopting-csp.html
- https://developers.google.com/web/fundamentals/security/csp

### Notion Card Links
https://www.notion.so/quill/Missing-Content-Security-Policy-Header-a285458ffd204229ad0999ba7118da8c

PR Checklist | Your Answer
------------ | -------------
Have you added and/or updated tests? |  Yes, manually (see 'Discussion')


services/QuillLMS/config/initializers/secure_headers.rb
```diff
@@ -1,5 +1,55 @@
 SecureHeaders::Configuration.default do |config|
-  config.csp = SecureHeaders::OPT_OUT
+  config.csp = {
+    default_src: [
+      "'self'",
```
happythenewsad, line 4: This double+single quote syntax is intentional. Consult one of the linked reference sites for more info.

services/QuillLMS/config/initializers/secure_headers.rb
```diff
@@ -1,5 +1,68 @@
...
+  config.csp = {
+    default_src: [
+      "'self'", 
+      "https://*.quill.org",
+      "'unsafe-inline'"                                           # TODO: remove once nonce strategy is in place
+    ],                                                            # fallback for more specific directives
+
+    object_src: %w('none'),                                       # addresses <embed>, <object>, and <applet>
+
+    script_src: [
```
dandrabik, line 11: So reviewing this is a bit tricky, but there are some frontend services that we use that aren't in this list (maybe they don't need to be, based on how they are used).

However, to be safe, it would be worth confirming if the following need to be added to one of these safelists:
- Intercom (support chat window)
- Coview (Intercom plugin to share browser info)
- Sentry (frontend error reporting)
- Clever Login (third-party login)
- Google Login (third-party login - may already be covered in the google urls)

dandrabik (changes requested):
Thanks for putting this together. The explanation links were very helpful.

I'm 'requesting changes' since I listed a few services we should confirm we don't need to add to this list before we deploy. 

Also, it is probably worth having @emilia-friedberg or @eadams17 look this over to see if there are any other services that we might be missing.

emilia-friedberg (commented):
I can't think of services other than what you have here and the ones Dan listed, but I think to your testing plan we should add:

a) Clever log-in
b) Google log-in
c) a Quill Lessons activity (previewing one is probably fine for this purpose)

services/QuillLMS/config/initializers/secure_headers.rb
happythenewsad: I added these items as a checklist at the top of the page.

anathomical (approved):
The approach looks good.  I also can't think of additional services that might need to be added here, but agree with Emilia that testing as close to real-world circumstances as we can is going to be the thing to do here.

dandrabik (approved):
:+1:  Nice work on this. Thanks for making these additions/changes and doing the additional testing.  

Let's deploy this outside of peak traffic times, just in case. So some time after 5pm ET on a weekday would work.

PR #8067, opened 2021-08-11
rename comprehension to evidence everywhere user-facing

emilia-friedberg:
## WHAT
Rename "Comprehension" to "Evidence" everywhere user-facing, in the public urls, and in the code where it would have been more confusing to leave as is.

## WHY
We are moving forward with Quill Evidence as the name for now.

## HOW
Changed the name and key locally, then did a global search to find places that needed to be updated either because they explicitly used the wrong name or because they relied on the old value.

### Screenshots

### Notion Card Links
https://www.notion.so/quill/Rename-Comprehension-to-Quill-Evidence-ab7ce4bdfeb741aa9cff5ada100b83a3

PR Checklist | Your Answer
------------ | -------------


eadams17 (approved):
the evidence is clear, this is good to go ✨

services/QuillLMS/app/assets/stylesheets/shared/icons.scss
```diff
@@ -166,7 +166,7 @@
   background-position: center;
   height: 40px;
 }
-.icon-comprehension-green-no-border, .icon-9-green-no-border {
+.icon-evidence-green-no-border, .icon-9-green-no-border {
```
anathomical, line 169: Appreciate clean-up even in things like class-names here.

services/QuillLMS/app/controllers/application_controller.rb
```diff
@@ -8,6 +8,7 @@ class ApplicationController < ActionController::Base
   GOOGLE_OR_CLEVER_JUST_SET = :google_or_clever_just_set
   KEEP_ME_SIGNED_IN = :keep_me_signed_in
   COMPREHENSION = 'comprehension'
+  EVIDENCE = 'evidence'
```
anathomical, line 11: Do we need to remove line 10 here with the `COMPREHENSION` definition?

services/QuillLMS/app/controllers/pages_controller.rb
```diff
@@ -467,7 +467,7 @@ def backpack
     @style_file = 'staff'
   end
 
-  def comprehension
+  def evidence
     allow_iframe
     @style_file = ApplicationController::COMPREHENSION
```
anathomical, line 472: Does this need to be updated?  It looks like the answer is probably "no" (same thing on line 524 below), but I wanted to double check.

anathomical (approved):
Looks good to me.  I assume we also need to actually update the values in the `ActivityClassification` table.  Should we include a rake task for that here, or just update them by hand around the time this deploys?

services/QuillLMS/app/controllers/application_controller.rb
emilia-friedberg: Good question - no, this has to stay until we change the webpack definition for the file names, which I think we are going to do as part of a code-name change at a later date.

services/QuillLMS/app/controllers/pages_controller.rb
emilia-friedberg: Same as above!

PR #8076, opened 2021-08-12
Teacher center/ beta access updates

eadams17:
## WHAT
add comprehension resources for beta users to teacher center

## WHY
we want only beta users to have access to these resources

## HOW
add logic for checking if app setting is enabled for conditional rendering

### Screenshots
Screen Shot 2021-08-12 at 2 08 21 PM
Screen Shot 2021-08-12 at 2 08 29 PM

### Notion Card Links
https://www.notion.so/quill/Restrict-non-beta-users-from-accessing-teacher-center-articles-f54c34d37e0d44d19dd1b48c65f46d2c

PR Checklist | Your Answer
------------ | -------------


emilia-friedberg (approved):
This looks good to me. I would maybe just double check that we still want to call this section "reading comprehension" in light of the fact that we're calling the tool Quill Evidence.

services/QuillLMS/app/views/navbar/_teacher_center.html.erb
```diff
@@ -10,6 +10,16 @@
       link: '/teacher-center',
       text: 'All Resources'
     %>
+    <% if current_user && AppSetting.enabled?(name: 'comprehension', user: current_user) %>
```
emilia-friedberg: It probably makes sense to make `'comprehension'` a const in the `AppSetting` model file.

services/QuillLMS/app/views/blog_posts/index.html.erb
```diff
@@ -1,4 +1,14 @@
...
+  <%= react_component('BlogPostsApp', props: {
+    route: 'index',
+    blogPosts: @blog_posts,
+    announcement: @announcement,
+    query: @query,
+    topics: @topics,
+    studentTopics: @student_topics,
+    role: current_user&.role,
+    title: @title,
+    isComprehensionUser: current_user && AppSetting.enabled?(name: 'comprehension', user: current_user)
```
emilia-friedberg: See below point about the `comprehension` const.

services/QuillLMS/client/app/bundles/Teacher/components/blog_posts/header_section.jsx
```diff
@@ -1,17 +1,30 @@
...
-    <div className="text">
-      <h1>{props.title}</h1>
-      <p>{props.subtitle}</p>
+export const HeaderSection = ({ title, subtitle, showCancelSearchButton, query }) => {
+
+  function renderTitle() {
+    if(title === USING_QUILL_FOR_READING_COMPREHENSION) {
+      const words = title.split(" ");
+      const quill = words[1].charAt(0).toUpperCase() + words[1].slice(1);
+      return words[0] + ' ' + quill + ' ' + words.slice(2).join(" ");
```
emilia-friedberg: It might be easiest just to return a string literal? Otherwise, I would just do `title.replace('quill', 'Quill')`.

services/QuillLMS/app/views/blog_posts/index.html.erb
eadams17: @emilia-friedberg good call. Peter signed off on this last night though so I think we're using reading comprehension here more as the concept rather than the tool name

services/QuillLMS/client/app/bundles/Teacher/components/blog_posts/header_section.jsx
eadams17: totally forgot you could do that! great suggestion, thank you

cissyxyu (approved):

PR #8078, opened 2021-08-12
CMS - Multiple Choice Responses View

dandrabik:
## WHAT
A follow-on PR to #8061. Add an additional materialized view to power the `multiple_choice_options` endpoint.
## WHY
After testing production data, it's not possible to power the `multiple_choice_options` from just `GradedResponse` since there are not two nonoptimal-graded responses for every question.
## HOW
Abstract some of the code used for `GradedResponse` and add another materialized view migration using `scenic`.

**UPDATE**
Additionally, to have this query finish on production, I needed to scope it to responses that have a `count > 10`. I've added some logic in the controller endpoint to fallback on other responses if a question doesn't have enough responses that fit this setup. So unfortunately, the DB constraints made the code a touch more complicated.

PR Checklist | Your Answer
------------ | -------------
Self-Review: Have you done an initial self-review of the code below on Github? | Yup
Spec Review: Have you reviewed the spec and ensured this meets requirements and/or matches design mockups? | Yes.


services/QuillCMS/app/workers/refresh_responses_view_worker.rb
```diff
@@ -1,15 +1,21 @@
-class RefreshGradedResponsesWorker
+class RefreshResponsesViewWorker
   include Sidekiq::Worker
   sidekiq_options queue: SidekiqQueue::LOW
 
   REFRESH_TIMEOUT = ENV["VIEW_STATEMENT_TIMEOUT"] || '10min'
```
brendanshean: Would fetch apply here:  `ENV.fetch("VIEW_STATEMENT_TIMEOUT", '10min')`

services/QuillCMS/db/migrate/20210812215158_create_multiple_choice_responses.rb
```diff
@@ -0,0 +1,28 @@
...
+
+    add_index :multiple_choice_responses, :id, unique: true
+    add_index :multiple_choice_responses, :question_uid
+    add_index :multiple_choice_responses, :optimal
+  ensure
+    db_set_timeout(original_timeout)
+  end
+
+  def down
+    drop_view :multiple_choice_responses, materialized: true
```
brendanshean, line 20: do the three `add_index` methods need corresponding `drop_index` in the `down` step?

services/QuillCMS/app/workers/refresh_responses_view_worker.rb
```diff
@@ -1,15 +1,21 @@
...
+  VIEWS = %w(GradedResponse MultipleChoiceResponse)
+
+  class InvalidView < StandardError; end
 
   # Make the DB statement timeout longer while refreshing the materialized view
-  def perform
+  def perform(view)
     original_timeout = db_timeout
```
brendanshean, line 12: Is there ever a case where we don't want to extend the timeout on a materialized view?  In other words, should the `db_timeout` wrapper be in the concern `refresh` method (as opposed to requiring any future workers to make the call)?

I also noticed another configuration via the `establish_connection` [here](https://github.com/scenic-views/scenic/issues/201#issuecomment-434993888)

services/QuillCMS/spec/models/graded_response_spec.rb
```diff
@@ -2,9 +2,9 @@
 
 RSpec.describe GradedResponse, type: :model do
   context 'basic queries' do
-    let!(:response_optimal) {create(:response, question_uid: '123', optimal: true)}
-    let!(:response_nonoptimal) {create(:response, question_uid: '123', optimal: false)}
-    let!(:response_ungraded) {create(:response, question_uid: '123', optimal: nil)}
+    let!(:optimal) {create(:response, question_uid: '123', optimal: true)}
```
brendanshean: Maybe time for factory aliases for optimal_response, non_optimal_response, etc.

brendanshean (approved):
Looks good to me.   Some comments you can take or leave.

services/QuillCMS/db/migrate/20210812215158_create_multiple_choice_responses.rb
dandrabik: Good question! In my testing, using only `drop_view` doesn't error, but you're right, it's probably better to add the drop_indexes to ensure they are cleaned up.

services/QuillCMS/spec/models/graded_response_spec.rb
dandrabik: Hmm, maybe. I'm a bit on the fence here, since being explicit inline is about the same length as the abstraction (and some of the names can be confusing, i.e. distinguishing between `optimal` `false` and `nil`, which many times are treated the same). Let me think about this.

```ruby
create(:optimal_response)
create(:response, optimal: true)

Comment (brendanshean) [reply services/QuillCMS/spec/models/graded_response_spec.rb]: While length is about the same,  `create(:optimal_response)` prevents the accidentally specifying wrong value for `optimal`.  It also makes the `let(:optimal) { create(:optimal_response) }` easier to read.

Comment (brendanshean) [reply services/QuillCMS/spec/models/graded_response_spec.rb]: One other point is that it provides an opportunity to standardize language around what's being created.

**Review by happythenewsad — APPROVED** [review]

**Review by anathomical — APPROVED** [review]

File: services/QuillCMS/app/workers/refresh_responses_view_worker.rb (hunk 1 above)
Comment (dandrabik) [reply services/QuillCMS/app/workers/refresh_responses_view_worker.rb]: Yup, good call, will replace.

File: services/QuillCMS/app/workers/refresh_responses_view_worker.rb (hunk 3 above)
Comment (dandrabik) [reply services/QuillCMS/app/workers/refresh_responses_view_worker.rb:12]: To be honest, I went back and forth on where to keep the wrapper. I kept it in the job since we don't need the wrapper for `test` and `development` environments, it's only needed on `production` and `staging`.  The test suite has to call `refresh` relatively often, so I thought this made tests slightly less complicated. I could be convinced otherwise.

Also, I saw the `establish_connection` method, but that's for every call by the model, even queries, so I'd rather keep this scoped to refresh.

Comment (brendanshean) [reply services/QuillCMS/app/workers/refresh_responses_view_worker.rb:12]: Those are good reasons to keep it where you have it, especially the point about queries.

File: services/QuillCMS/spec/models/graded_response_spec.rb (hunk 4 above)
Comment (dandrabik) [reply services/QuillCMS/spec/models/graded_response_spec.rb]: Ok, you've made several good points, I'll add the aliases. Thanks for the feedback on this.

**Comment by dandrabik** [comment]
I added an update after testing the queries a bit. Re-opening this for review:



> UPDATE
> Additionally, to have this query finish on production, I needed to scope it to responses that have a count > 10. I've added some logic in the controller endpoint to fallback on other responses if a question doesn't have enough responses that fit this setup. So unfortunately, the DB constraints made the code a touch more complicated.

File: services/QuillCMS/spec/controllers/questions_controller_spec.rb [hunk 5]
```diff
@@ -116,5 +118,34 @@
         expect(null_optimal_count).to eq 0
       end
     end
+
+    context 'fallback responses needed' do
+      let!(:nonoptimal) {create(:graded_nonoptimal_response, question_uid: '123', count: 17)}
```
brendanshean: I would rename
`nonoptimal` -> `graded_nonoptimal`

services/QuillCMS/spec/controllers/questions_controller_spec.rb
```diff
@@ -116,5 +118,34 @@
...
+      # Note, this expectation is bound to QuestionsController::MULTIPLE_CHOICE_LIMIT
+      it 'should return graded responses, 2 nonoptimal' do
+        get :multiple_choice_options, params: {question_uid: '123'}
+
+        expect(response.status).to eq 200
+
+        json = JSON.parse(response.body)
+        optimal_count = json.count {|gr| gr['optimal']}
+        nonoptimal_count = json.count {|gr| gr['optimal'] == false}
+        null_optimal_count = json.count {|gr| gr['optimal'].nil?}
```
brendanshean: I would rename:
`nonoptimal_count` -> `graded_nonoptimal_count`
`null_optimal_count` -> `ungraded_count`

services/QuillCMS/spec/controllers/questions_controller_spec.rb
```diff
@@ -14,9 +14,9 @@
     end
 
     context 'with data' do
-      let!(:response_optimal) {create(:response, question_uid: '123', optimal: true)}
-      let!(:response_nonoptimal) {create(:response, question_uid: '123', optimal: false)}
-      let!(:response_ungraded) {create(:response, question_uid: '123', optimal: nil)}
+      let!(:optimal) {create(:optimal_response, question_uid: '123')}
+      let!(:nonoptimal) {create(:graded_nonoptimal_response, question_uid: '123')}
```
brendanshean: I would rename 
`nonoptimal` ->`graded_nonoptimal`

services/QuillCMS/spec/controllers/questions_controller_spec.rb
```diff
@@ -46,35 +46,36 @@
...
-      let!(:optimal2) {create(:response, question_uid: '123', optimal: true, count: 7)}
-      let!(:optimal3) {create(:response, question_uid: '123', optimal: true, count: 9)}
+      let!(:optimal1) {create(:optimal_response, question_uid: '123', count: 5)}
+      let!(:optimal2) {create(:optimal_response, question_uid: '123', count: 7)}
+      let!(:optimal3) {create(:optimal_response, question_uid: '123', count: 9)}
 
-      let!(:nonoptimal1) {create(:response, question_uid: '123', optimal: false, count: 7)}
-      let!(:nonoptimal2) {create(:response, question_uid: '123', optimal: false, count: 5)}
-      let!(:nonoptimal3) {create(:response, question_uid: '123', optimal: false, count: 9)}
+      let!(:nonoptimal1) {create(:graded_nonoptimal_response, question_uid: '123', count: 17)}
```
brendanshean: I would rename 
`nonoptimal1` -> `graded_nonoptimal1`
`nonoptimal2` -> `graded_nonoptimal2`
`nonoptimal3` -> `graded_nonoptimal3`

services/QuillCMS/spec/controllers/questions_controller_spec.rb
```diff
@@ -46,35 +46,36 @@
...
-      it 'should return graded responses, 2 optimal, 2 nonoptimal' do
+      it 'should return graded responses, 2 optimal, 2 from nonoptimal/ungraded' do
         get :multiple_choice_options, params: {question_uid: '123'}
 
         expect(response.status).to eq 200
 
         json = JSON.parse(response.body)
         optimal_count = json.count {|gr| gr['optimal']}
-        nonoptimal_count = json.count {|gr| !gr['optimal']}
+        nonoptimal_count = json.count {|gr| gr['optimal'] == false }
```
brendanshean: I would rename:
`nonoptimal_count` -> `graded_nonoptimal_count` 
`null_optimal_count` -> `ungraded_count`

services/QuillCMS/spec/models/graded_response_spec.rb
```diff
@@ -2,9 +2,9 @@
 
 RSpec.describe GradedResponse, type: :model do
   context 'basic queries' do
-    let!(:response_optimal) {create(:response, question_uid: '123', optimal: true)}
-    let!(:response_nonoptimal) {create(:response, question_uid: '123', optimal: false)}
-    let!(:response_ungraded) {create(:response, question_uid: '123', optimal: nil)}
+    let!(:optimal) {create(:optimal_response, question_uid: '123')}
+    let!(:nonoptimal) {create(:graded_nonoptimal_response, question_uid: '123')}
```
brendanshean: Rename:
`nonoptimal` -> `graded_non_optimal`

services/QuillCMS/spec/models/graded_response_spec.rb
```diff
@@ -17,10 +17,12 @@
...
       GradedResponse.refresh
-      graded_responses = GradedResponse.where(question_uid: '123').sort_by(&:id)
+      responses = GradedResponse.where(question_uid: '123').sort_by(&:id)
+      response_ids = responses.map(&:id).sort
+      graded_ids = [optimal.id, nonoptimal.id].sort
 
-      expect(graded_responses.first.id).to be response_optimal.id
-      expect(graded_responses.second.id).to be response_nonoptimal.id
+      expect(response_ids).to eq graded_ids
+      expect(responses.first.attributes.keys.sort).to eq nonoptimal.attributes.keys.sort
```
brendanshean: I'm not I understand the purpose of this expectation.

services/QuillCMS/app/controllers/questions_controller.rb
```diff
@@ -6,11 +6,37 @@ def responses
...
+      .to_a
+
+    nonoptimal = nonoptimal_main.concat(nonoptimal_fallback(nonoptimal_main.count))
+
+    render json: optimal.concat(nonoptimal)
+  end
+
+  # The MultipleChoiceResponse model contains responses
+  # that have a count > MultipleChoiceResponse::MIN_COUNT
+  # All questions, like new ones, may may not have enough responses,
```
brendanshean: typo with "may may"

services/QuillCMS/spec/models/multiple_choice_response_spec.rb
```diff
@@ -0,0 +1,54 @@
+require 'rails_helper'
+
+RSpec.describe MultipleChoiceResponse, type: :model do
+  context 'basic queries' do
+    let!(:ungraded) {create(:ungraded_response, question_uid: '123', count: 15)}
+    let!(:nonoptimal) {create(:graded_nonoptimal_response, question_uid: '123', count: 19)}
+    let!(:ungraded2) {create(:ungraded_response, question_uid: '123', count: 17)}
+    let!(:optimal) {create(:optimal_response, question_uid: '123', count: 9999)}
+
+
```
brendanshean: whitespace

services/QuillCMS/spec/models/multiple_choice_response_spec.rb
```diff
@@ -0,0 +1,54 @@
+require 'rails_helper'
+
+RSpec.describe MultipleChoiceResponse, type: :model do
+  context 'basic queries' do
+    let!(:ungraded) {create(:ungraded_response, question_uid: '123', count: 15)}
+    let!(:nonoptimal) {create(:graded_nonoptimal_response, question_uid: '123', count: 19)}
```
brendanshean: I would rename 
`nonoptimal` -> `graded_nonoptimal`

services/QuillCMS/spec/models/multiple_choice_response_spec.rb
```diff
@@ -0,0 +1,54 @@
...
+      response_ids = responses.map(&:id).sort
+      highest_two_count_ids = [nonoptimal.id, ungraded2.id].sort
+
+      expect(response_ids).to eq highest_two_count_ids
+      expect(responses.first.attributes.keys.sort).to eq nonoptimal.attributes.keys.sort
+    end
+  end
+
+  context 'rank ties' do
+    let!(:ungraded) {create(:ungraded_response, question_uid: '123', count: 15)}
```
brendanshean: Since the variables in 33-36 aren't used in the spec itself, you could throw them in a `before` clause.

services/QuillCMS/spec/models/multiple_choice_response_spec.rb
```diff
@@ -0,0 +1,54 @@
...
+
+    it 'should return additional records when there is a tie for 2nd rank' do
+      # return 3 returns for counts 9,5,5
+      MultipleChoiceResponse.refresh
+      expect(MultipleChoiceResponse.count).to be 3
+    end
+  end
+
+  context 'records with less than 10 count' do
+    let!(:nonoptimal_less_than10) {create(:graded_nonoptimal_response, question_uid: '123', count: 10)}
```
brendanshean: Since the variables in 46-47 aren't used in the spec itself, you could throw them in a `before` clause.

brendanshean (approved):
I think this is really close.  Great work here, this has been fun to see the development of this materialized view.

Most of the comments are on the redefining the scope [nonoptimal](https://github.com/empirical-org/Empirical-Core/pull/8078/commits/53e1f97fc1927254bcd09f274c24ab911245eb96#diff-2e42f05e854c93cd8b96d485035062fef859b6fc9bb5e38345add73874c8fd5fR10).  Since `nonoptimal` now covers two states: `false` and `nil`, renaming  `let(:nonoptimal)` to `let(:graded_nonoptimal)` would be more informative.  I'm not suggesting a similar renaming for `ungraded` to `ungraded_nonoptimal` because that would be redundant.

brendanshean:
@dandrabik This [PR](https://github.com/empirical-org/Empirical-Core/pull/8089) should fix those failing specs.

services/QuillCMS/spec/models/graded_response_spec.rb
dandrabik: This is confirmation that the `Response` model json and the `GradedResponse` model json have the exact same keys. If an attribute in the `Response` model changes, say we add a column via a migration, this test would fail, marking that we need to update the `GradedResponse` model, via and equivalent migration.

dandrabik: Also, this may not be obvious, but the columns/attributes of the materialized view are locked after the view migration, i.e. you can't do `responses.*` in the query and expect new columns to come in when the original table is migrated. You also have to migrate the view.

https://github.com/scenic-views/scenic#faqs

dandrabik:
> I think this is really close. Great work here, this has been fun to see the development of this materialized view.
> 
> Most of the comments are on the redefining the scope [nonoptimal](https://github.com/empirical-org/Empirical-Core/pull/8078/commits/53e1f97fc1927254bcd09f274c24ab911245eb96#diff-2e42f05e854c93cd8b96d485035062fef859b6fc9bb5e38345add73874c8fd5fR10). Since `nonoptimal` now covers two states: `false` and `nil`, renaming `let(:nonoptimal)` to `let(:graded_nonoptimal)` would be more informative. I'm not suggesting a similar renaming for `ungraded` to `ungraded_nonoptimal` because that would be redundant.

Fair points about making the naming more explicit, I've made those changes. Thanks for taking a close look at this!

services/QuillCMS/spec/models/graded_response_spec.rb
brendanshean: Thanks for the explanation.  I was unaware of that.   I think it's worth documenting that in its own spec:

```
it 'should have the same keys as Responses table' do
  GradedResponse.refresh
  responses = GradedResponse.where(question_uid: '123').sort_by(&:id)
  expect(responses.first.attributes.keys.sort).to eq graded_nonoptimal.attributes.keys.sort
end
```
and then add  same check for `multiple_choice_response_spec.rb` too?

PR #8086, opened 2021-08-13
Prevent default submission behavior on Comprehension Turk buttons

cissyxyu:
## WHAT
The "copy", "edit" and "delete" buttons on the admin Comprehension Turk Sessions page were submitting by default and redirecting the user to a 404 page because the submission would take them to a broken link. This fixes that by preventing the default submission behavior.

## WHY
We don't want admin to go to a broken page when they click these links.

## HOW
Add a line to prevent the default event behavior on button click.

### Screenshots

### Notion Card Links
https://www.notion.so/quill/Broken-buttons-in-Comprehension-internal-tool-a5a55e8a87c14a538f797fb5050ac4e3

PR Checklist | Your Answer
------------ | -------------
Have you added and/or updated tests? |  NO, tiny change

### DIFF
--- a/services/QuillLMS/client/app/bundles/Staff/components/comprehension/__tests__/__snapshots__/turkSessions.test.tsx.snap
+++ b/services/QuillLMS/client/app/bundles/Staff/components/comprehension/__tests__/__snapshots__/turkSessions.test.tsx.snap
@@ -1169,7 +1169,7 @@ exports[`TurkSessions component should render TurkSessions 1`] = `
               clickHandler={[Function]}
               id={1}
               label="copy"
-              value="undefined/comprehension/#/turk?uid=17&id=1"
+              value="undefined/evidence/#/turk?uid=17&id=1"
             />,
             "delete": <Unknown
               clickHandler={[Function]}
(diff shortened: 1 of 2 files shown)


emilia-friedberg (changes requested):
Since this is an admin page, it isn't a huge deal, but in general, we don't want to be explicitly giving buttons `type="submit"` and then using the actual click handler to say "actually, don't treat this like a submit button", both for accessibility reasons and also because it's confusing to specify one behavior in one place and then change it in another. I think a better option here would be to change the `type` in the `TurkSessionButton` file to `button`, since this is the only place I think we use it. If we did sometimes want it to be `type="submit"`, we could pass in the type as a prop.

cissyxyu:
> Since this is an admin page, it isn't a huge deal, but in general, we don't want to be explicitly giving buttons `type="submit"` and then using the actual click handler to say "actually, don't treat this like a submit button", both for accessibility reasons and also because it's confusing to specify one behavior in one place and then change it in another. I think a better option here would be to change the `type` in the `TurkSessionButton` file to `button`, since this is the only place I think we use it. If we did sometimes want it to be `type="submit"`, we could pass in the type as a prop.

So after giving this another look, I think I was on the wrong track with `e.preventDefault()`. We thought the button was redirecting improperly because of the button type, but `type="submit"` is actually not the issue. The code to provide the `ReactTable` row and cell is providing that button with a `link` attribute which makes the button redirect to an improper link. That link is not used or necessary, so I just removed it and the whole thing is working fine now!

eadams17 (approved):
Looks good! You'll just want to also remove the link key in the headers array too, and probably adjust the widths of the other columns

cissyxyu:
> Looks good! You'll just want to also remove the link key in the headers array too, and probably adjust the widths of the other columns

My bad! On looking over it AGAIN I realized that the link is actually fine and needed, but it was just wrongly triggering the data-table component's code because the `DataTable` component interprets `link` as a redirect link for the entire table row. It's just a matter of mixed up semantics. 

I changed "link" to "turkLink" in the turk component so it doesn't wrongly trigger any weird behavior, and this time it worked.

emilia-friedberg (approved):

PR #8087, opened 2021-08-13
Fix bug involving google imports of classes with existing with already existing names

brendanshean:
## WHAT
Fix a bug involving a teacher importing of google classrooms where there is a naming collision between an imported classroom and existing one owned by the teacher.

## WHY
During renaming of classrooms within the Quill UI, users will be warned dynamically when trying to update names to invalid ones.  
Screen Shot 2021-08-13 at 3 26 40 PM
This validation also runs when teachers import from Google Classroom, however, when validation fails, the error message does not propagate to the UI and the import cycles indefinitely:
Screen Shot 2021-08-13 at 3 41 15 PM

## HOW
Simply append `_1` to the existing name if there is a collision.  In the unlikely event that rename collides with another owned classroom, the logic provides a [loop](https://github.com/empirical-org/Empirical-Core/pull/8087/files#diff-dd8e7d46eefefd3f89a7bb8ab5aa3c563de8e7b43ee0053ed1d96fca27275efaR59) that eventually settles on a temporary name.

### Screenshots

### Notion Card Links
https://www.notion.so/quill/Sentry-Error-ActiveRecord-RecordInvalid-Teachers-ClassroomManagerController-update_google_classr-2b522bc57318430397f0272fdf0b743b

PR Checklist | Your Answer
------------ | -------------

### DIFF
--- a/services/QuillLMS/app/services/google_integration/classrooms/creator.rb
+++ b/services/QuillLMS/app/services/google_integration/classrooms/creator.rb
@@ -21,7 +21,7 @@ module GoogleIntegration
           teacher_id: teacher_id,
           classrooms_teachers_attributes: [
             {
-              user_id: user_id,
+              user_id: teacher_id,
               role: role
             }
           ]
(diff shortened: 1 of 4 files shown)


cissyxyu (approved):
Looks good!

PR #8101, opened 2021-08-17
proofreader bug

happythenewsad:
## WHAT
Fixes controller params permissioning broken by Rails 5.0, so that `passage` data is persisted to the database.

## WHY
So that students can resume editing their work after hitting `Save & Exit` or reloading.

## DISCUSSION
Thomas's related PR: https://github.com/empirical-org/Empirical-Core/pull/8102

### Screenshots
(If applicable. Also, please censor any sensitive data)

### Notion Card Links
https://www.notion.so/quill/b417e5b89fe04af89cba62d4198f7031?v=77f294a5b0df466e98633244de7ebfef&p=4622f65cdd714be08cf71f6f7fecdf3c

PR Checklist | Your Answer
------------ | -------------
Have you added and/or updated tests? |  tested manually. Creating a followup card for testing (we were not able to create an effective spec at this time)
Have you deployed to Staging? | not yet
Self-Review: Have you done an initial self-review of the code below on Github? | yes
Spec Review: Have you reviewed the spec and ensured this meets requirements and/or matches design mockups? | yes

PR #8102, opened 2021-08-17
proofreader/no-override-with-empty-passage

anathomical:
## WHAT
Do not clobber Passages with empty arrays from Active Sessions
## WHY
We've found a bug that mistakenly saves empty arrays to the `passage` key of ActiveActivitySession for Proofreader activities, and that empty value is clobbering the actual intended `passage` value rendering activities unplayable on reload
## HOW
Ignore the `passage` key from `ActiveActivitySession` if it is an empty array (which can be caused by a bug we are working on fixing now).  That lets the activity play through as expected, even if it doesn't persist.

### Notion Card Links
https://www.notion.so/quill/Quill-Proofreader-passage-not-loading-for-students-4622f65cdd714be08cf71f6f7fecdf3c

PR Checklist | Your Answer
------------ | -------------
Have you added and/or updated tests? |  No tests on this very specific edge case on the front-end


brendanshean (approved):

emilia-friedberg (approved):

dandrabik (approved):
Approved. I believe this is the change that Emilia suggested: https://github.com/empirical-org/Empirical-Core/pull/8101/files#r690465463

PR #8105, opened 2021-08-17
Evidence/ add caption and attribution for passage images

eadams17:
## WHAT
add functionality to add caption and attribution properties for Evidence passage images
 
## WHY
we want to have captions for users and we need to have attribution for CC legal reasons

## HOW
add two new inputs on the internal tool, refactor Evidence student play to display caption and image credit, add warning modal for opening new link

### Screenshots
Screen Shot 2021-08-17 at 12 35 15 PM
Screen Shot 2021-08-17 at 12 38 44 PM
Screen Shot 2021-08-17 at 12 38 52 PM

### Notion Card Links
https://www.notion.so/quill/Caption-and-attribution-feature-5ac5e5dcbc3441c686dff5f47efbe46b

PR Checklist | Your Answer
------------ | -------------


cissyxyu (approved):

services/QuillLMS/client/app/bundles/Shared/components/shared/tooltip.tsx
```diff
@@ -25,22 +29,29 @@ export class Tooltip extends React.Component<TooltipProps, {}> {
...
   hideTooltipOnClick(e) {
+    const { setNavigationUrl, toggleWarningModal } = this.props;
     if (this.tooltip && e.target !== this.tooltip && e.target !== this.tooltipTrigger) {
       this.hideTooltip()
     }
+    if(e.target.href) {
+      e.preventDefault();
+      setNavigationUrl(e.target.href)
+      toggleWarningModal();
+    }
```
emilia-friedberg: This logic is too specific for this shared component to know about - we can refactor this so you're just passing an `onClick` handler from the parent which then does all this logic for you, instead of passing in these props that are hyperspecific to this one implementation.

services/QuillLMS/client/app/bundles/Shared/components/shared/tooltip.tsx
```diff
@@ -1,6 +1,10 @@
 import * as React from 'react'
+import marked from 'marked';
 
 interface TooltipProps {
+  isMarkdown?: boolean,
```
emilia-friedberg: I think instead of handling the conversion from Markdown in this file, it would be better to do the conversion in the parent and pass it in as a prop, since we don't really have reason to think we're going to want to be doing this all that frequently.

services/QuillLMS/client/package.json
```diff
@@ -89,7 +90,7 @@
     "react-dropzone": "^4.2.8",
     "react-html-parser": "^2.0.2",
     "react-i18next": "^8.4.0",
-    "react-markdown": "^2.4.6",
+    "react-markdown": "^2.5.1",
```
emilia-friedberg, line 93: If we already have `react-markdown` in here we probably don't also need `marked`, right?

services/QuillLMS/client/app/bundles/Comprehension/components/studentView/container.tsx
```diff
@@ -61,6 +63,8 @@ interface StudentViewContainerState {
   explanationSlidesCompleted: boolean;
   explanationSlideStep:  number;
   completedSteps: Array<number>;
+  imageAttributionModalOpen: boolean;
+  imageUrlToVisit: string;
```
emilia-friedberg: As far as I can tell, the container component doesn't need to know about any of this at all, and this file is already really bloated. I think we need to pull out all the image-related code into its own component and file and you can handle all of this in there.

emilia-friedberg (changes requested):
Interface looks great, but there are some refactors I think we need to make to keep our files clean. Let me know if you have questions!

emilia-friedberg (approved):
Looks great! Thanks for making those changes!

PR #8108, opened 2021-08-17
lms/fix-activity-count-race-condition

anathomical:
## WHAT
Wrap activity count in a retry block to avoid race conditions
## WHY
I don't know how this happens, but sometimes a user triggers `ActivitySession.update` twice in parallel with both payloads trying to mark the session as `complete`.  When this happens, if the user doesn't already have a `UserActivityClassification` record for the activity classification of the referenced activity, there can be a race condition about which thread gets to create the record for the first time.
## HOW
Just wrap the block that does the creation in a `begin` block and retry in cases where the race condition is detected.

### Notion Card Links
https://www.notion.so/quill/Race-Condition-in-User-completed-activity-counts-8c234b1100a048399d6819c47cbbd5fa

PR Checklist | Your Answer
------------ | -------------
Have you added and/or updated tests? |  No.  I don't actually know of a good way to simulate a race condition.
Have you deployed to Staging? | No - tiny change

### DIFF
--- a/services/QuillLMS/app/controllers/api/v1/activity_sessions_controller.rb
+++ b/services/QuillLMS/app/controllers/api/v1/activity_sessions_controller.rb
@@ -101,8 +101,12 @@ class Api::V1::ActivitySessionsController < Api::ApiController
   private def count_completed_activity
     return unless @activity_session.finished?
 
-    counter = UserActivityClassification.find_or_create_by(user: @activity_session.user, activity_classification: @activity_session.classification)
-    counter.increment_count
+    begin
+      counter = UserActivityClassification.find_or_create_by(user: @activity_session.user, activity_classification: @activity_session.classification)
+      counter.increment_count
+    rescue ActiveRecord::RecordNotUnique
+      retry
+    end
   end
 
   private def activity_session_params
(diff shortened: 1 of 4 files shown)


brendanshean (commented):
Per the referenced [article](https://blog.saeloun.com/2019/02/23/rails-6-adds-create_or_find_by.html#overcoming-duplicate-inserts), should the `find_or_create_by` be wrapped in a transaction:

```ruby
begin
  UserActivityClassification.transaction(requires_new: true) do
    counter = UserActivityClassification.find_or_create_by(user: ...
    counter.increment_count
  end
rescue
  retry
end
```

Also, the Rails 6 method `create_or_find_by` method uses a similar [pattern](https://github.com/rails/rails/blob/6f519dc9440b883f1d88af5ae01ec84dc11aeafc/activerecord/lib/active_record/relation.rb#L216)

happythenewsad:
> I don't actually know of a good way to simulate a race condition.

You could try using [Timecop](https://github.com/travisjeffery/timecop) in RSpec

brendanshean (approved):
I forgot to suggest this [arkency](https://blog.arkency.com/2015/09/testing-race-conditions/) post but it looks like you already found it.  Great work here!

happythenewsad (approved):

PR #8123, opened 2021-08-18
CMS - Enable new response and multiple_choice_options endpoints

dandrabik:
## WHAT
Testing of the new endpoints powered by views looks good on `production`, so putting them in place.
## WHY
These are much more performant than the current endpoints, without even using caching.
## HOW
Removing old routes, controller endpoints, and tests.  Since these are materialized views that need to be refreshed, queueing a job to refresh each view over night.

PR Checklist | Your Answer
------------ | -------------

### DIFF
--- a/services/QuillCMS/app/models/cron.rb
+++ b/services/QuillCMS/app/models/cron.rb
@@ -24,7 +24,7 @@ class Cron
     timeout = RefreshResponsesViewWorker::REFRESH_TIMEOUT
     default_interval = 600 # 10 minutes
 
-    interval = minute_string.in?(timeout) ? (timeout.gsub(minute_string,'').to_i * 60) : default_inteveral
+    interval = minute_string.in?(timeout) ? (timeout.gsub(minute_string,'').to_i * 60) : default_interval
 
     RefreshResponsesViewWorker::VIEWS.each.with_index do |view, index|
       RefreshResponsesViewWorker.perform_in(index * interval, view)
--- a/services/QuillCMS/app/models/cron.rb
+++ b/services/QuillCMS/app/models/cron.rb
@@ -2,11 +2,32 @@ class Cron
 
   def self.run
     current_time = Time.zone.now
-    if current_time.hour == 3
-      beginning_of_yesterday = current_time.yesterday.beginning_of_day
-      end_of_yesterday = current_time.yesterday.end_of_day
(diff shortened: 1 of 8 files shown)


services/QuillCMS/app/models/cron.rb
```diff
@@ -2,11 +2,32 @@ class Cron
...
+    # default to 10 minutes
+    minute_string = 'min'
+    timeout = RefreshResponsesViewWorker::REFRESH_TIMEOUT
+    default_interval = 600 # 10 minutes
+
+    interval = minute_string.in?(timeout) ? (timeout.gsub(minute_string,'').to_i * 60) : default_interval
+
+    RefreshResponsesViewWorker::VIEWS.each.with_index do |view, index|
+      RefreshResponsesViewWorker.perform_in(index * interval, view)
     end
```
anathomical: This is a question about approach, not code, but do we want our `Cron` model to have a bunch of logic in it, or should it just be a registry of Workers to call?  I could imagine moving the logic in these private classes into a sort of "wrapper" worker that does this work which lets us (try) to keep `Cron` as just a list of jobs that get run and when they get run which is how I'd been thinking of it.  However, it's possible that it's more valuable to do this sort of set-up work directly on the cron run in order to avoid delays.  I'm open to discussion on this.

anathomical (changes requested):
Question about code organization that I think bears discussion before merging, but the code itself all looks good to me.

happythenewsad (approved):

services/QuillCMS/app/models/cron.rb
happythenewsad: I see your point, Thomas. I'm kinda on the fence on this one. This PR is injecting logic, but it's minimal and all related to timing the executions of the workers.

dandrabik: I don't have strong feelings either way. I don't mind moving the logic I added to a Worker (it's probably enough logic to justify it). The `elasticsearch` method has a little less logic, so those seem a little odd to move. I could just put that logic back in the main run class. 

I'll take a few minutes to refactor.

dandrabik: @anathomical The refactor is up, I think this is a little cleaner. Easy to move stuff around because of the tests.

anathomical (approved):

PR #8124, opened 2021-08-18
Evidence/remove csrf from apis

anathomical:
## WHAT
Consolidate the logic that allows Evidence API calls to be made without a CSRF token, and then remove the front-end code for Evidence applications that sort of hackily worked to attach CSRF tokens to requests now that they're not necessary.
## WHY
While the workaround worked, it's not really the best way to handle these situations, and added code that wasn't necessary to the front-end.  Removing it makes us behave more like we should regarding APIs and clients.
## HOW
- Create a new `ApiController` in Evidence that inherits from `ApplicationController` but always skips CSRF verification.  Then update all of our various API controllers to inherit from the intermediate `ApiController`.
- Remove the code that we'd put in the Evidence front-ends that sort of hackily extracted the CSRF token from the underlying Rails pages and attached them to API call headers since we no longer need it.

### Notion Card Links
https://www.notion.so/quill/Remove-CSRF-requirements-from-all-Comprehension-API-Endpoints-e10079c799df44b1a86f0eed58ef59ba

PR Checklist | Your Answer
------------ | -------------
Have you added and/or updated tests? |  No, this was basically just removal of some imports and `localStorage` calls


brendanshean (approved):
Looks good to me.

cissyxyu (approved):

eadams17 (approved):

PR #8129, opened 2021-08-20
Evidence/link activity title to lms

anathomical:
## WHAT
Add `after_save` hooks to both base `Activity` and `Comprehension::Activity` so that when you change a specific value in one model, it updates the corresponding value in the other.  Specifically, we are linking `Activity.name` and `Comprehension::Activity.title`.  This allows Curriculum developers to edit the value via either the Evidence admin interface OR the LMS Activity editing interface and not have to worry about updating the corresponding value in the other place.
## WHY
We were running into instances where Curriculum would update the `Comprehension::Activity.title` value in the Evidence admin tool, but forgetting to update the LMS value which resulted in students and teachers seeing one name when looking at their dashboards, and a different name when actually playing the activity.
## HOW
Use `after_save` hooks in both models to update the value in the associated parent/child model when the relevant values are changed.
Note, we also tweaked the way that we defined the `Comprehension::Activity.parent_activity` relationship so that it namespaces properly.  (I'll leave an in-line note in the PR with more detail.)

### Notion Card Links
https://www.notion.so/quill/Display-a-warning-when-changing-the-activity-title-in-the-Activity-Settings-c5eac6ea4eaa44aeb9e0e972d01a97db

PR Checklist | Your Answer
------------ | -------------

### DIFF
--- a/services/QuillLMS/app/models/activity.rb
+++ b/services/QuillLMS/app/models/activity.rb
@@ -60,6 +60,7 @@ class Activity < ApplicationRecord
   has_many :topics, through: :activity_topics
   before_create :flag_as_beta, unless: :flags?
   after_commit :clear_activity_search_cache
+  after_save :update_evidence_child_title, if: :update_evidence_title?
 
   delegate :form_url, to: :classification, prefix: true
 
@@ -219,6 +220,19 @@ class Activity < ApplicationRecord
 
   def is_evidence?
     classification&.key == ActivityClassification::EVIDENCE_KEY
+  end
+
+  def child_activity
+    return unless is_evidence?
+    Comprehension::Activity.find_by(parent_activity_id: id)
+  end
+
+  private def update_evidence_title?
+    is_evidence? && name_changed?
+  end
+
+  private def update_evidence_child_title
+    child_activity&.update(title: name)
   end
 
   private def data_must_be_hash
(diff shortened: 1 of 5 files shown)


services/QuillLMS/app/models/activity.rb
```diff
@@ -221,6 +222,19 @@ def is_evidence?
     classification&.key == ActivityClassification::EVIDENCE_KEY
   end
 
+  def child_activity
```
brendanshean, line 225: Since this method's implementation is coupled to Evidence, would `evidence_child_activity` be more informative?  (This would also be consistent with the methods `update_evidence_title?` and `update_evidence_child_title`)

brendanshean (approved):
One comment you can take or leave.  I really love the thorough unit testing.

cissyxyu (approved):

services/QuillLMS/app/models/activity.rb
anathomical: That's a reasonable suggestion, but I actually wanted to leave this in its more generic form in case we later decide to add additional activities with relationships here.  I can imagine us adding new activity types, or modifying old ones, such that this would be a useful function to hook up to.  I'm not married to it, but it seemed worth leaving that possibility open.

PR #8140, opened 2021-08-23
Validate Turk Sessions before Turkers can play Comprehension activity

cissyxyu:
## WHAT
We don't want Turkers to be able to access old turk links that are 1) expired or 2) attached to Comprehension activities that have been archived.

So this PR adds a validation endpoint to check on these conditions. If the turk link is invalid, it redirects them to the Quill 404 page.

## WHY
Turkers should not be able to play activities that are in either of these invalid states.

## HOW
1. Add a back end API endpoint that checks if a turk link is expired.
2. Add a front end call to this endpoint when the turk session is rendered.
3. Redirect the user if the turk link is expired.

### Screenshots

### Notion Card Links
https://www.notion.so/quill/Write-up-404-message-for-archived-links-0934f77ee48946f9b8b907492abe7aa9

PR Checklist | Your Answer
------------ | -------------


services/QuillLMS/engines/comprehension/spec/dummy/app/models/activity.rb
```diff
@@ -1,2 +1,7 @@
 class Activity < ApplicationRecord
+
+  def flag flag = nil
```
brendanshean: Per style guide, make sure to wrap method arguments in parentheses: `def flag(flag = nil)`

services/QuillLMS/engines/comprehension/app/models/comprehension/activity.rb
```diff
@@ -15,7 +15,7 @@ class Activity < ApplicationRecord
     has_many :passages, inverse_of: :activity, dependent: :destroy
     has_many :prompts, inverse_of: :activity, dependent: :destroy
     has_many :turking_rounds, inverse_of: :activity
-    belongs_to :parent_activity, class_name: Comprehension.parent_activity_classname
```
brendanshean: We should keep the method call here over the hard-coded `::Activity`

brendanshean (commented):
I think of `GET` as requesting a resource while `POST` as creating a resource.  Would `GET` be more appropriate here?

services/QuillLMS/engines/comprehension/app/models/comprehension/activity.rb
cissyxyu: So while I generally agree that we should not hard code strings here, this wasn't working as expected when we had it as `Comprehension.parent_activity_classname`. Rails was interpreting it as the `Comprehension::Activity` module class instead of the parent class. I tried a bunch of configurations and the only one that would work was hard coding it as a string as I've done here.

cissyxyu:
> I think of `GET` as requesting a resource while `POST` as creating a resource. Would `GET` be more appropriate here?

Hmm I think of `GET` requests as requests that hit an endpoint to return something static, like static HTML content or data that will be the same every time. Because this request is sending a message body with multiple parameters, and the server has to parse those parameters and make a decision on what to send based on that, and then send differing data each time, I think of this more as a `POST` request.

https://stackoverflow.com/questions/978061/http-get-with-request-body

What do you think?

services/QuillLMS/engines/comprehension/app/models/comprehension/activity.rb
brendanshean: Can you write a spec that shows this resolution failure and I will take a looK?

brendanshean:
> > I think of `GET` as requesting a resource while `POST` as creating a resource. Would `GET` be more appropriate here?
> 
> Hmm I think of `GET` requests as requests that hit an endpoint to return something static, like static HTML content or data that will be the same every time. Because this request is sending a message body with multiple parameters, and the server has to parse those parameters and make a decision on what to send based on that, and then send differing data each time, I think of this more as a `POST` request.
> 
> https://stackoverflow.com/questions/978061/http-get-with-request-body
> 
> What do you think?

GET can return static HTML, content and data, but the returned doesn't have to be the same on every request.
For example, imagine we're making a GET request to a user endpoint:  `/users/34`.  Let's say that all last week, that link successfully returned user data.  Yesterday, however, the user closed their account.  Upon hitting that endpoint again, now we see a 404 (or some error).  The GET request is still idempotent, it's just that other business rules changed the state of User with id 34..  Also, think about GET requests to /index routes, those are certainly dynamic. 

As far as the request body is concerned, what about passing the activity_id and turking_round_id in the url itself either through routing `api/v1/comprehension/turking_round_activity_sessions/activities/5/turking_round/3.json` or query_string params:  `api/v1/comprehension/turking_round_activity_sessions/?activity_id=5&turking_round_id=3`

The semantics of POST are that the request is making a change to the system.  It's not clear to me what change is being made by this request.

cissyxyu:
> ed, what about passing the activity_id and turking_round_id in the url itself either through routing `api/v1/comprehension/turking_round_activity_sessions/activities/5/turking_round/3.json` or query_string params: `api/v1/comprehension/turking_round_activity_sessions/?activity_id=5&turking_round_id=3`

Gotcha, I think I understand it now -- hadn't thought of sending the parameters inside of the URL string.
I changed the API endpoint to a GET request!

emilia-friedberg (approved):

brendanshean (approved):
Looks good!

PR #8153, opened 2021-08-25
Updates to stabilize CMS Deploy Script

cissyxyu:
## WHAT
Remove the `git stash` and `git stash apply` components of the CMS Deploy Script. Instead, check the working tree for changes to existing files, and exit the deploy script if any changes exist. 
Also add a dateTime stamp to deploy labels so that multiple deploys of the same commit can go through.

## WHY
Having `git stash` and `git stash apply` in a deploy script is unconventional as it modifies the developer's working tree, which is not expected for a deploy script. It's not always necessary to run `git stash` and we had an outage where `git stash apply` was unnecessarily applied.

In the future, it makes more sense for a developer to resolve changed files on their own before deploying the CMS. This avoids any future complications where the git commands are run inadvertently and cause outages.

## HOW
Add a check before deploying to make sure the developer's working tree is clean (except for untracked files). 
Add a dateTime segment to the deploy label so that every deploy has a unique label name.

### Screenshots

### Notion Card Links
https://www.notion.so/quill/Post-mortem-2021-08-11-CMS-Merge-Conflict-Downtime-f83b5ed5f9dd45eda6bbd65bba312559
https://www.notion.so/quill/ea241f31a84c4b79ad56e83e99f7ee93?v=a9d41c7da6794f429933c6369ee1e230&p=7230aae8578946d89c64e14480a98eeb

PR Checklist | Your Answer
------------ | -------------
Have you added and/or updated tests? |  NO, tested manually.


services/QuillCMS/deploy.sh
```diff
@@ -29,23 +29,26 @@ if [[ "$response" =~ ^([y])$ ]]
...
+    if [[ -z $(git status --untracked-files=no -s) ]]
+    then
+      # update local production branch
+      # you would need to already be on production to get here
+      # but checking out production just in case
+      git fetch origin production
+      git checkout production
+      git pull origin production
+    else
+      echo "Your working tree is dirty, please stash or commit changes before deploying."
```
happythenewsad: Let's keep the code block but remove the if/else logic. When we checkout & pull production, there are two possible bad scenarios:
- normal uncommitted changes are present
- a merge conflict is present, which is a special form of uncommitted change

If we sequence the `git status` check after the checkout & pull, I believe all edge cases would be handled.

services/QuillCMS/deploy.sh
```diff
@@ -29,23 +29,26 @@ if [[ "$response" =~ ^([y])$ ]]
...
-    git fetch origin production
-    git checkout production
-    git pull origin production
-    git stash apply
+    if [[ -z $(git status --untracked-files=no -s) ]]
+    then
+      # update local production branch
+      # you would need to already be on production to get here
+      # but checking out production just in case
+      git fetch origin production
```
happythenewsad: You can remove this line because `pull` is a concatenation of the `fetch` and `merge` commands.

happythenewsad (changes requested):

services/QuillCMS/deploy.sh
```diff
@@ -29,23 +29,26 @@ if [[ "$response" =~ ^([y])$ ]]
...
   then
-    # stash local changes and update local production branch
-    # you would need to already be on production to get here
-    # but checking out production just in case
-    git stash
-    git fetch origin production
-    git checkout production
-    git pull origin production
-    git stash apply
+    if [[ -z $(git status --untracked-files=no -s) ]]
```
happythenewsad: Great job finding this flag, which makes the script much more elegant. This flag will probably catch the additional edge case I mention below, but it's worth checking.

brendanshean: Great use of the `--untracked-files` flag here.

happythenewsad (approved):
LGTM

brendanshean (approved):
Looks great!  Thank you for updating this.

anathomical (approved):

PR #8159, opened 2021-08-26
Add ProviderClassroomUser STI model, migration

brendanshean:
## WHAT
Add a STI join table `ProviderClassroomUser` along with subtypes `GoogleClassroomUser` and `CleverClassroomUser`

## WHY
In order to display what students are in a classroom on one of our third party providers, we need a model to track their status.  
More details in [RFC](https://www.notion.so/quill/RFC-Google-Classroom-and-Clever-Import-Object-291c61b3d05a46c3834c5621ef222a0e)

## HOW
Create the requisite migration and model files.  There's also a timestamp attribute `deleted_at` which will keep track of when a student is removed from a classroom on one of the providers.

### Notion Card Links
https://www.notion.so/quill/Add-ProviderClassroomUser-STI-model-migration-97991855dab6443cb2325df044ca406f

PR Checklist | Your Answer
------------ | -------------

### DIFF
--- a/services/QuillLMS/app/models/provider_classroom_user.rb
+++ b/services/QuillLMS/app/models/provider_classroom_user.rb
@@ -1,0 +1,42 @@
+# == Schema Information
+#
+# Table name: provider_classroom_users
+#
+#  id                    :bigint           not null, primary key
+#  deleted_at            :datetime
+#  type                  :string           not null
+#  created_at            :datetime         not null
+#  updated_at            :datetime         not null
+#  provider_classroom_id :string           not null
+#  provider_user_id      :string           not null
+#
+# Indexes
+#
+#  index_provider_classroom_users_on_provider_classroom_id    (provider_classroom_id)
+#  index_provider_classroom_users_on_provider_user_id         (provider_user_id)
+#  index_provider_user_id_and_provider_classroom_id_and_type  (provider_user_id,provider_classroom_id,type) UNIQUE
+#
+class ProviderClassroomUser < ApplicationRecord
+  scope :active, -> { where(deleted_at: nil) }
+  scope :deleted, -> { where.not(deleted_at: nil) }
+
+  ACTIVE = :active
+  DELETED = :deleted
+
+  def active?
+    deleted_at.nil?
+  end
+
+  def deleted?
+    !active?
+  end
+
+  def status
+    active? ? ACTIVE : DELETED
+  end
+
+  def soft_delete
+    update!(deleted_at: Time.now)
+  end
+end
+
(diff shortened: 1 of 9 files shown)


services/QuillLMS/app/models/provider_classroom_user.rb
```diff
@@ -0,0 +1,42 @@
...
+# Indexes
+#
+#  index_provider_classroom_users_on_provider_classroom_id    (provider_classroom_id)
+#  index_provider_classroom_users_on_provider_user_id         (provider_user_id)
+#  index_provider_user_id_and_provider_classroom_id_and_type  (provider_user_id,provider_classroom_id,type) UNIQUE
+#
+class ProviderClassroomUser < ApplicationRecord
+  scope :active, -> { where(deleted_at: nil) }
+  scope :deleted, -> { where.not(deleted_at: nil) }
+
```
dandrabik, line 25: You might consider an inclusion validation for types, e.g. `validates :type, inclusion: {in: TYPES}`. 

I'm also curious if there are any other validations we could put on the `provider_classroom_id` or `provider_user_id` like max length.

services/QuillLMS/db/migrate/20210824114552_create_provider_classroom_user.rb
```diff
@@ -0,0 +1,17 @@
...
+      t.string :type, null: false
+      t.string :provider_classroom_id, index: true, null: false
+      t.string :provider_user_id, index: true, null: false
+      t.timestamp :deleted_at
+
+      t.timestamps
+    end
+
+    add_index :provider_classroom_users,
+      [:provider_user_id, :provider_classroom_id, :type],
```
dandrabik: Flagging this to double check that this index order is optimal for common queries.

(As you likely know, order matters in compound indexes i.e. the index can only be used by a query if the left-most index column is in the query (`provider_user_id` here). So, for example, would it be common to query by just `type` and `provider_classroom_id`? If so, we might consider a different order (or different index setup).)

dandrabik (approved):
Overall, this looks like a good setup. I flagged a few things to look at, but looks good.

emilia-friedberg (approved):

services/QuillLMS/db/migrate/20210824114552_create_provider_classroom_user.rb
brendanshean: This is a good point.  Spoke with Dan about this offline and we agreed on just starting with a `[:type, :provider_classroom_id, :provider_user_id]` compound since `type` is almost always be used, and most queries currently use `:provider_classroom_id`.

services/QuillLMS/app/models/provider_classroom_user.rb
brendanshean: Good ideas.  I've added validations for these

dandrabik (approved):
:+1: Looks good to me.

PR #8174, opened 2021-08-31
app setting script improvements

happythenewsad:
## WHAT
- adds extra input validation AppSetting user id ingestion script
- accepts csv rather than flat file
- appends user allowlist instead of overwriting
- 'best effort' strategy allows for successful script completion when Users do not exist
- creates admin-protected detail page for app settings, so that non-engineer admins can view allowlists easily

## WHY
These pain points arose after Lindsey and I started coordinating on AppSetting user-based rollouts last week. 

## HOW
- Adds JSON API endpoint for AppSetting status that is accessible to all staff
- rake task edits

### Screenshots

### Notion Card Links
https://www.notion.so/quill/AppSetting-system-enhancements-23699c437b46421e8903d83e68de4693

PR Checklist | Your Answer
------------ | -------------
Have you added and/or updated tests? |  yes)
Have you deployed to Staging? |about to

### DIFF
--- a/services/QuillLMS/spec/controllers/api/v1/app_settings_controller_spec.rb
+++ b/services/QuillLMS/spec/controllers/api/v1/app_settings_controller_spec.rb
@@ -1,7 +1,7 @@
 require 'rails_helper'
 
 RSpec.describe Api::V1::AppSettingsController, type: :controller do
-  let(:user) { create(:user) }
+  let(:user) { create(:user, role: 'staff') }
 
   before { allow(controller).to receive(:current_user) { user } }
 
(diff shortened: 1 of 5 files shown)


services/QuillLMS/spec/controllers/api/v1/app_settings_controller_spec.rb
```diff
@@ -1,7 +1,7 @@
 require 'rails_helper'
 
 RSpec.describe Api::V1::AppSettingsController, type: :controller do
-  let(:user) { create(:user) }
+  let(:user) { create(:user, role: 'staff') }
```
brendanshean, line 4: Should `role: staff` need to be applied to all endpoints or just `#admin_show`?

brendanshean (approved):
A minor question in the spec, but looks good otherwise.

services/QuillLMS/spec/controllers/api/v1/app_settings_controller_spec.rb
happythenewsad: Yes, staff authentication only applies to #admin_show, since only this action has the potential to leak user emails.

cissyxyu (approved):

PR #8179, opened 2021-09-01
Evidence/rename comprehension engine

anathomical:
## WHAT
Rename the "Comprehension" engine to "Evidence", and update as many places as we can to refer to "evidence".
## WHY
So that we don't get confused later when everyone is used to talking about "Evidence" and can't find what we're looking for because it's still using the old terminology.
## HOW
- Moved everything in the engine from `/comprehension` subdirectories to `/evidence` subdirectories
- Replaced all references to the `Comprehension` module with `Evidence`
- Hard-coded table names so that Evidence models can still refer to the old "comprehension" tables (so we can avoid migrations)
- Updated the config for the Engine to call it `Evidence`
- Updated all references to "comprehension" in the Engine to be "evidence" EXCEPT in cases where the references were direct database table or column references
- Updated the LMS `ChangeLog` model so that it can reference `Evidence` rather than `Comprehension` values and models
- Updated all tests and factories

## PRE-DEPLOY CHECKLIST
- [x] Use `db:drop`, `db:create`, and `db:migrate` on the test environment, ensure all tests pass
- [x] Ensure that activities are playable by playing through one start to finish
- [x] Ensure that activities are editable by creating a new test one and modifying an existing one
- [x] Ensure that `FeedbackHistory` and `ChangeLog` reporting for the test activity in the previous step are accessible and complete
- [x] Deploy code to prod during off-hours (since we have a data migration that needs to be run)
- [x] Run rake task to complete data migraton

### Notion Card Links
https://www.notion.so/quill/Rename-Comprehension-Engine-Evidence-06227c425fe2481cabdc36c513564db4

PR Checklist | Your Answer
------------ | -------------


happythenewsad:
Have stakeholders outside of engineering been notified that this will impact analytics?

Example: https://github.com/empirical-org/Empirical-Core/pull/8179/files#diff-3a17b7ca82348d50db00521e1c0d875da3b2b44623117a6da6985685bc2dbf52R38

services/QuillLMS/app/models/change_log.rb
```diff
@@ -80,16 +80,16 @@ class ChangeLog < ApplicationRecord
...
-    'Comprehension::Prompt',
-    'Comprehension::Rule',
-    'Comprehension::Passage',
-    'Comprehension::AutomlModel',
-    'Comprehension::Label',
-    'Comprehension::Feedback',
-    'Comprehension::Highlight',
-    'Comprehension::RegexRule',
-    'Comprehension::PlagiarismText'
+    'Evidence::Activity',
```
dandrabik, line 83: This will require a data migration, correct? Will that be part of this PR?

services/QuillLMS/app/models/feedback_history.rb
```diff
@@ -29,7 +29,7 @@
 class FeedbackHistory < ApplicationRecord
   CONCEPT_UID_LENGTH = 22
   DEFAULT_PAGE_SIZE = 25
-  DEFAULT_PROMPT_TYPE = "Comprehension::Prompt"
+  DEFAULT_PROMPT_TYPE = "Evidence::Prompt"
```
dandrabik, line 32: Another data migration will be needed for this.

happythenewsad (approved):
Looks good; I respect the grind.

Since this will be a high impact deploy, how about adding a pre-deploy checklist? This could include:

- [ ] play through a Comprehension activity
- [ ] load a new test DB from schema and confirm specs pass
- [ ] run migrations from first comprehension migration and confirm specs pass
- [ ] TBD

dandrabik (changes requested):
So overall, looks like you hit the places that need to be renamed (without having to migrate the engine tables yet). Thanks for working through all of this!

I'm 'requesting changes' since there are two tables that I believe need a data migration for this to work properly (changing references stored in DB columns from `Comprehension::` to `Evidence::` ):
- change_logs
- feedback_history

(If I'm wrong on this, or there is a separate PR for this, let me know, but I think those references will cause errors if not updated).

services/QuillLMS/app/models/change_log.rb
anathomical: Good catch on these.  I've added a `rake` task to handle namespaces in both `ChangeLog` and `FeedbackHistory` polymorphic references.

anathomical:
@happythenewsad I liked your suggestion of a checklist, and I've added one to the main PR description if you want to review and provide feedback.

dandrabik (approved):
Thanks for adding the migration. I like the idea of a migration checklist and deploying off-peak for this one (given there's a reasonable chance we missed something).

PR #8187, opened 2021-09-01
Lms/update deprecations

anathomical:
## WHAT
Went through the test suite and updated all the places where we were getting warnings about deprecation and went ahead and updated the code to the most recent syntax.
## WHY
To ease later upgrades so that upgrading future versions of Rails doesn't break a bunch of stuff (some of these deprecated behaviors are intended to do the opposite of what they do now in the next version).  And to declutter my console when running local tests.
## HOW
Most of these updates fell into one of two categories:
- Changes to how `<attribute>_changed?` functions in lifecycle calls such as `after_save`.  Basically, `<attribute>_changed?`'s behavior is changing, and to get the old behavior we need to use `saved_change_to_<attribute>?` instead.  Similarly, `<attribute>_was` needs to be changed to `<attribute>_before_last_save`.  In any cases where I found warnings about this, I updated the code to use the new syntax.
- The use of `stub` in RSpec tests is deprecated in favor of `expect(<thing>).to receive`.  In any tests I found with the old syntax, I updated to the new syntax.

There were also a couple of other miscellaneous thing in there, such as:
- a `puts` call in one of our specs which was cluttering up console output
- a couple of spots where we were using `expect .to raise_error` or `expect .not_to raise_error(SomethingSpecific)`, both of which can pass erroneously (in both cases when errors you don't anticipate get raised)
- a number of tests for `response.should redirect_to '...'` which is deprecated in favor of `expect(response).to redirect_to '...'` which I updated

### Notion Card Links
https://www.notion.so/quill/Fix-DEPRECATION-WARNINGS-in-LMS-and-Evidence-engine-tests-24cde87455b24d0eacdfe1366472dae7

PR Checklist | Your Answer
------------ | -------------
Have you deployed to Staging? | No, no functional changes


happythenewsad (approved):

brendanshean (approved):
Thank you for cleaning this up!  It will be nice to have those deprecation messages eliminated.

PR #8194, opened 2021-09-03
CMS - Add fallback for new questions added to system.

dandrabik:
## WHAT
The materialized view (`GradedResponse`) that serves question responses is refreshed nightly. If a question is added by our Curriculum Team during the day, the endpoint wouldn't find any responses.

This adds a fallback query on the `Response` table if no responses are found to handle this edge case.
## WHY
We want newly added questions to work immediately. Also, since this table is refreshed nightly, this issue was impossible to replicate the next day (since it only happens the first day of a live question).
## HOW
Use the old controller code as a fallback if no responses are found.
### Screenshots
Investigation into the support ticket found the activity and question were updated recently.
Activity
Screen Shot 2021-09-03 at 10 28 25 AM
Question
Screen Shot 2021-09-03 at 12 07 02 PM

PR Checklist | Your Answer
------------ | -------------
Spec Review: Have you reviewed the spec and ensured this meets requirements and/or matches design mockups? |  N/a

### DIFF
--- a/services/QuillCMS/app/controllers/questions_controller.rb
+++ b/services/QuillCMS/app/controllers/questions_controller.rb
@@ -1,8 +1,13 @@
 class QuestionsController < ApplicationController
   MULTIPLE_CHOICE_LIMIT = 2
+  CACHE_EXPIRY = 24.hours.to_i
 
   def responses
-    render json: GradedResponse.no_parent.where(question_uid: params[:question_uid])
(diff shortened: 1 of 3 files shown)


cissyxyu (approved):

emilia-friedberg (approved):
Good catch! Thank you for getting this fix up so quickly!

PR #8204, opened 2021-09-07
Workers and Rake task for bulk reindexing CMS Responses

cissyxyu:
## WHAT
Adding a worker that will collect response data for a slice of the Responses table and make an Elasticsearch Bulk call to reindex those responses. Then, add a rake task that divides all the Responses into batches and fires off many instances of the Bulk Reindex worker.

## WHY
First step in the Reindex CMS Responses project.

## HOW
Add workers, tasks, and tests.

### Screenshots

### Notion Card Links
https://www.notion.so/quill/RFC-for-CMS-Elasticsearch-reindex-2981c23f600c4908b032d42c0a4dbb3c

PR Checklist | Your Answer
------------ | -------------


services/QuillCMS/app/workers/bulk_reindex_responses_worker.rb
```diff
@@ -0,0 +1,20 @@
+class BulkReindexResponsesWorker
+  include Sidekiq::Worker
+  sidekiq_options retry: 5, queue: 'low'
+
+  def client
```
brendanshean: How is this method used?

services/QuillCMS/lib/tasks/reindex_responses.rake
```diff
@@ -0,0 +1,8 @@
+namespace :elasticsearch do
+  desc 'Reindex Responses in Elasticsearch'
+  task :build_response_index, [:batch_size] => [:environment] do |task, args|
+    (1..Response.last.id).step(args[:batch_size].to_i).each do |starting_index|
+      BulkReindexResponsesWorker.perform_async(starting_index, args[:batch_size].to_i - 1)
```
brendanshean: You could use [find_each](https://api.rubyonrails.org/v5.1.7/classes/ActiveRecord/Batches.html#method-i-find_each) and pass the batch size as a parameter in there.  This would optimize the querying by ensuring getting the full batch size each iteration (`starting_index..(starting_index+batch_size)` just ensures that you're getting at most batch size records)

I would probably use the rake task to call a `MainBulkReindexResponsesWorker` that spawns all of the smaller `BulkReindexResponsesWorker` that way the whole thing is running in Sidekiq.  Also, for testing purposes, you probably need some kind of parameter for `MainBulkReindexResponsesWorker` that sets a limit on the percent of responses that you reindex.

services/QuillCMS/app/workers/bulk_reindex_responses_worker.rb
```diff
@@ -0,0 +1,20 @@
...
+    @client = Elasticsearch::Client.new url: ENV['ELASTICSEARCH_URL']
+  end
+
+  def perform(starting_index, batch_size)
+    batch_for_bulk = []
```
brendanshean: This would be a good place to add logic for [cancelling jobs](https://github.com/mperham/sidekiq/wiki/FAQ#how-do-i-cancel-a-sidekiq-job)  It's much easier to add this logic and get it working now then later with the full dataset.

brendanshean (commented):
This looks really good so far.  I've added some comments on things that you probably want in there (unless those are for a subsequent PR).

services/QuillCMS/app/workers/bulk_reindex_responses_worker.rb
```diff
@@ -0,0 +1,20 @@
...
+    Response.where(id: starting_index..(starting_index+batch_size)).each do |response|
```
dandrabik: You should look to use Rails' [`find_in_batches`](https://apidock.com/rails/v6.1.3.1/ActiveRecord/Batches/find_in_batches), I believe it does everything you're doing here.

services/QuillCMS/app/workers/bulk_reindex_responses_worker.rb
```diff
@@ -0,0 +1,20 @@
...
+      batch_for_bulk.push({ index: { _id: response.id, data: response.as_indexed_json } })
```
dandrabik: If you are returning the whole collection with an action performed on each item, it's cleaner and shorter to use `.map`, i.e.

instead of this pattern
```ruby
result = []
array.each do |item|
  result.push(item.do_something)
end
return result
```
you can do
```ruby
result = array.map {|item| item.do_something}
```
So this could be something like:
```ruby
Response.__elasticsearch__.client.bulk(
  index: "responses",
  type: "response",
  body: responses.map {|r| { index: { _id: r.id, data: r.as_indexed_json } } }
)

File: services/QuillCMS/lib/tasks/reindex_responses.rake (hunk 2 above)
Comment (dandrabik) [reply services/QuillCMS/lib/tasks/reindex_responses.rake]: Let's pass the full batch size into the job. I think it will be surprising that this rake task does `args[:batch_size].to_i -1` before passing it along, e.g. the rake task is sent `batch_size` = 100, and then pass the job a `batch_size` = 99.

Comment (dandrabik) [reply services/QuillCMS/lib/tasks/reindex_responses.rake]: Agreed that there should be a run this rake task without running it for all responses for testing. Might be worth having `starting_index` and ending_index` params in case you need to stop it and start it again.

**Review by dandrabik — CHANGES_REQUESTED** [review]
Nice work getting started on this. There are some code cleanups I suggested. Also, let's make sure there is a way to run this without having to do the entire Response table (that will make your testing a lot easier!).

Let me know tomorrow if you have any questions about my comments, I'm happy to talk through them.

File: services/QuillCMS/app/workers/bulk_reindex_responses_worker.rb (hunk 1 above)
Comment (cissyxyu) [reply services/QuillCMS/app/workers/bulk_reindex_responses_worker.rb]: Hmm I initially thought I needed it to form a connection to the ElasticSearch service, but I think it's actually redundant. Removing this!

File: services/QuillCMS/lib/tasks/reindex_responses.rake (hunk 2 above)
Comment (cissyxyu) [reply services/QuillCMS/lib/tasks/reindex_responses.rake]: > You could use [find_each](https://api.rubyonrails.org/v5.1.7/classes/ActiveRecord/Batches.html#method-i-find_each) and pass the batch size as a parameter in there. This would optimize the querying by ensuring getting the full batch size each iteration (`starting_index..(starting_index+batch_size)` just ensures that you're getting at most batch size records)
> 
> I would probably use the rake task to call a `MainBulkReindexResponsesWorker` that spawns all of the smaller `BulkReindexResponsesWorker` that way the whole thing is running in Sidekiq. Also, for testing purposes, you probably need some kind of parameter for `MainBulkReindexResponsesWorker` that sets a limit on the percent of responses that you reindex.

For simplicity's sake I changed it to just the one worker that loops through the responses and sends bulk requests for each batch. Testing this on staging now to see how this goes.

File: services/QuillCMS/app/workers/bulk_reindex_responses_worker.rb [hunk 6]
```diff
@@ -0,0 +1,19 @@
+require 'sidekiq/api'
+
+class BulkReindexResponsesWorker
+  include Sidekiq::Worker
+  sidekiq_options retry: 5, queue: 'low'
+
+  def perform(starting_index, end_index, batch_size)
+    responses = Response.find_each(start: starting_index, finish: end_index, batch_size: batch_size)
```
dandrabik: Nice refactor on this. So since you are using the entire batch instead of individual records, you'd want to use [`find_in_batches`](https://apidock.com/rails/ActiveRecord/Batches/find_in_batches) (it's a related method to `find_each`).

So something like this:
```ruby
Response.find_in_batches(...) do |responses|
  Response.__elasticsearch__.client.bulk(
      index: "responses",
      type: "response",
      body: responses.map { |r| {index: {_id: r.id, data: r.as_indexed_json}}}
    )
end
```

You can now use this job to loop through the responses, not just a single set, e..g. to loop through the first 100K records in chunks of 1,000, you can call:
```ruby
BulkReindexResponsesWorker.perform_async(1, 100_000, 1_000)

**Review by dandrabik — APPROVED** [review]
Requested a minor code change to make this more flexible, so approving assuming that will be put in place. :+1: for testing this on staging.

Nice work.

**Review by happythenewsad — APPROVED** [review]

File: services/QuillCMS/spec/workers/bulk_reindex_responses_worker_spec.rb [hunk 7]
```diff
@@ -0,0 +1,35 @@
+require 'rails_helper'
+
+describe BulkReindexResponsesWorker do
+  let(:worker) { described_class.new }
+  let!(:response) { create(:response, question_uid: 'blah', count: 1, first_attempt_count: 1) }
+  let!(:response2) { create(:response, question_uid: 'blah', parent_id: response.id, count: 1, first_attempt_count: 1) }
```
brendanshean, line 6: More of something to keep in mind in the future. 

If I'm using creating the same factory object multiple times, I'll usually use `create_list` to cut down on boilerplate and you have an ordered object if you need it:

`let!(:responses) { create_list(:response, 4, question_uid: 'blah', parent_id: response.id, count: 1, first_attempt_count: 1) }`

brendanshean (approved):
Looks great!  Well done on this!

PR #8206, opened 2021-09-07
LMS - Don't return body in ActiveActivitySession#update

dandrabik:
## WHAT
This is our most hit `ActiveActivitySessionsController#update` is our most hit endpoint. We're returning a `json` body that isn't used by any response (that I can find). This PR changes the endpoint to return a 204 and no content.
## WHY
Slight speed improvement and less unused data being sent to the client (slight over-optimization, but given how much this endpoint is hit, we should see gains). Also, a REST update technically should return a 204 and no content.
## HOW
return `head no_content`
### Screenshots
Load for this endpoint
Screen Shot 2021-09-07 at 6 04 34 PM
**Before large 'with passages' request**
Screen Shot 2021-09-07 at 5 54 45 PM
**After**
Screen Shot 2021-09-07 at 5 56 31 PM

**Before smaller request** 
Screen Shot 2021-09-07 at 5 09 43 PM

**After**
Screen Shot 2021-09-07 at 5 10 20 PM

PR Checklist | Your Answer
------------ | -------------


brendanshean (approved):

anathomical (approved):
Yeah, this makes sense.  I think any case where the front-end needs to confirm that data has been updated as expected it can just fetch the `ActiveActivitySession` on the GET endpoint.

happythenewsad (approved):

PR #8218, opened 2021-09-09
Refactor google integration namespace, part 1

brendanshean:
## WHAT
1.  Flatten out the namespaces `GoogleIntegration::Classrooms` and `GoogleIntegration::Users`.  
2.  Change compact style module-class declarations for some of the GoogleIntegration classes to the nested form.

Per @dandrabik [insight](https://github.com/empirical-org/Empirical-Core/pull/8218#pullrequestreview-750576772), I'm breaking this up into two parts

## WHY
1.  I recently created a couple namespaces `GoogleIntegration::Classrooms` and `GoogleIntegration::Users`.  However, after getting deeper into the code, there was sometimes a use for `classrooms_[object]` and `classroom_[object]`.  Removing this extra layer of hierarchy allows for precise, unambiguous naming of classes.

2.  As far as the nested class declaration, it's better to have nested reasons due to constant resolution. For example, see this [post](https://github.com/rubocop/ruby-style-guide#namespace-definition)

## HOW
1.  Rename classes with appropriate prefix (e.g. Classroom, Classrooms, Teacher) and move respective tests and references.
2. Wrap classes in `module GoogleIntegration`

### Notion Card Links
https://www.notion.so/quill/Google-Classroom-and-Clever-Refactor-google-integration-namespace-6dc6ecc026954b7f880445bb7916ca57

PR Checklist | Your Answer
------------ | -------------
Have you deployed to Staging? | Not yet, deploying now.

### DIFF
--- a/services/QuillLMS/app/controllers/auth/google_controller.rb
+++ b/services/QuillLMS/app/controllers/auth/google_controller.rb
@@ -42,7 +42,7 @@ class Auth::GoogleController < ApplicationController
 
   private def run_background_jobs
     if @user.teacher?
-      GoogleIntegration::Users::UpdateImportedClassroomsWorker.perform_async(@user.id)
+      GoogleIntegration::TeacherUpdateImportedClassroomsWorker.perform_async(@user.id)
       GoogleStudentImporterWorker.perform_async(@user.id, 'Auth::GoogleController')
     elsif @user.student?
       GoogleStudentClassroomWorker.perform_async(@user.id)
--- a/services/QuillLMS/app/controllers/teachers/classroom_manager_controller.rb
+++ b/services/QuillLMS/app/controllers/teachers/classroom_manager_controller.rb
@@ -171,7 +171,7 @@ class Teachers::ClassroomManagerController < ApplicationController
   end
 
   def retrieve_google_classrooms
-    serialized_google_classrooms = GoogleIntegration::Users::ClassroomsCache.get(current_user.id)
+    serialized_google_classrooms = GoogleIntegration::TeacherClassroomsCache.get(current_user.id)
     if serialized_google_classrooms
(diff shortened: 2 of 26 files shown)


happythenewsad (approved):

dandrabik (changes requested):
So overall the renaming is :+1:.

But I think this needs to be two deploys because some of the old classes/modules are used by `sidekiq` jobs (or are `sidekiq` jobs) that will persist past a deploy and cause errors if the code is no longer there.

So, something like this:
1. Deploy new names/modules, 
2. Wait until enqueued Sidekiq jobs have no references to old classes/modules 
3. Deploy the removal of the old classes/modules.

Otherwise you'll end up with a bunch of failed jobs that can't find the code required to run. Note, given that these imports are jobs the user is waiting on, we'd also like to avoid the "manually fix any failed jobs" approach which is usually an alternative, but would be too slow here.

brendanshean:
> So overall the renaming is 👍.
> 
> But I think this needs to be two deploys because some of the old classes/modules are used by `sidekiq` jobs (or are `sidekiq` jobs) that will persist past a deploy and cause errors if the code is no longer there.
> 
> So, something like this:
> 
> 1. Deploy new names/modules,
> 2. Wait until enqueued Sidekiq jobs have no references to old classes/modules
> 3. Deploy the removal of the old classes/modules.
> 
> Otherwise you'll end up with a bunch of failed jobs that can't find the code required to run. Note, given that these imports are jobs the user is waiting on, we'd also like to avoid the "manually fix any failed jobs" approach which is usually an alternative, but would be too slow here.

@dandrabik This makes complete sense.  Thank you for anticipating these issues.

dandrabik (approved):
:+1: LGTM. Thanks for making the requested changes.

PR #8245, opened 2021-09-10
lms/update-migrations

anathomical:
## WHAT
Update old migration files so that they're runnable
## WHY
So that new developers, or anyone who needs to wipe and rebuild their database, can run through all the migrations
## HOW
For the most part, just add a Rails version (`[4.2]`) to old migrations.  In a couple of places, comment out some code that refers to no-longer-existent models.

PR Checklist | Your Answer
------------ | -------------
Have you added and/or updated tests? |  No test behavior changed, but tests do pass.
Have you deployed to Staging? | No - not really relevant to staging

### DIFF
--- a/services/QuillLMS/db/migrate/20200604165331_migrate_lessons_to_activity.rb
+++ b/services/QuillLMS/db/migrate/20200604165331_migrate_lessons_to_activity.rb
@@ -1,72 +1,68 @@
 class MigrateLessonsToActivity < ActiveRecord::Migration[4.2]
   def change
-    if defined?(Lesson)
-      # Before running this part, make sure that the inconsistent names in this
-      # spreadsheet are resolved:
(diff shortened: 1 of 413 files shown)


services/QuillLMS/db/migrate/20200604165331_migrate_lessons_to_activity.rb
```diff
@@ -18,43 +18,47 @@ def change
                           '-LKld_Wf0lzgNnEvY5aA'
                         ]
 
-      Lesson.where(uid: lessons_to_archive).each do |lesson|
-        lesson[:data]["flag"] = "archived"
-        lesson.save!
-      end
+      # The Lesson model is no longer in the code base, so this
```
dandrabik: The `if defined?(Lesson)` work to skip this? I'm guessing that was it's intention. Either way is likely fine, but we could also update the `if`

dandrabik (approved):
:+1: LGTM thanks for working through this.

happythenewsad (approved):

emilia-friedberg (approved):

PR #8256, opened 2021-09-14
LMS - Small Internal Tool performance improvements

dandrabik:
## WHAT
Two small CMS performance improvements.
1. User Search - do an exact match search for email
2. School#show - Don't pull in student sessions.

**Note, I made some layout changes since it was requested the partial email search be put back it, since it is occasionally used.**
## WHY
Performance improvements.

## HOW
There's a larger refactor that could be done here (I started doing one, then decided it wasn't worth it right now), but I went for easy wins.

1. The email search is slow because it does an ILIKE, but all the searches in the logs are exact match searches. The team isn't using the fuzzy search, let's give them the fast version.
2. Removing `activity_sessions` from the school's page makes it much faster for large schools. This data isn't used by Partnerships, since that data is in Vitally now.

Screenshots
Before
Screen Shot 2021-09-16 at 9 19 42 AM
After
Screen Shot 2021-09-16 at 9 19 21 AM
Toggle open
Screen Shot 2021-09-16 at 9 19 27 AM

### Notion Card Links
https://www.notion.so/quill/Performance-update-to-user-search-7002f74412ce443c8089febda06cc6cd

https://www.notion.so/quill/Performance-Updates-to-CMS-Schools-index-show-f4633f9bb2b04020aaabc6a2cd571772

PR Checklist | Your Answer
------------ | -------------
Have you added and/or updated tests? |  No, tiny changes

### DIFF
--- a/services/QuillLMS/client/app/bundles/Teacher/startup/CmsUserIndexAppClient.jsx
+++ b/services/QuillLMS/client/app/bundles/Teacher/startup/CmsUserIndexAppClient.jsx
@@ -100,7 +100,7 @@ export default class CmsUserIndex extends React.Component {
     }
   };
 
-  toggleAdvanced = () => { this.setState({showAdvanced: !this.state.showAdvanced})};
+  toggleAdvanced = () => { this.setState(prevState => ({ showAdvanced: !prevState.showAdvanced }))};
 
   search = (e) => {
     e ? e.preventDefault() : null;
--- a/services/QuillLMS/client/app/bundles/Teacher/startup/CmsUserIndexAppClient.jsx
+++ b/services/QuillLMS/client/app/bundles/Teacher/startup/CmsUserIndexAppClient.jsx
@@ -13,7 +13,8 @@ export default class CmsUserIndex extends React.Component {
       columns: this.getColumns(),
       data: props.queryResults,
       query: props.query,
-      loading: false
+      loading: false,
+      showAdvanced: false,
     }
   }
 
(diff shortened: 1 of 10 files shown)


cissyxyu (approved):

brendanshean (approved):
I think the email exact search is a good call here.  Nice work!

emilia-friedberg (approved):

services/QuillLMS/client/app/bundles/Teacher/startup/CmsUserIndexAppClient.jsx
```diff
@@ -99,6 +100,8 @@ export default class CmsUserIndex extends React.Component {
     }
   };
 
+  toggleAdvanced = () => { this.setState({showAdvanced: !this.state.showAdvanced})};
```
emilia-friedberg: If you want to use the current state in a `setState` function, you can pass a function like `this.setState(prevState => ({ showAdvanced: !prevState.showAdvanced }))`. Theoretically that avoids an edge case where the state has changed between the function being invoked and `this.state` gets called. Practically it doesn't really matter in this instance.

emilia-friedberg (approved):
One small suggestion you can take or leave. Normally I would request that you deconstruct state calls at the top of the function, but since this is a large old file and we aren't doing that in it already I don't think it's worth the refactor here.

services/QuillLMS/client/app/bundles/Teacher/startup/CmsUserIndexAppClient.jsx
dandrabik: Ah, nice, I like that pattern, I'll change this.

PR #8263, opened 2021-09-15
LMS - Add a weekly job to clean up obsolete ActiveActivitySessions.

dandrabik:
## WHAT
Add a weekly job that deletes `ActiveActivitySessions` that are no longer needed, either because the student finished the activity or
## WHY
This is the second largest table in the DB, so dropping some records should free up some storage. Also, this is table powers the most frequently accessed API endpoint in our system, so keeping the table smaller might have performance benefits (I would guess the gains would be small though)
## HOW
Add a query to search for obsolete records, loop through them in small batches, and `delete_all`

PR Checklist | Your Answer
------------ | -------------

### DIFF
--- a/services/QuillLMS/spec/factories/active_activity_session.rb
+++ b/services/QuillLMS/spec/factories/active_activity_session.rb
@@ -3,7 +3,7 @@ FactoryBot.define do
     foo: 'bar'
   }
   factory :active_activity_session do
-    uid                   SecureRandom.uuid
-    data                  data
+    uid   {SecureRandom.uuid}
+    data  data
   end
 end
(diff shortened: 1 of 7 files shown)


services/QuillLMS/spec/factories/active_activity_session.rb
```diff
@@ -3,7 +3,7 @@
     foo: 'bar'
   }
   factory :active_activity_session do
-    uid                   SecureRandom.uuid
-    data                  data
+    uid   {SecureRandom.uuid}
```
brendanshean, line 6: Good catch on this one.  We might want to consider this [cop](https://www.rubydoc.info/gems/rubocop-rspec/RuboCop/Cop/RSpec/FactoryBot/AttributeDefinedStatically) since this can be a subtle source of bugs.

brendanshean (approved):
LGTM!  Nice here template here for how to write a cron job.  

NB, there's a  lint error with one of the lambdas.

PR #8275, opened 2021-09-17
LMS - CircleCI, ignore tests on deploy-* branches

dandrabik:
## WHAT
Filter out `deploy-*` branches from our CircleCI run.
## WHY
Might be causing issues with tests being skipped on PRs. Also, these branches aren't monitored and are only used for deploy (when they are direct copies of branches that already have test runs), so there is no value in running the tests here.
## HOW
Use CircleCI's filtering on each job (you can't do it on the whole workflow).
### Screenshots
Screen Shot 2021-09-17 at 10 52 05 AM

### Notion Card Links
https://www.notion.so/quill/CircleCI-fake-test-passes-edafcd5b52d444839163755dc96a7f81

PR Checklist | Your Answer
------------ | -------------
Have you added and/or updated tests? |  No, config code

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -1,4 +1,4 @@
-version: 2
+version: 2.1
 jobs:
   lms_rails_build:
     working_directory: ~/Empirical-Core
@@ -337,7 +337,6 @@ jobs:
           command: |
             flake8
 workflows:
-  version: 2.1
   build-test:
     filters:
       branches:
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -1,4 +1,4 @@
-version: 2.1
+version: 2
 jobs:
   lms_rails_build:
     working_directory: ~/Empirical-Core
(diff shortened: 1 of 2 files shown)


happythenewsad (commented):
This looks like a good feature to ship. However, using the current syntax, any future configurations to `lms_rails_build` will be applied to most everything else.

How about abstracting `default_filter` into a separate yaml entity, and have `lms_rails_build` include it the same way that its sibling jobs include it?

dandrabik:
> This looks like a good feature to ship. However, using the current syntax, any future configurations to `lms_rails_build` will be applied to most everything else.
> 
> How about abstracting `default_filter` into a separate yaml entity, and have `lms_rails_build` include it the same way that its sibling jobs include it?

@happythenewsad In general, I'd agree that we'd want this abstracted, but unfortunately CircleCI validates the keys and format of this file, so it would lead to errors like this if we added a key to hold this info. What's in the PR was the cleanest way without adding a new key or fake job.
Screen Shot 2021-09-17 at 11 23 55 AM

happythenewsad (approved):

brendanshean (approved):
Thank you for configuring this!

PR #8281, opened 2021-09-20
LMS - Add better caching for Concepts / Questions / ConceptFeedback

dandrabik:
## WHAT
This adds/adjusts caching in 2 places and preps a third to be cached:
1. Use action caching to cache concepts
2. Make the `questions#index` and `#show` caching longer, and refactor to refresh cache on update.
3. Prep `ConceptFeedback#index` for future/easier caching by moving a param to the route (leave a TODO, will use `action_caching`)
## WHY
These are all high volume endpoints, so there are caching gains to be had here.
## HOW
I'm putting standard Rails caching in place rather than custom caching code. I'm using:
1. `action_caching` for data that rarely changes (less complex)
2. `Rails.cache.fetch` for data that does change, and expiring/refreshing it as is needed (more complex)
### Screenshots
Screen Shot 2021-09-20 at 11 17 32 AM

PR Checklist | Your Answer
------------ | -------------
Spec Review: Have you reviewed the spec and ensured this meets requirements and/or matches design mockups? | N/a

### DIFF
--- a/services/QuillLMS/app/controllers/api/v1/concept_feedback_controller.rb
+++ b/services/QuillLMS/app/controllers/api/v1/concept_feedback_controller.rb
@@ -2,6 +2,7 @@ class Api::V1::ConceptFeedbackController < Api::ApiController
   before_action :activity_type
   before_action :concept_feedback_by_uid, except: [:index, :create, :update]
 
+  # TODO: Cache this once routes are fixed
   def index
     all_concept_feedbacks = ConceptFeedback.where(activity_type: @activity_type).all.reduce({}) { |agg, q| agg.update({q.uid => q.as_json}) }
     render(json: all_concept_feedbacks)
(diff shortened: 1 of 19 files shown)


services/QuillLMS/app/controllers/api/v1/concept_feedback_controller.rb
```diff
@@ -2,6 +2,7 @@ class Api::V1::ConceptFeedbackController < Api::ApiController
   before_action :activity_type
   before_action :concept_feedback_by_uid, except: [:index, :create, :update]
 
+  # TODO: Cache this once routes are fixed
```
dandrabik, line 5: The goal in this PR is to change the route from `api/v1/concept_feedback.json?activity_type=grammar` to `api/v1/activity_type/grammar/concept_feedback.json` so that action_cache can be used here.

cissyxyu (approved):

PR #8284, opened 2021-09-21
Csp3

happythenewsad:
## WHAT
- adds a very permissive CSP for initial rollout, which is expected to work in production
- I created a [test environment](https://dashboard.heroku.com/apps/agile-savannah-89665) and confirmed that this config works on both staging and the new environment. I plan to buy and manage a new domain name for a more complete test. **I don't want to block this PR until the new domain is set up, but will if there are concerns.**
- adds the full CSP config in report-only mode. This allows use to inspect CSP violations in production via console, but avoids breaking the site for users. This will allow us to incrementally add restrictions. For example, whitelisting fully qualified subdomains, instead of using `*.quill.org` syntax. There is no documented difference between these syntaxes, but there may be a real-life difference.

## WHY
- We don't know why the initial CSP broke production, so we want as many debugging and incremental rollout tools as possible. UPDATE: I have high confidence from a recent test that the issue is that the `*.domain.org` CSP syntax does not encompass `domain.org` assets. In spite of this, I plan to use the same rollout strategy that this PR captures.

## HOW
See secure_headers.rb

### Screenshots

### Notion Card Links
https://www.notion.so/quill/Missing-Content-Security-Policy-Header-a285458ffd204229ad0999ba7118da8c

PR Checklist | Your Answer
------------ | -------------

### DIFF
--- a/services/QuillLMS/config/initializers/secure_headers.rb
+++ b/services/QuillLMS/config/initializers/secure_headers.rb
@@ -1,5 +1,24 @@
+csp_types = %w(default_src script_src font_src img_src style_src connect_src)
+permissive_config = csp_types.each_with_object({}) do |n, memo|
+  memo[n.to_sym] = [
+    "*" # wildcard directive must not be quoted
+  ]
+  memo
+end
+permissive_config[:script_src] = permissive_config[:script_src].concat(
+  [
+    "'unsafe-inline'",
+    "'unsafe-eval'"    
+  ]
+)
+permissive_config[:style_src] = permissive_config[:style_src].concat(
+  [
+    "'unsafe-inline'"
+  ]
+)
+
 SecureHeaders::Configuration.default do |config|
-  config.csp = {
+  default_config = {
     default_src: [
       "'self'", 
       "https://*.quill.org",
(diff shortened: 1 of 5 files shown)


services/QuillLMS/config/initializers/secure_headers.rb
```diff
@@ -1,5 +1,24 @@
+csp_types = %w(default_src script_src font_src img_src style_src connect_src)
+permissive_config = csp_types.each_with_object({}) do |n, memo|
```
dandrabik, line 2: There are some shorter ways to write this current code using enumerable methods, but it might just be clearer (and possibly shorter overall) to just explicitly write out the full config, e.g.
```ruby
permissive_config = {
  default_src: ["*"],
  font_src: ["*"],
  script_src: [
      "*", 
      "'unsafe-inline'",
      "'unsafe-eval'",
  ],
  ...
```
  I would have found this easier to parse and compare with the `default_config`

dandrabik (approved):
Overall, I like this approach of making this permissive and turning on report-only. Nice work on this.

For readability, I think it would be better to make the `permissible_config` a static hash, but otherwise :+1:

services/QuillLMS/config/initializers/secure_headers.rb
happythenewsad: Yes - this is not the best syntax for long term code. However, it is helpful when I need to iterate rapidly over the config - sometimes applying changes to all keys, and sometimes applying changes to individual keys. I plan to remove all of `permissive_config` in a matter of days when it becomes obsolete.

PR #8269, opened 2021-09-16
Show unsynced Google Classroom students and allow teachers to remove/move those students

brendanshean:
## WHAT
The new model `ProviderClassroomUser` tracks which students belong to a classroom on a particular provider (i.e. Google Classroom or Clever). Each of these records has a `deleted_at` attribute which when null corresponds to an active student in the provider classroom; a non-null value indicates the student used to be in that classroom but is no longer. This PR addresses requirements for incorporating this data on the teacher classrooms page:

1. Don't show the "Remove from class" option for students that are currently synced with Google Classroom or Clever: 
2. Show "Remove from Class" and "Move class" options for unsynced Google Classroom students
3. Change Synced column details for a 'Google Classroom' classroom to 'Yes' or 'No'.  In the case of 'No' add a hover indicating the 'Student is no longer in this class in Google Classroom'
4. Remove synced column for classrooms without a provider (i.e. Google Classroom or Clever)

## WHY
Actions on Quill should be consistent with the logic and source of truth on Google Classroom provider.

## HOW
To solve these related items, we first add a couple of pieces to the classrooms#index JSON:
a) Add a new `providerClassroom` key to the classrooms JSON.
b) Add a new `synced` key to the  classrooms_students JSON.  The value of `synced` is computed via `ProviderClassroomUser` query to see if a particular record exists and is `active`.  The `ProviderClassroom` uses the decorator pattern to target a specific use case of `Classroom` (i.e. those classrooms with a Provider).  

1 & 2.  Assign actions based on [synced status](https://github.com/empirical-org/Empirical-Core/pull/8269/files#diff-2b27ac339f8a8cd829c6efb0b7a8181c8853cc41c77493e210bc2c635ee1e6b0R188)
3. The crux of the front-end implementation was getting the tooltip to work for with the DataTable Component.  @emilia-friedberg uncovered that an `overflow:  hidden` directive was preventing the tooltip to load.  After adding a small css [rule](https://github.com/empirical-org/Empirical-Core/pull/8269/files#diff-2b27ac339f8a8cd829c6efb0b7a8181c8853cc41c77493e210bc2c635ee1e6b0R35) on the synced column, the tooltip worked.  Once that was in place, utilize the [syncedStatus](https://github.com/empirical-org/Empirical-Core/pull/8269/files#diff-2b27ac339f8a8cd829c6efb0b7a8181c8853cc41c77493e210bc2c635ee1e6b0R428) for display.
[...]

PR #8288, opened 2021-09-22
Show unsynced Clever Classroom students and allow teachers to remove/move those students

brendanshean:
## WHAT
Most of this is a repeat of this `GoogleClassroomUser` [PR](https://github.com/empirical-org/Empirical-Core/pull/8269) but applied to Clever classrooms.

The new model ProviderClassroomUser tracks which students belong to a classroom on a particular provider (i.e. Google Classroom or Clever). Each of these records has a deleted_at attribute which when null corresponds to an active student in the provider classroom; a non-null value indicates the student used to be in that classroom but is no longer. This PR addresses requirements for incorporating this data on the teacher classrooms page:

1.  Don't show the "Remove from class" option for students that are currently synced with Clever:
1. Show "Remove from Class" and "Move class" options for unsynced Clever students
1.  Change Synced column details for a 'Clever' classroom to 'Yes' or 'No'. In the case of 'No' add a hover indicating the 'Student is no longer in this class in Clever Classroom'

## WHY
Actions on Quill should be consistent with the logic and source of truth on Clever classroom provider.

## HOW
1 & 2.  Update actions to [check](https://github.com/empirical-org/Empirical-Core/pull/8288/files#diff-2b27ac339f8a8cd829c6efb0b7a8181c8853cc41c77493e210bc2c635ee1e6b0R187) for `clever_id`
3. Utilize the [syncedStatus](https://github.com/empirical-org/Empirical-Core/pull/8269/files#diff-2b27ac339f8a8cd829c6efb0b7a8181c8853cc41c77493e210bc2c635ee1e6b0R428) for display.

NB.  There's also some refactoring of the `CleverIntegration::Classrooms` namespace to a flatter representation.  For example, move `CleverIntegration::Classrooms::Importer` to `CleverIntegration::ClassroomImporter`.

### Screenshots
Screen Shot 2021-09-22 at 3 15 44 PM

### Notion Card Links

PR Checklist | Your Answer
------------ | -------------
Spec Review: Have you reviewed the spec and ensured this meets requirements and/or matches design mockups? | Not yet.


services/QuillLMS/app/services/clever_integration/classroom_updater.rb
```diff
@@ -0,0 +1,40 @@
...
+
+    def initialize(data)
+      @data = data
+      @clever_id = data[:clever_id]
+      @grade = data[:grade]
+    end
+
+    def run
+      update!
+      classroom
```
dandrabik, line 13: Is this `classroom` unneeded since `update!` would return the classroom?

services/QuillLMS/app/services/clever_integration/classroom_updater.rb
```diff
@@ -0,0 +1,40 @@
...
+
+    private def custom_name?
+      classroom.name != classroom.synced_name
+    end
+
+    private def name
+      custom_name? ? classroom.name : data[:name]
+    end
+
+    private def synced_name
```
dandrabik: Should `synced_name` be an `attr_reader` for consistency? Also, you use `data[:name]` in the `def name`, should that be `synced_name`?

services/QuillLMS/app/services/clever_integration/classroom_students_importer.rb
```diff
@@ -0,0 +1,45 @@
...
+  class ClassroomStudentsImporter
+    attr_reader :classroom, :students_data
+
+    def initialize(classroom, students_data)
+      @classroom = classroom
+      @students_data = students_data
+    end
+
+    def run
+      return if students_data.nil? || students_data.empty?
```
dandrabik, line 11: Rails has a shorter version of this (you can take or leave)
```suggestion
      return if students_data.blank?
```

services/QuillLMS/app/services/clever_integration/classroom_creator.rb
```diff
@@ -0,0 +1,35 @@
...
+      Classroom.create!(
+        clever_id: clever_id,
+        grade: grade,
+        name: name,
+        synced_name: synced_name
+      )
+    end
+
+
+    private def clever_id
```
dandrabik: take or leave, these could be `attr_reader`.

services/QuillLMS/app/services/clever_integration/classroom_students_importer.rb
```diff
@@ -0,0 +1,45 @@
...
+
+    private def parsed_students_data
+      students_data.map { |student_data| parsed_student_data(student_data['id'], student_data['name']) }
+    end
+
+    private def parsed_student_data(id, name)
+      {
+        clever_id: id,
+        name: "#{name['first']} #{name['middle']} #{name['last']}".squish,
+        username: id
```
dandrabik, line 33: Just double checking that this is intentional and not a typo:
```ruby
username: id

**Review by dandrabik — APPROVED** [review]
I made some small take or leave comments, but overall this looks good. :+1:

I do have a larger question about this setup (that can be addressed in a later PR, since the current code matches an existing pattern): 

Should this all be wrapped in a `transaction` block? 

All of the saves raise on error, but they aren't connected in anyway, so you could create the `classroom`, but have the `teacher`->`classroom` association creations or `student` creations fail. Would we want this to be an all-or-nothing import?

Note that partial imports have happened and caused issues (this card captures an example: https://www.notion.so/quill/Clever-Classrooms-missing-Teachers-1be9032fb94e4af48a1afa3e8156804d). Happy to discuss this.

File: services/QuillLMS/app/services/clever_integration/classroom_updater.rb (hunk 1 above)
Comment (brendanshean) [reply services/QuillLMS/app/services/clever_integration/classroom_updater.rb:13]: I think `classroom.update!` returns a boolean

File: services/QuillLMS/app/services/clever_integration/classroom_creator.rb (hunk 4 above)
Comment (brendanshean) [reply services/QuillLMS/app/services/clever_integration/classroom_creator.rb]: After reading this [post](https://www.codewithjason.com/dont-wrap-instance-variables-attr_reader-unless-necessary/) last week I was re-evaluating my use of `attr_reader` since it breaks encapsulation.  On second thought I will put them back in since it's easier to read and more consistent with previous code.

File: services/QuillLMS/app/services/clever_integration/classroom_students_importer.rb (hunk 5 above)
Comment (brendanshean) [reply services/QuillLMS/app/services/clever_integration/classroom_students_importer.rb:33]: I definitely had to double-check, but this is how it was originally [setup](https://github.com/empirical-org/Empirical-Core/pull/8288/files#diff-4b1500e72eabdf0dbdddcfc51568e9bb021aa83bfe06d94baa66245ab4456a4fL15)

File: services/QuillLMS/app/services/clever_integration/libary_student_importer.rb [hunk 6]
```diff
@@ -0,0 +1,23 @@
+module CleverIntegration
+  class LibaryStudentImporter
```
emilia-friedberg: nbd but Library is missing an R here and in the file name!

emilia-friedberg (approved):

services/QuillLMS/app/services/clever_integration/libary_student_importer.rb
brendanshean: Good catch.  I must have been channeling my high school Latin teacher who pronounced it that way.

brendanshean:
> All of the saves raise on error, but they aren't connected in anyway, so you could create the `classroom`, but have the `teacher`->`classroom` association creations or `student` creations fail. Would we want this to be an all-or-nothing import?
> 
> Note that partial imports have happened and caused issues (this card captures an example: https://www.notion.so/quill/Clever-Classrooms-missing-Teachers-1be9032fb94e4af48a1afa3e8156804d). Happy to discuss this.

I think this is a very valid concern and I hope to address it in the aforementioned card.

Issue #344, opened 2014-09-15
Students not being correctly tied to classroom


RyanNovas:
If a space is added to the class code during initial login, students cannot join the class.
As a result, we should make spaces an invalid character in the class code field.

PR #8309, opened 2021-09-24
Evidence/validate highlights

anathomical:
## WHAT
Add visual indicators to the admin tool to indicate which Activities have Highlights associated with them that don't have working matches with the associated passage.
## WHY
When a Highlight does not exactly match the associated passage, it doesn't actually work at highlighting for students.  Currently there is no easy way for Curriculum to tell when this has happened (either because of a typo, or just because the passage has been updated), so an unknown number of highlights are not working and there's no way for Curriculum to figure out which ones that is without an arduous manual audit process.  Automating the flagging of these items will make it much easier for Curriculum to get them fixed.
## HOW
- Added a set of functions to the `Highlight`, `PlagiarismText` and `Activity` models to flag cases where highlights have not valid match.
- Added the results of those functions to serialized payloads so that they are returned as part of the API responses.
- Modified the front-end to display:
- - Which activities have any highlights that are invalid in the overall activities list
- - Links to specific rules that have highlights that are invalid on the activities detail page
- - An error message on rules pages that flags specific highlights that are not valid

### Screenshots
image
image
image

### Notion Card Links
https://www.notion.so/quill/Implementing-Evidence-Highlighting-validation-35beda02b591499fb3ce8c06bb3590d4

PR Checklist | Your Answer
------------ | -------------
Spec Review: Have you reviewed the spec and ensured this meets requirements and/or matches design mockups? | IN PROGRESS


services/QuillLMS/client/app/bundles/Staff/components/evidence/configureSettings/activityForm.tsx
```diff
@@ -146,6 +166,30 @@ const ActivityForm = ({ activity, handleClickArchiveActivity, requestErrors, sub
...
+      const ruleName = RULE_TYPE_TO_NAME[rule_type]
+      const invalidHighlightLink = (<Link to={`/activities/${id}/${ruleTypePart}/${rule_id}`}>{ruleName} Rule #{rule_id}</Link>);
+      return {
+        id: rule_id,
+        link: invalidHighlightLink
+      }
+    });
+
+    const dataTableFields = [
+      { name: "Invalid Highlights", attribute:"link", width: "100%" }
```
eadams17: based off seeing that `[Object object]` tooltip in the second screenshot and the fact that all the values will be short strings like "Rule #344", you can add `noTooltip: true` to this hash to remove that

services/QuillLMS/client/app/bundles/Staff/components/evidence/configureSettings/activityForm.tsx
```diff
@@ -253,6 +297,7 @@ const ActivityForm = ({ activity, handleClickArchiveActivity, requestErrors, sub
           handleSetPrompt={handleSetPrompt}
         />
       </form>
+      {invalid_highlights && invalid_highlights.length > 0 && renderInvalidHighlightLinks(invalid_highlights)}
```
eadams17: I think it would be nice to have this bool check above the return as a variable, something like `const invalidHighlightsPresent = invalid_highlights && invalid_highlights.length > 0`

eadams17 (approved):
Looks good! just two small comments

PR #8314, opened 2021-09-24
universal rules update rake task

happythenewsad:
## WHAT
idempotent CSV uploader for universal rules and concepts

## WHY
So that we can update these records in a standardized and safe manner (and prep the way for a GUI interface later)

## HOW
new rake task 

### Screenshots

### Notion Card Links
https://www.notion.so/quill/Update-the-concepts-feedback-for-universal-rules-in-the-internal-tool-ccc6d55ea35845bab690f1dae966111c

PR Checklist | Your Answer
------------ | -------------

### DIFF
--- a/services/QuillLMS/app/lib/universal_rule_loader.rb
+++ b/services/QuillLMS/app/lib/universal_rule_loader.rb
@@ -1,0 +1,76 @@
+# This loader assumes a specific CSV shape. The canonical copy is here:
+# https://www.notion.so/quill/860620e6d04f448784182a878be45b50?v=6e07319b3db14e42b850a27bac37ca73
+
+class UniversalRuleLoader 
+  def self.update_from_csv(type:, iostream:)
+    if !Evidence::Rule::TYPES.include?(type)
+      puts "Invalid rule type #{type}"
+      return
+    end
+
+    
+    if (CSV.parse(iostream, headers: true).headers & ['Rule UID', 'Rule', 'Concept UID', 'Feedback - Revised']).count != 4
+      puts 'Invalid headers. Exiting.'
+      return
+    end 
+
+    CSV.parse(iostream, headers: true) do |row|
+      rule = Evidence::Rule.find_by_uid(row['Rule UID'])
+      if rule.nil? || !rule.universal || rule.rule_type != type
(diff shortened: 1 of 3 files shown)


happythenewsad:
@anathomical @emilia-friedberg

services/QuillLMS/app/lib/universal_rule_loader.rb
```diff
@@ -0,0 +1,58 @@
...
+        puts "\nCannot find universal #{type} Rule with UID #{row['Rule UID']}. Creating new rule."
+        rule = Evidence::Rule.new(
+          uid: row['Rule UID'],
+          name: row['Rule'],
+          universal: true,
+          rule_type: type,
+          suborder: row['Overall Rule Priority'],
+          optimal: true, # Hardcoded, b/c the input file does not currently include an 'optimal' column
+          state: Evidence::Rule::STATES.first # same as above
+        )
```
anathomical, line 44: If this fires because of the second two conditions (not universal or not the right type) you'll attempt to create a new rule with the same UID which will fail due to uniqueness validation.

services/QuillLMS/spec/lib/universal_rule_loader_spec.rb
```diff
@@ -0,0 +1,93 @@
...
+        rule = Evidence::Rule.find_by_uid '1d66a'
+        expect(rule.concept_uid).to eq '4d5e3'
+
+        feedback = Evidence::Feedback.first 
+        expect([feedback.rule_id, feedback.text]).to eq(
+          [rule.id, 'the feedback']
+        )
+      end
+
+      it 'should create a new rule and new feedback' do
```
anathomical, line 90: Need to figure out what behavior we want when there's an existing rule with the UID, but it's not universal or is the wrong type and add some test cases for that edge case behavior (which might just be to throw an error or something).

anathomical (changes requested):
Need to figure out what behavior we actually want if there's a rule with the specified UID but it's somehow not universal or has the wrong type.

emilia-friedberg (approved):
LGTM other than Thomas's points, don't need to review again when those are made.

anathomical (approved):

PR #8363, opened 2021-10-06
Add finer logic to Merge Units to preserve completed activity sessions

cissyxyu:
## WHAT
When merging two classroom units, sometimes a student will have completed a session on one of the Classroom Units and started the session on the other Classroom Unit. In these cases, Partnerships always wants to preserve only the finished session and get rid of the started session.

## WHY
Teachers don't want to suddenly lose their students' completed Activity Session data -- they would rather keep that data and discard the half-completed session data.

## HOW
Add more branching logic to this function to account for this case and make sure this logic takes precedent over the "keep the most recently updated session" logic.

### Screenshots

### Notion Card Links
https://www.notion.so/quill/Merge-Classes-Teacher-Fix-Did-Not-Carry-Over-All-Completed-Data-one-teacher-9a8254fecbf5425998beb97afe853cfa

PR Checklist | Your Answer
------------ | -------------


services/QuillLMS/app/controllers/concerns/teacher_fixes.rb
```diff
@@ -48,6 +48,11 @@ def self.merge_activity_sessions_between_two_classroom_units(cu1, cu2)
       act_sesh2 = ActivitySession.find_by(classroom_unit: cu2, activity: act_sesh1.activity, user: act_sesh1.user)
       if act_sesh2.blank?
         act_sesh1.update!(classroom_unit_id: cu2.id)
+      elsif act_sesh2.state == ActivitySession::STATE_STARTED && act_sesh1.state == ActivitySession::STATE_FINISHED
```
brendanshean: I would move these equality checks into the activity_session.rb file.  One of them is already [written](https://github.com/empirical-org/Empirical-Core/blob/develop/services/QuillLMS/app/models/activity_session.rb#L151) and then you could write a similar method for `started?`.  Then the condition becomes `elsif act_sesh2.started? && act_sesh1.finished?`.  This is important since the logic of `started?` is now encapsulated inside of activity_session.rb

services/QuillLMS/app/controllers/concerns/teacher_fixes.rb
```diff
@@ -48,6 +48,11 @@ def self.merge_activity_sessions_between_two_classroom_units(cu1, cu2)
...
+        act_sesh2.update!(visible: false)
+        act_sesh1.update!(classroom_unit_id: cu2.id)
```
brendanshean, line 53: I might be missing it, but there doesn't seem to be a spec for this update on `classroom_unit_id`.

brendanshean (changes requested):

brendanshean (approved):
Looks good!  Nice clear tests

happythenewsad (approved):

PR #8365, opened 2021-10-07
opinion LMS endpoint

happythenewsad:
## WHAT
Move business logic from Opinion API into a new LMS endpoint, which calls Opinion API.

## WHY
So that OAPI and LMS are maximally decoupled, and so that we don't need to push complex config changes to oapi.

## HOW
New LMS endpoint. Also, a modification to OAPI's payload to support this.

### Screenshots

### Notion Card Links
https://www.notion.so/quill/Refactor-Evidence-Opinion-API-Flow-to-go-through-the-engine-93614b68aade43cc995e3277a491d4d3

PR Checklist | Your Answer
------------ | -------------


services/QuillLMS/engines/evidence/evidence.gemspec
```diff
@@ -1,25 +1,25 @@
...
-  s.test_files = Dir["spec/**/*"]
+  s.files = Dir['{app,config,db,lib}/**/*', 'MIT-LICENSE', 'Rakefile', 'README.rdoc']
+  s.test_files = Dir['spec/**/*']
 
-  s.add_dependency "google-cloud-automl", "~> 1.0.2"
-  s.add_dependency "google-cloud-automl-v1", "~> 0.3.0"
-  s.add_dependency "rails", "5.1.7"
+  s.add_dependency 'google-cloud-automl', '~> 1.0.2'
+  s.add_dependency 'google-cloud-automl-v1', '~> 0.3.0'
+  s.add_dependency 'rails', '5.1.7'
```
happythenewsad, line 22: No logic changes, just single quoting

brendanshean (approved):
Looks good.  Nice comment on the rspec stub_const item.

services/QuillLMS/engines/evidence/lib/evidence/evidence/opinion_feedback_assembler.rb
```diff
@@ -0,0 +1,41 @@
...
+      'Starts with Verb Check' => '4f4ed261-16f8-44ae-905c-0ad7c6449af4'
+    }
+
+    def initialize(oapi_response)
+      @oapi_response = oapi_response
+      error = @oapi_response['oapi_error']
+      rule_uid = OAPI_ERROR_TO_RULE_UID[error]
+      @rule = Evidence::Rule.where(uid: rule_uid).includes(:feedbacks).first
+      feedbacks = @rule&.feedbacks
+      @top_feedback = feedbacks.empty? ? nil : feedbacks.min_by { |e| e.order }
```
dandrabik: These two lines could be:
```suggestion
      @top_feedback = @rule&.feedbacks.min_by(&:order)  
```

services/QuillLMS/engines/evidence/lib/evidence/evidence/opinion_feedback_assembler.rb
```diff
@@ -0,0 +1,41 @@
+module Evidence 
+  class OpinionFeedbackAssembler
+    OAPI_ERROR_TO_RULE_UID = {
+      'Using Must' => '5545d756-9ba5-44d5-829a-0479cbbe941e',
```
dandrabik: I'm a little wary of using phrases with spaces, some capitalization, and some dashes as codes shared between two systems. In my experience, it's makes it easier to miss typos in coding and code review, e.g. if you listed `'Starts With Verb Check'` as a key in the opinion API, I'm not sure I'd notice the difference in review. Can we use snake case here, e.g. `using_must`?

Also, naming-wise, it would be nice to make these more consistent. Some have "Check", some don't. There are different verb forms, e.g. "Using Must", "Starts with Verb Check", and some don't have verbs (but could). 

These keys should have a clear pattern so if a check needs to be added, the format would be obvious, e.g. `must_check`, `first_person_reference_check`, etc.

services/QuillLMS/engines/evidence/app/controllers/evidence/opinion_controller.rb
```diff
@@ -0,0 +1,20 @@
+module Evidence
+  class OpinionController < ApiController
+    API_TIMEOUT = 5
+    def fetch
+      oapi_response = Timeout.timeout(API_TIMEOUT) do
```
dandrabik: I think it would be better to move this code into an API wrapper that abstracts these request details and has methods for the two pieces of info that will be used, `rule_uid` and `highlight`, e.g. 
```ruby
request = OpinionAPIRequest.new(params['entry'], params['prompt_text'])
request.rule_uid
request.highlight

```
I think this will make the code more testable and maintainable.

services/QuillLMS/engines/evidence/app/controllers/evidence/opinion_controller.rb
```diff
@@ -0,0 +1,20 @@
...
+          ENV['OPINION_API_DOMAIN'], 
+          headers:  {'Content-Type': 'application/json'},
+          body:     {
+            entry: params['entry'],
+            prompt_text: params['prompt_text']
+          }.to_json
+        )
+      end
+
+      assembler = OpinionFeedbackAssembler.new(oapi_response)
```
dandrabik: Following the comment above. We only need two pieces of information for this method, `rule_uid` and `highlight`. I think it would make the code clearer and more testable if those two were passed in, rather than the entire `HTTParty` response, since most of it isn't used.

services/QuillLMS/engines/evidence/lib/evidence/evidence/opinion_feedback_assembler.rb
```diff
@@ -0,0 +1,41 @@
...
+      feedbacks = @rule&.feedbacks
+      @top_feedback = feedbacks.empty? ? nil : feedbacks.min_by { |e| e.order }      
+    end
+
+    def to_payload 
+      {
+        'concept_uid': @rule&.concept_uid,
+        'feedback': @top_feedback&.text,
+        'feedback_type': 'opinion',
+        'optimal': @rule&.optimal,
```
dandrabik: So if a rule is not found (an API error or a bug), I think this payload would return `optimal: nil`, when I believe we want this to be `optimal: true` so that the student doesn't receive this feedback.

services/QuillLMS/engines/evidence/spec/controllers/evidence/opinion_controller_spec.rb
```diff
@@ -0,0 +1,83 @@
...
+          'response_id' => '0',
+          'highlight' => [
+            { 'type' => 'response', 'text' => 'someText', 'character' => 0 }
+          ],
+          'labels' => '',
+          'rule_uid' => example_rule_uid.to_s
+        })
+      end
+
+      context 'Rule.feedback exists' do
```
dandrabik, line 59: Let's add some tests for:
1. Opinion API raises a 500.
2. A key that isn't in the `OAPI_ERROR_TO_RULE_UID`.

I believe the payloads using the current code aren't what we'd want them to be.

services/QuillLMS/engines/evidence/spec/controllers/evidence/opinion_controller_spec.rb
```diff
@@ -0,0 +1,83 @@
...
+          { example_error => example_rule_uid }
+        )
+        allow(::HTTParty).to receive(:post).and_return(oapi_response)
+      end
+
+      it 'should return a valid json response' do 
+        post :fetch, params: incoming_payload, as: :json
+
+        expect(JSON.parse(response.body)).to eq({
+          'concept_uid' => 'ConceptUID',
```
dandrabik: If we are checking the expectation of `concept_uid` and `optimal`, we should explicitly set those when creating the evidence rule, or confirm they match the `rule.concept_uid`

dandrabik (changes requested):
Thanks for putting this together and good start on this. I'm requesting changes, since I believe the logic needs to be changed to properly handle API errors. I've also requested some structural changes to keep this code maintainable. 

Happy to jump on a call to walk through anything.

services/QuillLMS/engines/evidence/lib/evidence/evidence/opinion_feedback_assembler.rb
happythenewsad: I initially wrote it in one line, similar to your suggestion, but rubocop raised this error:
`Lint/SafeNavigationChain: Do not chain ordinary method call after safe navigation operator.`

services/QuillLMS/engines/evidence/lib/evidence/evidence/opinion_feedback_assembler.rb
happythenewsad: Yes, we certainly could do something like this, but it requires an intermediate step of updating our entire ecosystem with these changes: gapi, oapi, golang, postman, LMS (Rule), and [rules.csv.](https://www.notion.so/quill/860620e6d04f448784182a878be45b50?v=6e07319b3db14e42b850a27bac37ca73) 

Do you want me to do this work before merging this PR, or after?

services/QuillLMS/engines/evidence/spec/controllers/evidence/opinion_controller_spec.rb
happythenewsad: While I initially planned to use NewRelic.notice_error for these cases, I ended up letting them fail 'hard'. What's the desired behavior, if not failing hard (5xx)? I don't see a lot of value in returning HTTP 2xx and using NewRelic.notice_error, since the go server could interpret that as a green light to use the Oapi response.

services/QuillLMS/engines/evidence/lib/evidence/evidence/opinion_feedback_assembler.rb
brendanshean: I think rubocop wants an extra `&` before the`min_by`:

`@top_feedback = @rule&.feedbacks&.min_by(&:order)`.

services/QuillLMS/engines/evidence/spec/controllers/evidence/opinion_controller_spec.rb
dandrabik: I agree that's a possible path, but I don't think that's what the code will do. That is why a test would be helpful for these two scenarios. 

1. For the 500 error, I believe that `HTTParty` swallows 500 errors (e.g. you need to check if response was a `.success?` or use its `.code` method).
2. For the missing uid, you are essentially swallowing this error with `nil` checks. That works but I don't think the response is what we want, i.e. `optimal: nil`.

services/QuillLMS/engines/evidence/lib/evidence/evidence/opinion_feedback_assembler.rb
dandrabik: I think since this is a new, unused API endpoint, I think it makes sense to migrate it before putting this endpoint in use (rather than putting it in use and having to migrate a live endpoint). I don't believe it involves all of those items, but either way, let's migrate it now.

happythenewsad: @dandrabik OK, I'll create a ticket to capture this new work today, and pausing this PR for now

CC @brendanshean for awareness

services/QuillLMS/engines/evidence/lib/evidence/evidence/opinion/feedback_assembler.rb
```diff
@@ -0,0 +1,43 @@
...
+        'using_maybe' => '6e8c268a-9aac-4f0a-a373-4987bf861e9c',
+        'using_perhaps' => '2ca42c1d-9d0e-4d2b-a762-2eb10261f90b',
+        'using_please' => '1cb76abe-cd2d-4c87-86aa-b3bc7907c77a',
+        'starts_with_a_verb' => '4f4ed261-16f8-44ae-905c-0ad7c6449af4'
+      }
+
+      def initialize(client_response)
+        @client_response = client_response
+        error = @client_response['oapi_error']
+        rule_uid = OAPI_ERROR_TO_RULE_UID.fetch(error)
```
happythenewsad, line 22: fetch raises an exception when a key is not found.

services/QuillLMS/engines/evidence/lib/evidence/evidence/opinion/client.rb
```diff
@@ -0,0 +1,33 @@
...
+          response = HTTParty.post(
+            ENV['OPINION_API_DOMAIN'], 
+            headers:  {'Content-Type': 'application/json'},
+            body:     {
+              entry: @entry,
+              prompt_text: @prompt_text
+            }.to_json
+          )
+          if !response.success? 
+            raise "Encountered upstream error: #{response}"
```
dandrabik: Let's name this error type, e.g. `OpinionAPIError`

services/QuillLMS/engines/evidence/lib/evidence/evidence/opinion/client.rb
```diff
@@ -0,0 +1,33 @@
+module Evidence
+  module Opinion 
+    class Client
+      API_TIMEOUT = 500
```
dandrabik, line 4: I like this short timeout. Noting that the longest request in the last 7 days was 132 ms, so this looks reasonable.

Screen Shot 2021-10-20 at 6 36 48 PM

dandrabik (approved):
:+1: LGTM. Small comment about giving an error a name (this helps with error monitoring and debugging), but once that is in place, :+1:

PR #8371, opened 2021-10-08
lms/add-connection-pooling

anathomical:
## WHAT
Tweak database config to connect to PGBouncer pools when configured instead of connecting directly to the database
## WHY
This should give us virtually unlimited database connections so that we can avoid errors when making more than 500 database connections to the database.
## HOW
Using these instructions (https://devcenter.heroku.com/articles/postgres-connection-pooling) we've configured a PGBouncer connection pool.  Connect to that endpoint instead of the database directly (note that the only difference between the PGBouncer instance connection and the raw database connection is the port being connected to) when it's defined in the config, or connect directly to the db if there is no pool defined.

### Notion Card Links
https://www.notion.so/quill/Implement-LMS-Heroku-DB-connection-Pooling-964161abd8684b678ea4831ac3263c31

PR Checklist | Your Answer
------------ | -------------
Have you added and/or updated tests? |  No tests on configs

### DIFF
--- a/services/QuillLMS/config/database.yml
+++ b/services/QuillLMS/config/database.yml
@@ -18,11 +18,7 @@ test_env: &test_env
 # here are intended to simplify later decision to use followers
 # as well as disambiguate similarly-named ENV variables.
 leader_db_env: &leader_db_env
-  database: <%= ENV['LEADER_DB_NAME'] %>
-  username: <%= ENV['LEADER_DB_USERNAME'] %>
-  password: <%= ENV['LEADER_DB_PASSWORD'] %>
-  host: <%= ENV['LEADER_DB_HOSTNAME'] %>
-  port: <%= ENV['LEADER_DB_PORT'] %>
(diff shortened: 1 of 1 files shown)


brendanshean (approved):
Awesome. I'm glad this is being setup.

Would this also obviate the need for scaling servers down after deploys?

anathomical:
@brendanshean Yeah, that's the goal.  Basically, we get thousands of effective connections (while still adhering to the specified limits).

dandrabik (approved):
:+1: I think our application should be good to use this feature. We should do this off hours and monitor.

Before deploy, I think it's worth reading through the [configuration options outlined here](https://devcenter.heroku.com/articles/best-practices-pgbouncer-configuration) to make sure the pbbouncer configuration will fit our setup (I believe you are proposing "server-side" configuration, in "transaction" mode):

Also, this is the list of Postgres Commands we can't use in this setup (I don't believe we use any of these in our app, but it's worth a double check, and it's also helpful to know we can't do things lie `SET statement_timeout=10min` with this setup):

Screen Shot 2021-10-11 at 3 47 40 PM

PR #8373, opened 2021-10-08
Delete Rails cache on question right after saving

cissyxyu:
## WHAT
There is currently a race condition where we fire off a worker `RefreshQuestionCacheWorker` to refresh the question cache right after we save a question. However sometimes this worker can take a while to run, in which case the question cache is not deleted and it has incorrect data. This PR fixes this by manually deleting the cache data right away so a race condition where a GET request returns the wrong data will not occur.

## WHY
Admin were getting frustrated because they would modify questions and then not see their changes saved, or see unexpected data in questions because of the lag with the workers not refreshing the cache in time.

## HOW
Just delete the cache before firing off the worker, so in the time it takes for the worker to complete we don't have the wrong data in there.

### Screenshots

### Notion Card Links
https://www.notion.so/quill/Adding-New-Incorrect-Sequence-doesn-t-always-show-up-right-away-ecc95344f3f14a0b9a6fabc395c49643

PR Checklist | Your Answer
------------ | -------------


brendanshean (approved):
The WHAT/WHY/HOW really explained this small PR well.

PR #8378, opened 2021-10-11
Fix clever student importing bug where email and clever_id are disjointed

brendanshean:
## WHAT
Fix an issue where a clever student is imported but the email associated with the clever_id is not saved to that account.

## WHY
We'd like to have consistent behavior across clever imports so that teachers can expect emails, where provided, are linked with clever_id.

## HOW
There are a couple of fixes in this PR.  

1.  Our system doesn't [allow](https://github.com/empirical-org/Empirical-Core/blob/develop/services/QuillLMS/app/models/user.rb#L129) email addresses as usernames on the `User` model, yet sometimes clever data coming in has an email address as the username.  This [fix](https://github.com/empirical-org/Empirical-Core/pull/8378/files#diff-079dfa23ae95a4ba2ef35ac85475755cdff63e0d272126a3ad54e882ad6c82c3R7) addresses this constraint.

2.  The larger fix concerns a scenario where the clever_id and email of student becomes disjointed across two users in our database.  For example, consider:
```
user1 =  { clever_id: '324234234', google_id: nil, email: nil }
user2 =  { email: 'student@gmail.com`, google_id: '234234234', clever_id: nil }
```
When a teacher tries to import their clever roster containing a student with the user1.clever_id and user2.email, the current importing scheme imports user1 onto the roster, but user1.email remains nil.  In order to get the email displayed properly, we have to [merge](https://github.com/empirical-org/Empirical-Core/pull/8378/files#diff-06a255fce9946ff26367c32c1d511c466b1ca3f39e7b2f083e7727dec1893b7bR31) the student accounts and [update](https://github.com/empirical-org/Empirical-Core/pull/8378/files#diff-06a255fce9946ff26367c32c1d511c466b1ca3f39e7b2f083e7727dec1893b7bR32) the email.

Due to three separate uniqueness constraints on `email`, `clever_id` and `username`,  when updating a user we first have to determine which attribute of pre-existing student [already exists](https://github.com/empirical-org/Empirical-Core/pull/8378/files#diff-06a255fce9946ff26367c32c1d511c466b1ca3f39e7b2f083e7727dec1893b7bR39)  

### Screenshots

### Notion Card Links
https://www.notion.so/quill/Clever-Some-students-synced-from-Clever-don-t-have-their-email-address-added-to-their-Quill-account-74d35eaf071a4660af6c14ec2d5fde13

PR Checklist | Your Answer
------------ | -------------


services/QuillLMS/app/services/clever_integration/parsers/student.rb
```diff
@@ -4,6 +4,7 @@ def self.run(student)
     name_hash = student.name
     name = generate_name(name_hash.first, name_hash.last)
     username = student.credentials ? student.credentials.district_username : nil
+    username = username.split('@').first if username =~ ::User::VALID_EMAIL_REGEX
```
dandrabik, line 7: Are we worried about DB uniqueness constraints on `username` since we are removing the domain?

services/QuillLMS/app/services/clever_integration/student_updater.rb
```diff
@@ -0,0 +1,57 @@
...
+    private def fix_disjointed_clever_id_and_email
+      return unless clever_id_and_email_disjointed?
+
+      clever_student = User.find_by(clever_id: clever_id)
+      student.merge_student_account(clever_student)
+      clever_student.update!(clever_id: nil, account_type: 'unknown')
+    end
+
+    private def fix_username_conflict
+      data[:username] = nil if username_conflict?
```
dandrabik, line 36: Ah, I see you clear up the potential `username` conflict here.

Hmm, so you assign the `username`, then later in the code un-assign `username` if the  `username` exists in the DB. It might be cleaner to only assign it if it doesn't exist in the DB.

Also, that being said, there's a bit of a race condition for all of the checks here, since we do uniqueness checks [time passes] then we save (that check could become invalid in the [time passes]). That might be worth looking at.

services/QuillLMS/app/services/clever_integration/student_updater.rb
```diff
@@ -0,0 +1,57 @@
...
+    end
+
+    private def set_student
+      if email.present? && User.exists?(email: email)
```
dandrabik: Note, you use `User.exists?(email: email.downcase)` in the [other file](https://github.com/empirical-org/Empirical-Core/pull/8378/files#diff-bf463aae2e6f0346be374c1928527a529c4ec781e219eab86a7484b257f0b07fR29). 

It seems the booleans in that file, e.g. `student_with_email?`, need to be identical the checks here (if they drift, there will be bugs, e.g. `set_student` might not set a student), so maybe they should use shared methods in the `User` model or a mixin?

dandrabik (changes requested):
Nice work on this! Overall this looks good. 

I'm requesting changes since I think it would be worth shoring up two areas:
1. We are aiming to perform the same checks in different files, but we are not using shared methods (there was already a small drift between the methods in this PR). We should DRY this up to prevent bugs.
2. There is likely a subtle race-condition issue with how we are doing these checks that's worth looking at.

Note, this PR is an improvement to the code, so if either is a larger code change, we could ship this and fix as a follow-on card.

services/QuillLMS/app/services/clever_integration/student_updater.rb
brendanshean: Good point.  I will clean that up.

services/QuillLMS/app/services/clever_integration/student_updater.rb
brendanshean: I'm not sure I follow the "cleaner to only assign" recommendation.

The race condition is definitely a concern, but I think it would make more sense to do when I tackle this [story](https://www.notion.so/quill/Clever-Classrooms-missing-Teachers-1be9032fb94e4af48a1afa3e8156804d).

services/QuillLMS/app/services/clever_integration/parsers/student.rb
brendanshean: Another option would be `username = nil if username =~ ::User::VALID_EMAIL_REGEX`.

services/QuillLMS/app/services/clever_integration/student_updater.rb
dandrabik: Sorry for the late reply on this. To clarify, "the cleaner to only assign" comment, you are potentially assigning invalid data, then un-assigning that invalid data, so it seems like it might be cleaner to ONLY assign it if it is valid, e.g. 

Proposed:
`(valid -> assign, invalid -> don't assign)` 

Current: 
`(valid -> assign, invalid -> assign -> unassign)`

This is really minor (so you can take it or leave it), and if you are going to address the race condition in a future ticket, switching this to approved.

dandrabik (approved):
:+1: Thanks for your work on this.

services/QuillLMS/app/services/clever_integration/student_updater.rb
brendanshean: Got it.  That makes sense.  I'll make a note to clean that up when I look at the race conditions.

PR #8379, opened 2021-10-11
Only select visible activity sessions in the Activity Scores report

cissyxyu:
## WHAT
Our query for activity scores was not filtering for activity sessions with `visible: true`. As a result, some teachers were seeing unexpected data in the Activity Scores (e.g. students who did not belong to the class anymore and had long been archived). This PR adds that filter in the query to hide those old reports.

## WHY
Teachers were seeing phantom students in their Activity Scores reports, because these students may have been moved to a different class but their old archived activity sessions were still showing up in this Activity Scores report.

## HOW
Just add `activity_sessions.visible=true` to the SQL query we're making.

### Screenshots

### Notion Card Links
https://www.notion.so/quill/Students-showing-up-in-extra-classes-on-Activity-Scores-report-a7501155f2ab43e7a7a7bff3b6d2aa0c

PR Checklist | Your Answer
------------ | -------------


services/QuillLMS/spec/queries/progress_reports/activities_scores_by_classroom_spec.rb
```diff
@@ -3,7 +3,9 @@
 describe 'ActivitiesScoresByClassroom' do
   let(:classroom) {create(:classroom_with_classroom_units)}
 
-  it "returns a row for each student that completed an activity session" do
+  it "returns a row for each student that completed a visible activity session" do
+    activity_sessions = classroom.classroom_units.first.activity_sessions
+    ActivitySession.unscoped.each {|as| as.update(visible: true)}
```
brendanshean, line 8: Should line 8 be:  `activity_sessions.unscoped.each { |as| as.update(visible: true) }`?

**Optional:**
I've noticed our codebase uses `update(visible: true)` and `update(visible: false)`.  You could create a concern `Archivable` 

```
def unarchive
  update(visible: true)
end

def archive
  update(visible: false)
end
```
then extend in `class ActivitySession`.  

This would allow for cleaner:
Then merge line 7-8 into: `classroom.classroom_units.first.activity_sessions.unscoped.each(&:unarchive)`

brendanshean (approved):
I think one very minor fix needed in the spec, but should be good to go otherwise.

services/QuillLMS/spec/queries/progress_reports/activities_scores_by_classroom_spec.rb
brendanshean: @cissyxyu What is the purpose of line 7?

PR #8388, opened 2021-10-13
Fix an intermittent CircleCI failure

cissyxyu:
## WHAT
The `DashboardSpec` was intermittently failing because it capped the number of students at 2, when there are 3 student records that can be pulled. We need to actually be able to pull all of the records to ensure that there's never the condition where the third student is missing and the spec fails.

## WHY
On line 32, `expect(results.to_s).to include(name_of_lowest)`, the spec tries to find the name of the lowest scoring student in the pulled results. Because all 3 students have the same score, and the query capped results at 2 students, the results actually contain 2 out of the 3 students. We need to pull 3 out of the 3 students to ensure that the lowest scoring student is always in the results.

## HOW
Just raise the limit to 3.

### Screenshots

### Notion Card Links
https://www.notion.so/quill/Fix-intermittent-CI-failure-spec-mailers-csv_export_mailer_spec-rb-26-40284a5331e04d73b4b0b4cab879438f

PR Checklist | Your Answer
------------ | -------------


brendanshean (approved):
Thank you for fixing this!

PR #8406, opened 2021-10-18
go/reduce-api-timeout

anathomical:
## WHAT
Set a 5 second client-side timeout for API calls and test behavior
## WHY
Now that we do serial feedback we don't want to wait forever to send the user feedback if one of the APIs is taking forever to return.
## HOW
- Configure a client-side timeout in our globally-defined `client`.
- Update calls that use a client to take `client` as an argument (for easier testing)
- Write a test that times the client request out immediately and confirm that we get back the expected response (an `InternalAPIResponse` with `Error == true`.

### Notion Card Links
https://www.notion.so/quill/Evidence-feedback-sometimes-takes-between-30-sec-1-5-minutes-to-load-6587ba81246444b9b66d5ae6b4540d92

PR Checklist | Your Answer
------------ | -------------

### DIFF
--- a/services/comprehension/feedback-api-main/src/endpoint/endpoint.go
+++ b/services/comprehension/feedback-api-main/src/endpoint/endpoint.go
@@ -24,7 +24,7 @@ var default_api_response = APIResponse{
 }
 
 // TODO: This is a temporary replacement `http` that allows us to bypass SSL security checks
-var client = &http.Client {
+var base_client = &http.Client {
 	Transport: &http.Transport {
 		TLSClientConfig: &tls.Config{InsecureSkipVerify: true},
 	},
@@ -109,7 +109,7 @@ func Endpoint(context *gin.Context) {
 	var returnable_result APIResponse
 
 	for index, url := range urls {
-		results[index] = getAPIResponse(url, index, request_body, client)
+		results[index] = getAPIResponse(url, index, request_body, base_client)
 		return_index, finished := processResults(results, len(urls))
 
 		if finished {
@@ -135,7 +135,7 @@ func Endpoint(context *gin.Context) {
 		return
 	}
 
-	recordFeedback(request_object, returnable_result, GetFeedbackHistoryUrl(), client)
+	recordFeedback(request_object, returnable_result, GetFeedbackHistoryUrl(), base_client)
 
 	context.Header("Access-Control-Allow-Origin", "*")
 	context.Header("Content-Type", "application/json")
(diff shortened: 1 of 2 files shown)


services/comprehension/feedback-api-main/src/endpoint/endpoint_test.go
```diff
@@ -83,6 +83,33 @@ func TestDefaultFeedbackFallback(t *testing.T) {
...
+
+	url := backend.URL
+
+	r := getAPIResponse(url, 1, nil, testClient)
+
+	if !r.Error {
+		t.Errorf("Error attribute on response was not true")
+	}
+
+	if r.APIResponse.Optimal {
```
dandrabik, line 108: Should this be Optimal?, i.e. if a response times out, we have to act like there was no feedback to give.

Edit: I just looked at this code and remembered we broke that out that logic into a separate `Error` attribute. Ignore this.

dandrabik (approved):
Looks good to me. :+1:

services/comprehension/feedback-api-main/src/endpoint/endpoint.go
```diff
@@ -108,7 +109,7 @@ func Endpoint(context *gin.Context) {
 	var returnable_result APIResponse
 
 	for index, url := range urls {
-		results[index] = getAPIResponse(url, index, request_body)
+		results[index] = getAPIResponse(url, index, request_body, base_client)
```
dandrabik, line 112: So is the intention to pass a pointer to the single base_client? I think you'd need to use `&base_client`

https://goinbigdata.com/golang-pass-by-pointer-vs-pass-by-value/

happythenewsad (approved):

services/comprehension/feedback-api-main/src/endpoint/endpoint.go
anathomical: So it looks like declaring the incoming param as `client *http.Client` in the function signature means that passing the base variable name in the call automatically gets a pointer.  When I make the call with `&base_client` I get a double-pointer error:
Screen Shot 2021-10-18 at 1 38 47 PM
So I think this is correct, or alternatively I think I could rewrite the function signature without the `*` and then use `&` when making calls.

anathomical: Actually, testing reveals that the rewrite doesn't work.  Possibly because `base_client` is already assigned a pointer (since it is set to `&http.Client ...`)?

dandrabik: Discussed with Thomas, I missed that base_client is already set as a pointer `base_client = &http_client...`, so this should be satisfied.

PR #8407, opened 2021-10-18
LMS Bugfix - Fix call to our Geo API service

dandrabik:
## WHAT
We are getting errors for calls to Pointpin using their gem (this is our top error today). The gem uses [`http`](https://github.com/pointpin/pointpin-ruby/blob/master/lib/pointpin/configuration.rb#L4), which they seem to not support anymore (`https` calls work).

Removing the gem and making the `https` request ourselves.
## WHY
We want to fix this error. There is an added bonus of removing a gem dependency.
## HOW
Don't use their API wrapper gem, just make the `https` calls directly. Note, we could just specify the api_endpoint in the gem, but I'd rather just remove this external dependency.

### Notion Card Links
https://sentry.io/organizations/quillorg-5s/issues/2728230665/?project=11238&query=is%3Aunresolved

PR Checklist | Your Answer
------------ | -------------

### DIFF
--- a/services/QuillLMS/app/workers/ip_location_worker.rb
+++ b/services/QuillLMS/app/workers/ip_location_worker.rb
@@ -1,16 +1,34 @@
 class IpLocationWorker
   include Sidekiq::Worker
+  API_KEY = ENV["POINTPIN_KEY"]
+  BASE_URL = 'https://geo.pointp.in'
+
+  class PinpointAPIError < StandardError; end
 
   def perform(id, ip_address, blacklist = [])
-    @user = User.find(id)
-    location = Pointpin.locate(ip_address)
-    postcode = location ? location["postcode"] : nil
-    country_name = location ? location["country_name"] : nil
-    region_name = location ? location["region_name"] : nil
-    city_name = location ? location["city_name"] : nil
-    unless blacklist.include?(postcode)
-      new_ip_location = IpLocation.new
-      IpLocation.create(country: country_name, state: region_name, city: city_name, zip: postcode.to_i, user_id: @user.id)
+    user = User.find(id)
+    response = HTTParty.get(pinpoint_url(ip_address))
+
+    if !response.success?
+      raise PinpointAPIError.new("#{response.code}: #{response}")
     end
+
+    postcode = response["postcode"]
+
+    return if blacklist.include?(postcode)
+
+    IpLocation.create(
+      country: response["country_name"],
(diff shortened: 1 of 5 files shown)


services/QuillLMS/app/workers/ip_location_worker.rb
```diff
@@ -1,16 +1,34 @@
...
 
   def perform(id, ip_address, blacklist = [])
-    @user = User.find(id)
-    location = Pointpin.locate(ip_address)
-    postcode = location ? location["postcode"] : nil
-    country_name = location ? location["country_name"] : nil
-    region_name = location ? location["region_name"] : nil
-    city_name = location ? location["city_name"] : nil
-    unless blacklist.include?(postcode)
-      new_ip_location = IpLocation.new
```
anathomical, line 12: Wait, did this line just not do anything?  Haha.

anathomical (approved):
Always love dumping a dependency.

services/QuillLMS/spec/workers/ip_location_worker_spec.rb
```diff
@@ -5,30 +5,60 @@
...
+    let(:api_base) { 'https://testurl.com' }
+    let(:ip_address) {'12.34.56.78'}
+    let(:api_url) { "#{api_base}/#{api_key}/json/#{ip_address}"}
+    # This is needed for proper HTTParty stubbing
+    let(:response_headers) { {content_type: 'application/json'} }
 
     before do
-      allow(Pointpin).to receive(:locate) {
+      stub_const('IpLocationWorker::API_KEY', api_key)
+      stub_const('IpLocationWorker::BASE_URL', api_base)
```
happythenewsad, line 17: Why are you stubbing BASE_URL here?

happythenewsad (approved):
Approved, pending the lint fix.

services/QuillLMS/spec/workers/ip_location_worker_spec.rb
dandrabik: This may be overkill, but my aim here was to avoid tests from accidentally calling the actual API. This mock setup is for a very specific url, so if someone changes that url, e.g. by changing a param, the test would hit the actual API endpoint. If you mock the base_url, you remove that possibility.

Since mock API responses typically match real API responses, if you don't mock the base url, you could have a test that is calling the real API and passing without anyone noticing.

PR #8409, opened 2021-10-18
evidence/set-bing-timeout

anathomical:
## WHAT
Set a client-side timeout for Bing 
## WHY
So that we don't block on providing feedback for longer than is useful.  We have cases where this call takes 60 or 90 seconds, and we definitely don't want to sit around that long.  So we'll treat calls that take more than 5 seconds optimal.
## HOW
Just add a `timeout` param to the HTTParty call to Bing's spelling API.

### Notion Card Links
https://www.notion.so/quill/Evidence-feedback-sometimes-takes-between-30-sec-1-5-minutes-to-load-6587ba81246444b9b66d5ae6b4540d92

PR Checklist | Your Answer
------------ | -------------


dandrabik (approved):
Thanks for tightening this up.

PR #8420, opened 2021-10-21
diagnostic reports landing page

emilia-friedberg:
## WHAT
Update the diagnostic reports landing page to show the new pre- and post-tests.

## WHY
To have an overview that visually shows the relationship between the pre and post tests. 

## HOW
Update the backend query to rearrange the data in a way that makes more sense for the new view, and includes some methods that utilize our new skills and skill-groups tables. Then write the new frontend for the page.

### Screenshots
Screen Shot 2021-10-19 at 9 17 57 AM
Screen Shot 2021-10-19 at 9 17 20 AM

### Notion Card Links
https://www.notion.so/quill/Diagnostic-landing-page-c91a8eaaf65f454b8bdacc83750b05e0?d=972c7e504c7e470894028ce43c3a9bfe

PR Checklist | Your Answer
------------ | -------------
Have you added and/or updated tests? | NO - waiting to do that in a final PR to keep momentum/avoid having to rewrite them when I inevitably make changes throughout the rest of the project
Have you deployed to Staging? | NO - going to do that when I have Tom review it all at once


eadams17 (approved):
🎉

services/QuillLMS/app/models/activity_session.rb
```diff
@@ -505,6 +505,14 @@ def minutes_to_complete
     ((completed_at - started_at)/60).round
   end
 
+  def correct_skill_count
+    skills = activity&.skill_groups&.map(&:skills)&.flatten&.uniq
```
brendanshean: what about `activity.skills.distinct` since here's a `has_many :skills, through: :skill_groups`

services/QuillLMS/app/controllers/teachers/units_controller.rb
```diff
@@ -337,49 +320,96 @@ def score_info
...
       }
-      assignments.push(activity)
+      classrooms.push(classroom)
+    end
+    classrooms
+  end
+
+  private def record_with_aggregated_activity_sessions_and_skill_count(diagnostic_records, activity_id, classroom_id)
+    records = diagnostic_records.select { |record| record['activity_id'] == activity_id && record['classroom_id'] == classroom_id }
+    classroom_unit_ids = records.map { |record| record['classroom_unit_id'] }
```
brendanshean, line 393: What about a guard clause `return if records.empty?` on line 392 since the classroom_unit_ids and activity_sessions aren't relevant if there are no records.

services/QuillLMS/app/models/activity_session.rb
```diff
@@ -505,6 +505,14 @@ def minutes_to_complete
...
+    skills.reduce(0) do |sum, skill|
+      concept_results = ConceptResult.where(activity_session_id: id, concept_id: [skill.concept_ids])
+      sum += concept_results.length && concept_results.all? { |cr| cr.metadata['correct'] > 0 } ? 1 : 0
```
brendanshean: It looks like there's a method on concept_result `:correct?` that you could use:

`sum += concept_results.length && concept_results.all?(&:correct?) ? 1 : 0`

services/QuillLMS/spec/models/activity_spec.rb
```diff
@@ -33,6 +33,12 @@
 require 'rails_helper'
 
 describe Activity, type: :model, redis: true do
+  has_many :skill_group_activities
+  has_many :activities, through: :skill_group_activities
```
anathomical: Are these two lines doing anything here?

services/QuillLMS/app/controllers/teachers/units_controller.rb
```diff
@@ -337,49 +320,97 @@ def score_info
...
+        pre: record
+      }
+
+      if record['post_test_id']
+        grouped_record['pre'] = record_with_aggregated_activity_sessions_and_skill_count(diagnostic_records, record['activity_id'], record['classroom_id'])
+        post_test = record_with_aggregated_activity_sessions_and_skill_count(diagnostic_records, record['post_test_id'], record['classroom_id'])
+        grouped_record['post'] = post_test || { unit_template_id: ActivitiesUnitTemplate.find_by_activity_id(record['post_test_id'])&.unit_template_id }
+      else
+        grouped_record[:pre]['completed_count'] = ActivitySession.where(activity_id: record['activity_id'], classroom_unit_id: record['classroom_unit_id'], state: 'finished').size
+        grouped_record[:pre]['assigned_count'] = record['assigned_student_ids'].size
```
anathomical: Not necessarily a problem, but is there a reason we're using string-formatted keys in the `if` and symbol-formatted keys in the `else` here?  Looks like we're using string-formatted keys in most of the rest of this file.

anathomical (approved):

services/QuillLMS/app/models/activity_session.rb
brendanshean: Optional:  convert using a [pretzel colon](https://dcjones.medium.com/the-pretzel-colon-75df46dde0c7)

`concept_results.all? { |cr| cr.correct? }` becomes `concept_results.all?(&:correct?)`.  Instead of passing a block to the `all?` iterator, you can pass it the name of a proc which is more idiomatic Ruby.  It's similar conceptually to passing functions as arguments in JavaScript like `map` and `reduce`.

PR #8436, opened 2021-10-22
grammar API endpoint

happythenewsad:
## WHAT
Sibling to https://github.com/empirical-org/Empirical-Core/pull/8365 
Creates LMS Grammar endpoint, which intermediates between golang and GAPI

## WHY
So that GAPI and LMS are maximally decoupled, and so that we don't need to push complex config changes to gapi.

### Screenshots

### Notion Card Links
https://www.notion.so/quill/Make-Grammar-API-endpoint-generic-74f14400f587408c90c006cefce186da

PR Checklist | Your Answer
------------ | -------------
Self-Review: Have you done an initial self-review of the code below on Github? | ues


services/QuillLMS/engines/evidence/lib/evidence/evidence/grammar/feedback_assembler.rb
```diff
@@ -0,0 +1,14 @@
+module Evidence 
+  module Grammar
+    class FeedbackAssembler < Evidence::FeedbackAssembler
+      def self.error_to_rule_uid 
+        { } # specify in class descendents
```
happythenewsad: This mapping will be added in a separate ticket: https://www.notion.so/quill/Grammar-API-add-rule-lookup-and-update-rules-csv-with-snake-cased-rules-21e0f209b0e34f6a80b7571f92bb6f93

happythenewsad:
@dandrabik @anathomical nudge.

services/QuillLMS/engines/evidence/lib/evidence/evidence/grammar/client.rb
```diff
@@ -0,0 +1,35 @@
+module Evidence
+  module Grammar
+    class Client
+      API_TIMEOUT = 500
+      ALLOWED_PAYLOAD_KEYS = ['oapi_error', 'highlight']
+
+      class GAPIError < StandardError; end
```
dandrabik: I think we should refer to this API always as the `GrammarAPI`. Our app integrates with `Google` in many ways, so `GAPI` is more ambiguous than it needs to be.

services/QuillLMS/engines/evidence/lib/evidence/evidence/grammar/feedback_assembler.rb
```diff
@@ -0,0 +1,60 @@
+module Evidence 
+  module Grammar
+    class FeedbackAssembler < Evidence::FeedbackAssembler
+      def self.error_to_rule_uid
```
dandrabik, line 52: I think this should still be a constant. If you're trying to conform the subclass structure, this method could just return that constant, e.g.
```ruby
def self.error_to_rule_uid
  RULE_MAPPINGS
end

File: services/QuillLMS/engines/evidence/lib/evidence/evidence/opinion/feedback_assembler.rb [hunk 4]
```diff
@@ -1,50 +1,28 @@
 module Evidence 
   module Opinion
-    class FeedbackAssembler
-      OAPI_ERROR_TO_RULE_UID = {
```
dandrabik, line 4: Same comment as above, but it would be good to keep this constant.

services/QuillLMS/engines/evidence/lib/evidence/evidence/grammar/feedback_assembler.rb
```diff
@@ -0,0 +1,60 @@
...
+  module Grammar
+    class FeedbackAssembler < Evidence::FeedbackAssembler
+      def self.error_to_rule_uid 
+        {
+          "accept_vs_except" => '99257c28-9281-4ace-a131-952719f56a17',
+          "advise_vs_advice" => 'da27c3fe-67cc-471f-974e-b89743be6fc1',
+          "affect_vs_effect" => 'aad27eb2-9ca4-417a-82dc-ba31f7417ef1',
+          "allcaps" => 'a9e29976-3d99-4db2-a209-e215e2c3e34a',
+          "apart_vs_a_part" => '94a8b3e4-1aa9-4f9d-bf6a-09b75be68e77',
+          "articles_-_a_optimal" => '4c2962f1-03c4-449c-b821-8213688a9aa1',
```
dandrabik: I think we should remove the `-` from these names, e.g. `articles_a_optimal` instead of `articles_-_a_optimal`

services/QuillLMS/engines/evidence/lib/evidence/evidence/grammar/feedback_assembler.rb
```diff
@@ -0,0 +1,60 @@
...
+          "articles_-_an_optimal" => '04b69d41-f235-4a98-9fcd-1e8677961a2e',
+          "cite_vs_sight_vs_site" => '110da2cc-9acd-43e9-a137-774b3b994f08',
+          "contractions" => '112cc9e1-75db-47ab-9be1-7b9084838575',
+          "council_vs_counsel" => '31c3118f-0952-485b-8318-c3feba9fb5c4',
+          "further_vs_farther" => 'ba42ab74-e0f8-4abe-998e-e11ffeb060a5',
+          "in_regards_to" => 'b5068c7d-30f7-496b-832e-a6eca792ffe5',
+          "its_versus_it's_-_it's_optimal" => '8d8fe106-e0a5-440a-8bd4-4e34d8d0a16e',
```
dandrabik: Let's remove the `'` from this and replace it wish a `_`, e.g. `its_versus_it_s_it_s_optimal` instead of `its_versus_it's_-_it's_optimal`.

I'm less concerned about the actual name (If you have a better name feel free to use it), but we should restrict the keys to lowercase letters and underscores.

dandrabik (changes requested):
Overall looks good. :+1: Thanks for putting this together.

I'm requesting some small naming changes, the main one being that we should only use snake_case for the keys (only lowercase and underscores) to be consistent. There are few that have quotes and dashes, let's do a sweep and clean those up, then this should be good to go!

anathomical (approved):
I really dig the `FeedbackAssembler` abstraction.  Would love to expand that to other types of feedback generation (regex, automl, etc.)

dandrabik (approved):
:+1: LGTM. As we discussed, once the `.` are removed from the keys, ship it. :ship:

PR #8437, opened 2021-10-25
Modify Regex Rules to be optionally conditional

cissyxyu:
## WHAT
Curriculum team asked us to implement "Conditional Regex Rules".
These rules are configured such that **if** any of the incorrect sequences are caught in the response, **then** we will require the response to have a required sequence. However, if no incorrect sequences are present, we do not require the required sequences.

For reference, our current system only has non-conditional rules, which work like this: a response must always have a required sequence, whether or not it is also flagged on an incorrect sequence.

## WHY
Curriculum team runs into this case a lot where they want to be more granular in their regex rule grading (both in Evidence and Grammar/Connect/Diagnostic). 

## HOW
On the back end, I added the `conditional` field to `RegexRule` to flag whether a sequence should be graded conditionally or non-conditionally. I also changed Rule grading to grade differently based on this flag.

On the front end, I added a checkbox for Curriculum to flag whether a Rule should be graded conditionally or non-conditionally.

### Screenshots
Screen Shot 2021-10-25 at 2 58 18 PM

### Notion Card Links
https://www.notion.so/quill/Regex-w-both-Required-Incorrect-Sequences-a8e5353bc4e84a629a2ce965eb03a63f

PR Checklist | Your Answer
------------ | -------------


happythenewsad (approved):
This looks good - had a few questions that Cissy and I talked out. 

My only concern now is that the business logic for Regex rules is getting rather complex, and difficult to infer from the code. This might be a good opportunity to link to an explanatory document from a code comment.

services/QuillLMS/engines/evidence/db/migrate/20211022145011_add_dual_field_to_regex_rule.rb
```diff
@@ -0,0 +1,5 @@
+class AddDualFieldToRegexRule < ActiveRecord::Migration[5.1]
+  def change
+    add_column :comprehension_regex_rules, :conditional, :boolean, default: false
```
anathomical, line 3: It seems weird to keep a value that impacts all sequences on the table for individual sequences.  The UI doesn't seem to support the possibility, but the database would definitely allow some sequences to have `conditional` true and others not to.  However, from what I can tell, we want this to be all or nothing, which implies that this value should really be on the `Rule` model.  But that's weird because we really only want it for one of the Rule types, so it feels odd to add a whole column to a more generic record type.

anathomical (commented):
This looks like it would work as-written.  No real complaints about implementation other than my one comment which doesn't offer a better solution.

That said, I wonder if a more generic abstraction might make sense.  I wonder if, instead of making this special logic for regex (which we might want to do at least for now just for simplicity sake), it might not make more sense to consider a system by which Rules can be dependent on other Rules.  So instead of having a bunch of sequences with very special logic around a "conditional" check-box, you could just write two `Rules`.  One with inorrect_sequences in it, and another with required_sequences that depends on the first triggering.  Then we write some logic into Rules that processes dependent rule only when the "parent" rule triggers.

Might be more complex than is worth implementing at the moment, but thoughts?

services/QuillLMS/engines/evidence/db/migrate/20211022145011_add_dual_field_to_regex_rule.rb
cissyxyu: Yeah, I don't think either approach makes a ton of sense -- it doesn't feel exactly right to add the field on either of the models. I don't love my approach but I don't love the idea of a `conditional` field on the generic `Rule` database either, so this is my compromise for now.

cissyxyu:
> This looks like it would work as-written. No real complaints about implementation other than my one comment which doesn't offer a better solution.
> 
> That said, I wonder if a more generic abstraction might make sense. I wonder if, instead of making this special logic for regex (which we might want to do at least for now just for simplicity sake), it might not make more sense to consider a system by which Rules can be dependent on other Rules. So instead of having a bunch of sequences with very special logic around a "conditional" check-box, you could just write two `Rules`. One with inorrect_sequences in it, and another with required_sequences that depends on the first triggering. Then we write some logic into Rules that processes dependent rule only when the "parent" rule triggers.
> 
> Might be more complex than is worth implementing at the moment, but thoughts?

I think this approach could also work, although I can't think of any cases where we would want any other rule type (Grammar, Plagiarism, AutoML...) to have this dependent behavior. If we implemented the solution you suggest, we would have to do considerably more work around the grading logic to change all Rules to have this optional behavior, and then changing the rule-grading logic to grade rules interdependently.

Because Curriculum team is eager to get started on implementing some conditional regex ASAP, I vote that we go with this simpler solution for now and think about implementing more complex interdependency later. Thoughts?

anathomical (approved):
> I think this approach could also work, although I can't think of any cases where we would want any other rule type (Grammar, Plagiarism, AutoML...) to have this dependent behavior. If we implemented the solution you suggest, we would have to do considerably more work around the grading logic to change all Rules to have this optional behavior, and then changing the rule-grading logic to grade rules interdependently.
> 
> Because Curriculum team is eager to get started on implementing some conditional regex ASAP, I vote that we go with this simpler solution for now and think about implementing more complex interdependency later. Thoughts?

I'm basically imagining Curriculum wanting to add a Regex rule that triggers if, say, Plagiarism gets triggered, and conditionally override the default feedback.  But maybe we can just wait on that until it comes up?  I'll approve this implementation for now.

services/QuillLMS/engines/evidence/app/models/evidence/rule.rb
```diff
@@ -73,13 +73,22 @@ def determine_feedback_from_history(feedback_history)
...
       end
     end
 
+    def grade_sequences_separately(entry)
+      all_incorrect_sequences_passing?(entry) && at_least_one_required_sequence_passing?(entry)
+    end
+
+    def grade_sequences_conditionally(entry)
+      return true if all_incorrect_sequences_passing?(entry)
+      return at_least_one_required_sequence_passing?(entry)
```
dandrabik: You can remove the last return since Ruby always returns the last line.

services/QuillLMS/engines/evidence/app/models/evidence/rule.rb
```diff
@@ -141,18 +150,33 @@ def conjunctions
       prompts.map(&:conjunction)
     end
 
-    private def all_regex_rules_passing?(entry)
-      regex_rules.none? do |regex_rule|
+    def conditional
+      return nil if !regex? || regex_rules.empty?
+      return regex_rules.first.conditional?
```
dandrabik: I think depending on order as a flag here is a little fragile.

services/QuillLMS/engines/evidence/app/models/evidence/rule.rb
```diff
@@ -141,18 +150,33 @@ def conjunctions
       prompts.map(&:conjunction)
     end
 
-    private def all_regex_rules_passing?(entry)
-      regex_rules.none? do |regex_rule|
+    def conditional
```
dandrabik, line 146: So I think I might have misunderstood how this would be set up when we chatted. I thought this implementation would look like the flow below. This might be worth a chat (I might be misunderstanding this). 

- A rule has many regex_rules that can each either be `always` or `conditionally` run
- To run a rule:
  - Run all of its `always` regex_rules
    - If they `don't match`
      - Return `non_match`
    - If they `match`
        - Run all of its `conditional` regex_rules
          - If there are none
            - return  `match`
          - If the `conditional` regex_rules all `match`
            -  return `match`
          - If the `conditional` regex_rules all `don't match`
            -  return `non_match`

dandrabik (changes requested):
Nice progress on this!

I'm marking this as `Request Changes` since there's a piece of logic I think we should discuss (there's a chance I'm just misunderstanding the intended functionality). 

Also, It's probably faster to just chat through. I left a comment about how I expected the logic to work, so if you could read through that, it would be a good primer for our discussion.

dandrabik (approved):
:+1: LGTM. Thanks for making the changes and talking through this with me!

PR #8446, opened 2021-10-29
prefilter backend

happythenewsad:
## WHAT
Part 1 of the prefilter project

Includes the LMS prefilter API. 

Separate PRs will add:
- client side adapter
- rake task for populated new rules
- lambdas for the prefilters
- configure golang to use this new endpoint

## WHY
So that golang can hit an API for prefilters, using the same architecture as our existing APIs.

## HOW
new API 

### Screenshots

### Notion Card Links

PR Checklist | Your Answer
------------ | -------------

### DIFF
--- a/services/QuillLMS/engines/evidence/app/controllers/evidence/feedback_controller.rb
+++ b/services/QuillLMS/engines/evidence/app/controllers/evidence/feedback_controller.rb
@@ -2,11 +2,10 @@ module Evidence
   require 'json'
 
   class FeedbackController < ApiController
-    before_action :set_params, only: [:automl, :plagiarism, :regex, :spelling, :prefilter]
+    before_action :set_params, only: [:automl, :plagiarism, :regex, :spelling]
 
     def prefilter 
-      prefilter_check = Evidence::PrefilterCheck.new(@entry, @prompt)
-      return render :body => nil, :status => 404 if prefilter_check.feedback_object[:error]
+      prefilter_check = Evidence::PrefilterCheck.new(prefilter_params)
       render json: prefilter_check.feedback_object
     end
 
@@ -39,6 +38,10 @@ module Evidence
       return render :body => {:error => spelling_check.error }.to_json, :status => 500 if spelling_check.error.present?
       render json: spelling_check.feedback_object
     end
+
+    private def prefilter_params
+      params.require(:entry)
+    end 
 
     private def set_params
       @entry = params[:entry]
--- a/services/QuillLMS/engines/evidence/app/controllers/evidence/feedback_controller.rb
+++ b/services/QuillLMS/engines/evidence/app/controllers/evidence/feedback_controller.rb
@@ -2,7 +2,13 @@ module Evidence
   require 'json'
 
   class FeedbackController < ApiController
-    before_action :set_params, only: [:automl, :plagiarism, :regex, :spelling]
+    before_action :set_params, only: [:automl, :plagiarism, :regex, :spelling, :prefilter]
+
+    def prefilter 
+      prefilter_check = Evidence::PrefilterCheck.new(@entry, @prompt)
+      return render :body => nil, :status => 404 if prefilter_check.feedback_object[:error]
+      render json: prefilter_check.feedback_object
+    end
 
     def plagiarism
       rule = @prompt.rules&.find_by(rule_type: Evidence::Rule::TYPE_PLAGIARISM)
--- a/services/QuillLMS/engines/evidence/app/models/evidence/rule.rb
+++ b/services/QuillLMS/engines/evidence/app/models/evidence/rule.rb
@@ -12,16 +12,27 @@ module Evidence
       STATE_ACTIVE      = 'active',
       STATE_INACTIVE    = 'inactive'
     ]
+
+    TYPE_AUTOML       = 'autoML'
+    TYPE_GRAMMAR      = 'grammar'
+    TYPE_OPINION      = 'opinion'
+    TYPE_PLAGIARISM   = 'plagiarism'
+    TYPE_REGEX_ONE    = 'rules-based-1'
+    TYPE_REGEX_TWO    = 'rules-based-2'
+    TYPE_REGEX_THREE  = 'rules-based-3'
+    TYPE_SPELLING     = 'spelling'
+    TYPE_PREFILTER    = 'prefilter'
+
     TYPES = [
-      TYPE_AUTOML       = 'autoML',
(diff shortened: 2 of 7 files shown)


services/QuillLMS/engines/evidence/lib/evidence/evidence/prefilter_check.rb
```diff
@@ -0,0 +1,57 @@
+module Evidence
+  class PrefilterCheck
+    MINIMUM_WORD_COUNT = 3
+    OPTIMAL_RULE_UID = '' # TODO: populate
+
+    PREFILTERS = [
+      'this is a rule uid for TOO_LONG' => lambda do |text|
```
happythenewsad: Lambdas for each existing precheck would be stored inline here, since
- we don't want to marshal/unmarshal arbitrary ruby into Postgres
- doing so would be a departure from `Rule`'s normal behavior

services/QuillLMS/engines/evidence/lib/evidence/evidence/prefilter_check.rb
```diff
@@ -0,0 +1,57 @@
+module Evidence
+  class PrefilterCheck
+    MINIMUM_WORD_COUNT = 3
+    OPTIMAL_RULE_UID = '' # TODO: populate
```
happythenewsad, line 5: A rake task will populate this `Rule`, and the other prefilter rules and feedbacks.

services/QuillLMS/engines/evidence/lib/evidence/evidence/prefilter_check.rb
```diff
@@ -0,0 +1,53 @@
+module Evidence
+  class PrefilterCheck
+    attr_accessor :prefilter_rules
+    MINIMUM_WORD_COUNT = 3
+    OPTIMAL_RULE_UID = '' # TODO: populate
+
+    PREFILTERS = {
+      # TODO: populate
```
happythenewsad, line 8: Will be populated in subsequent PR.

services/QuillLMS/engines/evidence/spec/lib/prefilter_check_spec.rb
```diff
@@ -0,0 +1,58 @@
...
+      let!(:prefilter_rule1) do 
+        create(:evidence_rule, **rule_factory_overrides, uid: '123')
+      end
+
+      let(:non_prefilter_rule) { create(:evidence_rule, **rule_factory_overrides, rule_type: 'notPrefilter') }
+
+      context 'PREFILTERS query hit' do 
+        before do 
+          Evidence::PrefilterCheck && stub_const('Evidence::PrefilterCheck::PREFILTERS', {
+            '123' => ->(x) { false }
```
happythenewsad, line 25: i prefer using the keyword `lambda`, but Rubocop wants the arrow syntax for one-liners

services/QuillLMS/engines/evidence/app/controllers/evidence/feedback_controller.rb
```diff
@@ -34,6 +39,10 @@ def spelling
       render json: spelling_check.feedback_object
     end
 
+    private def prefilter_params
+      params.require(:entry)
+    end 
+
```
anathomical, line 45: Any reason not to just use the `set_params` `before_action` that the other endpoints use?  Obviously the `prefilter` doesn't need a lot of that data, but I kind of wonder if consistency isn't a value here.  Open to discussion.

services/QuillLMS/engines/evidence/app/models/evidence/rule.rb
```diff
@@ -12,15 +12,16 @@ class Rule < ApplicationRecord
...
+    TYPES = [
+      TYPE_AUTOML       = 'autoML',
+      TYPE_GRAMMAR      = 'grammar',
+      TYPE_OPINION      = 'opinion',
+      TYPE_PLAGIARISM   = 'plagiarism',
+      TYPE_REGEX_ONE    = 'rules-based-1',
+      TYPE_REGEX_TWO    = 'rules-based-2',
+      TYPE_REGEX_THREE  = 'rules-based-3',
+      TYPE_SPELLING     = 'spelling',
+      TYPE_PREFILTER    = 'prefilter'
```
anathomical: Not a huge deal, but it looks like we may be keeping these values in alphabetical order.  I wonder if we should add `TYPE_PREFILTER` just before `TYPE_PLAGIARISM`.

services/QuillLMS/engines/evidence/lib/evidence/evidence/prefilter_check.rb
```diff
@@ -0,0 +1,53 @@
...
+    end
+
+    def feedback_object
+      violated_rule = @prefilter_rules.find do |rule| 
+        next unless PREFILTERS[rule.uid]
+        !PREFILTERS[rule.uid].call(@entry) 
+      end
+      return default_response unless violated_rule
+
+      feedback = violated_rule.feedbacks.first
```
anathomical, line 37: Not for this PR, but I wonder if in the long term we actually want some sort of logic to live on the `Rule` model that returns the appropriate `Feedback` object.  I'm assuming that such logic is pretty universal and not tied to specific algorithms.  I could imagine something like `Rule.get_feedback()`.

anathomical (approved):
Looks good to me.  I like the lambda approach you've sketched out here.

services/QuillLMS/engines/evidence/lib/evidence/evidence/prefilter_check.rb
```diff
@@ -0,0 +1,53 @@
...
+      @prefilter_rules = Evidence::Rule.where(rule_type: Evidence::Rule::TYPE_PREFILTER).includes(:feedbacks)
+      
+    end
+
+    def default_response
+      {
+        feedback: '',
+        feedback_type: Evidence::Rule::TYPE_PREFILTER,
+        optimal: true,
+        response_id: '',
```
cissyxyu, line 22: I think we're using "0" for this in some places in the code. Not a big deal because it isn't used for anything yet, but maybe we should standardize whether this is empty string or "0".

cissyxyu (approved):

services/QuillLMS/engines/evidence/lib/evidence/evidence/prefilter_check.rb
happythenewsad: SpellingCheck and Regex both use `''` 

Want to raise in the eng. meeting? I'm happy to change all instances then, + documentation, if that's what the consensus indicates.

PR #8447, opened 2021-10-29
evidence/plagiarized-response-analysis

anathomical:
## WHAT
Add a rake task that can analyze feedback history for plagiarism.  NOTE: this code doesn't really need to be deployed or anything, but I figured I'd save it for posterity.  But that makes it NOT URGENT.
## WHY
We want to know how adding minimum edit distance to the plagiarism algorithm would impact its performance, so this script retroactively implements that logic against historical entries to see which ones that weren't being treated as plagiarism before might start being treated that way now
## HOW
Take a CSV input file of `FeedbackHistory` rows and related `plagiarism_text`, then for each row do the plagiarism analysis (using ten word windows over both `entry` and `plagiarism_text`), but checking the Levenshtein distance between each compared window pair.  For each entry, find the smallest Levenshtein window across all window pairs and record whether applying that Levenshtein distance would have resulted in a flag of "plagiarism" if the maximum distance was set at 5, 10, 15, or 20 characters.

Include some logic to bail on the comparisons early if the distance is going to be greater than 20 or if we've identified at least one window pair with a distance less than 5.

### Notion Card Links
https://www.notion.so/quill/Pull-data-from-Evidence-for-plagiarism-3bbc481c04ce4397abcae5fad3796e60

PR Checklist | Your Answer
------------ | -------------
Have you added and/or updated tests? |  No tests on script
Have you deployed to Staging? | No.  This was run entirely locally.


services/QuillLMS/engines/evidence/lib/tasks/plagiarized_responses.rake
```diff
@@ -28,4 +28,89 @@ namespace :plagiarized_responses do
...
+      entry_arr = entry.gsub(/[[:punct:]]/, '').downcase.split
+      plagiarism_arr = plagiarism_text.gsub(/[[:punct:]]/, '').downcase.split
+
+      slices = entry_arr.each_cons(match_minimum).map { |slice| slice.join(' ') }
+      plagiarism_slices = plagiarism_arr.each_cons(match_minimum).map { |slice| slice.join(' ') }
+      min_distance = nil
+      slices.each do |slice|
+        plagiarism_slices.each do |pslice|
+          min_distance = levenshtein_distance(slice, pslice, 21)
+          return min_distance if min_distance <= 5
```
happythenewsad: integers should be created as variables so that they are not 'magic'

happythenewsad (approved):

services/QuillLMS/engines/evidence/lib/tasks/plagiarized_responses.rake
anathomical: Good call out.

cissyxyu (approved):

services/QuillLMS/engines/evidence/lib/tasks/plagiarized_responses.rake
```diff
@@ -28,4 +28,90 @@ namespace :plagiarized_responses do
...
+  #     ON feedback_histories.prompt_id = comprehension_prompts.id
+  # JOIN comprehension_prompts_rules
+  #     ON feedback_histories.prompt_id = comprehension_prompts_rules.prompt_id
+  # JOIN comprehension_plagiarism_texts
+  #     ON comprehension_prompts_rules.rule_id = comprehension_plagiarism_texts.rule_id
+
+  task :test, [:input_csv_path] => :environment do |t, args|
+
+    # Levenshtein implementation lifted from https://stackoverflow.com/questions/16323571/measure-the-distance-between-two-strings-with-ruby
+    def levenshtein_distance(src, target, max = nil)
```
dandrabik: Googling a bit, Ruby 2.3 ships with this gem:
https://github.com/ruby/did_you_mean:

So you can do this:
```ruby
DidYouMean::Levenshtein.distance(string1, string2)
```
Running this 10K times takes 0.35 seconds, so probably better to use this performance tuned version.
Screen Shot 2021-11-04 at 4 02 09 PM

dandrabik (changes requested):
One small change requested, we should use the gem included with Ruby for the distance calculation for speed/confidence reasons, you can call it like this: `DidYouMean::Levenshtein.distance('string1', 'string2')`

services/QuillLMS/engines/evidence/lib/tasks/plagiarized_responses.rake
anathomical: I was going to point out that one of the reasons we rolled our own was to specify an optional `max` value that would allow us to short-circuit the calculation if the distance was going to be too long.  However, I tried running a benchmark with the built-in, and it looks like the optimization to the algorithm more than makes up for the difference, so using the built-in is faster than having the early bail-out.  I'll update the code to reference the built-in.

dandrabik (approved):
:+1:

PR #8449, opened 2021-10-29
Evidence/ fix activity completion check for skipping onboarding slides

eadams17:
## WHAT
add fix for logic that handles skipping onboarding slides in Evidence

## WHY
the original logic would skip past onboarding slides for students that completed 3 or more Evidence activities, but put students who have completed between 1 and 3 activities into a state where they couldn't start activities. A fix was put out for this, but the current state skips students past the reading/highlight phase after completing 3 or more activities. This accounts for both making sure students who have completed between 1 and 3 activities can start an activity, and makes sure students who completed 4 or more won't be skipped past the reading/highlighting phase

## HOW
just reverted to previous logic, tweaking to use `&&` operators instead of `||` and adding a check for a null `activityCompletionCount` value to account for first time users

### Screenshots

### Notion Card Links
https://www.notion.so/quill/Some-students-report-that-they-don-t-get-asked-to-complete-the-Evidence-highlighting-f72eed51d8af4ed8913841e92f80a094

PR Checklist | Your Answer
------------ | -------------
Have you added and/or updated tests? |  yes-- manually tested (note, we have a task for fixing test coverage for this file; refactoring to a functional component rendered the old tests unusable)


services/QuillLMS/client/app/bundles/Evidence/components/studentView/container.tsx
```diff
@@ -506,7 +506,7 @@ export const StudentViewContainer = ({ dispatch, session, isTurk, location, acti
 
   const className = `activity-container ${showFocusState ? '' : 'hide-focus-outline'} ${activeStep === READ_PASSAGE_STEP ? 'on-read-passage' : ''}`
 
-  if(!explanationSlidesCompleted) {
+  if(!explanationSlidesCompleted && (!activityCompletionCount || (activityCompletionCount && activityCompletionCount < 4))) {
```
anathomical: What if we just move this extra conditional up to line 75 where we set `explanationSlidesCompleted`.  That keeps the skip logic concentrated up near the key variables rather than spreading it out down here.  Or is that actually more confusing?

anathomical (changes requested):
Good catch on my breaking things.  What do you think of consolidating logic into the `explanationSlidesCompleted` variable up top?

services/QuillLMS/client/app/bundles/Evidence/components/studentView/container.tsx
```diff
@@ -59,7 +59,7 @@ const MINIMUM_STUDENT_HIGHLIGHT_COUNT = 2
 
 export const StudentViewContainer = ({ dispatch, session, isTurk, location, activities, handleFinishActivity, user, }: StudentViewContainerProps) => {
   const activityCompletionCount: number = parseInt(getParameterByName('activities', window.location.href, '0'));
```
emilia-friedberg: I think something funny is going on here - '0' is being submitted as a third argument to `getParameterByName` but that function only takes two arguments. My guess is that it's supposed to be a default, and if we rewrite this line as I have it below it'll clean up some of the conditional logic below.

` const activityCompletionCount: number = parseInt(getParameterByName('activities', window.location.href)) || 0;`

services/QuillLMS/client/app/bundles/Evidence/components/studentView/container.tsx
emilia-friedberg: If we give `activityCompletionCount` a default value, as suggested above, this line can just be `if (!explanationSlidesCompleted && activityCompletionCount > 3)`. Also, I think we should make whatever the minimum number of activities completed to skip the onboarding slide a constant at the top of the file for ease of changing - something like `ACTIVITY_COMPLETION_MAXIMUM_FOR_ONBOARDING`.

eadams17: good suggestion!

services/QuillLMS/client/app/bundles/Evidence/components/studentView/container.tsx
anathomical: This is probably something left-over from the reversion.  In the code that I wrote that broke things, I added a third param to the `getParameterByName` function, and you are right it was a default.  I bet that went away when Eric reverted my changes, but this line got missed.

anathomical (approved):

PR #8450, opened 2021-10-29
results summary and growth results summary page

emilia-friedberg:
## WHAT
The most exciting part of the new diagnostic reports - the results summary and growth results summary pages, which show how students performed broken down by skill and skill group and also compares the pre- and post-test results.

## WHY
This is a huge chunk of the work of this project!

## HOW
A lot of backend work that I am hoping to optimize later on and a ton, a ton, a ton of struggling with CSS and HTML to get the scrolling/overlapping dynamics of this page working correctly. But it does work now!

### Screenshots

### Notion Card Links
https://www.notion.so/quill/ea241f31a84c4b79ad56e83e99f7ee93?v=0c466a1f8c3846db94b8c959984cbb9d#7e4fef480c574465913528f841f214e6

PR Checklist | Your Answer
------------ | -------------
Have you added and/or updated tests? | NO - will be doing testing at the end because the project is so big and I will have a chance to do it while Tom reviews all the frontend
Have you deployed to Staging? | NO - pushing it up all at once for Tom's review


services/QuillLMS/client/app/bundles/Teacher/components/progress_reports/diagnostic_reports/diagnostics/growthResults.tsx
```diff
@@ -0,0 +1,160 @@
...
+    </div>
+    {cardContent}
+  </section>)
+}
+
+const GrowthResults = ({ passedStudentResults, passedSkillGroupSummaries, match, mobileNavigation, }) => {
+  const [loading, setLoading] = React.useState<boolean>(!passedStudentResults);
+  const [studentResults, setStudentResults] = React.useState(passedStudentResults || []);
+  const [skillGroupSummaries, setSkillGroupSummaries] = React.useState(passedSkillGroupSummaries || []);
+  const [openPopover, setOpenPopover] = React.useState({})
```
eadams17: would be nice to have interface defs for these pieces of state

eadams17 (approved):
looks great! I know this was a huge undertaking so big props to you for pushing through and getting this all together 🎉

services/QuillLMS/client/app/bundles/Teacher/components/progress_reports/diagnostic_reports/diagnostics/growthResults.tsx
emilia-friedberg: you're right - I've already made some edits to this component but I'll go through and add them in the PR that's branched off of this.

services/QuillLMS/app/controllers/concerns/growth_results_summary.rb
```diff
@@ -0,0 +1,151 @@
+module GrowthResultsSummary
+  extend ActiveSupport::Concern
+
+  NOT_PRESENT = 'Not present'
```
cissyxyu: Optional - you could put these constants in arrays that describe what they are like here: https://github.com/empirical-org/Empirical-Core/blob/8605d8e7c0031337b2c16a5e64980a89ee8c8cce/services/QuillLMS/engines/evidence/app/models/evidence/rule.rb#L11

services/QuillLMS/app/controllers/concerns/growth_results_summary.rb
```diff
@@ -0,0 +1,151 @@
...
+
+  NO_PROFICIENCY = 'No proficiency'
+  PARTIAL_PROFICIENCY = 'Partial proficiency'
+  PROFICIENCY = 'Proficiency'
+  GAINED_PROFICIENCY = 'Gained proficiency'
+  MAINTAINED_PROFICIENCY = 'Maintained proficiency'
+
+  extend self
+
+  def growth_results_summary(current_user, pre_test_activity_id, post_test_activity_id, classroom_id)
```
cissyxyu, line 7: Do we have rspec unit tests for this method and the method in ResultsSummary?

emilia-friedberg: Not yet, see note in PR checklist!

cissyxyu (approved):

services/QuillLMS/app/controllers/concerns/growth_results_summary.rb
```diff
@@ -0,0 +1,151 @@
...
+    end
+
+    {
+      student_results: student_results,
+      skill_group_summaries: @skill_group_summaries
+    }
+  end
+
+  private def set_pre_test_activity_sessions_and_assigned_students(activity_id, classroom_id)
+    unit_ids = @current_user.units.joins("JOIN unit_activities ON unit_activities.activity_id = #{activity_id}")
```
anathomical: I'm not sure this works as intended.  As-written, I'm pretty sure that this creates a merge of the `@current_user.units` results joined against ALL `unit_activity` records with the input `activity_id`.  That means that it probably joins `unit_activity` results that aren't specific to this user or even tied to their `units`.

Maybe this should be `@current_user.units.joins(:unit_activities).where(units: {unit_activities: {activity_id: activity_id}})`?

services/QuillLMS/app/controllers/concerns/growth_results_summary.rb
```diff
@@ -0,0 +1,151 @@
...
+    assigned_student_ids = classroom_units.map { |cu| cu.assigned_student_ids }.flatten.uniq
+    @pre_test_assigned_students = User.where(id: assigned_student_ids).sort_by { |u| u.last_name }
+    @pre_test_activity_sessions = ActivitySession.where(activity_id: activity_id, classroom_unit_id: classroom_units.ids, state: 'finished').order(completed_at: :desc).uniq { |activity_session| activity_session.user_id }
+  end
+
+  private def set_post_test_activity_sessions_and_assigned_students(activity_id, classroom_id)
+    unit_ids = @current_user.units.joins("JOIN unit_activities ON unit_activities.activity_id = #{activity_id}")
+    classroom_units = ClassroomUnit.where(unit_id: unit_ids, classroom_id: classroom_id)
+    assigned_student_ids = classroom_units.map { |cu| cu.assigned_student_ids }.flatten.uniq
+    @post_test_assigned_students = User.where(id: assigned_student_ids).sort_by { |u| u.last_name }
```
anathomical: `sort_by` is a nice work around to the lack of `last_name` in the database.  Sweet.

services/QuillLMS/app/controllers/concerns/growth_results_summary.rb
```diff
@@ -0,0 +1,151 @@
...
+    @post_test_activity_sessions = ActivitySession.where(activity_id: activity_id, classroom_unit_id: classroom_units.ids, state: 'finished').order(completed_at: :desc).uniq { |activity_session| activity_session.user_id }
+  end
+
+  private def student_results
+    @post_test_assigned_students.map do |assigned_student|
+      post_test_activity_session = @post_test_activity_sessions.find { |as| as.user_id == assigned_student.id }
+      pre_test_activity_session = @pre_test_activity_sessions.find { |as| as.user_id == assigned_student.id }
```
anathomical: These could be slightly faster to do look-ups on.  If, instead of storing `@post_test_activity_sessions` in an array, you could but it in a hash: `@post_test_activity_sessions = @post_test_activity_sessions.map { |session| [session.user_id, session] }.to_h` (passing an array of size-2 arrays to `to_h` gives you a hash that uses the first value as a key and the second value as the value for the hash).  That lets you do your lookups on `@post_test_activity_session[assigned_student.id]`.  That means you don't have to loop over the arrays for each record in `@post_test_assigned_students`, and thus it is slightly faster.  (Takes longer to write the original value, but makes it faster to look up specific sections of it.)

services/QuillLMS/app/controllers/concerns/growth_results_summary.rb
```diff
@@ -0,0 +1,151 @@
...
+      else
+        { name: assigned_student.name }
+      end
+    end
+  end
+
+  private def data_for_skill_by_activity_session(activity_session_id, skill)
+    concept_results = ConceptResult.where(activity_session_id: activity_session_id, concept_id: [skill.concept_ids])
+    number_correct = concept_results.select(&:correct?).length
+    number_incorrect = concept_results.reject { |cr| cr.correct? }.length
```
anathomical: Should be able to use `reject(&:correct?)` here the same way you used it for `select` immediately above.

services/QuillLMS/app/controllers/concerns/results_summary.rb
```diff
@@ -0,0 +1,117 @@
...
+      classroom_units = ClassroomUnit.where(unit_id: unit_ids, classroom_id: classroom_id)
+      assigned_student_ids = classroom_units.map { |cu| cu.assigned_student_ids }.flatten.uniq
+      @assigned_students = User.where(id: assigned_student_ids).sort_by { |u| u.last_name }
+      @activity_sessions = ActivitySession.where(activity_id: activity_id, classroom_unit_id: classroom_units.ids, state: 'finished').order(completed_at: :desc).uniq { |activity_session| activity_session.user_id }
+    end
+  end
+
+  private def student_results
+    @assigned_students.map do |assigned_student|
+      activity_session = @activity_sessions.find { |as| as.user_id == assigned_student.id }
```
anathomical: Same note as above about this being a faster look-up if `@activity_sessions` is a hash instead of an array.

services/QuillLMS/app/controllers/concerns/results_summary.rb
```diff
@@ -0,0 +1,117 @@
...
+      else
+        { name: assigned_student.name }
+      end
+    end
+  end
+
+  private def data_for_skill_by_activity_session(activity_session_id, skill)
+    concept_results = ConceptResult.where(activity_session_id: activity_session_id, concept_id: [skill.concept_ids])
+    number_correct = concept_results.select(&:correct?).length
+    number_incorrect = concept_results.reject { |cr| cr.correct? }.length
```
anathomical: Again, this could probably just be `reject(&:correct?)`

anathomical (changes requested):
One call-out where I think the query you have is too permissive.

A couple of other places where there could be more optimization, but I'll leave those changes up to your discretion.  The main thing I care about is that query.

services/QuillLMS/app/controllers/concerns/growth_results_summary.rb
emilia-friedberg: Eek, thanks for calling this out. My limited local dataset obviously didn't reveal that issue and it didn't occur to me! These methods have undergone some change since this PR but this hasn't been corrected, so I'm going to make that change to the last PR for this project to minimize merging weirdness if that's okay with you.

emilia-friedberg:
@anathomical addressed all your comments in this PR here: https://github.com/empirical-org/Empirical-Core/pull/8488

anathomical (approved):

PR #8462, opened 2021-11-03
Log Google users out if their auth credential is expired

cissyxyu:
## WHAT
If Google users have an expired auth credential, we should log them out and force them to log back in, which resets their access token. 

## WHY
So teachers don't run into the situation where they can't perform Google actions even though they're logged into Quill.

## HOW
Check the expiration date of the auth credential directly instead of just estimating it.

### Screenshots

### Notion Card Links
https://www.notion.so/quill/Log-Google-users-out-if-their-access-token-has-expired-fea79aa0fee045f387974c8bdb28cc52

PR Checklist | Your Answer
------------ | -------------


services/QuillLMS/app/controllers/application_controller.rb
```diff
@@ -147,7 +147,7 @@ def non_standard_route_redirect?(route)
 
     # Assuming that the refresh_token expires at (current_user.auth_credential.created_at  + 6 months),
     # we can reset the session whenever (Time.now > (current_user.auth_credential.created_at + 5 months))
-    return reset_session if current_user.google_id && current_user.auth_credential && Time.now > (current_user.auth_credential.created_at + 5.months)
+    return reset_session if current_user.google_id && current_user.auth_credential && (Time.now > (current_user.auth_credential.expires_at) || current_user.auth_credential.expires_at.nil?)
```
brendanshean, line 150: I think you could consolidate the logic and avoid violating the Law of Demeter with:
`return reset_session if current_user.google_id && !current_user.google_authorized?`

brendanshean (approved):
Just a minor code suggestion, but otherwise looks good.  Thank you for doing this!

anathomical (approved):

PR #8464, opened 2021-11-04
FeedbackAssembler specs, enhancements

happythenewsad:
## WHAT
- The FeedbackAssembler abstract parent class had Opinion-specific logic, which would have caused grammar API to break if traffic were going to it. Fixed the bug, and added specs.
- point golang to the new LMS endpoint.

## WHY
Bugs should be fixed. 

### Screenshots

### Notion Card Links

PR Checklist | Your Answer
------------ | -------------
Spec Review: Have you reviewed the spec and ensured this meets requirements and/or matches design mockups? | n/a


services/comprehension/feedback-api-main/src/endpoint/endpoint.go
```diff
@@ -58,20 +58,31 @@ func GetOpinionDomain() (string) {
...
+	var maybe_domain = os.Getenv("grammar_domain")
+
+	var grammar_domain = "https://www.quill.org"
+
+	if len(maybe_domain) > 0 {
+		grammar_domain = maybe_domain
+	}
+	return grammar_domain
+}
+
```
anathomical: Why not re-use the pattern we have for other Evidence-engine-based APIs.  Like `automl_api` just uses `lms_domain`.

services/QuillLMS/engines/evidence/app/controllers/evidence/grammar_controller.rb
```diff
@@ -2,8 +2,8 @@ module Evidence
   class GrammarController < ApiController
 
     def fetch
-      oapi_client = Grammar::Client.new(entry: params['entry'], prompt_text: params['prompt_text'])
-      render json: Grammar::FeedbackAssembler.run(oapi_client.post)
+      grammar_client = Grammar::Client.new(entry: params['entry'], prompt_text: params['prompt_text'])
+      render json: Grammar::FeedbackAssembler.run(grammar_client.post)
     end
   end
 end
```
anathomical: What if we just put this (and the Opinion API Controller) into the `FeedbackController`.  There are already endpoints there for different feedback types.  Consolidation of these single-action controllers is probably easier to keep track of.

services/QuillLMS/engines/evidence/lib/evidence/evidence/feedback_assembler.rb
```diff
@@ -33,6 +33,9 @@ def self.default_payload
         'rule_uid': ''
       }  
     end
-  end
-    
+
+    def self.error_name 
+      'abstract_error'
```
anathomical: Maybe just `raise NoMethodError` or something here in the abstract class?  That'll actually cause errors if someone fails to do a sub-class implementation which I think is the behavior we want?

anathomical (changes requested):
I like the implementation in general, and appreciate the tests a ton.  A couple of thoughts on consolidation that might make things simpler?  Some of these probably should have been recommended back when the original OAPI changes were merged.

cissyxyu (approved):

happythenewsad:
@anathomical @cissyxyu Thanks for reviewing. This PR is blocked on https://www.notion.so/quill/blocked-Set-up-staging-Spell-model-servers-562b17d99b0b4266930c0fb227765420

I will address your changes once the PR becomes unblocked.

cissyxyu (approved):

anathomical (approved):

PR #8470, opened 2021-11-04
cms/update-sidekiq-config

anathomical:
## WHAT
Modify Sidekiq config to use namespaced DB0 in Redis
## WHY
Apparently the Redis provider that we use in Heroku don't allow connection to Redis "DBs" higher than 0.  The `/2` in the old code made Sidekiq attempt to connect to DB2, which worked fine in AWS, but not in Heroku.

I have literally no idea why this worked on Monday November 1, 2021 when we first swapped over to Heroku only to break the morning of Tuesday November 2, 2021.  Or why it was continuing to work on Staging prior to me deploying this change there to test it.  A discussion with the support team for the Redis provider revealed no reason that these would have ever worked.

Anyway, this update lets us isolate Sidekiq data on DB0, which is guaranteed to be accessible via our provider.
## HOW
Connect to the default DB (0), but add a namespace in order to make sure we don't collide with any cache or other keys set by Ruby.

### Notion Card Links
https://www.notion.so/quill/RFC-For-Migrating-CMS-Webservers-from-ElasticBeanstalk-to-Heroku-d7740f01090e4c3380578dd42edabd19

PR Checklist | Your Answer
------------ | -------------
Have you added and/or updated tests? |  No tests on config


brendanshean (approved):
Good sleuthing on this!

happythenewsad (approved):

PR #8475, opened 2021-11-05
update golang for prefilter url, adds prefilter lambdas

happythenewsad:
## WHAT
- completes the prefilter project (besides deletion of old client side code)
- points golang to the new prefilter api
- adds lambdas for all prefilters. profanity and sentence checking required quite a bit of new code
- the profanity checkers bad words list is the exact same list previously used in our client side library

## WHY
So that prefilters can be handle via LMS, rather than client side code. This has various benefits. One big behavior change is that prefilter violations will now be tracked as attempts, same as automl or grammar.

## HOW
- add pragmatic_segmenter gem for sentence boundary detection
- add profanity list and profanity checker

## ROLLOUT PLAN
- deploy this PR to LMS. (no user facing change; this configuration is stable)
- deploy golang (users will now encounter two sets of prefilter logic - server side and client side.)
- deploy the client side PR 

### Screenshots

### Notion Card Links
https://www.notion.so/quill/Save-the-pre-filter-response-data-to-the-backend-5b096b59aada4d4e8e52663d02cc48cd

PR Checklist | Your Answer
------------ | -------------


happythenewsad:
You can demo the full integration on pkong.quill.org now (any evidence test play)

services/QuillLMS/engines/evidence/config/initializers/bad_words.rb
```diff
@@ -0,0 +1,3 @@
+path = File.join(Evidence::Engine.root, 'config/initializers/bad_words.json')
```
happythenewsad, line 1: Copied from: https://github.com/web-mech/badwords/blob/master/lib/lang.json

services/QuillLMS/engines/evidence/lib/evidence/evidence/profanity.rb
```diff
@@ -0,0 +1,26 @@
+module Evidence 
+  class Profanity 
+    def self.profane?(word)
+      return false unless word.respond_to?(:length) && word.length > 1
+      word.downcase! 
+
+      BAD_WORDS.find do |badword|
+        stripped_badword = badword.gsub('*', '')
```
happythenewsad: This logic is because the bad words list uses a glob-style syntax that cannot be shoved directly into `Regexp.new` and work as expected.

services/QuillLMS/engines/evidence/lib/evidence/evidence/prefilter_check.rb
```diff
@@ -1,17 +1,22 @@
...
   class PrefilterCheck
     attr_accessor :prefilter_rules
     MINIMUM_WORD_COUNT = 3
-    OPTIMAL_RULE_UID = '' # TODO: populate
+    OPTIMAL_RULE_UID = 'a7410335-5dae-4fc7-832a-ce9cf8d5dffb'
 
+    # When a prefilter lambda identifies a violation, it returns true
     PREFILTERS = {
-      # TODO: populate
+      'f576dadc-7eec-4e27-8c95-7763e6550141' => ->(entry) { entry.match?(/\?$/) },
```
happythenewsad: These UIDs are now consistent across staging and prod DBs.

services/QuillLMS/engines/evidence/lib/evidence/evidence/profanity.rb
```diff
@@ -0,0 +1,26 @@
+module Evidence 
+  class Profanity 
+    def self.profane?(word)
```
happythenewsad: Good opportunity to use a Bloom filter as a future optimization

services/QuillLMS/engines/evidence/lib/evidence/evidence/prefilter_check.rb
brendanshean: These keys are rather opaque.  It might be worth extracting them as constants like you have with `OPTIMAL_RULE_ID`.  Then you can refer to the constants in [specs](https://github.com/empirical-org/Empirical-Core/pull/8475/files#diff-9666bfbd5482cf107a573c604f3e9b5ad0c1f0ccd14a58175e1eb74d8f07d080R23)

services/QuillLMS/engines/evidence/lib/evidence/evidence/prefilter_check.rb
```diff
@@ -45,9 +51,17 @@ def feedback_object
       )
     end
 
-    def self.to_word_array(entry)
+    def self.words(entry)
       entry.split(' ')
     end
 
+    def self.word_count(entry)
+      words(entry).filter{ |s| !s.empty? }.count
```
brendanshean: What about `words(entry).reject(&:empty?).count`?

services/QuillLMS/engines/evidence/lib/evidence/evidence/profanity.rb
```diff
@@ -0,0 +1,26 @@
...
+      return false unless word.respond_to?(:length) && word.length > 1
```
brendanshean: is the `respond_to?(:length)` checking for objects outside of `String` (maybe symbols)?  If not, would word.is_a?(String) be more appropriate?  Otherwise, it seems like you would need similar methods for `word.respond_to?(:downcase)` and `word.respond_to?(:match)`

services/QuillLMS/engines/evidence/lib/evidence/evidence/profanity.rb
```diff
@@ -0,0 +1,26 @@
...
+      word = word.downcase.gsub(/[.!?]/, '')
+
+      BAD_WORDS.find do |badword|
+        stripped_badword = badword.gsub('*', '')
+        
+        if badword[0] == '*' && badword[-1] == '*'
+          regex = ::Regexp.new(stripped_badword)
+          return word.match?(regex)
```
brendanshean: I think `return` in these clauses is unnecessary

services/QuillLMS/engines/evidence/lib/evidence/evidence/profanity.rb
```diff
@@ -0,0 +1,26 @@
+module Evidence 
+  class Profanity 
+    def self.profane?(word)
+      return false unless word.respond_to?(:length) && word.length > 1
+      word = word.downcase.gsub(/[.!?]/, '')
+
+      BAD_WORDS.find do |badword|
+        stripped_badword = badword.gsub('*', '')
+        
+        if badword[0] == '*' && badword[-1] == '*'
```
brendanshean: Would `badword.start_with?('*') && badword.end_with?('*')` be more readable?

services/QuillLMS/engines/evidence/spec/controllers/evidence/feedback_controller_spec.rb
```diff
@@ -19,15 +19,38 @@ module Evidence
...
-
-        expect(response.status).to eq 200
+      let!(:profanity_rule) do 
+        create(:evidence_rule, rule_type: 'prefilter', uid: 'fdee458a-f017-4f9a-a7d4-a72d1143abeb')
+      end 
+
+      context 'profanity violation' do 
+        it 'should return correct rule uid' do 
+          post "prefilter", :params => ({ 
+            :entry => "nero was an ahole",
```
brendanshean, line 29: :fire:

services/QuillLMS/engines/evidence/spec/lib/prefilter_check_spec.rb
```diff
@@ -12,42 +12,61 @@ module Evidence
       end
     end
 
-    describe '#feedback_object' do
-      let!(:prefilter_rule1) do 
-        create(:evidence_rule, **rule_factory_overrides, uid: '123')
+    describe '#sentence_count' do 
+      it 'should enumerate sentences' do 
+        text = <<~HEREDOC
+          Hello world. My name is Mr. Smith. I work for the U.S. Government and I live in the U.S. I live in New York.
```
brendanshean, line 18: I like this test. Succinctly shows the segmenter in action.

services/comprehension/feedback-api-main/src/endpoint/endpoint_test.go
```diff
@@ -226,6 +226,7 @@ func TestIdentifyUsedFeedbackIndex(t *testing.T) {
 	feedbacks[1] = optimal_response
 	feedbacks[2] = usable_used_response
 	feedbacks[3] = usable_unused_response
+	feedbacks[4] = usable_unused_response
```
brendanshean, line 229: What's the reason for storing this twice?

services/comprehension/feedback-api-main/src/endpoint/endpoint.go
```diff
@@ -63,18 +63,19 @@ func AssembleUrls() ([api_count]string) {
...
-		automl_api = fmt.Sprintf("%s/api/v1/evidence/feedback/automl.json", lms_domain)
-		plagiarism_api = fmt.Sprintf("%s/api/v1/evidence/feedback/plagiarism.json", lms_domain)
-		sentence_structure_regex_api = fmt.Sprintf("%s/api/v1/evidence/feedback/regex/rules-based-1.json", lms_domain)
-		post_topic_regex_api         = fmt.Sprintf("%s/api/v1/evidence/feedback/regex/rules-based-2.json", lms_domain)
-		typo_regex_api               = fmt.Sprintf("%s/api/v1/evidence/feedback/regex/rules-based-3.json", lms_domain)
-		spell_check_bing             = fmt.Sprintf("%s/api/v1/evidence/feedback/spelling.json", lms_domain)
-
-		grammar_check_api = "https://quill.spell.services/Quill/grammar/predict"
-		opinion_check_api = fmt.Sprintf("%s/", opinion_domain)
+		automl_api 										= fmt.Sprintf("%s/api/v1/evidence/feedback/automl.json", lms_domain)
```
brendanshean: The formatting here seems inconsistent.

brendanshean (commented):
Nice work on this.  Very easy to follow.

Some minor refactoring comments.  I think it's worth getting another review on this since I haven't yet internalized the overall flow of the prefilters.

services/QuillLMS/engines/evidence/lib/evidence/evidence/prefilter_check.rb
dandrabik: Seconding this. This section of code would benefit from something having a name, the uid and/or the lambda. It would make it much easier to scan to understand what checks are being run.

services/QuillLMS/engines/evidence/config/initializers/bad_words.rb
```diff
@@ -0,0 +1,3 @@
...
+io = File.open(path).read 
+BAD_WORDS = JSON.parse(io)['words'].map(&:downcase)
```
dandrabik: Let's wrap this all in a module so it's not in the global scope, e.g. `Evidence::BAD_WORDS` or `BadWords::ALL` or whatever you feel is appropriate.

services/QuillLMS/engines/evidence/lib/evidence/evidence/prefilter_check.rb
```diff
@@ -1,17 +1,22 @@
+require 'pragmatic_segmenter'
+
 module Evidence
   class PrefilterCheck
     attr_accessor :prefilter_rules
```
dandrabik: You're defining the accessor, but not using it in the code (you directly call `@prefilter_rules`). Just to be consistent, we should either:
1. Remove `attr_accessor :prefilter_rules`
2. Use `attr_reader :prefilter_rules, :entry` and use `prefilter_rules` and `entry` throughout the code (non-instance vars)

services/QuillLMS/engines/evidence/lib/evidence/evidence/prefilter_check.rb
```diff
@@ -45,9 +51,17 @@ def feedback_object
...
-    def self.to_word_array(entry)
+    def self.words(entry)
       entry.split(' ')
     end
 
+    def self.word_count(entry)
+      words(entry).filter{ |s| !s.empty? }.count
+    end
+
+    def self.sentence_count(entry)
```
dandrabik, line 67: Take or leave. It's a judgement call of when to encapsulate logic, but there's enough here in class methods and lambdas, that you might consider breaking the pre-check logic into it's own class or classes, e.g. by making an entry object and passing that to the lambdas.
```ruby
entry = PreFilterEntry.new(entry_text)
UID -> (entry) { entry.contains_profanity? }
UID2 -> (entry) { entry.multiple_sentences? }
# or via meta programming and send
UID => :contains_profanity?,
UID2 => :multiple_sentences?,

File: services/QuillLMS/engines/evidence/lib/evidence/evidence/profanity.rb (hunk 4 above)
Comment (dandrabik) [inline services/QuillLMS/engines/evidence/lib/evidence/evidence/profanity.rb]: This might be a premature optimization (this may be super fast as is), but do we need to check for profanity in the entry word by word? With 450 "bad words", if there are 50 words in an entry, that's 50*450 or 22,500 comparisons. vs. 450 comparisons if you just checked for profane words in the entire entry at once.

Could we do something like this to trim everything that isn't in the entry and do the deeper inspection ONLY on items in the entry.

```ruby
words_in_entry = bad_words.gsub('*','').select {|word| entry.include?(word) } 
# then deeper check on words_in_entry

File: services/QuillLMS/engines/evidence/lib/evidence/evidence/profanity.rb [hunk 16]
```diff
@@ -0,0 +1,26 @@
+module Evidence 
+  class Profanity 
+    def self.profane?(word)
+      return false unless word.respond_to?(:length) && word.length > 1
+      word = word.downcase.gsub(/[.!?]/, '')
+
+      BAD_WORDS.find do |badword|
+        stripped_badword = badword.gsub('*', '')
```
dandrabik: There's enough logic in this loop, you might consider breaking it into its own method, i.e.
```ruby
BAD_WORDS.find {|bad_word| word_match?(bad_word, word)}

File: services/QuillLMS/engines/evidence/spec/lib/prefilter_check_spec.rb [hunk 17]
```diff
@@ -12,42 +12,61 @@ module Evidence
...
+          rule_uid: 'fdee458a-f017-4f9a-a7d4-a72d1143abeb'
+        }
+      ]
 
-        it 'returns a valid payload with feedback' do
-          feedback = create(:evidence_feedback, text: 'lorem ipsum', rule_id: prefilter_rule1.id)
-          prefilter_check = Evidence::PrefilterCheck.new('example entry')
-          response = prefilter_check.feedback_object
-          expect(response[:feedback]).to eq feedback.text
+      violations.each do |violation|
```
dandrabik, line 48: Good test coverage!

dandrabik (changes requested):
Overall, this looks good. I left a few refactoring and optimization comments that feel like enough to flag as "request changes". I don't think any of them are major, so it might be worth looking at today (I'm out tomorrow and Friday)

services/QuillLMS/engines/evidence/lib/evidence/evidence/profanity.rb
happythenewsad: It's pretty fast currently, and most entries are far less than 50 words. I prefer to use a Bloom filter if/when we want to optimize.

services/QuillLMS/engines/evidence/lib/evidence/evidence/prefilter_check.rb
happythenewsad: Passing on this for now - it looks much more legible after implementing your other suggestion of associating rule UIDs with constants.

happythenewsad:
@dandrabik I addressed or responded to all your feedback. I'm halfway through addressing Brendan's feedback.

services/QuillLMS/engines/evidence/lib/evidence/evidence/profanity.rb
dandrabik: @happythenewsad I don't believe this one was addressed. What are your thoughts on this? Most students won't write profanity, so the most common run case will be running through all of the words.

happythenewsad: @dandrabik Sorry for the confusion. I self-commented a reply (above), but it was displaying only for me until I hit "Start a review" just now. Hopefully you see it now. But to reiterate, I think we should ship it as is, or do a once-and-for-all optimization (Bloom filter).

services/comprehension/feedback-api-main/src/endpoint/endpoint_test.go
happythenewsad: By convention, golang returns the default automl response if everything else is optimal. And the existing code makes assumptions about the relative precedence of the automl API. Thus, we need to this massaging.

services/comprehension/feedback-api-main/src/endpoint/endpoint.go
happythenewsad: ~Notice the line change. It is consistent, but github's diffing UI makes it appear so. Check the raw file in an IDE to see what I mean.~ 

My mistake - this displays correctly in VSCode, but not `less`, which indicates there is a formatting issue.

brendanshean: Screen Shot 2021-11-11 at 11 59 31 AM

It looks weird in my IDE too.

happythenewsad: Fixed!

dandrabik (approved):
:+1: Approved. (I just merged my optimization in).

PR #8479, opened 2021-11-08
Fix nil error edge case on some google users

cissyxyu:
## WHAT
Some google users are seeing a nil error when logging into their accounts because their `expires_at` value is nil.

## WHY
I had accidentally put the two function calls in reverse order - the nil check came after the call on the value, resulting in the nil check not actually working.                      

## HOW
Put the nil check before the value call.

### Screenshots

### Notion Card Links

PR Checklist | Your Answer
------------ | -------------

### DIFF
--- a/services/QuillLMS/app/controllers/application_controller.rb
+++ b/services/QuillLMS/app/controllers/application_controller.rb
@@ -147,7 +147,7 @@ class ApplicationController < ActionController::Base
 
     # Assuming that the refresh_token expires at (current_user.auth_credential.created_at  + 6 months),
     # we can reset the session whenever (Time.now > (current_user.auth_credential.created_at + 5 months))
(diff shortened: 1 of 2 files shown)


happythenewsad (approved):

brendanshean:
@cissyxyu I think this bug is a good example of the need to consolidate complicated logic, see my recommendation on the last [PR](https://github.com/empirical-org/Empirical-Core/pull/8462#discussion_r742219342)

cissyxyu:
> @cissyxyu I think this bug is a good example of the need to consolidate complicated logic, see my recommendation on the last [PR](https://github.com/empirical-org/Empirical-Core/pull/8462#discussion_r742219342)

You're right - I had chosen not to go with your suggestion of using `google_authorized?` because it used a slightly different logic from what I was looking for (it added six months to the expiration date instead of using the date). But I think a better approach would have been to add some consolidation logic to my code and made it another method in `auth_credential`.

PR #8486, opened 2021-11-10
add evidence completion button before ActivityFollowup

happythenewsad:
## WHAT
Adds 'Complete Activity' button after an activity is completed, but before the user is routed to the ActivityFollowup component. 

## WHY
This is the intended UX.

## HOW
Re-arrange callbacks, make a new button component.

### Screenshots

Screen Shot 2021-11-10 at 12 25 44 PM

### Notion Card Links
[(Please provide links to any relevant Notion card(s) relevant to this PR.)](https://www.notion.so/quill/b417e5b89fe04af89cba62d4198f7031?v=77f294a5b0df466e98633244de7ebfef&p=02c5cb240f91444c8ec71a41b389b6b2)

PR Checklist | Your Answer
------------ | -------------
Have you added and/or updated tests? |  tested. manually
Spec Review: Have you reviewed the spec and ensured this meets requirements and/or matches design mockups? | Tom is evaluating the styling.

### DIFF
--- a/services/QuillLMS/client/app/bundles/Evidence/components/studentView/container.tsx
+++ b/services/QuillLMS/client/app/bundles/Evidence/components/studentView/container.tsx
@@ -77,6 +77,7 @@ export const StudentViewContainer = ({ dispatch, session, isTurk, location, acti
   const [activeStep, setActiveStep] = React.useState(shouldSkipToPrompts ? READ_PASSAGE_STEP + 1: READ_PASSAGE_STEP)
   const [activityIsComplete, setActivityIsComplete] = React.useState(false)
   const [activityIsReadyForSubmission, setActivityIsReadyForSubmission] = React.useState(false)
+  const [completeButtonClicked, setCompleteButtonClicked] = React.useState(false)
   const [completedSteps, setCompletedSteps] = React.useState(defaultCompletedSteps)
   const [showFocusState, setShowFocusState] = React.useState(false)
   const [startTime, setStartTime] = React.useState(Date.now())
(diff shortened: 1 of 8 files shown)


services/QuillLMS/client/app/bundles/Evidence/components/studentView/container.tsx
```diff
@@ -36,7 +36,6 @@ interface StudentViewContainerProps {
 interface StudentViewContainerState {
   activeStep?: number;
   activityIsComplete: boolean;
-  activityIsReadyForSubmission: boolean;
```
happythenewsad, line 39: This state var was not used for anything, so I removed it.

services/QuillLMS/client/app/bundles/Evidence/helpers/containerRenderHelpers.tsx
```diff
@@ -131,6 +143,7 @@ export const renderPromptSteps = ({
   return (<div className="prompt-steps">
     {renderDirectionsSectionAndModal({ className: 'hide-on-mobile', closeReadTheDirectionsModal, activeStep, doneHighlighting, showReadTheDirectionsModal, activities })}
     {steps}
+    {activityIsComplete && renderCompletionButton()}
```
emilia-friedberg: I think you should pass the `completionButtonCallback` and the `activityIsComplete` props into `PromptStep` and then use them [here](https://github.com/empirical-org/Empirical-Core/blame/production/services/QuillLMS/client/app/bundles/Evidence/components/studentView/promptStep.tsx#L293). It looks like the logic at one point was supposed to do something similar with the `everyOtherStepCompleted ? 'Done : ...` thing but evidently that stopped working at some point. It might be worth just trying to fix that logic, but alternately in there you could just change the `onClick` handler based on the value of `activityIsComplete` and then you wouldn't have to render this whole other button.

emilia-friedberg (changes requested):
This approach makes sense, but I think there's a way to do this that involves either fixing the existing code or just adjusting your approach so we don't need an additional design element.

services/QuillLMS/client/app/bundles/Evidence/helpers/containerRenderHelpers.tsx
happythenewsad: Good idea - addressed.

happythenewsad:
~Currently looking into the spec failure~ fixed; had to update the snapshot

emilia-friedberg (approved):
LGTM! Thanks for making that change!

PR #8487, opened 2021-11-10
Add error logging for Clever importing of teacher classrooms 

brendanshean:
## WHAT
During Clever Importing of teachers and classrooms:
1.  Fix an edge case involving classrooms with two owners
2. Log errors to New Relic caused by `ClassroomsTeacher.find_or_create_by!`

## WHY
1.  In the case of a classroom having two owners, `classroom.owner` is non-deterministic since it uses (`find_by_role`).  When assigning a role for clever import we should have predictable assignment.
2. There's still a larger bug that I'm working on, but it would be useful to add in some logging in the interim to get more information during failure.

## HOW
1.  Check to see if the current teacher is in the set of owners rather than equal to `classroom.owner`
2.  Add in `find_or_create_by!` method and rescue errors and log them to New Relic

### Screenshots

### Notion Card Links
https://www.notion.so/quill/Clever-Clever-Classrooms-missing-Teachers-1be9032fb94e4af48a1afa3e8156804d

PR Checklist | Your Answer
------------ | -------------

### DIFF
--- a/services/QuillLMS/app/services/clever_integration/associators/classrooms_to_teacher.rb
+++ b/services/QuillLMS/app/services/clever_integration/associators/classrooms_to_teacher.rb
@@ -1,11 +1,22 @@
 module CleverIntegration::Associators::ClassroomsToTeacher
 
   def self.run(classrooms, teacher)
-    updated_classrooms = classrooms.map do |classroom|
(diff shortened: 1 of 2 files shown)


dandrabik (approved):
:+1: LGTM. Always in favor of adding more logging.

PR #8489, opened 2021-11-10
skipping onboarding should not mess up timetracking

emilia-friedberg:
## WHAT
Fix issue where students who skipped onboarding weren't having their timetracking calculated correctly.

## WHY
We want students to have their time tracked regardless of whether or not they see the onboarding slides.  

## HOW
If students had completed enough Evidence activities to be able to skip the onboarding, they were never actually having the `explanationSlidesCompleted` boolean set to true, which is how we determine whether to add time to the onboarding step or one of the later ones. I cleaned up the code by determining the default value for `explanationSlidesCompleted` when we initially set it, which also lets us take it out of the conditional for rendering it.

### Screenshots

### Notion Card Links
https://www.notion.so/quill/Some-time-tracking-data-is-missing-61b632cf6112402a82b2f375ff038c5e

PR Checklist | Your Answer
------------ | -------------


anathomical (approved):

cissyxyu (approved):
If we use this in a couple of places, it might be helpful to save it as a constant for readability.

PR #8491, opened 2021-11-10
add rake task for skills and skill groups for the rest of the diagnostics

emilia-friedberg:
## WHAT
Add rake task for creating skills and skill groups for all of the other diagnostics.

## WHY
We initially just had them for the Starter, Intermediate, and Advanced diagnostic pre- and post-tests, but then we realized the reports wouldn't really be functional without having this data for all activities, so here we are.

## HOW
Basically just formatting the list Hannah sent me.

### Screenshots

### Notion Card Links
](https://www.notion.so/quill/Update-rake-task-with-new-skill-groups-073d69c35baa4162aeadd3225cb1f9be)

PR Checklist | Your Answer
------------ | -------------

### DIFF
--- a/services/QuillLMS/lib/tasks/create_skills_and_skill_groups_part_two.rake
+++ b/services/QuillLMS/lib/tasks/create_skills_and_skill_groups_part_two.rake
@@ -22,7 +22,7 @@ namespace :skills_and_skill_groups_part_two do
   end
 
   def create_data(skill_group_data, activity_ids)
-    skill_group_data.map do |sg|
+    skill_group_data.each do |sg|
       skill_group = SkillGroup.create({ name: sg[:name], order_number: sg[:order_number] })
       activity_ids.each do |id|
         SkillGroupActivity.create(activity_id: id, skill_group: skill_group)
@@ -196,7 +196,7 @@ namespace :skills_and_skill_groups_part_two do
             concepts: [
               { uid: 'Tlhrx6Igxn6cR_SD1U5efA' }
             ]
-          },
+          }
         ]
       },
       {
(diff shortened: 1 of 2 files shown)


services/QuillLMS/lib/tasks/create_skills_and_skill_groups_part_two.rake
```diff
@@ -0,0 +1,791 @@
...
+    create_data(pre_ap2_skill_groups, [pre_ap2.id])
+    create_data(springboard_skill_groups, [springboard.id])
+    create_data(ap_skill_groups, [ap.id])
+    create_data(ell_starter_skill_groups, [ell_starter.id])
+    create_data(ell_intermediate_skill_groups, [ell_intermediate.id])
+    create_data(ell_advanced_skill_groups, [ell_advanced.id])
+  end
+
+  def create_data(skill_group_data, activity_ids)
+    skill_group_data.map do |sg|
```
brendanshean: Since the values don't seem to be used downstream, `each` would be more appropriate here instead of `map`.

brendanshean (approved):

cissyxyu (approved):
To be safe, you might consider wrapping this whole block in a transaction block in case something fails in the middle. That way we won't have any half-created records if there is a failure in this task.

https://api.rubyonrails.org/v6.1.4/classes/ActiveRecord/Transactions/ClassMethods.html

emilia-friedberg:
> To be safe, you might consider wrapping this whole block in a transaction block in case something fails in the middle. That way we won't have any half-created records if there is a failure in this task.
> 
> https://api.rubyonrails.org/v6.1.4/classes/ActiveRecord/Transactions/ClassMethods.html

Brilliant, thanks Cissy!

PR #8499, opened 2021-11-12
When rake commands are run on production, post the command to quill-developer slack channel

brendanshean:
## WHAT
When a user runs a rake task on our production LMS, post a message to quill-developer channel.

## WHY
It's useful to see commands that alter the production database.

## HOW
Add an after_hook following this [post](https://t2013anurag.medium.com/rails-adding-global-hooks-for-rake-tasks-7faf0844364).  While it would be useful to also post the user who executed the rake task, I wasn't able to find that info easily.

### Screenshots
Screen Shot 2021-11-12 at 10 20 28 AM

### Notion Card Links

PR Checklist | Your Answer
------------ | -------------

### DIFF
--- a/services/QuillLMS/Rakefile
+++ b/services/QuillLMS/Rakefile
@@ -3,29 +3,18 @@
 
 require File.expand_path('../config/application', __FILE__)
 require 'dotenv/tasks'
+require 'tasks/modules/slack_tasks'
 
 EmpiricalGrammar::Application.load_tasks
 
 task default: [:spec]
 
 task after_hook: :dotenv do
+  include SlackTasks
+
   post_slack_rake_task
 end
 
-Rake.application.tasks.reject do |task|
-  next if [Rake::Task['environment'], Rake::Task['dotenv'], Rake::Task['after_hook']].include?(task)
-
-  task.enhance([:after_hook])
-end
-
-def post_slack_rake_task
-  return unless ENV.fetch('RAILS_ENV', '') == 'production'
-
-  webhook_url = ENV.fetch('SLACK_API_WEBHOOK', '')
-
-  return unless webhook_url.present?
-
-  message = ":rake: rake #{ARGV.join(' ')}"
-
-  sh %(curl -X POST -H 'Content-type: application/json' --data "{'text':'#{message}'}" #{webhook_url})
+Rake.application.tasks.each do |task|
+  task.enhance([:after_hook]) unless %w[after_hook dotenv environment].include?(task.name)
 end
--- a/services/QuillLMS/Rakefile
+++ b/services/QuillLMS/Rakefile
@@ -2,7 +2,30 @@
 # for example lib/tasks/capistrano.rake, and they will automatically be available to Rake.
 
 require File.expand_path('../config/application', __FILE__)
+require 'dotenv/tasks'
 
 EmpiricalGrammar::Application.load_tasks
 
 task default: [:spec]
+
+task after_hook: :dotenv do
+  post_slack_rake_task
+end
+
+Rake.application.tasks.reject do |task|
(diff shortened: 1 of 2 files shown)


services/QuillLMS/Rakefile
```diff
@@ -2,7 +2,30 @@
...
+end
+
+def post_slack_rake_task
+  return unless ENV.fetch('RAILS_ENV', '') == 'production'
+
+  webhook_url = ENV.fetch('SLACK_API_WEBHOOK', '')
+
+  return unless webhook_url.present?
+
+  message = ":rake: rake #{ARGV.join(' ')}"
```
dandrabik: I like the emoji 🤣

services/QuillLMS/Rakefile
```diff
@@ -2,7 +2,30 @@
...
 
 EmpiricalGrammar::Application.load_tasks
 
 task default: [:spec]
+
+task after_hook: :dotenv do
+  post_slack_rake_task
+end
+
+Rake.application.tasks.reject do |task|
```
dandrabik: There's a few rake tasks I think we'd want to ignore:

1. Asset compilation during deploy, e.g. `rake assets:precompile`, `rake assets:clean`, [our enhance method](https://github.com/empirical-org/Empirical-Core/blob/develop/services/QuillLMS/lib/tasks/assets.rake#L7). Note, you used these in the PR example, so you might be thinking differently. Happy to discuss.
2. [Our generic cron rake tasks](https://github.com/empirical-org/Empirical-Core/blob/develop/services/QuillLMS/lib/tasks/cron.rake#L1)

Scanning `lib/tasks`, I think that's it (but if others pop up, we can add them later).

services/QuillLMS/Rakefile
```diff
@@ -2,7 +2,30 @@
...
+  next if [Rake::Task['environment'], Rake::Task['dotenv'], Rake::Task['after_hook']].include?(task)
+
+  task.enhance([:after_hook])
+end
+
+def post_slack_rake_task
```
dandrabik: It would be good to move this method out of the global namespace (it's a very specific method name, so it wouldn't collide, but just to keep the global namespace clean), e.g.
```ruby
module SlackTasks
  def post_slack_rake_task
  end
end
task after_hook: :dotenv do
  include SlackTasks
  post_slack_rake_task
end
```

Also, this might be premature, but you might consider moving this method out of tasks and making this reusable, e.g. 

```ruby
# create interface
Slack.post_to_developer(message)

# use (you can pass a ruby block to enhance)
task.enhance do
  Slack.post_to_developer(":rake: rake #{ARGV.join(' ')}")
end

dandrabik (approved):
I really like this idea and setup, thanks for putting this together! :100:

I've made a few comments about tasks we should likely ignore, and some small refactor suggestions. Marking as 'Approved' since they are all small, but you should read through them.

services/QuillLMS/Rakefile
brendanshean: While testing this on staging, I used `db:migrate:status`  and `db:rollback STEP=0`.  The `rake assets:precompile`, `rake assets:clean` fired errantly, so I would agree that we need these on an ignore list.
Screen Shot 2021-11-15 at 1 28 14 PM

services/QuillLMS/Rakefile
brendanshean: Great idea on moving out of global namespace. I will extract that out.  

I thought about extracting the method out but the [other](https://github.com/empirical-org/Empirical-Core/blob/develop/scripts/post_slack_deploy.sh) [scripts](https://github.com/empirical-org/Empirical-Core/blob/develop/scripts/post_slack_deploy_description.sh) that post to slack function a little differently, calling rake locally `rake db:migrate` vs this PR where they are called remotely `heroku run rake db:migrate -r app-name`.

PR #8505, opened 2021-11-12
Fix bug involving ClassroomsTeacher creation during Clever importing

brendanshean:
## WHAT
Clean up issue where clever classrooms are not associated with a particular teacher.

## WHY
Having clever classrooms without teachers causes many issues elsewhere.

## HOW
1.  Previously, ClassroomsTeacher records were being created downstream from the classroom creation.  Instead of having separate transactions, we can construct classroom and classrooms_teacher record in a single transaction using [nested attributes](https://github.com/empirical-org/Empirical-Core/pull/8505/files#diff-bd92c44b1dd8abf9bb8e3c90940dba94aa1f291e4c41593005209e30489bd71fR20)

2.  In the case of classrooms that already exist, move associations code into the [updater](https://github.com/empirical-org/Empirical-Core/pull/8505/files#diff-2018c8e8d3e29bd293c0b0938a1a03e570656dba81e3604f2b4c94227583a2d1R22) service object

### Notion Card Links
https://www.notion.so/quill/Clever-Clever-Classrooms-missing-Teachers-1be9032fb94e4af48a1afa3e8156804d

PR Checklist | Your Answer
------------ | -------------


dandrabik (approved):
:+1: Thanks for cleaning this up. I think this is a much better structure!

PR #8531, opened 2021-11-18
Temporarily log clever importing for school admin

brendanshean:
## WHAT
Save one clever import auth_hash for each user_type to the database.

## WHY
We currently have no way of testing library integration imports. These records will give the correct payload to model tests after.

## HOW
Save auth_hash to changelog record.

### Screenshots

### Notion Card Links

PR Checklist | Your Answer
------------ | -------------
Have you added and/or updated tests? |  Manual testing on staging.

### DIFF
--- a/services/QuillLMS/app/services/clever_integration/sign_up/school_admin.rb
+++ b/services/QuillLMS/app/services/clever_integration/sign_up/school_admin.rb
@@ -12,10 +12,12 @@ module CleverIntegration::SignUp::SchoolAdmin
   end
 
   def self.library_integration(auth_hash)
+    log_import(:district_integration, auth_hash) # TODO: remove this temporary call
     CleverIntegration::Importers::Library.run(auth_hash)
   end
 
   def self.district_integration(auth_hash, district)
+    log_import(:library_integration, auth_hash) # TODO: remove this temporary call
     user = create_user(auth_hash)
     if user.present?
       associate_user_to_district(user, district)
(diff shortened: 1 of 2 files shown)


services/QuillLMS/app/services/clever_integration/sign_up/school_admin.rb
```diff
@@ -26,6 +28,16 @@ def self.district_integration(auth_hash, district)
...
   end
 
+  #TODO: remove this method
+  def self.log_import(action, auth_hash)
+    user = User.find(ENV['CLEVER_IMPORT_LOG_USER_ID'])
+    changed_attr = auth_hash.dig(:info, :user_type)
+
+    return if ChangeLog.exists?(changed_record: user, changed_attribute: changed_attr, action: action, user: user)
+
+    ChangeLog.create!(explanation: auth_hash.to_json, changed_record: user, changed_attribute: changed_attr, action: action, user: user)
```
dandrabik: I might have missed this on the other PR, but do you want to raise if the Changelog isn't valid? 

it might be better to fail silently (or log an error) rather than crash, since it's just logging. (You are closer to this code, so you might have a reason to raise).

dandrabik (approved):
One comment about whether of not to raise on a `Changelog` error (that I'll defer to your decision on), otherwise looks good to go!

services/QuillLMS/app/services/clever_integration/sign_up/school_admin.rb
brendanshean: That's a good point.   I will remove from here and the other PR.

PR #8540, opened 2021-11-22
Mrfreeze

happythenewsad:
## WHAT
- commit 1 adds magic comments to all LMS Ruby files, auto-freezing string literals.
- commit 2 removes now-redundant `.freeze` invocations and unfreezes strings that actually need to be modified (via `dup`)

## WHY
- reduces string allocation, for faster performance 
- This is [Mike Perham](https://www.mikeperham.com/2018/02/28/ruby-optimization-with-one-magic-comment/)'s preferred method, as opposed to always remembering to add `.freeze` to every constant string

## HOW
- Add the magic comment to all .rb files 

## DISCUSSION
This PR is one implementation to solve the performance problem of unfrozen strings. I'm leaning toward this solution because it is set & forget. We don't have to be adding `.freeze` everywhere. One downside is that when strings do need to be modified, you need to use `dup`. This doesn't happen very often - in fact, just 4x in LMS (see the diff).

Other options we can implement are:
- do nothing
- use `.freeze` everywhere
- use `.freeze` in a more scoped way - only when actually declared as CONSTANT, as opposed to any string literal

There's further discussion in this official Ruby bug thread:
https://bugs.ruby-lang.org/issues/16047

### Notion Card Links
[(Please provide links to any relevant Notion card(s) relevant to this PR.)](https://www.notion.so/quill/Product-Board-30f97e4eb01246dbb8264253fb385073?p=76110756aad74befa4314812de1eee2f)

PR Checklist | Your Answer
------------ | -------------
Have you added and/or updated tests? |  No - no logic change


services/QuillLMS/app/models/subscription.rb
```diff
@@ -344,7 +346,7 @@ def self.set_cb_lifetime_expiration_and_start_date
   end
 
   def self.create_with_school_or_user_join school_or_user_id, type, attributes
-    type.capitalize!
+    type = type.capitalize
```
brendanshean, line 349: What's the purpose of this change?

brendanshean (approved):
Thank you for doing this. 

I think it would be worth adding a [cop](https://www.rubydoc.info/gems/rubocop/RuboCop/Cop/Style/FrozenStringLiteralComment) to ensure that future ruby code includes this pragma comment.

services/QuillLMS/app/models/subscription.rb
happythenewsad: @brendanshean I haven't requested reviews yet as I'm going to push up a couple more commits. But this change is because the `type` argument is not a frozen string, so it cannot be modified.

brendanshean (approved):

anathomical (approved):
I like this approach a lot more than trying to remember to put `freeze` everywhere.  The mutable scenario is such a small segment of our use-cases that it makes sense for THAT to be the place we have additional syntax rather than the more common use-case.

Good call.

happythenewsad:
before: 
bundle exec rspec spec/controllers  124.51s user 14.10s system 79% cpu 2:54.66 total 

after:
bundle exec rspec spec/controllers  115.53s user 13.23s system 79% cpu 2:42.26 total

PR #8544, opened 2021-11-23
Add length validation on User.username

cissyxyu:
## WHAT
We noticed in Sentry that a student intentionally put a huge chunk of text in the username field and caused an error. So we want to prevent this in the future for all the student-editable fields.

## WHY
We don't want the user creation process to break when a student tries to mess with a text input field.

## HOW
Put a validation check on all the fields that the student can manually enter.

### Screenshots
Screen Shot 2021-11-23 at 2 52 01 PM

### Notion Card Links
https://www.notion.so/quill/Sentry-Error-AccountsController-create-ActiveRecord-ValueTooLong-3f8841ca8447423a8f9e707540324276

PR Checklist | Your Answer
------------ | -------------

### DIFF
--- a/services/QuillLMS/app/models/user.rb
+++ b/services/QuillLMS/app/models/user.rb
@@ -61,6 +61,8 @@ class User < ApplicationRecord
   attr_accessor :validate_username,
                 :require_password_confirmation_when_password_present
 
+  CHAR_FIELD_MAX_LENGTH = 255
+
   before_validation :generate_student_username_if_absent
   before_validation :prep_authentication_terms
   before_save :capitalize_name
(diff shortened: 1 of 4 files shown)


services/QuillLMS/app/models/user.rb
```diff
@@ -116,14 +116,17 @@ class User < ApplicationRecord
   delegate :name, :mail_city, :mail_state, to: :school, allow_nil: true, prefix: :school
 
   validates :name,                  presence: true,
-                                    format:       {without: /\t/, message: 'cannot contain tabs'}
+                                    format:       {without: /\t/, message: 'cannot contain tabs'},
+                                    length:       { maximum: 255}
```
happythenewsad: Let's move the value 255 into a constant - we want to avoid magic numbers.

happythenewsad (changes requested):

services/QuillLMS/spec/models/user_spec.rb
```diff
@@ -96,6 +96,10 @@
   #it { should validate_uniqueness_of(:username).on(:create) }
 
   it { should validate_presence_of(:username).on(:create) }
+  it { should validate_length_of(:username).is_at_most(255) }
```
happythenewsad: @cissyxyu here too :-) 
Using the constant means we can change the value later without breaking specs.

cissyxyu: Whoops, makes sense! I added it in the spec too.

happythenewsad (approved):

PR #8545, opened 2021-11-23
Add a nil check in UserAdminSerializer

cissyxyu:
## WHAT
Add a nil check for when the `admins_teachers` method call returns `nil`. So the worker doesn't error out when it reaches this edge case.

## WHY
To prevent nil errors in this worker.

## HOW
Add a nil check to the method call to return the `else` block in cases of `nil`.

### Screenshots

### Notion Card Links
https://www.notion.so/quill/Sentry-Error-FindAdminUsersWorker-NoMethodError-23898c188e6f41e4b3f7aa64f7c0960c

PR Checklist | Your Answer
------------ | -------------

### DIFF
--- a/services/QuillLMS/app/models/user.rb
+++ b/services/QuillLMS/app/models/user.rb
@@ -116,14 +116,17 @@ class User < ApplicationRecord
   delegate :name, :mail_city, :mail_state, to: :school, allow_nil: true, prefix: :school
 
   validates :name,                  presence: true,
-                                    format:       {without: /\t/, message: 'cannot contain tabs'}
+                                    format:       {without: /\t/, message: 'cannot contain tabs'},
+                                    length:       { maximum: 255}
 
   validates_with ::FullnameValidator
 
-  validates :password,              presence:     { if: :requires_password? }
+  validates :password,              presence:     { if: :requires_password? },
+                                    length:       { maximum: 255}
 
   validates :email,                 presence:     { if: :email_required? },
-                                    uniqueness:   { message: :taken, if: :email_required_or_present?}
+                                    uniqueness:   { message: :taken, if: :email_required_or_present?},
+                                    length:       { maximum: 255}
 
   validate :username_cannot_be_an_email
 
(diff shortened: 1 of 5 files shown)


services/QuillLMS/app/models/user.rb
```diff
@@ -116,14 +116,17 @@ class User < ApplicationRecord
...
   validates :name,                  presence: true,
-                                    format:       {without: /\t/, message: 'cannot contain tabs'}
+                                    format:       {without: /\t/, message: 'cannot contain tabs'},
+                                    length:       { maximum: 255}
 
   validates_with ::FullnameValidator
 
-  validates :password,              presence:     { if: :requires_password? }
+  validates :password,              presence:     { if: :requires_password? },
+                                    length:       { maximum: 255}
```
happythenewsad: Same comment regarding magic numbers.

services/QuillLMS/app/serializers/user_admin_serializer.rb
```diff
@@ -3,7 +3,7 @@ class UserAdminSerializer < ActiveModel::Serializer
 
   def teachers
     teacher_ids = User.find(object.id).admins_teachers
-    if teacher_ids.any?
+    if teacher_ids&.any?
```
brendanshean: What about moving the check for empty_ids inside the `TeachersData.run` method itself:

```
  def self.run(teacher_ids)
   return [] if teacher_ids.blank?

    teacher_ids_str = teacher_ids.join(', ')
    ...
```
That way, the `TeachersData.run` is now protected from other objects calling it with nil/empty params.

happythenewsad (approved):
LGTM after the magic number fix.

services/QuillLMS/app/serializers/user_admin_serializer.rb
cissyxyu: Makes sense. I moved the whole check into the query method.

brendanshean: Perfect!

brendanshean (approved):
Good to go!

PR #8556, opened 2021-11-30
nil check for pre test activity session

emilia-friedberg:
## WHAT
Handle case where students have completed a post-test activity but not the pre-test for the `diagnostic_units` method.

## WHY
We shouldn't be in this state - it isn't possible for students to start a post-test from their dashboard without having completed the pre-test first, but I think I overlooked an edge case where the new Google Classroom activity links will still work if a teacher assigns both at the same time. Going to fix that in another PR, but in the interim this should allow the teacher's page to load. There may be other related errors which we will need to fix, but I will have better insight into what those are once this page is working.

## HOW
Nil check the activity session before calling `correct_skill_ids` on it.

### Screenshots

### Notion Card Links
](https://www.notion.so/quill/New-diagnostic-reports-not-loading-for-1-user-even-after-clearing-cache-742072097680429984e90d5eaaeba3ae)

PR Checklist | Your Answer
------------ | -------------

### DIFF
--- a/services/QuillLMS/app/controllers/concerns/growth_results_summary.rb
+++ b/services/QuillLMS/app/controllers/concerns/growth_results_summary.rb
@@ -31,7 +31,7 @@ module GrowthResultsSummary
     @post_test_assigned_students.map do |assigned_student|
       post_test_activity_session = @post_test_activity_sessions[assigned_student.id]
       pre_test_activity_session = @pre_test_activity_sessions[assigned_student.id]
-      if post_test_activity_session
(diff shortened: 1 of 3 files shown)


happythenewsad (approved):
LGTM

Please add unit tests to capture the bug your are fixing in your follow-on PR (or add here). IMO using `send` to test `student_results`, or making `student_results` a non-private method are both acceptable techniques.

emilia-friedberg:
@happythenewsad added a test!

PR #8558, opened 2021-11-30
change default parent activity flagset

happythenewsad:
## WHAT
When an Evidence activity is created, set the parent activity's flagset to `alpha` instead of beta 

## WHY
So that beta users are not immediately exposed to in-development content 

## HOW
See diff. 

### Screenshots

### Notion Card Links
[(Please provide links to any relevant Notion card(s) relevant to this PR.)](https://www.notion.so/quill/Flag-new-Quill-Evidence-activities-with-the-alpha-flag-instead-of-beta-4b27813ce676482c8b494a9ebabf390b)

PR Checklist | Your Answer
------------ | -------------

### DIFF
--- a/services/QuillLMS/engines/evidence/app/models/evidence/activity.rb
+++ b/services/QuillLMS/engines/evidence/app/models/evidence/activity.rb
@@ -12,6 +12,10 @@ module Evidence
     MAX_TITLE_LENGTH = 100
     MAX_SCORED_LEVEL_LENGTH = 100
 
+    # See activity.rb in the enclosing app for the ur-constant,
+    # which is not accessible from here
+    LMS_ACTIVITY_DEFAULT_FLAG = 'alpha'
+
     before_destroy :expire_turking_rounds
     before_validation :set_parent_activity, on: :create
     after_save :update_parent_activity_name, if: :saved_change_to_title?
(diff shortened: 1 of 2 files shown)


services/QuillLMS/engines/evidence/app/models/evidence/activity.rb
```diff
@@ -42,7 +46,9 @@ def set_parent_activity
         self.parent_activity = Evidence.parent_activity_class.find_or_create_by(
           name: title,
           activity_classification_id: Evidence.parent_activity_classification_class.evidence&.id
-        )
+        ) do |activity|
```
brendanshean: For greater readability, maybe rename the block parameter `parent_activity` or `p_activity`  to distinguish from the evidence activity:

```) { |parent_activity| parent_activity.flags =  [LMS_ACTIVITY_DEFAULT_FLAG] }```

brendanshean (approved):
Made a minor comment for readability, but looks good otherwise.

anathomical (approved):
This solution seems just fine, but a bit limited.  You'd have to check with Curriculum, but I wonder if the better fix here isn't just to give the LMS `Activity` model default "alpha" flags for all activity types, not just Evidence.  I doubt it super matters, but I suspect that that's desired behavior that's just being handled already in other activity types.

happythenewsad:
> This solution seems just fine, but a bit limited. You'd have to check with Curriculum, but I wonder if the better fix here isn't just to give the LMS `Activity` model default "alpha" flags for all activity types, not just Evidence. I doubt it super matters, but I suspect that that's desired behavior that's just being handled already in other activity types.

Going with the current implementation, as non-Evidence activities already default to alpha. See full convo here: https://www.notion.so/quill/b417e5b89fe04af89cba62d4198f7031?v=77f294a5b0df466e98633244de7ebfef&p=4b27813ce676482c8b494a9ebabf390b

PR #8562, opened 2021-11-30
assignment flow/ share activity links with students and to Google Classroom (part 1)

eadams17:
## WHAT
add the functionality for teachers to share direct, clickable links to their students or post to Google Classroom. This PR adds the functionality while in the assignment flow (note: there is a tweak some logic in the second PR which handles the case where only one classroom is selected; the link text is rendered rather than the dropdown)

## WHY
we want to give teachers the ability to send direct activity links to their students and post assignments to Google Classroom.

## HOW
added a new stage to the assignment flow `ShareToStudents` and a `ShareActivityPackModal` which handles all the logic of fetching `classrooms` and `classroom_units` in order to construct the appropriate link

### Screenshots
Screen Shot 2021-11-30 at 2 42 01 PM
Screen Shot 2021-11-30 at 2 42 15 PM
Screen Shot 2021-11-30 at 2 49 45 PM
Screen Shot 2021-11-30 at 2 43 56 PM
Screen Shot 2021-11-30 at 2 44 15 PM

### Notion Card Links
https://www.notion.so/quill/Google-Classroom-Frontend-Add-share-this-activity-to-the-assignment-workflow-89fd37179e904ea482480ae279d99240

PR Checklist | Your Answer
------------ | -------------


services/QuillLMS/client/app/bundles/Shared/components/shared/expandableCard.tsx
```diff
@@ -0,0 +1,66 @@
...
+        })}
+      </section>
+    );
+  }
+
+  const img = imgSrc ? <img alt={imgAlt} className="expandable-card-icon" src={imgSrc} /> : null;
+  const cardClass = `expandable-quill-card ${isExpanded ? 'open' : ''}`;
+
+  return (
+    <div className={cardClass} onKeyDown={handleKeyDownOnCard} role="button" tabIndex={0}>
```
emilia-friedberg: do we not want an `onClick` handler here? just a keydown?

services/QuillLMS/client/app/bundles/Teacher/components/assignment_flow/create_unit/create_unit.jsx
```diff
@@ -374,10 +380,31 @@ export default class CreateUnit extends React.Component {
...
+      );
+    }
+  }
+
+  stage4specificComponents = () => {
     const { referralCode, location, history, } = this.props
-    const { classrooms, selectedActivities, name, } = this.state
-    if ((this.state.assignSuccess)) {
+    const { classrooms, selectedActivities, name, assignSuccess } = this.state
+    if ((assignSuccess)) {
```
emilia-friedberg: This was leftover from the previous version but there are double parentheses here that might be worth removing - up to you.

services/QuillLMS/client/app/bundles/Teacher/components/assignment_flow/create_unit/share_activity_pack/activityPackInformation.tsx
```diff
@@ -0,0 +1,38 @@
...
+
+  function getStudentCount() {
+    let count = 0;
+    classroomData.map(classroom => {
+      const { students } = classroom;
+      if(students && students.length) {
+        count += students.length;
+      }
+    });
+    return count;
```
emilia-friedberg: I think a slightly cleaner way to do this would be like this:

```
  return classroomData.reduce((previousValue, classroom) => {
    const { students, } = classroom
    return students && students.length ? previousValue += students.length : previousValue
  }, 0)

```

services/QuillLMS/client/app/bundles/Teacher/components/assignment_flow/create_unit/share_activity_pack/shareActivityPackModal.tsx
```diff
@@ -0,0 +1,191 @@
...
+
+export const ShareActivityPackModal = ({ activityPackData, closeModal, singleActivity, unitId }) => {
+  const [showSnackbar, setShowSnackbar] = React.useState<boolean>(false);
+  const [selectedClass, setSelectedClass] = React.useState<DropdownObject>({ label: '', value: '' });
+  const [classrooms, setClassrooms] = React.useState<Classroom[]>([]);
+  const [classroomUnits, setClassroomUnits] = React.useState<ClassroomUnit[]>([]);
+  const [link, setLink] = React.useState<string>(classrooms && classrooms.length && getDefaultLink());
+
+  React.useEffect(() => {
+    if(!classrooms.length || (!classroomUnits.length && unitId)) {
```
emilia-friedberg: Shouldn't a `useEffect` that just has an empty array as a second parameter only run once? If that's true, I don't think you should need this line, because you're always instantiating these values with empty arrays.

services/QuillLMS/client/app/bundles/Teacher/components/assignment_flow/create_unit/share_activity_pack/shareActivityPackModal.tsx
```diff
@@ -0,0 +1,191 @@
...
+        setClassrooms(body.classrooms && body.classrooms.classrooms_and_their_students)
+        setClassroomUnits(body.classroom_units)
+      })
+    }
+  }, [])
+
+  function getDefaultLink() {
+    if(classrooms.length !== 1) {
+      return ''
+    }
```
emilia-friedberg: This can be one line.

services/QuillLMS/client/app/bundles/Teacher/components/assignment_flow/create_unit/share_activity_pack/shareActivityPackModal.tsx
```diff
@@ -0,0 +1,191 @@
...
+    const classroomObject = classrooms[0];
+    const classroomUnitObject = classroomUnits[0];
+    const { classroom } = classroomObject;
+    if(!singleActivity) {
```
emilia-friedberg, line 95: Maybe it makes more sense to start with the positive conditional?

services/QuillLMS/client/app/bundles/Teacher/components/assignment_flow/create_unit/share_activity_pack/shareActivityPackModal.tsx
```diff
@@ -0,0 +1,191 @@
...
+
+  function handleCopyLink() {
+    setShowSnackbar(true);
+    setTimeout(() => setShowSnackbar(false), defaultSnackbarTimeout);
+  }
+
+  function handleClassChange(classOption) {
+    const { value } = classOption;
+    setSelectedClass(classOption);
+    if(!singleActivity) {
```
emilia-friedberg: Again I think it would be more readable to start with the case where there `is` a `singleActivity`.

services/QuillLMS/client/app/bundles/Teacher/components/assignment_flow/create_unit/share_activity_pack/shareToStudents.tsx
```diff
@@ -0,0 +1,171 @@
...
+
+  React.useEffect(() => {
+    setActivityPack(activityPackData);
+  }, [activityPackData]);
+
+  React.useEffect(() => {
+    if (leaving) {
+      handleUnmount()
+    }
+  }, [leaving]);
```
emilia-friedberg: Why do we need this when the effect below should already handle the cleanup?

emilia-friedberg (approved):
This looks really good! I left some small suggestions you can take or leave, and some questions that are just for my own edification. Nice work! Teachers will be so happy they can do this.

services/QuillLMS/client/app/bundles/Shared/components/shared/expandableCard.tsx
eadams17: this is actually something Peter S just brought up, I'll be adding that

services/QuillLMS/client/app/bundles/Teacher/components/assignment_flow/create_unit/share_activity_pack/activityPackInformation.tsx
eadams17: good call! thanks for the suggestion

services/QuillLMS/client/app/bundles/Teacher/components/assignment_flow/create_unit/share_activity_pack/shareActivityPackModal.tsx
eadams17: ahh good catch, I didn't realize I had added that array argument

services/QuillLMS/client/app/bundles/Teacher/components/assignment_flow/create_unit/share_activity_pack/shareToStudents.tsx
eadams17: yea sorry there's a slight difference in the second PR. There's a redirect to the classrooms section of the dashboard that we don't want to happen when the user leaves this stage to move to the `Stage4` component. I'll add a comment

services/QuillLMS/app/controllers/activities_controller.rb
```diff
@@ -62,15 +62,18 @@ def customize_lesson
   end
 
   def activity_session
+    return redirect_to profile_path if !current_user.student?
```
anathomical: Pretty minor Ruby thing, but I think our current pattern is to use `unless` instead of `if !`.  So this could be: `return redirect_to profile_path unless current_user.student?`

services/QuillLMS/app/controllers/students_controller.rb
```diff
@@ -102,7 +101,13 @@ def join_classroom
   private def redirect_to_profile
     @current_user = current_user
     classroom_id = params["classroom"]
-    if classroom_id && (Classroom.find_by(id: classroom_id).nil? || StudentsClassrooms.find_by(student_id: @current_user.id, classroom_id: classroom_id).nil?)
+    unit_id = params["unit_id"]
+
+    return redirect_to profile_path if current_user && !current_user.student?
```
anathomical: You could combine Ruby's built-in nil-checking with `unless` here: `return redirect_to profile_path unless current_user&.student?`

(`current_user&.student?` will return `nil` if `current_user` is `nil`, and it won't try to check the value of `student?`)

services/QuillLMS/app/controllers/students_controller.rb
```diff
@@ -113,12 +118,12 @@ def join_classroom
     classroom_id = params["classroom"]
     unit_id = params["unit_id"]
 
-    return unless classroom_id && unit_id
-
     classroom_unit = ClassroomUnit.find_by(classroom_id: classroom_id, unit_id: unit_id)
 
     unless classroom_unit && classroom_unit.assigned_student_ids.include?(current_user.id)
-      flash[:error] = 'Sorry, you do not have access to this activity pack because it has not been assigned to you. Please contact your teacher.'
+      flash[:error] = t('activity_link.errors.activity_pack_not_assigned')
```
anathomical, line 124: I do like the move to translation strings in these situations.  Just figured I'd cll that out since we should probably do more of it.

services/QuillLMS/app/controllers/teachers/classroom_manager_controller.rb
```diff
@@ -46,6 +46,14 @@ def retrieve_classrooms_i_teach_for_custom_assigning_activities
     render json: classroom_with_students_json(current_user.classrooms_i_teach)
   end
 
+  def classrooms_and_classroom_units_for_activity_share
+    unit_id = params["unit_id"]
+    render json: {
+      classrooms: classroom_with_students_json(current_user.classrooms_i_teach),
+      classroom_units: unit_id ? ClassroomUnit.where(unit_id: unit_id) : []
```
anathomical: So the `ClassroomUnit` database definition doesn't allow `unit_id` to be `nil`.  This means that no records can be found with that value.  If you call `ClassroomUnit.where(unit_id: nil)` you'll actually get an empty array back.  So you don't need the ternary here.  You can just make this line:`classroom_units: ClassroomUnit.where(unit_id: unit_id)`

anathomical (approved):
A couple of notes on Ruby syntax stuff, but nothing to stop a merge.

services/QuillLMS/app/controllers/students_controller.rb
eadams17: good call-- I always forget that ruby has that `&.` feature

services/QuillLMS/app/controllers/teachers/classroom_manager_controller.rb
eadams17: nice! I'll update

PR #8569, opened 2021-12-03
don't try to save timespent values that are larger than the maximum

emilia-friedberg:
## WHAT
If a timespent value is larger than a 4-byte integer size (what's allowed in our database), save the maximum 4-byte integer size instead. Also add tracking so I can see more detail about why we might be getting really large values for timespent in the first place.

## WHY
We have a [sentry error](https://sentry.io/organizations/quillorg-5s/issues/2823293016/?project=11238&query=is%3Aunresolved) that came about as a result of a student with an insanely large timespent value. All instances of this error are from the same student. This value is the equivalent of 68 years. I'm not sure how this value made its way in there, since even if the timetracking never stopped running (which it does, when the user is idle for more than 30 seconds or they switch tabs, according to my testing) this is way way way longer than it should ever possibly be. I'm hoping that adding this Sentry tracking for all sessions longer than an hour will give me some more information about how this might be occurring - possibly browser-specific.

## HOW
Just check that the timespent value is not larger than the maximum size before saving it and, if it is, default to the largest possible size, just so the student can actually save their session.

### Screenshots

### Notion Card Links
](https://www.notion.so/quill/Sentry-Error-Api-V1-ActivitySessionsController-update-Timetracking-Out-of-Range-f7ed4c129700466b8d7e6319d5d64f6b)

PR Checklist | Your Answer
------------ | -------------

### DIFF
--- a/services/QuillLMS/app/controllers/api/v1/activity_sessions_controller.rb
+++ b/services/QuillLMS/app/controllers/api/v1/activity_sessions_controller.rb
@@ -105,7 +105,7 @@ class Api::V1::ActivitySessionsController < Api::ApiController
     timespent = @activity_session&.timespent || ActivitySession.calculate_timespent(time_tracking)
     max_4_bit_integer_size = 2147483647
 
-    if timespent > 3600
+    if timespent && timespent > 3600
       begin
         raise "#{timespent} seconds for user #{@activity_session.user_id} and activity session #{@activity_session.id}"
       rescue => e
--- a/services/QuillLMS/app/controllers/api/v1/activity_sessions_controller.rb
+++ b/services/QuillLMS/app/controllers/api/v1/activity_sessions_controller.rb
@@ -105,6 +105,14 @@ class Api::V1::ActivitySessionsController < Api::ApiController
     timespent = @activity_session&.timespent || ActivitySession.calculate_timespent(time_tracking)
     max_4_bit_integer_size = 2147483647
 
+    if timespent > 3600
+      begin
+        raise "#{timespent} seconds for user #{@activity_session.user_id} and activity session #{@activity_session.id}"
+      rescue => e
+        Sentry.capture_exception(e)
+      end
+    end
+
     params
       .permit(activity_session_permitted_params)
       .merge(data: data)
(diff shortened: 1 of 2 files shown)


services/QuillLMS/app/controllers/api/v1/activity_sessions_controller.rb
```diff
@@ -103,12 +103,21 @@ def destroy
     data = params.delete(:data)&.permit!
     time_tracking = data && data['time_tracking']
     timespent = @activity_session&.timespent || ActivitySession.calculate_timespent(time_tracking)
+    max_4_bit_integer_size = 2147483647
```
anathomical: Probably worth making this a constant at the top of the file since it never changes.

services/QuillLMS/app/controllers/api/v1/activity_sessions_controller.rb
```diff
@@ -103,12 +103,21 @@ def destroy
...
+        Raven.capture_exception(e)
+      end
+    end
 
     params
       .permit(activity_session_permitted_params)
       .merge(data: data)
       .reject { |_, v| v.nil? }
-      .merge(timespent: timespent)
+      .merge(timespent: max_4_bit_integer_size > (timespent || 0) ? timespent : max_4_bit_integer_size)
```
anathomical: I think this is slightly more readable if we use `min`: `[(timespent || 0), max_4_bit_integer_size].min`

anathomical (approved):
A couple of syntax notes, but this approach seems reasonable, and you know I always love tests.

services/QuillLMS/app/controllers/api/v1/activity_sessions_controller.rb
emilia-friedberg: I modified your version slightly to account for the case where timespent is actually nil, in which case I think we'd better save nil than 0. I don't think this should ever happen in reality because we have timetracking in all of our tools, but that's a closer preservation of the current behavior.

happythenewsad (changes requested):
That integer ceiling is 4 bytes, not 4 bits. Not trying to be nit-picky but it might confuse someone later. Otherwise, LGTM

emilia-friedberg:
@happythenewsad not nitpicky, thank you for the correction! I blame post-booster brain fog for that one.

happythenewsad (approved):

PR #8574, opened 2021-12-03
Demo account for teachers

cissyxyu:
## WHAT
Publish demo account to all teachers. Add links and buttons to direct teachers to explore the demo account, as well as a banner that easily logs them out and back into their own account.

## WHY
We want to see if this new feature will help improve teacher retention with newly signed up teachers.

## HOW
Some styling changes to add the "Explore demo" features to the teacher dashboard and welcome modal.

For the back end, the functionality is similar to the "View as Student" feature we already built. Instead of logging a teacher out of their account and into the Demo account when they go to `quill.org/demo,` we're now saving the new demo account ID as `session[:demo_id]` and saving their real account ID in `session[:user_id]` so they can still jump back out of the demo into their own account at any time.

### Screenshots
Screen Shot 2021-12-03 at 4 52 57 PM
Screen Shot 2021-12-03 at 4 52 53 PM
Screen Shot 2021-12-03 at 4 51 59 PM

### Notion Card Links
https://www.notion.so/quill/Display-the-fifth-attempt-feedback-on-Evidence-Activity-Analysis-e58f0720da714cdebe3c91df4328bdb1

PR Checklist | Your Answer
------------ | -------------


services/QuillLMS/app/lib/quill_authentication.rb
```diff
@@ -68,10 +70,22 @@ def sign_in(user)
...
+      @current_user = User.find(session[:user_id])
+    end
+  end
+
   def preview_student_id=(student_id)
     session[:preview_student_id] = student_id
     if student_id
       @current_user = User.find(session[:preview_student_id])
+    elsif session[:demo_id].present?
+      @current_user = User.find(session[:demo_id])
```
emilia-friedberg, line 90: Why is this here? It seems like it would case weird behavior if you tried to preview a student account from inside the demo account.

services/QuillLMS/client/app/bundles/Teacher/components/dashboard/welcome_modal.tsx
```diff
@@ -16,13 +22,22 @@ const WelcomeModal = ({ close, size }) => {
...
+          <img alt="Teacher at projector in classroom" src="https://assets.quill.org/images/pages/dashboard/illustrations-classroom-activities-2.svg" />
+        </div>
+        <div>
+          <div className="welcome-modal-text">
+            <p className="welcome-modal-header">Want to see Quill&apos;s full potential?</p>
+            <p>Play around with a demo account to see sample student data and reports.</p>
+          </div>
+          <br />
+          <div className="welcome-modal-buttons">
+            <button className="quill-button contained secondary medium outlined welcome-modal-demo" onClick={goToDemo} type="button">Explore demo</button>
```
emilia-friedberg: You can just make this an anchor tag with the `quill-button` styling and then you can use a normal href rather than an onClick function for it.

services/QuillLMS/client/app/bundles/Teacher/components/dashboard/handy_actions.tsx
```diff
@@ -17,13 +18,21 @@ const HandyAction = ({ icon, text, link, }) => (
   </a>
 )
 
-const HandyActions = ({ linkedToClever, }) => (
+const ExploreDemoAction = ({setShowDemoModal}) => (
+  <a className="handy-action focus-on-light" onClick={() => setShowDemoModal(true)}>
```
emilia-friedberg: Since this isn't opening a link, we shouldn't use an anchor tag. Ideally this would be a button and you can adjust the styling so it looks like the rest of the handy actions.

emilia-friedberg (changes requested):
Looks really good overall! Small changes requested where you're using an HTML element with the behavior of a different HTML element, which can be an accessibility issue: https://www.w3.org/TR/WCAG20-TECHS/G115.html.

services/QuillLMS/app/lib/quill_authentication.rb
cissyxyu: This is here because if the user is previewing as a student from inside the Demo account, we need to bring them back into the demo account when they exit the student preview.

We can longer bring them back to the demo account through `session[:user_id]` because that field now contains the ID of the actual teacher, not the demo teacher.

You can think of it as a double-nested action. A teacher jumps into the Demo account, then from the Demo account jumps into a student account. When they jump out of the student account, we want to let them jump out one level back into the Demo account, not two levels back into their own account.

emilia-friedberg: That makes sense! Thanks for clarifying.

emilia-friedberg (approved):

services/QuillLMS/app/lib/quill_authentication.rb
```diff
@@ -68,10 +70,22 @@ def sign_in(user)
     @current_user = user
   end
 
+  def demo_id=(demo_id)
```
happythenewsad: What about `current_user_demo_id` instead of `demo_id`? This method modifies `@current user`, but the naming indicates that it is only modifying an instance variable.

happythenewsad (approved):
Looks good, besides my method naming suggestion.

services/QuillLMS/app/lib/quill_authentication.rb
```diff
@@ -81,6 +95,14 @@ def previewing_student_dashboard?
     !session[:preview_student_id].nil?
   end
 
+  def viewing_demo_account?
+    !session[:demo_id].nil?
```
anathomical: This is super-minor, but it might be cleaner to just make this `session[:demo_id].present?`.  Easier to parse without the `!` if it works here.

services/QuillLMS/app/lib/quill_authentication.rb
```diff
@@ -81,6 +95,14 @@ def previewing_student_dashboard?
...
+  end
+
+  def signed_in_outside_demo?
+    !session[:user_id].nil?
```
anathomical: Same note as above.

anathomical (approved):
Looks good to me.

PR #8579, opened 2021-12-06
fix bug with students removed from class still showing in diagnostic reports

emilia-friedberg:
## WHAT
Fix bug where students that had been removed from the class still showed in the diagnostic reports, and also bug where students weren't having their id removed from the `assigned_student_ids` array when their `students_classrooms` relationship got archived.

## WHY
If a teacher removes a student from a classroom, they shouldn't be able to see their data on any reports.

## HOW
Just scope the various queries we're using to only use ids from `assigned_student_ids` arrays, and also update the `remove_students` controller action to use a method that won't ignore validations and callbacks, since we are relying on a callback to remove the student id from `assigned_student_ids` arrays that it shouldn't be part of any longer.

### Screenshots

### Notion Card Links
https://www.notion.so/quill/Students-removed-from-class-still-appear-on-new-diagnostic-reports-cd83df57ae7d4a58b027fadc8da50f59

PR Checklist | Your Answer
------------ | -------------

### DIFF
--- a/services/QuillLMS/app/controllers/teachers/classrooms_controller.rb
+++ b/services/QuillLMS/app/controllers/teachers/classrooms_controller.rb
@@ -60,8 +60,7 @@ class Teachers::ClassroomsController < ApplicationController
   def remove_students
     StudentsClassrooms
       .where(student_id: params[:student_ids], classroom_id: params[:classroom_id])
-      .archive_all
-
+      .each { |sc| sc.archive }
     render json: {}
   end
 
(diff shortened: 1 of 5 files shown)


services/QuillLMS/app/controllers/teachers/classrooms_controller.rb
```diff
@@ -60,8 +60,7 @@ def create_students
   def remove_students
     StudentsClassrooms
       .where(student_id: params[:student_ids], classroom_id: params[:classroom_id])
-      .archive_all
```
brendanshean, line 63: Good catch on this one.  I learned something new.  https://stackoverflow.com/a/39646223

I noticed we use `update_all` in many places so there might be other places where this lack of callbacks might happen.

brendanshean (approved):
Nice work on fixing this bug.

PR #8584, opened 2021-12-07
Introduce a Caching Pattern

dandrabik:
## WHAT
This is a proposal for a caching pattern that I wanted feedback from the team. This is somewhere between a PR and an RFC. I'd love feedback about gotchas, what could be better, edge cases, concerns, etc.

The proposal is this: When child objects such as `activity_sessions` and `classroom_teachers` are updated, we would also change the `updated_at` on the parent (and/or parent of the parent), so then we could use the parent's `updated_at` as a canonical cache_key to signify "nothing has changed in this classroom if the `updated_at` didn't change"

## WHY
Using caching to improve site performance.

Most of our reports are aggregations of `activity_sessions` and `concept_results` for a given `student`, `classroom_unit`, `classroom`, or `teacher`. `ActivitySessions` rarely change after they are written, but we generate all of our reports from scratch on every page load.
## HOW
Use Rails' `touch: true` to update relations. Use a `CacheKey` class as a template for which cache_keys to use. Start implementing this in some controllers.

Note, the trade-off with this approach is that there are more DB writes on associations for every save. The upside is caching is easier.

We'd also need to eliminate anywhere that `updated_at` is used for different app logic.

### TODO
- [x] Add Tests
- [x] Eliminate any application logic based on `updated_at`
- [x] Ensure all applicable children update the appropriate parents

PR Checklist | Your Answer
------------ | -------------
Have you added and/or updated tests? |  Not yet, very early.
Have you deployed to Staging? | Not yet, still gathering feedback.


services/QuillLMS/app/models/cache_key.rb
```diff
@@ -0,0 +1,35 @@
+# Used to define common cache_keys in the app and force a common pattern
+module CacheKey
+
+  def self.all_classrooms(user, name:, groups: [])
+    classroom = user.classrooms_i_teach.maximum('classrooms.updated_at')
+
+    model_key(classroom, name: name, groups: groups)
+  end
+
+  def self.classroom(classroom, name:, groups: [])
```
happythenewsad: We can call `CacheKey.model_key` for these cases, since they are pure pass-through methods

services/QuillLMS/app/models/cache_key.rb
```diff
@@ -0,0 +1,35 @@
...
+      ClassroomUnit.where(classroom_id: classroom_id)
+        .joins(:unit, :unit_activities)
+        .where(unit: {unit_activities: {activity_id: activity_id}})
+        .maximum('classroom_units.updated_at')
+    end
+
+    model_key(classroom_unit, name: name, groups: groups)
+  end
+
+  def self.model_key(object, name:, groups: [])
```
happythenewsad: let's do some type checking here if we decide to deploy. Also, the first param can be a keyword param, like its siblings

services/QuillLMS/app/models/cache_key.rb
```diff
@@ -0,0 +1,35 @@
+# Used to define common cache_keys in the app and force a common pattern
+module CacheKey
+
+  def self.all_classrooms(user, name:, groups: [])
```
happythenewsad: I think it's better to not have app-specific logic in this module, but rather use `model_key` from the caller context, and pass in all the necessary ingredients from the caller context.

We could have `model_key` accept a block argument, so that it can execute arbitrary lambdas that are passed from the caller context.

happythenewsad:
Great start, this looks promising.

 I'm a bit leery of using `updated_at` for cache purposes, since it's a rails primitive and may be touched by other Rails code in non-obvious ways. I prefer a dedicated column, like `cached_at` or something. Or, we can just be ok with a lot of cache misses that are not true cache misses.

In addition, there's an opportunity for more automation for this feature. We can include a helper module in all of our active record models, which can introspect on their `belongs_to` relations, identify parents, and touch `parent.cached_at` by default.

services/QuillLMS/app/models/concerns/teacher.rb
```diff
@@ -32,10 +32,6 @@ def has_classrooms?
     classrooms_i_teach.any? && !classrooms_i_teach.all?(&:new_record?)
   end
 
-  def classrooms_i_teach
```
brendanshean, line 35: What's the reason for this removal?

services/QuillLMS/app/models/cache_key.rb
brendanshean: It's helpful in this RFC to see all of the proposed methods within CacheKey module, but I would agree with @happythenewsad, the methods should be in the caller context.

brendanshean (commented):
Based on what you've described about these particular objects not being updated often, I think the approach sounds sensible.  

As far as gotchas, I've used `touch: true` in the past for cache invalidation, but sometimes it's difficult to figure out why/why not `updated_at` was changed.  There's also a worry about the `touch: true` causing unexpected load on the database so it might be worth breaking out these caching calls incrementally in different PRs.  That way you can monitor their effect more systematically.

[read_multi](https://api.rubyonrails.org/v5.1.7/classes/ActiveSupport/Cache/Store.html#method-i-read_multi) might be worth looking into for the more complicated object dependencies (e.g. `CacheKey.classroom_unit_by_ids`).  Here's a [post](https://www.justinweiss.com/articles/a-faster-way-to-cache-complicated-data-models/) on it.

services/QuillLMS/app/models/concerns/teacher.rb
emilia-friedberg: Seconding this - I think it gets used pretty extensively.

services/QuillLMS/app/models/unit.rb
```diff
@@ -37,6 +37,7 @@ class Unit < ApplicationRecord
   validates_with UniqueNameWhenVisible
   belongs_to :user
   has_many :unit_activities, dependent: :destroy
+  # TODO might need a touch here
```
emilia-friedberg: I think you will - can you do to a touch to `classroom_units`? Otherwise I don't think your cache for the classroom unit will get invalidated if a user changes the name of the unit, or adds/removes an activity from it.

emilia-friedberg (commented):
Just added some very minor comments - I will return to this later if I think of anything else. I did a search for `updated_at` in the repo and I actually don't think the changes you're making will impact the places we're using it. I like the approach overall.

services/QuillLMS/app/controllers/teachers/progress_reports/diagnostic_reports_controller.rb
```diff
@@ -79,7 +90,20 @@ def recommendations_for_classroom
   end
 
   def lesson_recommendations_for_classroom
-    render json: {lessonsRecommendations: get_recommended_lessons(current_user, params[:unit_id], params[:classroom_id], params[:activity_id])}
+    classroom = current_user.classrooms.find(params[:classroom_id])
```
cissyxyu: I think we want to use `classrooms_i_teach` here.

services/QuillLMS/app/models/concerns/teacher.rb
anathomical: I suspect (and Dan can confirm), that the raw SQL used in the join here might mess up the auto caching built into Rails that we're trying to use here.  That said, I agree with Emilia that this is used extensively, so if that IS the problem, we probably just have to refactor this function to work with the caching system we want to use here.

services/QuillLMS/app/models/cache_key.rb
anathomical: I'm a bit torn.  Counter-point would be that it a central registry of cache-keys can help you avoid both double-naming things (and therefore having inaccurate "misses") when some other part of the code is already caching what you care about, and (more worryingly) collisions.

That said, if we use this pattern extensively, I could see this file having dozens of registered key patterns, at which point it's so unwieldy that it might not help solve the risk of collision anymore anyway.

dandrabik: @happythenewsad @brendanshean @anathomical 

I think it would be helpful to explain my thinking on this class further. 

We technically don't need this `CacheKey` class, it's entirely an object lookup and pass-through methods to form an array for `Rails.cache.fetch`. You can use `Rails.cache.fetch` directly fine enough.

My concern is that when using `Rails.cache.fetch`, it's relatively complicated to know which model objects are usable for their `updated_at` timestamp as cache keys, e.g. you have to confirm that every child object is properly updating its parent objects (which may have edge cases), and you have to validate that an object's `updated_at` ensures everything attached is up to date. I'm concerned that this overhead will discourage developers from using caching (or encourage them to just copy/paste from an example somewhere else in the code base).

My thought with this class is that it might be easier to have an `"approved list"` of `cache_keys` that have been vetted/validated so that:

1.  Developers can use these approved keys with only a little knowledge of the caching setup, and just knowledge of "Is the data I want to cache `all teacher classrooms` level, `individual classroom` level, `classroom_unit` level, etc."
1. It lowers the barrier of entry for adding caching.
1. The cache keys follow a standard pattern

I'd be interested to hear feedback on this.

dandrabik:
> Great start, this looks promising.
> 
> I'm a bit leery of using `updated_at` for cache purposes, since it's a rails primitive and may be touched by other Rails code in non-obvious ways. I prefer a dedicated column, like `cached_at` or something. Or, we can just be ok with a lot of cache misses that are not true cache misses.
> 
> In addition, there's an opportunity for more automation for this feature. We can include a helper module in all of our active record models, which can introspect on their `belongs_to` relations, identify parents, and touch `parent.cached_at` by default.

@happythenewsad It's a very common Rails pattern to use the `updated_at` column for caching. It's why the `touch: true` modifier on associations exists and why if you pass a model object to `Rails.cache.fetch` it forms a key based on its `id` and `updated_at` column. I think we'd want to follow this standard pattern.

services/QuillLMS/app/models/concerns/teacher.rb
dandrabik: I'll put this back in since that removal is unrelated to this PR. 

I removed this because `classrooms_i_teach` is actually defined in two places (it's also [defined on the User model as an association](https://github.com/empirical-org/Empirical-Core/blob/develop/services/QuillLMS/app/models/user.rb#L98)). I thiiink that other definition overwrites this one since Teacher is a mixin to User (and it appears above the association), but I'm not 100% sure.

I'm going to write up a card to address this (and not do it in this PR), since we should only have one definition.

services/QuillLMS/app/models/cache_key.rb
dandrabik: > We can call `CacheKey.model_key` for these cases, since they are pure pass-through methods

Yeah, see my longer comment about this. I'm aware this is a pass-through. My intention here was to mark `classroom` as a validated cache key that has all of the other updates in place to make it work properly. Happy to get feedback on that.

services/QuillLMS/app/models/unit.rb
dandrabik: Yeah, I'll write a custom callback to touch these items (there is no built in `touch` for `has_many`)

dandrabik:
> Based on what you've described about these particular objects not being updated often, I think the approach sounds sensible.
> 
> As far as gotchas, I've used `touch: true` in the past for cache invalidation, but sometimes it's difficult to figure out why/why not `updated_at` was changed. There's also a worry about the `touch: true` causing unexpected load on the database so it might be worth breaking out these caching calls incrementally in different PRs. That way you can monitor their effect more systematically.
> 
> [read_multi](https://api.rubyonrails.org/v5.1.7/classes/ActiveSupport/Cache/Store.html#method-i-read_multi) might be worth looking into for the more complicated object dependencies (e.g. `CacheKey.classroom_unit_by_ids`). Here's a [post](https://www.justinweiss.com/articles/a-faster-way-to-cache-complicated-data-models/) on it.

I really like the idea of rolling out some of the `touch: true` calls slowly to judge performance (I'm also concerned about the hit there).

When writing this, I did look into `read_multi` and its counterpart `fetch_multi`. My use case in this PR is to cache entire controller actions, i.e. cache 'single' items (the example in the blog post is caching multiple items and their associations). For our use case of a single end item, the downside of using `fetch_multi` to split logic is it computes all of the keys you pass in, and the downside of using  `read_multi` to split logic is you then have to have separate logic to write to the cache. I think, for our use case of a single item per call, I think finding the appropriate model item and then using `fetch` will be slightly more straightforward. But these are definitely methods we might use in the future.

services/QuillLMS/app/models/concerns/teacher.rb
eadams17: just want to drop a note that the share activity/activity pack feature that will be released tomorrow utilizes `classrooms_i_teach` but via the User model association

services/QuillLMS/app/models/cache_key.rb
brendanshean: Those are strong points.  I like the idea of the class being a place for "approved" cache_keys makes sense as this is a pattern we should be careful when using.   While I was worried about the file becoming too bloated, I think part of that was due to the generic name `CacheKey`.  What about a more specific name like: `LowLevelCacheKey`?  This would communicate the intent of the keys more effectively while preventing accidental inclusion of other types of cache keys.

services/QuillLMS/app/controllers/teachers/progress_reports/diagnostic_reports_controller.rb
emilia-friedberg: Cissy is right, calling `.classrooms` on a user looks for `classroom_students` relationships, not `classrooms_teacher`.

dandrabik: Nice catch, I'll make that change.

services/QuillLMS/app/models/cache_group_key.rb
```diff
@@ -0,0 +1,35 @@
...
+        .joins(:unit, :unit_activities)
+        .where(unit: {unit_activities: {activity_id: activity_id}})
+        .maximum('classroom_units.updated_at')
+    end
+
+    model_key(classroom_unit, name: name, groups: groups)
+  end
+
+  def self.model_key(object, name:, groups: [])
+    [name, *groups, object]
```
brendanshean: @dandrabik Not sure if it matters but to we care about groups being sorted?

dandrabik:
Closing this to re-open an updated PR (it's changed enough to warrant a new PR).

PR #8587, opened 2021-12-09
Save clever access_tokens when user authenticates

brendanshean:
## WHAT
When a teacher authenticates via clever, save their access token to the appropriate `auth_credential`. Also, delete the same auth_credential when it has expired.

## WHY
Currently, when a teacher logs in, all of the students and classes are imported immediately.  This is achieved without storing the access token since the controller immediately uses the access token after login.  If we want to use the access token later in a background job or ajax request, we need to store the appropriate token. 

## HOW
Use the AuthCredentialSaver to save the new credential.  Use the `clever_library` or `clever_district` provider depending on the type of integration.  After assigning the new credential, kick off a background job to run in 23 hours that will remove the expired worker.  I'm choosing to expire 1.hour before the actual expiration so as to avoid some edge cases.

### Screenshots

### Notion Card Links

PR Checklist | Your Answer
------------ | -------------

### DIFF
--- a/services/QuillLMS/app/controllers/api/v1/users_controller.rb
+++ b/services/QuillLMS/app/controllers/api/v1/users_controller.rb
@@ -3,25 +3,7 @@
 class Api::V1::UsersController < Api::ApiController
 
   def index
-    has_refresh_token = false
-    refresh_token_expires_at = nil
-    if current_user
-      user_id = current_user.id
-      auth_credential = AuthCredential.where(user_id: user_id).first
-      if auth_credential.present?
-        if auth_credential.refresh_token
-          has_refresh_token = true
-          refresh_token_expires_at = auth_credential&.expires_at&.in_time_zone&.utc&.to_s&.sub(' ','T')
-        end
-      end
-    end
-
-    render json: {
-      user: current_user,
-      text: "Hi",
-      has_refresh_token: has_refresh_token,
-      refresh_token_expires_at: refresh_token_expires_at
-    }
+    render json: { user: current_user }
   end
 
   def current_user_and_coteachers
--- a/services/QuillLMS/app/models/auth_credential.rb
+++ b/services/QuillLMS/app/models/auth_credential.rb
@@ -28,7 +28,11 @@ class AuthCredential < ApplicationRecord
   belongs_to :user
 
   GOOGLE_PROVIDER = 'google'
-  EXPIRATION_DURATION = 6.months
+  GOOGLE_EXPIRATION_DURATION = 6.months
+
+  CLEVER_DISTRICT_PROVIDER = 'clever_district'
+  CLEVER_LIBRARY_PROVIDER = 'clever_library'
+  CLEVER_EXPIRATION_DURATION = 24.hours
 
   def google_authorized?
     provider == GOOGLE_PROVIDER && refresh_token_valid?
@@ -41,6 +45,6 @@ class AuthCredential < ApplicationRecord
   end
 
   def refresh_token_expires_at
-    expires_at + EXPIRATION_DURATION
+    expires_at + GOOGLE_EXPIRATION_DURATION
   end
 end
--- a/services/QuillLMS/app/models/auth_credential.rb
+++ b/services/QuillLMS/app/models/auth_credential.rb
@@ -32,19 +32,25 @@ class AuthCredential < ApplicationRecord
 
   CLEVER_DISTRICT_PROVIDER = 'clever_district'
   CLEVER_LIBRARY_PROVIDER = 'clever_library'
-  CLEVER_EXPIRATION_DURATION = 24.hours
+  CLEVER_EXPIRATION_DURATION = 23.hours
 
   def google_authorized?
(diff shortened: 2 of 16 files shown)


services/QuillLMS/app/models/auth_credential.rb
```diff
@@ -41,6 +45,6 @@ def refresh_token_valid?
   end
 
   def refresh_token_expires_at
```
dandrabik, line 51: Note, I was looking for uses of this method since this is now a general method with a provider specific implementation, and there's a different calculation for [refresh_token_expires_at](https://github.com/empirical-org/Empirical-Core/blob/develop/services/QuillLMS/app/controllers/api/v1/users_controller.rb#L14) in the API controller.

services/QuillLMS/app/models/auth_credential.rb
```diff
@@ -41,6 +45,6 @@ def refresh_token_valid?
...
-    expires_at + EXPIRATION_DURATION
+    expires_at + GOOGLE_EXPIRATION_DURATION
```
dandrabik, line 54: This might be overkill, but since this generic code is now serving multiple providers, it might be worth defensively coding these methods to only return values for `google` auth credentials to prevent future edge cases, e.g.
```ruby
def refresh_token_expires_at
  return nil unless google?
  
  expires_at + GOOGLE_EXPIRATION_DURATION
end

def refresh_token_valid?
 return nil unless google?
 ...
end

File: services/QuillLMS/app/models/auth_credential.rb [hunk 3]
```diff
@@ -28,7 +28,11 @@ class AuthCredential < ApplicationRecord
   belongs_to :user
 
   GOOGLE_PROVIDER = 'google'
-  EXPIRATION_DURATION = 6.months
+  GOOGLE_EXPIRATION_DURATION = 6.months
+
+  CLEVER_DISTRICT_PROVIDER = 'clever_district'
+  CLEVER_LIBRARY_PROVIDER = 'clever_library'
+  CLEVER_EXPIRATION_DURATION = 24.hours
```
dandrabik: Take or leave, but if we are expiring the credential in `23.hours`, you could consider just storing that here for simplicity.

dandrabik (approved):
:+1: Small take or leave coding suggestions, but looks good to me.

services/QuillLMS/app/models/auth_credential.rb
brendanshean: Having to keep track of the offset elsewhere is error-prone so I definitely like this idea.

services/QuillLMS/app/models/auth_credential.rb
brendanshean: I saw that one.  At some point I would like to refactor all of the refresh_token stuff but I think we can leave as-is for now since this will just return nil for `refresh_token_expires_at` since the `auth_credential.refresh_token` conditional will return false for clever login.

services/QuillLMS/app/models/auth_credential.rb
brendanshean: This is a good point.

services/QuillLMS/app/models/auth_credential.rb
brendanshean: @dandrabik I looked at this code yesterday and decided it wasn't relevant.  On second pass today, as far as I can tell, those `has_refresh_token` and `refresh_token_expires_at` aren't used by that API.  I'm removing them but wanted to get your review again

services/QuillLMS/app/controllers/api/v1/users_controller.rb
```diff
@@ -3,25 +3,7 @@
...
-      end
-    end
-
-    render json: {
-      user: current_user,
-      text: "Hi",
-      has_refresh_token: has_refresh_token,
-      refresh_token_expires_at: refresh_token_expires_at
-    }
+    render json: { user: current_user }
```
dandrabik, line 6: Yeah, it looks like you removed the usage of the refresh token fields on the frontend in [this PR in July](https://github.com/empirical-org/Empirical-Core/pull/7851/files#diff-3cc3eb04a48dcab4cc4fd7b4aa44b4c9c831272f251878399631be49eeca036e) when removing the google re-auth code. So if that's no longer needed, this change seems good. 

Side note, I'm assuming/hoping the `text: "Hi!` is also vestigial (likely test) code. It was put in place in the initial controller code [in 2017](https://github.com/empirical-org/Empirical-Core/commit/d642a95212ba64b52d7faab1b75f1d91984fa11d) and kept being carried forward. I can't find references to it, but `text` is generic enough to be hard to search for.

dandrabik (approved):
:+1: Comment with a reference to confirm, but looks good to me! 🚢

services/QuillLMS/app/controllers/api/v1/users_controller.rb
brendanshean: I wasn't able to find anything and the requests to that API don't seem to use it.

PR #8588, opened 2021-12-09
query refactor

happythenewsad:
## WHAT & WHY

Optimizes an expensive query that was impacting our database a few weeks ago. Reduces query time by 142x (!) by removing two layers of subqueries and breaking the large query into three separate ones. 

I admit this is not the prettiest code, but it's easier to read and obviously way more performant. 

### Screenshots

### Notion Card Links
[(Please provide links to any relevant Notion card(s) relevant to this PR.)](https://www.notion.so/quill/Refactor-FindAdminUsersWorker-7807b13d460641598dc2dd91fcf896f9)

PR Checklist | Your Answer
------------ | -------------

### DIFF
--- a/services/QuillLMS/app/queries/teachers_data.rb
+++ b/services/QuillLMS/app/queries/teachers_data.rb
@@ -18,7 +18,7 @@ module TeachersData
     return [] if teacher_ids.blank?
 
     teacher_ids_str = teacher_ids.join(', ')
-    number_of_students = User.find_by_sql(
+    number_of_students_query = User.find_by_sql(
       "SELECT
         users.id,
         users.name,
@@ -31,7 +31,8 @@ module TeachersData
       WHERE users.id IN (#{teacher_ids_str})
       GROUP BY users.id"
     )
-    number_of_questions_completed = User.find_by_sql(
+
+    question_count_query = User.find_by_sql(
       "SELECT 
         users.id,
         COUNT(DISTINCT concept_results.id) AS number_of_questions_completed
@@ -44,6 +45,7 @@ module TeachersData
       AND activity_sessions.state = 'finished'
       GROUP BY users.id"
     )
+
     time_spent_query = User.find_by_sql(
       "SELECT 
         users.id,
@@ -58,7 +60,7 @@ module TeachersData
     )
 
     combiner = {}
-    number_of_students.each do |row|
+    number_of_students_query.each do |row|
       combiner[row.id] = {
         name: row.name,
         email: row.email, 
@@ -66,14 +68,14 @@ module TeachersData
       }
     end
 
-    number_of_questions_completed.each do |row|
+    question_count_query.each do |row|
       combiner[row.id][:number_of_questions_completed] = row.number_of_questions_completed
     end
 
     time_spent_query.each do |row|
       combiner[row.id][:time_spent] = row.time_spent
     end
-    
+
     result = combiner.keys.map do |key|
       hash_value = combiner[key]
       UserStruct.new(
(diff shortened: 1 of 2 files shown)


services/QuillLMS/app/queries/teachers_data.rb
```diff
@@ -12,34 +12,95 @@ module TeachersData
...
+      combiner[row.id][:number_of_questions_completed] = row.number_of_questions_completed
+    end
+
+    time_spent_query.each do |row|
+      combiner[row.id][:time_spent] = row.time_spent
+    end
+
+    result = combiner.keys.map do |key|
+      hash_value = combiner[key]
+      user = User.new(
```
happythenewsad, line 80: I attempted to use a Struct here, but we need AR models here because the serializer expects them.

services/QuillLMS/app/queries/teachers_data.rb
```diff
@@ -16,30 +16,82 @@ def self.run(teacher_ids)
...
+      FROM users 
+      INNER JOIN units ON users.id = units.user_id
+      INNER JOIN classroom_units ON units.id = classroom_units.unit_id
+      INNER JOIN activity_sessions ON classroom_units.id = activity_sessions.classroom_unit_id
+      WHERE users.id IN (#{teacher_ids_str})
+      AND activity_sessions.state = 'finished'
+      GROUP BY users.id"
+    )
+
+    combiner = {}
```
happythenewsad, line 61: Using a hash lookup table here to avoid an O(n^2) merge operation.

services/QuillLMS/app/queries/teachers_data.rb
```diff
@@ -16,30 +16,82 @@ def self.run(teacher_ids)
...
+    end
+
+    result = combiner.keys.map do |key|
+      hash_value = combiner[key]
+      user = User.new(
+        id: key,
+        name: hash_value[:name],
+        email: hash_value[:email]
+      )
+      user.define_singleton_method(:number_of_students) do
```
happythenewsad, line 85: This isn't as weird as it seems - `find_by_sql`, which was used previously, does this under the hood.

services/QuillLMS/spec/queries/teachers_data_spec.rb
```diff
@@ -41,27 +41,27 @@
   let!(:concept_result2) { create(:concept_result, concept: concept2, activity_session: activity_session2) }
 
   before :each do
-    @result = teachers_data_module.run(teacher_ids).first
+    @result = teachers_data_module.run(teacher_ids)
```
brendanshean: If `.first` is removed here, should the instance variable be `@results`

services/QuillLMS/app/queries/teachers_data.rb
brendanshean: It might make sense to introduce a tableless class to actually model this since it's not really a `User` class anymore.  Here's a POC: https://github.com/empirical-org/Empirical-Core/commit/ced2b8442d57b61abe495c88bb41241bd9eb60b7

brendanshean (approved):
Great work on breaking this up.  I think it's definitely more readable.  I added a note about moving that `user` object into a new model that might make this more transparent while eliminating the `singleton_methods`.

services/QuillLMS/app/queries/teachers_data.rb
brendanshean: This is a very interesting approach.

services/QuillLMS/app/queries/teachers_data.rb
happythenewsad: Thanks for floating that option, but my preference is not adding a new shadow model.

services/QuillLMS/spec/queries/teachers_data_spec.rb
happythenewsad: 👍 Addressing now

happythenewsad:
Although @anathomical is ill today, I plan to wait until he can review this PR (rather than delegating to someone else), since he has already partially reviewed an earlier version of the PR.

anathomical (approved):

PR #8590, opened 2021-12-09
use the react-router location, not the window.location for the querystring in diagnostic reports

emilia-friedberg:
## WHAT
Make sure we're pulling `location` from props rather than from the window in the diagnostic reports.

## WHY
We are calling `location.search` to get the unit id in these reports, but it was using `window.location` instead of the `react-router` location and coming up undefined.

## HOW
Just make sure we're actually using the prop.

### Screenshots

### Notion Card Links
https://www.notion.so/quill/Skills-Survey-results-overriden-when-reassigned-f61b4f38d5d34923b7e1ba78a217080c

PR Checklist | Your Answer
------------ | -------------

### DIFF
--- a/services/QuillLMS/client/app/bundles/Teacher/components/progress_reports/diagnostic_reports/diagnostics/results.tsx
+++ b/services/QuillLMS/client/app/bundles/Teacher/components/progress_reports/diagnostic_reports/diagnostics/results.tsx
@@ -89,8 +89,6 @@ export const Results = ({ passedStudentResults, passedSkillGroupSummaries, match
   }, [openPopover])
 
   function getResults() {
-    debugger;
-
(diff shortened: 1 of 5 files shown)


services/QuillLMS/client/app/bundles/Teacher/components/progress_reports/diagnostic_reports/diagnostics/results.tsx
```diff
@@ -89,6 +89,7 @@ export const Results = ({ passedStudentResults, passedSkillGroupSummaries, match
   }, [openPopover])
 
   function getResults() {
+    debugger;
```
eadams17: forgot to remove this

eadams17 (approved):

PR #8597, opened 2021-12-10
Nr for evidence

happythenewsad:
## WHAT
We want to surface user_id in Evidence controllers in New Relic for diagnostic purposes. This behavior is already in place for the root app. 

## HOW
Check that NewRelicAttributable module is loaded before including it in ApiController. This way, the concern is completely managed by the root app, and the engine doesn't need to manage a separate NewRelic config, or even load NewRelicAttributable.

### Screenshots

### Notion Card Links
[(Please provide links to any relevant Notion card(s) relevant to this PR.)](https://www.notion.so/quill/Add-user_id-to-NewRelic-for-Quill-Evidence-279bf0eaf8434b3ab89ccaa092fd9d4f)

PR Checklist | Your Answer
------------ | -------------

### DIFF
--- a/services/QuillLMS/spec/controllers/concerns/new_relic_attributable_spec.rb
+++ b/services/QuillLMS/spec/controllers/concerns/new_relic_attributable_spec.rb
@@ -34,4 +34,19 @@ describe NewRelicAttributable, type: :controller do
       get :index
     end
   end
+
+  context 'Evidence engine' do 
+    controller(Evidence::ApiController) do
+  
+      def index
+        head :ok
+      end
+    end
+
+    it 'should pass user_id to NewRelic' do
+      allow(controller).to receive(:current_user) { user }
+      expect(::NewRelic::Agent).to receive(:add_custom_attributes).with({user_id: user.id})
+      get :index
+    end
+  end
 end
--- a/services/QuillLMS/engines/evidence/app/controllers/evidence/api_controller.rb
+++ b/services/QuillLMS/engines/evidence/app/controllers/evidence/api_controller.rb
@@ -4,6 +4,10 @@ require_dependency 'evidence/application_controller'
 
 module Evidence
   class ApiController < ApplicationController
+    if Object.const_defined?('NewRelicAttributable') 
+      include NewRelicAttributable 
+    end
+    
     skip_before_action :verify_authenticity_token
   end
 end
--- a/services/QuillLMS/engines/evidence/lib/evidence/engine.rb
+++ b/services/QuillLMS/engines/evidence/lib/evidence/engine.rb
@@ -20,5 +20,11 @@ module Evidence
 
       g.fallbacks[:shoulda] = :test_unit
     end
+
+    config.after_initialize do 
+      if self.const_defined? (:NewRelicAttributable) 
+        ApplicationController.module_eval { include NewRelicAttributable }
+      end
+    end
   end
 end
--- a/services/QuillLMS/engines/evidence/lib/evidence/engine.rb
+++ b/services/QuillLMS/engines/evidence/lib/evidence/engine.rb
@@ -20,11 +20,5 @@ module Evidence
 
       g.fallbacks[:shoulda] = :test_unit
     end
-
-    config.after_initialize do 
-      if self.const_defined? (:NewRelicAttributable) 
-        ApplicationController.module_eval { include NewRelicAttributable }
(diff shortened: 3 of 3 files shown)


happythenewsad:
@dandrabik This PR could use some love

services/QuillLMS/spec/controllers/concerns/new_relic_attributable_spec.rb
```diff
@@ -34,4 +34,19 @@ def index
...
+  context 'Evidence engine' do 
+    controller(Evidence::ApiController) do
+  
+      def index
+        head :ok
+      end
+    end
+
+    it 'should pass user_id to NewRelic' do
+      allow(controller).to receive(:current_user) { user }
```
dandrabik, line 47: Have you tested this on a staging environment or with New Relic locally? I think this test setup (stubbing `current_user` and putting the test in the LMS) might be covering up that `current_user` isn't accessible by `Evidence::ApiController` controllers and that the `New::Relic` agent also may not be accessible.

It might we'd need to pass the New Relic agent to the engine [here](https://github.com/empirical-org/Empirical-Core/blob/develop/services/QuillLMS/config/initializers/evidence.rb) and copy and re-write the concern for the engine (it's duplication, but it's only a few lines of code). You should be able to put a test in the Evidence engine that uses the dummy app and passes. That should be the threshold.

dandrabik (changes requested):
Thanks for starting on this. I believe this needs more testing and a different setup to work properly. See my comments for additional details (also happy to have a quick chat).

services/QuillLMS/spec/controllers/concerns/new_relic_attributable_spec.rb
happythenewsad: > Have you tested this on a staging environment or with New Relic locally? I think this test setup (stubbing `current_user` and putting the test in the LMS) might be covering up that `current_user` isn't accessible by `Evidence::ApiController` controllers and that the `New::Relic` agent also may not be accessible.
> 
> It might we'd need to pass the New Relic agent to the engine [here](https://github.com/empirical-org/Empirical-Core/blob/develop/services/QuillLMS/config/initializers/evidence.rb) and copy and re-write the concern for the engine (it's duplication, but it's only a few lines of code). You should be able to put a test in the Evidence engine that uses the dummy app and passes. That should be the threshold.

Thanks for raising these concerns. Here is a demonstration of this PR currently working as is:
Screen Shot 2021-12-15 at 1 51 31 PM

happythenewsad: link from the screenshot: https://one.newrelic.com/launcher/nr1-core.explorer?platform[timeRange][duration]=1800000&pane=eyJuZXJkbGV0SWQiOiJlcnJvcnMtdWkub3ZlcnZpZXciLCJiYXJjaGFydCI6ImJhcmNoYXJ0IiwidG9wRmFjZXQiOiJ0cmFuc2FjdGlvblVpTmFtZSIsInBhZ2UiOiJzaG93IiwicHJpbWFyeUZhY2V0IjoiZXJyb3IuY2xhc3MiLCJ0cmFjZUlkIjoiODIyNGNjMTYtNWRlOC0xMWVjLWI5YzEtMDI0MmFjMTEwMDBhXzBfNjg1NyIsImVudGl0eUd1aWQiOiJNall6T1RFeE0zeEJVRTE4UVZCUVRFbERRVlJKVDA1OE5UVXhPRFE0TVRRdyJ9&sidebars[0]=eyJuZXJkbGV0SWQiOiJucjEtY29yZS5hY3Rpb25zIiwiZW50aXR5R3VpZCI6Ik1qWXpPVEV4TTN4QlVFMThRVkJRVEVsRFFWUkpUMDU4TlRVeE9EUTRNVFF3Iiwic2VsZWN0ZWROZXJkbGV0Ijp7Im5lcmRsZXRJZCI6ImVycm9ycy11aS5vdmVydmlldyJ9fQ==&state=6be8de5e-a806-144b-0a00-cb44bca49115

dandrabik (approved):
After discussing, @happythenewsad confirmed that this is working, but we'll break off a different card to clean up HOW it's working (from Peter's investigation, it's mixing in `current_user` via `RailsAdmin`, which is confusing). Shipping this so we get this additional tracking.

PR #8601, opened 2021-12-10
Add app_settings to Demo front end to show only 50% of users

cissyxyu:
## WHAT
We want to try rolling out this feature with `app_settings` to see if we can notice any effects on users, and also to see how the app settings feature integrates with Heap (for the bigger v2 Demo rollout later).

The app setting record I will create in the DB will be called `demo` and I will run this rake task to create it:
`bundle exec rake 'app_settings:create[demo,true,true,50,[]]'`

## WHY
So we can get more comfortable with app_settings and doing more A/B testing rollouts in the future.

## HOW
Add a check before showing the link to see if the app settings permits this user to see the link.
I'll run a rake task in production to add this particular app setting to the DB.

### Screenshots

### Notion Card Links

PR Checklist | Your Answer
------------ | -------------

### DIFF
--- a/services/QuillLMS/client/app/bundles/Teacher/components/dashboard/handy_actions.tsx
+++ b/services/QuillLMS/client/app/bundles/Teacher/components/dashboard/handy_actions.tsx
@@ -10,6 +10,7 @@ import {
   groupAccountIcon,
   googleClassroomIcon,
 } from '../../../Shared/index'
+import { handleHasAppSetting } from "../../../Shared/utils/appSettingAPIs";
 
 const HandyAction = ({ icon, text, link, }) => (
   <a className="handy-action focus-on-light" href={link}>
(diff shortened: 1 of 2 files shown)


services/QuillLMS/client/app/bundles/Teacher/components/dashboard/handy_actions.tsx
```diff
@@ -25,18 +26,22 @@ const ExploreDemoAction = ({setShowDemoModal}) => (
   </button>
 )
 
-const HandyActions = ({ linkedToClever, setShowDemoModal}) => (
-  <section className="handy-actions">
+const HandyActions = ({ linkedToClever, setShowDemoModal}) => {
+  const [errors, setErrors] = React.useState<string[]>([])
+  const [hasAppSetting, setHasAppSetting] = React.useState<boolean>(false);
+  if (!hasAppSetting) { handleHasAppSetting({appSettingSetter: setHasAppSetting, errorSetter: setErrors, key: 'demo', }) }
```
emilia-friedberg: This should go in a `React.useEffect` hook, with the second argument as an empty array - that's effectively the functional equivalent of the old `componentDidMount` method. That will have it only run once. Right now it's going to run every time the component rerenders, which shouldn't be a huge deal in your use case but isn't a pattern we should be following.

emilia-friedberg (changes requested):
Small change requested, looks good overall!

happythenewsad (approved):
Approving, pending Emilia's change request

services/QuillLMS/client/app/bundles/Teacher/components/dashboard/handy_actions.tsx
```diff
@@ -25,18 +26,24 @@ const ExploreDemoAction = ({setShowDemoModal}) => (
...
 )
 
-const HandyActions = ({ linkedToClever, setShowDemoModal}) => (
-  <section className="handy-actions">
+const HandyActions = ({ linkedToClever, setShowDemoModal}) => {
+  const [errors, setErrors] = React.useState<string[]>([])
+  const [hasAppSetting, setHasAppSetting] = React.useState<boolean>(false);
+  React.useEffect(() => {
+    handleHasAppSetting({appSettingSetter: setHasAppSetting, errorSetter: setErrors, key: 'demo', })
+  });
```
emilia-friedberg: Nitpicky but you want this line to be `}, [])` - otherwise this will still run on every re-render.

emilia-friedberg (approved):
Approving when final change is made.

PR #8624, opened 2021-12-17
make sure we can create a new evidence activity

emilia-friedberg:
## WHAT
Fix issue where new Evidence activities weren't saving because they didn't yet have a parent activity.

## WHY
We want new Evidence activities to be creatable.

## HOW
Just call `set_parent_activity` at the top of the flags file, and also hide the `Play Activity` buttons until there is actually an activity saved to play because they won't work otherwise (unrelated to the bug but something called out in the card).

### Screenshots

### Notion Card Links
https://www.notion.so/quill/Save-button-not-working-in-Evidence-CMS-ff888c7cbb9b433aac0aa8acf80300e0

PR Checklist | Your Answer
------------ | -------------


emilia-friedberg:
Trying to figure out what's going on with that one test

brendanshean (approved):
Thank you for the quick fix on this!

PR #8631, opened 2021-12-20
update activity summary etc links to go to diagnostic reports

emilia-friedberg:
## WHAT
Update all links to reports for a diagnostic activity to go to the diagnostic reports, and add an empty state for old diagnostics that don't have skills associated with them.

## WHY
We were having teachers who followed a link to a report for a diagnostic activity from a non-diagnostic page see the normal report, but they were getting confused about where to find the recommendations. This way they will always be taken to the view with the recommendations.

## HOW
Just add some logic to handle the case where the activity is a diagnostic in the existing redirect links, and add a new path for redirection so we can use it for the dashboard activity feed.

### Screenshots

### Notion Card Links
https://www.notion.so/quill/Help-teachers-find-their-practice-recommendations-and-diagnostics-assigned-multiple-times-be23c243d38d4d6194e9275a4741cfdf

PR Checklist | Your Answer
------------ | -------------


services/QuillLMS/app/models/concerns/public_progress_reports.rb
```diff
@@ -14,11 +14,11 @@ def last_completed_diagnostic
                   .first
     end
 
-    def activity_session_report(classroom_unit_id, user_id, activity_id)
-      classroom_unit = ClassroomUnit.find(classroom_unit_id)
-      unit_id = classroom_unit.unit_id
-      classroom_id = classroom_unit.classroom_id
-      if unit_id && activity_id && classroom_id
+    def activity_session_report(unit_id, classroom_id, user_id, activity_id)
+      if Activity.diagnostic_activity_ids.include?(activity_id.to_i)
```
anathomical, line 18: Not worth doing in this PR, I don't think, but the number of places we have this particular `include?` check makes me think that maybe we should have a an `Activity.is_diagnostic_id?()` function that does the `to_i` and `include?` check just to simplify all the call sites.  Something to consider for later.

services/QuillLMS/spec/controllers/teachers/progress_reports/diagnostic_reports_controller_spec.rb
```diff
@@ -19,12 +19,72 @@
...
       it "redirects to the correct page" do
-        get :report_from_classroom_unit_activity_and_user, params: ({classroom_unit_id: classroom_unit.id, user_id: student.id, activity_id: activity.id})
+        get :report_from_classroom_unit_and_activity_and_user, params: ({classroom_unit_id: classroom_unit.id, user_id: student.id, activity_id: activity.id})
         expect(response).to redirect_to("/teachers/progress_reports/diagnostic_reports#/u/#{unit.id}/a/#{activity.id}/c/#{classroom.id}/student_report/#{student.id}")
       end
     end
   end
 
+  describe '#redirect_to_report_for_most_recent_activity_session_associated_with_activity_and_unit' do
+  describe 'when the activity is a diagnostic' do
```
anathomical: Very tiny detail, but it looks like you're missing a level of indentation here which leads to the `end` on line 86 being under-indented.  This doesn't break the tests, obviously, but does make it a bit harder to read.

anathomical (approved):
One indentation comment in your test files, but this overall looks really good.  I like how you organized your test files and covered all the edge cases that we know we care about.

PR #8636, opened 2021-12-22
deleting active activity sessions

emilia-friedberg:
## WHAT
Update how we delete active activity sessions so that we do it from the controller action that completes the `activity_session`, rather than making another request from the frontend once the `activity_session` has been completed.

## WHY
I discovered, as a result of investigating this Sentry error: https://sentry.io/organizations/quillorg-5s/issues/1631395843/?project=210579&query=is%3Aunresolved&statsPeriod=14d, that the request to delete the active activity session wasn't actually always being sent, because it often didn't finish executing before the redirect occurred. Rather than rewrite how that worked for every activity type, I just removed the code that was supposed to be handling it on the frontend and updated the controller action to implement that functionality itself. I think there is also a case to be made that this should just be done in a callback on `activity_session` when `completed_at` changes, but this seemed like the closest mimic of the existing behavior.

## HOW
Have the `update` controller action for `activity_sessions` delete the `active_activity_session` if there is one and the activity session has successfully been completed, and delete all the code that pertained to the older way of doing things.

### Screenshots

### Notion Card Links
https://www.notion.so/quill/Sentry-Error-JS-Deleting-ActiveActivitySession-TypeError-a7ef31f5008b484c949dc16c6a747848

PR Checklist | Your Answer
------------ | -------------
Spec Review: Have you reviewed the spec and ensured this meets requirements and/or matches design mockups? |YES

### DIFF
--- a/services/QuillLMS/app/controllers/api/v1/active_activity_sessions_controller.rb
+++ b/services/QuillLMS/app/controllers/api/v1/active_activity_sessions_controller.rb
@@ -1,7 +1,7 @@
 # frozen_string_literal: true
 
 class Api::V1::ActiveActivitySessionsController < Api::ApiController
-  before_action :activity_session_by_uid, only: [:show, :destroy]
+  before_action :activity_session_by_uid, only: [:show]
 
   def show
     render json: @activity_session.as_json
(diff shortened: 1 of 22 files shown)


anathomical (approved):
Looks good.  I very much approve of this approach.  Way better than the old way that was so easy to break.

happythenewsad (approved):
👍

PR #8637, opened 2021-12-23
Bug-fix/ fix activity completion date for Activity Summary activity tooltip

eadams17:
## WHAT
fix activity completion date for Activity Summary activity tooltip

## WHY
we want to display the date for the most recently completed activity session

## HOW
use `updated` value rather than the first score's `completed_at` value in the `scores` array

### Screenshots

### Notion Card Links
https://www.notion.so/quill/Activity-Summary-date-does-not-update-with-latest-attemp-45227e8bf7d943e58e6e827fd96b231e

PR Checklist | Your Answer
------------ | -------------

### DIFF
--- a/services/QuillLMS/client/app/bundles/Teacher/components/general_components/tooltip/activity_details.jsx
+++ b/services/QuillLMS/client/app/bundles/Teacher/components/general_components/tooltip/activity_details.jsx
@@ -1,54 +1,58 @@
 import React from 'react';
 import moment from 'moment';
-import activityTypeFromClassificationId from '../../modules/activity_type_from_classification_id.js';
 
-export default class ActivityDetails extends React.Component {
-  getClassName = () => {
-    if (this.props.data.concept_results && this.props.data.concept_results.length) {
+export const ActivityDetails = ({ data }) => {
+
+  if (!Object.keys(data).length) { return <span /> }
+
+  const { concept_results, started_at, updated, scores, activity_description } = data;
+
+  function getClassName() {
+    if (concept_results && concept_results.length) {
       return 'activity-details';
     }
     return 'activity-details no-concept-results';
   };
 
-  detailOrNot = () => {
-    let dateTitle,
-    dateBody;
-    if (!this.props.data.concept_results || !this.props.data.concept_results.length) {
-      if (this.props.data.started_at) {
+  function detailOrNot() {
+    let dateTitle, dateBody;
+    if (!concept_results || !concept_results.length) {
+      if (started_at) {
         dateTitle = 'Started'
-        dateBody = this.props.data.started_at
+        dateBody = started_at
       }
     } else {
-      const firstScore = this.props.data.scores[0]
-      const firstCr = this.props.data.concept_results[0];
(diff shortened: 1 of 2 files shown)


anathomical:
Quick note: this payload is generated by `app/queries/scorebook/query.rb:self.run`.  It may make sense to add a `completed_at` column to that query and then use that instead of `updated_at`, but in all honesty there's a good chance those two values will be identical.  (As I'm not sure we ever update a record once it's completed, so the `updated_at` should have the same value.)

services/QuillLMS/client/app/bundles/Teacher/components/general_components/tooltip/activity_details.jsx
```diff
@@ -1,54 +1,58 @@
...
-import activityTypeFromClassificationId from '../../modules/activity_type_from_classification_id.js';
 
-export default class ActivityDetails extends React.Component {
-  getClassName = () => {
-    if (this.props.data.concept_results && this.props.data.concept_results.length) {
+export const ActivityDetails = ({ data }) => {
+
+  if (!Object.keys(data).length) { return <span /> }
+
+  const { concept_results, started_at, updated, scores, activity_description } = data;
```
emilia-friedberg, line 8: If this is coming from the place Thomas mentioned, which I think it is, shouldn't this be `updated_at`?

services/QuillLMS/client/app/bundles/Teacher/components/general_components/tooltip/activity_details.jsx
```diff
@@ -1,54 +1,58 @@
...
-      const firstScore = this.props.data.scores[0]
-      const firstCr = this.props.data.concept_results[0];
-      if (firstScore && firstScore.completed_at) {
+      const scoresExist = scores && scores.length;
+      const firstCr = concept_results[0];
+      // check if scores exist and use updated for most recent activity completion date
+      if (scoresExist && updated) {
         dateTitle = 'Completed';
-        dateBody = firstScore.completed_at;
+        dateBody = updated;
```
emilia-friedberg, line 30: Same here?

services/QuillLMS/client/app/bundles/Teacher/components/general_components/tooltip/activity_details.jsx
eadams17: @emilia-friedberg @anathomical that's correct, the reason it's `updated` is because of how the data was organized here in the Scorebook component: https://github.com/empirical-org/Empirical-Core/blob/develop/services/QuillLMS/client/app/bundles/Teacher/containers/Scorebook.jsx#L209

so I suppose I could change that to `updated_at` if y'all would prefer that?

emilia-friedberg: Oh, I see. I would just go with what's already there - though I hate it. It would be nice to clean up some of these files at some point.

emilia-friedberg (approved):

PR #8638, opened 2021-12-29
Allow teachers to choose which Clever classes are synced to Quill

brendanshean:
## WHAT
There are ~~a couple~~ three new flows involving Clever importing.

Flow 1: Clever Classroom importing
1.  Users will notice 'Import from Clever Classrooms' button at the top:
import-clever-classrooms
 2.  When users click, a dialog box will show any classes that are able to be imported:
import-clever-classrooms-modal
3.  If there are no classes left to import, and empty modal will be show instead:
clever-empty-classrooms-modal

Flow 2: Clever Classroom Student Importing
4.  If the user opens a specific classroom, there will be a "Import Clever classroom students' button:
import-clever-students-button
5. A modal will show information pertaining to the import of students for that classroom:
import-clever-students-modal

Flow 3: While Assigning Activities to students, import a clever classroom.
This basically follows the same flow as Flow 1 except it occurs inside of the Assigning Activities flow
Screen Shot 2022-01-19 at 1 33 49 PM

NB:  In both of these flows if the teacher's access token is expired, a reauthorization modal will appear before any subsequent importing can be performed:
reauthorize-clever

## WHY
Currently, every time a clever teacher logs in, all of their classrooms and students are imported. Allowing teachers to choose which classes they want to import would be a preferable system.

## HOW
From a high level, we add a new controller `CleverIntegration::TeachersController` with three routes which will handle importing classrooms, importing students, and retrieving classrooms.  From there I modeled the design similar to the way GoogleClassrooms were implemented.  The major difference lies with the ReauthorizeCleverModal.  

With GoogleClassroom, users access tokens (via refresh tokens) can be used for up to 6 months before expiring.  Clever does not offer refresh tokens so we are left with access tokens that expire in 24 hours.  As a result, we need to ask the user to Reauthorize access for import_classrooms and import_students methods when their auth_credential is `nil`.

The following is a list of commits with a little more description

1) Restore previous scoping for provider classrooms methods: in the previous PR, I had made the `google_classrooms` and `clever_classrooms` methods unscoped, but I've opted to remove the `unscoped` so as to preserve functionality of `google_classrooms`.
1) On signin, only import existing clever classrooms.  The rest of the PR will address the use case where teachers select which clever classrooms they would like to import. 
1) Refactor TeacherClassroomsRetriever to `TeacherClassroomsCacheHydrator`: while we are technically retrieving the classroom data, that data is really only cached and then used by another method.
1) Add `CleverIntegration::TeachersController#import_classrooms` route:  This endpoint will be used to import specific clever classrooms and their respective students that a teacher selects via the UI.
1) Add `CleverIntegration::TeachersController#import_students` route: This endpoint will be used to import students for a specific clever classroom which a teacher selects via the UI.
1) Add `CleverIntegration::TeachersController#retrieve_classrooms` route: This endpoint will retrieve cached classrooms data for a given teacher.  If the teacher's access token is expired it will first route for reauthorization.  If the cache is empty for that teacher, then the cache will be hydrated
1) Add some styling for clever modals
1) Add `clever_link` and `CLEVER_REDIRECT` so that reauthorization modal functions:  The clever reauthorization link is needed so that users can reauthorize.  Once reauthorized, utilize the `CLEVER_REDIRECT`
1) Add Clever Classroom data:  this data will be used in the various frontend tests
1) Add `ImportCleverClassroomStudentsModal`: this modal is responsible for importing students for an individual classroom
1) Add `ImportCleverClassroomModal`: this modal is responsible for importing classrooms and their respective students
1) Add `CleverClassroomsEmptyModal`: this modal informs the user there are no classrooms to import.
1) Add `ReauthorizeCleverModal`: this modal informs a teacher that the need to reauthorize with clever since their access token has expired
1) Add Clever Classroom importing to `ActiveClassrooms`: orchestrate all the relevant clever modals with the ActiveClassroom component.

### Notion Card Links
https://www.notion.so/quill/Clever-Allow-teachers-to-choose-which-Clever-classes-are-synced-to-Quill-f3dd936034bb40b7a746818f586d87c1
https://www.notion.so/quill/Clever-Allow-individual-Clever-classes-to-be-synced-on-demand-cbc3b5ded2bb40c888db1b88a9b78e8a

PR Checklist | Your Answer
------------ | -------------
Have you added and/or updated tests? |   YES


services/QuillLMS/client/app/bundles/Teacher/components/classrooms/__tests__/__snapshots__/import_google_classroom_students_modal.test.jsx.snap
```diff
@@ -33,7 +33,7 @@ exports[`ImportGoogleClassroomStudentsModal component should render ImportGoogle
           onClick={[Function]}
         />
         <span>
-          I understand that newly imported students have access to the activities that have already been assigned to the entire class.
+          I understand that newly imported students have access to the activities that have already been assigned the entire class.
```
emilia-friedberg: I think we want to keep the `to` in this sentence, no?

services/QuillLMS/client/app/bundles/Teacher/components/classrooms/active_classrooms.tsx
```diff
@@ -62,26 +72,17 @@ export default class ActiveClassrooms extends React.Component<ActiveClassroomsPr
...
-
-    this.openModal = this.openModal.bind(this)
-    this.closeModal = this.closeModal.bind(this)
-    this.showSnackbar = this.showSnackbar.bind(this)
-    this.onSuccess = this.onSuccess.bind(this)
-    this.clickClassroomHeader = this.clickClassroomHeader.bind(this)
-    this.clickImportGoogleClassrooms = this.clickImportGoogleClassrooms.bind(this)
-    this.getClassroomsAndCoteacherInvitations = this.getClassroomsAndCoteacherInvitations.bind(this)
-    this.getGoogleClassrooms = this.getGoogleClassrooms.bind(this)
-    this.setStateBasedOnParams = this.setStateBasedOnParams.bind(this)
```
emilia-friedberg, line 76: Thanks for this refactor!

services/QuillLMS/client/app/bundles/Teacher/components/classrooms/active_classrooms.tsx
```diff
@@ -101,19 +102,61 @@ export default class ActiveClassrooms extends React.Component<ActiveClassroomsPr
...
       this.setState({ showModal, selectedClassroomId })
     }
+  }
+
+  getCleverClassrooms = () => {
+    const { user } = this.props
+    const { clever_id } = user
+    const { attemptedImportCleverClassrooms } = this.state
+
+    if (clever_id) {
```
emilia-friedberg: I'd do a `if (!clever_id) { return }` block at the top to avoid at least one of the levels of nesting here.

services/QuillLMS/client/app/bundles/Teacher/components/classrooms/active_classrooms.tsx
```diff
@@ -101,19 +102,61 @@ export default class ActiveClassrooms extends React.Component<ActiveClassroomsPr
...
   }
 
-  getGoogleClassrooms() {
-    if (this.props.user.google_id) {
+  getGoogleClassrooms = () => {
+    const { user } = this.props
+    const { clever_id, google_id } = user
+    const { attemptedImportGoogleClassrooms } = this.state
+
+    if (!clever_id && google_id) {
```
emilia-friedberg, line 164: Just curious about the thinking behind this - theoretically I don't think we should have users that are both Google and Clever users but if we did, we might still want them to access their Google classrooms, right?

services/QuillLMS/client/app/bundles/Teacher/components/classrooms/active_classrooms.tsx
```diff
@@ -125,52 +168,85 @@ export default class ActiveClassrooms extends React.Component<ActiveClassroomsPr
     }
   }
 
-  initializePusherForGoogleClassrooms(id) {
-    if (process.env.RAILS_ENV === 'development') {
-      Pusher.logToConsole = true;
-    }
+  initializePusherForCleverClassrooms(user_id: String) {
```
emilia-friedberg: we should camelCase `user_id` since we're in JS-land.

services/QuillLMS/client/app/bundles/Teacher/components/classrooms/active_classrooms.tsx
```diff
@@ -125,52 +168,85 @@ export default class ActiveClassrooms extends React.Component<ActiveClassroomsPr
...
+    requestGet('/teachers/classrooms', (body) => this.setState({
+      classrooms: body.classrooms.filter(classroom => classroom.visible),
+      coteacherInvitations: body.coteacher_invitations
+    }))
   }
 
-  onSuccess(snackbarCopy) {
+  onSuccess = (snackbarCopy) => {
     this.getClassroomsAndCoteacherInvitations()
     this.getGoogleClassrooms()
```
emilia-friedberg, line 218: We probably also want to call `getCleverClassrooms` here now, right?

services/QuillLMS/client/app/bundles/Teacher/components/classrooms/active_classrooms.tsx
```diff
@@ -235,30 +312,44 @@ export default class ActiveClassrooms extends React.Component<ActiveClassroomsPr
...
-      />)
+
+      return (
+        <Classroom
+          archiveClass={() => this.openModal(archiveClassModal)}
+          changeGrade={() => this.openModal(changeGradeModal)}
+          classroom={classroom}
+          classrooms={ownActiveClassrooms}
+          clickClassroomHeader={this.clickClassroomHeader}
+          importCleverClassroomStudents={() => this.clickImportCleverClassroomStudents()}
```
emilia-friedberg: You can just do `importCleverClassroomStudents={this.clickImportCleverClassroomStudents}` since there are no arguments.

services/QuillLMS/client/app/bundles/Teacher/components/classrooms/active_classrooms.tsx
```diff
@@ -296,163 +391,301 @@ export default class ActiveClassrooms extends React.Component<ActiveClassroomsPr
...
+
+  renderImportCleverClassroomsModal() {
+    const { user } = this.props
+    const { cleverClassrooms, showModal } = this.state
+
+    if (showModal === importCleverClassroomsModal) {
+      return (
+        <ImportCleverClassroomsModal
+          classrooms={cleverClassrooms}
+          close={() => this.closeModal(() => this.setState({attemptedImportCleverClassrooms: false}))}
```
emilia-friedberg: We try to avoid binding in props in general (https://github.com/yannickcr/eslint-plugin-react/blob/master/docs/rules/jsx-no-bind.md), but there is an existing pattern of doing it in this file since it was written before we codified our JS style guide so I think it's okay to follow that in this instance. I think we should definitely define the argument you're passing to closeModal as its own function, though, so we aren't passing `two` arrow functions as a prop.

services/QuillLMS/client/app/bundles/Teacher/components/classrooms/classroom_student_section.tsx
```diff
@@ -223,9 +226,15 @@ export default class ClassroomStudentSection extends React.Component<ClassroomSt
     this.setState({ selectedStudentIds: [] })
   }
 
-  handleClickViewAsStudentButton = () => this.props.viewAsStudent()
+  handleClickViewAsStudentButton = (e) => {
+    const { viewAsStudent } = this.props
+    viewAsStudent(e)
```
emilia-friedberg: I think passing `e` here is going to create issues since it looks like `viewAsStudent` is only expecting an argument that is an individual student's id.

services/QuillLMS/client/app/bundles/Teacher/components/classrooms/clever_classrooms_empty_modal.tsx
```diff
@@ -0,0 +1,30 @@
+import * as React from 'react'
+const emptyClassSrc = `${process.env.CDN_URL}/images/illustrations/empty-class.svg`
+
+interface CleverClassroomsEmptyModalProps {
+  close: () => void;
+}
+
+export default class CleverClassroomsEmptyModal extends React.Component<CleverClassroomsEmptyModalProps, {}> {
```
emilia-friedberg, line 8: You don't need to change it, but this is a prime example of a component that would be really easy to write as a functional component, which is what we're trying to move to for all React code ultimately. Random example of a functional component here: https://github.com/empirical-org/Empirical-Core/blob/feat/bolding-word-diff/services/QuillLMS/client/app/bundles/Teacher/components/dashboard/welcome_modal.tsx#L14.

services/QuillLMS/client/app/bundles/Teacher/components/classrooms/import_clever_classroom_students_modal.tsx
```diff
@@ -0,0 +1,141 @@
...
+  constructor(props) {
+    super(props)
+
+    this.state = {
+      checkboxOne: false,
+      waiting: false
+    }
+
+    this.handleToggleCheckbox = this.handleToggleCheckbox.bind(this)
+    this.handleImportStudents = this.handleImportStudents.bind(this)
```
emilia-friedberg: I know you did a ton of work refactoring files like `active_classrooms`, and that this is based on an existing file, but since it's new code it does seem to me like we should probably avoid breaking linter rules as much as possible (like these binds here, and calling `this.props.onSuccess` below, etc).

services/QuillLMS/client/app/bundles/Teacher/components/classrooms/import_google_classroom_students_modal.tsx
```diff
@@ -64,35 +61,54 @@ export default class ImportGoogleClassroomStudentsModal extends React.Component<
...
-    return (<div className="checkboxes">
-      <div className="checkbox-row">
-        {this.renderCheckbox()}
-        <span>I understand that newly imported students have access to the activities that have already been assigned to the entire class.</span>
+    return (
+      <div className="checkboxes">
+        <div className="checkbox-row">
+          {this.renderCheckbox()}
+          <span>
+            I understand that newly imported students have access to the activities that have already been assigned
```
emilia-friedberg: I think this is where the `to` got removed (but shouldn't have been) that I called out in the test file.

services/QuillLMS/client/app/bundles/Teacher/components/classrooms/reauthorize_clever_modal.tsx
```diff
@@ -0,0 +1,54 @@
...
+interface ReauthorizeCleverModalProps {
+  cleverLink: string;
+  close: () => void;
+}
+
+export default class ReauthorizeCleverModal extends React.Component<ReauthorizeCleverModalProps, {}> {
+  constructor(props) {
+    super(props)
+    this.handleReauthorizeClick = this.handleReauthorizeClick.bind(this)
+  }
```
emilia-friedberg: Shouldn't use bind here - and in fact you don't need a constructor at all.

emilia-friedberg (changes requested):
Mostly reviewed from a frontend perspective - looks really good overall and I appreciate the refactors of existing code that you did here. This was interesting to look back on because it doesn't feel like all that long ago that I wrote this and yet the style of how we write React code has changed a lot! I have some minor changes requested, mostly stylistic and a few I'm worried will cause minor bugs. Thank you again for all your work on this!

services/QuillLMS/client/app/bundles/Teacher/components/classrooms/active_classrooms.tsx
brendanshean: The `getGoogleClassrooms` call is a pull from the third-party API that utilizes the users auth_credential.  Users can only have on auth_credential so I'm making an assumption here that if they have a clever_id, we don't want to be pulling anything from Google Classrooms since it will result in an error.  That being said, any google classrooms that exist on the Quill server will be accessible to teachers (they just won't be able to pull new information unless they relink their account with google).

services/QuillLMS/client/app/bundles/Teacher/components/classrooms/active_classrooms.tsx
brendanshean: Good idea.  I love guard clauses.

services/QuillLMS/client/app/bundles/Teacher/components/classrooms/import_clever_classroom_students_modal.tsx
brendanshean: Good catch on these.  I actually fixed a bunch of the binds and props destructuring in other existing files but missed ones in the files I added.  Go figure.  I will fix these.

services/QuillLMS/client/app/bundles/Teacher/components/classrooms/active_classrooms.tsx
brendanshean: Good question.  In the case of Google Classrooms, the getGoogleClassrooms call is basically reloading the cache in case we need that info later.  In the case of Clever, it's probably not what we want since this will effectively check for clever authorization more frequently than necessary, asking the user for clever reauthorization daily.  Instead, I'm only calling it when it's explicitly needed to prevent the reauthorization modal from being required.

services/QuillLMS/client/app/bundles/Teacher/components/classrooms/clever_classrooms_empty_modal.tsx
brendanshean: Normally, I would take you up on this refactor, but I would like to get this shipped.  I will make a note to do this in a future refactor.

services/QuillLMS/client/app/bundles/Teacher/components/classrooms/reauthorize_clever_modal.tsx
brendanshean: I thought I  need the constructor so I have access to the props?

services/QuillLMS/client/app/bundles/Teacher/components/classrooms/active_classrooms.tsx
brendanshean: Good idea!

services/QuillLMS/app/controllers/teachers/classroom_manager_controller.rb
```diff
@@ -22,9 +23,11 @@ def lesson_planner
 
   def assign
     session[GOOGLE_REDIRECT] = request.env['PATH_INFO']
+    session[CLEVER_REDIRECT] = request.env['PATH_INFO']
```
dandrabik, line 26: Not a major deal, but any reason to not make this generic between the two providers?, e.g. `session[PROVIDER_REDIRECT]` or whatever.

dandrabik (approved):
LGTM to me. Nice work on this. 🚢

services/QuillLMS/app/controllers/teachers/classroom_manager_controller.rb
brendanshean: It's a good idea, but I think I'll save it for a future refactor.

services/QuillLMS/client/app/bundles/Teacher/components/classrooms/reauthorize_clever_modal.tsx
emilia-friedberg: `constructor` always gets called under the hood, so all you're doing here once you remove the unnecessary bind is defining it as calling `super`, which is the same as not including it.

emilia-friedberg (approved):
Looks great! Thanks for making those changes.

PR #8640, opened 2022-01-03
Fix new year's bug with Vitally Workers

cissyxyu:
## WHAT
We encountered a bug when the year changed to 2022 with Vitally workers. This fixes it by accurately calculating the last school year using the `school_year_start` method.

## WHY
All the jobs are failing right now because it is trying to calculate previous year data for the year 2021 (which is a still-ongoing school year).

## HOW
Our code is using a quick-and-dirty way to calculate the last school year, subtracting one year from the current date. However this doesn't work in 2022 because the last school year varies based on whether the current date is before or after July (e.g. the last school year right now is 2020, because the last school year began on August 1, 2020, but in October 2022, the last school year will be 2021 because we've entered the 2022-2023 school year).

We already wrote the logic to calculate the date of the beginning of the school year in `School.school_year_start`, so I'm just calling this method with today's date minus one year to get the accurate year of the last school year.

### Screenshots

### Notion Card Links
https://www.notion.so/quill/Product-Board-30f97e4eb01246dbb8264253fb385073?p=3736cb90594e4b3fb3f114cab8382aad

PR Checklist | Your Answer
------------ | -------------

### DIFF
--- a/services/QuillLMS/app/services/vitally_integration/serialize_vitally_sales_account.rb
+++ b/services/QuillLMS/app/services/vitally_integration/serialize_vitally_sales_account.rb
@@ -57,7 +57,7 @@ class SerializeVitallySalesAccount
   end
 
   private def get_from_cache(key)
-    last_school_year = Date.today.year - 1
+    last_school_year = School.school_year_start(Date.today - 1.year).year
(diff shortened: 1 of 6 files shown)


anathomical (approved):

happythenewsad (approved):
Thanks for the explainer 👊

PR #8645, opened 2022-01-05
Rules index performance improvements

happythenewsad:
## WHAT, HOW
- removes the possibility of hitting `RulesController#index` without a `rule_type` param. This avoids an expensive, unnecessary query. See further details and decision tree in the notion ticket.
- audited all existing calls to `RulesController#index` and ensured that this change does not degrade the product
- adds a new endpoint for specifically requesting `universal` rules. This improves load time of the staff Universal Rules page. 
- I considered adding a partial index on `Rule where universal = true`, but found that the query optimizer used a table-level filter instead of the new index, so I removed it.

## WHY
Performance improvements 

### Screenshots

### Notion Card Links
[(Please provide links to any relevant Notion card(s) relevant to this PR.)](https://www.notion.so/quill/ea241f31a84c4b79ad56e83e99f7ee93?v=388007d5669c4347b4215e24770ebbcd&p=1ca5c09bac1e45e2a33f2ae7c0cc650e)

PR Checklist | Your Answer
------------ | -------------

### DIFF
--- a/services/QuillLMS/client/app/bundles/Staff/helpers/evidence/ruleHelpers.tsx
+++ b/services/QuillLMS/client/app/bundles/Staff/helpers/evidence/ruleHelpers.tsx
@@ -418,15 +418,15 @@ export async function handleSubmitRule({
 }
 
 export function getRulesUrl(activityId: string, promptId: string, ruleType: string) {
-  const url = `activities/${activityId}/rules`;
-  if(promptId && !ruleType) {
-    return `rules?prompt_id=${promptId}`
-  } else if(!promptId && ruleType) {
+  if (activityId) {
+    return `activities/${activityId}/rules`
+  } else if (!ruleType) {
+    throw new Error('A rule type must be specified.')
+  } else if(!promptId) {
     return `rules?rule_type=${ruleType}`
-  } else if(promptId && ruleType) {
+  } else {
     return `rules?prompt_id=${promptId}&rule_type=${ruleType}`
   }
-  return url;
 }
 
 export function getReturnLinkRuleType(ruleType) {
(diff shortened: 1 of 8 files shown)


services/QuillLMS/client/app/bundles/Staff/helpers/evidence/ruleHelpers.tsx
```diff
@@ -418,15 +418,15 @@ export async function handleSubmitRule({
...
 
 export function getRulesUrl(activityId: string, promptId: string, ruleType: string) {
-  const url = `activities/${activityId}/rules`;
-  if(promptId && !ruleType) {
-    return `rules?prompt_id=${promptId}`
-  } else if(!promptId && ruleType) {
+  if (activityId) {
+    return `activities/${activityId}/rules`
+  } else if (!ruleType) {
+    throw new Error('A rule type must be specified.')
```
happythenewsad, line 424: This code path is not currently hit; it's meant to inform future developers. I'm open to alternatives, like 
- returning a null string
- catching the error in the caller context.
- don't document anything here, and just let the controller throw an exception.

dandrabik: Hmm, it might be better to have the controller throw this error since our backend reporting is a little noisier. It's a staff endpoint, so it's not the biggest deal either way.

That might also allow you to turn this code branching into just passing along `prompt_id` and `rule_type` params to the url and letting the backend deal with it.

dandrabik (approved):
:+1: Small comment about when to throw an error / small suggestion. Nothing blocking, both take or leave. Thanks for fixing this up!

services/QuillLMS/client/app/bundles/Staff/helpers/evidence/ruleHelpers.tsx
happythenewsad: FYI The controller currently throws an error as well, via strong params. In light of this, I'll keep this slightly redundant front end check for now (my bias leans toward code clarity and intentionality).

PR #8651, opened 2022-01-05
Add readability field to JSON hash for UnitTemplate

cissyxyu:
## WHAT
For our next round of Activity Pack changes, we're going to need a `readability` field for `UnitTemplate` that shows the range of readability levels within an Activity Pack. 

E.g. When an activity pack has an activity of readability 4th-5th grade and another activity of 6th-7th grade, the activity pack's total readability is 4th-7th.

## WHY
This data point is useful for teachers when considering which activity packs to assign to a class, and it's also useful for Curriculum developers to see how readable an activity pack's activities are.

## HOW
Add this field to the serializer for `UnitTemplate`.

Add a method `readability` in the `UnitTemplate` model that parses the readability of all its activities and connects the lower and upper bound readabilities.

Note: In implementing this method, I chose to sort all of a pack's activities by their `raw_score_id`. This sorting only works because right now, all the `RawScore` records happen to be in ascending order of readability (e.g. the 1st record is the lowest difficulty, the last record is the highest difficulty). **This is not explicitly coded anywhere** and could change at some point in the future, breaking this code. 

Nevertheless, I chose to implement it this way because it made the code more readable. Because we store the readability level as strings only (e.g. "500-600"), there is no simple way to sort all of them except by string splitting and then converting them to integers (including special cases like "BR100"). I also thought the `RawScore` records are unlikely to change over time because the things they represent (grade levels K-12) do not change.

But I'm open to doing it the more complicated but failsafe way, if other people prefer!

### Screenshots

### Notion Card Links
https://www.notion.so/quill/Overhauls-to-Individual-Activity-Pack-view-page-20cf3f183848451f94db618c1a356122

PR Checklist | Your Answer
------------ | -------------

### DIFF
--- a/services/QuillLMS/app/models/unit_template.rb
+++ b/services/QuillLMS/app/models/unit_template.rb
@@ -55,7 +55,7 @@ class UnitTemplate < ApplicationRecord
   DIAGNOSTIC = 'Diagnostic'
 
   def readability
-    activities_with_raw_scores = activities.select {|act| act.raw_score.present? }.sort_by {|act| act.raw_score.order }
+    activities_with_raw_scores = activities.joins(:raw_score).where.not(raw_score: nil).order("raw_scores.order ASC")
     return nil if activities_with_raw_scores.empty?
 
     lowest_raw_score_activity = activities_with_raw_scores.first
--- a/services/QuillLMS/app/models/unit_template.rb
+++ b/services/QuillLMS/app/models/unit_template.rb
@@ -54,6 +54,21 @@ class UnitTemplate < ApplicationRecord
   INDEPENDENT_PRACTICE = 'Independent practice'
   DIAGNOSTIC = 'Diagnostic'
 
+  def readability
+    activities_with_raw_scores = activities.select {|act| act.raw_score.present? }
+    return nil if activities_with_raw_scores.empty?
+
+    lowest_raw_score_activity = activities_with_raw_scores.min_by(&:raw_score_id)
+    highest_raw_score_activity = activities_with_raw_scores.max_by(&:raw_score_id)
+
+    return nil if lowest_raw_score_activity&.raw_score&.nil? || highest_raw_score_activity&.raw_score&.nil?
+
+    lowest_readability_range = lowest_raw_score_activity.readability_grade_level
+    highest_readability_range = highest_raw_score_activity.readability_grade_level
+
+    "#{lowest_readability_range.split("-")[0]}-#{highest_readability_range.split("-")[1]}"
+  end
+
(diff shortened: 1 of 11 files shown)


brendanshean (commented):
Since the `RawScore` table is very small, I would make the ordering in that table explicit by adding a required integer attribute `order`;  relying on the primary key, like you mention, is asking for trouble later.

Also, if we're worried about people adding records `RawScore` incorrectly to the table in the future, you could add an [inclusion](https://guides.rubyonrails.org/active_record_validations.html#inclusion) that essentially locks the table unless items are explicitly added in the code to the validation.  I tend to prefer using these for small, closed classes that aren't expected to grow over time.

emilia-friedberg (approved):
I like Brendan's suggestion; approving once that's in place.

services/QuillLMS/db/migrate/20220106193721_add_order_to_raw_scores.rb
```diff
@@ -0,0 +1,9 @@
+# frozen_string_literal: true
+
+class AddOrderToRawScores < ActiveRecord::Migration[5.1]
+  def change
+    add_column :raw_scores, :order, :integer, null: true
+    RawScore.all.each {|rs| rs.update(order: rs.id)}
+    change_column_null :raw_scores, :order, false
```
brendanshean, line 7: Nice job with the `change_column_null` call!

services/QuillLMS/app/models/unit_template.rb
```diff
@@ -54,6 +54,19 @@ class UnitTemplate < ApplicationRecord
   INDEPENDENT_PRACTICE = 'Independent practice'
   DIAGNOSTIC = 'Diagnostic'
 
+  def readability
+    activities_with_raw_scores = activities.select {|act| act.raw_score.present? }.sort_by {|act| act.raw_score.order }
```
brendanshean: What about a database only call:

`activities.joins(:raw_score).where.not(raw_score: nil).order(raw_score: :order)`

services/QuillLMS/spec/serializers/cms/unit_template_serializer_spec.rb
```diff
@@ -0,0 +1,27 @@
+# frozen_string_literal: true
+
+require 'rails_helper'
+
+describe Cms::UnitTemplateSerializer do
```
brendanshean, line 5: Good idea on adding this spec.

brendanshean (approved):
One minor comment on querying that I think would be useful.  Other than that, looks good to me.

PR #8653, opened 2022-01-06
Backend prework for teachers choose clever classroom sync

brendanshean:
## WHAT
This commit represents a large scale refactor of existing functionality for  [clever_integration/sign_up/teacher](https://github.com/empirical-org/Empirical-Core/blob/develop/services/QuillLMS/app/services/clever_integration/sign_up/teacher.rb) in preparation for a frontend PR with new functionality.

## WHY
1) Add `CleverIntegration::LibraryTeacherIntegration`: This will save an authorization token that is used when accessing the LibraryClient in background jobs
2) move `CleverLibrary::Api::Client` to `CleverIntegration::LibraryClient`:  The new name here reflects the type of client used, in contrast to `DistrictClient`.
3) Move `CleverLibrary::Api::Oauth` to `CleverIntegration::OauthClient`: Get the client under `CleverIntegration` namespace.  This client is used to create oauth tokens for `DistrictClient`
4) Have `CleverIntegration::Importers::CleverDistrict` use `OauthClient` and `LibraryClient`
5) Add `CleverIntegration::DistrictTeacherIntegration`: This will save an authorization token that is used when accessing the `DistrictClient` in background jobs.  Also, for district integration, we associate teachers to schools and districts.
6) Add `CleverIntegration::Error` class: this class will be inherited by various Error classes within the `CleverIntegration` namespace to cut down on boilerplate .
7) Add `TeacherDataAdapter` class: adapt teacher data from clever api, specifically the `OmniAuth::AuthHash` object.
8) Add `CleverIntegration::Teacher{Importer, Updater, Creator}` classes:  Importer determines whether to use Creator or Updater on incoming teacher data
9) Add `CleverIntegration::TeacherClassroomsRetriever` and `CleverIntergration::TeacherClassroomsCache` classes:  retrieve teacher classroom data from Clever api and cache it for later use.
10) Add `CleverIntegration::TeacherImportedClassroomUpdater` class: import all clever classrooms for a teacher.  NB: The name of this class in this intermediate step is not quite correct but it will be when its commented out methods are utilized in the subsequent frontend PR.
11) Add `CleverIntegration::TeacherIntegration` class: this class orchestrates all of the importing pertaining to clever teachers.
12) Add `CleverIntegration::Library{Classroom,Student}DataAdapter` classes: these data adapters will parse and structure incoming data from `get_teacher_classrooms` and `get_classroom_students` calls to `LibraryClient` so that the data is in the same form as that corresponding calls to the `DistrictClient`.
13) Add `CleverIntegration::District{Classroom,Student}DataAdapter` classes: these data adapters will parse and structure incoming data from `get_teacher_classrooms` and `get_classroom_students` calls to `DistrictClient` so that the data is in the same form as that corresponding calls to the `LibraryClient`.
14) Add `CleverIntegration::ClientFetcher` classes: this fetches the appropriate clever api client (`DistrictClient`, `LibraryClient`) based on the type of `AuthCredential`
15) Add service objects and worker classes for classroom student importing
16) Add `CleverIntegration::TeacherClassroomsData` class:  this iterator facilitates importing of classrooms
17)  Utilize new TeacherIntegration with sub_main: replace sign_up/teacher.rb module with `CleverIntegration::TeacherIntegration`.
18) Odds and ends:  removing various clever integration files from the older design.
19) Add fix for duplicate clever classroom names

## HOW
The `TeacherIntegration` service object orchestrates most of what is going in this PR:

1.  Teacher logs in
2.  Save the teachers auth credential
3.  If district integration, associate classroom to school, and district
4.  Cache the classroom data
5.  update the classroom
6.  update the classroom students

The rest is of the code is creating lots of service objects along with much needed specs to keep track of the computations.

### Screenshots

### Notion Card Links

PR Checklist | Your Answer
------------ | -------------


services/QuillLMS/app/models/clever_integration/teacher_classrooms_cache.rb
```diff
@@ -0,0 +1,32 @@
+# frozen_string_literal: true
+
+module CleverIntegration
+  class TeacherClassroomsCache
```
dandrabik, line 4: Is there reason to use the raw `redis` actions vs. `Rails.cache.fetch/read/write`? If not, it would be nice to move towards that pattern. 

(Note, I used `redis` directly for [RedisFeed](https://github.com/empirical-org/Empirical-Core/blob/develop/services/QuillLMS/app/models/redis_feed.rb), because I was manipulating Redis arrays, but the functionality here seems in line with standard cache setting/expiration).

services/QuillLMS/app/services/clever_integration/classroom_creator.rb
```diff
@@ -28,12 +28,33 @@ def run
...
+    private def teacher
+      ::User.find(teacher_id)
+    end
+
+    private def valid_name
+      temp_name = data[:name]
+
+      loop do
+        return temp_name unless other_owned_classroom_names.include?(temp_name)
+        temp_name += '_1'
```
dandrabik: So if duplicates existed, this would `loop` would return:
```
'some_name'
'some_name_1'
'some_name_1_1'
'some_name_1_1_1'
```
Is that the intention, or should it add `_1`, `_2`

services/QuillLMS/app/services/clever_integration/classroom_creator.rb
```diff
@@ -28,12 +28,33 @@ def run
...
+    end
+
+    private def teacher
+      ::User.find(teacher_id)
+    end
+
+    private def valid_name
+      temp_name = data[:name]
+
+      loop do
```
dandrabik: Even if they are unlikely, I think we should avoid potential infinite loops, so let's put a cap on this (maybe 10). You could default to adding a hash to the classroom if you want this to always succeed.

services/QuillLMS/app/services/clever_integration/classroom_students_importer.rb
```diff
@@ -12,6 +12,7 @@ def initialize(classroom, students_data)
     def run
       return if students_data.nil? || students_data.empty?
 
+      import_students
```
dandrabik, line 15: Maybe you are doing this for clarity, but you can remove this line and the `import_students` method and the functionality would stay the same.

services/QuillLMS/app/services/clever_integration/classroom_updater.rb
```diff
@@ -34,16 +33,46 @@ def run
...
+    private def valid_name
+      temp_name = data[:name]
+
+      loop do
+        return temp_name unless other_owned_classroom_names.include?(temp_name)
+        temp_name += '_1'
+      end
+    end
+
+    private def visible
```
dandrabik: It might be clearer to just remove this method and pass `visible: true` into the `update` method above.

services/QuillLMS/app/services/clever_integration/classroom_updater.rb
```diff
@@ -34,16 +33,46 @@ def run
...
-        grade: grade
+        grade: grade,
+        visible: visible
       )
     end
+
+    private def valid_name
+      temp_name = data[:name]
+
+      loop do
```
dandrabik: Infinite loop, let's cap.

services/QuillLMS/app/services/clever_integration/client_fetcher.rb
```diff
@@ -0,0 +1,53 @@
...
+    end
+
+    private def auth_credential
+      user.auth_credential
+    end
+
+    private def client
+      raise NilAuthCredentialError if auth_credential.nil?
+
+      case provider
```
dandrabik, line 34: Take or leave (if you find this more confusing, stick with what you have), but you can convert this to a hash lookup for open/close extensibility in the future, e.g.

```ruby
PROVIDER_CLIENT_MAP = {
  AuthCredential::CLEVER_DISTRICT_PROVIDER => -> (token) { DistrictClient.new(token) }
  AuthCredential::CLEVER_LIBRARY_PROVIDER => -> (token) { LibraryClient.new(token) }
}

def run
  raise NilAuthCredentialError if auth_credential.nil?

  PROVIDER_CLIENT_MAP.fetch(provider).call(token) # will raise a KeyError if provider doesn't exist
end

File: services/QuillLMS/app/services/clever_integration/district_teacher_integration.rb [hunk 8]
```diff
@@ -0,0 +1,44 @@
+# frozen_string_literal: true
+
+module CleverIntegration
+  class DistrictTeacherIntegration < ApplicationService
+    class NilDistrictError < ::CleverIntegration::Error
```
dandrabik, line 5: Take or leave, if you always want the same message, you can do this without a custom subclass:
```ruby
class NilDistrictError < StandardError
  def message
    'No district was found'
  end
end

# or this if you just want a default
class NilDistrictError < StandardError
  def initialize(msg='No district was found')
    super
  end
end

File: services/QuillLMS/app/services/clever_integration/importers/clever_district.rb [hunk 9]
```diff
@@ -2,17 +2,19 @@
...
-    client = CleverLibrary::Api::Client.new(district_oauth_token)
+    district_oauth_token = district_tokens[0]["access_token"] if district_tokens
+    client = CleverIntegration::LibraryClient.new(district_oauth_token)
     clever_district_data = client.get_district(district_id: district_id)
-    if district_oauth_token.nil?
-      raise "district_oauth_token cannot be nil"
-    end
-    CleverIntegration::Creators::CleverDistrict.run(clever_id: district_id, name: clever_district_data["name"], token: district_oauth_token)
+
+    raise "district_oauth_token cannot be nil" if district_oauth_token.nil?
```
dandrabik: Take or leave, I don't think this is your code, but this `raise` seems like it can move up in the method (we're setting a var as `nil`, then raising if it's `nil`). 

```ruby
district_tokens = oclient.get_district_token(district_id: district_id) # you might consider making this method raise
raise "some error" unless district_tokens

district_oauth_token = district_tokens[0]["access_token"]
raise "some error" if district_oauth_token.nil?

File: services/QuillLMS/app/services/clever_integration/teacher_classrooms_retriever.rb [hunk 10]
```diff
@@ -0,0 +1,58 @@
...
+
+    private def client
+      ClientFetcher.run(teacher)
+    end
+
+    private def data
+      { classrooms: client.get_teacher_classrooms(teacher.clever_id) }
+    end
+
+    private def load_teacher
```
dandrabik, line 37: Take or leave, consider removing this method and putting the `raise` in `run`, since it reads clearly as a one-liner without a method abstraction.

services/QuillLMS/app/services/clever_integration/teacher_classrooms_retriever.rb
```diff
@@ -0,0 +1,58 @@
...
+      notify_pusher
+    rescue StandardError => e
+      NewRelic::Agent.notice_error(e, user_id: teacher.id)
+    end
+
+    private def cache_classrooms_data
+      TeacherClassroomsCache.set(teacher_id, data.to_json)
+    end
+
+    private def client
```
dandrabik, line 29: In other places you memoize this method, should that happen here?

services/QuillLMS/app/services/clever_integration/teacher_updater.rb
```diff
@@ -0,0 +1,28 @@
...
+
+    ACCOUNT_TYPE = ::User::CLEVER_ACCOUNT
+    ROLE = ::User::TEACHER
+
+    def initialize(teacher, data)
+      @teacher = teacher
+      @data = data
+    end
+
+    def run
```
dandrabik, line 15: Could this method just be `teacher.update!(teacher_attrs)` (won't that return the teacher)

services/QuillLMS/app/services/clever_integration/teacher_creator.rb
```diff
@@ -0,0 +1,31 @@
...
+
+    attr_reader :data, :name
+
+    def initialize(data)
+      @data = data
+      @name = data[:name]
+    end
+
+    def run
+      teacher
```
dandrabik, line 16: Take or leave, but looking at all of these `run` methods, there are times that you put the operation in `run` and times you've abstracted into another method. 

In my opinion, this is one in which having the operation in `run` would be easier to scan/understand quickly, e.g. 
```ruby
def run
  ::User.create!(teacher_attrs)
end
```
Slightly easier to scan as "this Service Object creates a teacher" vs. "this Service Object calls a method called `teacher`, that method creates a teacher". Might be worth revisiting some of the `run` methods to determine if the abstracted methods make it easier to understand (note, this purely opinion based, so take or leave).

dandrabik (approved):
Overall this looks good, thanks for putting this together. I made a bunch of comments, but I believe that most of them are small / cosmetic, so no need for a re-review (most are take or leave).

services/QuillLMS/app/models/clever_integration/teacher_classrooms_cache.rb
brendanshean: Our codebase uses raw `$redis` throughout, so that was my initial choice, but now that you mentioned it, I agree.  After updating to `Rails.cache` it looks much [cleaner](https://github.com/empirical-org/Empirical-Core/pull/8653/commits/b25f5ec0bbceeee7beb0377affda073237d12057)

services/QuillLMS/app/services/clever_integration/classroom_creator.rb
brendanshean: That was the intention, but I agree that `_1`, `_2` makes more sense.  I've extracted out a new [method](https://github.com/empirical-org/Empirical-Core/pull/8653/commits/142911eb16a5df27a6685e322a912f53046f8c96) to handle this since `{Google,Clever}Integration::Classroom{Creator,Updater}` were all using this same pattern.

services/QuillLMS/app/services/clever_integration/classroom_students_importer.rb
brendanshean: I'm including that more for clarity.

services/QuillLMS/app/services/clever_integration/classroom_updater.rb
brendanshean: I was debating about how to flag this hard-coded value on object construction.

services/QuillLMS/app/services/clever_integration/importers/clever_district.rb
brendanshean: I agree.  Good point.

services/QuillLMS/app/services/clever_integration/teacher_classrooms_retriever.rb
brendanshean: Can you point me to the other places I memoize this?  I'm not finding them.

services/QuillLMS/app/services/clever_integration/teacher_updater.rb
brendanshean: This is a rails gotcha.  `update!` calls `save!` which either raises an exception or returns true.

brendanshean:
@dandrabik A couple of the review commits I've added (namely, DuplicateNameResolver and Cache refactoring) seem more than cosmetic so re-requesting.

services/QuillLMS/app/models/clever_integration/teacher_classrooms_cache.rb
dandrabik: Yeah, we definitely use `$redis` in other places, this was more a note that I'd like us to move away from that (no way for you to know that).

services/QuillLMS/app/services/clever_integration/teacher_classrooms_students_importer.rb
```diff
@@ -0,0 +1,33 @@
...
+    end
+
+    private def classrooms
+      ::Classroom
+        .where(id: classroom_ids)
+        .where.not(clever_id: nil)
+    end
+
+    private def client
+      @client ||= ClientFetcher.run(teacher)
```
dandrabik, line 23: Memoized here.

services/QuillLMS/app/services/duplicate_name_resolver.rb
```diff
@@ -0,0 +1,32 @@
+# frozen_string_literal: true
+
+class DuplicateNameResolver < ApplicationService
```
dandrabik, line 3: Nice Abstraction!

dandrabik (approved):
This looks good to me. Thanks for putting this together.

One note based on our discussion, if you are viewing the PR conversation view, you might have missed some of my comments from my initial review(they aren't resolved, but they don't appear on the conversation view for some reason). So I'd look at the files link: https://github.com/empirical-org/Empirical-Core/pull/8653/files (just search for my name).

The remaining comments are take or leave so they aren't blockers and you can ship without any other changes.

PR #8660, opened 2022-01-07
lms/add-subscription-refresh-task

anathomical:
## WHAT
Add a rake task that takes a list of user ids to refresh subscriptions by checking their schools to see if there are newer subscriptions to use
## WHY
Somehow a group of around 250 users don't have their subscriptions updated from the current subscription from the schools that they're assigned to.  We want to trigger the code that updates their subscriptions individually.
## HOW
Take a pipe of a CSV file and parse it to figure out which users need to be refreshed, then refresh each one.

### Notion Card Links
](https://www.notion.so/quill/Some-users-not-getting-premium-access-from-the-school-subscription-702a8b272c0146c2b3d8872bcb48b52c)

PR Checklist | Your Answer
------------ | -------------
Have you added and/or updated tests? |  No tests on tasks

### DIFF
--- a/services/QuillLMS/lib/tasks/users.rake
+++ b/services/QuillLMS/lib/tasks/users.rake
@@ -13,4 +13,22 @@ namespace :users do
       progress_bar.increment
     end
   end
+
+  task refresh_school_subscriptions: :environment do
+    pipe_data = STDIN.read unless STDIN.tty?
+
+    unless pipe_data
+      puts 'No data detected on STDIN.  You must pass data to the task for it to run.  Example:'
+      puts '  rake users:refresh_school_subscriptions < path/to/local/file.csv'
+      puts ''
+      puts 'If you are piping data into Heroku, you need to include the --no-tty flag:'
+      puts '  heroku run rake schools:update_clever_ids -a empirical-grammar --no-tty < path/to/local/file.csv'
+      exit 1
+    end
+
+    CSV.parse(pipe_data, headers: true) do |row|
+      user = User.find(row['user_id'])
+      user.updated_school(user.school.id)
+    end
+  end
 end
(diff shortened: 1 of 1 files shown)


services/QuillLMS/lib/tasks/users.rake
```diff
@@ -13,4 +13,22 @@ namespace :users do
...
+
+  task refresh_school_subscriptions: :environment do
+    pipe_data = STDIN.read unless STDIN.tty?
+
+    unless pipe_data
+      puts 'No data detected on STDIN.  You must pass data to the task for it to run.  Example:'
+      puts '  rake users:refresh_school_subscriptions < path/to/local/file.csv'
+      puts ''
+      puts 'If you are piping data into Heroku, you need to include the --no-tty flag:'
+      puts '  heroku run rake schools:update_clever_ids -a empirical-grammar --no-tty < path/to/local/file.csv'
```
dandrabik, line 25: I think this comment is a copy/paste error.

dandrabik (approved):
This looks fine to me. Given this task is tiny, you might just consider running this in the Heroku rails console without deploying code.

happythenewsad (approved):
I'm fast-track approving this given your comment, but I recommend:

1. Wrapping the main loop in a db transaction
2. Using find_by_id so that if there's one missing user it doesn't kill the whole rake task / transaction.

services/QuillLMS/lib/tasks/users.rake
anathomical: Good catch.  I actually explicitly remember changing this text, but maybe I re-pasted it?  Will fix.

anathomical:
> I'm fast-track approving this given your comment, but I recommend:
> 
> 1. Wrapping the main loop in a db transaction
> 2. Using find_by_id so that if there's one missing user it doesn't kill the whole rake task / transaction.

1. Not sure if this changes your thought, but the command is idempotent, so having it all succeed or all fail doesn't matter since re-running a partially completed task won't cause any unexpected errors.
2. Probably not a bad idea, but then we need a good way to report back which users didn't get updated so further work can be done with them.  Probably not hard to do, but do we have a pattern for this kind of thing in any of our existing tasks?

PR #8664, opened 2022-01-11
Delete archived classrooms

happythenewsad:
## WHAT
Safely delete Classrooms and join table objects. 

## WHY
Teacher requested this. 

## HOW
override `Classroom.destroy`
### Screenshots

### Notion Card Links
https://www.notion.so/quill/Teacher-would-like-archived-classes-deleted-8a987659e2554212b0350f25ee7cbce6

PR Checklist | Your Answer
------------ | -------------

### DIFF
--- a/services/QuillLMS/app/controllers/teachers/classrooms_controller.rb
+++ b/services/QuillLMS/app/controllers/teachers/classrooms_controller.rb
@@ -80,15 +80,6 @@ class Teachers::ClassroomsController < ApplicationController
     end
   end
 
-  def destroy
-    authorize_owner! { params[:id] }
-    Classroom.find(params[:id]).destroy
-    # we need a performed? check here to avoid a double render, since
-    # authorize_owner can trigger a render, which is an anti-pattern
-    # more info: https://medium.com/cedarcode/abstractcontroller-doublerendererror-fix-d18881b80476
-    redirect_to teachers_classrooms_path unless performed?
-  end
-
   def bulk_archive
     Classroom.where(id: params[:ids]).each do |classroom|
       next if classroom.owner != current_user
--- a/services/QuillLMS/app/models/classroom.rb
+++ b/services/QuillLMS/app/models/classroom.rb
@@ -30,7 +30,7 @@ class Classroom < ApplicationRecord
   include CheckboxCallback
 
   GRADES = %w(1 2 3 4 5 6 7 8 9 10 11 12 University)
-  UNIVERSITY= "University"
+  UNIVERSITY = "University"
   GRADE_INTEGERS = {Kindergarten: 0, University: 13, PostGraduate: 14}
 
   validates_uniqueness_of :code
@@ -41,13 +41,13 @@ class Classroom < ApplicationRecord
   after_commit :hide_appropriate_classroom_units
   after_commit :trigger_analytics_for_classroom_creation, on: :create
 
-  has_many :classroom_units
+  has_many :classroom_units, dependent: :destroy
   has_many :units, through: :classroom_units
   has_many :unit_activities, through: :units
   has_many :activities, through: :unit_activities
   has_many :activity_sessions, through: :classroom_units
   has_many :standard_levels, through: :activities
-  has_many :coteacher_classroom_invitations
+  has_many :coteacher_classroom_invitations, dependent: :destroy
 
   has_many :students_classrooms, foreign_key: 'classroom_id', dependent: :destroy, class_name: "StudentsClassrooms"
   has_many :students, through: :students_classrooms, source: :student, inverse_of: :classrooms, class_name: "User"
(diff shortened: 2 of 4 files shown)


services/QuillLMS/app/controllers/teachers/classrooms_controller.rb
```diff
@@ -80,15 +80,6 @@ def update
     end
   end
 
-  def destroy
```
happythenewsad, line 83: This wasn't operational and we weren't using it, so I'm deleting it.

services/QuillLMS/app/models/classroom.rb
```diff
@@ -59,6 +59,12 @@ class Classroom < ApplicationRecord
 
   accepts_nested_attributes_for :classrooms_teachers
 
+  def destroy
```
happythenewsad, line 62: Delete the join records, but don't delete units, activities, and other records that could be owned by other entities.

services/QuillLMS/app/models/classroom.rb
```diff
@@ -59,6 +59,12 @@ class Classroom < ApplicationRecord
 
   accepts_nested_attributes_for :classrooms_teachers
 
+  def destroy 
+    classroom_units.each(&:destroy)
+    ClassroomsTeacher.where(classroom_id: self.id).destroy_all
```
happythenewsad: We need a query here, since this is a foreign key relationship (i.e. not accessible via `record.classrooms_teachers`)

anathomical:
Would it make more sense to just modify the various 'belongs_to' declarations with 'dependent: destroy'?  Rails magic really should be able to take care of this forums, I think.

anathomical (approved):

services/QuillLMS/app/controllers/teachers/classrooms_controller.rb
brendanshean: I would remove the corresponding route then as well: https://github.com/empirical-org/Empirical-Core/blob/develop/services/QuillLMS/config/routes.rb#L290

services/QuillLMS/spec/models/classroom_spec.rb
```diff
@@ -51,6 +51,36 @@
...
+        classroom_to_destroy.destroy 
+      end
+      .to change { Classroom.count }.by(-1)
+      .and change { Unit.count }.by(0)
+      .and change { ClassroomUnit.count }.by(-1)
+      .and change { User.count }.by(0)
+      .and change { StudentsClassrooms.count }.by(-1)
+      .and change { Activity.count }.by(0) 
+      .and change { CoteacherClassroomInvitation.count }.by(-1)
+      expect(ClassroomsTeacher.where(classroom_id: classroom_to_destroy.id).count).to eq 0
```
brendanshean, line 80: The ClassroomsTeacher association lacks a "dependent: :destroy" within Classroom.rb.  How is this getting destroyed?

services/QuillLMS/app/models/classroom.rb
```diff
@@ -59,6 +59,10 @@ class Classroom < ApplicationRecord
 
   accepts_nested_attributes_for :classrooms_teachers
 
+  def destroy 
+    ClassroomsTeacher.where(classroom_id: id).destroy_all
```
emilia-friedberg, line 66: Couldn't we just add a `has_many` relationship to `classrooms_teachers` and then do `dependent: :destroy` on it?

services/QuillLMS/spec/models/classroom_spec.rb
happythenewsad: By this line: `ClassroomsTeacher.where(classroom_id: id).destroy_all`

services/QuillLMS/app/models/classroom.rb
happythenewsad: `dependent: :destroy` is a no-op when used in conjunction with the `foreign_key` designator, which requires that I explicitly do this:

`ClassroomsTeacher.where(classroom_id: id).destroy_all`

You can verify this by removing the above line, and re-adding `dependent: :destroy` to the relationship. The join table records will not be deleted (it's a silent 'failure', but the specs will complain).

brendanshean: So how are students classrooms getting destroyed?

brendanshean: Maybe, the class_name: "StudentsClassrooms" is giving the dependent: :destroy macro the information it needs to do the cascading deleting?

happythenewsad: I tried that already - unfortunately it does not affect the issue.

brendanshean: That still begs the question, why doesn't `dependent: :destroy` work for classrooms_teachers but it does for students_classrooms

happythenewsad: Yes, it's odd. But I'd rather ship this legitimate solution than spend potentially a lot of time on investigating a different syntax.

brendanshean: I think you should add a comment / todo or something indicating the weirdness so that future developers don't go down this rabbit hole again with the `dependent: :destroy`

happythenewsad: Good idea - added.

PR #8667, opened 2022-01-11
lms/add-script-to-set-activitities-private

anathomical:
## WHAT
Add a rake task that can set activities as "private" based on names in a CSV
## WHY
Curriculum has a list of 120 activities that they want to mass set to "private"
## HOW
Just take a pipe of a CSV file and parse each row to find an Activity by name and set the flag to `private`

### Notion Card Links
https://www.notion.so/quill/Change-these-activities-from-production-to-private-1966233fd36d47b8bf113191a8c8a96f

PR Checklist | Your Answer
------------ | -------------
Have you added and/or updated tests? | No tests on rake tasks

### DIFF
--- a/services/QuillLMS/lib/tasks/activities.rake
+++ b/services/QuillLMS/lib/tasks/activities.rake
@@ -15,6 +15,27 @@ namespace :activities do
       end
       activity.save!
     end
+  end
 
+  desc 'Take pipe of CSV with activity names to be set to "private"'
+  task set_to_private_by_name: :environment do
+    pipe_data = STDIN.read unless STDIN.tty?
+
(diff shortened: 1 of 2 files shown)


happythenewsad (changes requested):
lets use `save!` instead of `save` so we know that persistence took place. Otherwise, looks good

anathomical:
> lets use `save!` instead of `save` so we know that persistence took place. Otherwise, looks good

@happythenewsad Good call.  Will update.

emilia-friedberg (approved):

cissyxyu (approved):

happythenewsad (approved):

PR #8671, opened 2022-01-12
Caching Pattern.

dandrabik:
## WHAT
Further work on this original PR #8584. Add a reusable way to cache reports. Note, I'm planning on deploying all of the `touch` items first, I'll break that off into a separate PR>
## WHY
ActivitySessions change very often while a class is in session, but basically never change once a class/unit/activity are finished. This is an opportunity for caching.
## HOW
I moved the caching inside the a `User` mixin and integrated it with `Rails.cache.fetch` in an attempt to make it easier / more foolproof to use. 

This has the advantages of:
1. Ensures `nil` cache objects aren't accidentally shared between users since it allows us to default to the `user` object as a `cache_key` (found this edge case in testing)
2. Makes the syntax a little shorter (you just pass in a block, rather that create a key and use `Rails.cache.fetch`).

Sample controller usage:

```ruby
response = user.classroom_unit_cache(classroom_unit, key: 'controller.method') do
  # some expensive action to cache that only needs to update
  # when a classroom_unit or its children change
end
render json: response
```

PR Checklist | Your Answer
------------ | -------------


services/QuillLMS/spec/models/concerns/user_cacheable_spec.rb
```diff
@@ -0,0 +1,244 @@
...
+    let(:classroom) {create(:classroom) }
+
+    it 'should call model_cache' do
+      expect(teacher).to receive(:model_cache).with(classroom, key: 'test.key', groups: {page: 1})
+
+      teacher.classroom_cache(classroom, key: 'test.key', groups: {page: 1})
+    end
+
+    it 'should yield to model_cache' do
+      expect {|b| teacher.classroom_cache(classroom, key: 'test.key', groups: {page: 1}, &b) }.to yield_control.exactly(1).times
```
brendanshean, line 89: I haven't seen `yield_control` before.  Cool!

services/QuillLMS/app/serializers/lesson_planner/unit_serializer.rb
```diff
@@ -73,7 +73,7 @@ class LessonPlanner::UnitSerializer < ActiveModel::Serializer
   end
 
   private def dueDates
-    object.reload.unit_activities.uniq(&:activity).each_with_object({}) do |unit_activity, acc|
```
brendanshean: Was the reload here messing up the touches?

services/QuillLMS/spec/models/concerns/user_cacheable_spec.rb
```diff
@@ -0,0 +1,244 @@
...
+    end
+
+    it 'should call model_cache with first matching unit if there is a unit_id' do
+      expect(teacher).to receive(:model_cache).with(classroom_unit1, key: 'test.key', groups: {page: 1})
+
+      teacher.classroom_unit_by_ids_cache(classroom_id: classroom.id, unit_id: unit_activity1.unit_id, activity_id: activity.id, key: 'test.key', groups: {page: 1})
+    end
+
+    it 'should yield to model_cache' do
+      expect {|b| teacher.classroom_unit_by_ids_cache(classroom_id: classroom.id, unit_id: unit_activity1.unit_id, activity_id: activity.id, key: 'test.key', groups: {page: 1}, &b) }.to yield_control.exactly(1).times
```
brendanshean, line 159: This line is a little long.  I think breaking it up aids in scanning:

```
  expect { 
    teacher.classroom_unit_by_ids_cache(
      classroom_id: classroom.id, 
      unit_id: unit_activity1.unit_id, 
      activity_id: activity.id, 
      key: 'test.key', 
      groups: {page: 1}, 
      &b
    ) 
  }.to yield_control.exactly(1).times
```

brendanshean (approved):
This looks great.  I really like the utilization of `Rails.cache.fetch` with the block and the subsequent testing of it with`{ i+= 1 }` approach very clever.  That testing code is very thorough and will help immensely with future caching improvements.

One general comment that seems optional is the strings used for keys:
'classroom_manager.teacher_dashboard_metrics'
'classroom_manager.scores'
'diagnostic_reports.lesson_recommendations_for_classroom'

Would it make sense to constantize these?

emilia-friedberg (approved):
LGTM!

services/QuillLMS/app/serializers/lesson_planner/unit_serializer.rb
dandrabik: This `reload` was causing an ordering problem. This reload causes this (it was just surfaced by adding the touches):
- Old object used by: "methods called before `dueDates` is called"
- New object used by: "`dueDates` and any methods called after it OR all methods if you call `as_json` a second time"

As an aside, I couldn't find this serializer used anywhere, but there's the gem has some implicit usage by path/file/model name, so I'm not 100% sure, so I haven't removed it yet.

services/QuillLMS/app/models/concerns/user_cacheable.rb
```diff
@@ -0,0 +1,77 @@
...
+    [key, *group_array, object || self]
+  end
+
+  private def model_cache(object, key:, groups:)
+    Rails.cache.fetch(model_cache_key(object, key: key, groups: groups)) do
+      yield
+    end
+  end
+
+  private def last_updated_classroom
```
happythenewsad, line 53: Have you considered caching this method itself?

anathomical (approved):

PR #8672, opened 2022-01-12
LMS: Add touches for use by caching

dandrabik:
## WHAT
Pre-requisite deploy for #8671 to add model `touch`es to keep `updated_at` dates up to date and usable by caching.
## WHY
I want to put these changes out separately, before the caching to monitor and make sure they don't cause issue.
## HOW
Adding lots of touches to associations.

PR Checklist | Your Answer
------------ | -------------
Have you added and/or updated tests? |  yes, small changes

### DIFF
--- a/services/QuillLMS/app/models/classroom_unit.rb
+++ b/services/QuillLMS/app/models/classroom_unit.rb
@@ -28,7 +28,7 @@ class ClassroomUnit < ApplicationRecord
   include ::NewRelic::Agent
   include AtomicArrays
 
-  belongs_to :unit #, touch: true
+  belongs_to :unit
   belongs_to :classroom, touch: true
   has_many :activity_sessions
   has_many :unit_activities, through: :unit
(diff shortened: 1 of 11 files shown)


happythenewsad (approved):
Minor request - can you squash, or otherwise modify #8671 after merging this PR, so that reviewers don't have to review this changeset twice?

services/QuillLMS/app/models/classroom_unit.rb
```diff
@@ -28,7 +28,7 @@ class ClassroomUnit < ApplicationRecord
   include ::NewRelic::Agent
   include AtomicArrays
 
-  belongs_to :unit #, touch: true
+  belongs_to :unit
```
emilia-friedberg: Why don't we want a touch here?

brendanshean (approved):
I guess i did these in the wrong order.  And now Peter's comment makes more sense regarding the changeset.

services/QuillLMS/app/models/classroom_unit.rb
dandrabik: So the `touch` goes in the other direction: if a `unit` changes, each of its `classroom_unit`s is touched. I believe that's the direction we want (let me know if that seem incorrect). 

(Note, we'd have to only pick one direction, if we add a `touch` here, it would create an infinite loop of touches back and forth).

PR #8673, opened 2022-01-12
Evidence/side panel updates 1

eadams17:
## WHAT
first series of updates to the Evidence side panel: update the step counter display and add counter to the header

## WHY
these are part of a series of changes that were implementing to improve the UX of Evidence

## HOW
updated `StepOverview` and `Header` components to show these update step counter changes

### Screenshots
Screen Shot 2022-01-12 at 3 49 54 PM
Screen Shot 2022-01-12 at 3 50 08 PM

### Notion Card Links
https://www.notion.so/quill/Update-the-side-panel-95de1b95e3a2488390275254d80b9bf2

PR Checklist | Your Answer
------------ | -------------


cissyxyu (approved):

services/QuillLMS/client/app/bundles/Evidence/components/studentView/container.tsx
```diff
@@ -94,6 +94,14 @@ export const StudentViewContainer = ({ dispatch, session, isTurk, location, acti
     4: 0
   })
 
+  React.useEffect(() => {
+    dispatch(setActiveStepForSession(activeStep));
+  }, [activeStep])
```
emilia-friedberg: We should have a single source of truth here - so rather than updating the local `activeStep` and then calling `setActiveStepForSession` when it changes, we should just call `setActiveStepForSession` every time we'd update it locally and use the one passed down from the reducer as our `activeStep` in this file.

services/QuillLMS/client/app/bundles/Evidence/components/studentView/container.tsx
```diff
@@ -94,6 +94,14 @@ export const StudentViewContainer = ({ dispatch, session, isTurk, location, acti
...
+
+  React.useEffect(() => {
+    dispatch(setExplanationSlidesCompletedForSession(explanationSlidesCompleted));
+  }, [explanationSlidesCompleted])
```
emilia-friedberg: Same thing here - we shouldn't track the same value in two places.

services/QuillLMS/client/app/bundles/Evidence/components/studentView/stepOverview.tsx
```diff
@@ -1,43 +1,63 @@
...
+  {
+    html: <p>Expand a sentence using <u>but</u> to provide an opposing idea.</p>,
+    stepNumber: BUT_PASSAGE_STEP
   },
   {
-    text: 'Write three sentences using evidence'
+    html: <p>Expand a sentence using <u>so</u> to provide a result.</p>,
+    stepNumber: SO_PASSAGE_STEP
   }
 ]
```
emilia-friedberg: If you just define these individually here (so like:

```const readPassageStep = {
    html: <p>Read a text and highlight sentences.</p>,
    stepNumber: READ_PASSAGE_STEP
}```

You can just call those individually when you pass them as the `step` prop to the `Step` component throughout the file and then you don't need to mess with the stepNumber - 1 thing. It doesn't look like you actually need them as an array at all.

File: services/QuillLMS/client/app/bundles/Shared/utils/constants.ts [hunk 4]
```diff
@@ -2,3 +2,7 @@ export const DEFAULT_HIGHLIGHT_PROMPT = "As you read, highlight two sentences "
 export const BECAUSE = 'because';
 export const BUT = 'but';
 export const SO = 'so';
+export const READ_PASSAGE_STEP = 1
+export const BECAUSE_PASSAGE_STEP = 2
+export const BUT_PASSAGE_STEP = 3
+export const SO_PASSAGE_STEP = 4
```
emilia-friedberg: I might call these constants `READ_PASSAGE_STEP_NUMBER` etc to be fully descriptive.

emilia-friedberg (changes requested):
Small refactors suggested but looks great overall!

emilia-friedberg (approved):
Looks great! Thanks for making those changes.

PR #8675, opened 2022-01-13
Correct school year start to July 1

cissyxyu:
## WHAT
When I was correcting a new year's bug last week, I noticed that we have the wrong month for the first month of the school year. I'm correcting it to July.

## WHY
July 1 was the month we agreed to make the "beginning" of the school year for data recording purposes. This is a holdover from when we used to have it as August, it will slightly skew the data we're rporting.

## HOW
Just correct the month.

### Screenshots

### Notion Card Links

PR Checklist | Your Answer
------------ | -------------

### DIFF
--- a/services/QuillLMS/app/models/school.rb
+++ b/services/QuillLMS/app/models/school.rb
@@ -129,7 +129,7 @@ class School < ApplicationRecord
   end
 
   def self.school_year_start(time)
-    time.month >= 8 ? time.beginning_of_year + 7.months : time.beginning_of_year - 5.months
+    time.month >= 7 ? time.beginning_of_year + 6.months : time.beginning_of_year - 6.months
   end
 
   private def generate_leap_csv_row(student, teacher, classroom, activity_session)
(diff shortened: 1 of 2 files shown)


brendanshean (approved):

services/QuillLMS/app/models/school.rb
```diff
@@ -129,7 +129,7 @@ def students
   end
 
   def self.school_year_start(time)
-    time.month >= 8 ? time.beginning_of_year + 7.months : time.beginning_of_year - 5.months
+    time.month >= 7 ? time.beginning_of_year + 6.months : time.beginning_of_year - 6.months
```
happythenewsad: Good catch, but let's use a named variable rather than the magic number `7` here, so that future developers better know your intent. While you're at it, how about replacing `6` with `HALF_A_YEAR = 6`?

happythenewsad (changes requested):

happythenewsad (approved):

PR #8678, opened 2022-01-14
Fix bug with teachers changing from google to clever

brendanshean:
## WHAT
Fix a bug involving teachers re-linking their accounts from google back to clever.

## WHY
We want to allow teachers to re-link with clever

## HOW
Check to see if the user exists by email if they do not already exist by clever_id.

### Screenshots

### Notion Card Links
](https://www.notion.so/quill/Teachers-not-able-to-re-link-their-accounts-to-Clever-6a1e78ae783e4d5fb64f5b82846bbaa5)

PR Checklist | Your Answer
------------ | -------------

### DIFF
--- a/services/QuillLMS/app/services/clever_integration/teacher_updater.rb
+++ b/services/QuillLMS/app/services/clever_integration/teacher_updater.rb
@@ -18,7 +18,7 @@ module CleverIntegration
     end
 
     private def teacher_attrs
-      data.merge(account_type: ACCOUNT_TYPE, role: ROLE, google_id: nil)
+      data.merge(account_type: ACCOUNT_TYPE, role: ROLE)
     end
 
     private def update
--- a/services/QuillLMS/app/services/clever_integration/teacher_updater.rb
+++ b/services/QuillLMS/app/services/clever_integration/teacher_updater.rb
@@ -18,7 +18,7 @@ module CleverIntegration
     end
 
     private def teacher_attrs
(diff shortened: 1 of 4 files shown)


services/QuillLMS/app/services/clever_integration/teacher_updater.rb
```diff
@@ -18,7 +18,7 @@ def run
     end
 
     private def teacher_attrs
-      data.merge(account_type: ACCOUNT_TYPE, role: ROLE)
+      data.merge(account_type: ACCOUNT_TYPE, role: ROLE, google_id: nil)
```
dandrabik: So just so I understand, is this to ensure that a user only has one provider id (google or clever)? 

Would this change prevent them from logging in with google?

Given this is an edge case, probably ok to ship this as, since the most recent provider is more important for syncing.

dandrabik (approved):
I think this looks good. I have a question that I'd want to understand more, but given this is preventing users from syncing, it's likely ok to ship as is.

services/QuillLMS/app/services/clever_integration/teacher_updater.rb
brendanshean: I guess the unlinking would be handled elsewhere.  I'll remove it.

brendanshean: Setting the `google_id: nil` would prevent SSO via google as well as preventing any importing from google classroom.

PR #8679, opened 2022-01-14
Evidence/fuzzy plagiarism match

anathomical:
## WHAT
Add code to do a second "fuzzy match" plagiarism pass if the basic plagiarism algorithm comes back as "optimal".
## WHY
In cases where users make typos during a plagiarized sequence, we want to give them plagiarism feedback rather than giving them other feedback that, once implemented, gives them plagiarized feedback late in the cycle. 
## HOW
Sub-class the `PlagiarismFeedback` lib class, and refactor it slightly so that a minimum sub-class can do Levenshtein-distance matching.

### Notion Card Links
](https://www.notion.so/quill/Identify-plagiarized-responses-that-contain-misspellings-Fuzzy-Match-Plagiarism-99fa9c49b8df45cabe2a4d2a62a65c12)

PR Checklist | Your Answer
------------ | -------------


cissyxyu (approved):

services/QuillLMS/engines/evidence/lib/evidence/evidence/fuzzy_plagiarism_check.rb
```diff
@@ -0,0 +1,20 @@
...
+  class FuzzyPlagiarismCheck < PlagiarismCheck
+
+    FUZZY_CHARACTER_THRESHOLD = 5
+
+    private def minimum_overlap?
+      !clean_passage.empty? && [clean_entry.split.size, clean_passage.split.size].min >= MATCH_MINIMUM
+    end
+
+    private def identify_matched_slices(entry_slices, passage_slices)
+      entry_slices.select do |k,v|
```
dandrabik: In ruby, if you don't use a value, it's common to name it `_`, so this would be `|_,v|`. Also, let's give `v` a better name (style guide, avoid single letter variables outside of inline blocks).

happythenewsad (approved):
I think it would be cleaner to bake the fuzzy check into the existing PlagiarismCheck, because:

- we could slim down the controller by moving the double loops into the check file
- we probably want to call the fuzzy check whenever we call any plagiarism check, so they don't need to be in a parent<>child hierarchy 

but if you consider that out of scope for this ticket, I'm OK with that.

services/QuillLMS/engines/evidence/lib/evidence/evidence/plagiarism_check.rb
```diff
@@ -100,12 +105,27 @@ def feedback_object
...
+      slices = entry_arr.each_cons(MATCH_MINIMUM)
+      passage_slices = passage_word_arrays.each_with_index.to_a.map(&:reverse).to_h
+      matched_slices = identify_matched_slices(passage_slices, slices)
+      return "" if matched_slices.empty?
+
+      build_longest_continuous_slice(matched_slices, passage_slices)
+    end
+
+    private def identify_matched_slices(indexable_slices, reference_slices)
+      indexable_slices.select {|k,v| v.in?(reference_slices) }.keys
```
dandrabik: Another place to use `_` for unused variable.

dandrabik (approved):
I also think it might make this cleaner to just have a single plagiarism check rather subclassing (also, if we are going to subclass, I think we'd want a base class that both `ExactMatch` and `FuzzyMatch` inherit from to make the shared/unique code distinction clearer). 

Also, has the fuzzy match made the exact match redundant?, i.e. if we just run the `FuzzyMatch`, would that return the same matches as running both the `FuzzyMatch` and the `ExactMatch`? If so, should we eliminate the `ExactMatch`?

So, I think it's worth revisiting this setup for a quick refactor. So you could either ship this PR and refactor this in a new PR, or just refactor in this PR (up to you, this seems fine to ship). 

Also, happy to chat through the different options with you.

anathomical:
@happythenewsad @dandrabik I originally had it broken out as a separate class because one of the original goals for the feature was that "fuzzy" and "direct" plagiarism would be separate rules.  However, further discussion about requirements has made that no longer the case and I agree with both of you that it's A) worth consolidating down to a single class, and B) "fuzzy" is a super-set of all plagiarism so we really only need one checker to run.

I'll make these updates today.

PR #8682, opened 2022-01-14
Changes to Edit Activity Pack page for Admin

cissyxyu:
## WHAT
We're revamping this page -- adding some new fields and new functionality like Upload a PDF and a full HTML editor.

## WHY
Curriculum is getting their long-awaited changes to these pages to make them more usable and relevent.

## HOW
Add some new fields, `readability` and `diagnostics` that give staff more info about an activity pack.
Add an upload file feature (re-using code from other identical features on admin panels).
Add an HTML editor using `TextEditor` component.
Add a script that will run once to convert all the activity descriptions in the database from Markdown to HTML (because we are no longer using Markdown to create, edit, or display this field).

### Screenshots
Screen Shot 2022-01-12 at 11 48 05 AM

### Notion Card Links
https://www.notion.so/quill/Overhauls-to-Individual-Activity-Pack-view-page-20cf3f183848451f94db618c1a356122

PR Checklist | Your Answer
------------ | -------------


services/QuillLMS/app/models/unit_template.rb
```diff
@@ -68,6 +68,10 @@ def readability
     "#{lowest_readability_range.split('-')[0]}-#{highest_readability_range.split('-')[1]}"
   end
 
+  def diagnostic_names
+    diagnostics_recommended_by.map {|d| d.name}
```
emilia-friedberg: I think it would technically be more efficient to do `.pluck(:name)` here since then you can get the activities and the names with just one query. Since this is unlikely to ever return more than a handful of results in practice it probably doesn't make a difference.

emilia-friedberg (approved):
One small comment, otherwise LGTM!

eadams17 (approved):

PR #8691, opened 2022-01-18
Refactor finish_lesson to find ActivitySessions one time only

cissyxyu:
## WHAT
We're having performance issues on this endpoint because we're making the same call `ActivitySession.find()` four different times on different methods.

Instead of calling these methods so many times, I'm refactoring the code to call this one time and then pass the resulting array through to the helper methods.

## WHY
This will hopefully speed up this endpoint.

## HOW
Only make one call to the database to find all the ActivitySessions. Then pass these sessions through to the helper methods instead of making each helper method replicate the same call again.

### Screenshots

### Notion Card Links
https://www.notion.so/quill/Performance-Api-V1-ClassroomUnitsController-finish_lesson-e6181d777ce64742b484584bb23cef84

PR Checklist | Your Answer
------------ | -------------


services/QuillLMS/app/models/activity_session.rb
```diff
@@ -389,13 +383,8 @@ def self.save_concept_results(classroom_unit_id, activity_id, concept_results)
...
 
-  def self.delete_activity_sessions_with_no_concept_results(classroom_unit_id, activity_id)
+  def self.delete_activity_sessions_with_no_concept_results(activity_sessions)
     incomplete_activity_session_ids = []
-    activity = Activity.find_by_id_or_uid(activity_id)
-    activity_sessions = ActivitySession.where(
-      classroom_unit_id: classroom_unit_id,
-      activity: activity
-    )
     activity_sessions.each do |as|
```
happythenewsad, line 388: Optional nit: you could make the previous code a little tighter with `select` + `map`

services/QuillLMS/spec/controllers/api/v1/classroom_units_controller_spec.rb
```diff
@@ -75,6 +75,9 @@
   context '#finish_lesson' do
     let(:follow_up_activity) { create(:activity) }
     let(:activity) { create(:activity, follow_up_activity: follow_up_activity) }
+    # let(:student) { create(:student)}
```
happythenewsad: forgot to delete?

happythenewsad (approved):
Love it! But you should delete the commented out lines unless there's a strong reason to keep them.

emilia-friedberg (approved):

anathomical (approved):
Nice solution.

PR #8695, opened 2022-01-19
have apis just use quill authentication

emilia-friedberg:
## WHAT
Include `QuillAuthentication` module in the `api_controller` so we don't need a separate `current_user` method.

## WHY
The API version of the `current_user` method had once been the same as the `QuillAuthentication` version, but diverged as we made changes, which caused at least one bug where the demo account couldn't really present Quill Lessons (because nothing was coming through as the current user). Rather than try to keep two separate methods in sync, there didn't really seem to be a downside to just including QuillAuthentication here, and we were already using it in some specific API controllers.

## HOW
Just include the module and remove inclusions in sub-controllers.

### Screenshots

### Notion Card Links
https://www.notion.so/quill/Unable-to-go-to-Next-Slide-on-Quill-Lessons-ba94a468e7244278808d675738eaf87d

PR Checklist | Your Answer
------------ | -------------

### DIFF
--- a/services/QuillLMS/app/controllers/api/api_controller.rb
+++ b/services/QuillLMS/app/controllers/api/api_controller.rb
@@ -35,10 +35,4 @@ class Api::ApiController < ActionController::Base
     response.headers['X-Platform-Spec'] = 'https://github.com/interagent/http-api-design'
     response.headers['X-API-Reference'] = 'http://docs.empirical.org/api-reference/'
   end
-
-  private def doorkeeper_token
-    return @token if instance_variable_defined?(:@token)
-    methods = Doorkeeper.configuration.access_token_methods
-    @token = Doorkeeper::OAuth::Token.authenticate(request, *methods)
-  end
 end
--- a/services/QuillLMS/app/controllers/api/api_controller.rb
+++ b/services/QuillLMS/app/controllers/api/api_controller.rb
@@ -1,6 +1,7 @@
 # frozen_string_literal: true
 
 class Api::ApiController < ActionController::Base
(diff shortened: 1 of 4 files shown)


services/QuillLMS/app/controllers/api/api_controller.rb
```diff
@@ -35,23 +36,6 @@ class AccessForbidden < StandardError; end
...
-          return @current_user ||= User.find_by_token!(username) if username.present?
-        end
-      end
-    rescue ActiveRecord::RecordNotFound
-      sign_out
-      nil
-    end
-  end
-
   private def doorkeeper_token
```
brendanshean: Is this method still needed?

brendanshean (approved):
I really appreciate that you are doing this.

services/QuillLMS/app/controllers/api/api_controller.rb
anathomical: I think we no longer use doorkeeper, but I couldn't guarantee it off-hand.

anathomical (approved):
Good call.  Glad you're solving this.

services/QuillLMS/app/controllers/api/api_controller.rb
brendanshean: QuillAuthentication already has a doorkeeper_token method, so I think the method here is redundant.

emilia-friedberg: @brendanshean I'll remove it, thanks for flagging.

PR #8696, opened 2022-01-19
Fix Deadlock with AssignRecommendationsWorker

dandrabik:
## WHAT
Adding the touches need by caching caused some [deadlock errors](https://one.newrelic.com/launcher/nr1-core.launcher?platform[timeRange][duration]=604800000&pane=eyJuZXJkbGV0SWQiOiJlcnJvcnMtdWkub3ZlcnZpZXciLCJiYXJjaGFydCI6ImJhcmNoYXJ0IiwidG9wRmFjZXQiOiJ0cmFuc2FjdGlvblVpTmFtZSIsInBhZ2UiOiJzaG93IiwicHJpbWFyeUZhY2V0IjoiZXJyb3IuY2xhc3MiLCJ0cmFjZUlkIjoiM2FiNDlkZTktNzYxZi0xMWVjLWFmYzYtMDI0MmFjMTEwMDEzXzBfNjc4MiIsImVudGl0eUd1aWQiOiJNall6T1RFeE0zeEJVRTE4UVZCUVRFbERRVlJKVDA1OE5UUTRPRFUyT0RjMSJ9&sidebars[0]=eyJuZXJkbGV0SWQiOiJucjEtY29yZS5hY3Rpb25zIiwiZW50aXR5R3VpZCI6Ik1qWXpPVEV4TTN4QlVFMThRVkJRVEVsRFFWUkpUMDU4TlRRNE9EVTJPRGMxIiwic2VsZWN0ZWROZXJkbGV0Ijp7Im5lcmRsZXRJZCI6ImVycm9ycy11aS5vdmVydmlldyJ9fQ==&state=8237a9af-e401-8426-39e5-9d3f3cd1c2e0) in the `AssignRecommendationsWorker.` 

I have not been able to replicate that deadlock in a test or in general testing, but it was caused by a `touch` on classrooms (it went away when the `touch` was removed). 

This PR converts two sets of `touch`es into `"touches without callbacks"` which will hopefully, prevent circular deadlocks on resources. As I stated, I can't replicate the issue (in production it was occurring in a small percentage of `AssignRecommendationsWorker` runs), but cascading touches seem to be the cause.
## WHY
Deadlocks are bad, and I'd like to get the caching code out with accurate `updated_at` timestamps.
## HOW
Attempt to avoid deadlocks by triggering less callbacks

PR Checklist | Your Answer
------------ | -------------


services/QuillLMS/spec/services/units/creator_spec.rb
```diff
@@ -5,11 +5,11 @@
 Sidekiq::Testing.fake!
 
 describe Units::Creator do
```
dandrabik, line 7: This file was a small cleanup that is unrelated (I attempted to trigger the deadlock from many different angles). Leaving it in, since it's slightly cleaner.

happythenewsad (approved):
Let's snapshot the controller's current execution time so that we can compare after this PR is deployed.

brendanshean (approved):
I think this seems like a sensible approach to sorting out the deadlocks.

PR #8721, opened 2022-01-21
fix bug where demo creator was trying to use the same student account for different demo teachers

emilia-friedberg:
## WHAT
Update the report demo creator script to only generate an Angie Thomas student with an email address for the standard, teacher-facing demo account.

## WHY
Using a student with a hardcoded email in the demo script, and deleting other students with that email in the database, was resulting in a bug because we use multiple demo teacher accounts and so one of them would always end up missing a student.

## HOW
Only hardcode the email for the student if there is no email for the teacher passed in, which is the same as creating the generic teacher demo account.

### Screenshots

### Notion Card Links
https://www.notion.so/quill/Quill-demo-report-shows-5-of-4-students-completed-1351a3b2da704a8394879970ef9475b4

PR Checklist | Your Answer
------------ | -------------

### DIFF
--- a/services/QuillLMS/app/services/demo/report_demo_creator.rb
+++ b/services/QuillLMS/app/services/demo/report_demo_creator.rb
@@ -210,7 +210,7 @@ module Demo::ReportDemoCreator
   def self.create_demo(email)
     teacher = create_teacher(email)
     classroom = create_classroom(teacher)
-    students = create_students(classroom)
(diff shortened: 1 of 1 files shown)


anathomical (approved):

cissyxyu (approved):
Good catch on this! Sorry for missing this when I deployed these changes, should have tested beforehand to account for this edge case.

PR #8725, opened 2022-01-21
Profanity highlight

happythenewsad:
## WHAT
Adds profanity highlighting in student responses. 

## WHY
Product request.

## HOW
This required a bit of refactoring.
- the `profanity` library now returns the actual profane word or nil, rather than a boolean. This allows us to check for profanity only once rather than twice (it's an expensive operation)
- the profanity check is the only check that depends on instance variables (the instance variable memoizes the expensive operation). Thus, it cannot be initialized as a lambda-constant. I moved all checks from lamba-constants to lambda-instance variables for consistency.
- 
Screen Shot 2022-01-21 at 10 21 31 AM

### Notion Card Links
https://www.notion.so/quill/Product-Board-30f97e4eb01246dbb8264253fb385073?p=44d09dfa7511488c98e11186eb774648

PR Checklist | Your Answer
------------ | -------------

### DIFF
--- a/services/QuillLMS/engines/evidence/app/models/evidence/highlight.rb
+++ b/services/QuillLMS/engines/evidence/app/models/evidence/highlight.rb
@@ -9,16 +9,11 @@ module Evidence
     MIN_TEXT_LENGTH = 1
     MAX_TEXT_LENGTH = 5000
 
-    TYPE_PASSAGE  = 'passage'
-    TYPE_RESPONSE = 'response'
-    TYPE_PROMPT   = 'prompt'
-
     TYPES = [
-      TYPE_PASSAGE,
-      TYPE_RESPONSE,
-      TYPE_PROMPT
+      TYPE_PASSAGE = 'passage',
+      TYPE_RESPONSE = 'response',
+      TYPE_PROMPT = 'prompt'
     ]
-
     belongs_to :feedback, inverse_of: :highlights
 
     validates :text, presence: true, length: {minimum: MIN_TEXT_LENGTH, maximum: MAX_TEXT_LENGTH}
--- a/services/QuillLMS/engines/evidence/app/models/evidence/highlight.rb
+++ b/services/QuillLMS/engines/evidence/app/models/evidence/highlight.rb
@@ -8,10 +8,10 @@ module Evidence
     
     MIN_TEXT_LENGTH = 1
     MAX_TEXT_LENGTH = 5000
-    TYPES= [
-      'passage',
-      'response',
-      'prompt'
+    TYPES = [
+      PASSAGE = 'passage',
+      RESPONSE = 'response',
+      PROMPT = 'prompt'
     ]
 
     belongs_to :feedback, inverse_of: :highlights
--- a/services/QuillLMS/engines/evidence/app/models/evidence/highlight.rb
+++ b/services/QuillLMS/engines/evidence/app/models/evidence/highlight.rb
@@ -8,10 +8,15 @@ module Evidence
     
     MIN_TEXT_LENGTH = 1
     MAX_TEXT_LENGTH = 5000
+
+    TYPE_PASSAGE  = 'passage'
+    TYPE_RESPONSE = 'response'
+    TYPE_PROMPT   = 'prompt'
+
     TYPES = [
-      PASSAGE = 'passage',
-      RESPONSE = 'response',
-      PROMPT = 'prompt'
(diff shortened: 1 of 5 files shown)


services/QuillLMS/engines/evidence/lib/evidence/evidence/prefilter_check.rb
```diff
@@ -13,17 +13,30 @@ class PrefilterCheck
...
-      QUESTION_MARK_RULE_UID      => ->(entry) { entry.match?(/\?$/) },
-      MULTIPLE_SENTENCE_RULE_UID  => ->(entry) { sentence_count(entry) > 1 },
-      PROFANITY_RULE_UID          => ->(entry) { Profanity.profane?(entry) },
-      MINIMUM_WORD_RULE_UID       => ->(entry) { word_count(entry) < MINIMUM_WORD_COUNT}
-    }
-
     def initialize(entry)
+      # When a prefilter lambda identifies a violation, it returns true
+      @prefilters = {
+        QUESTION_MARK_RULE_UID      => ->(the_entry) { the_entry.match?(/\?$/) },
```
happythenewsad, line 20: The lambdas' arg name is `the_entry` to avoid variable shadowing, which rubocop complains about.

services/QuillLMS/engines/evidence/app/models/evidence/highlight.rb
```diff
@@ -8,10 +8,15 @@ class Highlight < ApplicationRecord
     
     MIN_TEXT_LENGTH = 1
     MAX_TEXT_LENGTH = 5000
-    TYPES= [
-      'passage',
-      'response',
-      'prompt'
+
+    TYPE_PASSAGE  = 'passage'
```
happythenewsad: This re-org allows use to reference Highlight constants from PrefilterCheck, for data integrity.

anathomical: Just a syntax note, but you can combine these two things.  We often do this just to keep things consolidated:
```ruby
TYPES = [
  TYPE_PASSAGE = 'passage,
  TYPE_RESPONSE = 'response',
  TYPE_PROMPT = 'prompt'
]
```

anathomical (approved):
This approach feels a little bit odd to me, but I think that it's still probably right to keep all the pre-filters in a single file, which is what leads to some of the weird code here, so it's probably not worth a big refactor right now.  Looks good to me.

services/QuillLMS/engines/evidence/app/models/evidence/highlight.rb
happythenewsad: Addresed.

services/QuillLMS/engines/evidence/lib/evidence/evidence/prefilter_check.rb
```diff
@@ -13,17 +13,30 @@ class PrefilterCheck
...
+        MULTIPLE_SENTENCE_RULE_UID  => ->(the_entry) { PrefilterCheck.sentence_count(the_entry) > 1 },
+        PROFANITY_RULE_UID          => ->(the_entry) { profanity? },
+        MINIMUM_WORD_RULE_UID       => ->(the_entry) { PrefilterCheck.word_count(the_entry) < MINIMUM_WORD_COUNT}
+      }
+
       @entry = entry
       @prefilter_rules = Evidence::Rule.where(rule_type: Evidence::Rule::TYPE_PREFILTER).includes(:feedbacks)
+      @violated_rule = nil
+      @profanity_instance = nil
```
brendanshean: Could we set `@profanity_instance = Profanity.profane(entry)`?  This would remove the side effect in `profanity?` and simplify that method:

```
def profanity?
  @prefilter_instance.present?
end
```

I realize `nil?` and `present?` aren't quite opposites, but I'm assuming that profanity is a non-empty string.

brendanshean (approved):
This looks good.

I would probably refactor the lambdas into separate files and then standardize the interface to those prefilter checks, but that might be more anticipating future additions to this process and such a refactor, as Thomas pointed out, might not be needed at this time.

services/QuillLMS/engines/evidence/lib/evidence/evidence/prefilter_check.rb
happythenewsad: Yeah, I agree that's cleaner. Addressed.

PR #8728, opened 2022-01-21
evidence/fuzzy-plagiarism-performance-enhancements

anathomical:
## WHAT
This is a first pass to tweak fuzzy plagiarism match to be more performant.  On my local machine benchmark time fell from approximately 270ms to 170ms per run with the test dataset.

Also add a benchmarking rake task to calculate performance improvements
## WHY
By not running Levenshtein comparisons if there aren't enough exact matching words to allow the maximum fuzziness to exist.
## HOW
We know that we only care about Levenshtein distance if it is low enough.  Given the size of the word arrays we compare, that means that for the Levenshtein distance to be as low as we want requires that a minimum number of words in the comparison be identical.  This new code basically confirms that there are enough matching words that it's possible for the Levenshtein distance to be as low as we want before computing said distance, allowing us to reduce the required CPU time by short-circuiting the most expensive part of the algorithm when we know it won't find a match anyway.

NOTE: The benchmarked performance improvement (a 40%-ish reduction in run time) may not actually achieve our target of 500ms for all production runs, but I want to get this into production to see how the benchmark correlates with production times to figure out how much more aggressive we might need to be with algorithm revisions.

### Notion Card Links

PR Checklist | Your Answer
------------ | -------------
Have you added and/or updated tests? |  No.  We don't actually want to change behavior/outcomes here

### DIFF
--- a/services/QuillLMS/engines/evidence/lib/evidence/evidence/plagiarism_check.rb
+++ b/services/QuillLMS/engines/evidence/lib/evidence/evidence/plagiarism_check.rb
@@ -125,10 +125,21 @@ module Evidence
 
     private def identify_matched_slices(entry_slices, passage_slices)
       entry_slices.select do |_, slice|
+        # Check to see if there's enough similarity for it to be worth doing a Levenshtein comparison
+        next false unless confirm_minimum_overlap(slice, passage_slices)
         slice_string = slice.join(' ')
         passage_slice_strings = passage_slices.map { |s| s.join(' ') }
         passage_slice_strings.any? { |passage_string| DidYouMean::Levenshtein.distance(slice_string, passage_string) <= FUZZY_CHARACTER_THRESHOLD }
       end.keys
+    end
+
+    private def confirm_minimum_overlap(target_array, source_arrays)
+      # Since we allow character deviation that's smaller than the total number of words compared
+      # we know that there's a minimum number of words that have to be identical.  This function
+      # confirms that minimum amount of overlap to justify doing a set of Levenshtein calculations.
+      source_arrays.any? do |source_array|
+        (source_array & target_array).length >= (MATCH_MINIMUM - FUZZY_CHARACTER_THRESHOLD)
+      end
     end
 
     # using the indices of plagiarized slices, find the longest continuous plagiarized section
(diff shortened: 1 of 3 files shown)


services/QuillLMS/engines/evidence/lib/tasks/benchmark.rake
```diff
@@ -0,0 +1,31 @@
+# frozen_string_literal: true
+
+require 'benchmark'
+
+namespace :benchmark do
+
+  desc "Runs a benchmark on the Plagiarism algorithm"
+
+  task :plagiarism => :environment do |t, args|
```
dandrabik, line 9: I like these reusable benchmarks! This will be very helpful if we need to optimize more.

services/QuillLMS/engines/evidence/lib/evidence/evidence/plagiarism_check.rb
```diff
@@ -125,12 +125,23 @@ def feedback_object
...
       entry_slices.select do |_, slice|
+        # Check to see if there's enough similarity for it to be worth doing a Levenshtein comparison
+        next false unless confirm_minimum_overlap(slice, passage_slices)
         slice_string = slice.join(' ')
         passage_slice_strings = passage_slices.map { |s| s.join(' ') }
         passage_slice_strings.any? { |passage_string| DidYouMean::Levenshtein.distance(slice_string, passage_string) <= FUZZY_CHARACTER_THRESHOLD }
       end.keys
     end
 
+    private def confirm_minimum_overlap(target_array, source_arrays)
```
dandrabik: take or leave, consider using Ruby boolean method naming conventions with question mark and readable verb, e.g.

```ruby
def has_minimum_overlap?(target_array, source_arrays)

dandrabik (approved):
Thanks for improving this. I like the iterative approach of trying a small refactor and seeing where that lands us. And it's nice to have the benchmarks for further testing.

services/QuillLMS/engines/evidence/lib/evidence/evidence/plagiarism_check.rb
anathomical: Good call. I actually prefer the convention for readability, so thanks for reminding me I'd missed it here.

PR #8732, opened 2022-01-24
Update Rubocop

brendanshean:
## WHAT
This is a preliminary upgrade of rubocop to 1.25.0.  We spoke about linting at the last Engineering meeting.  I'm including everyone on this PR so that they can inspect all of the potential cops that will be added to our codebase.  There's quite a few but they're pretty straightforward for the most part.  I left out the more complicated ones for later PRs.

## WHY
rubocop v1.0 was the first stable release of the gem.  Upon its release authors decided to start partitioning different "departments" of rubocop into separate gems (e.g. `rubocop-rspec`).  Upgrading the base `rubocop` library will allow us to use these different departments more effectively. 

## HOW
1)  upgrade the `rubocop.sh` to utilize the appropriate version.
2) Incrementally each of new cops and evaluate their inclusion in our build process (e.g. `rubocop -A --only Style/QuotedSymbols`).  There are quite a few commits in this PR but most of three types:  
a) `Enable cop`:  No violations were found and the cop is simply added.
b) `Disable cop`: Too many false positives were found so this cop is disabled. 
c) `Fix violations`: A cop is added and the existing violations in our codebase are fixed.

### Notion Card Links
https://www.notion.so/quill/Add-Rubocop-rules-for-rspec-89b82487ec734ebe9d1a7b04db015a55

PR Checklist | Your Answer
------------ | -------------


services/QuillLMS/app/controllers/errors_controller.rb
```diff
@@ -1,11 +1,11 @@
 # frozen_string_literal: true
 
 class ErrorsController < ApplicationController
-  def error_404
+  def error404
     render 'error_404', :status => 404
```
dandrabik: If we are renaming these paths, we should rename the views to match. (`error_500` and `error_404`)

services/QuillLMS/app/controllers/teachers_controller.rb
```diff
@@ -126,9 +126,7 @@ def completed_diagnostic_unit_info
   end
 
   def lessons_info_for_dashboard_mini
-    if !current_user
-      units = []
-    else
+    if current_user
```
dandrabik: I think this change is fine (especially for such a large changeset already), but this might be cleaner with an early return, e.g.

```ruby
if current_user.nil?
  return render json: { units: [] }
end

File: services/QuillLMS/app/graphql/mutations/concepts/create.rb [hunk 3]
```diff
@@ -2,10 +2,10 @@
 
 class Mutations::Concepts::Create < Mutations::BaseMutation
   def self.authorized?(value, context)
-    if !context[:current_user].staff?
-      raise GraphQL::ExecutionError, "Only staff can run this mutation"
-    else
+    if context[:current_user].staff?
```
dandrabik: Early return possible

```ruby
return if context[:current_user].staff?

raise GraphQL::ExecutionError, "Only staff can run this mutation"

File: services/QuillLMS/app/graphql/mutations/concepts/edit.rb [hunk 4]
```diff
@@ -2,10 +2,10 @@
 
 class Mutations::Concepts::Edit < Mutations::BaseMutation
   def self.authorized?(value, context)
-    if !context[:current_user].staff?
-      raise GraphQL::ExecutionError, "Only staff can run this mutation"
-    else
+    if context[:current_user].staff?
```
dandrabik: (take or leave) early return / eliminate branching possible.

services/QuillLMS/app/graphql/mutations/concepts/replace.rb
```diff
@@ -2,10 +2,10 @@
 
 class Mutations::Concepts::Replace < Mutations::BaseMutation
   def self.authorized?(value, context)
-    if !context[:current_user].staff?
-      raise GraphQL::ExecutionError, "Only staff can run this mutation"
-    else
+    if context[:current_user].staff?
```
dandrabik: (take or leave) early return / eliminate branching possible.

services/QuillLMS/app/graphql/types/user_type.rb
```diff
@@ -54,7 +54,7 @@ def completed_diagnostic
 
   private def get_recommended_units(concept_result_scores)
     units = []
-    recommendations = Recommendation.where(activity_id: 413, category: 0).map do |activity_pack_recommendation|
+    Recommendation.where(activity_id: 413, category: 0).limit(1).each do |activity_pack_recommendation|
```
dandrabik: Where does the `.limit(1)` come in? I could be reading this wrong, but I think this might be different logic than the original if there are more than 1 recommendations.

services/QuillLMS/app/models/classroom.rb
```diff
@@ -123,7 +123,7 @@ def unique_standard_count_array
   end
 
   def archived_classrooms_manager
-    coteachers = !self.coteachers.empty? ? self.coteachers.map { |ct| { name: ct.name, id: ct.id, email: ct.email } } : []
+    coteachers = self.coteachers.empty? ? [] : self.coteachers.map { |ct| { name: ct.name, id: ct.id, email: ct.email } }
```
dandrabik: This is redundant to:
```ruby
coteachers = self.coteachers.map { |ct| { name: ct.name, id: ct.id, email: ct.email } }

File: services/QuillLMS/app/models/concept.rb [hunk 8]
```diff
@@ -78,13 +78,9 @@ def self.level_one_only
...
   def self.find_by_id_or_uid(arg)
-    begin
-      find(arg)
-    rescue ActiveRecord::RecordNotFound
-      find_by(uid: arg)
-    rescue ActiveRecord::RecordNotFound
-      raise ActiveRecord::RecordNotFound, "Couldn't find Concept with 'id' or 'uid'=#{arg}"
-    end
+    find_by(uid: arg) || find(arg)
+  rescue ActiveRecord::RecordNotFound
```
dandrabik: Debatable value to the rescue and re-raise. Probably removable.

services/QuillLMS/app/models/concerns/user_cacheable.rb
```diff
@@ -18,36 +18,36 @@
 module UserCacheable
   extend ActiveSupport::Concern
 
-  def all_classrooms_cache(key:, groups: {})
-    model_cache(last_updated_classroom, key: key, groups: groups) { yield }
+  def all_classrooms_cache(key:, groups: {}, &block)
```
dandrabik, line 21: Oh, interesting, I wasn't sure the best way to write this. I do like this explicit `&block` better

services/QuillLMS/app/models/subscription.rb
```diff
@@ -196,22 +196,22 @@ def self.new_school_premium_sub(school, user)
   def self.give_teacher_premium_if_charge_succeeds(user)
     teacher_premium_sub = new_teacher_premium_sub(user)
     teacher_premium_sub.save_if_charge_succeeds('teacher')
-    if !teacher_premium_sub.new_record?
+    if teacher_premium_sub.new_record?
```
dandrabik: Consider `return false if teacher_premium_sub.new_record?`

services/QuillLMS/app/models/unit_template.rb
```diff
@@ -87,13 +87,14 @@ def activity_ids
   end
 
   def self.user_scope(user_flag)
-    if user_flag == PRIVATE
+    case user_flag
```
dandrabik: A style guide discussion for another date, but simple `case` statements can be hash lookups, e.g.
```ruby
USER_SCOPES = {
  PRIVATE => UnitTemplate.private_user,
  ALPHA => UnitTemplate.alpha_user
...
}

USER_SCOPES.fetch(user_flag, UnitTemplate.production)

File: services/QuillLMS/app/models/user.rb [hunk 12]
```diff
@@ -441,11 +438,11 @@ def last_name= last_name
   end
 
   def first_name
-    @first_name ||= name.to_s.split("\s")[0]
+    @first_name ||= name.to_s.split[0]
```
dandrabik, line 441: I think any whitespace regex is different than the default `' '`. 
`/\s/ - A whitespace character: /[ \t\r\n\f\v]/`

services/QuillLMS/lib/eg_form_builder.rb
```diff
@@ -181,12 +179,12 @@ def field *args, &block
       input_options[:placeholder] = ((placeholder = options.delete(:placeholder)) == true ? name.to_s.humanize : placeholder)
     end
 
-    unless options[:hidden].nil?
-      input_options[:class] = 'hidden' if options[:hidden] == true
+    if !options[:hidden].nil? && (options[:hidden] == true)
```
dandrabik: I think this could be 
```ruby
if options[:hidden] == true

File: services/QuillLMS/lib/tasks/schools.rake [hunk 14]
```diff
@@ -6,22 +6,11 @@ require 'open-uri'
...
-      nil
-  when 'N'
-      nil
-  when 'UG'
-      nil
-  when 'AE'
-      nil
-  when '13'
-      12
+  when 'PK', 'KG' then 0
```
dandrabik, line 9: Another case statement that's probably better as a hash.

dandrabik (approved):
I tried to leave comments whenever something might need to be looked at again. Almost everything is take or leave.

Thanks for working through this all and helping to make our code better!

anathomical (approved):

services/QuillLMS/app/controllers/teachers_controller.rb
brendanshean: I agree.  I'll make that a guard clause.

services/QuillLMS/app/graphql/types/user_type.rb
brendanshean: Good question.  I was a little confused by the original code.  There was a `return units` statement at the end of the loop, rendering subsequent iterations unreachable.  I'm going to update this to just use a `find_by` and remove the outer loop to make this more readable.

services/QuillLMS/app/models/classroom.rb
brendanshean: Good point!  Sounds like a rubocop needs a new cop.

services/QuillLMS/app/models/concept.rb
brendanshean: I would agree.  Will remove along with app/models/activity.rb one.

services/QuillLMS/app/models/concerns/user_cacheable.rb
brendanshean: One gotcha to note is that the `&block` does not raise [LocalJumpError](https://github.com/empirical-org/Empirical-Core/pull/8732/files#diff-e2f5b86283bb42eea52843a7d2be4c02f8f44285f137ccb95c702c7d3bfa04f6R48) like the implicit block from yield does.  Good discussion [here](https://stackoverflow.com/a/47237614)

brendanshean: I think was my favorite cop that I've added in this PR since it refactored some code I recently wrote in a more Ruby way (e.g. https://github.com/empirical-org/Empirical-Core/pull/8732/files#diff-8c18a29c1f1d227f4fe816ed5f3c3e133a415311e74b6ceb142fe2c2fd6f8134R14)

services/QuillLMS/app/models/subscription.rb
brendanshean: Definitely.  You're reminding me of a [GuardClause cop](https://www.rubydoc.info/gems/rubocop/RuboCop/Cop/Style/GuardClause) that I'm guessing we don't have enabled.

services/QuillLMS/app/models/user.rb
brendanshean: I was actually surprised by this cop too:

```
[1] pry(main)> "A\tb\rc\nd\fe\v".split("\s")
=> ["A", "b", "c", "d", "e"]
[2] pry(main)> "A\tb\rc\nd\fe\v".split
=> ["A", "b", "c", "d", "e"]
```
The default for split is not `' '` but instead just [whitespace](https://ruby-doc.org/core-2.4.0/String.html#method-i-split).

services/QuillLMS/lib/tasks/schools.rake
brendanshean: I'll probably just leave this one as-is since it's a rake task.

services/QuillLMS/app/controllers/errors_controller.rb
brendanshean: Agreed

services/.rubocop.yml
```diff
@@ -9,292 +9,381 @@ AllCops:
...
+  Enabled: true
+
+Layout/EmptyLinesAroundBeginBody:
+  Enabled: true
+
+Layout/EmptyLinesAroundExceptionHandlingKeywords:
+  Enabled: true
+
+Layout/EndAlignment:
+  Enabled: false
```
happythenewsad, line 70: Note to self: this isn't a new cop, just moved

happythenewsad (commented):
Can't say I evaluated every single change, but everything I saw looks good :-)

PR #8737, opened 2022-01-25
Add rubocop-rspec cops

brendanshean:
## WHAT
Add rubocop-rspec gem to our circleci build

## WHY
Adding these cops will improve the quality of our specs in the LMS

## HOW
Add `rubocop-rspec` to our `.rubocop.yml` configuration.

### Screenshots

### Notion Card Links
https://www.notion.so/quill/Add-Rubocop-rules-for-rspec-89b82487ec734ebe9d1a7b04db015a55

PR Checklist | Your Answer
------------ | -------------


services/.rubocop.yml
```diff
@@ -1,13 +1,24 @@
...
   merge:
     - Exclude
 
 AllCops:
   Exclude:
+    - 'QuillConnect/node_modules/**/*'
+    - 'QuillComprehension/node_modules/**/*'
+    - 'QuillDiagnostic/node_modules/**/*'
+    - 'QuillGrammar/node_modules/**/*'
+    - 'QuillLessons/node_modules/**/*'
```
anathomical: There's no problem with adding these directories to the exclusion, but I'm actually pretty sure we just want to remove these directories entirely.  In addition to the `QuillProofreader` directory below.  I think all this code was moved into `QuillLMS/client`

anathomical (approved):
A note on "dead" directories being excluded, but all these rules seem okay to me.

services/QuillLMS/engines/evidence/spec/controllers/evidence/feedback_controller_spec.rb
```diff
@@ -208,53 +210,53 @@ module Evidence
...
         post("plagiarism", :params => ({ :entry => "No plagiarism here.", :prompt_id => 100, :session_id => 1, :previous_feedback => ([]) }), :as => :json)
-        expect(404).to(eq(response.status))
+        expect(response.status).to(eq(404))
       end
 
       it 'should return successfully when there is plagiarism that matches the first plagiarism text string' do
         post("plagiarism", :params => ({ :entry => ("bla bla bla #{plagiarized_text1}"), :prompt_id => prompt.id, :session_id => 1, :previous_feedback => ([]) }), :as => :json)
         parsed_response = JSON.parse(response.body)
-        expect(false).to(eq(parsed_response["optimal"]))
+        expect(parsed_response["optimal"]).to(eq(false))
```
dandrabik: Is it preferred to use `be false` vs `eq false`? (I don't have a strong opinion here)

services/QuillLMS/spec/controllers/activities_controller_spec.rb
```diff
@@ -93,15 +93,14 @@
     context 'student is assigned to classroom_unit' do
       before { classroom_unit.update(assigned_student_ids: [student.id]) }
 
-      it '' do
+      it do
```
dandrabik: It might be better to fill these in instead of dropping.

services/.rubocop.yml
```diff
@@ -336,6 +341,120 @@ Naming/VariableName:
...
   Enabled: true
 
+#######################################################################################################################
+# RSPEC
+#######################################################################################################################
+
+RSpec/AroundBlock:
+  Enabled: true
+
+RSpec/BeEql:
```
dandrabik, line 351: Surprised this didn't flag the `eq false`? I guess it only flags ` eql false`

dandrabik (approved):
Looks good to me, thanks for cleaning this all up. Two very minor take or leave comments.

Also, big :+1: for the cop which enforces the order of the expectations, e.g. swapping order of `expect("Dan").to eq user.name` (oddly, I can't find this in the config list, but I think it should be [this one](https://www.rubydoc.info/gems/rubocop-rspec/1.15.0/RuboCop/Cop/RSpec/ExpectActual)). So much better.

services/QuillLMS/engines/evidence/spec/controllers/evidence/feedback_controller_spec.rb
brendanshean: I think it depends on how restrictive you want the matching to be.  I tend to favor strict matching so I'll change this to be false.

brendanshean: Some of the cops, like `RSpec/ExpectActual` are enabled by [default](https://docs.rubocop.org/rubocop-rspec/cops_rspec.html#rspecexpectactual) so we don't have to explicitly reference them in the config.

PR #8738, opened 2022-01-25
Evidence/narrow fuzzy match threshold

anathomical:
## WHAT
Narrow our fuzzy-matching plagiarism algorithm so that only student entries that are within 3 characters (instead of the old 5) will be flagged as plagiarism
## WHY
We have some code within the algorithm that short-circuits the relatively CPU-intensive Levenshtein distance calculation if there isn't already enough similarity between the student entry and the `plagiarism_text` of the prompt being responded to.  This short-circuit is allowed to be more aggressive the fewer characters of difference we allow, which lets us more often skip the need to do any Levenshtein calculations at all for most entries (which have no plagiarism).

Basically, while we derive no pedagogical benefits from this change, we do derive performance benefits, which will hopefully bring the max run-time of this algorithm down below our target of 500ms.

One nice thing worth noting here, is that an analysis of 10k student entries from the `FeedbackHistory` table reveals that 3-character matching and 5-character matching have 99.97% the same results.  This means that we shouldn't be reducing our pedagogical rigor with this change.
## HOW
Fortunately, the specificity of fuzzy-matching is a constant, so I simply reduced the value from 5 to 3.

### Notion Card Links

PR Checklist | Your Answer
------------ | -------------

### DIFF
--- a/services/QuillLMS/engines/evidence/lib/evidence/evidence/plagiarism_check.rb
+++ b/services/QuillLMS/engines/evidence/lib/evidence/evidence/plagiarism_check.rb
@@ -8,7 +8,7 @@ module Evidence
     ENTRY_TYPE = 'response'
     MATCH_MINIMUM = 10
     OPTIMAL_RULE_KEY = 'optimal_plagiarism_rule_serialized'
-    FUZZY_CHARACTER_THRESHOLD = 5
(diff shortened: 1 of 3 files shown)


dandrabik (approved):
Thanks for these iterative changes!

PR #8741, opened 2022-01-26
active activity session performance

happythenewsad:
## WHAT
Part one of an initiative to speed up active_activity_session#update
Stops the returning of an unnecessary payload.

I've also included the benchmark I created that motivates this change. Including benchmarks within specs is a new pattern for us. I think it's a good idea, because it's a clear way to document how something was benchmarked, and it can be disabled when not working with it directly.

## WHY
This controller action is not scaling well. 

### Screenshots

### Notion Card Links
[(Please provide links to any relevant Notion card(s) relevant to this PR.)](https://www.notion.so/quill/Product-Board-30f97e4eb01246dbb8264253fb385073?p=767be0eaa3204811abc7763eb3b7a0f4)

PR Checklist | Your Answer
------------ | -------------
Have you added and/or updated tests? |  Yes. Also manually tested Connect & Evidence plays.

### DIFF
--- a/services/QuillLMS/app/controllers/api/v1/active_activity_sessions_controller.rb
+++ b/services/QuillLMS/app/controllers/api/v1/active_activity_sessions_controller.rb
@@ -12,7 +12,7 @@ class Api::V1::ActiveActivitySessionsController < Api::ApiController
     begin
       @activity_session = ActiveActivitySession.find_or_initialize_by(uid: params[:id])
       @activity_session.data ||= {}
-      @activity_session.data = @activity_session.data.merge(working_params)
+      @activity_session.data = params#@activity_session.data.merge(working_params)
       @activity_session.save!
     rescue ActiveRecord::RecordNotUnique => e
       # Due to the way that ActiveRecord handles unique validations such as the one on UID,
@@ -26,7 +26,7 @@ class Api::V1::ActiveActivitySessionsController < Api::ApiController
       retried = true
       retry
     end
-    render json: {}, status: 204
+    head :no_content
   end
 
   private def working_params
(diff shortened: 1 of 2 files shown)


anathomical:
Love the benchmarking, but have a question about the approach.  Given that the benchmark doesn't have a pass/fail condition, I'm not sure it belongs in our specs.  I've been implementing benchmarks this sprint as part of my work on the Plagiarism algorithm, and have been doing so as rake tasks.  I'm not sure that's better, but it does feel to me like we shouldn't put things in specs that can't fail somehow.

happythenewsad:
> Love the benchmarking, but have a question about the approach. Given that the benchmark doesn't have a pass/fail condition, I'm not sure it belongs in our specs. I've been implementing benchmarks this sprint as part of my work on the Plagiarism algorithm, and have been doing so as rake tasks. I'm not sure that's better, but it does feel to me like we shouldn't put things in specs that can't fail somehow.

Note that I've used `xit`, so that the benchmark doesn't actually run except when you explicitly want it to. I prefer to keep it in specs, since then all the created records get garbage collected within rspec by default. But I'm also happy to remove the benchmark code entirely as well.

anathomical:
> 
> Note that I've used `xit`, so that the benchmark doesn't actually run except when you explicitly want it to. I prefer to keep it in specs, since then all the created records get garbage collected within rspec by default. But I'm also happy to remove the benchmark code entirely as well.

On further consideration, you are benchmarking controller actions, which is more complicated than just unit testing a lib (which is what I've been doing in the Plagiarism module).  So I suppose you can't really easily do this in `rake`, and probably do need the scaffolding provided by `Rspec`.  So this makes sense to me.  We probably want to consolidate our approaches so that all our benchmarks are run through the same process just to avoid confusion in the future, but that probably means we should err on the side of putting them in Rspec rather than in Rake tasks.

anathomical (approved):

dandrabik (approved):
This looks good to me. 

Since we are removing the payload, let's confirm that the returned payload from this controller endpoint isn't being used anywhere. Other than that. Good to go!

happythenewsad:
> This looks good to me.
> 
> Since we are removing the payload, let's confirm that the returned payload from this controller endpoint isn't being used anywhere. Other than that. Good to go!

Confirmed.

PR #8743, opened 2022-01-26
Fix bug where repurposed old clever classrooms are not showing up on new teacher pages.

brendanshean:
## WHAT
Fix a bug in how the `CleverIntegration::TeacherImportedClassroomsUpdater` is run.

## WHY
We'd like for all applicable clever classrooms for a given teacher to be updated in our system.

## HOW
The crux of this has to do with classroom clever ids.  There are two arrays of clever ids that are compared:
1) The classroom_data clever_ids that are pulled from the clever api
2) `teacher.clever_classrooms.pluck(:clever_id)`

`CleverIntegration::TeacherImportedClassroomsUpdater` runs only items occurring in 1) and 2).  However, there might be classroom clever_ids from 1) not in the intersection that still exist in our system (i.e. `existing_classrooms_where_teacher_was_added_in_clever`)  We need to iterate over those items and add the appropriate `ClassroomsTeacher` object and role so that those classrooms show up on their dashboard.

### Screenshots

### Notion Card Links
https://www.notion.so/quill/Teachers-Classes-not-showing-up-through-Import-From-Clever-button-e0f4a9880c08478bb86a88690f8df06a

PR Checklist | Your Answer
------------ | -------------

### DIFF
--- a/services/QuillLMS/app/services/clever_integration/teacher_imported_classrooms_updater.rb
+++ b/services/QuillLMS/app/services/clever_integration/teacher_imported_classrooms_updater.rb
@@ -4,11 +4,15 @@ module CleverIntegration
   class TeacherImportedClassroomsUpdater < ApplicationService
     attr_reader :teacher_id
 
+    COTEACHER = ClassroomsTeacher::ROLE_TYPES[:coteacher]
+    OWNER = ClassroomsTeacher::ROLE_TYPES[:owner]
+
     def initialize(teacher_id)
       @teacher_id = teacher_id
     end
 
     def run
+      update_classrooms_teachers
       update_classrooms
       update_classrooms_students
     end
@@ -25,6 +29,10 @@ module CleverIntegration
       @classrooms_data ||= TeacherClassroomsData.new(teacher, serialized_classrooms_data)
     end
 
+    private def existing_classrooms_where_teacher_was_added_in_clever
+      ::Classroom.where(clever_id: classroom_clever_ids - imported_classrooms.pluck(:clever_id))
+    end
+
     private def imported_classrooms
       teacher.clever_classrooms.where(clever_id: classroom_clever_ids)
     end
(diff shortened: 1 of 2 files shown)


services/QuillLMS/app/services/clever_integration/teacher_imported_classrooms_updater.rb
```diff
@@ -38,12 +46,21 @@ def run
...
+     imported_classrooms.each { |classroom| ClassroomUpdater.run(classroom, classroom_data(classroom) )}
     end
 
     private def update_classrooms_students
       ImportClassroomStudentsWorker.perform_async(teacher_id, imported_classrooms.map(&:id))
     end
+
+    private def update_classrooms_teachers
+      existing_classrooms_where_teacher_was_added_in_clever.each do |classroom|
+        teacher.classrooms_teachers.create!(
```
dandrabik: So in this loop, there is a tiny race condition on the `.exists?` followed by the `.create!`. That's probably not the biggest deal, given the use.

I have a generic question though: Looking at this collection loop that could `raise`, will a `raise` in the middle of the collection cause issues when it's re-run? If it does, should this be different, e.g. a `find_or_create!` to avoid errors with the unique table index: `unique_classroom_and_user_ids_on_classrooms_teachers  (user_id,classroom_id) UNIQUE`

You have tests that cover if a `classroom_teacher`  exists already in the DB, so that may cover all scenarios, but it's hard for me to fully reason through all scenarios (errors, re-runs) at a glance.

E.g. say this `each` loop has 3 records (a,b,c) to `create!` and the second one (b) errors (race condition, some unforeseen duplicate unforeseen) will the re-run of this job not hit the previous first create of (a) because the finder is different? Would (b) succeed on the second run if it was a race condition issue?

dandrabik (approved):
Thanks for fixing this up! Approved!

Note, I left a comment about a scenario (small race-condition, possible issue with errors and re-runs) that I think may be fine but I think is worth confirming that to be true.

services/QuillLMS/app/services/clever_integration/teacher_imported_classrooms_updater.rb
brendanshean: That's a good point.  I think a `find_or_create!` is the right call.  I will update.

PR #8745, opened 2022-01-26
Evidence/simplify plagiarism highlighting

anathomical:
## WHAT
Modify the plagiarism algorithm to only find the first instance of plagiarism in an entry even if there happened to be multiple discreet instances.
## WHY
This allows us to short-circuit the most expensive part of the algorithm (the Levenshtein distance calculation) once we've confirmed the discovery of a single contiguous block of plagiarized text in the entry.  As soon as that block ends confirmed by finding non-plagiarized text), we stop checking to see if even more text from the response might be plagiarized, letting us skip some number of Levenshtein calculations.

NOTE: With the updated benchmark rake task, which I added some more real-world examples to using data from NewRelic, this only results in a 10% speed-up of the algorithm.  However, it should show more speed-up on some of our "worst case" sets of inputs where things get past our initial screens to start running Levenshtein distances, so I think it's worth getting into production to see how it impacts our longest-running queries.
## HOW
Rather than using streamlined `select` and `map` calls, which return `nil` if you `break` them, I've cobbled together a slightly dirtier `each` loop with an accumulator array that gets populated under the conditions we were using for the old `select` loop.  But once we reach conditions of an early exit, we `break` the loop and return the accumulator without doing further iterations on the input data.

### Notion Card Links

PR Checklist | Your Answer
------------ | -------------


services/QuillLMS/engines/evidence/lib/evidence/evidence/plagiarism_check.rb
```diff
@@ -124,13 +124,20 @@ def feedback_object
     end
 
     private def identify_matched_slices(entry_slices, passage_slices)
```
dandrabik: Small note here, I noticed that you are passing in the args: `passage_slices, slices` but they are labeled `entry_slices, passage_slices` (reversed order) in the method definition. Order may not matter for this, but we should make this accurate.

services/QuillLMS/engines/evidence/lib/evidence/evidence/plagiarism_check.rb
```diff
@@ -124,13 +124,20 @@ def feedback_object
...
-      entry_slices.select do |_, slice|
-        # Check to see if there's enough similarity for it to be worth doing a Levenshtein comparison
+      contiguous_matched_indexes = []
+      # Placeholder to be overridden during the first run of the loop that makes it past the confirm_minimum_overlap? guard statement.  If that never happens, we don't have to calculate this value
+      passage_slice_strings = nil
+      entry_slices.each do |index, slice|
```
dandrabik: I wonder if some of this code can be simplified, since we've simplified the logic requirements.

So the logic was previously to `"find all of the matching indexes, then find the longest set of consecutive indexes, form the string from them"`, 

Now we are doing `"Find all of the matching indexes, then stop once they aren't consecutive, then find the longest set of consecutive indexes, form the string from them"`. There's a bit of redundancy there.

dandrabik (approved):
Some small possible refactors, but this looks ok to test.

services/QuillLMS/engines/evidence/lib/evidence/evidence/plagiarism_check.rb
anathomical: Because we use the `identify_matched_slices` function to generate highlights in two different directions (one where we pass `(passage_slices, entry_slices)` and one where we pass `(entry_slices, passage_slices)`, each being used to generate separate highlights (one for the entry, and one for the passage), it probably makes sense to tweak the names of the input params to be less prescriptive.  I'll do that.

services/QuillLMS/engines/evidence/lib/evidence/evidence/plagiarism_check.rb
anathomical: Good catch.  I've updated that code so that the main Levenshtein loop just assembles the string, rather than a series of indexes that gets combined into a string later.  Benchmarks don't show this speeding things up very much, but it's probably much more readable than it was.

happythenewsad (approved):

cissyxyu (approved):

PR #8756, opened 2022-01-27
lms/cache-teacher-scores-reporting

anathomical:
## WHAT
Update basic Teacher `/scores` caching
## WHY
To reduce database load on commonly-viewed reports.  This particular report is especially prevalent as it's one of the "default" report pages that a lot teachers land on even if they're going somewhere else.
## HOW
Wrap the payload construction code with the new user `classroom`-level caching code

### Notion Card Links
https://www.notion.so/quill/Implement-Caching-on-3-endpoints-07b97e6538134ea090ba0f7fd5df3d8b

PR Checklist | Your Answer
------------ | -------------

### DIFF
--- a/services/QuillLMS/app/controllers/teachers/classroom_manager_controller.rb
+++ b/services/QuillLMS/app/controllers/teachers/classroom_manager_controller.rb
@@ -135,14 +135,27 @@ class Teachers::ClassroomManagerController < ApplicationController
   end
 
   def scores
-    scores = Scorebook::Query.run(params[:classroom_id], params[:current_page], params[:unit_id], params[:begin_date], params[:end_date], current_user.utc_offset)
-
-    last_page = scores.length < 200
-
-    render json: {
-      scores: scores,
-      is_last_page: last_page
+    classroom = Classroom.find(params[:classroom_id])
+    cache_groups = {
+      unit: params[:unit_id],
+      page: params[:current_page],
+      begin: params[:begin_date],
+      end: params[:end_date],
+      offset: current_user.utc_offset
     }
+
+    json = current_user.classroom_cache(classroom, key: 'classroom_manager.teacher_dashboard_metrics', groups: cache_groups) do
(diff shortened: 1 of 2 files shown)


services/QuillLMS/app/controllers/teachers/classroom_manager_controller.rb
```diff
@@ -135,14 +135,27 @@ def getting_started
...
+    cache_groups = {
+      unit: params[:unit_id],
+      page: params[:current_page],
+      begin: params[:begin_date],
+      end: params[:end_date],
+      offset: current_user.utc_offset
+    }
 
-    last_page = scores.length < 200
+    json = current_user.classroom_cache(classroom, key: 'classroom_manager.teacher_dashboard_metrics', groups: cache_groups) do
```
emilia-friedberg: I don't think this key can be right - shouldn't that be for the metrics that the teacher sees on their dashboard?

emilia-friedberg (changes requested):

services/QuillLMS/app/controllers/teachers/classroom_manager_controller.rb
anathomical: Good catch!  That was definitely a copy-paste error on my part.  I've fixed it now.

services/QuillLMS/app/controllers/teachers/classroom_manager_controller.rb
```diff
@@ -135,14 +135,27 @@ def getting_started
...
 
-    last_page = scores.length < 200
+    json = current_user.classroom_cache(classroom, key: 'classroom_manager.scores', groups: cache_groups) do
+      scores = Scorebook::Query.run(params[:classroom_id], params[:current_page], params[:unit_id], params[:begin_date], params[:end_date], current_user.utc_offset)
 
-    render json: {
-      scores: scores,
-      is_last_page: last_page
-    }
+      last_page = scores.length < 200
```
happythenewsad, line 150: nit: this should be `is_last_page` because of type boolean.

happythenewsad (approved):

emilia-friedberg (approved):

PR #8760, opened 2022-01-28
lms/cache-teachers-units-scores-info

anathomical:
## WHAT
Update controller to cache on classroom_unit
## WHY
It takes half the time to load the cached version of this common reporting endpoint
## HOW
1. Wrap the controller payload code in a cache block
2. Add a `rescue` clause to handle cases where `classroom_id` is not present or valid (the tests make it clear that we expect a specific static payload in those cases, nice job, tests!)

### Notion Card Links
https://www.notion.so/quill/Implement-Caching-on-3-endpoints-07b97e6538134ea090ba0f7fd5df3d8b

PR Checklist | Your Answer
------------ | -------------
Have you added and/or updated tests? |  No behavior changes, so no need to update tests

### DIFF
--- a/services/QuillLMS/app/controllers/teachers/units_controller.rb
+++ b/services/QuillLMS/app/controllers/teachers/units_controller.rb
@@ -139,17 +139,31 @@ class Teachers::UnitsController < ApplicationController
   # :activity_id (in url)
   # :classroom_unit_id
   def score_info
-    completed = ActivitySession.where(
-      classroom_unit_id: params[:classroom_unit_id],
-      activity_id: params[:activity_id],
-      is_final_score: true,
-      visible: true
-    )
+    classroom_unit = ClassroomUnit.find(params[:classroom_unit_id])
+    cache_groups = {
+      activity: params[:activity_id]
+    }
 
-    completed_count = completed.count
-    cumulative_score = completed.sum(:percentage) * 100
+    json = current_user.classroom_unit_cache(classroom_unit, key: 'units.score_info' , groups: cache_groups) do
+      completed = ActivitySession.where(
+        classroom_unit_id: params[:classroom_unit_id],
+        activity_id: params[:activity_id],
+        is_final_score: true,
+        visible: true
+      )
 
-    render json: {cumulative_score: cumulative_score, completed_count: completed_count}
+      completed_count = completed.count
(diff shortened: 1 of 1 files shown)


services/QuillLMS/app/controllers/teachers/units_controller.rb
```diff
@@ -139,17 +139,31 @@ def destroy
...
-      activity_id: params[:activity_id],
-      is_final_score: true,
-      visible: true
-    )
+    classroom_unit = ClassroomUnit.find(params[:classroom_unit_id])
+    cache_groups = {
+      activity: params[:activity_id]
+    }
+
+    json = current_user.classroom_unit_cache(classroom_unit, key: 'units.score_info', groups: cache_groups) do
```
brendanshean: For consistency with the other caching PRs, maybe make `key: 'teachers.units.score_info'`?

services/QuillLMS/app/controllers/teachers/units_controller.rb
```diff
@@ -139,17 +139,31 @@ def destroy
   # :activity_id (in url)
   # :classroom_unit_id
   def score_info
-    completed = ActivitySession.where(
-      classroom_unit_id: params[:classroom_unit_id],
-      activity_id: params[:activity_id],
-      is_final_score: true,
-      visible: true
-    )
+    classroom_unit = ClassroomUnit.find(params[:classroom_unit_id])
```
brendanshean: Since the exception clause below is being used for control flow, what about using a `find_by` and guard clause instead:

```
classroom_unit = ClassroomUnit.find_by(params[:classroom_unit_id])
return render json: { cumulative_score: 0, completed_count: 0 } if classroom_unit.nil?
```

brendanshean (approved):
A couple minor comments, but otherwise looks good.  Thank you for caching all these endpoints!

cissyxyu (approved):

services/QuillLMS/app/controllers/teachers/units_controller.rb
anathomical: Yeah, good idea.

services/QuillLMS/app/controllers/teachers/units_controller.rb
anathomical: Not opposed.  Refactored.

PR #8769, opened 2022-01-31
Fix a couple of bugs with Google Student importing

brendanshean:
## WHAT
Fix a bug involving workers that make calls to the Google Classroom when the access token is expired.

## WHY
The workers will fail and then run again and fail, etc, since the access token has expired.

## HOW
Add a check for `user.google_authorized?` to both workers.

### Screenshots
Screen Shot 2022-02-01 at 12 29 25 PM

### Notion Card Links

PR Checklist | Your Answer
------------ | -------------

### DIFF
--- a/services/QuillLMS/app/workers/google_student_importer_worker.rb
+++ b/services/QuillLMS/app/workers/google_student_importer_worker.rb
@@ -8,13 +8,11 @@ class GoogleStudentImporterWorker
 
   def perform(teacher_id, context = "none", selected_classroom_ids=nil)
     teacher = User.find(teacher_id)
+    return unless teacher.google_authorized?
+
     GoogleIntegration::TeacherClassroomsStudentsImporter.run(teacher, selected_classroom_ids)
     PusherTrigger.run(teacher_id, PUSHER_EVENT_CHANNEL, "Google classroom students imported for #{teacher_id}.")
   rescue StandardError => e
-    if Rails.env.development?
-      puts 'ERROR', e
-    else
-      NewRelic::Agent.notice_error(e, context: context)
-    end
+    NewRelic::Agent.notice_error(e, context: context)
   end
 end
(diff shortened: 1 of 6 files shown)


services/QuillLMS/app/workers/google_student_importer_worker.rb
```diff
@@ -8,13 +8,11 @@ class GoogleStudentImporterWorker
 
   def perform(teacher_id, context = "none", selected_classroom_ids=nil)
     teacher = User.find(teacher_id)
+    return unless teacher.google_authorized?
```
dandrabik: Do we know how this state happens: a google import job for the teacher is fired, but the teacher isn't `google_authorized`?

Wondering because:
1. Is there is a bug further upstream where this job shouldn't fire?
2. Should we send a `Pusher` notification for that non-authorized state (we send one for the success state)? If it's actionable and they will be on a page to receive the message, that might be helpful information.

dandrabik (approved):
Thanks for cleaning this up. Approved!

I left a question around potential additional improvements, but those aren't blockers for shipping this.

services/QuillLMS/app/workers/google_student_importer_worker.rb
brendanshean: 1)  That's a good call.  I have it checked for authorization [here](https://github.com/empirical-org/Empirical-Core/blob/develop/services/QuillLMS/app/controllers/auth/google_controller.rb#L48) but I missed the check [here](https://github.com/empirical-org/Empirical-Core/blob/develop/services/QuillLMS/app/controllers/teachers/classroom_manager_controller.rb#L217).  I will fix that.
2) The pusher notification is a great suggestion and a good fallback until we implement the reauthorization UI like we have for clever.  I will add that in.

brendanshean: @dandrabik Re-requesting review due to the new changes.  I've also added a screenshot showing the new pusher notification.

dandrabik (approved):
This looks good, thanks for making these change. I like the pusher notification, I think that's a nice bridge to when we add the full UI.

PR #8770, opened 2022-01-31
Add exclusion filter for benchmarking specs

brendanshean:
## WHAT
Add a metadata tag `benchmarking` that can be applied to performance related specs.

## WHY
We'd like to exclude benchmarking specs from being run in circleci flow without using `xit` labeling.

## HOW
Update the rails_helper with the appropriate [exclusion config](https://relishapp.com/rspec/rspec-core/v/3-10/docs/filtering/exclusion-filters) and then apply the tag to the relevant spec.

### Screenshots

### Notion Card Links

PR Checklist | Your Answer
------------ | -------------

### DIFF
--- a/services/QuillLMS/spec/controllers/api/v1/active_activity_session_controller_spec.rb
+++ b/services/QuillLMS/spec/controllers/api/v1/active_activity_session_controller_spec.rb
@@ -20,12 +20,9 @@ describe Api::V1::ActiveActivitySessionsController, type: :controller do
   end
 
   describe "#update" do
-
-    # rubocop:disable RSpec/Pending, RSpec/ExampleLength
-    xit 'performance testing' do
-      100.times do |i|
-        create(:active_activity_session)
-      end
+    # rubocop:disable RSpec/ExampleLength
+    it 'performance testing', :benchmarking do
+      create_list(:active_activity_session, 100)
 
       data = {
           "updatedAt": 1593741284430,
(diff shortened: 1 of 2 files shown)


services/QuillLMS/spec/controllers/api/v1/active_activity_session_controller_spec.rb
```diff
@@ -341,7 +338,7 @@
         end
       end
     end
-    # rubocop:enable RSpec/Pending, RSpec/ExampleLength
+    # rubocop:enable RSpec/ExampleLength
```
happythenewsad, line 341: I vote to remove this cop, rather than add exceptions. In general, the pros of relaxed linting in testland outweigh the cons.

happythenewsad (approved):

anathomical (approved):

services/QuillLMS/spec/controllers/api/v1/active_activity_session_controller_spec.rb
anathomical: I think I generally agree with the sentiment.  Might be worth excluding our `spec` directories from this cop.

brendanshean: Which cop?

brendanshean: Are we talking about RSpec/ExampleLength or RSpec/Pending?

Currently, in our codebase, this is the sole violation of `RSpec/ExampleLength` due to the test containing 300 lines of data that should be in a separate data file.

As far as `RSpec/Pending` goes, I can't really think of any good reasons to keep pending specs in our codebase.  They basically get committed and then people forget about them and we end up deleting them later.

PR #8771, opened 2022-01-31
Add job to reset Ghost Inspector account

brendanshean:
## WHAT
Add a cron job that resets the Ghost Inspector account at :20 and :50 every hour.

## WHY
At the end of the ghost inspector test  "Student: (1) Assign Starter Diagnostic (2) Complete Starter Diagnostic (3) Delete Starter Diagnostic (v2)", there are a few steps that handle cleanup of the Ghost Inspector account.  
Periodically there are spurious failures with the test preventing the cleanup from happening.  Ultimately, the next run through of the test has previous state associated with the Ghost Inspector account resulting in a continually failing test.  We need to cleanup the account before running the next test.

## HOW
The test runs every 30 minutes at :00 and :30, so I've added a new cron job set to run at :20 and :50 that will cleanup the units associated with the ghost inspector account.  The cleanup adapts the [hide method](https://github.com/empirical-org/Empirical-Core/blob/develop/services/QuillLMS/app/controllers/teachers/units_controller.rb#L126) across all units for that teacher.

### Screenshots

### Notion Card Links

PR Checklist | Your Answer
------------ | -------------

### DIFF
--- a/services/QuillLMS/app/models/cron.rb
+++ b/services/QuillLMS/app/models/cron.rb
@@ -3,6 +3,8 @@
 class Cron
   # Configured in Heroku Scheduler to run every 10 minutes
   def self.interval_10_min
+    run_at_20_minute_mark if Time.now.min == 20
+    run_at_50_minute_mark if Time.now.min == 50
   end
 
   # Configured in Heroku Scheduler to run every hour on the XX:30 mark
@@ -25,22 +27,22 @@ class Cron
     MaterializedViewRefreshWorker.perform_async
     RematchUpdatedQuestionsWorker.perform_async(date.beginning_of_day, date.end_of_day)
 
-    Question::TYPES.each {|type| RefreshQuestionCacheWorker.perform_async(type) }
+    Question::TYPES.each { |type| RefreshQuestionCacheWorker.perform_async(type) }
   end
 
   # Configured in Heroku Scheduler to run at XX:20
-  def self.run_hourly_at_20_minute_mark
+  def self.run_at_20_minute_mark
     ResetGhostInspectorAccountWorker.perform_async
   end
 
   # Configured in Heroku Scheduler to run at XX:50
-  def self.run_hourly_at_50_minute_mark
+  def self.run_at_50_minute_mark
     ResetGhostInspectorAccountWorker.perform_async
   end
 
   def self.run_saturday
     SetImpactMetricsWorker.perform_async
-    UploadLeapReportWorker.perform_async(29087)
+    UploadLeapReportWorker.perform_async(29_087)
     PopulateAllActivityHealthsWorker.perform_async
     DeleteObsoleteActiveActivitySessionsWorker.perform_async
   end
--- a/services/QuillLMS/app/models/cron.rb
+++ b/services/QuillLMS/app/models/cron.rb
@@ -28,8 +28,13 @@ class Cron
     Question::TYPES.each {|type| RefreshQuestionCacheWorker.perform_async(type) }
   end
 
-  # Configured in Heroku Scheduler to run at XX:20 and XX:50 marks every hour
-  def self.run_at_20_and_50_on_the_hour
+  # Configured in Heroku Scheduler to run at XX:20
+  def self.run_hourly_at_20
+    ResetGhostInspectorAccountWorker.perform_async
+  end
+
+  # Configured in Heroku Scheduler to run at XX:50
+  def self.run_hourly_at_50
     ResetGhostInspectorAccountWorker.perform_async
   end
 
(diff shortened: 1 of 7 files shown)


happythenewsad (approved):
Nice fix.

services/QuillLMS/app/models/cron.rb
```diff
@@ -28,6 +28,16 @@ def self.interval_1_day
...
+  # Configured in Heroku Scheduler to run at XX:20
+  def self.run_hourly_at_20_minute_mark
+    ResetGhostInspectorAccountWorker.perform_async
+  end
+
+  # Configured in Heroku Scheduler to run at XX:50
+  def self.run_hourly_at_50_minute_mark
+    ResetGhostInspectorAccountWorker.perform_async
+  end
+
```
anathomical: Note that Heroku Scheduler is configured to trigger rake tasks, so I think you'll need to add corresponding tasks to `lib/tasks/cron.rake` if you want these to be run on that timer.  An alternative might be to add code to the "ten minute" interval that calls these if certain conditions are true (the way we trigger `run_saturday_ inside `interval_1_day`).  Not sure which approach is better.

services/QuillLMS/spec/services/ghost_inspector_account_resetter_spec.rb
```diff
@@ -0,0 +1,29 @@
...
+
+require 'rails_helper'
+
+RSpec.describe GhostInspectorAccountResetter do
+  subject { described_class.run }
+
+  let!(:ghost_inspector_account) { create(:user, email: described_class::DEFAULT_GHOST_INSPECTOR_ACCOUNT_EMAIL) }
+
+  context 'no units associated with ghost_inspector account exist' do
+    it 'does reset anything' do
```
anathomical: I think this is supposed to be `it 'does NOT reset anything' do`

anathomical (changes requested):
Generally speaking, this looks fine to me, but as per one of the comments: I don't think this works as-written.

services/QuillLMS/app/models/cron.rb
brendanshean: Thank you!  I was banging my head on this yesterday since the scheduler was reporting that they were run:
Screen Shot 2022-02-01 at 8 08 52 AM

yet I wasn't seeing anything happening.

services/QuillLMS/app/models/cron.rb
```diff
@@ -3,6 +3,8 @@
 class Cron
   # Configured in Heroku Scheduler to run every 10 minutes
   def self.interval_10_min
+    run_at_20_minute_mark if now.min == 20
+    run_at_50_minute_mark if now.min == 50
```
anathomical: I don't actually know how precise Heroku's "every ten minutes" is.  These might need to be ranges like `if now.min.between?(20, 29)` or something?

brendanshean: That's a very astute point.  I do notice a lag sometimes with stuff kicking off in the logs.

brendanshean: Screen Shot 2022-02-01 at 11 54 25 AM

Your suspicions are confirmed

brendanshean: Also in the heroku ui:
Screen Shot 2022-02-01 at 11 56 01 AM

brendanshean: Also noticed this warning too:
Screen Shot 2022-02-01 at 12 20 20 PM

services/QuillLMS/app/models/cron.rb
```diff
@@ -3,6 +3,8 @@
 class Cron
   # Configured in Heroku Scheduler to run every 10 minutes
   def self.interval_10_min
+    run_at_20_minute_mark if Time.now.min.between?(20, 29)
+    run_at_50_minute_mark if Time.now.min.between?(50, 59)
```
anathomical: You could take or leave this, but I'd be inclined to make sure your `now` is in UTC.  Just in case this somehow runs on servers where local time is in one of those "off by 30 minutes" timezones.

anathomical (approved):

services/QuillLMS/app/models/cron.rb
brendanshean: I'd rather not have to troubleshoot some bizarre time edge case in the future, so I will peg this to UTC.

brendanshean: @anathomical So, the `self.now` method uses a memoized variable which carried through with the different specs.  I've removed the timecop blocks and just used a stub instead.

PR #8779, opened 2022-02-01
Evidence/ exemplars migrations

eadams17:
## WHAT
add migrations for Evidence prompt exemplars

## WHY
we want to show these after a student completes an activity

## HOW
add `first_strong_example` and `second_strong_example` to Evidence prompt

### Screenshots

### Notion Card Links
https://www.notion.so/quill/Update-the-Completion-Page-efe12085e4c4470d921a5f1f095d6eed

PR Checklist | Your Answer
------------ | -------------
Have you added and/or updated tests? |  no-- migration files PR

### DIFF
--- a/services/QuillLMS/client/app/bundles/Evidence/tests/components/studentView/__snapshots__/promptStep.test.tsx.snap
+++ b/services/QuillLMS/client/app/bundles/Evidence/tests/components/studentView/__snapshots__/promptStep.test.tsx.snap
@@ -13,7 +13,7 @@ exports[`PromptStep component active state before any responses have been submit
     Object {
       "conjunction": "because",
       "max_attempts": 5,
-      "max_attempts_feedback": "Nice effort! You made some strong revisions. Here is an example of a strong response. What is similar or different about your response? 
+      "max_attempts_feedback": "Nice effort! You made some strong revisions. Here is an example of a strong response. What is similar or different about your response?
 
  Governments should make voting compulsory because otherwise not everyone will vote.",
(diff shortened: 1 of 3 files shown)


brendanshean (approved):

brendanshean:
If there were a need to add `third_strong_example`, ..., `nth_strong_example`, I would probably throw this in an array of strings attribute, but as is, the two attributes seem fine.

emilia-friedberg (approved):
Edit: just saw Brendan's comment, I think that makes sense. Maybe we should confirm with curriculum that we will likely only ever have two here?

Original comment: I'd leave it up to a backend dev probably but attributes on the `prompts` table with names like `first` and `second` start to feel like a code smell to me... like maybe it would make more sense to have these as a separate table? Or even as an array type? This is definitely the easiest solution so maybe we should go with it for now...

eadams17:
> Edit: just saw Brendan's comment, I think that makes sense. Maybe we should confirm with curriculum that we will likely only ever have two here?
> 
> Original comment: I'd leave it up to a backend dev probably but attributes on the `prompts` table with names like `first` and `second` start to feel like a code smell to me... like maybe it would make more sense to have these as a separate table? Or even as an array type? This is definitely the easiest solution so maybe we should go with it for now...

@emilia-friedberg had a chat with @brendanshean -- I understand the thinking on this, I was following the structure we have with plagiarism feedback (which was decided that we'd only ever want two). I'm checking with curriculum to see what they say before proceeding

PR #8783, opened 2022-02-01
lms/report-downloads-bug

anathomical:
## WHAT
Refactor some of the logic that toggles download links
## WHY
Pretty sure there's some sort of conditional race condition going on.  On some pages, this always component works exactly as intended, but on some pages it always seems to mis-identify the state that the page is in when it checks to see if it should hide the download links.  The updated code consistently works on all pages.
## HOW
1. Move all of our relevant class names into constants
2. Add one new class name to our set of constants
3. Reference the new constants when deciding if the element triggering the global `click` handler instead of copies of the classNames as strings

### Screenshots
Current result of clicking on "Download Report" button
image
Updated result of clicking on "Download Report" button
image

### Notion Card Links
https://www.notion.so/quill/Admin-Dashboard-Reports-Download-Button-Not-Working-e8fdaa83e649429395c1aabf88601228
https://www.notion.so/quill/Download-report-button-does-not-work-on-Concepts-report-043f42744afb41a88d2c9e28f174f428

PR Checklist | Your Answer
------------ | -------------
Have you added and/or updated tests? |  No clear way to update the tests since we don't fully understand what causes the discrepancy in behavior between pages.

### DIFF
--- a/services/QuillLMS/client/app/bundles/Teacher/components/progress_reports/csv_download_for_progress_report.jsx
+++ b/services/QuillLMS/client/app/bundles/Teacher/components/progress_reports/csv_download_for_progress_report.jsx
@@ -11,6 +11,11 @@ function formatData(data) {
   })
 }
 
+const unclickedDownloadBtnClass = "btn button-green"
+const clickedDownloadBtnClass = "quill-button medium primary contained focus-on-light"
+const printBtnClass = "print-button"
+const printImgClass = "print-img"
+
 export default class CSVDownloadForProgressReports extends React.Component {
   constructor(props) {
     super(props)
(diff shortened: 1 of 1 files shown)


emilia-friedberg (approved):

services/QuillLMS/client/app/bundles/Teacher/components/progress_reports/csv_download_for_progress_report.jsx
```diff
@@ -75,7 +80,8 @@ export default class CSVDownloadForProgressReports extends React.Component {
   }
 
   closeDropdownIfOpen(e) {
-    if (this.state.showDropdown && e.target.classList.value !== 'btn button-green' && e.target.classList.value !== 'print-button' && e.target.classList.value !== 'print-img') {
+    const ignoreClasses = [printBtnClass, printImgClass, clickedDownloadBtnClass, unclickedDownloadBtnClass]
```
emilia-friedberg: I think we probably still also want to check for `this.props.className` in here in case one is getting passed down?

anathomical: Good call.  That seems reasonable.

PR #8796, opened 2022-02-03
add support for hints in Evidence

emilia-friedberg:
## WHAT
Add support for "hints" in Evidence.

## WHY
To give students clearer examples of 

## HOW
- Add backend support for the existence of hints (migration/model files/controllers)
- Add frontend internal tool component so that Curriculum team members can add, remove, and edit hints
- Update our feedback controllers and the Go endpoint to correctly return the hint as part of the feedback payload when one exists
- Show the hint on the student frontend when one exists

### Screenshots
Screen Shot 2022-02-07 at 8 38 37 AM

### Notion Card Links
https://www.notion.so/quill/Hint-examples-for-Evidence-508f84ec978d4ebc98660a04e3e09537

PR Checklist | Your Answer
------------ | -------------


services/QuillLMS/client/app/bundles/Staff/components/evidence/configureRules/ruleHint.tsx
```diff
@@ -0,0 +1,89 @@
...
+
+  function handleRemoveHint() {
+    setHint({ ...hint, _destroy: true, })
+  }
+
+  function handleHintAltTextChange(e: TextAreaEvent) {
+    setHint({ ...hint, image_alt_text: e.target.value })
+  }
+
+  function handleDrop(acceptedFiles) {
```
happythenewsad, line 35: Unless I'm missing something, this is an opportunity to use async/await.

services/QuillLMS/engines/evidence/app/models/evidence/hint.rb
```diff
@@ -0,0 +1,20 @@
+# frozen_string_literal: true
+
+module Evidence
+  class Hint < ApplicationRecord
+    belongs_to :rule, inverse_of: :hint
+
+    validates_presence_of :explanation
+    validates_presence_of :image_link
```
happythenewsad, line 8: I suggest calling it `image_name` or something, since `link` could be construed as a fully qualified url. URLs are best managed by the app, not persisted in the DB (so that staging and prod can have full separate environments, or if we need to change the image URL we don't want to backfill the DB).

Tangentially: this PR uses the `Image` class for uploading the images. Have you considered adding the relation `Hint has_one Image`?

services/QuillLMS/app/controllers/cms/images_controller.rb
```diff
@@ -9,6 +9,8 @@ def index
   def create
     i = Image.new
     file = params[:file]
+    split_filename = file.original_filename.split('.')
+    file.original_filename = "#{split_filename[0]}_#{Image.last.id + 1}.#{split_filename[1]}"
```
anathomical: So I think the goal here is to cut off the extension for the file, add a number to guarantee uniqueness, and then add it back.  However, there's an edge case here where the file is named something like "file.name.extension" where your code will change the name to "file_1.name".

A harder to read, but safer version of this code is `split_filename = file.original_filename.split/(.*)\.([^\.]*)$/)`, and then use `file.original_filename = "#{split_filename[1]}_#{Image.last.id + 1}.#{split_filename[2]}"`

services/QuillLMS/client/app/bundles/Staff/components/evidence/configureRules/ruleUniversalAttributes.tsx
```diff
@@ -2,7 +2,7 @@ import * as React from "react";
 import { EditorState, ContentState } from 'draft-js'
 
 import { numericalWordOptions } from '../../../../../constants/evidence';
-import { RuleFeedbackInterface, ClickEvent } from '../../../interfaces/evidenceInterfaces';
+import { RuleFeedbackInterface, ClickEvent, TextAreaEvent,  } from '../../../interfaces/evidenceInterfaces';
```
anathomical: I don't see where this import is used in this component.  Is the import necessary for other reasons?

services/QuillLMS/engines/evidence/app/models/evidence/hint.rb
```diff
@@ -0,0 +1,20 @@
...
+    validates_presence_of :image_alt_text
```
anathomical, line 9: Not sure if this matters, but it might be worth adding some max length validators on these strings just in case?  And you do end up doing that, let's also make sure to add tests for those validators.  (If we decide not to bother, no need to change tests.)

services/evidence/feedback-api-main/src/endpoint/endpoint.go
```diff
@@ -251,6 +253,13 @@ type Highlight struct {
 	Character int `json:"character,omitempty"`
 }
 
+type Hint struct {
+  Id int `json:"id,omitempty"`
+	Explanation string `json:"explanation,omitempty"`
+	Image_link string `json:"image_link,omitempty"`
+  Image_alt_text string `json:"image_alt_text,omitempty"`
```
anathomical: Some weird indentation stuff going on in here.  Go should be enforcing tabs-only, but something seems off here for some reason.

anathomical (approved):
A spacing issue in the Go code that I'd like to see get cleaned up, and a couple of suggestions here and there, but overall this looks like a solid approach to me.  I especially appreciate how much test coverage has been added to make sure it all stays working.

services/QuillLMS/client/app/bundles/Staff/components/evidence/configureRules/ruleHint.tsx
emilia-friedberg: We could use async/await here, but I don't see the benefit in this instance?

services/QuillLMS/client/app/bundles/Staff/components/evidence/configureRules/ruleHint.tsx
```diff
@@ -0,0 +1,89 @@
...
+import { TextEditor } from '../../../../Shared/index';
+import getAuthToken from '../../../../Teacher/components/modules/get_auth_token'
+
+const RuleHint = ({
+  errors,
+  hint,
+  setHint,
+}) => {
+
+  function handleAddHint() {
```
eadams17: this creates an emtpy hint? maybe `handleCreateHint` or `handleHintCreation` might make more sense

eadams17 (approved):

services/QuillLMS/client/app/bundles/Staff/components/evidence/configureRules/ruleHint.tsx
emilia-friedberg: We definitely could use async/await here, but I'm not sure what the benefit would be in this instance.

services/QuillLMS/engines/evidence/app/models/evidence/hint.rb
emilia-friedberg: image_link is, as it stands at the moment, a fully qualified url. I see what you're saying about the benefits of having the app handle the actual linking, but I think if we were going to build out a different system here we'd probably also want to address the other places we use the Image class for image uploads, and that feels outside the scope of this project. The other thing is that from talking to Hannah and Lindsey, my understanding is that they wanted this hint system in place quickly and that it would probably get iterated on further later to not use the static images, so I don't want to do too much fiddling with it here.

PR #8806, opened 2022-02-08
Fix bug involving auto-renew feature not triggering

brendanshean:
## WHAT
There's an issue where users with recurring subscriptions are not being renewed.  We need to add a check that purchasers exist as customers in Stripe's system before attempting charges for renewal.

## WHY
Our `RenewExpiringRecurringSubscriptionsWorker` is called daily.  When it runs, it finds subscriptions where 
1) `expiration <= Date.today`
2) `recurring: true`
3) `de_activated_date: nil`
and then iterates over each subscription, making a API call to stripe to make a [charge](https://github.com/empirical-org/Empirical-Core/blob/develop/services/QuillLMS/app/models/subscription.rb#L295).
 
However, if a particular stripe_customer_id either doesn't exist in stripe or has been deleted (i.e. archived), and exception will be raised an all subsequent renewals for other users are not performed.

## HOW
~~First we should update the worker to set `stripe_customer_id` to nil. Once this PR is merged into develop, then run the one-off rake task that updates these records.~~ After discussing with DD, I am scrapping this approach since we'd still like to keep the `stripe_customer_id` in case we need it for auditing.

For the second attempt, instead I am first checking to see if the user exists in Stripe.  When `Stripe::Customer.retrieve` is called [here](https://github.com/empirical-org/Empirical-Core/pull/8806/files#diff-f12ad37316715e6b76bd333b1ca6ae145da224367875fa5e4bcf76fe601faff1R311), three cases follow:
1) the customer_id exists in stripe (return true)
2) the customer_id exists in stripe but the customer has been archived (returns false)
3) an `Stripe::InvalidRequestError` is thrown that the customer_id does not exist in Stripe (returns false)
In the event that some other exception is raised, we log the exception to NewRelic.

### Screenshots

### Notion Card Links
https://www.notion.so/quill/Auto-Renew-Feature-did-not-trigger-in-teacher-account-673034ba857f48498cc1348d42cca383

PR Checklist | Your Answer
------------ | -------------
Have you added and/or updated tests? |  No.  There is no new functionality here except setting `stripe_customer_id` to nil.  

### DIFF
--- a/services/QuillLMS/app/models/subscription.rb
+++ b/services/QuillLMS/app/models/subscription.rb
@@ -278,19 +278,19 @@ class Subscription < ApplicationRecord
   end
 
   protected def charge_user_for_teacher_premium
-    return unless purchaser&.stripe_customer_id
+    return unless purchaser&.stripe_customer?
 
     Stripe::Charge.create(amount: TEACHER_PRICE, currency: 'usd', customer: purchaser.stripe_customer_id)
   end
 
   protected def charge_user_for_school_premium(school)
-    return unless purchaser&.stripe_customer_id
+    return unless purchaser&.stripe_customer?
 
     Stripe::Charge.create(amount: SCHOOL_FIRST_PURCHASE_PRICE, currency: 'usd', customer: purchaser.stripe_customer_id)
   end
 
   protected def charge_user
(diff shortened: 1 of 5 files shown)


dandrabik (approved):
:+1: Looks good to me!

PR #8819, opened 2022-02-11
Fix Select School for new teachers

cissyxyu:
## WHAT
We are not creating a new `SchoolsUsers` record when teachers sign up, instead we're just initializing the object but not saving it. This means teachers are not getting their schools saved when they sign up for Quill.

## WHY
We want teacher data to be saved and updated.

## HOW
Change the initialize call to a create call so the object gets saved in the database.

### Screenshots
Screen Shot 2022-02-11 at 4 06 30 PM

### Notion Card Links
https://www.notion.so/quill/b417e5b89fe04af89cba62d4198f7031?v=77f294a5b0df466e98633244de7ebfef&p=8afc588b010d44f2b136f7883215a961

PR Checklist | Your Answer
------------ | -------------

### DIFF
--- a/services/QuillLMS/app/controllers/schools_controller.rb
+++ b/services/QuillLMS/app/controllers/schools_controller.rb
@@ -101,8 +101,9 @@ class SchoolsController < ApplicationController
             name: school_params[:school_id_or_type]
           )
         end
-        school_user = SchoolsUsers.find_or_initialize_by(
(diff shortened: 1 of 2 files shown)


anathomical (approved):
Glad we're adding test for the case.  Yay!

eadams17 (approved):

PR #8830, opened 2022-02-15
Fix admin show

happythenewsad:
## WHAT
- adds nil guard so that quill.org/api/v1/app_settings/comprehension/admin_show does not throw exception when `User.email` is nil
- adds additional diagnostic property to the admin json payload: users without emails
- dedupes `AppSetting.user_ids_allow_list` during the rake task upload (duplicates were invisibly accumulating in the database)

## WHY
So that the admin AppSetting tool can become usable
So that unnecessary data does not accumulate in AppSetting.user_ids_allow_list 

### Screenshots

### Notion Card Links
https://www.notion.so/quill/app_settings-admin_show-throws-exception-40f76a502b2a4dc1ad33e917aece999c

PR Checklist | Your Answer
------------ | -------------

### DIFF
--- a/services/QuillLMS/lib/tasks/app_settings.rake
+++ b/services/QuillLMS/lib/tasks/app_settings.rake
@@ -35,7 +35,7 @@ namespace :app_settings do
 
     app_setting = AppSetting.find_by_name!(args[:name])
 
-    app_setting.update!(user_ids_allow_list: app_setting.user_ids_allow_list.uniq.concat(user_ids))
+    app_setting.update!(user_ids_allow_list: app_setting.user_ids_allow_list.concat(user_ids).uniq)
     puts "AppSetting #{app_setting.name} has been updated with user list: #{user_ids}."
   end
 end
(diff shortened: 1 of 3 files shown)


dandrabik (approved):
Thanks for making this change!

cissyxyu (approved):

services/QuillLMS/lib/tasks/app_settings.rake
```diff
@@ -35,7 +35,7 @@ namespace :app_settings do
 
     app_setting = AppSetting.find_by_name!(args[:name])
 
-    app_setting.update!(user_ids_allow_list: app_setting.user_ids_allow_list.concat(user_ids))
+    app_setting.update!(user_ids_allow_list: app_setting.user_ids_allow_list.uniq.concat(user_ids))
```
emilia-friedberg: Does it make sense to do the `.uniq` after the `.concat` so that if the new user_ids include dupes, we still aren't saving duplicates?

emilia-friedberg (approved):
One small question but LGTM overall

services/QuillLMS/lib/tasks/app_settings.rake
happythenewsad: Good idea

PR #8832, opened 2022-02-15
lms/confirm-touch-chains

anathomical:
## WHAT
Update the way we bubble-up touches with custom callbacks
## WHY
To avoid some performance issues, we've had to write some custom callbacks to simulate `touch` behavior on a couple of models.  One of those custom callbacks wasn't being called in all situations.  The hope is that this fixes our cache invalidation process so that we can turn caching back on if this code doesn't cause performance issues of its own.
## HOW
1. Add an `after_touch` hook to `ClassroomUnit` as this is a likely cause of our failure to invalidate caches (`ActivitySession` calls `touch` on `ClassroomUnit`, but that wasn't triggering the `touch_classroom_without_callbacks` function, so we weren't managing to update the `Classroom.updated_at` values which are the key value in our current caching regime).
2. Instead of using `after_touch :save` in `Unit`, call our custom touch function directly.  `save` may trigger other callbacks which we don't want to pay the overhead for, so this is a little bit safer (even though the original way we were doing it didn't seem to be causing issues at the moment).
3. Add tests to our models to ensure that our touch callbacks trigger on `save` AND `touch`

### Notion Card Links
https://www.notion.so/quill/Cache-invalidation-fix-a5902d4aa1f04598926a635949ba098b

PR Checklist | Your Answer
------------ | -------------

### DIFF
--- a/services/QuillLMS/app/models/classroom_unit.rb
+++ b/services/QuillLMS/app/models/classroom_unit.rb
@@ -38,8 +38,9 @@ class ClassroomUnit < ApplicationRecord
   validates :unit, uniqueness: { scope: :classroom }
   before_save :check_for_assign_on_join_and_update_students_array_if_true
   after_save  :hide_appropriate_activity_sessions
-  after_touch :touch_classroom_without_callbacks
-  after_save :touch_classroom_without_callbacks
+  # Using an after_commit hook here because we want to trigger the callback
+  # on save or touch, and touch explicitly bypasses after_save hooks
+  after_commit :touch_classroom_without_callbacks
 
   # this method does not seem to be getting used, but leaving it in for the tests for now
   def assigned_students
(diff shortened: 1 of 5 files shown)


services/QuillLMS/app/models/classroom_unit.rb
```diff
@@ -38,6 +38,7 @@ class ClassroomUnit < ApplicationRecord
   validates :unit, uniqueness: { scope: :classroom }
   before_save :check_for_assign_on_join_and_update_students_array_if_true
   after_save  :hide_appropriate_activity_sessions
+  after_touch :touch_classroom_without_callbacks
```
dandrabik: If we're doing `after_save` and `after_touch`, I think we could just do one `after_commit` instead (might be worth confirming that).

dandrabik (approved):
I left a comment that there might be a way to make this shorter, but otherwise, looks good to me.

services/QuillLMS/app/models/classroom_unit.rb
anathomical: That seems reasonable.  Let me do some testing on that front to confirm.

anathomical: Testing and reading shows that this should work as intended, so I'm making the suggested update.

cissyxyu (approved):

PR #8854, opened 2022-02-22
lms/fix-inconsistent-diagnostic-reports-test

anathomical:
## WHAT
Use `include` instead of `eq` when testing array contents
## WHY
We don't use an explicit sort when retrieving this data, so we can't guarantee the order that the array will return in.  Most of the time, it returns in the order the test expects, but sometimes it doesn't and that triggers spurious failures.
## HOW
Replace `eq` comparisons with `include` in cases where the `eq` is comparing against a multi-element array

### Notion Card Links
https://app.circleci.com/pipelines/github/empirical-org/Empirical-Core/14572/workflows/324ea59e-2017-4aa7-a78c-e5a12bfc323d/jobs/210323

PR Checklist | Your Answer
------------ | -------------
Have you added and/or updated tests? |  Yes.  Only tests have changed
Have you deployed to Staging? | No, this is only a change to tests


emilia-friedberg (approved):

anathomical:
Realized that this was a duplicate fix as another PR I already wrote.  Whoops.

PR #8862, opened 2022-02-23
lms/expose-sidekiq-retry-queue-in-status-controller

anathomical:
## WHAT
Add a "retry" key to the sidekiq queue statuses endpoint
## WHY
So that we can set up synthetic monitors in New Relic (and/or other platforms) to send alerts if the retry queue gets too big
## HOW
Create a new hash that pulls from Sidekiq's `RetrySet` size, then merge that hash with the one we're already taking from Sidekiq's queue stats data.  Return the merged hash as our JSON payload.

PR Checklist | Your Answer
------------ | -------------

### DIFF
--- a/services/QuillLMS/app/controllers/statuses_controller.rb
+++ b/services/QuillLMS/app/controllers/statuses_controller.rb
@@ -33,7 +33,11 @@ class StatusesController < ApplicationController
   end
 
   def sidekiq_queue_length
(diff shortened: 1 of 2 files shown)


brendanshean (approved):
This is great!  I was unaware of this controller.

happythenewsad (approved):

PR #8864, opened 2022-02-23
Fix bug involving the stripe daily auto-renew job

brendanshean:
## WHAT
Fix an issue involving the double billing of users with multiple subscriptions.

## WHY
We'd like the daily auto renew functionality to work as expected and not double charge users.
 
## HOW
The `def self.update_todays_expired_recurring_subscriptions` method gets called every morning via a cron job which checks to see if there are any recurring subscriptions that have expired that haven't been deactivated yet.  It then [attempts](https://github.com/empirical-org/Empirical-Core/blob/develop/services/QuillLMS/app/models/subscription.rb#L268) to renew that subscription (via a charge with Stripe).  However, if the user has an existing active subscription, the system currently will charge the user for a second subscription. 

To fix this, we simply add a few checks.

First, ensure that the subscription has exactly one user.
Then, check if the user has any active subscriptions.   If they do, deactivate the current subscription.  Otherwise, charge stripe and renew the subscription.

### Screenshots

### Notion Card Links

PR Checklist | Your Answer
------------ | -------------

### DIFF
--- a/services/QuillLMS/app/models/subscription.rb
+++ b/services/QuillLMS/app/models/subscription.rb
@@ -262,10 +262,16 @@ class Subscription < ApplicationRecord
     end
   end
 
-
   def self.update_todays_expired_recurring_subscriptions
-    expired_today_or_previously_and_recurring.each do |s|
-      s.update_if_charge_succeeds unless s.users.empty?
+    expired_today_or_previously_and_recurring.each do |subscription|
+      # TODO: Deactivate subscriptions with multiple users
+      next unless subscription.users.count == 1
+
+      if subscription.users.first.subscriptions.active.empty?
+        subscription.update_if_charge_succeeds
+      else
+        subscription.update(de_activated_date: Date.today)
+      end
     end
   end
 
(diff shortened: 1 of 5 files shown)


anathomical (approved):

services/QuillLMS/app/models/subscription.rb
```diff
@@ -262,10 +262,16 @@ def save_if_charge_succeeds(premium_type, school=nil)
     end
   end
 
-
   def self.update_todays_expired_recurring_subscriptions
-    expired_today_or_previously_and_recurring.each do |s|
-      s.update_if_charge_succeeds unless s.users.empty?
+    expired_today_or_previously_and_recurring.each do |subscription|
+      # TODO: Deactivate subscriptions with multiple users
+      next unless subscription.users.count == 1
```
emilia-friedberg, line 268: Do we have a card for this already?

emilia-friedberg (approved):
LGTM! Just wanted to make sure we had a plan to follow-up on the multi-user subscriptions since one we enable direct purchases of school subscriptions that'll be more important.

dandrabik (approved):
Looks good to me. We discussed this yesterday, but just a reminder to manually test the queries on production and confirm that the users pulled for the next few days match our expectations. If anything seems off, let's pull this job out of the cron scheduler (you could also consider doing that now and running this manually to make monitoring easier).

services/QuillLMS/app/models/subscription.rb
brendanshean: That's a good point.  I will make one.

PR #8870, opened 2022-02-24
active_activity_session#update relax params

happythenewsad:
## WHAT
removes strong param checking 

## WHY
to speed up transaction time 

### Screenshots

### Notion Card Links
https://www.notion.so/quill/Product-Board-30f97e4eb01246dbb8264253fb385073?p=8493348417744fd295797d37a1a4d5f4

PR Checklist | Your Answer
------------ | -------------


services/QuillLMS/app/controllers/api/v1/active_activity_sessions_controller.rb
```diff
@@ -8,11 +8,12 @@ def show
   end
 
   def update
+    ActionController::Parameters.permit_all_parameters = true
```
happythenewsad: This allows arbitrary hashes to be accepted from the incoming payload. More details here: https://api.rubyonrails.org/v5.0.0/classes/ActionController/Parameters.html

dandrabik: This is a global setting change, not just for this action, so I think we need a different way to do this.
Screen Shot 2022-02-24 at 3 24 43 PM

dandrabik (changes requested):
We'd have to find a way to do this without changing a global setting (Puma is a multi-threaded server, so setting `ActionController::Parameters.permit_all_parameters = true` would turn that off for ALL controllers).

services/QuillLMS/app/controllers/api/v1/active_activity_sessions_controller.rb
happythenewsad: Thanks for catching that - ~looking into a workaround~ looks like we can use `to_unsafe_hash`

services/QuillLMS/app/controllers/api/v1/active_activity_sessions_controller.rb
```diff
@@ -30,22 +30,10 @@ def update
...
-  end
-
-  private def valid_params
-    params
-      .permit(active_activity_session: {})
-      .require(:active_activity_session)
-      .except(:uid)
+  private def update_params
+    params['active_activity_session']
+      .reject {|key, _| key == 'uid' }
```
dandrabik, line 35: You can use `except` here.
```suggestion
      .except('uid')
```

dandrabik (approved):
Small syntax suggestion, but otherwise looks good to me.

Thanks for making the changes. Interested to see the effect this has on performance.

anathomical (approved):

PR #8873, opened 2022-02-24
Evidence/ optimize tooltip for mobile

eadams17:
## WHAT
add functionality for tooltip to be clickable on mobile (and also on desktop 

## WHY
we want the tooltip to function on mobile

## HOW
conditionally pass an `onClick` prop if the user is on mobile

### Screenshots

### Notion Card Links
https://www.notion.so/quill/Image-credit-tooltip-doesn-t-work-on-mobile-afb46e6ac6a74016b713443a87dc8401

PR Checklist | Your Answer
------------ | -------------


services/QuillLMS/client/app/bundles/Shared/components/shared/tooltip.tsx
```diff
@@ -71,6 +73,7 @@ export class Tooltip extends React.Component<TooltipProps, {}> {
       >
         <span
           className={`${tooltipTriggerTextClass}`}
+          onClick={onMobile() ? this.handleTooltipClick : () => false} // eslint-disable-next-line react/jsx-no-bind
```
emilia-friedberg: If you just do the `onMobile` check inside of `handleTooltipClick` you don't need the bind here.

services/QuillLMS/client/app/bundles/Shared/components/shared/tooltip.tsx
```diff
@@ -27,42 +28,43 @@ export class Tooltip extends React.Component<TooltipProps, {}> {
...
   startTimer() {
     this.timer = setTimeout(this.hideTooltip, 1500)
   }
 
+  handleTooltipClick = (e) => {
+    const { tooltipVisible } = this.state
+
+    if(tooltipVisible) {
+      this.setState({ tooltipVisible: false })
+      this.hideTooltip()
```
emilia-friedberg: Shouldn't we not need to call `this.setState` here since we call it in `hideTooltip`?

services/QuillLMS/client/app/bundles/Shared/components/shared/tooltip.tsx
```diff
@@ -27,42 +28,43 @@ export class Tooltip extends React.Component<TooltipProps, {}> {
...
+    } else {
+      this.setState({ tooltipVisible: true })
+      this.showTooltip()
```
emilia-friedberg: Same here - since we call `setState` in `showTooltip` we shouldn't need it here too.

services/QuillLMS/client/app/bundles/Shared/components/shared/tooltip.tsx
```diff
@@ -27,42 +28,43 @@ export class Tooltip extends React.Component<TooltipProps, {}> {
...
-  }
-
-  hideTooltipOnClick(e) {
-    const { handleClick } = this.props;
-    if (this.tooltip && e.target !== this.tooltip && e.target !== this.tooltipTrigger) {
-      this.hideTooltip()
-    }
-    if(handleClick) {
-      handleClick(e);
-    }
```
emilia-friedberg, line 42: I think there is behavior here that isn't really captured by the new setup - namely, that the tooltip is hidden when you click somewhere on the page that isn't the tooltip (behavior we probably want to keep), and also that if a `handleClick` prop is passed in we do call it when it's clicked. I'm not sure where the latter case is actually in use, so maybe it isn't, but we should confirm that before we remove it.

emilia-friedberg (commented):
I like this general approach but I have a couple of concerns/clean up notes.

services/QuillLMS/client/app/bundles/Shared/components/shared/tooltip.tsx
emilia-friedberg: I actually think the only place it's currently in use is from that `handleImageAttributionClick` in Evidence.

emilia-friedberg (approved):
LGTM!

PR #8883, opened 2022-02-28
Make Quill Evidence section of Teacher Center private

brendanshean:
## WHAT
The Teacher Center Reading Comprehension section can be found via Google. This should be visible to those with users with the Comprehension app setting.

## WHY
We don't want this to be public.

## HOW
Add a check for AppSettings before rendering this particular show_topic page.

### Screenshots

### Notion Card Links
https://www.notion.so/quill/Quill-Evidence-section-of-Teacher-Center-findable-via-Google-search-should-be-private-c4aa1797772140749e9c012be7a25876

PR Checklist | Your Answer
------------ | -------------


services/QuillLMS/app/models/blog_post.rb
```diff
@@ -75,6 +75,7 @@ class BlogPost < ApplicationRecord
   ]
 
   TEACHER_TOPICS = TOPICS.reject { |t| [PRESS_RELEASES, IN_THE_NEWS].include?(t) }
+  PUBLIC_TEACHER_TOPICS = TEACHER_TOPICS - [USING_QUILL_FOR_READING_COMPREHENSION]
```
happythenewsad: How about moving this to the spec file, so that we're not adding app code for the sake of testing?

happythenewsad (changes requested):
Note on your PR text: we should refer to this 'flag' as the 'comprehension app setting', since 'beta flag' refers to a different thing. If you actually meant to apply this logic to what Curriculum refers to as 'beta and gamma users', then use `User.flags` instead of `AppSetting::COMPREHENSION`

brendanshean:
> Note on your PR text: we should refer to this 'flag' as the 'comprehension app setting', since 'beta flag' refers to a different thing. If you actually meant to apply this logic to what Curriculum refers to as 'beta and gamma users', then use `User.flags` instead of `AppSetting::COMPREHENSION`

Good point.  I've adjusted the PR text language.

happythenewsad (approved):

PR #8898, opened 2022-03-04
Fix evidence bug with backspace over prompt in student response

brendanshean:
## WHAT
This PR fixes a bug where a user attempts to delete their response and a portion of the prompt, and then the text box is replaced with the prompt followed by the first letter of the prompt that the user did not delete.

## WHY
This is unexpected behavior.

## HOW
There's an assignment for `indexOfLettersToKeepFromDiffWord` which will return `-1` if the index is not found.  In this case there's no text after the prompt that we need to add therefore we should continue on in the loop.

### Screenshots

### Notion Card Links
https://www.notion.so/quill/Extra-e-added-when-typing-a-student-response-in-Evidence-on-Firefox-5ae25574ac58412d9d2d8bf4d693463c

PR Checklist | Your Answer
------------ | -------------


emilia-friedberg (approved):
Good catch!

PR #8906, opened 2022-03-07
Upgrade bundler to v2

brendanshean:
## WHAT
Upgrade bundler from 1.17.3 to 2.2.33.

## WHY
There's an [incompatibility](https://github.com/rubyjs/mini_racer/issues/218) with M1 ARM architecture and our current version of `mini_racer` and the dependency [libv8](https://github.com/rubyjs/libv8/issues/319#issuecomment-969179803).  One workaround involves using the rosetta2 emulator to build gems using the older intel architecture.  However, this eliminates the benefit of the ARM processor.  By upgrading bundler we don't have to use rosetta2 anymore.  A more standard solution is simply upgrading bundler to v2.

Also, Heroku supports bundler v2 and future upgrades to ruby will be streamlined with using a newer version of bundler.

## HOW
After `gem install -v 2.2.33`, run `bundle update --bundler`.

### Screenshots

### Notion Card Links

PR Checklist | Your Answer
------------ | -------------
Have you added and/or updated tests? |  No.  Configuration change.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -25,6 +25,7 @@ jobs:
           name: Bundle Install if cache isn't present.
           command: |
             cd services/QuillLMS
+            gem install bundler:2.2.33
             # BUNDLE_GEMS__CONTRIBSYS__COM defined in https://circleci.com/gh/empirical-org/Empirical-Core/edit#env-vars
             bundle config --local gems.contribsys.com ${BUNDLE_GEMS__CONTRIBSYS__COM}
             bundle check || bundle install --path vendor/bundle
--- a/services/QuillLMS/Gemfile
+++ b/services/QuillLMS/Gemfile
@@ -199,7 +199,7 @@ group :test do
   gem 'capybara-screenshot'
   gem 'codeclimate-test-reporter', require: nil
   gem 'codecov'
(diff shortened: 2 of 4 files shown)


happythenewsad (approved):

anathomical (changes requested):
I don't actually see any place where this is actually specified.  I presume that that's because `bundler` is sort of outside the framework.  That said, we should at least update `bin/dev/bootstrap.sh` which has a note about pinning an older version due to incompatibility.

brendanshean:
> I don't actually see any place where this is actually specified. I presume that that's because `bundler` is sort of outside the framework. That said, we should at least update `bin/dev/bootstrap.sh` which has a note about pinning an older version due to incompatibility.

My guess is that circleci picks up the bundler version in the [Gemfile.lock](https://github.com/empirical-org/Empirical-Core/blob/develop/services/QuillLMS/Gemfile.lock#L928) but I also couldn't find anything explicit as to why that current version of bundler is there without a `gem install` call.

Good point on the bin/dev/bootstrap.sh
I will update that.

anathomical (approved):

dandrabik (approved):
:+1: Thanks for testing this helpful upgrade!

PR #8913, opened 2022-03-08
Evidence/speed up plagiarism algorithm

anathomical:
## WHAT
Do a simple replacement of `DidYouMean`'s Levenshtein distance function with the one from the `hotwater` gem.
## WHY
Our current algorithm performs fine (if not great) for 95% of cases, but we'd like it to perform better if it can be done easily.  Local benchmarking shows `hotwater` to be about 40x faster, so we should use it.  (Note, testing on Staging suggests that plagiarized request times are reduced by at least 3x.)
## HOW
- Add the gem to our Gemfile
- Require the gem in our plagiarism lib
- Use `hotwater`'s implementation of `levenshtein_distance` whenever we need that distance

### Notion Card Links
https://www.notion.so/quill/Test-alternative-algorithms-of-Fuzzy-Matching-448bea4b93d247588e9f322b03c5fb26

PR Checklist | Your Answer
------------ | -------------


brendanshean (approved):
Nice find on this gem!

PR #8924, opened 2022-03-10
Update user-facing subscription plan names

cissyxyu:
## WHAT
We came up with a new set of official user-facing subscription names to show users on the "My Subscription" page. This PR changes that page to only show these new user-facing names and no longer show the old `account_type` and `subscriptionType`.

## WHY
This will provide a more unified and coherent user experience. It's confusing for users to see a bunch of slightly different plan names on their My Subscription page.

## HOW
Add a file `constants.js` that contains the official user-facing plan names.
Change the front end components to reference only the names in this file when displaying the plan name to users.

### Screenshots
Screen Shot 2022-03-10 at 2 21 56 PM
Screen Shot 2022-03-09 at 1 46 23 PM
Screen Shot 2022-03-08 at 2 20 00 PM
Screen Shot 2022-03-10 at 3 27 53 PM

### Notion Card Links
https://www.notion.so/quill/Display-consistent-plan-names-on-MySubscriptions-page-f8e1225db4d9415cb76aeaf1b2292638

PR Checklist | Your Answer
------------ | -------------


services/QuillLMS/client/app/bundles/Teacher/components/subscriptions/current_subscription.jsx
```diff
@@ -19,18 +20,15 @@ export default class CurrentSubscription extends React.Component {
...
   getCondition() {
     const { subscriptionType, } = this.props
-    switch (subscriptionType) {
-      case 'School':
-        return 'school';
-      case 'School Sponsored':
-        return 'school sponsored';
-      default:
-        return 'other';
+    if (['School Premium', 'District Premium'].includes(subscriptionType)) {
```
emilia-friedberg: We should make these their own constants in that `constants.js` file and then import them here since they're used for logic.

services/QuillLMS/client/app/bundles/Teacher/components/subscriptions/current_subscription.jsx
```diff
@@ -42,7 +40,7 @@ export default class CurrentSubscription extends React.Component {
       return this.editCreditCardElement();
     } else if (subscriptionStatus && subscriptionStatus.payment_method === 'Credit Card') {
       return <span>Credit Card</span>;
-    } else if (subscriptionType === 'School Sponsored' || subscriptionType === 'Trial') {
+    } else if (subscriptionType === 'Teacher Premium Trial') {
```
emilia-friedberg: Same with this and the rest of the places we're changing the display based on a plan name (and not just displaying it directly).

emilia-friedberg (changes requested):
This approach looks good to me, but I think since we are already introducing constants it makes sense to take things as a step further and solidify those constants in one place so that if we change a name later we don't need to go hunting through all these files.

services/QuillLMS/client/app/bundles/Teacher/components/subscriptions/__tests__/__snapshots__/current_subscription.test.jsx.snap
```diff
@@ -87,7 +87,23 @@ exports[`CurrentSubscription container when there is a current subscription shou
             >
               <div>
                 <_default
-                  content="School Paid"
+                  content={
+                    <span>
+                      School Premium
+                      <Tooltip
+                        tooltipText="You have a School Paid subscription"
```
emilia-friedberg, line 94: Don't we want this to say School Premium now?

cissyxyu: We actually want the tooltip to still display the account type. I think the idea is for Quill staff to more easily investigate and debug people's subscriptions.

emilia-friedberg (approved):
Awesome work, Cissy, thanks for cleaning that up.

brendanshean (approved):
Nice work on cleaning this up Cissy!

PR #8930, opened 2022-03-11
Ruby Feedback Endpoint

dandrabik:
## WHAT
Porting the Go endpoint to Ruby as a backup plan and potential speed improvement.
## WHY
We aren't currently using the async properties of Go due to some memory leakage, so it might be worth porting that endpoint to Ruby to cut down on dependencies.
## HOW
I went in with the goal of not touching any of the existing code. I created a base class called `Check::Base` and a child class for each check, e.g. `Check::Plagiarism`. Each child class is responsible for defining a single method `run` that populates a `response` variable. Everything else is boiler plate from the base class.

The module method `Check.run_all` has all of the logic.

For the individual checks, I just copied the logic from the current controller endpoints, so (for now), there is some duplication. That would be easy to clean up if we so decide.  I copied the tests from each endpoint to a `post :create` test and changed the constant in the create endpoint to only do that individual check. (To gain confidence that it's a direct port).

To store the `FeedbackHistory` I created a seam `Evidence.feedback_history_class` that has one required method `.save_feedback`.

I aimed for feature parity with the Go endpoint, so didn't add any additional features. One I considered was enforcing a timeout on each check (we've discussed this in the past).

### Notion Card Links
[(Please provide links to any relevant Notion card(s) relevant to this PR.)](https://www.notion.so/quill/Ruby-Endpoint-for-Feedback-Port-from-Go-97187ac76e514b7daa63aa24eed069a9)

PR Checklist | Your Answer
------------ | -------------
Have you added and/or updated tests? |  'YES'. I might write some more, but there's decent coverage here.


services/QuillLMS/engines/evidence/lib/evidence/evidence/check/regex_post_topic.rb
```diff
@@ -0,0 +1,12 @@
+# frozen_string_literal: true
+
+module Evidence
+  class Check::RegexPostTopic < Check::Base
```
dandrabik, line 4: Note, the three `Check::Regex...` classes are all the same except for they pass in a different rule type. I could have subclassed them, but I wasn't convinced that would be cleaner given these are such small classes.

services/QuillLMS/engines/evidence/lib/evidence/evidence/check/base.rb
```diff
@@ -0,0 +1,46 @@
...
+    end
+
+    # run method of subclass should populate @response with the API response
+    # it can optionally populate @error
+    def run
+      raise NotImplementedError
+    end
+
+    def self.run(entry, prompt, previous_feedback)
+      new(entry, prompt, previous_feedback).tap do |check|
```
anathomical, line 21: Oh, didn't know about `tap`.  Neat method.

services/QuillLMS/engines/evidence/lib/evidence/evidence/check/plagiarism.rb
```diff
@@ -0,0 +1,31 @@
...
+        @response = plagiarism_check.feedback_object
+      end
+    end
+
+    # rubocop:disable Metrics/CyclomaticComplexity
+    private def get_plagiarism_feedback_from_previous_feedback(prev, rule)
+      previous_plagiarism = prev.select {|f| f["feedback_type"] == Evidence::Rule::TYPE_PLAGIARISM && f["optimal"] == false }
+      feedbacks = rule&.feedbacks
+      previous_plagiarism.empty? ? feedbacks&.find_by(order: 0)&.text : feedbacks&.find_by(order: 1)&.text
+    end
```
anathomical, line 28: I think you could shortcut this by using `rule.determine_feedback_from_history(prev)`.  In fact, you might be able to use that look-up more generally as part of the `Check::Base` class since it's handles any number of `Feedback`s that might be attached to a `Rule` as long as there's at least one, and thus should work for all sub-classes.

services/QuillLMS/spec/models/feedback_history_spec.rb
```diff
@@ -138,6 +138,44 @@
...
+        optimal: false,
+        highlight: [{text: 'some', type: 'entry', category: 'grammar', character: 0}],
+        labels: [],
+        response_id: 'abc123',
+        hint: 'a hint'
+      }
+    }
+
+    it 'should store the data properly' do
+      feedback = FeedbackHistory.save_feedback(feedback_hash, entry, 99, session_uid, 5)
```
anathomical: Purely personal preference, but if I'm going to pass a value into a function and then `expect` that value anywhere on the other side, I like to make it a variable.  Might be worth explicitly setting `prompt_id` and `attempt` variables here?  No strong preference on making any changes here.

services/QuillLMS/engines/evidence/spec/controllers/evidence/feedback_controller_spec.rb
```diff
@@ -23,6 +24,287 @@ module Evidence
...
 
+    describe '#create' do
+
+      let(:feedback) { double('feedback', response: {key1: 'some value'} ) }
+
+      it "should call Check.run_all, save history, and return feedback.response" do
+        expect(Check).to receive(:run_all).with(entry, prompt, []).and_return(feedback)
+        expect(Evidence.feedback_history_class).to receive(:save_feedback).with(feedback, entry, prompt.id, 99, 3)
+
+        post :create, params: {entry: entry, prompt_id: prompt.id, session_id: 99, previous_feedback: ([]), attempt: 3 }, as: :json
```
anathomical: I've mentioned this elsewhere, but my personal privilege is that if you're `expect`ing a value that you're passing into a function, it should be a variable.  Might be worth pulling out `session_id` and `attempt` here?

services/QuillLMS/engines/evidence/spec/controllers/evidence/feedback_controller_spec.rb
```diff
@@ -23,6 +24,287 @@ module Evidence
...
+          parsed_response = JSON.parse(response.body)
+          expect(parsed_response["optimal"]).to be false
+          expect(second_feedback.text).to(eq(parsed_response["feedback"]))
+        end
+      end
+
+      context 'prefilter test' do
+        let!(:profanity_rule) do
+          create(:evidence_rule, rule_type: 'prefilter', uid: 'fdee458a-f017-4f9a-a7d4-a72d1143abeb')
+        end
```
anathomical: I don't think we need to specify the `uid` in this `create` call since the factory should provide one if unspecified. 
 Also, pure style thing, but does this need to be a multi-line block?

services/QuillLMS/engines/evidence/spec/controllers/evidence/feedback_controller_spec.rb
```diff
@@ -23,6 +24,287 @@ module Evidence
...
+
+        before do
+          stub_const("Evidence::Check::ALL_CHECKS", [Check::Prefilter])
+        end
+
+        context 'profanity violation' do
+          it 'should return correct rule uid' do
+            post :create, params: {
+              entry: "nero was an ahole",
```
anathomical, line 161: Love this example.  Too true.

services/QuillLMS/engines/evidence/spec/controllers/evidence/feedback_controller_spec.rb
```diff
@@ -23,6 +24,287 @@ module Evidence
...
+            post :create, params: {
+              entry: "the quick brown fox",
+              prompt_id: prompt.id,
+              session_id: 1,
+              previous_feedback: ([])
+            }, :as => :json
+
+            expect(response.status).to eq 200
+            parsed_response = JSON.parse(response.body)
+            expect(parsed_response).to eq({})
```
anathomical: This is not a comment on this particular test, but I think this is the only case we test that confirms that the controller returns `{}` even if all the checks error or fail somehow.  Might be worth having an explicit case for the "fallback" case rather than just having an incidental test for that case inside the "prefilter" context.

services/QuillLMS/engines/evidence/spec/controllers/evidence/feedback_controller_spec.rb
```diff
@@ -23,6 +24,287 @@ module Evidence
...
+              'type': 'response',
+              'text': 'someText',
+              'character': 0
+            }]
+          }
+        end
+
+        let!(:grammar_rule) do
+          create(:evidence_rule, uid: example_rule_uid, optimal: false, concept_uid: 'xyz')
+        end
```
anathomical: No objection to it, but does this need to be a multi-line block?

services/QuillLMS/engines/evidence/spec/controllers/evidence/feedback_controller_spec.rb
```diff
@@ -23,6 +24,287 @@ module Evidence
...
+          )
+          allow_any_instance_of(Grammar::Client).to receive(:post).and_return(client_response)
+        end
+
+        it 'should return a valid json response' do
+          stub_const("Evidence::Check::ALL_CHECKS", [Check::Grammar])
+          post :create, params: incoming_payload, as: :json
+
+          expect(JSON.parse(response.body)).to eq({
+            'concept_uid' => 'xyz',
```
anathomical, line 211: Is it worth referencing this as `grammar_rule.concept_uid` in case that value ever changes for some reason?

services/QuillLMS/engines/evidence/spec/controllers/evidence/feedback_controller_spec.rb
```diff
@@ -23,6 +24,287 @@ module Evidence
...
+            'feedback' => nil,
+            'feedback_type' => 'grammar',
+            'optimal' => false,
+            'response_id' => '0',
+            'highlight' => [
+              { 'type' => 'response', 'text' => 'someText', 'character' => 0 }
```
anathomical, line 217: Is it worth referencing this whole thing as `client_response['highlight']` in case those values change somehow?

services/QuillLMS/engines/evidence/spec/controllers/evidence/feedback_controller_spec.rb
```diff
@@ -23,6 +24,287 @@ module Evidence
...
+            { example_error => example_rule_uid }
+          )
+          allow_any_instance_of(Opinion::Client).to receive(:post).and_return(client_response)
+        end
+
+        it 'should return a valid json response' do
+          post :create, params: incoming_payload, as: :json
+
+          expect(JSON.parse(response.body)).to eq({
+            'concept_uid' => 'xyz',
```
anathomical, line 265: Same note as with grammar: should we reference `opinion_rule.concept_uid` here?

services/QuillLMS/engines/evidence/spec/controllers/evidence/feedback_controller_spec.rb
```diff
@@ -23,6 +24,287 @@ module Evidence
...
+            'feedback' => nil,
+            'feedback_type' => 'opinion',
+            'optimal' => false,
+            'response_id' => '0',
+            'highlight' => [
+              { 'type' => 'response', 'text' => 'someText', 'character' => 0 }
```
anathomical, line 271: Same note as with grammar: should we reference `client_response['highlight']` here?

services/QuillLMS/engines/evidence/spec/controllers/evidence/feedback_controller_spec.rb
```diff
@@ -23,6 +24,287 @@ module Evidence
...
+          before do
+            create(:evidence_feedback, rule_id: opinion_rule.id, text: 'lorem ipsum')
+          end
+
+          it 'should return a valid json response' do
+            post :create, params: incoming_payload, as: :json
+
+            expect(JSON.parse(response.body)).to eq({
+              'concept_uid' => 'xyz',
+              'feedback' => 'lorem ipsum',
```
anathomical, line 289: Maybe we should wrap the creation of `evidence_feedback` in a `let` so we can reference `evidence_feedback.text` here?

anathomical (approved):
Mostly comments on syntax and style with one specific call-out on how we might handle `Rule`s with multiple `Feedback`s (specifically in `Check::Plagiarism` here).  The extensive test coverage is great and leads me to believe that even if you don't refactor the process of finding `Feedback` in cases where a `Rule` has more than one available, this code should still be okay to ship as-is.

services/QuillLMS/engines/evidence/lib/evidence/evidence/check/plagiarism.rb
dandrabik: So, I'll add this to the description, but I think I could have been more clear on where some of this code came from:

For this PR, I copied the existing code from each of the check endpoints. Then I copied the tests from each endpoint to a `post :create` test and changed the constant in the `create` endpoint to only do that individual check. (To gain confidence that it's a direct port).

I don't disagree that there might be a refactor here, and I'd love feedback on this, but I think I'd rather refactor the endpoints in a separate PR, rather than do this move and refactor all at once.

services/QuillLMS/engines/evidence/spec/controllers/evidence/feedback_controller_spec.rb
dandrabik: Good point, I'll just remove this test and add one for the general case.

services/QuillLMS/engines/evidence/spec/controllers/evidence/feedback_controller_spec.rb
dandrabik: So, since I'm copying current tests, I'd want to avoid refactoring those tests at the same time. So I'll change it to a single line, but for now, will leave the test(s) as is.

services/QuillLMS/engines/evidence/spec/controllers/evidence/feedback_controller_spec.rb
dandrabik: For this PR, I'd want to avoid refactoring any of the test substance, just because it's a direct port.

services/QuillLMS/engines/evidence/spec/controllers/evidence/feedback_controller_spec.rb
dandrabik: For this PR, I'd want to avoid refactoring any of the test substance, just because it's a direct port.

services/QuillLMS/engines/evidence/spec/controllers/evidence/feedback_controller_spec.rb
dandrabik: For this PR, I'd want to avoid refactoring any of the test substance, just because it's a direct port.

services/QuillLMS/spec/models/feedback_history_spec.rb
dandrabik: That's fair, will adjust.

happythenewsad (commented):
Looks solid at first pass. You may want to document somewhere that we need to touch

https://github.com/empirical-org/Empirical-Core/blob/develop/services/QuillLMS/client/app/bundles/Evidence/actions/session.ts#L174

to toggle this on.

What are the next intended steps after this code is merged? Run load tests in prod?

cissyxyu (approved):
Looks good to me!

dandrabik:
> Looks solid at first pass. You may want to document somewhere that we need to touch
> 
> https://github.com/empirical-org/Empirical-Core/blob/develop/services/QuillLMS/client/app/bundles/Evidence/actions/session.ts#L174
> 
> to toggle this on.
> 
> What are the next intended steps after this code is merged? Run load tests in prod?

@happythenewsad The plan would be to deploy this for more testing, but not adjust the `GO_FANOUT_URL` yet. Once we do more testing (load testing, performance testing, acceptance testing) we'll make the decision whether to use this endpoint or the Go endpoint.

services/QuillLMS/engines/evidence/spec/controllers/evidence/feedback_controller_spec.rb
dandrabik: As referenced above, for now, I'm avoiding editing existing tests.

services/QuillLMS/engines/evidence/lib/evidence/evidence/check.rb
```diff
@@ -0,0 +1,51 @@
...
+      Grammar,
+      Spelling,
+      RegexTypo
+    ]
+
+    FALLBACK_RESPONSE = {
+      feedback: "Thank you for your response.",
+      feedback_type: "autoML",
+      optimal: true,
+    }
```
anathomical, line 21: I like the fallback feedback here.  Might make sense in a later pass to add a new `feedback_type` for fallback to differentiate it from legitimate autoML feedback, but that can wait.

services/QuillLMS/engines/evidence/lib/evidence/evidence/check.rb
```diff
@@ -0,0 +1,51 @@
...
+
+    def self.feedback(entry, prompt, previous_feedback)
```
anathomical: Pure style question, but does this make more sense as `get_feedback` or `fetch_feedback`?  If a method might create significant work/delay in a workstream, I like to give it a name that implies that it will do some work.  Not critical, just a preference.

anathomical (approved):
One comment on method naming, but this looks like it's good to go to me.

services/QuillLMS/engines/evidence/lib/evidence/evidence/check.rb
dandrabik: Yeah, this is a direct copy from the Go endpoint, but agreed long-term we might want to differentiate.

services/QuillLMS/engines/evidence/lib/evidence/evidence/check.rb
dandrabik: I was debating using `get_feedback` here, and given you would find that clearer, I've changed the name to that.

PR #8931, opened 2022-03-11
evidence/addressable-version-bump

anathomical:
## WHAT
Bump Addressable version in Evidence per Dependabot
## WHY
To stay up to date on our dependencies
## HOW
Update minimum `addressable` version in Gemfile

PR Checklist | Your Answer
------------ | -------------
Have you added and/or updated tests? |  No tests for dependencies

### DIFF
--- a/services/QuillLMS/engines/evidence/Gemfile.lock
+++ b/services/QuillLMS/engines/evidence/Gemfile.lock
@@ -48,7 +48,7 @@ GEM
       i18n (>= 0.7, < 2)
       minitest (~> 5.1)
       tzinfo (~> 1.1)
-    addressable (2.7.0)
+    addressable (2.8.0)
       public_suffix (>= 2.0.2, < 5.0)
     arel (8.0.0)
     builder (3.2.4)


happythenewsad (approved):
any reason not to specify 
`gem 'addressable', '~> 2.8.0'`

in `Gemfile`?

anathomical:
> any reason not to specify `gem 'addressable', '~> 2.8.0'`
> 
> in `Gemfile`?

I didn't specify it there because `addressable` is not a direct dependency (we don't use it in our code), but a sub-dependency (of `webmock` and `signet`).  So it didn't feel appropriate to add it to our Gemfile.  I could be convinced to do otherwise, though?

happythenewsad:
> > any reason not to specify `gem 'addressable', '~> 2.8.0'`
> > in `Gemfile`?
> 
> I didn't specify it there because `addressable` is not a direct dependency (we don't use it in our code), but a sub-dependency (of `webmock` and `signet`). So it didn't feel appropriate to add it to our Gemfile. I could be convinced to do otherwise, though?

I asked because it's currently a direct dependency in the LMS Gemfile, but not the Evidence Gemfile: https://github.com/empirical-org/Empirical-Core/blob/develop/services/QuillLMS/Gemfile#L85

I just ran LMS specs after removing that line, and they pass. How about removing the direct dependency while you're at it?

anathomical:
> > > any reason not to specify `gem 'addressable', '~> 2.8.0'`
> > > in `Gemfile`?
> > 
> > 
> > I didn't specify it there because `addressable` is not a direct dependency (we don't use it in our code), but a sub-dependency (of `webmock` and `signet`). So it didn't feel appropriate to add it to our Gemfile. I could be convinced to do otherwise, though?
> 
> I asked because it's currently a direct dependency in the LMS Gemfile, but not the Evidence Gemfile: https://github.com/empirical-org/Empirical-Core/blob/develop/services/QuillLMS/Gemfile#L85
> 
> I just ran LMS specs after removing that line, and they pass. How about removing the direct dependency while you're at it?

Specs may pass, but the `Addressable` library is used in the `Activity` model in the base LMS application.  It is therefor a real dependency of the LMS.  It is also a sub-dependency, so it gets installed whether we declare it directly or not, but it's good manners to declare code you actually use directly as a direct dependency, I think.

cissyxyu (approved):

happythenewsad:
> > > > any reason not to specify `gem 'addressable', '~> 2.8.0'`
> > > > in `Gemfile`?
> > > 
> > > 
> > > I didn't specify it there because `addressable` is not a direct dependency (we don't use it in our code), but a sub-dependency (of `webmock` and `signet`). So it didn't feel appropriate to add it to our Gemfile. I could be convinced to do otherwise, though?
> > 
> > 
> > I asked because it's currently a direct dependency in the LMS Gemfile, but not the Evidence Gemfile: https://github.com/empirical-org/Empirical-Core/blob/develop/services/QuillLMS/Gemfile#L85
> > I just ran LMS specs after removing that line, and they pass. How about removing the direct dependency while you're at it?
> 
> Specs may pass, but the `Addressable` library is used in the `Activity` model in the base LMS application. It is therefor a real dependency of the LMS. It is also a sub-dependency, so it gets installed whether we declare it directly or not, but it's good manners to declare code you actually use directly as a direct dependency, I think.

Thanks for the explanation - carry on!

PR #8941, opened 2022-03-15
Create Plan model

brendanshean:
## WHAT
As part of the [Stripe Schema RFC](https://www.notion.so/quill/Stripe-Schema-RFC-96821751a3ad4205b497ea32c1373b79) this commit establishes the `Plan` model which will be added as a foreign key to Subscriptions object.

For greater transparency as well as auditing purposes, the Plan object is readonly since we don't want the Plans to be updated or destroyed after creation.

## WHY
Each plan instance will capture relevant details associated with that plan will streamline the integration of plan details in the UI.  In the case of the two Stripe-related plans, we'll also associate the `price_id` found on Stripe with the respective object.

## HOW
1. `rails generate migration Plan  [attributes]`
2.  Add validations to the model. 
2. Using the Stripe CLI creating products and prices for teacher and school premium plans, noting the returned `price_id`.
4. Run the rake task the seeds these plans using the price_id found in the previous step.

### Screenshots

### Notion Card Links

PR Checklist | Your Answer
------------ | -------------

### DIFF
--- a/services/QuillLMS/app/models/plan.rb
+++ b/services/QuillLMS/app/models/plan.rb
@@ -10,7 +10,7 @@
 #  interval        :string
 #  interval_count  :integer
 #  name            :string           not null
-#  paid            :boolean          not null
+#  price           :integer          default(0)
 #  created_at      :datetime         not null
 #  updated_at      :datetime         not null
 #  stripe_price_id :string
@@ -33,6 +33,7 @@ class Plan < ApplicationRecord
   ]
 
   validates :name, presence: true, uniqueness: true
+  validates :price, numericality: { greater_than_or_equal_to: 0 }
   validates :audience, presence: true, inclusion: { in: AUDIENCE_TYPES }
   validates :interval, presence: true, inclusion: { in: INTERVAL_TYPES }
   validates :interval_count, numericality: { greater_than_or_equal_to: 0 }
--- a/services/QuillLMS/app/models/plan.rb
+++ b/services/QuillLMS/app/models/plan.rb
@@ -32,6 +32,8 @@ class Plan < ApplicationRecord
     DAILY_INTERVAL_TYPE = 'daily'
   ]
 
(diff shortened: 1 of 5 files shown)


emilia-friedberg (approved):
This looks great to me overall. My only concern is that I can foresee a case where we'd want to change a display name without having to create a new plan entirely and move all of the existing ones over to it. Is there a way to make that attribute editable while the rest of it is readonly?

brendanshean:
> This looks great to me overall. My only concern is that I can foresee a case where we'd want to change a display name without having to create a new plan entirely and move all of the existing ones over to it. Is there a way to make that attribute editable while the rest of it is readonly?

That's a good question and one that I was also wrangling with @dandrabik yesterday.  I think the copy for `display_name` for Plan instances much like the `name` is unlikely to change much (if at all), but in the very unlikely event that we need to change it, one possible way around the readonly protection is to use an `update_all` where you restrict the relation to the set of records you want to update.  

For example, let's say you have the following plans:

Plan 
id: 1
name: 'Teacher Paid'
display_name: 'Teacher Premium'

Plan
id: 2
name: 'School Paid'
display_name

`Plan.find(1).update(name: 'Teacher Paid v2')` will raise a `ActiveRecord::ReadOnlyRecord: Plan is marked as readonly`

However, if I make it `Plan.where(id: 1).update_all(name: 'Teacher Paid v2')` will [skip](https://api.rubyonrails.org/v5.2/classes/ActiveRecord/Relation.html#method-i-update_all) callbacks and validations which includes the `readonly` tag.

brendanshean:
@emilia-friedberg Another option would be to add all but `display_name` as  [attr_readonly](https://api.rubyonrails.org/classes/ActiveRecord/ReadonlyAttributes/ClassMethods.html#method-i-attr_readonly). 
That's probably what we want.

emilia-friedberg (approved):

cissyxyu (approved):

PR #8943, opened 2022-03-15
Cache `/teachers/classrooms/classrooms_i_teach`

anathomical:
## WHAT
Cache `/teachers/classrooms/classrooms_i_teach` payload
## WHY
In our effort to cache all of our reporting endpoints to reduce db load
## HOW
Wrap the code that generates the payload in a cache call at the all classrooms level

### Notion Card Links
https://www.notion.so/quill/Turn-Caching-Back-On-4bf7669d517443909640387fbafdc958

PR Checklist | Your Answer
------------ | -------------

### DIFF
--- a/services/QuillLMS/app/controllers/teachers/classrooms_controller.rb
+++ b/services/QuillLMS/app/controllers/teachers/classrooms_controller.rb
@@ -31,8 +31,11 @@ class Teachers::ClassroomsController < ApplicationController
   end
 
   def classrooms_i_teach
-    @classrooms = current_user.classrooms_i_teach
-    render json: @classrooms.sort_by { |c| c[:update_at] }, each_serializer: ClassroomSerializer
+    data = current_user.all_classrooms_cache(key: 'teachers.classrooms.classrooms_i_teach') do
+      @classrooms = current_user.classrooms_i_teach
+      @classrooms.sort_by { |c| c[:update_at] }
+    end
+    render json: data, each_serializer: ClassroomSerializer
   end
 
   def regenerate_code
--- a/services/QuillLMS/app/controllers/teachers/classrooms_controller.rb
+++ b/services/QuillLMS/app/controllers/teachers/classrooms_controller.rb
@@ -160,7 +160,6 @@ class Teachers::ClassroomsController < ApplicationController
   private def fetch_classrooms_i_teach_cache
     current_user.all_classrooms_cache(key: 'teachers.classrooms.classrooms_i_teach') do
       @classrooms = current_user.classrooms_i_teach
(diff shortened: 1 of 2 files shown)


services/QuillLMS/app/controllers/teachers/classrooms_controller.rb
```diff
@@ -158,6 +157,13 @@ def transfer_ownership
     render json: {}
   end
 
+  private def fetch_classrooms_i_teach_cache
+    current_user.all_classrooms_cache(key: 'teachers.classrooms.classrooms_i_teach') do
+      @classrooms = current_user.classrooms_i_teach
+      @classrooms.sort_by { |c| c[:update_at] }
```
brendanshean: I'm guessing this is an older bug, but should this be `:updated_at`?

services/QuillLMS/spec/controllers/teachers/classrooms_controller_spec.rb
```diff
@@ -316,14 +316,25 @@
     let(:classroom) { create(:classroom) }
 
     before do
-      allow(teacher).to receive(:classrooms_i_teach) { [classroom] }
+      allow(teacher).to receive(:classrooms_i_teach).once.and_return([classroom])
```
brendanshean, line 319: This is a very clever approach to testing the caching.

brendanshean (approved):
There's a minor typo that should be fixed but nothing that should hold this up.  Nice work!

eadams17 (approved):

services/QuillLMS/app/controllers/teachers/classrooms_controller.rb
anathomical: Yeah, you can see that this code used to live on line 35, and I moved it.  While this probably should be fixed, I'm hesitant to actually do so in case there are now expectations on the current (probably broken) behavior.  Thoughts?

brendanshean: That's a good point. What about removing the sort_by altogether.

brendanshean: The [docs]( https://ruby-doc.org/core-2.7.0/Enumerable.html#method-i-sort_by) mention that `sort_by` maps each value in the array to the block, but if the mappings are equal, the result is unpredictable.  Trying it out in the console a few times seems to return the original array.

anathomical: Seems reasonable.  I'll pull the sort and make sure tests pass, and we can cross our fingers that there's no weird gotchas in there.

PR #8945, opened 2022-03-15
Evidence/ add additional logic for slow feedback response times

eadams17:
## WHAT
update feedback loading state for Evidence in cases where feedback takes too long to load

## WHY
we want to show some messaging so that users won't be confused if feedback response time lags

## HOW
add additional logic to display an error message if feedback takes more than 30 seconds to load

### Screenshots
Screen Shot 2022-03-15 at 5 00 12 PM
Screen Shot 2022-03-15 at 5 00 24 PM
Screen Shot 2022-03-15 at 5 00 47 PM

### Notion Card Links
https://www.notion.so/quill/Feedback-Not-Loading-Message-7c21b367bbd34526b089baa8f9672f7c

PR Checklist | Your Answer
------------ | -------------


services/QuillLMS/client/app/bundles/Evidence/components/studentView/container.tsx
```diff
@@ -481,13 +481,14 @@ export const StudentViewContainer = ({ dispatch, session, isTurk, location, acti
       const innerElements = node.children.map((n, i) => convertNodeToElement(n, i, transformMarkTags))
       const stringifiedInnerElements = node.children.map(n => n.data ? n.data : n.children[0].data).join('')
       let className = ''
+      const key = `${Math.random()}`;
```
emilia-friedberg, line 484: Having a random key defeats the purpose of a key, because this will be regenerated on every rerender and then the key will change so React will have to rerender the components. I'd key it off the contents or maybe the index.

services/QuillLMS/client/app/bundles/Evidence/components/studentView/promptStep.tsx
```diff
@@ -53,20 +58,35 @@ export class PromptStep extends React.Component<PromptStepProps, PromptStepState
       customFeedback: null,
       customFeedbackKey: null,
       responseOverCharacterLimit: false,
-      submissionTime: null
+      submissionTime: 0,
+      failedToLoadFeedback: false,
+      currentTime: 0
```
emilia-friedberg: I think `currentTime` is a confusing name here because I expect the `currentTime` to be whatever `new Date()).getTime()` returns at a given time (ie, not a stored value). Why didn't it work to just calculate the `currentTime` the way you were before?

eadams17: @emilia-friedberg the component was only updating when the mouse was moving or a keyboard event fired, so if the user just clicks submit and does nothing after, the time wasn't getting calculated. What do you think would be a better variable name? I can update it, just wasn't sure of a better name because at the time of the check, we are checking the current time against submission time

emilia-friedberg (approved):

PR #8954, opened 2022-03-17
Bump redis and sidekiq LMS versions for Dependabot

cissyxyu:
## WHAT
Bump the versions on these packages because we have a high severity security issue on the older version.

## WHY
To keep our app secure and up to date.

## HOW
Bump these versions in the Gemfile and Gemfile.lock files.

### Screenshots

### Notion Card Links
https://www.notion.so/quill/Update-LMS-Sidekiq-161905d0bd3e461f91a294286e3d523d

PR Checklist | Your Answer
------------ | -------------

### DIFF
--- a/services/QuillLMS/spec/models/unit_template_spec.rb
+++ b/services/QuillLMS/spec/models/unit_template_spec.rb
@@ -195,23 +195,23 @@ describe UnitTemplate, redis: true, type: :model do
       flag_types = ['beta_unit_templates', 'production_unit_templates', 'gamma_unit_templates', 'alpha_unit_templates']
       exist_count = 0
       flag_types.each do |flag|
-        exist_count += $redis.exists(flag) ? 1 : 0
+        exist_count += $redis.exists(flag) == 1 ? 1 : 0
       end
       exist_count
     end
 
     it "deletes the cache of the saved unit" do
       $redis.set("unit_template_id:#{unit_template.id}_serialized", 'something')
-      expect($redis.exists("unit_template_id:#{unit_template.id}_serialized")).to be
+      expect($redis.exists("unit_template_id:#{unit_template.id}_serialized")).to eq(1)
       unit_template.update(name: 'something else')
-      expect($redis.exists("unit_template_id:#{unit_template.id}_serialized")).not_to be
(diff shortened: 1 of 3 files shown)


services/QuillLMS/spec/models/unit_template_spec.rb
```diff
@@ -195,23 +195,23 @@ def exist_count
...
-        exist_count += $redis.exists(flag) ? 1 : 0
+        exist_count += $redis.exists(flag) == 1 ? 1 : 0
       end
       exist_count
     end
 
     it "deletes the cache of the saved unit" do
       $redis.set("unit_template_id:#{unit_template.id}_serialized", 'something')
-      expect($redis.exists("unit_template_id:#{unit_template.id}_serialized")).to be
+      expect($redis.exists("unit_template_id:#{unit_template.id}_serialized")).to eq(1)
```
brendanshean, line 205: Based on this [comment](https://github.com/redis/redis-rb/blob/master/lib/redis.rb#L21), I would probably use `exists?` here to keep the spirit of the spec the same.  That being said, it only looks like `$redis.exists` is used in our codebase in this spec, so it's not super important.

brendanshean (approved):
Thank you for updating this.  One minor comment you can take or leave.

services/QuillLMS/spec/models/unit_template_spec.rb
cissyxyu: The `exists?` command isn't working as expected -- it evaluates to `false` always. I haven't able to find good documentation about it and the console is warning me that it will be deprecated on the next version of `redis-namespace`.

Because we don't actually use this command anywhere else in the codebase except in this test, I think I would be okay with testing this way with the `exists` command for the sake of time. If we ever decide to use this command in the actual code we should investigate how `exists` / `exists?` works, but for now I think this test does suffice.

happythenewsad (approved):

services/QuillLMS/spec/models/unit_template_spec.rb
brendanshean: That sounds like a good plan.

PR #8982, opened 2022-03-22
Cache `/teachers/progress_reports/activity_with_recommendations_ids`

anathomical:
## WHAT
Cache `/teachers/progress_reports/activity_with_recommendations_ids`
## WHY
In our effort to cache all of our reporting endpoints to reduce db load
## HOW
Since this is actually a set of global data (basically it's a unique list of all `activity_ids` that are in the `Recommendations` table), we can cache this farther away from the controller.  IN this case, I've added caching within the class method that's used to retrieve this data since it's the same for all users.

### Notion Card Links
https://www.notion.so/quill/Turn-Caching-Back-On-4bf7669d517443909640387fbafdc958

PR Checklist | Your Answer
------------ | -------------

### DIFF
--- a/services/QuillLMS/spec/models/activity_spec.rb
+++ b/services/QuillLMS/spec/models/activity_spec.rb
@@ -608,7 +608,7 @@ describe Activity, type: :model, redis: true do
     let!(:recommendation1) { create(:recommendation) }
     let!(:recommendation2) { create(:recommendation) }
 
-    after(:each) do
+    after do
       Rails.cache.clear
     end
 
--- a/services/QuillLMS/spec/models/activity_spec.rb
+++ b/services/QuillLMS/spec/models/activity_spec.rb
@@ -603,4 +603,26 @@ describe Activity, type: :model, redis: true do
       end
     end
   end
+
+  context '#activity_with_recommendations_ids' do
+    let!(:recommendation1) { create(:recommendation) }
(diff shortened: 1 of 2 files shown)


services/QuillLMS/spec/models/activity_spec.rb
```diff
@@ -603,4 +603,26 @@
       end
     end
   end
+
+  context '#activity_with_recommendations_ids' do
+    let!(:recommendation1) { create(:recommendation) }
+    let!(:recommendation2) { create(:recommendation) }
+
+    after do
+      Rails.cache.clear
```
brendanshean, line 612: I'm not sure I understand the purpose of this after hook. Is there some cache state we're trying to avoid in line 616?

anathomical: This could have been a `before` call, too, but basically if you don't clear the cache then the run in 616 caches data causing the 621 expectation to fail AND causing 624 to fail (the cached values do NOT match the ids that are created for the test since they were cached during a previous test run using models that were cleaned up and replaced).  I figured that given the caching of invalid values, it made sense to clear cache `after` to make sure that no subsequent tests get impacted.

brendanshean: That makes sense.  Thank you for the detailed explanation.

brendanshean (approved):

PR #8985, opened 2022-03-23
lms/pre-cache-admin-dashboards

anathomical:
## WHAT
Add a cron job to pre-cache admin dashboards overnight
## WHY
This data can take from a couple of seconds up to a handful of minutes to generate, and while we do cache it, we don't currently generate it until the first time an admin logs in.  This makes the first interaction that an admin has with our system each day incredibly frustrating.  By generating the cache data overnight, we can ensure that the first interaction an admin has with our system each day feels performant.
## HOW
Get a list of all admin users who have logged in this academic year and pre-generate their dashboard data during our nightly cron jobs

### Notion Card Links
https://www.notion.so/quill/School-Admin-Report-Caching-Pre-caching-d9182688be4c4fe78f13ef4d16d15bf2

PR Checklist | Your Answer
------------ | -------------

### DIFF
--- a/services/QuillLMS/app/workers/pre_cache_admin_dashboards_worker.rb
+++ b/services/QuillLMS/app/workers/pre_cache_admin_dashboards_worker.rb
@@ -5,10 +5,10 @@ class PreCacheAdminDashboardsWorker
   sidekiq_options queue: SidekiqQueue::DEFAULT
 
   def perform
-    active_admins = User.where('last_sign_in >= ?', School.school_year_start(Time.now)).joins(:schools_admins).all
+    active_admin_ids = User.where('last_sign_in >= ?', School.school_year_start(Time.now)).joins(:schools_admins).pluck(:id)
 
-    active_admins.each do |user|
-      FindAdminUsersWorker.perform_async(user.id)
+    active_admin_ids.each do |id|
+      FindAdminUsersWorker.perform_async(id)
     end
   end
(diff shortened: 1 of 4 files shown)


happythenewsad (approved):

services/QuillLMS/app/workers/pre_cache_admin_dashboards_worker.rb
```diff
@@ -0,0 +1,14 @@
+# frozen_string_literal: true
+
+class PreCacheAdminDashboardsWorker
+  include Sidekiq::Worker
+  sidekiq_options queue: SidekiqQueue::DEFAULT
+
+  def perform
+    active_admins = User.where('last_sign_in >= ?', School.school_year_start(Time.now)).joins(:schools_admins).all
```
brendanshean: Is `joins(:school_admins)` needed here?

brendanshean: Also, what about `pluck(:id)` instead of all since we're just using the id?

anathomical: The `joins` is necessary because that's an inner join and means that we only pull users who have an entry in the `schools_admins` table, which is how we know that the user is an admin.  `pluck` is reasonable, so I'll make that change.

PR #8986, opened 2022-03-23
Evidence - Add AutoML predict timeout

dandrabik:
## WHAT
We discussed adding a time out to the Evidence APIs generically, but I've found a few sources (e.g. https://adamhooper.medium.com/in-ruby-dont-use-timeout-77d9d4e5a001) that convinced me that a generic ruby timeout has issues and is less ideal than scoping a timeout to specific actions, like an `http` request. 

So I'm adding timeouts to the `http` calls instead.  AutoML is the only one without a timeout, so adding that one.
## WHY

If the AutoML API takes more that 5 seconds, it's likely an issue with Google's service, so we shouldn't stop the student from continuing on an activity.
## HOW
Add an `config.timeout = 5.0` setting specifically to the Prediction Service.

Note, I couldn't find a way to trigger the timeout in smoke testing, but from the google api code and [this guide](https://github.com/ankane/the-ultimate-guide-to-ruby-timeouts#google-cloud), it will raise a `Google::Cloud::Error` so that's what I'm expecting.

### Notion Card Links
https://www.notion.so/quill/Add-timeouts-to-Ruby-feedback-endpoint-code-2459e2cc99b94e6fa2c5b3018eb8a970

PR Checklist | Your Answer
------------ | -------------
Spec Review: Have you reviewed the spec and ensured this meets requirements and/or matches design mockups? | n/a

### DIFF
--- a/services/QuillLMS/engines/evidence/app/models/evidence/automl_model.rb
+++ b/services/QuillLMS/engines/evidence/app/models/evidence/automl_model.rb
@@ -13,6 +13,7 @@ module Evidence
       STATE_ACTIVE = 'active',
       STATE_INACTIVE = 'inactive'
     ]
+    PREDICT_API_TIMEOUT = 5.0
 
     attr_readonly :automl_model_id, :name, :labels
 
@@ -126,7 +127,9 @@ module Evidence
     end
 
     private def automl_prediction_client
-      @automl_prediction_client ||= Google::Cloud::AutoML.prediction_service
+      @automl_prediction_client ||= Google::Cloud::AutoML.prediction_service do |config|
+        config.timeout = PREDICT_API_TIMEOUT
+      end
     end
 
     private def automl_model_full_id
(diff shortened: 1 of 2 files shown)


happythenewsad (approved):
FYI in future PRs, [Timecop](https://github.com/travisjeffery/timecop) should be able to simulate time passing and organically trigger the timeout exception.

anathomical (approved):

dandrabik:
> FYI in future PRs, [Timecop](https://github.com/travisjeffery/timecop) should be able to simulate time passing and organically trigger the timeout exception.

So the tricky part here is that that API timeout is WITHIN the Google library method call and the method call also includes google's authentication request (so I don't think Timecop would work here. We'd have to stub the auth request and mock a timeout for the predict request that happen within the library. I'm still working on that test, but am merging this PR since I'm fairly confident the test I added represents what would happen (some error would be raised).

PR #9004, opened 2022-03-24
evidence/make-all-feedback-html

anathomical:
## WHAT
Add a rake task to ensure Feedback is always stored as HTML in case we ever need to depend on it being consistent
## WHY
For consistency.  The instances where they aren't HTML come from feedbacks that weren't entered via the admin interface which makes them HTML by default.
## HOW
Write a rake task to look over the two places we store feedback (`Evidence::Feedback.text` and `FeedbackHistory.feedback_text`), find all entries that aren't in HTML (as determined by not starting with `<p>`).  Then for each of those items, update it as an HTML string by mirroring the transformation done by our front-end interface in Evidence admin:
1. Replacing single quotes with `&#x27;`
2. Replacing double quotes with `&quot;`
3. Wrapping the full string in `<p>` and `</p>`

### Notion Card Links
https://www.notion.so/quill/Make-all-feedback-HTML-42a134d164e04ce083667002fde29af0

PR Checklist | Your Answer
------------ | -------------
Have you added and/or updated tests? |  No tests for rake task

### DIFF
--- a/services/QuillLMS/lib/tasks/evidence_feedback.rake
+++ b/services/QuillLMS/lib/tasks/evidence_feedback.rake
@@ -1,0 +1,26 @@
+# frozen_string_literal: true
+
+namespace :evidence_feedback do
+  desc 'Find any feedback that is not in HTML format and update it'
+  task :ensure_html => :environment do
+    def htmlize(input)
+      return input if input.start_with?('<p>')
+      input = input.sub('\'', '&#x27;')
+      input = input.sub('"', '&quot;')
+      "<p>#{input}</p>"
+    end
+
+    Evidence::Feedback.where("text NOT LIKE '<p>%'").each do |feedback|
+      feedback.update(text: htmlize(feedback.text))
+    end
+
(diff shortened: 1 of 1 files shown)


brendanshean (approved):
Depending on how rigid you need the sweep to be, you might consider adding some HTML validation (e.g. via [Nokogiri](  https://nokogiri.org/tutorials/ensuring_well_formed_markup.html#gentle-well-formedness) ).  Take for instance, the input string contains a closing tag like `</p>` but does not start with `<p>`.  This might be overkill, but thought i would mention.

anathomical:
> Depending on how rigid you need the sweep to be, you might consider adding some HTML validation (e.g. via [Nokogiri](https://nokogiri.org/tutorials/ensuring_well_formed_markup.html#gentle-well-formedness) ). Take for instance, the input string contains a closing tag like `</p>` but does not start with `<p>`. This might be overkill, but thought i would mention.

Totally reasonable suggestion.  I took a look at some possibilities, but I think we'd need to have some samples we wanted to correct for to make Nokogiri really deal with any problems.  Basically, since we don't know of any such examples, I'm inclined to skip it for now and keep things simple.

PR #9005, opened 2022-03-25
Create and renew teacher subscriptions using the Stripe Subscription API

brendanshean:
## WHAT
Move the handling of Teacher Subscriptions from server to the the Stripe Subscriptions API. 

NB: A subsequent PR will handle the transitioning of existing Subscriptions to renewal via Stripe.

## WHY
We'd like to streamline how we manage subscriptions

## HOW
There are three places where teachers have the ability to purchase subscriptions
1) At `/premium` there is a 'Buy Now' button
Screen Shot 2022-04-06 at 2 36 14 PM

2) If a user is in trial mode, `/premium` will also show 'Upgrade to Premium Now'
Screen Shot 2022-04-06 at 2 36 18 PM

3) If a user had a subscription but it has now expired, a 'Renew Subscription' button appears on `/subscriptions` page.
renew-subscription-via-ui

Each of these buttons has a new jsx component `StripeCheckoutSessionButton` which when clicked, triggers a [post](https://github.com/empirical-org/Empirical-Core/blob/f3500bd340e0e922c6a914d436dafa00472feb01/services/QuillLMS/client/app/bundles/Teacher/components/shared/StripeCheckoutSessionButton.jsx#L28) request to `StripeIntegration::CheckoutSessionsController`.  This controller [requests](https://github.com/empirical-org/Empirical-Core/blob/f3500bd340e0e922c6a914d436dafa00472feb01/services/QuillLMS/app/controllers/stripe_integration/checkout_sessions_controller.rb#L8) from Stripe a Checkout::Session object which has a url that that the frontend will then [redirect](https://github.com/empirical-org/Empirical-Core/blob/f3500bd340e0e922c6a914d436dafa00472feb01/services/QuillLMS/client/app/bundles/Teacher/components/shared/StripeCheckoutSessionButton.jsx#L39) to:  

stripe-checkout-flow

After the user successfully fills in the credit card info and clicks 'Subscribe', Stripe redirects to the `success_url` specified [here](https://github.com/empirical-org/Empirical-Core/blob/f3500bd340e0e922c6a914d436dafa00472feb01/services/QuillLMS/app/controllers/stripe_integration/checkout_sessions_controller.rb#L49) at which point Stripe immediately sends a bunch of webhook events, which I'll say more about later.

The browser redirects to the `subscriptions/?checkout_session_id={CHECKOUT_SESSION_ID}` where `CHECKOUT_SESSION_ID`  is used to  [determine](https://github.com/empirical-org/Empirical-Core/blob/f3500bd340e0e922c6a914d436dafa00472feb01/services/QuillLMS/app/controllers/subscriptions_controller.rb#L60) that the subscription was purchased successfully.  This boolean is used to determine if the `PremiumConfirmationModal` is show to the user after completion of purchase.
Screen Shot 2022-04-06 at 2 17 41 PM

From there the the "Payment Method on File" should list the credit card along with the last4, which has to be calculated [differently](https://github.com/empirical-org/Empirical-Core/blob/f3500bd340e0e922c6a914d436dafa00472feb01/services/QuillLMS/app/models/stripe_integration/subscription.rb#L5) since the previous [approach](https://github.com/empirical-org/Empirical-Core/blob/5cc36287f1ed2bdc3d7b2bedf5c04778b9e43a7a/services/QuillLMS/app/models/user.rb#L307) no longer works.

If the user clicks 'Edit Credit Card', a similar [flow](https://github.com/empirical-org/Empirical-Core/blob/f3500bd340e0e922c6a914d436dafa00472feb01/services/QuillLMS/app/controllers/stripe_integration/billing_portal_sessions_controller.rb#L6) to the CheckoutSession one occurs:

edit-card

#### Webhooks
From a high level, Stripe sends various messages our endpoint `/stripe_integration/webhooks`.  From there our controller must respond to the events [immediately](https://github.com/empirical-org/Empirical-Core/blob/f3500bd340e0e922c6a914d436dafa00472feb01/services/QuillLMS/app/controllers/stripe_integration/webhooks_controller.rb#L8) returning a 200 or 400 depending on the circumstances and kick off a background job `StripeIntegration::Webhooks::HandleEventWorker` to process the event.  The `EventHandlerFactory` determines which handler to use.  Currently there are:

`InvoicePaidEventHandler`
When [triggered](https://github.com/empirical-org/Empirical-Core/blob/f3500bd340e0e922c6a914d436dafa00472feb01/services/QuillLMS/app/services/stripe_integration/webhooks/invoice_paid_event_handler.rb#L13) provisions a new Subscription using the `SubscriptionCreator`  [service object](https://github.com/empirical-org/Empirical-Core/blob/f3500bd340e0e922c6a914d436dafa00472feb01/services/QuillLMS/app/services/stripe_integration/webhooks/subscription_creator.rb#L5)

NB.  In my Stripe RFC Schema, I had initially added a column `stripe_subscription_id` to the `Subscription` table.  However, Stripe persists that id across renewals, whereas our system creates a new subscription object each renewal.  To better reflect this modeling, I've instead added a `stripe_invoice_id` column which has a one-to-one correspondence with our `Subscription` table.  It also allows us to pull the relevant information from Stripe when provisioning our new Subscription objects.

`IgnoredEventHandler`
This is the current [list](https://github.com/empirical-org/Empirical-Core/blob/f3500bd340e0e922c6a914d436dafa00472feb01/services/QuillLMS/app/services/stripe_integration/webhooks/ignored_event_handler.rb#L6) of events that I have observed (and at this point am ignoring).

`UnknownEventHandler`
All other events are considered unknown and will [log](https://github.com/empirical-org/Empirical-Core/blob/f3500bd340e0e922c6a914d436dafa00472feb01/services/QuillLMS/app/services/stripe_integration/webhooks/unknown_event_handler.rb#L9) an error so that we can then either update our codebase with a new Handler or add that unknown event to the ignored list.

### Screenshots

### Notion Card Links
https://www.notion.so/quill/Implement-a-new-direct-subscription-process-that-utilizes-Stripe-subscriptions-229f2cba7d5e4dbf8bf95388f1ab4209

PR Checklist | Your Answer
------------ | -------------


emilia-friedberg (approved):
Thank you for this incredibly clear and thorough explainer - it definitely made the code much easier to follow. One question I have is around the `CHECKOUT_SUCCESS_URL`/banner logic - if the webhooks don't complete until after the redirect is completed, will the user just never see the banner? Maybe it's generally fast enough that we don't need to worry, but we might want to consider using Pusher or something when the background job completes if it has not yet succeeded.

brendanshean:
@emilia-friedberg That's a very astute observation.  I kind of settled with the hack [here](https://github.com/empirical-org/Empirical-Core/blob/111926fcd5578336362b2d363d073d7db50cc1c6/services/QuillLMS/app/models/subscription.rb#L350), but Pusher seems like a better solution.  Thank you for the suggestion.

services/QuillLMS/app/models/user.rb
```diff
@@ -616,6 +603,10 @@ def completed_activity_count
     user_activity_classifications.sum(:count)
   end
 
+  def subscription_status
+    subscription&.subscription_status || last_expired_subscription&.subscription_status
```
emilia-friedberg, line 605: I might be missing something, but is it a problem that the old version of code does return an object if the user has no subscription, and here we'd always be returning nil?

services/QuillLMS/app/models/subscription.rb
```diff
@@ -352,4 +342,33 @@ def self.create_with_school_or_user_join school_or_user_id, type, attributes
     subscription
   end
 
+  # rubocop:disable Metrics/CyclomaticComplexity
+  def self.stripe_purchase_completed?(checkout_session_id)
+    return false if checkout_session_id.nil?
+
+    # TODO: remove this blocking waits for webhooks to complete
+    sleep ENV.fetch('STRIPE_WAIT_FOR_WEBHOOK', 0).to_i
```
emilia-friedberg: Oh, I see, this is handling my concern about the webhooks completing after page load?

services/QuillLMS/client/app/bundles/Teacher/components/subscriptions/SubscriptionStatus.jsx
```diff
@@ -78,7 +86,7 @@ const SubscriptionStatus = ({ subscriptionType, showPurchaseModal, subscriptionS
     case SCHOOL_PREMIUM:
     case DISTRICT_PREMIUM:
     case SCHOOL_PREMIUM_SCHOLARSHIP:
-      content.pCopy = schoolPremiumCopy(subscriptionType);
+      content.premiumCopy = schoolPremiumCopy(subscriptionType);
```
emilia-friedberg, line 89: lol this is definitely an improvement

services/QuillLMS/client/app/bundles/Teacher/components/subscriptions/current_subscription.jsx
```diff
@@ -151,13 +148,30 @@ export default class CurrentSubscription extends React.Component {
...
 
   handleEditCreditCardClick = () => {
-    new EnterOrUpdateStripeCard(this.updateLastFour, 'Update');
-  };
+    const { subscriptionStatus } = this.props
+    const { stripe_customer_id } = subscriptionStatus
+
+    if ( !stripe_customer_id ) { return }
+
+    request.post({
```
emilia-friedberg: I don't think you'd have much way of knowing this, but we've been trying to use a centralized `request` module that @anathomical made a while back rather than invoking `request` directly (by importing from here: https://github.com/empirical-org/Empirical-Core/blob/develop/services/QuillLMS/client/app/modules/request/index.js#L48). This is definitely something we could change as a follow-up card.

services/QuillLMS/client/app/bundles/Teacher/components/shared/StripeCheckoutSessionButton.jsx
```diff
@@ -0,0 +1,56 @@
...
+    if (!userIsSignedIn) {
+      alert('You must be logged in to activate Premium.')
+    } else if (!userIsEligibleForNewSubscription) {
+      alert(
+        `You have an active subscription and cannot buy premium now. If your subscription is a school subscription,
+        you may buy Premium when it expires. If your subscription is a teacher one, please turn on recurring payments
+        and we will renew it automatically when your subscription ends.`
+      )
+    } else {
+      request.post({
```
emilia-friedberg: See comment elsewhere about `requestPost`.

emilia-friedberg (approved):
Submitted last comment too early! A few more notes for your consideration. I'm really impressed by this effort.

services/QuillLMS/app/models/subscription.rb
brendanshean: Definitely a hack 😜.  I think pusher will be a more sensible approach.

services/QuillLMS/client/app/bundles/Teacher/components/subscriptions/current_subscription.jsx
brendanshean: Thank you for drawing attention.  I did notice the `request` is deprecated so I will update this now.

services/QuillLMS/app/models/user.rb
brendanshean: This was tricky, so definitely worth discussing.  I extracted subscription_status from the [controller](https://github.com/empirical-org/Empirical-Core/pull/9005/files#diff-aa578c282750da885731bb9b3b1d6f9a2a74028b046729e15c1c8e18acff5c5aL72) and moved it to the User model since it's asking a question of the user.  The original method has three branches for what the user has: 
1) a current subscription
2) an expired subscription
3) no subscriptions

For case 3), the original code had an explicit return, but the new disjunction `subscription&.subscription_status || last_expired_subscription&.subscription_status` should now cover case 3) via the safe_navigation operator `&.`

services/QuillLMS/app/controllers/stripe_integration/checkout_sessions_controller.rb
```diff
@@ -0,0 +1,53 @@
...
+    private def stripe_price_id
+      params[:stripe_price_id]
+    end
+
+    private def stripe_customer_id
+      User.find_by(email: customer_email).stripe_customer_id
+    end
+
+    private def success_url
+      "#{ENV['DEFAULT_URL']}/subscriptions?checkout_session_id={CHECKOUT_SESSION_ID}"
```
dandrabik, line 49: The `checkout_session_id` is a constant? Just flagging in case that was a mistake.

services/QuillLMS/app/controllers/stripe_integration/webhooks_controller.rb
```diff
@@ -0,0 +1,37 @@
...
+  class WebhooksController < ApplicationController
+    protect_from_forgery except: :create
+
+    def create
+      Webhooks::HandleEventWorker.perform_async(stripe_webhook_event.id)
+      head 200
+    rescue ActiveRecord::RecordNotUnique
+      head 200
+    rescue => e
+      NewRelic::Agent.notice_error(e)
```
dandrabik: A few days ago, I added a `ErrorNotifier.report(error)`. I'd like to start using that in other places (we log them in both Sentry and NR for now).
```suggestion
      ErrorNotifier.report(e)
```

services/QuillLMS/app/controllers/stripe_integration/webhooks_controller.rb
```diff
@@ -0,0 +1,37 @@
...
+      head 400
+    end
+
+    private def endpoint_secret
```
dandrabik: Would this be better as a constant, e.g. `ENDPOINT_SECRET = ENV.fetch('STRIPE_ENDPOINT_SECRET', '')`

services/QuillLMS/app/models/subscription.rb
```diff
@@ -80,15 +83,26 @@ class Subscription < ApplicationRecord
...
-  PAYMENT_METHODS = ['Invoice', 'Credit Card', 'Premium Credit']
+  PAYMENT_METHODS = [
+    INVOICE_PAYMENT_METHOD = 'Invoice',
+    CREDIT_CARD_PAYMENT_METHOD = 'Credit Card',
+    PREMIUM_CREDIT_PAYMENT_METHOD = 'Premium Credit'
+  ]
+
   ALL_TYPES = OFFICIAL_FREE_TYPES.dup.concat(OFFICIAL_PAID_TYPES)
 
+  validates :stripe_invoice_id, allow_blank: true, format: { with: /\Ain_[0-9a-zA-Z]*\z/ }
```
dandrabik: Do we have documentation that this is a set format for `invoice.id`? Will it be loud if this fails (if they add an underscore or something).

services/QuillLMS/app/models/subscription.rb
```diff
@@ -80,15 +83,26 @@ class Subscription < ApplicationRecord
...
+
   scope :active, -> { where(de_activated_date: nil).where("expiration > ?", Date.today).order(expiration: :asc) }
 
   def is_trial?
     account_type && TRIAL_TYPES.include?(account_type)
   end
 
+  def expired?
+    expiration <= Date.today
```
dandrabik, line 103: We might want to add this cop and move away from `Date.today` (in favor of Date.current or `Time.now.utc.to_date`), just to ensure the timezone is utc.

https://www.rubydoc.info/gems/rubocop/0.41.2/RuboCop/Cop/Rails/Date

services/QuillLMS/app/models/user.rb
```diff
@@ -243,13 +243,7 @@ def subscription_authority_level(subscription_id)
...
 
   def eligible_for_new_subscription?
-    if subscription
-      # if they have a subscription it must be a trial one
-      Subscription::TRIAL_TYPES.include?(subscription.account_type)
-    else
-      # otherwise they are good for purchase
-      true
-    end
+    !subscription || Subscription::TRIAL_TYPES.include?(subscription.account_type)
```
dandrabik: I think(?) this is just a `nil` check, so consider `subscription.nil?`for readability?

services/QuillLMS/app/serializers/plan_serializer.rb
```diff
@@ -0,0 +1,29 @@
...
+#
+# Indexes
+#
+#  index_plans_on_name  (name) UNIQUE
+#
+class PlanSerializer < ActiveModel::Serializer
+  attributes :display_name, :price_in_dollars, :stripe_price_id
+
+  def price_in_dollars
+    object.price / 100
```
dandrabik: Price is an integer, let's make the denominator a float to avoid weird integer division.
```suggestion
    object.price / 100.0
```

Screen Shot 2022-04-07 at 5 15 44 PM

services/QuillLMS/app/services/stripe_integration/webhooks/subscription_creator.rb
```diff
@@ -0,0 +1,94 @@
...
+      private def start_date
+        Time.at(stripe_subscription.current_period_start).to_date
+      end
+
+      private def stripe_customer_id
+        stripe_subscription.customer
+      end
+
+      private def stripe_price_id
+        stripe_subscription&.items&.data&.first&.price&.id
```
dandrabik, line 68: It's nil checks all the way down! 🤣

services/QuillLMS/client/app/bundles/Teacher/components/scorebook/premium_banners/free_trial_status.jsx
```diff
@@ -1,27 +1,18 @@
...
-  const premiumButton = originPage == 'premium' ?
-    (
-      <button className='btn-orange' onClick={handleClickUpgradeNow} type='button'>Upgrade to Premium Now</button>
-    ) :
-    (
-      <a href='/premium'>
-        <button className='btn-orange' type='button'>Upgrade to Premium Now</button>
-      </a>
-    )
+  const premiumButton = (originPage == 'premium')
```
dandrabik: Copied code, but I think we want a threequal.
```suggestion
  const premiumButton = (originPage === 'premium')
```

services/QuillLMS/client/app/bundles/Teacher/components/subscriptions/current_subscription.jsx
```diff
@@ -35,18 +35,15 @@ export default class CurrentSubscription extends React.Component {
 
   getPaymentMethod() {
     const { subscriptionStatus, authorityLevel, subscriptionType, } = this.props
-    const { lastFour, } = this.state
 
-    if (subscriptionStatus && subscriptionStatus.payment_method === 'Credit Card' && lastFour && authorityLevel) {
+    if (subscriptionStatus && subscriptionStatus.payment_method === 'Credit Card' && authorityLevel) {
```
dandrabik, line 38: This isn't your code, so feel free to punt for now, but we should consider using more constants for strings that are used for logic in the payments code, e.g. `'Credit Card'`.

services/QuillLMS/client/app/bundles/Teacher/components/subscriptions/current_subscription.jsx
```diff
@@ -197,15 +234,21 @@ export default class CurrentSubscription extends React.Component {
...
+        <i className="fas fa-icon fa-lightbulb-o" />
+        {body}
+      </div>
+    )
   }
 
   nextPlanAlertOrButtons(condition, renewDate) {
-    const { lastFour, } = this.state
     const { authorityLevel, subscriptionStatus, } = this.props
+    const { last_four } = subscriptionStatus
```
dandrabik, line 239: I'm noticing more mixed variable naming (some `camelcase`, some `snakecase`). Maybe this is coming from Stripe, so we're not in control of it, but it might be worth discussing with @emilia-friedberg and @eadams17 how we should determine which to use, so that if someone else is working on this code, it's clear what to use (or it can lead to errors/bugs down the road).

services/QuillLMS/app/controllers/stripe_integration/billing_portal_sessions_controller.rb
```diff
@@ -0,0 +1,22 @@
...
+    private def billing_portal_args
+      { customer: customer, return_url: return_url }
+    end
+
+    private def customer
+      params[:stripe_customer_id]
+    end
+
+    private def return_url
+      "#{ENV['DEFAULT_URL']}/subscriptions"
```
dandrabik: Is there a route_url we can use? Trying to get away from these manual urls.

dandrabik (approved):
Overall, looks good to me! Thanks for putting this together, it looks like a lot of work! Thanks for being so thorough! 

I made a bunch of small comments, but none of them are blockers (I think all are take or leave).

👍

services/QuillLMS/app/controllers/stripe_integration/checkout_sessions_controller.rb
brendanshean: I was confused by this in the [docs](https://stripe.com/docs/payments/checkout/custom-success-page#modify-success-url), but it's apparently a "template variable", allowing Stripe to populate it later.

services/QuillLMS/app/controllers/stripe_integration/webhooks_controller.rb
brendanshean: Cool!  I will update.

services/QuillLMS/client/app/bundles/Teacher/components/subscriptions/current_subscription.jsx
brendanshean: I would agree.  I'm planning on addressing this in one of the future stripe PRs.

services/QuillLMS/client/app/bundles/Teacher/components/subscriptions/current_subscription.jsx
brendanshean: I spoke with Emilia and since the subscriptionStatus is older code, we'll leave as-is for now.

brendanshean: Long term I think the subscriptionStatus should be refactored and we can address it then.

services/QuillLMS/app/models/subscription.rb
brendanshean: @dandrabik I couldn't find any documentation concerning the exact form, but all of the different models in the Stripe documentation followed this same format.  I've added a [validator](https://github.com/empirical-org/Empirical-Core/blob/da3ea2f9f97db42f1a9939fecb00a6553368666a/services/QuillLMS/app/models/validators/stripe_uid_validator.rb) to group the pattern and give more informative error reporting when validation fails (e.g. subscription [creation](https://github.com/empirical-org/Empirical-Core/blob/da3ea2f9f97db42f1a9939fecb00a6553368666a/services/QuillLMS/app/services/stripe_integration/webhooks/subscription_creator.rb#L72))

services/QuillLMS/app/models/subscription.rb
brendanshean: I will update this instance and add then add the Rubocop setting in a future PR since it will require a new gem: rubocop-rails.

brendanshean:
@emilia-friedberg Just a heads-up, I'm re-requesting review due to the Pusher integration that wasn't in the original PR
It's basically these two commits:
pre-work: https://github.com/empirical-org/Empirical-Core/pull/9005/commits/a67fba8b8206a5d3d1d9a83e6211c3dc33de2af8
implementation:  https://github.com/empirical-org/Empirical-Core/pull/9005/commits/c6227750c3a638dfc9e058e175ad3ec809d602b8

emilia-friedberg (commented):
This looks great from the frontend but I'm not seeing the part where the backend is actually implementing the pusher notification? Maybe I've just got Monday morning brain.

brendanshean:
@emilia-friedberg My mistake, I forgot to include the third commit:  https://github.com/empirical-org/Empirical-Core/pull/9005/commits/c6227750c3a638dfc9e058e175ad3ec809d602b8

brendanshean:
@emilia-friedberg You're right.  It's not in there.  I guess it must have been succeeding the first time everytime.  I need to add in the code for `quill_retrieval_processing`

brendanshean:
@emilia-friedberg I forgot that I removed it from the background job due to the race condition we talked about earlier.  I've added it back in now since we still want it to trigger if it's slower than the redirect:  https://github.com/empirical-org/Empirical-Core/pull/9005/commits/c47acdceee5d8cf888b134ec9cc9bb032fbe8e07

brendanshean:
@emilia-friedberg Here is the commit for the pusher integration backend: https://github.com/empirical-org/Empirical-Core/pull/9005/commits/aad026e30b21bb644ae631b5f06b756bda697f8f

I verified it working by temporarily adding a `sleep 15` in the worker job.

services/QuillLMS/app/controllers/pages_controller.rb
```diff
@@ -439,8 +436,7 @@ def premium
   end
   # rubocop:enable Metrics/CyclomaticComplexity
 
-  def tutorials
-  end
+  def tutorials; end
```
cissyxyu, line 439: Might we just want to clean up these empty endpoints? I could be wrong but I don't think we're intending to use them for anything.

services/QuillLMS/app/controllers/subscriptions_controller.rb
```diff
@@ -56,40 +63,22 @@ def destroy
   private def set_index_variables
     @subscriptions = current_user.subscriptions
     @premium_credits = current_user.credit_transactions.map {|x| x.serializable_hash(methods: :action)}.compact
-    subscription_status
+    @stripe_invoice_id = StripeIntegration::StripeInvoiceIdFinder.run(checkout_session_id)
```
cissyxyu, line 66: Is this supposed to be `StripeIntegration::StripeInvoiceIdFinder.new(checkout_session_id).run`? The class definition says that the initialization takes an argument but `run` does not.

services/QuillLMS/app/models/subscription.rb
```diff
@@ -80,15 +83,26 @@ class Subscription < ApplicationRecord
...
 
+  validates :stripe_invoice_id, allow_blank: true, stripe_uid: { prefix: :in }
+
   scope :active, -> { where(de_activated_date: nil).where("expiration > ?", Date.today).order(expiration: :asc) }
 
   def is_trial?
     account_type && TRIAL_TYPES.include?(account_type)
   end
 
+  def expired?
```
cissyxyu, line 102: I think we're using `expiration <= Date.today` in some other places in the code to check if a subscription is expired. Might be worth double checking to make sure we're using this new method in all those old places.

services/QuillLMS/app/models/subscription.rb
```diff
@@ -147,7 +162,6 @@ def credit_user_and_de_activate
     end
   end
 
-
   def self.expired_today_or_previously_and_recurring
     Subscription.where('expiration <= ? AND recurring IS TRUE AND de_activated_date IS NULL', Date.today)
```
cissyxyu, line 166: see above comment (we can use the new method `expired?` here)

services/QuillLMS/app/models/subscription.rb
```diff
@@ -218,16 +207,23 @@ def update_if_charge_succeeds
...
+      redemption_start_date(school_or_user) + 1.year
+    elsif school_or_user.instance_of?(School)
+      promotional_dates[:expiration]
     else
-      raise "an incorrect premium type #{premium_type} was passed"
+      Date.today + 1.year
     end
+  end
+
+  def save_if_charge_succeeds(premium_type, school=nil)
```
cissyxyu, line 221: Looks like we're changing this method to only save school premium subs. Might want to change the method name as well to reflect that so it's not confusing.

services/QuillLMS/app/controllers/pages_controller.rb
brendanshean: It seems to be [pointing](https://github.com/empirical-org/Empirical-Core/blob/develop/services/QuillLMS/app/views/pages/tutorials.html.erb) to a 'TutorialsApp' which has some [components](https://github.com/empirical-org/Empirical-Core/tree/develop/services/QuillLMS/client/app/bundles/Teacher/components/tutorials)

services/QuillLMS/app/controllers/subscriptions_controller.rb
brendanshean: That particular service object inherits from [ApplicationService](https://github.com/empirical-org/Empirical-Core/blob/develop/services/QuillLMS/app/services/application_service.rb#L5) which is  syntactic sugar that allows to collapse `ServiceObject.new(*args, &block).run` into just `ServiceObject.run(*args, block)`

services/QuillLMS/app/models/subscription.rb
brendanshean: I will most likely be removing this method when integrating the school subscription stuff.  I decided to leave it for now so to keep the commit diff down.

services/QuillLMS/app/models/subscription.rb
brendanshean: That's a good point.  Dan made a [suggestion](https://github.com/empirical-org/Empirical-Core/pull/9005#discussion_r845563873) to change this so I did but I will probably revert to the original and then change all of them when adding in the Rubocop configuration.

emilia-friedberg (approved):
Pusher stuff looks good so everything LGTM! Cheers!

brendanshean:
@dandrabik There is one more [commit](https://github.com/empirical-org/Empirical-Core/pull/9005/commits/a5d9e1ddcb79ef79243247a3fd539bbccf61beac) based on your QA testing:

dandrabik (approved):

PR #9009, opened 2022-03-25
Evidence AutoML - Only perform auth once in predict call.

dandrabik:
## WHAT
The current AutoML `fetch_labels` task uses two API clients: one to fetch the model path, one to get the prediction. This changes it to use one client for both.
## WHY
This is causing us to authenticate with google twice for each prediction.
## HOW
The `prediction_client` also has a `model_path` method. Use that. Update tests to be more isolated as well.
### Screenshots
Before
Screen Shot 2022-03-25 at 1 09 32 PM

After
Screen Shot 2022-03-25 at 3 09 39 PM

PR Checklist | Your Answer
------------ | -------------

### DIFF
--- a/services/QuillLMS/engines/evidence/spec/models/evidence/automl_model_spec.rb
+++ b/services/QuillLMS/engines/evidence/spec/models/evidence/automl_model_spec.rb
@@ -164,8 +164,9 @@ module Evidence
       end
 
       it 'should strip any whitespace that is on "automl_model_id" before looking records up via Google' do
-        ENV['AUTOML_GOOGLE_PROJECT_ID'] = 'PROJECT_ID'
-        ENV['AUTOML_GOOGLE_LOCATION'] = 'LOCATION'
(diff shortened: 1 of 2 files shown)


services/QuillLMS/engines/evidence/spec/models/evidence/automl_model_spec.rb
```diff
@@ -164,16 +164,17 @@ def automl_model.labels_have_associated_rules
       end
 
       it 'should strip any whitespace that is on "automl_model_id" before looking records up via Google' do
-        ENV['AUTOML_GOOGLE_PROJECT_ID'] = 'PROJECT_ID'
```
dandrabik, line 167: Just a note, setting ENV variables is a global change that doesn't get reset after the test is run, and should be avoided. Assigning to a constant and using `stub_const` is one way to get around this.

anathomical (approved):

PR #9019, opened 2022-03-30
prefilter checks should return concept uids

happythenewsad:
## WHAT
Returns concept_uid in response payload

## WHY
So that the prefilter response payload conforms to the standard evidence schema 

### Notion Card Links
https://www.notion.so/quill/Product-Board-30f97e4eb01246dbb8264253fb385073?p=002fca860bad4a8f87cc6c721a3754c9

PR Checklist | Your Answer
------------ | -------------
Have you deployed to Staging? | yes - added a new postman directory for this endpoint too

### DIFF
--- a/services/QuillLMS/engines/evidence/spec/lib/prefilter_check_spec.rb
+++ b/services/QuillLMS/engines/evidence/spec/lib/prefilter_check_spec.rb
@@ -90,6 +90,7 @@ module Evidence
             expect(response[:rule_uid]).to eq violation[:rule_uid]
             expect(response[:optimal]).to eq false
             expect(response[:feedback]).to eq "#{violation[:name]} feedback"
+            expect(response[:concept_uid]).to eq 'ConceptUID'
           end
         end
       end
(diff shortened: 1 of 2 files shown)


services/QuillLMS/engines/evidence/spec/lib/prefilter_check_spec.rb
```diff
@@ -90,6 +90,7 @@ module Evidence
             expect(response[:rule_uid]).to eq violation[:rule_uid]
             expect(response[:optimal]).to eq false
             expect(response[:feedback]).to eq "#{violation[:name]} feedback"
+            expect(response[:concept_uid]).to eq 'ConceptUID'
```
anathomical: I suspect that the answer is "from the factory", but why do we expect this value here?  Is there a way to make it clear where we're sourcing this from?

anathomical (approved):
Note on test expectations, but it wouldn't be the end of the world if we merged this as-written.

services/QuillLMS/engines/evidence/spec/lib/prefilter_check_spec.rb
happythenewsad: Yes, it is from the factory. I just pushed a commit which makes the intention clearer.

anathomical: Very clear.  Thanks.

dandrabik (approved):
:+1:

PR #9034, opened 2022-04-06
evidence/fix-plagiarism-with-repeated-words

anathomical:
## WHAT
Update plagiarism matching so that we count repeated words
## WHY
We want to be able to successfully identify plagiarism even when the text that includes it has repeated words.
## HOW
We use the length of the intersection of two arrays to determine if "enough" words are the same between those arrays to be worth doing a Levenshtein distance calculation on them.  Naive array intersections in Ruby (using &) collapse duplicate values, so [1,1] & [1,1] becomes [1].  This leads to arrays that are shorter than we expect, and thus caused the algorithm to decide that it wasn't worth doing a Levenshtein even when we wanted it to.  This set of changes re-hydrates the intersection value with duplicates so that the length is what we would expect.

For the curious, the failing match in production was on the entry "the toxic chemicals and microplastic release into the air and" and plagiarism text "the toxic chemicals and microplastics released into the air and".  Note that "microplastic" is singular in the entry and plural in the plagiarism text, that "release" is present tense in the entry and past tense in the plagiarism text.  That would give us an expected overlap length of 8 (10 words, 2 not overlapping), which should go on to a Levenshtein comparison.  However, you'll also note that "the" appears twice in both strings, as does "and".  Elimination of duplicates reduces length of the intersection by another 2 to 6, which is below our Levenshtein threshold so the algorithm was erroneously failing to process this section of the plagiarism match.

### Notion Card Links
https://www.notion.so/quill/Sentry-Error-plagiarism_check-undefined-method-for-nil-NilClass-6ead38ec036a42fb9456330ecd71ad1d

PR Checklist | Your Answer
------------ | -------------

### DIFF
--- a/services/QuillLMS/engines/evidence/lib/evidence/evidence/plagiarism_check.rb
+++ b/services/QuillLMS/engines/evidence/lib/evidence/evidence/plagiarism_check.rb
@@ -154,10 +154,8 @@ module Evidence
       # we know that there's a minimum number of words that have to be identical.  This function
       # confirms that minimum amount of overlap to justify doing a set of Levenshtein calculations.
       source_arrays.any? do |source_array|
-        [
-         (target_array - source_array).length,
-         (source_array - target_array).length
-        ].max <= FUZZY_CHARACTER_THRESHOLD
+        ((target_array - source_array).length <= FUZZY_CHARACTER_THRESHOLD ||
+         (source_array - target_array).length <= FUZZY_CHARACTER_THRESHOLD)
       end
     end
 
--- a/services/QuillLMS/engines/evidence/lib/evidence/evidence/plagiarism_check.rb
+++ b/services/QuillLMS/engines/evidence/lib/evidence/evidence/plagiarism_check.rb
@@ -158,7 +158,13 @@ module Evidence
       # we know that there's a minimum number of words that have to be identical.  This function
       # confirms that minimum amount of overlap to justify doing a set of Levenshtein calculations.
       source_arrays.any? do |source_array|
-        (source_array & target_array).length >= (MATCH_MINIMUM - FUZZY_CHARACTER_THRESHOLD)
+        intersection = source_array & target_array
+        # '&' collapses duplicates so that ['the', 'the'] & ['the', 'the'] == ['the']
+        # This throws off the expected value of length, so we use this bit here to add the duplicates back into the array.
+        # Note, the order will be scrambled [1,2,1] & [1,2,1] run through this flat_map will be [1,1,2], but since we are using
+        # length in this calculation, order doesn't matter
+        with_duplicates = intersection.flat_map { |word| [word]*[source_array.count(word), target_array.count(word)].min }
+        with_duplicates.length >= (MATCH_MINIMUM - FUZZY_CHARACTER_THRESHOLD)
       end
     end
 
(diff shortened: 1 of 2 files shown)


happythenewsad (commented):
`get_highlight(text, cleaned_text, matched_slice)` can still raise the error referenced in sentry if the `matched_slice` param is not valid (i.e. there is no match)

anathomical:
> `get_highlight(text, cleaned_text, matched_slice)` can still raise the error referenced in sentry if the `matched_slice` param is not valid (i.e. there is no match)

That would be true, but `get_highlight` can't be called if `matched_slice` is not valid.  It's called through `def highlights` which checks `optimal?` before actually calling `get_highlight`.  `optimal?` will be `true` if `matched_slice.blank?` is true, and that will bypass highlight generation.

happythenewsad (approved):

services/QuillLMS/engines/evidence/lib/evidence/evidence/plagiarism_check.rb
```diff
@@ -158,7 +158,13 @@ def feedback_object
       # we know that there's a minimum number of words that have to be identical.  This function
       # confirms that minimum amount of overlap to justify doing a set of Levenshtein calculations.
       source_arrays.any? do |source_array|
-        (source_array & target_array).length >= (MATCH_MINIMUM - FUZZY_CHARACTER_THRESHOLD)
+        intersection = source_array & target_array
```
dandrabik: You could also consider using Array minus (`-`) here to preserve duplicates and get the words that don't match, e.g. the example from the description:

Screen Shot 2022-04-06 at 3 58 39 PM

I believe this could also be
```ruby
(source_array - target_array).length <= FUZZY_CHARACTER_THRESHOLD

**Review by dandrabik — CHANGES_REQUESTED** [review]
This is working code, so we could ship this, but I think there is a simpler algorithm we could use (I've suggested one to test).

Since you have test in place, this should be easy to optimize.

File: services/QuillLMS/engines/evidence/lib/evidence/evidence/plagiarism_check.rb (hunk 1 above)
Comment (anathomical) [reply services/QuillLMS/engines/evidence/lib/evidence/evidence/plagiarism_check.rb]: Good suggestion.  I'll make tweaks and let you know when they're done so you can take a look.

File: services/QuillLMS/engines/evidence/lib/evidence/evidence/plagiarism_check.rb [hunk 2]
```diff
@@ -158,7 +154,10 @@ def feedback_object
       # we know that there's a minimum number of words that have to be identical.  This function
       # confirms that minimum amount of overlap to justify doing a set of Levenshtein calculations.
       source_arrays.any? do |source_array|
-        (source_array & target_array).length >= (MATCH_MINIMUM - FUZZY_CHARACTER_THRESHOLD)
+        [
+         (target_array - source_array).length,
```
dandrabik: This looks good. The only other slight optimization I might suggest is that you can early return if the first is above the threshold, i.e.
```ruby
(target_array - source_array).length <= FUZZY_CHARACTER_THRESHOLD 
|| (source_array - target_array).length <= FUZZY_CHARACTER_THRESHOLD

dandrabik (approved):
👍  Small optimization suggested, but otherwise 🚢

services/QuillLMS/engines/evidence/lib/evidence/evidence/plagiarism_check.rb
anathomical: Good catch.  I'll make that update and deploy.

PR #9035, opened 2022-04-06
LMS Adjust Time Tracking Outliers

dandrabik:
## WHAT
We are doing nuanced time-tracking per prompt on the front end, but there is an existing bug in that calculation which is causing many time-tracking lists to have one element that is a very large number (likely from a resumed session). 

We should fix the bug on the frontend, but the code here is a backstop against clearly bad data being recorded. It replaces any individual time element that is 40x the median with the median, e.g. the total for the list `[1,2,3,4,999999]` would be calculated as `[1,2,3,4,3].sum`
## WHY
Time tracking is important to teachers, but these large outliers really distort the data. Replacing with the median of the other values seems like a reasonable stopgap.
## HOW
Add Array extension for `Array.median`. Add a bunch of tests. Replace individual outliers with median.

Edit:
I added a `TimeTrackingCleaner` class and am editing the granular time_tracking data, and I've returned the `timespent` calculation to just be a `sum`.

Note `40x median` was my initial guess at a reasonable outlier threshold. It's likely that could be lower.

### Screenshots
Examples (time is in seconds)
Screen Shot 2022-04-06 at 12 30 24 PM
Screen Shot 2022-04-06 at 12 29 59 PM
Screen Shot 2022-04-06 at 12 30 12 PM

### Notion Card Links
https://www.notion.so/quill/ActivitySession-Time-Tracking-Outliers-Backend-mitigation-dbf94350104d460f8a6f8ce5c95799a4

PR Checklist | Your Answer
------------ | -------------
Have you deployed to Staging? | (Possible answers:Not yet - deploying now!
Self-Review: Have you done an initial self-review of the code below on Github? | YEs

### DIFF
--- a/services/QuillLMS/app/lib/time_tracking_cleaner.rb
+++ b/services/QuillLMS/app/lib/time_tracking_cleaner.rb
@@ -1,4 +1,6 @@
-class TimeTrackingCleaner
+# frozen_string_literal: true
+
+class TimeTrackingCleaner < ApplicationService
   OUTLIER_MULTIPLIER = 40
 
   attr_reader :time_tracking, :data_params
@@ -8,7 +10,7 @@ class TimeTrackingCleaner
     @time_tracking = data_params&.fetch(ActivitySession::TIME_TRACKING_KEY, nil)
   end
 
-  def clean
+  def run
     return data_params unless time_tracking.respond_to?(:values)
     return data_params if outliers.empty?
 
--- a/services/QuillLMS/app/lib/time_tracking_cleaner.rb
+++ b/services/QuillLMS/app/lib/time_tracking_cleaner.rb
@@ -10,21 +10,26 @@ class TimeTrackingCleaner < ApplicationService
     @time_tracking = data_params&.fetch(ActivitySession::TIME_TRACKING_KEY, nil)
   end
 
+
+  # replace outlier time_tracking values with the list's median value
+  # if they are 40x larger than the median
+  # keep track of original data in data['time_tracking_edits']
+  # this is a backstop against a bug in the frontend time tracking with resumed sessions
   def run
     return data_params unless time_tracking.respond_to?(:values)
     return data_params if outliers.empty?
 
     data_params.merge(
-      ActivitySession::TIME_TRACKING_KEY => cleaned_time_tracking,
+      ActivitySession::TIME_TRACKING_KEY => time_tracking_outliers_replaced,
       ActivitySession::TIME_TRACKING_EDITS_KEY => outliers
     )
   end
 
-  def cleaned_time_tracking
+  private def time_tracking_outliers_replaced
     time_tracking.merge(outliers.transform_values {|_| median_value})
   end
 
-  def outliers
+  private def outliers
     @outliers ||= time_tracking.select {|_, v| v > outlier_threshold}
   end
 
--- a/services/QuillLMS/app/lib/time_tracking_cleaner.rb
+++ b/services/QuillLMS/app/lib/time_tracking_cleaner.rb
@@ -11,7 +11,6 @@ class TimeTrackingCleaner < ApplicationService
     @time_tracking_edits = data_params&.fetch(ActivitySession::TIME_TRACKING_EDITS_KEY, {})
   end
 
-
   # replace outlier time_tracking values with the list's median value
   # if they are 40x larger than the median
   # keep track of original data in data['time_tracking_edits']
--- a/services/QuillLMS/app/lib/time_tracking_cleaner.rb
+++ b/services/QuillLMS/app/lib/time_tracking_cleaner.rb
@@ -1,0 +1,38 @@
+class TimeTrackingCleaner
+  TIME_TRACKING_KEY = 'time_tracking'
+  EDITS_KEY = 'time_tracking_edits'
+  OUTLIER_MULTIPLIER = 40
+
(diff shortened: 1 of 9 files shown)


brendanshean (approved):
Nice work on this.  This seems like a reasonable stopgap.   If more robust outlier detection is desired, one approach would be to compute z-scores on the set of values and identifying `|z_score| > 3`.  (This means normalizing all the data and then finding values that are greater than 3 standard deviations from the mean).

emilia-friedberg (approved):
I really like this solution. Thank you for coming up with that.

dandrabik:
> Nice work on this. This seems like a reasonable stopgap. If more robust outlier detection is desired, one approach would be to compute z-scores on the set of values and identifying `|z_score| > 3`. (This means normalizing all the data and then finding values that are greater than 3 standard deviations from the mean).

I had initially dismissed some statistical methods like `z-score`/`t-score` since the sample size is so small (most are below 10 samples, some are as low as 5 samples) and the outliers are so different that it distorts the mean a lot, e.g. the screen shots below, but there may be ways to use them. 

I'm definitely interested in ways to make this more robust long-term. We do have a population of these answers, so we could construct tables as reference. Also `50x` the median is a huge outlier (it's a way to step our toe into this), but I could see us lowering that. I might ping you to pick your brain on some of these approaches.

Screen Shot 2022-04-07 at 10 25 46 AM
Screen Shot 2022-04-07 at 10 36 49 AM

brendanshean:
In the example given the z-scores are:
Screen Shot 2022-04-07 at 11 01 28 AM

Which means you could make the band tighter like |z_score| > 2.
Or like you mention construct a reference table for mu and sigma when computing z-scores.

dandrabik:
So, I've changed this PR a bit, so re-requesting reviews. @emilia-friedberg @brendanshean these are the changes

1. The replacement/calculation logic is basically the same, `40x median`. I changed it from 50x to 40x after looking at some data. I'm open to using a different calculation like z-score, but I'd like to do a little more testing on that, if we go that route (it would be easy to swap in), and I think it's better to get this version out.
2. Instead of just editing the `activity_session.timespent` calculation, this edits the granular `activity_session.data['time_tracking']` data so that data will match the sum in `activity_session.timespent`. This feels more future proof, in case we use the granular data (we already use it internally). Note, I record any edits in `activity_session.data['time_tracking_edits']` in case we need to audit.

emilia-friedberg (approved):

services/QuillLMS/spec/models/activity_session_spec.rb
```diff
@@ -864,9 +864,25 @@
...
+
+    it 'should return nil for nil' do
+      expect(ActivitySession.calculate_timespent(nil, nil)).to be_nil
+    end
+
+    it 'should return nil for a string' do
+      expect(ActivitySession.calculate_timespent(nil, 'hello')).to be_nil
+    end
+
+    it 'should return timestamp of session passed in' do
```
brendanshean: should this be 'timespent'

services/QuillLMS/app/models/activity_session.rb
```diff
@@ -158,8 +165,18 @@ def finished?
...
-  def self.calculate_timespent(time_tracking)
-    time_tracking&.values&.compact&.sum
+  def self.time_tracking_sum(time_tracking)
+    return nil unless time_tracking.respond_to?(:values)
+
+    time_tracking.values.compact.sum
+  end
+
+  def self.calculate_timespent(activity_session, time_tracking)
+    timespent = activity_session&.timespent || time_tracking_sum(time_tracking)
```
brendanshean, line 175: I'm not sure I understand the disjunction here.  Assuming, the there is cleaned data, why would we prefer `activity_session&.timespent` over `time_tracking_sum(time_tracking)`

services/QuillLMS/app/lib/time_tracking_cleaner.rb
```diff
@@ -0,0 +1,43 @@
...
+
+  def initialize(data_params)
+    @data_params = data_params
+    @time_tracking = data_params&.fetch(ActivitySession::TIME_TRACKING_KEY, nil)
+  end
+
+
+  # replace outlier time_tracking values with the list's median value
+  # if they are 40x larger than the median
+  # keep track of original data in data['time_tracking_edits']
```
brendanshean, line 19: Tracking edits is a great idea.

brendanshean (approved):
This feels like an improvement from the previous commit.  I have one question concerning the `calculate_timespent` method.  Otherwise, looks good to go.

I think using the median makes sense for now, but long term we should look into utilize population statistics for outlier detection. 

On a related note, even without using them for z-score calculations, I think it would be useful to compute landing(μ, σ), prompt_1(μ, σ), prompt_2(μ, σ), etc. as that might provide some interesting insight to individual items.

services/QuillLMS/spec/models/activity_session_spec.rb
dandrabik: Ah, nice catch, will update.

services/QuillLMS/app/models/activity_session.rb
dandrabik: So this is copying the existing code (and just moving it to the model). I agree, this might be better to change, but I think there may be some reason this was in place (time_tracking is new-ish, so I think it's to avoid overwriting older data). Anyway, I'm aiming to keep this changeset as small as possible, so will skip this potential refactor for now.

PR #9037, opened 2022-04-06
Add new columns to Districts table and new Districts behavior

cissyxyu:
## WHAT
Add a bunch of new fields to the districts table.
Establish a one-to-many relationship between Districts and Schools.
Establish a many-to-many relationship between Districts and Users through a new DistrictsAdmins table.
Create a staff panel in the Staff CMS to edit and see District info.
Create behavior that allows users to become District Admins, and hooks to associate District Admins with the schools in that district through SchoolsAdmins.

## WHY
This is part of a new implementation of Districts logic to make our codebase represent School Districts more robustly.

## HOW
Some migrations to create the new Districts and DistrictsAdmins tables.
Front end files to create Staff views for Districts.
The main logic of this PR is in the Districts -> SchoolsAdmins behavior. 
When Quill staff creates a `DistrictsAdmins` relationship, we want to create `SchoolsAdmins` records for every school with a subscription that belongs to that district.
Conversely, when a user stops being a district admin, we want to destroy all admin relationships between them and the schools in that district.
Relatedly, when a school buys a subscription, we want all District Admins for that school's district to become School Admins for that school. And when they lose their subscription, we want the district admins to stop being admins for that school.

### Screenshots
Screen Shot 2022-04-06 at 4 31 56 PM
Screen Shot 2022-04-06 at 4 33 52 PM

### Notion Card Links
https://www.notion.so/quill/Create-CMS-Staff-panels-for-Districts-ed62ccd4d4f144d5a975464fa087eba5
https://www.notion.so/quill/Add-new-fields-to-districts-table-9ee36231f8874d66b6a400d411f94720

PR Checklist | Your Answer
------------ | -------------


services/QuillLMS/app/models/school.rb
```diff
@@ -135,6 +139,15 @@ def self.school_year_start(time)
     time.month >= SCHOOL_YEAR_START_MONTH ? time.beginning_of_year + HALF_A_YEAR : time.beginning_of_year - HALF_A_YEAR
   end
 
+  def detach_from_existing_district_admins(district)
+    return unless district.present? && district.admins.count > 0
+
+    district.admins.each do |admin|
+      school_admin = SchoolsAdmins.find_by(school: self, user: admin)
+      school_admin&.destroy
+    end
```
anathomical: I think you can refactor this to remove the loop: `admins.where(user: district.admins).destroy_all`. When you pass an array in as the value of a `where` clause pair, ActiveRecord will do an `IN` clause in SQL.

services/QuillLMS/app/models/school.rb
```diff
@@ -165,4 +178,20 @@ def self.school_year_start(time)
...
+      detach_from_existing_district_admins(previous_district)
+    end
+
+    return unless district_id.present?
+
+    new_district = District.find(district_id)
+    new_district_admins = new_district&.admins || []
+    new_district_admins.each do |da|
+      SchoolsAdmins.find_or_create_by(school: self, user: da)
+    end
```
anathomical: I think you can simplify this with ActiveRecord:
```ruby
admins = new_district&.admins || []
save!
```

services/QuillLMS/app/models/subscription.rb
```diff
@@ -238,8 +238,15 @@ def save_if_charge_succeeds(premium_type, school=nil)
...
 
+  def detach_district_admins
+    schools.each do |school|
+      school.detach_from_existing_district_admins(school.district)
+    end
+  end
+
   def self.update_todays_expired_recurring_subscriptions
     expired_today_or_previously_and_recurring.each do |subscription|
+      subscription.detach_district_admins if OFFICIAL_SCHOOL_TYPES.include?(subscription.account_type)
```
anathomical: This looks like it works, but I'm not clear on why we're doing it.  We want to detach admins from schools whenever their subscription renews?

services/QuillLMS/db/migrate/20220328165932_add_columns_to_districts.rb
```diff
@@ -0,0 +1,14 @@
...
+
+class AddColumnsToDistricts < ActiveRecord::Migration[5.1]
+  def change
+    add_column :districts, :nces_id, :integer
+    add_column :districts, :city, :string
+    add_column :districts, :state, :string
+    add_column :districts, :zipcode, :string
+    add_column :districts, :phone, :string
+    add_column :districts, :total_students, :integer
+    add_column :districts, :total_schools, :integer
```
anathomical, line 11: Is this value different from the `count` of `School` records with the matching `district_id`?

services/QuillLMS/db/migrate/20220328165932_add_columns_to_districts.rb
```diff
@@ -0,0 +1,14 @@
+# frozen_string_literal: true
+
+class AddColumnsToDistricts < ActiveRecord::Migration[5.1]
+  def change
+    add_column :districts, :nces_id, :integer
+    add_column :districts, :city, :string
+    add_column :districts, :state, :string
+    add_column :districts, :zipcode, :string
+    add_column :districts, :phone, :string
+    add_column :districts, :total_students, :integer
```
anathomical, line 10: Is this dynamic?  How do we keep this up to date?

services/QuillLMS/spec/factories/districts.rb
```diff
@@ -4,12 +4,20 @@
...
+#  phone          :string
+#  state          :string
+#  token          :string
+#  total_schools  :integer
+#  total_students :integer
+#  zipcode        :string
+#  created_at     :datetime
+#  updated_at     :datetime
+#  clever_id      :string
+#  nces_id        :integer
```
anathomical: Is it worth adding some of these values to the factory’s generation code below?  Probably not necessary, but might be nice?

services/QuillLMS/app/models/districts_admins.rb
```diff
@@ -0,0 +1,47 @@
...
+#  updated_at  :datetime         not null
+#  district_id :bigint           not null
+#  user_id     :bigint           not null
+#
+# Indexes
+#
+#  index_districts_admins_on_district_id  (district_id)
+#  index_districts_admins_on_user_id      (user_id)
+#
+class DistrictsAdmins < ApplicationRecord
```
brendanshean: While the underlying table name is plural district_admins, the convention for Rails model naming is singular so this would be `DistrictsAdmin` (I think of the model name as being an individual resource for the table). The reason this is important is that there are various inflections that rails does when dealing with resources which assume a singular model name.

brendanshean (commented):
I will review more of this tomorrow morning, but I left a comment about the `DistrictAdmin` vs `DistrictAdmins` which I'd be curious what @dandrabik thinks.

services/QuillLMS/app/models/districts_admins.rb
brendanshean: While we already have `SchoolsAdmins` and `StudentsClassrooms`, there is `ClassroomsTeacher`.

services/QuillLMS/app/models/districts_admins.rb
```diff
@@ -0,0 +1,47 @@
...
+      school_admin = SchoolsAdmins.find_or_create_by(school: school, user: admin)
+      NewAdminEmailWorker.perform_async(user_id, school.id)
+    end
+  end
+
+  def detach_from_schools
+    schools_with_subscriptions.each do |school|
+      school_admin = SchoolsAdmins.find_by(school: school, user: admin)
+      school_admin&.destroy
+    end
```
anathomical: I think you can do this without the loop:
```ruby
SchoolsAdmins.where(school: schools_with_subscriptions, user: admin).destroy_all
```

services/QuillLMS/app/models/districts_admins.rb
```diff
@@ -0,0 +1,47 @@
...
+
+  after_create :attach_to_subscribed_schools
+  after_destroy :detach_from_schools
+
+  def admin
+    user
+  end
+
+  def attach_to_subscribed_schools
+    district = District.find(district_id)
```
anathomical: This should just be `district` per the `belongs_to` defined on line 19.  But I'm also not seeing it get used below so probably just needs to be deleted?

services/QuillLMS/app/models/districts_admins.rb
```diff
@@ -0,0 +1,47 @@
...
+    schools_with_subscriptions.each do |school|
+      school_admin = SchoolsAdmins.find_or_create_by(school: school, user: admin)
+      NewAdminEmailWorker.perform_async(user_id, school.id)
+    end
```
anathomical: You can create the records without the loop, but I think you need the loop for the worker calls.  Maybe...
```ruby
admin.administered_schools += schools_with_subscriptions
admin.save
schools_with_subscriptions.each { |school| NewAdminEmailWorker.perform_async(user_id, school.id) }
```

services/QuillLMS/spec/models/districts_admins_spec.rb
```diff
@@ -0,0 +1,56 @@
...
+  let(:subscription) { create(:subscription, account_type: 'School Paid') }
+  let(:admins) { create(:districts_admins, user: user, district: district) }
+
+  describe '#admin' do
+    it 'should return the user associated' do
+      expect(admins.admin).to eq(admins.user)
+    end
+  end
+
+  describe '#attach_to_subscribed_schools' do
```
anathomical: You may want a case here for situations where someone is already an admin for a school when this gets triggered?  Like if they were manually assigned or something like that?

services/QuillLMS/spec/models/school_spec.rb
```diff
@@ -160,4 +164,26 @@
       expect(School.school_year_start(time)).to eq(Date.parse('2019-07-01').beginning_of_day)
     end
   end
+
+  describe('district admin behavior') do
+    let(:school) { create(:school) }
+    let(:district) { create(:district) }
+    let(:admin) { create(:user)}
+
+    it 'creates new school admin record if a school is attached to a new district' do
```
anathomical, line 173: Again may be worth a test case for what happens if the admin is already attached to the school (manually or whatever) when this happens.

services/QuillLMS/app/controllers/cms/districts_controller.rb
```diff
@@ -0,0 +1,253 @@
...
+          districts.city,
+          districts.state,
+          districts.zipcode,
+          districts.total_students,
+          districts.total_schools,
+          districts.id
+        #{order_by_query_string}
+        #{pagination_query_string}
+      SQL
+    ).to_a
```
anathomical: I think we should generally try to avoid using raw SQL here if we can, and it seems pretty straightforward that we could.
```ruby
page = [district_query_params[:page].to_i - 1, 0].max
result = District.distinct.limit(DISTRICTS_PER_PAGE).offset(DISTRICTS_PER_PAGE * page)

sort = district_query_params[:sort]
sort_direction = district_query_params[:sort_direction]
if sort && sort_direction && sort != 'undefined' && sort_direction != 'undefined'
  result = result.order("#{sort} #{sort_direction}")
else
  result = result.order("total_students DESC")
end

result = add_where_conditions(result)
# then write `add_where_conditions` function that just does `result = result.where(...)` for each condition you already have logic for
```

anathomical (changes requested):
The overall shape of this looks good.  I've got a number of small nitpicks around efficiency that mostly boil down to "use more ActiveRecord tools".

services/QuillLMS/app/models/districts_admins.rb
dandrabik: Naming-wise, I think this should be `DistrictAdmin` and the table should be `district_admins`.

services/QuillLMS/app/controllers/cms/districts_controller.rb
```diff
@@ -0,0 +1,253 @@
...
+
+  def search
+    district_search_query = district_query_params
+    district_search_query_results = district_query(district_query_params)
+    district_search_query_results ||= []
+    number_of_pages = (number_of_districts_matched / DISTRICTS_PER_PAGE).ceil
+    render json: {numberOfPages: number_of_pages, districtSearchQueryResults: district_search_query_results}
+  end
+
+  def show
```
dandrabik, line 26: If you are using a Rails `erb` view, you don't need to make this a hash. You can just use `@district.name` in the view.

services/QuillLMS/app/controllers/cms/districts_controller.rb
```diff
@@ -0,0 +1,253 @@
...
+
+class Cms::DistrictsController < Cms::CmsController
+  before_action :signed_in!
+
+  before_action :text_search_inputs, only: [:index, :search]
+  before_action :set_district, only: [
+    :show
+  ]
+
+  DISTRICTS_PER_PAGE = 30.0
```
dandrabik: This should be an integer.
```suggestion
  DISTRICTS_PER_PAGE = 30
```

services/QuillLMS/app/controllers/cms/districts_controller.rb
```diff
@@ -0,0 +1,253 @@
+# frozen_string_literal: true
+
+class Cms::DistrictsController < Cms::CmsController
+  before_action :signed_in!
```
dandrabik: This line isn't need. The parent `Cms::CmsController` has a `before_action :staff!` check.
```suggestion

```

services/QuillLMS/app/controllers/cms/districts_controller.rb
```diff
@@ -0,0 +1,253 @@
...
+  end
+
+  def add_admin_by_email
+    begin
+      user = User.find_by(email: params[:email_address])
+      district = District.find(params[:id])
+      DistrictsAdmins.create(user_id: user.id, district_id: district.id)
+      flash[:success] = "Yay! It worked! 🎉"
+      redirect_to cms_district_path(params[:id])
+    rescue
```
dandrabik: This will only rescue if the district isn't found. Use `create!` if you'd like it to `raise`

services/QuillLMS/app/models/district.rb
```diff
@@ -4,14 +4,29 @@
...
 #
 class District < ApplicationRecord
 
-  has_and_belongs_to_many :users
+  has_many :schools
+  has_many :districts_admins, class_name: 'DistrictsAdmins'
+  has_many :admins, through: :districts_admins, source: :user
+
+  def total_invoice
+    schools.map { |s| s&.subscription&.payment_amount || 0 }.inject(0, :+)
```
dandrabik: This could be:
```suggestion
    schools.sum { |s| s&.subscription&.payment_amount || 0 }
```

services/QuillLMS/app/controllers/cms/districts_controller.rb
```diff
@@ -0,0 +1,253 @@
...
+          districts.total_students,
+          districts.total_schools,
+          districts.id
+        #{order_by_query_string}
+        #{pagination_query_string}
+      SQL
+    ).to_a
+  end
+
+  # rubocop:disable Metrics/CyclomaticComplexity
```
dandrabik: Let's avoid adding new methods that have high CyclomaticComplexity. Let's see if we can refactor this to something more simple.

dandrabik (changes requested):
This is a great first start! 

Since it involves adding new models, controllers, and views, I think this would be a good opportunity to walk this PR with someone and talk through Rails conventions and when to lean on Rails and when to write custom code. It would be a good learning opportunity.

 I'll schedule you some time with someone to do that.

services/QuillLMS/app/controllers/cms/districts_controller.rb
```diff
@@ -0,0 +1,253 @@
...
+      {
+        name: admin.user.name,
+        email: admin.user.email,
+        district_id: admin.district_id,
+        user_id: admin.user_id
+      }
+    end
+  end
+
+  def new_admin
```
cissyxyu, line 31: Note to self: write this as nested resource https://www.digitalocean.com/community/tutorials/how-to-create-nested-resources-for-a-ruby-on-rails-application

services/QuillLMS/app/models/subscription.rb
cissyxyu: I changed my approach here and made a new method to specifically deal with updating school subscriptions that have expired (see the new changes to this file).

services/QuillLMS/db/migrate/20220328165932_add_columns_to_districts.rb
cissyxyu: Yes, this is an imported value from the NCES dataset. It has no relation to the number of schools we have in our database.

services/QuillLMS/db/migrate/20220328165932_add_columns_to_districts.rb
cissyxyu: This is also a static number imported from the NCES dataset. We will update it annually when we do our annual data sync with national NCES data.

services/QuillLMS/app/controllers/cms/district_admins_controller.rb
```diff
@@ -0,0 +1,39 @@
+# frozen_string_literal: true
+
+class Cms::DistrictAdminsController < Cms::CmsController
+  before_action :get_district, only: [:create]
+  before_action :set_district_admin, only: [:destroy]
+
+  def create
+    @district_admin = @district.district_admins.build({user_id: User.find_by(email: district_admin_params)&.id})
```
brendanshean: I should add a cop here, but the braces, but a Rails convention is to omit braces for hash parameters if it is the last argument in a method call.  Compare with the `redirect_back` calls in lines 17, 23

services/QuillLMS/app/views/cms/districts/show.html.erb
```diff
@@ -0,0 +1,104 @@
+<div class='container'>
+  <article class='simple-rounded-box cms-container'>
+    <%= link_to 'District Directory', cms_districts_path %> > <%= @district['Name'] %>
+    <br /><br />
+    <h2><%= @district['Name'] %></h2>
```
brendanshean: Should this be `@district.name?`

services/QuillLMS/app/views/cms/districts/show.html.erb
```diff
@@ -0,0 +1,104 @@
+<div class='container'>
+  <article class='simple-rounded-box cms-container'>
+    <%= link_to 'District Directory', cms_districts_path %> > <%= @district['Name'] %>
```
brendanshean: Should this be @district.name?

services/QuillLMS/config/routes.rb
```diff
@@ -597,6 +597,17 @@
...
 
+    resources :districts do
+      collection do
+        post :search
+        get :search, to: 'districts#index'
+      end
+      member do
+        get :new_admin
+      end
+      resources :district_admins
```
brendanshean: Based on the controller actions, this should probably be  `resources :district_admins, only: [:create, :destroy]`

brendanshean (approved):
This looks much better.  I especially like the cleaned up ActiveRecord queries as well as the skinny controller actions.  Nice work Cissy!

services/QuillLMS/app/controllers/cms/district_admins_controller.rb
```diff
@@ -0,0 +1,37 @@
...
+class Cms::DistrictAdminsController < Cms::CmsController
+  before_action :set_district, only: [:create]
+  before_action :set_district_admin, only: [:destroy]
+
+  def create
+    @district_admin = @district.district_admins.build(user_id: User.find_by(email: district_admin_params)&.id)
+
+    begin
+      @district_admin.save!
+      flash[:success] = "Yay! It worked! 🎉"
```
dandrabik, line 12: 🎉

services/QuillLMS/app/controllers/cms/district_admins_controller.rb
```diff
@@ -0,0 +1,37 @@
...
+    rescue
```
dandrabik: Note, if you aren't doing anything with the exception, you might want to use the fact that `.save` (with the `!`) returns a boolean based on whether the save worked, e.g. this is equivalent code.
```ruby
if @district_admin.save
  flash[:success] = "Yay! It worked! 🎉"
else
  flash[:error] = "It didn't work! 😭😭😭"
end 
```
This scopes the branching specifically to whether the `save` worked, whereas the `rescue` would rescue ANY error within the save code, which could cover up a bug in a save logic. You can scope the rescue to this to avoid that.

```ruby
begin
  @district_admin.save!
  flash[:success] = "Yay! It worked! 🎉"
rescue ActiveRecord::RecordInvalid
  flash[:error] = "It didn't work! 😭😭😭"
end
```

services/QuillLMS/app/controllers/cms/district_admins_controller.rb
```diff
@@ -0,0 +1,37 @@
...
+      flash[:error] = "It didn't work! 😭😭😭"
+    end
+
+    redirect_back(fallback_location: cms_district_path(@district))
+  end
+
+  def destroy
+    flash[:error] = 'Something went wrong.' unless @district_admin.destroy
```
dandrabik: Hmm, this would populate both `flash[:error]` and `flash[:success]` if there was an issue with destroy? I think you'd want something like this.
```ruby
if @district_admin.destroy
 flash[:success] = 'Success! 🎉'
else
  flash[:error] = 'Something went wrong.'
end

File: services/QuillLMS/app/controllers/cms/districts_controller.rb [hunk 28]
```diff
@@ -0,0 +1,117 @@
...
+    @district_search_query = {}
+    @district_search_query_results = []
+    @number_of_pages = 0
+  end
+
+  def search
+    district_search_query = district_query_params
+    district_search_query_results = district_query(district_query_params)
+    district_search_query_results ||= []
+    number_of_pages = (district_search_query_results.count / DISTRICTS_PER_PAGE).ceil
```
dandrabik: You need a `to_f` here. Integer/Integer division in Ruby counts the number of whole items (it's confusing, but you usually want to avoid dividing an integer by an integer).
```suggestion
    number_of_pages = (district_search_query_results.count / DISTRICTS_PER_PAGE.to_f).ceil

```
Screen Shot 2022-04-14 at 4 47 05 PM

services/QuillLMS/app/controllers/cms/districts_controller.rb
```diff
@@ -0,0 +1,117 @@
...
+  private def edit_or_add_district_params
+    params.require(:district).permit(:id, editable_district_attributes.values)
+  end
+
+  private def school_query
+    @district.schools.select('schools.name, schools.id, subscriptions.account_type, count(schools_users.id) as number_teachers, count(schools_admins.id) as number_admins')
+      .joins(:schools_users)
+      .joins(:schools_admins)
+      .joins(school_subscription: :subscription)
+      .where('subscriptions.expiration > ? AND subscriptions.start_date <= ?', Date.today, Date.today)
```
dandrabik: Let's use Date.current to avoid local timezones.
```suggestion
      .where('subscriptions.expiration > ? AND subscriptions.start_date <= ?', Date.current, Date.current)
```

services/QuillLMS/app/models/user.rb
```diff
@@ -90,7 +90,9 @@ class User < ApplicationRecord
   has_many :schools_i_authorize, class_name: 'School', foreign_key: 'authorizer_id'
 
   has_many :schools_admins, class_name: 'SchoolsAdmins'
+  has_many :district_admins, class_name: 'DistrictAdmin', dependent: :destroy
```
dandrabik: I thiiink the `class_name` isn't needed here (You are using the standard naming pattern which is the default).

services/QuillLMS/app/models/district.rb
```diff
@@ -4,14 +4,35 @@
...
+  has_many :admins, through: :district_admins, source: :user
+
+  scope :by_name, ->(name) { where('lower(name) LIKE ?', "%#{name.downcase}%") }
+  scope :by_city, ->(city) { where('lower(city) LIKE ?', "%#{city.downcase}%") }
+  scope :by_state, ->(state) { where(:state => state.upcase) }
+  scope :by_zipcode, ->(zipcode) { where(:zipcode => zipcode) }
+  scope :by_nces_id, ->(nces_id) { where(:nces_id => nces_id) }
+
+  def total_invoice
+    schools.sum { |s| s&.subscription&.payment_amount || 0 } / 100
```
dandrabik: Another integer division item to watch out for. Make the denominator a float to avoid.
```suggestion
    schools.sum { |s| s&.subscription&.payment_amount || 0 } / 100.0
```

services/QuillLMS/client/app/bundles/Teacher/startup/CmsDistrictIndexAppClient.tsx
```diff
@@ -0,0 +1,236 @@
...
+      </div>
+    )
+  }
+
+  renderTableOrLoading() {
+    const { loading, data, columns, query } = this.state
+
+    if (loading) {
+      return <LoadingIndicator />
+    } else if (data && data.length) {
```
dandrabik: You don't need else if since the above returns.
```suggestion
    } 
    
    if (data && data.length) {
```

services/QuillLMS/client/app/bundles/Teacher/startup/CmsDistrictIndexAppClient.tsx
```diff
@@ -0,0 +1,236 @@
...
+            showPagination={false}
+            showPaginationBottom={false}
+            showPaginationTop={false}
+          />
+          <div className='cms-pagination-container'>
+            {this.renderPageSelector()}
+          </div>
+        </div>
+      )
+    } else {
```
dandrabik: Take or leave, but consider moving this clause to the top to have less branching, e.g. a few "guard" clauses, then the code.

```javascript
if (loading) { return <LoadingIndicator />}

if (!data || data.length === 0) { return <p>No records found for your query.</p> }

// The rest of the happy path code with no branching.

File: services/QuillLMS/spec/controllers/cms/districts_controller_spec.rb [hunk 34]
```diff
@@ -0,0 +1,105 @@
...
+  describe "DISTRICTS_PER_PAGE" do
+    it 'should have the correct value' do
+      expect(described_class::DISTRICTS_PER_PAGE).to eq 30.0
+    end
+  end
+
+  describe '#index' do
+    let(:district_hash) { {district_zip: "1234", district_name: 'Test District', district_city: 'Test City'} }
+
+    before { allow(RawSqlRunner).to receive(:execute) { [district_hash] } }
```
dandrabik: Is `RawSqlRunner` still used? This might have been refactored out.

services/QuillLMS/spec/controllers/cms/districts_controller_spec.rb
```diff
@@ -0,0 +1,105 @@
...
+      expect(assigns(:district_search_query_results)).to eq []
+      expect(assigns(:number_of_pages)).to eq 0
+    end
+  end
+
+  describe '#search' do
+    let!(:district) { create(:district, name: "Test District") }
+    let(:district_hash) { {id: district.id, name: district.name, nces_id: district.nces_id, city: district.city, state: district.state, zipcode: district.zipcode, phone: district.phone, total_students: district.total_students, total_schools: district.total_schools } }
+
+    before { allow(RawSqlRunner).to receive(:execute).and_return([district_hash]) }
```
dandrabik: I think this is removable.

services/QuillLMS/app/controllers/cms/districts_controller.rb
```diff
@@ -0,0 +1,117 @@
...
+    sort = district_query_params[:sort]
+    sort_direction = district_query_params[:sort_direction]
+    if sort && sort_direction && sort != 'undefined' && sort_direction != 'undefined'
+      result = result.order("#{sort} #{sort_direction}")
+    else
+      result = result.order("total_students DESC")
+    end
+
+    result = add_where_conditions(result).select(:id, :name, :city, :state, :zipcode, :phone, :total_students, :total_schools, :nces_id)
+    # then write `add_where_conditions` function that just does `result = result.where(...)` for each condition you already have logic for
```
dandrabik: remove comment?

dandrabik (approved):
Nice work on this, this is a big improvement!

I left some comments that you should look at, especially the two about Ruby integer division that you should definitely fix (it's a [bit of a gotcha](https://stackoverflow.com/questions/5502761/why-is-division-in-ruby-returning-an-integer-instead-of-decimal-value)  in Ruby). 

Approving since I don't think I need to re-review after you address my comments. Ping me if you have any questions and nice work!

services/QuillLMS/app/controllers/cms/districts_controller.rb
```diff
@@ -0,0 +1,117 @@
...
+  end
+
+  private def add_where_conditions(districts)
+    districts = districts.by_name(district_query_params[:district_name]) if district_query_params[:district_name].present?
```
happythenewsad, line 80: Very nice pattern.

services/QuillLMS/app/models/subscription.rb
```diff
@@ -152,6 +152,10 @@ def self.expired_today_or_previously_and_recurring
     Subscription.where('expiration <= ? AND recurring IS TRUE AND de_activated_date IS NULL', Date.today)
   end
 
+  def self.expired_today_or_previously_and_not_recurring
+    Subscription.where('expiration <= ? AND recurring IS FALSE AND de_activated_date IS NULL', Date.today)
```
happythenewsad: style nit: the latter two conditionals can be expressed in AREL syntax: `recurring: false...`

happythenewsad (approved):
Nice work!

PR #9041, opened 2022-04-07
fix bug with changing recurring status on subscriptions page

emilia-friedberg:
## WHAT
Fix bug with changing recurring status on subscriptions page.

## WHY
We want teachers to be able to do this.

## HOW
Just update it to use the value in state rather than a passed-in value, which was actually a mouse event.

### Screenshots

### Notion Card Links
https://www.notion.so/quill/Change-Plan-doesn-t-seem-to-be-working-3b3e40a93e32495cb68de41daac5c52f

PR Checklist | Your Answer
------------ | -------------


brendanshean (approved):
That was fast.  Thank you!

PR #9047, opened 2022-04-07
Quick reversion of a change that introduced a new bug

anathomical:
## WHAT
Revert the changes from #9028 that involve moving the cursor because they appear to be over-zealous and moving the cursor even when we don't want it to be moved
## WHY
The code in question looks like an attempt to address https://www.notion.so/quill/First-letter-of-response-sometimes-skipped-on-QE-activities-7d6e6218fee346668a2aa7a1e266b740

However, in addition to possibly fixing that, it also moves the cursor to the end of the entry text on every keystroke.  This makes it difficult to revise a response because if you're editing in the middle of the response, every keystroke moves the cursor to the end and you have to click to move the cursor back to where you are editing for each character you type.
## HOW
Basically just did a `git checkout` on the file to pre-merge, then manually made the changes in that file that don't seem relevant to this issue (they are around button references).

PR Checklist | Your Answer
------------ | -------------
Have you added and/or updated tests? |  No tests on this code

### DIFF
--- a/services/QuillLMS/client/app/bundles/Evidence/components/studentView/promptStep.tsx
+++ b/services/QuillLMS/client/app/bundles/Evidence/components/studentView/promptStep.tsx
@@ -48,6 +48,7 @@ export class PromptStep extends React.Component<PromptStepProps, PromptStepState
   private interval: any // eslint-disable-line react/sort-comp
   private button: any // eslint-disable-line react/sort-comp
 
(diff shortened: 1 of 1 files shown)


cissyxyu (approved):

dandrabik (approved):
:+1: If this is just a revision 🚢

PR #9049, opened 2022-04-08
check prompt param in FeedbackController

happythenewsad:
## WHAT
404 and return early if entry does not exist in FeedbackController::create 

## WHY
So we can error early, and so that malformed requests are more easily distinguished from legitimate bugs.

### Screenshots

### Notion Card Links
[(Please provide links to any relevant Notion card(s) relevant to this PR.)](https://www.notion.so/quill/Product-Board-30f97e4eb01246dbb8264253fb385073?p=5aef295dee3d44ecbc5472d79ddc9df6)

PR Checklist | Your Answer
------------ | -------------

### DIFF
--- a/services/QuillLMS/engines/evidence/app/controllers/evidence/feedback_controller.rb
+++ b/services/QuillLMS/engines/evidence/app/controllers/evidence/feedback_controller.rb
@@ -74,6 +74,8 @@ module Evidence
 
     private def set_params
       @entry = params[:entry]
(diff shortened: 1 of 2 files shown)


cissyxyu (approved):

anathomical (approved):
We should be 404ing missing entries anyway based on the `rescue` block below, but this seems reasonable as there's no point in going to the database if we don't even have an ID to look up.

PR #9050, opened 2022-04-08
individual react table instance upgrades pt 1

emilia-friedberg:
## WHAT
Create a new default `ReactTable` component that uses the hooks from the new version of `react-table` to mimic the behavior, requiring minimal (though still fairly extensive) code changes to match the previous version. Then start using that in the files where we are currently using `react-table-6`.

## WHY
Since the new `ReactTable` is all hooks based, we need to individually convert each component from using the old version to the new version. This is the first set of components to change over (all the internal-facing uses where I could confirm code parity).

## HOW
- Create the new custom component, adding behavior as necessary to create behavior parity with existing versions
- Replace the old imports with the new one and make tweaks to usage as necessary

### Screenshots

### Notion Card Links

PR Checklist | Your Answer
------------ | -------------


eadams17 (approved):
thanks for knocking all of this out !

cissyxyu (approved):

PR #9064, opened 2022-04-13
Replace Stripe Billings portal code with Stripe::Checkout::Session

brendanshean:
## WHAT
Update the mechanism for updating credit cards for Stripe subscription.

## WHY
Stripe's customer portal is a nice UI that allows users to update credit cards, subscriptions, etc.   Unfortunately, it requires the paid product "Billing" in order to function in live mode, but curiously runs with no warnings in test mode.

For now, we'll use a different approach using the Stripe::Checkout::Session api.

## HOW
1.  Remove the BillingPortalSessions controller since we don't want to use that API anymore.
2.  Create a new controller `SubscriptionPaymentMethodsController` that will handle the updating of payment methods.  On a related note, since this new [controller](https://github.com/empirical-org/Empirical-Core/blob/ce6f95b233a2333efe6e99dfa1bb5395acac48d8/services/QuillLMS/app/controllers/stripe_integration/subscription_payment_methods_controller.rb#L9) and the existing `CheckoutSessionsController` both [use](https://github.com/empirical-org/Empirical-Core/blob/ce6f95b233a2333efe6e99dfa1bb5395acac48d8/services/QuillLMS/app/controllers/stripe_integration/subscription_checkout_sessions_controller.rb#L8) `Stripe::Checkout::Session.create` calls, I'm renaming `CheckoutSessionsController` as `SubscriptionCheckoutSessionsController` to better differentiate the two.  
I also renamed `StripeCheckoutSessionButton` to `StripeSubscriptionCheckoutSessionButton` to further highlight this change.
3. We'll add in a [boolean](https://github.com/empirical-org/Empirical-Core/blob/ce6f95b233a2333efe6e99dfa1bb5395acac48d8/services/QuillLMS/app/controllers/subscriptions_controller.rb#L67) state `stripe_payment_method_updated` to let the Subscriptions frontend know when to expect updates to the payment method and then rely on a [similar](https://github.com/empirical-org/Empirical-Core/blob/ce6f95b233a2333efe6e99dfa1bb5395acac48d8/services/QuillLMS/app/services/stripe_integration/webhooks/setup_intent_succeeded_event_handler.rb#L16) Pusher [mechanism](https://github.com/empirical-org/Empirical-Core/blob/ce6f95b233a2333efe6e99dfa1bb5395acac48d8/services/QuillLMS/client/app/bundles/Teacher/containers/Subscriptions.jsx#L75) for handling those updates.
4. Add in a webhook event handler `PaymentIntentSucceededEventHandler` which will  set the default payment of the Subscription to this new payment method (see [#6](https://stripe.com/docs/payments/checkout/subscriptions/update-payment-details#set-default-payment-method) for details)

### Screenshots
Screen Shot 2022-04-13 at 2 43 34 PM

### Notion Card Links

PR Checklist | Your Answer
------------ | -------------
Spec Review: Have you reviewed the spec and ensured this meets requirements and/or matches design mockups? |  YES


services/QuillLMS/app/models/validators/stripe_uid_validator.rb
```diff
@@ -7,8 +7,10 @@ class StripeUidValidator < ActiveModel::EachValidator
     EVT = 'evt',
     IN = 'in',
     PLAN = 'plan',
+    PM = 'pm',
     PRICE = 'price',
     PROD = 'prod',
+    SETI = 'seti',
```
dandrabik, line 13: 👽 👽 👽 

https://www.seti.org/

Screen Shot 2022-04-13 at 3 42 58 PM

services/QuillLMS/client/app/bundles/Teacher/containers/Subscriptions.jsx
```diff
@@ -80,11 +95,28 @@ export default class Subscriptions extends React.Component {
     const channel = pusher.subscribe(channelName);
 
     channel.bind('stripe-subscription-created', () => {
+      debugger
```
dandrabik: Remove debugger
```suggestion
      
```

services/QuillLMS/client/app/bundles/Teacher/containers/Subscriptions.jsx
```diff
@@ -80,11 +95,28 @@ export default class Subscriptions extends React.Component {
...
       this.retrieveStripePurchasedConfirmation()
       pusher.unsubscribe(channelName)
     })
   }
 
+  initializePusherForStripeSubscriptionPaymentMethodUpdating() {
+    if (process.env.RAILS_ENV === 'development') { Pusher.logToConsole = true }
```
dandrabik: Take or leave, but should this be pulled out for release? 

I wonder if there is a way to set this for `development` in a config somewhere, e.g. something called by `npm build:dev:client`

dandrabik (approved):
Looks good to me, thanks for the quick turnaround here. I also tested it on brendan.quill.org and editing a card worked for me.

Small note about removing a debugger comment, but otherwise 🚢 .

services/QuillLMS/client/app/bundles/Teacher/containers/Subscriptions.jsx
brendanshean: Good point.  I will remove.

PR #9069, opened 2022-04-14
individual react table instance upgrades pt 2

emilia-friedberg:
## WHAT
Upgrade all teacher- and admin-facing components, leaving just one final staff-facing component that I need to figure out how to convert still.

## WHY
The older version of `react-table` is no longer being maintained, and we use it heavily, so we want to be on the latest version.

## HOW
Manually upgrade and test every instance where we're using it.

### Screenshots

### Notion Card Links
https://www.notion.so/quill/Upgrade-teacher-and-admin-facing-instances-of-React-Table-to-use-latest-version-397cac3b380345dfb51980d909972bf9

PR Checklist | Your Answer
------------ | -------------

### DIFF
--- a/services/QuillLMS/client/app/bundles/Shared/components/reactTable/reactTablePagination.tsx
+++ b/services/QuillLMS/client/app/bundles/Shared/components/reactTable/reactTablePagination.tsx
@@ -132,9 +132,7 @@ export default class ReactTablePagination extends React.Component {
 
   render () {
     const {
-      // Computed
       pages,
-      // Props
       page,
       showPageSizeOptions,
       pageSizeOptions,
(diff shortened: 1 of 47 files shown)


cissyxyu (approved):
Thank you for doing all of this!!

services/QuillLMS/client/app/bundles/Shared/components/reactTable/reactTablePagination.tsx
```diff
@@ -0,0 +1,205 @@
...
+      inputType: this.state.page === '' ? 'text' : 'number',
+      pageJumpText: this.props.pageJumpText,
+    }
+  }
+
+  render () {
+    const {
+      // Computed
+      pages,
+      // Props
```
eadams17: can we get rid of this commented out code?

eadams17 (approved):

services/QuillLMS/client/app/bundles/Shared/components/reactTable/reactTablePagination.tsx
emilia-friedberg: I don't think it's commented out code, I think it's descriptive - but it's not actually clear to me what it means so I'll remove it 😆

PR #9071, opened 2022-04-15
new grammar bolding algorithm

happythenewsad:
## WHAT
1. Implements a new grammar highlighting algorithm that incorporates offsets. Without offsets, words that occur twice in the student entry would be bolded. We only want to bold the specific words designated by the grammar engine. 
2. Also refactors stripEvidenceHtml into a shared library.

## WHY
So that user-facing highlighting is correct. 

### Screenshots
Working examples after the fix:
Screen Shot 2022-05-03 at 1 27 47 PM
Screen Shot 2022-05-03 at 10 15 06 AM

### Notion Card Links
https://www.notion.so/quill/Product-Board-30f97e4eb01246dbb8264253fb385073?p=3ceb2291f228404197b124d7dd7f8923

PR Checklist | Your Answer
------------ | -------------

### DIFF
--- a/services/QuillLMS/client/app/bundles/Evidence/components/studentView/promptStep.tsx
+++ b/services/QuillLMS/client/app/bundles/Evidence/components/studentView/promptStep.tsx
@@ -105,8 +105,6 @@ export class PromptStep extends React.Component<PromptStepProps, PromptStepState
     return submittedResponses.map(r => r.entry).concat(text)
   }
 
-  stripHtml = (html: string) => html.replace(/<p>|<\/p>|<u>|<\/u>|<b>|<\/b>|<br>|<br\/>/g, '').replace(/&nbsp;/g, ' ')
-
   formattedPrompt = (submittedResponses?: Array<string>) => {
     if (submittedResponses && submittedResponses.length) {
       const lastSubmission = submittedResponses[submittedResponses.length - 1]
--- a/services/QuillLMS/client/app/bundles/Evidence/components/studentView/promptStep.tsx
+++ b/services/QuillLMS/client/app/bundles/Evidence/components/studentView/promptStep.tsx
@@ -6,7 +6,7 @@ import Feedback from './feedback'
 
 import EditCaretPositioning from '../../helpers/EditCaretPositioning'
 import ButtonLoadingSpinner from '../shared/buttonLoadingSpinner'
-import { highlightSpellingGrammar } from '../../libs/stringFormatting'
+import { highlightGrammar, highlightSpelling, stripEvidenceHtml } from '../../libs/stringFormatting'
 
 interface PromptStepProps {
   activityIsComplete: Boolean;
@@ -142,9 +142,9 @@ export class PromptStep extends React.Component<PromptStepProps, PromptStepState
       return str
     }
 
-    let wordsToFormat = lastSubmittedResponse.highlight.filter(hl => hl.type === PROMPT).map(hl => this.stripHtml(hl.text))
+    let wordsToFormat = lastSubmittedResponse.highlight.filter(hl => hl.type === PROMPT).map(hl => stripEvidenceHtml(hl.text))
     wordsToFormat = wordsToFormat.length === 1 ? wordsToFormat[0] : wordsToFormat
-    return highlightSpellingGrammar(str, wordsToFormat)
+    return highlightSpelling(str, wordsToFormat)
   }
 
   formatStudentResponse = (str: string) => {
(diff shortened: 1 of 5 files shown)


services/QuillLMS/client/app/bundles/Evidence/components/studentView/promptStep.tsx
```diff
@@ -6,7 +6,7 @@ import Feedback from './feedback'
 
 import EditCaretPositioning from '../../helpers/EditCaretPositioning'
 import ButtonLoadingSpinner from '../shared/buttonLoadingSpinner'
-import { highlightSpellingGrammar } from '../../libs/stringFormatting'
+import { highlightGrammar, highlightSpelling, stripEvidenceHtml } from '../../libs/stringFormatting'
```
anathomical, line 9: Any reason not to refactor `highlightSpellingGrammar` to handle offsets for all types instead of splitting behavior like this?  I haven't fully thought this question through, so it's possible that there's an obvious reason not to, but my first thought was "we should probably respect offsets on either type".

services/QuillLMS/client/app/bundles/Evidence/components/studentView/promptStep.tsx
```diff
@@ -105,8 +105,6 @@ export class PromptStep extends React.Component<PromptStepProps, PromptStepState
     return submittedResponses.map(r => r.entry).concat(text)
   }
 
-  stripHtml = (html: string) => html.replace(/<p>|<\/p>|<u>|<\/u>|<b>|<\/b>|<br>|<br\/>/g, '').replace(/&nbsp;/g, ' ')
```
anathomical, line 108: Very much approve of refactoring this out.  Good call.

services/QuillLMS/client/app/bundles/Evidence/libs/stringFormatting.test.ts
```diff
@@ -1,34 +1,79 @@
...
+
+    it('should not bold a phrase before the given character', () => {
+      const result = highlightGrammar('lorem ipsum ipsum', [{character: 8, text: "ipsum"}])
+      expect(result).toEqual('lorem ipsum <b>ipsum</b>')
+    });
+
+    it('should not bold a phrase after the given character', () => {
+      const result = highlightGrammar('lorem ipsum ipsum', [{character: 0, text: "ipsum"}])
+      expect(result).toEqual('lorem ipsum ipsum')
+    });
```
anathomical: Is this really the condition we want here?  I would have thought that this should highlight the first instance of "ipsum" it finds rather than nothing.  What am I missing?

anathomical (approved):
Generally looks good to me.  Two minor items: 1) Should we just have a single highlight function that handles both "spelling" and "grammar" cases?  Feels like we really want them to behave the same way...  2) I'm not sure one of the test cases makes sense to me.

services/QuillLMS/client/app/bundles/Evidence/components/studentView/promptStep.tsx
happythenewsad: We can't currently do this because the spelling payload does not include an offset (payload schemas differ between grammar and spelling): https://github.com/empirical-org/Empirical-Core/blob/develop/services/QuillLMS/engines/evidence/lib/evidence/evidence/spelling_check.rb#L69

services/QuillLMS/client/app/bundles/Evidence/libs/stringFormatting.test.ts
happythenewsad: Good question. Grammar engine responses are 1-indexed, not zero-indexed. I've updated this spec to reflect that. Hopefully it's clearer now.

happythenewsad: Example grammar engine offset payload: https://quill-org.postman.co/workspace/Quill-Team-Workspace~be8cf879-9fa6-4dc5-8544-630985cdf233/request/14211978-14b72239-9722-4049-9ab9-a774f4696afd

services/QuillLMS/client/app/bundles/Evidence/components/studentView/promptStep.tsx
anathomical: Fair point.  We could just assume that not having `highlight.character` just means an offset of `0` (start at the beginning of the string) or something.  That solves the problem for spelling and future-proofs it if we do decide to add support for it later.  But maybe that's more work than we want to do in this PR?  Up to you.

services/QuillLMS/client/app/bundles/Evidence/libs/stringFormatting.test.ts
anathomical: I think I'm still confused here.  The `it` text here suggests that if the offset is high enough, nothing should be highlighted, but I don't understand why nothing would be highlighted if the offset were _too low_.  Again, I'm probably missing something here, but can't figure out what.

services/QuillLMS/client/app/bundles/Evidence/components/studentView/promptStep.tsx
happythenewsad: Yeah, let's treat that as out of scope for this PR.

services/QuillLMS/client/app/bundles/Evidence/libs/stringFormatting.ts
```diff
@@ -10,3 +10,33 @@ export const highlightSpellingGrammar = (str: string, wordsToFormat: string | st
...
   return newString
 }
+
+export const highlightGrammar = (str: string, highlightArray, promptLength: number=0) => {
+  if (highlightArray.length < 1) { return str }
+
+  highlightArray.sort(descendingOffsetComparator)
+
+  highlightArray.forEach((highlight) => {
+    let offset = highlight.character - promptLength - 1
```
emilia-friedberg: This doesn't get reassigned, so it should be a `const`.

services/QuillLMS/client/app/bundles/Evidence/libs/stringFormatting.ts
```diff
@@ -10,3 +10,33 @@ export const highlightSpellingGrammar = (str: string, wordsToFormat: string | st
...
+    let stringAfterOffset = str.slice(offset)
+    let stringBeforeOffset = str.slice(0, offset)
```
emilia-friedberg: Same here.

emilia-friedberg (approved):

services/QuillLMS/client/app/bundles/Evidence/libs/stringFormatting.test.ts
happythenewsad: (TR and I resolved this over zoom)

PR #9073, opened 2022-04-18
Have overnight pre_cache use the DEFAULT Sidekiq queue.

dandrabik:
## WHAT
Change the Sidekiq queue for overnight pre_cache `FindAdminUsersWorker` jobs from `critical` to `default` 
## WHY
These jobs are long-running and clog up the `critical` and `critical_external` queue. This is bad since it blocks `critical` jobs by users who are currently on the site (like importing a classroom), in favor of pre-caching for users who aren't currently on the site.

We have alerts that go off of every night about this. These are legitimate alerts flagging that users can't do things like import classrooms. This should "silence" those alerts by fixing that bad scenario, and allowing `critical` jobs to go through.

## HOW
You can override the queue for a job when you call it.
### Screenshots
Screen Shot 2022-04-18 at 5 31 52 PM

Screen Shot 2022-04-18 at 5 27 19 PM

PR Checklist | Your Answer
------------ | -------------


happythenewsad (approved):
👍

brendanshean (approved):
This sounds like a sensible improvement.

anathomical (approved):

PR #9075, opened 2022-04-19
LMS/ add SalesFormSubmissions table

eadams17:
## WHAT
add SalesFormSubmissions table

## WHY
we want to store these submissions in our database

## HOW
add migration and model for `SalesFormSubmission` with necessary columns

### Screenshots

### Notion Card Links
https://www.notion.so/quill/Capture-opportunities-from-forms-b2090e97282f4953973e470615404aa5

PR Checklist | Your Answer
------------ | -------------


services/QuillLMS/app/models/sales_form_submission.rb
```diff
@@ -0,0 +1,16 @@
+# frozen_string_literal: true
+
+class SalesFormSubmission < ApplicationRecord
+  validates :first_name, presence: true
+  validates :last_name, presence: true
+  validates :email, presence: true
```
brendanshean, line 36: What about adding some kind of email validation [similar](https://github.com/empirical-org/Empirical-Core/blob/develop/services/QuillLMS/app/models/user.rb#L141) to our User model:

services/QuillLMS/app/models/sales_form_submission.rb
```diff
@@ -0,0 +1,16 @@
...
+  validates :phone_number, presence: true
+  validates :zipcode, presence: true
+  validates :collection_type, presence: true
```
brendanshean: Are the :collection_type values going to be fixed (e.g. selected from a drop down)?  If so, it might be worth adding that [restriction](https://github.com/empirical-org/Empirical-Core/blob/1bb95b65638d1c784cf2d23a7e074c11ff7fd99e/services/QuillLMS/app/models/plan.rb#L41) now.

services/QuillLMS/app/models/sales_form_submission.rb
```diff
@@ -0,0 +1,16 @@
...
+  validates :school_name, presence: true
+  validates :district_name, presence: true
+  validates :school_premium_count_estimate, presence: true
+  validates :teacher_premium_count_estimate, presence: true
+  validates :student_premium_count_estimate, presence: true
+  validates :submission_type, presence: true
```
brendanshean: Are the :submission_type values going to be fixed (e.g. selected from a drop down)?  If so, it might be worth adding that [restriction](https://github.com/empirical-org/Empirical-Core/blob/1bb95b65638d1c784cf2d23a7e074c11ff7fd99e/services/QuillLMS/app/models/plan.rb#L41) now.

services/QuillLMS/db/migrate/20220418234207_create_sales_form_submissions.rb
```diff
@@ -0,0 +1,23 @@
...
+    create_table :sales_form_submissions do |t|
+      t.string :first_name, null: false
+      t.string :last_name, null: false
+      t.string :email, null: false
+      t.string :phone_number, null: false
+      t.string :zipcode, null: false
+      t.string :collection_type, null: false
+      t.string :school_name, null: false
+      t.string :district_name, null: false
+      t.integer :school_premium_count_estimate, null: false
```
brendanshean: For the three integer fields I might opt for a stronger condition: `default: 0`.  This makes constructing valid objects easier since it's no longer necessary to specify those values, while still preserving `null: false`.

services/QuillLMS/app/models/sales_form_submission.rb
```diff
@@ -0,0 +1,16 @@
...
+class SalesFormSubmission < ApplicationRecord
+  validates :first_name, presence: true
+  validates :last_name, presence: true
+  validates :email, presence: true
+  validates :phone_number, presence: true
+  validates :zipcode, presence: true
+  validates :collection_type, presence: true
+  validates :school_name, presence: true
+  validates :district_name, presence: true
+  validates :school_premium_count_estimate, presence: true
```
brendanshean: For the three integer fields, what about "validates [FIELD_NAME], numericality: { greater_than_or_equal_to: 0 }`

brendanshean (approved):
I left some notes about adding validations for some of the attributes which may or may not be of value now.  Otherwise, looks good to go.

services/QuillLMS/app/models/sales_form_submission.rb
eadams17: thanks for all of the validation suggestions! I'll add those in, as well as adding default values for the numerical fields

dandrabik (approved):
:+1: Looks good to me!

PR #9081, opened 2022-04-19
make data table headers accessible

emilia-friedberg:
## WHAT
Make data table headers accessible by not using buttons where there is no text/action, and generally using a containing `<th>` tag so it's accessible to screenreaders.

## WHY
This was something that confused our accessibility auditor when we were doing our walkthrough.

## HOW
Only generate button elements when there is an action to take, and wrap all header elements in `<th>` tags so the screenreader can identify them properly.

### Screenshots
No visible change.

### Notion Card Links
https://www.notion.so/quill/Replace-unlabeled-button-for-columns-without-names-in-DataTable-component-b855ba48d2f44ad8adc9c02bb7651b83

PR Checklist | Your Answer
------------ | -------------
Have you added and/or updated tests? | YES - snapshots


brendanshean (approved):
Thanks for cleaning this up.  I think `<th>` wrappers make the code more clear in general.

eadams17 (approved):

PR #9087, opened 2022-04-20
add screenreader instructions to all activity types

emilia-friedberg:
## WHAT
Add instructions that are only visible to tab-navigators and screenreader users to tell screenreader users to use their reading keys.

## WHY
This was highlighted by our accessibility auditor as a nice-to-have that would help make sure students weren't confused by missing informational content when tab-navigating.

## HOW
Just add a new element that is only visible on focus (but can still be read by screenreaders if already using reading keys) that tells screenreader users to use reading keys so they don't skip past important content.

### Screenshots

### Notion Card Links
https://www.notion.so/quill/Add-screenreader-legible-command-at-top-of-activity-pages-to-tell-students-to-use-their-reading-keys-497c53db2e56455c8212cdaddaa93162

PR Checklist | Your Answer
------------ | -------------

### DIFF
--- a/services/QuillLMS/client/app/bundles/Diagnostic/components/home.tsx
+++ b/services/QuillLMS/client/app/bundles/Diagnostic/components/home.tsx
@@ -7,7 +7,7 @@ import TeacherNavbar from './navbar/teacherNavbar';
 
 import { routes } from "../routes";
 import { getParameterByName } from '../libs/getParameterByName';
-import { TeacherPreviewMenu } from '../../Shared/components/shared/teacherPreviewMenu';
+import { TeacherPreviewMenu, ScreenreaderInstructions, } from '../../Shared/index';
 import { fetchUserRole } from '../../Shared/utils/userAPIs';
 import { addKeyDownListener } from '../../Shared/hooks/addKeyDownListener';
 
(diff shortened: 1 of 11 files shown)


services/QuillLMS/client/app/bundles/Diagnostic/components/home.tsx
```diff
@@ -79,7 +79,8 @@ export const Home = () => {
           />
         </aside>}
         <main style={{ height: '100vh', overflow: 'auto' }}>
-          {isPlaying && !isTeacherOrAdmin && <button className="skip-main" onClick={handleSkipToMainContentClick} type="button">Skip to main content</button>}
+          <ScreenreaderInstructions />
+          <button className="skip-main" onClick={handleSkipToMainContentClick} type="button">Skip to main content</button>
```
cissyxyu, line 83: Are we removing the two nil checks on purpose here?

cissyxyu (approved):

eadams17 (approved):

services/QuillLMS/client/app/bundles/Diagnostic/components/home.tsx
emilia-friedberg: Yes - I'm not totally sure why we were here initially but this is the only place in five similar files we had these nil checks in place, and I don't think it's necessary.

PR #9092, opened 2022-04-21
Migrate legacy teacher subscription renewals to stripe

brendanshean:
## WHAT
Transition legacy subscriptions that are up for renewal over to Stripe Subscriptions

## WHY
It's easier to manage subscriptions within Stripe

## HOW
Old System Overview
1.  Subscription renewals initiated via cron job that calls `Subscription.update_todays_expired_recurring_subscriptions` [daily](https://github.com/empirical-org/Empirical-Core/blob/61d2ac68769d49fde911404b6c1261404b668690/services/QuillLMS/app/models/subscription.rb#L247).  
2. From there call the Stripe API and attempt a [charge](https://github.com/empirical-org/Empirical-Core/blob/61d2ac68769d49fde911404b6c1261404b668690/services/QuillLMS/app/models/subscription.rb#L280) for the subscription in `charge_user`
3.  If the charge on Stripe is successful, then our system [calls](https://github.com/empirical-org/Empirical-Core/blob/61d2ac68769d49fde911404b6c1261404b668690/services/QuillLMS/app/models/subscription.rb#L136) `renew_subscription`

New System Overview
1.  Same as original.
2. [renew_via_stripe](https://github.com/empirical-org/Empirical-Core/blob/5e9298270bf27049dd6f3ed4f1c6fe215f7b6f2f/services/QuillLMS/app/models/subscription.rb#L232) will attempt create the new subscription object on Stripe.
3. Stripe will send webhooks back, and if successful, `invoice.paid` will be sent which our system handles [here](https://github.com/empirical-org/Empirical-Core/blob/develop/services/QuillLMS/app/services/stripe_integration/webhooks/invoice_paid_event_handler.rb#L14) This will then create the corresponding subscription in our database.

NB:  In calculating the last_four, we have to add in a [conditional](https://github.com/empirical-org/Empirical-Core/blob/5e9298270bf27049dd6f3ed4f1c6fe215f7b6f2f/services/QuillLMS/app/models/stripe_integration/subscription.rb#L22) to check for older accounts that use Stripe `sources` as a payment method.  Stripe [recommends](https://stripe.com/docs/sources/cards) moving away from the sources API.

### Screenshots

### Notion Card Links
https://www.notion.so/quill/Implement-a-new-direct-subscription-process-that-utilizes-Stripe-subscriptions-229f2cba7d5e4dbf8bf95388f1ab4209

PR Checklist | Your Answer
------------ | -------------


cissyxyu (approved):

services/QuillLMS/spec/models/stripe_integration/subscription_spec.rb
```diff
@@ -3,62 +3,79 @@
...
+  include_context 'Stripe Invoice'
+
   let(:subscription) { create(:subscription, stripe_invoice_id: stripe_invoice_id) }
 
   describe '#last_four' do
     subject { described_class.new(subscription).last_four }
 
+    let(:retrieve_invoice) { allow(Stripe::Invoice).to receive(:retrieve).with(stripe_invoice_id) }
+    let(:retrieve_subscription) { allow(Stripe::Subscription).to receive(:retrieve).with(stripe_subscription_id) }
+    let(:retrieve_payment_method) { allow(Stripe::PaymentMethod).to receive(:retrieve).with(stripe_payment_method_id) }
```
anathomical, line 15: Hadn't seen this `let...allow` pattern before.  Neat!

anathomical (approved):
Looks good to me.

emilia-friedberg (approved):

services/QuillLMS/app/models/stripe_integration/subscription.rb
```diff
@@ -14,32 +16,38 @@ def stripe_subscription_id
...
+    end
+
+    private def stripe_payment_method_id
+      stripe_subscription.default_payment_method || stripe_customer.invoice_settings.default_payment_method
     end
 
-    private def stripe_payment_method
-      Stripe::PaymentMethod.retrieve(stripe_default_payment_method_id)
+    private def stripe_source
+      Stripe::PaymentMethod.list(customer: stripe_customer_id, type: 'card')&.data&.first
```
dandrabik: It might not matter if there should only be one card in the list, but if there could be multiple, should this be `.last` to get a newer card(I'm not 100% sure how they would be ordered, I'm guessing newest would be last, but that could be wrong)

services/QuillLMS/app/models/subscription.rb
```diff
@@ -93,7 +95,9 @@ class Subscription < ApplicationRecord
 
   validates :stripe_invoice_id, allow_blank: true, stripe_uid: { prefix: :in }
 
-  scope :active, -> { where(de_activated_date: nil).where("expiration > ?", Date.today).order(expiration: :asc) }
+  scope :active, -> { not_expired.where(de_activated_date: nil).order(expiration: :asc) }
+  scope :expired, -> { where('expiration <= ?', Date.today) }
```
dandrabik, line 99: Consider using `Date.current`

services/QuillLMS/app/models/subscription.rb
```diff
@@ -244,16 +228,27 @@ def detach_district_admins
     end
   end
 
+  def renew_via_stripe
+    return ErrorNotifier.report(RenewalNilStripeCustomer.new, subscription_id: id) unless purchaser&.stripe_customer?
```
dandrabik: Since you already have a `rescue` / `report` in the method, you could consider using:

```ruby
raise RenewalNilStripeCustomer unless purchaser&.stripe_customer?
# other code
rescue Stripe::InvalidRequestError, RenewalNilStripeCustomer => e

dandrabik (approved):
:+1: LGTM! Small take or leave comments, but nothing major. Thanks for putting this together!

services/QuillLMS/app/models/subscription.rb
brendanshean: I like this suggestion.  I was assuming that the error handling would be handled higher up, but this keeps it contained within the method which seems safer.

services/QuillLMS/app/models/stripe_integration/subscription.rb
brendanshean: This is a good point.  I was basically reusing [old](https://github.com/empirical-org/Empirical-Core/blob/f215d13bcb43cb26f56dc3bf853154acdba55ed6/services/QuillLMS/app/models/user.rb#L300) code for source, but what we really need here is `default_source` since that would resolve the case where there are two sources and the older card for whatever reason is the default one.

PR #9094, opened 2022-04-21
Lms/update deploy scripts

anathomical:
## WHAT
Update deploy scripts to no longer rely on Heroku's Github integration now that it's been indefinitely disabled
## WHY
We want our deploy scripts to work in order to simplify our workflows
## HOW
Replace the push to a targeted github branch that used to trigger deploys with a push to a custom Heroku git branch that will trigger deploys

### Notion Card Links
https://www.notion.so/quill/Update-LMS-and-CMS-deploy-scripts-to-use-updated-Heroku-deploy-process-b2f06036dbcf4cf0a1cc6312b0fce7d4

PR Checklist | Your Answer
------------ | -------------
Have you added and/or updated tests? |  No tests on deploy scripts

### DIFF
--- a/services/QuillCMS/deploy.sh
+++ b/services/QuillCMS/deploy.sh
@@ -32,7 +32,7 @@ then
         # Heroku's git server keeps `push` calls open during the full deploy process
         # This means that under normal circumstances, this `push` call wouldn't terminate for, like, 20 minutes and the scriptwouldn't terminate
         # In order to avoid that, we apply `nohup` to disconnect stdin, pipe outputs to /dev/null, and end the command with `&` to run the command in the background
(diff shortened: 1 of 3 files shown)


happythenewsad (approved):
Thanks for doing this. I don't think we need `--quiet` or the `&` backgrounding directive, given that we're using `nohup` and piping output to /dev/null. What's your reasoning for using these two directives?

anathomical:
> Thanks for doing this. I don't think we need `--quiet` or the `&` backgrounding directive, given that we're using `nohup` and piping output to /dev/null. What's your reasoning for using these two directives?

Agreed on `--quiet`.  I thought I'd removed it, and will do so.  The `&` needs to be there as far as I know, because of the way Heroku runs their Git server, we need `&` to fully disconnect, at least per my testing.

PR #9103, opened 2022-04-26
Fix broken test on develop

cissyxyu:
## WHAT
Fix broken test on `develop` that was happening because we added an active record query to fetch activities without an explicit ordering, which broke the ordered tests we added.

### Screenshots

### Notion Card Links

PR Checklist | Your Answer
------------ | -------------

### DIFF
--- a/services/QuillLMS/app/workers/populate_all_activity_healths_worker.rb
+++ b/services/QuillLMS/app/workers/populate_all_activity_healths_worker.rb
@@ -15,7 +15,6 @@ class PopulateAllActivityHealthsWorker
     activities = Activity
       .not_archived
       .where(activity_classification_id: relevant_ids)
-      .order(:id)
 
     # spread these, to cut down on DB resource contention.
     activities.each.with_index do |activity, index|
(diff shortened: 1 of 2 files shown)


happythenewsad (changes requested):
Thanks for looking into the issue. While this fixes the failing spec, it introduces complexity (unnecessary ordering) in app code for the sake of a test. We should fix the test code for the sake of the app code. 

One way to do this would be to stub [populate_all_activity_healths_worker.rb](https://github.com/empirical-org/Empirical-Core/pull/9103/files#diff-446b29a43e7ac811a2c8357bd3137876a39f46e2149574522ff333d3d7f5f46c)  [line 15]

in the spec and enforce order that way. 

Another way would be to write a custom rspec matcher:
https://relishapp.com/rspec/rspec-mocks/v/3-2/docs/setting-constraints/matching-arguments#using-a-custom-matcher

There are likely other solutions as well.

cissyxyu:
> Thanks for looking into the issue. While this fixes the failing spec, it introduces complexity (unnecessary ordering) in app code for the sake of a test. We should fix the test code for the sake of the app code.
> 
> One way to do this would be to stub [populate_all_activity_healths_worker.rb](https://github.com/empirical-org/Empirical-Core/pull/9103/files#diff-446b29a43e7ac811a2c8357bd3137876a39f46e2149574522ff333d3d7f5f46c) [line 15]
> 
> in the spec and enforce order that way.
> 
> Another way would be to write a custom rspec matcher: https://relishapp.com/rspec/rspec-mocks/v/3-2/docs/setting-constraints/matching-arguments#using-a-custom-matcher
> 
> There are likely other solutions as well.

Gotcha, that makes a lot of sense! Tried a fix here.

happythenewsad (approved):

PR #9111, opened 2022-04-28
Improvements to CMS District Manager

cissyxyu:
## WHAT
A round of improvements to the Staff District Manager page, now that we've finished the big data upload.
-"Add New District" page added.
-Get rid of the `leanm` field on schools and directly display a school's district name through the rails `district` relationship
-Fix the "Schools" query on district to show all schools regardless of whether they have a subscription or not

## WHY
This is the last round of tweaks we need to make to get the Districts UI operational on Quill staff's end, so they can start going in and editing Districts in the CMS.

## HOW
Adding a few new views and boilerplate controller endpoints, and then modifying the code wherever we're fetching a school's district.

### Screenshots
Screen Shot 2022-04-28 at 5 12 43 PM
Screen Shot 2022-04-28 at 5 13 50 PM
Screen Shot 2022-04-28 at 5 13 59 PM

### Notion Card Links

PR Checklist | Your Answer
------------ | -------------

### DIFF
--- a/services/QuillLMS/app/controllers/cms/districts_controller.rb
+++ b/services/QuillLMS/app/controllers/cms/districts_controller.rb
@@ -28,6 +28,22 @@ class Cms::DistrictsController < Cms::CmsController
     @admins = DistrictAdmin.includes(:user).where(district_id: params[:id].to_i)
   end
 
+  def new
+    @district = District.new
+    @editable_attributes = editable_district_attributes
+  end
+
+  def create
+    new_district = District.new(edit_or_add_district_params)
+    begin
+      new_district.save!
+      redirect_to cms_district_path(new_district.id)
+    rescue => e
+      flash[:error] = e.message
+      redirect_to cms_districts_path
+    end
+  end
+
   def new_admin
     @district = District.find(params[:id])
   end
@@ -38,10 +54,12 @@ class Cms::DistrictsController < Cms::CmsController
   end
 
   def update
-    if District.find(edit_or_add_district_params[:id]).update(edit_or_add_district_params)
+    begin
+      District.find(edit_or_add_district_params[:id]).update!(edit_or_add_district_params)
       redirect_to cms_district_path(edit_or_add_district_params[:id])
-    else
-      render :edit
+    rescue => e
+      flash[:error] = e.message
+      redirect_to cms_district_path(edit_or_add_district_params[:id])
     end
   end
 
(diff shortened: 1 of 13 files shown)


happythenewsad (approved):
+1 on breaking up the project into multiple PRs, so reviewing is easier.

services/QuillLMS/app/controllers/cms/districts_controller.rb
```diff
@@ -28,6 +28,22 @@ def show
     @admins = DistrictAdmin.includes(:user).where(district_id: params[:id].to_i)
   end
 
+  def new
+    @district = District.new
+    @editable_attributes = editable_district_attributes
+  end
+
+  def create
```
brendanshean, line 36: Before getting into `create` and `update`, I'd like to first recommend renaming: 
`def edit_or_add_district_params` to just `def district_params`.  It's a Rails convention to name the params for use in create/update as just the name of the model followed by underscore params.  Note that this is in contrast to the query params definition elsewhere. 

https://github.com/empirical-org/Empirical-Core/blob/a0604d267a9307955f6e0b8fa434c315b7f05ffd/services/QuillLMS/app/controllers/cms/districts_controller.rb#L121

In this particular case, the exception handling isn't being use to notify anything outside of the control (e.g. ErrorNotifier), so I would opt to use a conditional with  boolean `save` over rescue block with `save!` since that adds in unnecessary overhead:

```
district =  District.new(district_params)

if district.save
  redirect_to cms_district_path(district)
else
  redirect_to cms_districts_path, error: district.errors
end
```

Also note that instead you can [assign](https://edgeapi.rubyonrails.org/classes/ActionController/Redirecting.html#method-i-redirect_to) flash messages via `redirect_to`

services/QuillLMS/app/controllers/cms/districts_controller.rb
```diff
@@ -38,10 +54,12 @@ def edit
   end
 
   def update
-    if District.find(edit_or_add_district_params[:id]).update(edit_or_add_district_params)
```
brendanshean, line 41: Similar to `create`:

```
def update
  district = District.find(district_params[:id])
   if district.update(district_params)
        redirect_to cms_district_path(params[:id])
  else
    redirect_to cms_district_path(district_params[:id]), error: district.errors.messages
  end
end
```

services/QuillLMS/app/models/district.rb
```diff
@@ -19,8 +19,14 @@
 #  clever_id      :string
 #  nces_id        :integer
 #
+# Indexes
+#
+#  index_districts_on_nces_id  (nces_id) UNIQUE
+#
 class District < ApplicationRecord
 
+  validate :validate_name_not_blank
```
brendanshean: I think the ActiveRecord validation [presence](https://guides.rubyonrails.org/active_record_validations.html#presence) would suffice: `validates :name, presence: true`?

Then you can get rid of the `validate_name_not_blank`

brendanshean (approved):
Some refactoring comments but nothing that should hold this up once addressed.

Nice work on this!

PR #9112, opened 2022-04-29
Cms/remove memcached

anathomical:
## WHAT
Remove memcached configuration from the CMS
## WHY
Looking at the memcached server usage showed that we basically never use memcached, but we're paying $160/month for it.  After examining the code, I found only one place where we used `Rails.cache` calls, which hooked through to memcached, so it seemed like we could probably find a way to remove memcached safely.
## HOW
I tested hooking `Rails.cache` into Redis instead (which we use for Sidekiq already). and tests proved successful on Staging with no apparent impact on performance.  It seems like we should be safe to make this minimal change in production.
NOTE: Once this gets merged and deployed, we'll be ready to remove the Memcached instance from our CMS production server.

### Screenshots
Screen Shot 2022-04-29 at 4 05 00 PM

### Notion Card Links
https://www.notion.so/quill/Shut-down-unneeded-CMS-servers-in-AWS-4d2ca736c9e24efda50346b4d0afbd60

PR Checklist | Your Answer
------------ | -------------
Have you added and/or updated tests? |  No changes to behavior, so no update to tests


happythenewsad (approved):
👍

brendanshean (approved):

PR #9088, opened 2022-04-20
Cut down on CMS weekend DB load - spread ActivityHealth jobs out.

dandrabik:
## WHAT
Every weekend, we collect health metrics for all grammar and connect activities. This puts a lot of load on the CMS. This PR aims to reduce that load a bit by:
1. Not doing this for archived activities (takes away about 1/3 of the workload).
2. Spreading the jobs out so they don't hit the DB at the same time.
## WHY
The hope is this will help keep the CMS DB performant over the weekend, and also eliminate the alarms and site issues that this causes in off-hours.
## HOW
Scope the finder to `.not_archived`, spread the jobs out by 3 minutes each.
### Screenshots
Screen Shot 2022-04-20 at 11 30 31 AM

Screen Shot 2022-04-20 at 11 33 35 AM

### Notion Card Links
(Please provide links to any relevant Notion card(s) relevant to this PR.)

PR Checklist | Your Answer
------------ | -------------
Have you added and/or updated tests? |  'YES'. 
Have you deployed to Staging? |  NO - tiny change
Self-Review: Have you done an initial self-review of the code below on Github? | Yes
Spec Review: Have you reviewed the spec and ensured this meets requirements and/or matches design mockups? | N/A

PR #9118, opened 2022-05-03
remove Activity archive flag everywhere

happythenewsad:
## WHAT
- Activities editor dropdown should display the current Activity's flag (currently, it displayed the first flag in the flag array).
- Historically, Activities have contained values 'archive' and 'archived' in the flag column. Remove all logic pertaining to the 'archive' value from the codebase, to eliminate tech debt.

## HOW
Files evaluated:

flags.rb x
activities_controller x
activity.rb x
units_controller x
unit_activity.rb x
ActivitySearch x
scorebook/query.rb x
ActivityForm x
UnitTemplatePseudoSerializer x
flagMap.js x
app/views/cms/activities/index.html.slim x
cms/activities_controller_spec.rb x

### Screenshots

### Notion Card Links
https://www.notion.so/quill/b417e5b89fe04af89cba62d4198f7031?v=77f294a5b0df466e98633244de7ebfef&p=ecfc8bd64efe4792a64cee6be90c00bd

PR Checklist | Your Answer
------------ | -------------


happythenewsad:
Hi folks, I could use your opinions here. @dandrabik @emilia-friedberg 

I investigated a low priority bug ticket today. The issue is explained [here](https://www.notion.so/quill/b417e5b89fe04af89cba62d4198f7031?v=77f294a5b0df466e98633244de7ebfef&p=ecfc8bd64efe4792a64cee6be90c00bd).

The code in this PR addresses the issue (with the future addition of converting 'archived' back to 'archive' on `article.save`), but it's a band-aid on the root problem. The root problem is that our Postgres data does not conform to the declared [schema](https://www.notion.so/quill/b417e5b89fe04af89cba62d4198f7031?v=77f294a5b0df466e98633244de7ebfef&p=ecfc8bd64efe4792a64cee6be90c00bd).

In flags.rb, we declare that the enumeration of allowable flag values is:
`FLAGS = %w(production archive alpha beta gamma private)` 

However, we have code that assumes that `archived` is a valid flag. Existing code [handles this with a band-aid already](https://github.com/empirical-org/Empirical-Core/blob/develop/services/QuillLMS/app/controllers/cms/activities_controller.rb#L11), but not everywhere.

So going forward, do we:

1. Address the current issue with a band-aid (this PR). 
2. Or do it the proper way, by:
- backfilling activities so that `activity.flags.archive -> activity.flags.archived` 
- Removing all existing band aids
- auditing the codebase to catch regressions. 

There are 200 instances of 'archive' in the codebase, and 155 instances of 'archived', which is why I'm questioning if option 2 is worth the effort right now. But perhaps you two have some historical knowledge that can guide this decision. Thanks!

dandrabik:
> Hi folks, I could use your opinions here. @dandrabik @emilia-friedberg
> 
> I investigated a low priority bug ticket today. The issue is explained [here](https://www.notion.so/quill/b417e5b89fe04af89cba62d4198f7031?v=77f294a5b0df466e98633244de7ebfef&p=ecfc8bd64efe4792a64cee6be90c00bd).
> 
> The code in this PR addresses the issue (with the future addition of converting 'archived' back to 'archive' on `article.save`), but it's a band-aid on the root problem. The root problem is that our Postgres data does not conform to the declared [schema](https://www.notion.so/quill/b417e5b89fe04af89cba62d4198f7031?v=77f294a5b0df466e98633244de7ebfef&p=ecfc8bd64efe4792a64cee6be90c00bd).
> 
> In flags.rb, we declare that the enumeration of allowable flag values is: `FLAGS = %w(production archive alpha beta gamma private)`
> 
> However, we have code that assumes that `archived` is a valid flag. Existing code [handles this with a band-aid already](https://github.com/empirical-org/Empirical-Core/blob/develop/services/QuillLMS/app/controllers/cms/activities_controller.rb#L11), but not everywhere.
> 
> So going forward, do we:
> 
> 1. Address the current issue with a band-aid (this PR).
> 2. Or do it the proper way, by:
> 
> * backfilling activities so that `activity.flags.archive -> activity.flags.archived`
> * Removing all existing band aids
> * auditing the codebase to catch regressions.
> 
> There are 200 instances of 'archive' in the codebase, and 155 instances of 'archived', which is why I'm questioning if option 2 is worth the effort right now. But perhaps you two have some historical knowledge that can guide this decision. Thanks!

@happythenewsad Even though it is slightly more work in the short-term, I think it'll be a long-term net time-saver and is worth the effort to converge everything to use `archived` (the second option). It's really surprising that there is a both an `archive` and `archived` flag, and allowing that to still exist will lead to more bugs and time spent tracking down those bugs (I literally introduced [code last week](https://github.com/empirical-org/Empirical-Core/pull/9088/files) that is susceptible to this bug). Let's fix the root cause.

services/QuillLMS/client/app/bundles/Staff/containers/ActivityForm.tsx
```diff
@@ -100,23 +99,29 @@ const ActivityForm = ({ activity, activityClassification, contentPartnerOptions,
...
-  const standardOptionElements = standardOptions.map(so => (<option value={so.id}>{so.name}</option>))
+  const standardOptionElements = standardOptions.map(so => (<option key={so.id} value={so.id}>{so.name}</option>))
 
-  const activityCategoryOptionElements = activityCategoryOptions.map(so => (<option value={so.id}>{so.name}</option>))
+  const activityCategoryOptionElements = activityCategoryOptions.map(so => (<option key={so.id} value={so.id}>{so.name}</option>))
 
   let followUpActivityField
-
+  /* eslint-disable jsx-a11y/label-has-associated-control */
+  /* eslint-disable jsx-a11y/control-has-associated-label */
```
happythenewsad: Disabling these 2 rules because:
- this PR did not cause the violations, just uncovered them
- removing the violations would [change the HTML structure, or something more invasive](https://github.com/jsx-eslint/eslint-plugin-jsx-a11y/blob/main/docs/rules/label-has-associated-control.md), without providing much value

emilia-friedberg: I don't think we should disable these rules here. They're only staged file linter rules, so they're not going to make your PR unmergeable, but they are a standard we'd like to hold to. I think it's fine to ignore it for now if you'd like to, since you didn't introduce the issue, but the fix is just adding an id to the `select` element and an `htmlFor={[selectElementId]}` to the `label` element.

emilia-friedberg (approved):
One request but looks good overall, don't need to review again.

anathomical (approved):

PR #9125, opened 2022-05-04
Sales forms submission/ add sales forms

eadams17:
## WHAT
add sales forms UI and backend logic for quote and renewal requests

## WHY
we want to store these submissions in our DB

## HOW
create new SalesFormAppClient in frontend and add backend logic to support creating `sales_form_submission` entries on submission

### Screenshots
Screen Shot 2022-05-03 at 9 47 02 PM

### Notion Card Links
https://www.notion.so/quill/Create-custom-sales-forms-8e23e452b9894d5abeed63af7323f41c#db4f40e8a2bb4e0e8badcd7835bc5583

PR Checklist | Your Answer
------------ | -------------


services/QuillLMS/client/app/bundles/Teacher/helpers/salesForms.tsx
```diff
@@ -0,0 +1,173 @@
...
+  );
+  const districtSearchInput = (
+    <SelectSearch
+      filterOptions={districtSearch}
+      onChange={handleDistrictSearchChange}
+      options={districts}
+      placeholder="Search for your district"
+      search={true}
+    />
+  );
```
emilia-friedberg: For consistent styling, we should probably use our custom `DropdownInput` backpack element rather than the default `SelectSearch` here. You can see an example of the searchable `DropdownInput` here under "Exposed Editable Dropdown Menu": https://www.quill.org/backpack#menus.

services/QuillLMS/app/controllers/sales_form_submission_controller.rb
```diff
@@ -0,0 +1,44 @@
...
+
+  def options_for_sales_form
+    type = params[:type]
+    schools_or_districts = type == SCHOOL ? School.all : District.all
+    options = schools_or_districts.map do |school_or_district|
+      {
+        name: school_or_district.name,
+        value: school_or_district.name
+      }
+    end
```
emilia-friedberg: This is a matter of style mostly, but I'd be inclined to say that the backend should just pass up the data and then the frontend should format it however it needs to pass it to the rendering component (so the mapping should happen there, not here). The backend should be agnostic to how the data gets used.

services/QuillLMS/client/app/bundles/Teacher/components/salesForm/index.tsx
```diff
@@ -0,0 +1,174 @@
...
+    [PHONE_NUMBER]: setPhoneNumber,
+    [ZIPCODE]: setZipcode,
+    [SCHOOL_PREMIUM_ESTIMATE]: setSchoolPremiumEstimate,
+    [TEACHER_PREMIUM_ESTIMATE]: setTeacherPremiumEstimate,
+    [STUDENT_PREMIUM_ESTIMATE]: setStudentPremiumEstimate,
+    [COMMENTS]: setComments,
+    [SCHOOL]: setSelectedSchool,
+    [DISTRICT]: setSelectedDistrict,
+    [SCHOOL_OR_DISTRICT]: setSchoolOrDistrict
+  }
```
emilia-friedberg, line 72: I like this setup!

services/QuillLMS/client/app/bundles/Teacher/components/salesForm/index.tsx
```diff
@@ -0,0 +1,174 @@
...
+          handleUpdateField={handleUpdateField}
+          schoolPremimumEstimate={schoolPremimumEstimate}
+          studentPremimumEstimate={studentPremimumEstimate}
+          teacherPremimumEstimate={teacherPremimumEstimate}
+        />
+        {errors[SUBMISSION_ERROR] && <p className="error-text">{errors[SUBMISSION_ERROR]}</p>}
+        <button className="submit-button quill-button contained primary medium" onClick={handleFormSubmission}>Submit</button>
+      </form>}
+      {formSubmitted && <div className="success-container"><h3>Request successfully submitted!</h3></div>}
+    </div>
```
emilia-friedberg: It probably makes sense to use a snackbar here.

services/QuillLMS/client/app/bundles/Teacher/components/salesForm/index.tsx
```diff
@@ -0,0 +1,174 @@
...
+        <LowerFormFields
+          comments={comments}
+          errors={errors}
+          handleUpdateField={handleUpdateField}
+          schoolPremimumEstimate={schoolPremimumEstimate}
+          studentPremimumEstimate={studentPremimumEstimate}
+          teacherPremimumEstimate={teacherPremimumEstimate}
+        />
+        {errors[SUBMISSION_ERROR] && <p className="error-text">{errors[SUBMISSION_ERROR]}</p>}
+        <button className="submit-button quill-button contained primary medium" onClick={handleFormSubmission}>Submit</button>
```
emilia-friedberg: Let's add a `focus-on-light` class here to help make this accessible.

services/QuillLMS/client/app/bundles/Teacher/components/salesForm/lowerFormFields.tsx
```diff
@@ -0,0 +1,50 @@
+import * as React from 'react';
+
+import { Input } from '../../../Shared';
+import { SCHOOL_PREMIUM_ESTIMATE, TEACHER_PREMIUM_ESTIMATE, STUDENT_PREMIUM_ESTIMATE, COMMENTS } from '../../../../constants/salesForm';
+
+export const LowerFormFields = ({ errors, handleUpdateField, schoolPremimumEstimate, teacherPremimumEstimate, studentPremimumEstimate, comments }) => {
```
emilia-friedberg: Looks like there's a typo in all the estimate fields here (`Premimum` instead of `Premium`)

services/QuillLMS/client/app/bundles/Teacher/components/salesForm/lowerFormFields.tsx
```diff
@@ -0,0 +1,50 @@
...
+        error={errors[STUDENT_PREMIUM_ESTIMATE]}
+        handleChange={handleUpdateField}
+        id={STUDENT_PREMIUM_ESTIMATE}
+        label={STUDENT_PREMIUM_ESTIMATE}
+        placeholder=""
+        value={studentPremimumEstimate}
+      />
+      <div className="comments-container">
+        <label htmlFor={COMMENTS} id="comments-label">Comments (optional)</label>
+        <textarea
```
emilia-friedberg: I think this component: services/QuillLMS/client/app/bundles/Shared/components/shared/textArea.tsx might serve your purpose and already be styled to match the rest of the inputs here.

emilia-friedberg (changes requested):
Some changes requested around using existing Backpack components rather than new ones for consistent styling, but looks really good in general! I really like your hash approach to all of the setter functions.

services/QuillLMS/client/app/bundles/Teacher/helpers/salesForms.tsx
eadams17: the reason I opted to use `SelectSearch` was because of [this part of the spec](https://www.notion.so/quill/Create-custom-sales-forms-8e23e452b9894d5abeed63af7323f41c#69f29dba85074c4184b570e5acc5c429). I can pass the custom search functions I've added in this PR to handle this use case. Is there a straightforward way to do this with the `DropdownInput` ?

services/QuillLMS/client/app/bundles/Teacher/components/salesForm/index.tsx
eadams17: added!

services/QuillLMS/client/app/bundles/Teacher/components/salesForm/lowerFormFields.tsx
eadams17: good catch! I'll update

services/QuillLMS/client/app/bundles/Teacher/components/salesForm/lowerFormFields.tsx
eadams17: ahh totally forgot about this one! I thought we only had the `TextEditor` component. I'll update

services/QuillLMS/app/controllers/sales_form_submission_controller.rb
eadams17: ok, that's a fair point. I can just pass up an array of all the schools and districts

brendanshean: If you're just using names (and don't care about the objects themselves), you could use:
`options = type.classify.constantize.all.pluck(:name)`

eadams17: @brendanshean this is perfect, thank you!

services/QuillLMS/client/app/bundles/Teacher/helpers/salesForms.tsx
emilia-friedberg: Isn't that part of the spec just asking for a final option that says "My school isn't listed"? You should be able to just append that to the list of options and then handle the behavior based on whether or not they check it. Otherwise, I'd look at the docs for the `react-select` library we're using for `DropdownInput` and see if there's a way to implement custom searching (from a brief scan, it looks like there is): https://react-select.com/advanced. Then you can just update the `DropdownInput` component to pass that function if it exists.

emilia-friedberg (approved):
Looks great! Thank you for making those changes!

services/QuillLMS/client/app/bundles/Teacher/components/salesForm/index.tsx
```diff
@@ -26,28 +26,26 @@ export const SalesForm = ({ type }) => {
...
+      if(response && response.options) {
+        setSchools(response.options);
+      }
+    });
+    getSchoolsAndDistricts('district', districtSearchQuery).then((response) => {
+      if(response && response.options) {
+        setDistricts(response.options);
+      }
+    });
+  }, [schoolSearchQuery, districtSearchQuery]);
```
emilia-friedberg: Does it make sense to split these two up and only do `getSchoolsAndDistricts('school',...` on `[schoolSearchQuery]` and same for districts? Because the value of one isn't dependent on the other, right? Otherwise I think we'll be unnecessarily hitting the server an extra time for every keystroke.

emilia-friedberg (approved):
Approving with one question/suggestion.

services/QuillLMS/app/controllers/sales_form_submission_controller.rb
```diff
@@ -27,10 +27,20 @@ def create
...
   def options_for_sales_form
     type = params[:type]
-    schools_or_districts = type.classify.constantize.all.pluck(:name)
+    search_query = params[:search]
+    prefix = search_query_prefix(search_query)
+    schools_or_districts = type.classify.constantize.all.where("lower(name) LIKE :prefix", prefix: "#{prefix.downcase}%").limit(10).pluck(:name)
     render json: { options: schools_or_districts }
   end
 
+  private def search_query_prefix(search)
```
brendanshean: A guard clause can help remove the temporary variable. Also since params[:search] is available to this method, you could just omit the parameter passing altogether.

```
private def search_query_prefix
  search = params[:search]
  
  return '' if search.blank?
   
  search.gsub(/\d*/, '').squish
end

File: services/QuillLMS/app/controllers/sales_form_submission_controller.rb [hunk 10]
```diff
@@ -27,10 +27,20 @@ def create
 
   def options_for_sales_form
     type = params[:type]
-    schools_or_districts = type.classify.constantize.all.pluck(:name)
+    search_query = params[:search]
+    prefix = search_query_prefix(search_query)
+    schools_or_districts = type.classify.constantize.all.where("lower(name) LIKE :prefix", prefix: "#{prefix.downcase}%").limit(10).pluck(:name)
     render json: { options: schools_or_districts }
   end
```
brendanshean, line 33: For the regex in this method, `/\d{5}/`, in some cases might not give us quite what we want:

```
pry(main)> search = "123 The 23456 GoodZ 7890124"                                            
=> "123 The 23456 GoodZ 7890124"
pry(main)> search.gsub(/\d{5}/,'').strip                                                           
=> "123 The  GoodZ 24"
```

`/d*/` might be more appropriate here: 
```
pry(main)> search.gsub(/\d*/,'').strip                                                             
=> "The  GoodZ"
```

This leaves double spaces in the middle but you can use [squish](https://duetngyen.wordpress.com/2016/07/20/hello/)
```
pry(main)> search.gsub(/\d*/,'').squish                                                            
=> "The GoodZ"
```

services/QuillLMS/app/controllers/sales_form_submission_controller.rb
```diff
@@ -0,0 +1,48 @@
...
+
+  def create
+    sales_form_submission = SalesFormSubmission.new(sales_form_submission_params)
+    if  sales_form_submission.save!
+      head :no_content, status: 200
+    else
+      render json: sales_form_submission.errors, status: :unprocessable_entity
+    end
+  end
+
```
brendanshean, line 27: Some refactoring suggestions:

1) Assign a temporary variable for `klass`
2) Remove the parameter passing in `search_query_prefix` so that it can be used directly in the where query
3) Use case insensitive search `where("name ILIKE ?", #{search_query_prefix}%"`

```
def options_for_sales_form
  klass = params[:type].classify.constantize
  schools_or_districts = klass.all.where("name ILIKE ?", "#{search_query_prefix}%").limit(10).pluck(:name)
  render json: { options: schools_or_districts }
end
```

brendanshean (approved):
Some backend refactoring comments but otherwise looks good.  I think the limit(10) on the auto complete is a good idea.

services/QuillLMS/client/app/bundles/Teacher/components/salesForm/index.tsx
eadams17: yeah that's a good point, I'll separate them out

services/QuillLMS/app/controllers/sales_form_submission_controller.rb
eadams17: great! this is much better, thanks for the examples

services/QuillLMS/app/controllers/sales_form_submission_controller.rb
eadams17: this is much cleaner, thanks for the suggestions!

PR #9127, opened 2022-05-04
Update customer.subscription.updated webhook handling

brendanshean:
## WHAT
Fix webhook event handling that is reporting failure

## WHY
The `customer.subscription.updated` handler was originally written to handle events where users toggle the recurring status of the subscription.  However, this same webhook is triggered when users purchase a subscription for the first time.  When the handler attempts to find the subscription it raises a `SubscriptionNotFound` exception since our db does not yet have the subscription (a separate webhook event/handler handles the creation of the subscription in our db). 

## HOW
Add a check to the handler to see if the subscription was previously incomplete.  If so, don't bother trying to update the subscription in our database since it doesn't exist yet.

In this case we can inspect the previous attributes value and check for `status: complete`

### Screenshots

### Notion Card Links

PR Checklist | Your Answer
------------ | -------------
Have you added and/or updated tests? |  YES'


emilia-friedberg (approved):

dandrabik (approved):
Thanks for fixing this up, LGTM.

PR #9131, opened 2022-05-05
evidence accessibility updates

emilia-friedberg:
## WHAT
Deep dive into screenreader navigability in Evidence, with the major changes being:

- Adding the new "useFocus" hook (will come in handy elsewhere!) and calling it strategically (largely on major component load) as the user moves through the activity so they don't end up in morasses at the bottom of the page
- Adding screenreader-only text where directions might be confusing to users who aren't interacting with the page visually
- Using `aria-hidden` to hide parts of the page that are non-interactable at a given time (for example, the read passage section when it is blurred out), and also just have redundant information that will be confusing if repeated

## WHY
During early testing, our accessibility consultant and I realized that Evidence really needed a deeper accessibility dive since the pages are just so complicated. This is my pass at using a screenreader and making improvements where I identified points of potential confusion or difficulty. In my next meeting with her, we'll go through Evidence again and hopefully be able to make much smaller adjustments to make it fully functional.

## HOW
Extensive use of manual focus control, aria-hidden, and the `sr-only` class.

### Screenshots

### Notion Card Links
https://www.notion.so/quill/Take-another-pass-at-using-a-screenreader-to-navigate-Evidence-58c3bb8cc8b94589ad521fe5f9e6058d

PR Checklist | Your Answer
------------ | -------------


services/QuillLMS/client/app/bundles/Evidence/components/studentView/container.tsx
```diff
@@ -475,11 +476,19 @@ export const StudentViewContainer = ({ dispatch, session, isTurk, location, acti
       const highlightIncludesElement = strippedPassageHighlights.find(ph => ph.includes(stringifiedInnerElements)) // handles case where passage highlight spans more than one paragraph
       const elementIncludesHighlight = strippedPassageHighlights.find(ph => stringifiedInnerElements.includes(ph)) // handles case where passage highlight is only part of paragraph
       if (highlightIncludesElement) {
-        return <p><span className="passage-highlight">{stringifiedInnerElements}</span></p>
+        return (
+          <p>
+            <span className="passage-highlight">
+              <span className="sr-only">(yellow underlined text begins here)</span>
+              {stringifiedInnerElements}
+              <span className="sr-only">(yellow underlined text ends here)</span>
```
anathomical, line 484: This seems like a powerful pattern.  I wonder if we'll eventually want a shared component that takes the "start" and "stop" texts and wraps things automatically.  May not be worth figuring out now, but I wanted to call out that this is really neat.

services/QuillLMS/client/app/bundles/Evidence/components/studentView/container.tsx
```diff
@@ -475,11 +476,19 @@ export const StudentViewContainer = ({ dispatch, session, isTurk, location, acti
...
+            </span>
+          </p>
+        )
       }
       if (elementIncludesHighlight) {
         let newStringifiedInnerElements = stringifiedInnerElements
-        strippedPassageHighlights.forEach(ph => newStringifiedInnerElements = newStringifiedInnerElements.replace(ph, `<span class="passage-highlight">${ph}</span>`))
+        strippedPassageHighlights.forEach(ph => newStringifiedInnerElements = newStringifiedInnerElements.replace(ph, `<span class="passage-highlight"><span class="sr-only">(yellow underlined text begins here)</span>${ph}<span class="sr-only">(yellow underlined text ends here)</span></span>`))
```
anathomical, line 491: Don't know that it'd be simpler, but here's a place where a shared component could be used, but again, that might be overkill at this stage.

services/QuillLMS/client/app/bundles/Evidence/components/studentView/headerImage.tsx
```diff
@@ -57,7 +57,8 @@ export const HeaderImage = ({ headerImage, passage }) => {
         {headerImage}
         <section className="header-image-information">
           {passage.image_caption && <p className="header-image-caption">{stripHtml(passage.image_caption)}</p>}
-          {passage.image_attribution && <Tooltip handleClick={handleImageAttributionClick} isTabbable={true} tooltipText={tooltipText} tooltipTriggerText="Image credit" tooltipTriggerTextClass="image-attribution-tooltip" />}
+          <span className="sr-only">Image credit: {tooltipText}</span>
```
anathomical, line 60: A concern well beyond what we're doing in this PR, but should we be considering some sort of `sr-only` element that's _always_ part of `<Tooltip>` components?

services/QuillLMS/client/app/bundles/Evidence/components/studentView/readAndHighlightInstructions.tsx
```diff
@@ -4,14 +4,25 @@ import DisplayStudentHighlight from './displayStudentHighlight'
...
 const ReadAndHighlightInstructions = ({ passage, activeStep, studentHighlights, removeHighlight, inReflection, }) => {
+  const [containerRef, setContainerFocus] = useFocus()
+
+  React.useEffect(() => {
+    setContainerFocus()
+  }, [])
+
+  React.useEffect(() => {
+    setContainerFocus()
+  }, [inReflection])
```
anathomical, line 18: Do we want both of these declarations?  Seems like the first one would mean that our attempt to not run it all the time in the second one would fail, or am I missing something?

anathomical (approved):
Looks good to me!  One comment about possible duplicate code (that may just be me missing something), and a few notes about `sr-only` text which I really like.

services/QuillLMS/client/app/bundles/Evidence/components/studentView/headerImage.tsx
emilia-friedberg: Absolutely. I have taking a deeper look at tooltips from an accessibility perspective as an upcoming task!

services/QuillLMS/client/app/bundles/Evidence/components/studentView/readAndHighlightInstructions.tsx
emilia-friedberg: Good question. This is just a property of the `useEffect` hook. An empty array as the second argument is effectively the equivalent of the `componentDidMount` lifecycle method, and will only run on first load. An array with the name of a prop or state value as the second argument will only run when that value changes. So this setup means "run on load, and also when `inReflection` becomes true". There is a brief explainer of that here: https://reactjs.org/docs/hooks-effect.html#tip-optimizing-performance-by-skipping-effects.

services/QuillLMS/client/app/bundles/Evidence/components/studentView/container.tsx
emilia-friedberg: Yeah, I thought about just making it a shared component here, but it's a little weird because in the first instance we're rendering it as a JSX element and here it's a string (hence the `className` vs `class` difference) that then gets converted into HTML, so there actually wasn't a super easy way to do that.

services/QuillLMS/client/app/bundles/Evidence/components/studentView/container.tsx
emilia-friedberg: Totally agree that this is a pattern I think we'd like to follow in the future. I think the idea of a shared component is probably a good one - will contemplate that as I go through the rest of the accessibility work. I think I'll be doing something similar with underlined input fields in Proofreader, so I will try to think then about how we might share that logic most effectively.

services/QuillLMS/client/app/bundles/Evidence/components/studentView/readAndHighlightInstructions.tsx
anathomical: Aha!  I didn't know that's what the empty array did.  Interesting, and sort of weird way of handling it, but that makes sense to me!

services/QuillLMS/client/app/bundles/Evidence/components/studentView/container.tsx
anathomical: Yeah, that makes sense.  Giving ourselves time to see how we want to use something like this before trying to make shared component makes sense.

services/QuillLMS/client/app/bundles/Evidence/components/studentView/promptStepContainer.tsx
```diff
@@ -0,0 +1,57 @@
...
+  activities,
+  session,
+  activeStep,
+  doneHighlighting,
+  reportAProblem,
+}) => {
+  const { currentActivity, } = activities
+  const { submittedResponses, hasReceivedData, } = session
+
+  if (!currentActivity || !hasReceivedData) return
```
eadams17: do we also want to be returning an empty `<span/>` here?

eadams17 (approved):
LGTM!

PR #9136, opened 2022-05-05
End of year webinar banner

cissyxyu:
## WHAT
End of year webinar banner for teachers only.

## WHY
We want to publicize end of school year webinars.

## HOW
Add the new banner in

### Screenshots

### Notion Card Links
https://www.notion.so/quill/End-of-year-webinar-banners-363c3127cf4448cbb17889d9e1a8d65a

PR Checklist | Your Answer
------------ | -------------


emilia-friedberg (approved):
LGTM. My only question is about when we want to clean up one-off webinars that are now over - it isn't pressing but we presumably want to do sometime before next year rolls around, right? Does it make sense to just remove past ones when we are adding new ones?

cissyxyu:
> LGTM. My only question is about when we want to clean up one-off webinars that are now over - it isn't pressing but we presumably want to do sometime before next year rolls around, right? Does it make sense to just remove past ones when we are adding new ones?

Good call out -- I'll remove the banners we don't need in this PR, and in the future we can try to clean up old banners every time we add new ones.

PR #9159, opened 2022-05-11
Add bonus days to new school subscriptions on stripe

brendanshean:
## WHAT
Add parameter bonus days to new School Premium Subscriptions provisioned through Stripe.

## WHY
We'd like to get Schools on the same renewal cycle either June 30 or Dec 31 while also providing a bonus to new School subscribers.

It would also make the school manual plans' bonus days consistent with those purchased on stripe.

## HOW
Stripe [provides](https://stripe.com/docs/billing/subscriptions/trials) a `trial_period_days` parameter that can be [merged](https://github.com/empirical-org/Empirical-Core/pull/9159/commits/3ac63b26da3b994721b80a3dc784671480fe07db#diff-aaa793cd1f752bdf6b145438ed06ad642f56786d928238e2977c3c9f564c9592R21) with the existing parameters sent to the `Stripe:Checkout:Session.create`.

### Screenshots
Screen Shot 2022-05-11 at 9 45 56 AM

### Notion Card Links
](https://www.notion.so/quill/Implement-a-new-direct-subscription-process-that-utilizes-Stripe-subscriptions-229f2cba7d5e4dbf8bf95388f1ab4209)

PR Checklist | Your Answer
------------ | -------------

### DIFF
--- a/services/QuillLMS/app/models/bonus_days_calculator.rb
+++ b/services/QuillLMS/app/models/bonus_days_calculator.rb
@@ -1,0 +1,34 @@
+# frozen_string_literal: true
+
+class BonusDaysCalculator < ApplicationService
+  JUNE = 6
+  JULY = 7
+  DECEMBER = 12
+
+  attr_reader :user
+
+  delegate :school, to: :user
+
+  def initialize(user)
+    @user = user
+  end
+
+  def run
+    return 0 if user.nil? || school.nil? || school_has_paid_before?
+
+    (plan_ends - today).to_i
+  end
+
+  private def plan_ends
+    today.month < JULY ? today.change(month: JUNE, day: 30) : today.change(month: DECEMBER, day: 31)
+  end
+
+  private def school_has_paid_before?
+    ::Subscription.school_or_user_has_ever_paid?(school)
+  end
+
+  private def today
+    @today ||= Date.current
+  end
+end
+
--- a/services/QuillLMS/app/models/bonus_days_calculator.rb
+++ b/services/QuillLMS/app/models/bonus_days_calculator.rb
@@ -5,30 +5,27 @@ class BonusDaysCalculator < ApplicationService
   JULY = 7
   DECEMBER = 12
 
-  attr_reader :user
+  attr_reader :user, :start
 
   delegate :school, to: :user
 
-  def initialize(user)
+  def initialize(user, start: Date.current)
     @user = user
+    @start = start
   end
 
   def run
     return 0 if user.nil? || school.nil? || school_has_paid_before?
 
-    (plan_ends - today).to_i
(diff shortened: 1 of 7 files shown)


services/QuillLMS/app/models/bonus_days_calculator.rb
```diff
@@ -0,0 +1,34 @@
...
+  end
+
+  def run
+    return 0 if user.nil? || school.nil? || school_has_paid_before?
+
+    (plan_ends - today).to_i
+  end
+
+  private def plan_ends
+    today.month < JULY ? today.change(month: JUNE, day: 30) : today.change(month: DECEMBER, day: 31)
```
dandrabik: Oh, interesting, I didn't know about the `.change` method. Very handy!

services/QuillLMS/app/models/bonus_days_calculator.rb
```diff
@@ -0,0 +1,34 @@
...
+class BonusDaysCalculator < ApplicationService
+  JUNE = 6
+  JULY = 7
+  DECEMBER = 12
+
+  attr_reader :user
+
+  delegate :school, to: :user
+
+  def initialize(user)
```
dandrabik: I think you have a really good testing setup, so you can ignore this, but just something to think about in the future: 

Having a `current date/time` within a function can be tricky to test (hence callbacks and time travel), BUT if you can refactor the function to pass the `current date/time` as a parameter, then it becomes a pure function, in that it no longer has the "changing time side effect". 

Pure functions are very easy to test (there are other benefits that may not apply here, like being able to run it for the past or future). This is what testing would look like with this approach:

```ruby
attr_reader :user, :start

def initialize(user, start: Date.current)
  @user = user
  @start = start
end

# tests (no time travel required)
expect BonusDaysCalculator.run(user, start: Date.new(2003,6,1)).to eq(29)

dandrabik (approved):
Looks good to me. One take or leave comment about making the `BonusDaysCalculator` more functional.

services/QuillLMS/app/models/bonus_days_calculator.rb
brendanshean: I would agree.  I will just refactor now.

brendanshean: I would agree that making this a pure function would be an improvement.  I will refactor now.

services/QuillLMS/app/models/bonus_days_calculator.rb
emilia-friedberg: I didn't know about it either and something about the naming strikes me as funny.

emilia-friedberg (approved):

PR #9166, opened 2022-05-13
Cancel subscriptions on Quill that are canceled via Stripe UI

brendanshean:
## WHAT
When staff cancels a user's subscription via the Stripe Dashboard, propagate that change to the Quill server.

Also, remove a couple of shared_contexts that are no longer used.

## WHY
We'd like subscription status to be in sync with Stripe.

## HOW
When the `customer.subscription.deleted` webhook is received, set that user's subscription to expire today and turn off recurring.

### Screenshots

### Notion Card Links
https://www.notion.so/quill/Stripe-Subscription-Cancellation-via-Stripe-UI-6c76a91b7f6f4494a543ad936152b1dc

PR Checklist | Your Answer
------------ | -------------
Have you deployed to Staging? |Not yet - deploying now!
Self-Review: Have you done an initial self-review of the code below on Github? | No.


emilia-friedberg (approved):

services/QuillLMS/app/services/stripe_integration/webhooks/customer_subscription_deleted_event_handler.rb
```diff
@@ -0,0 +1,20 @@
+# frozen_string_literal: true
+
+module StripeIntegration
+  module Webhooks
+    class CustomerSubscriptionDeletedEventHandler < EventHandler
+      EVENT_TYPE = 'customer.subscription.deleted'
+
+      def self.handles?(event_type)
```
dandrabik: Take or leave, but you could eliminate this `handles?` boiler plate by making the `EVENT_HANDLERS` a map instead of an array, e.g.
```ruby
EVENT_HANDLERS = {
  'customer.subscription.deleted' => CustomerSubscriptionDeletedEventHandler,
  'invoice.paid' => InvoicePaidEventHandler,
 ...
}
EVENT_HANDLERS.fetch(event_type, UnknownEventHandler)...
```
If you are worried about the large ignored events list you could keep that as an array (note, there might be a better way to do this conversion), e.g.
```ruby
IGNORED_EVENTS.reduce({}) {|hash, item| hash[item] = IgnoredEventHandler; hash}

**Review by dandrabik — APPROVED** [review]
Approved. I like this approach. One small take or leave comment about removing some boiler plate.

File: services/QuillLMS/app/services/stripe_integration/webhooks/customer_subscription_deleted_event_handler.rb (hunk 1 above)
Comment (brendanshean) [reply services/QuillLMS/app/services/stripe_integration/webhooks/customer_subscription_deleted_event_handler.rb]: Great idea!  That boilerplate was really bothering me but I couldn't figure out how to get rid of it without throwing out the inheritance.

Comment (brendanshean) [reply services/QuillLMS/app/services/stripe_integration/webhooks/customer_subscription_deleted_event_handler.rb]: @dandrabik Re-requesting since I added a little [more](https://github.com/empirical-org/Empirical-Core/pull/9166/commits/d13984110b1638393b3c4d5a65e05986a6c8861d#diff-46345af568b3961173ca2654e6bbc3a2e5100a4b6e4b686c7bac2eed2bcc24f1R9) than your suggestion.

File: services/QuillLMS/app/services/stripe_integration/webhooks/event_handler.rb [hunk 2]
```diff
@@ -5,6 +5,10 @@ module Webhooks
     class EventHandler < ApplicationService
       attr_reader :stripe_webhook_event
 
+      def self.event_type
```
dandrabik: Interesting setup! This is clever. 

The only concern I would have is that the contract between the Stripe Event name (which we don't control) and the handler class is abstracted away, so if 2 years from now, someone renames `InvoicePaidEventHandler` to `InvoicePaymentEventHandler` (or something), the code would stop receiving the event, but, I believe, the tests would pass (and the future developer would likely be surprised by this).

So I'd suggest, either:
1. Document this very clearly with warnings/explanations about naming requirements, and consider adding tests using the explicit external string.
2. Make the external string to event handler relationship explicit (possibly with documentation as well).

brendanshean: That's a good point.  I will fix this.

dandrabik (approved):
LGTM to me! Thanks for cleaning this up.

cissyxyu (approved):

PR #9167, opened 2022-05-13
Set `cancel_at_period_end` true for teacher premium subscriptions that have received a school subscriptions

brendanshean:
## WHAT
When a user with an active Teacher Premium subscription gets a School Premium or District Premium Subscription, set the `cancel_at_period_end` attribute to `true` on the teacher subscription.

## WHY
We want to turn off renewal for the teacher subscription since they are now under a different plan.

## HOW
Call `stripe_cancel_at_period` end whenever setting `recurring: false` on a subscription.  If the subscription is a stripe one, then it will in turn call `Stripe::Subscription.update(stripe_subscription_id, cancel_at_period_end: true)` . 

### Screenshots

### Notion Card Links
https://www.notion.so/quill/Stripe-Subscription-Cancellation-by-user-5742cd1a90184ed48de1efc4482a1b1a

PR Checklist | Your Answer
------------ | -------------
Have you added and/or updated tests? |   YES

### DIFF
--- a/services/QuillLMS/app/models/stripe_integration/subscription.rb
+++ b/services/QuillLMS/app/models/stripe_integration/subscription.rb
@@ -3,7 +3,7 @@
 module StripeIntegration
   class Subscription < SimpleDelegator
     def stripe_cancel_at_period_end
-      return nil if stripe_invoice_id.nil?
+      return if stripe_invoice_id.nil?
 
       Stripe::Subscription.update(stripe_subscription_id, cancel_at_period_end: true)
     end
--- a/services/QuillLMS/app/models/stripe_integration/subscription.rb
+++ b/services/QuillLMS/app/models/stripe_integration/subscription.rb
@@ -2,6 +2,12 @@
 
 module StripeIntegration
   class Subscription < SimpleDelegator
+    def stripe_cancel_at_period_end
+      return nil if stripe_invoice_id.nil?
+
+      Stripe::Subscription.update(stripe_subscription_id, cancel_at_period_end: true)
+    end
+
     def last_four
       return nil if stripe_invoice_id.nil?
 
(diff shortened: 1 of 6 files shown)


services/QuillLMS/app/models/stripe_integration/subscription.rb
```diff
@@ -2,6 +2,12 @@
 
 module StripeIntegration
   class Subscription < SimpleDelegator
+    def stripe_cancel_at_period_end
+      return nil if stripe_invoice_id.nil?
```
dandrabik: I don't think this needs to return a value, e.g.
```suggestion
      return if stripe_invoice_id.nil?
```
(Side note, this is a minor change that would be caught by Sorbet/typing).

dandrabik (approved):
LGTM! One minor edit suggested.

services/QuillLMS/app/models/subscription.rb
```diff
@@ -165,13 +169,18 @@ def credit_user_and_de_activate
...
       # subtract later of start date or today's date from expiration date to calculate amount to credit
       # amount_to_credit = self.expiration - [self.start_date, Date.current].max
       amount_to_credit = expiration - start_date
       CreditTransaction.create(user_id: user_subscriptions.first.user_id, amount: amount_to_credit.to_i, source: self)
     end
   end
 
+  def disable_stripe_auto_renew
+    Stripe::Subscription.update(stripe_subscription_id, cancel_at_period_end: true)
+  end
```
anathomical: Given that you're doing the actually `update` through a delegation, is this stand-alone method still required on the class?  I don't see it being called anywhere.

anathomical (approved):
Looks like there might be some left-over dead code, but other than that looks good to me.

services/QuillLMS/app/models/subscription.rb
brendanshean: Good catch.  I completely missed this.

emilia-friedberg (approved):

PR #9176, opened 2022-05-17
Add Demo Account option to Welcome Modal for 30% of new teachers

cissyxyu:
## WHAT
Add a new Welcome Modal that encourages new teachers to explore the demo account. I'll add an AppSetting that triggers this modal so that only 30% of teachers see it for now.

## WHY
To test the outcome of letting teachers explore the demo account when they first sign up.

## HOW
Add this welcome modal component and styling, and have it conditionally appear only when the AppSetting is applied.

### Screenshots
Screen Shot 2022-05-17 at 2 15 19 PM

### Notion Card Links
https://www.notion.so/quill/Product-Board-30f97e4eb01246dbb8264253fb385073?p=d4b49293112b4b378666e1ed470c167f

PR Checklist | Your Answer
------------ | -------------

### DIFF
--- a/services/QuillLMS/client/app/bundles/Teacher/components/dashboard/welcome_modal.tsx
+++ b/services/QuillLMS/client/app/bundles/Teacher/components/dashboard/welcome_modal.tsx
@@ -1,12 +1,59 @@
 import * as React from 'react'
 import Confetti from 'react-confetti'
 
+import { handleHasAppSetting } from "../../../Shared/utils/appSettingAPIs";
+
 const classroomActivitiesSrc = `${process.env.CDN_URL}/images/pages/dashboard/illustrations-classroom-activities-2.svg`
 
 const exploreDemoLink = `${process.env.DEFAULT_URL}/teachers/view_demo`
 
+const WelcomeModalWithoutDemo = ({close}) => (
+  <div className="welcome-modal-without-demo quill-modal">
+    <div className="modal-body">
+      <h1>Welcome to Quill!</h1>
+      <div className="welcome-modal-text">
+        <p>Our mission as a non-profit is to help students become strong writers, so all our content is completely free to use with an unlimited number of students.</p>
+      </div>
+      <button className="quill-button contained primary medium" onClick={close} type="button">Let&apos;s go!</button>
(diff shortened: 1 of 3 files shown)


services/QuillLMS/client/app/bundles/Teacher/components/dashboard/welcome_modal.tsx
```diff
@@ -1,10 +1,59 @@
...
+      </div>
+      <div>
+        <div className="welcome-modal-text">
+          <p className="welcome-modal-header">Want to see Quill&apos;s full potential?</p>
+          <p>Play around with a demo account to see sample student data and reports.</p>
+        </div>
+        <br />
+        <div className="welcome-modal-buttons">
+          <a className="quill-button contained secondary medium outlined welcome-modal-demo" href={exploreDemoLink}>Explore demo</a>
+          <button className="quill-button contained primary medium welcome-modal-skip" onClick={close} type="button">Maybe later, skip</button>
```
emilia-friedberg: I think that since we want to encourage people to click this, maybe we should swap these buttons so that the "Explore demo" button is the "primary contained" class and comes after the "Maybe later, skip" button? Not sure if this is something that got mocked up previously or not. Unrelated: it would be nice to put `focus-on-light` classes on both of these buttons for accessibility purposes.

emilia-friedberg (approved):
Approving with one suggestion about a design tweak (feel free to ignore).

services/QuillLMS/client/app/bundles/Teacher/components/dashboard/welcome_modal.tsx
cissyxyu: This was a discussion I had with Tom C before he left (because he made this mockup). I think his reasoning was that we still want to emphasize the option that lets teachers get started as quickly as possible (which is "Skip"). We only want to encourage the users who are actually interested in exploring the demo to click it, instead of having everyone automatically explore the demo without realizing they're going into a demo experience, if that makes sense?

I will put the `focus-on-light` class on both buttons!

eadams17 (approved):

PR #9187, opened 2022-05-19
Lms/normalize concept results

anathomical:
## WHAT
Add a suite of models to support the normalization of `ConceptResult` records along with code to actually handle that migration
## WHY
We want to slow the rapid increase in database storage necessitated by the current denormalized and highly-repeated nature of current `ConceptResult` records.  This new approach should vastly reduce the amount of required space for the same data.
## HOW
- Create a series of migrations to create all the necessary tables we'll be using for denormalization
- Define models with basic tests and validation for all of the models
- Ensure that the root model (`StudentResponse`) can create new normalized records from either existing `ConceptResult` records, or from the JSON payload that's currently used by `ActivitySession`s to create new `ConceptResult`s.
- Ensure that the root model can present itself in the same shape as the old `ConceptResult` records so that we can ultimately use the new model as a drop-in replacement
- Create a series of Sidekiq jobs to copy existing `ConceptResult` records into the new `StudentResponse` model

### Notion Card Links
https://www.notion.so/quill/ConceptResult-RFC-Implementation-33dd662d331a4a0896b1a5bf00f3298d
https://www.notion.so/quill/ConceptResults-Pure-PostgreSQL-Normalization-RFC-b0fd4252f72c4ef68e10d91e5a52afca
https://www.notion.so/quill/Determine-ConceptResult-metadata-schema-3d1b92ff70b8435db5752c00acf78eb6

PR Checklist | Your Answer
------------ | -------------


services/QuillLMS/app/models/concept_result.rb
```diff
@@ -29,6 +29,8 @@ class ConceptResult < ApplicationRecord
   validates :concept, presence: true
   validates :activity_session_id, presence: true
 
+  has_one :student_response_concept_result, dependent: :destroy
```
happythenewsad: This is to avoid adding a foreign key column to a 1.4B row table, right? 👍

services/QuillLMS/app/models/student_response.rb
```diff
@@ -0,0 +1,255 @@
...
+
+
+  validates_exclusion_of :correct, in: [nil]
+  validates_presence_of :attempt_number, :question_number
+
+  # This is a list of keys from the old ConceptResults records
+  # that this model knows how to normalize.  Any keys that are
+  # fed into create_from_json that aren't from this list,
+  # will be stashed in a related StudentResponseExtraMetadata
+  # record
```
happythenewsad: If we're relegating the extra metadata to a separate table, why not throw it away entirely?

services/QuillLMS/spec/models/student_response_spec.rb
```diff
@@ -0,0 +1,345 @@
...
+          .and change(StudentResponsePreviousFeedbackText, :count).by(1)
+          .and change(StudentResponsePromptText, :count).by(1)
+          .and change(StudentResponseQuestionType, :count).by(1)
+          .and not_change(StudentResponseInstructionsText, :count)
+        # No change expected above when "instructions" aren't in the payload
+      end
+
+      it 'should not link to records when the appropriate keys are not provided' do
+        response = StudentResponse.create_from_json(json)
+        metadata = json[:metadata]
```
happythenewsad: remove unnecessary line

services/QuillLMS/spec/workers/enqueue_concept_results_migration_worker_spec.rb
```diff
@@ -0,0 +1,54 @@
...
+
+    it 'should enqueue a CopyConceptResultsToStudentResponsesWorker' do
+      concept_result_ids = ConceptResult.select(:id).all.map(&:id)
+
+      expect(CopyConceptResultsToStudentResponsesWorker).to receive(:perform_async).with(concept_result_ids)
+
+      subject.perform(nil, nil)
+    end
+
+    it 'should take start params to skip existing ConceptResults' do
```
happythenewsad: Nit: These two specs would provide more test coverage if they actually filtered out ConceptResults via the values of `start` and `finish`.

happythenewsad (changes requested):
Solid!

services/QuillLMS/app/models/concept_result.rb
anathomical: I hadn't fully thought that through, actually, but that is a very good reason not to put a ForeignKey constraint in there.

services/QuillLMS/app/models/student_response.rb
anathomical: This is a fair question.  I think a good chunk of this is a sense of paranoia, but on the more practical side of things, one of the things we're planning to do is test this extensively in staging.  Doing so should allow us to use the `extra_metadadata` table to ACTUALLY inventory the keys that we don't currently support.  The hope is that there will be a lot fewer rows in that table than in the full `concept_results` table because most of the `ConceptResults` records should transfer into `StudentResponse` records that have no `extra_metadata`.  We may, once we take a look at what sorts of data ends up in `extra_metadata`, drop the table.  We just want to make sure we look at it first.

services/QuillLMS/spec/workers/enqueue_concept_results_migration_worker_spec.rb
anathomical: So, secretly they both are because the `context` block includes a `let!` that creates a `concept_result`.  That means that there are two in the DB, and the test conditions check to ensure only one is getting processed, but that's definitely not obvious when I look at them again.  I think it's reasonable to tweak the tests to make it more obvious that they are testing the conditions that they are to avoid confusion.

services/QuillLMS/spec/models/student_response_spec.rb
anathomical: Good catch.  Done.

services/QuillLMS/app/models/student_response.rb
```diff
@@ -0,0 +1,255 @@
...
+
+  private_class_method def self.extract_extra_metadata(metadata)
+    extra_metadata = metadata.except(*KNOWN_METADATA_KEYS).reject{ |_,v| v.nil? || (v.is_a?(Enumerable) && v.empty?) }
+
+    return if extra_metadata.empty?
+
+    StudentResponseExtraMetadata.new(metadata: extra_metadata)
+  end
+
+  private_class_method def self.set_normalized_text(target, data_hash)
```
dandrabik: shouldn't this be an instance method?

services/QuillLMS/app/models/student_response.rb
```diff
@@ -0,0 +1,255 @@
...
+#  fk_rails_...  (student_response_directions_text_id => student_response_directions_texts.id)
+#  fk_rails_...  (student_response_instructions_text_id => student_response_instructions_texts.id)
+#  fk_rails_...  (student_response_previous_feedback_text_id => student_response_previous_feedback_texts.id)
+#  fk_rails_...  (student_response_prompt_text_id => student_response_prompt_texts.id)
+#  fk_rails_...  (student_response_question_type_id => student_response_question_types.id)
+#
+class StudentResponse < ApplicationRecord
+  belongs_to :activity_session
+  belongs_to :question
+  belongs_to :student_response_answer_text
```
dandrabik: Do we need `_text` in all of these names? Could this be `student_response_answer`? Also, I'm not sure `student` is needed in these names as well. Maybe you are trying to avoid a similar name to the CMS, but it might be better to  call all of these `response`, `response_answer`, etc. It would make the code a lot more brief and increase readability. We should discuss.

services/QuillLMS/app/models/student_response.rb
```diff
@@ -0,0 +1,255 @@
...
+      concept_ids: data_hash[:concept_ids].uniq,
+      attempt_number: metadata[:attemptNumber],
+      correct: metadata[:correct],
+      question: question,
+      question_number: metadata[:questionNumber],
+      question_score: metadata[:questionScore],
+      student_response_extra_metadata: extra_metadata
+    )
+
+    set_normalized_text(student_response, data_hash)
```
dandrabik: Seems like `set_normalized_text` should be an instance method.

services/QuillLMS/app/models/concerns/is_student_response_normalized_text.rb
```diff
@@ -0,0 +1,15 @@
...
+# This concern is intended to describe the many different tables that we use
+# to normalize different types of text.  Basically these tables all do the
+# same thing, but we wanted to keep the various concerns separated.
+module IsStudentResponseNormalizedText
+  extend ActiveSupport::Concern
+
+  included do
+    has_many :student_responses
+
+    validates :text, uniqueness: true, length: { minimum: 0, allow_nil: false }
```
dandrabik: Why does this validate a minimum length of zero?

services/QuillLMS/app/models/concerns/is_student_response_normalized_text.rb
```diff
@@ -0,0 +1,15 @@
+# frozen_string_literal: true
+
+# This concern is intended to describe the many different tables that we use
+# to normalize different types of text.  Basically these tables all do the
+# same thing, but we wanted to keep the various concerns separated.
+module IsStudentResponseNormalizedText
```
dandrabik: I'm on the fence about this module. I'm not sure if this abstraction is necessary, given how little code is here. It might be clearer to follow for future developers to just have the two lines in each model.

services/QuillLMS/app/models/student_response.rb
```diff
@@ -0,0 +1,257 @@
...
+  private_class_method def self.extract_extra_metadata(metadata)
+    extra_metadata = metadata.except(*KNOWN_METADATA_KEYS).reject{ |_,v| v.nil? || (v.is_a?(Enumerable) && v.empty?) }
+
+    return if extra_metadata.empty?
+
+    StudentResponseExtraMetadata.new(metadata: extra_metadata)
+  end
+
+  private_class_method def self.set_normalized_text(target, data_hash)
+    metadata = data_hash[:metadata]
```
dandrabik: Would it be better to just pass in `metadata` to this method?

services/QuillLMS/app/models/student_response.rb
```diff
@@ -0,0 +1,257 @@
...
+    # same process for all of them:
+    #  1) If the key for the string isn't provided, skip it
+    #  2) Find the ID for this exact string if we've seen it before
+    #  3) Create a new normalized record if this string is brand new
+    #  4) Link the normalized reference to the StudentResponse model
+    # This array of arrays is used to execute that logic on each of the
+    # six strings we care about without actually writing the logic six
+    # different times.
+
+    attribute_model_value_tuples = [
```
dandrabik: So I wonder if the long names are driving this meta-programming. If 1) this were an instance method (as it should be), 2) the names were shorter, and 3) the texts had validations for empty values, the non-meta-programming equivalent of this code could be:
```ruby
self.response_answer = ResponseAnswer.find_or_create(text: metadata[:answer])
self.response_direction = ResponseDirection.find_or_create(text: metadata[:directions])
self.response_instruction = ResponseInstruction.find_or_create(text: metadata[:instructions])
...
```
There's duplication, but that would be a pretty straightforward method to follow.

services/QuillLMS/app/models/student_response.rb
```diff
@@ -0,0 +1,257 @@
...
+  end
+
+  private def legacy_format_metadata
+    metadata = legacy_format_base_metadata
+    metadata.merge!(student_response_extra_metadata || {})
+    metadata.reject { |_,v| v.blank? }
+  end
+
+  private def legacy_format_base_metadata
+    metadata = {
```
dandrabik: This `metadata=` is not used.
```suggestion
     {
```

services/QuillLMS/app/models/student_response.rb
```diff
@@ -0,0 +1,257 @@
...
+      related_model = model.find_or_create_by(text: value)
+      target.send(attribute, related_model)
+    end
+  end
+
+  # Takes an input of hashes representing old-style ConceptResult
+  # and rolls them up so that a single response has only one entry
+  # plus an array of concept_ids that match to it
+  private_class_method def self.roll_up_concepts(data_hash_array)
+    data_hash_array.each_with_object({}) do |value, rollup|
```
dandrabik: I'm slightly confused here by the specific `key` construction that you are later throwing away with `.values`, but if this is your approach, I think you want to look at `.group_by`.

This is equivalent
```ruby
data_hash_array
  .group_by {|value| [value[:metadata][:questionNumber], value[:metadata][:attemptNumber]]}
  .map {|_,value| {concept_ids: value.map {|v| v[:concept_id}}
```
And returns something of the shape `[{concept_ids: [1,2,3]}, {concept_ids: [1,2,3]}]`. However, this shape isn't what the comment above it states. If it doesn't have them let's add a test for this method.

services/QuillLMS/app/models/student_response.rb
```diff
@@ -0,0 +1,257 @@
...
+      question_uid = activity_session.activity.data['questions'][question_index]['key']
+    end
+
+    return if !question_uid
+
+    Question.find_by_uid(question_uid)
+  end
+
+  private def legacy_format_metadata
+    metadata = legacy_format_base_metadata
```
dandrabik: To do this in an immutable way, consider the equivalent code:
```ruby
legacy_format_base_metadata
  .merge(student_response_extra_metadata || {})
  .reject { |_,v| v.blank? }

File: services/QuillLMS/db/migrate/20220505154357_create_student_response_answer_text.rb [hunk 15]
```diff
@@ -0,0 +1,18 @@
+# frozen_string_literal: true
+
+class CreateStudentResponseAnswerText < ActiveRecord::Migration[5.1]
+  def change
+    create_table :student_response_answer_texts do |t|
```
dandrabik: I think this is really surprising (and I think you agree based on this 6-line comment). We should not do this. I think trying to abstract this has led us a bit astray here.

services/QuillLMS/app/workers/copy_concept_results_to_student_responses_worker.rb
```diff
@@ -0,0 +1,13 @@
+# frozen_string_literal: true
+
+class CopyConceptResultsToStudentResponsesWorker
```
dandrabik: What's the case for the job looping through a batch of `concept_ids` vs. 1 job per `concept_id`? This isn't idempotent, so wouldn't we end up with duplicate `student_responses` if this job errors midway through?

services/QuillLMS/db/migrate/20220505155016_create_student_response.rb
```diff
@@ -0,0 +1,22 @@
...
+class CreateStudentResponse < ActiveRecord::Migration[5.1]
+  def change
+    create_table :student_responses do |t|
+      t.references :activity_session, null: false, foreign_key: true
+      t.integer :attempt_number
+      t.boolean :correct, null: false
+      t.references :question
+      t.integer :question_number
+      t.float :question_score
+      t.references :student_response_answer_text, foreign_key: true
```
dandrabik: Do we have any concerns that these foreign keys will cause performance issues on this giant table. FK's do slow down inserts and they often removed from very, very large tables for performance reasons. Probably worth doing a little bit of research on this and/or keeping an eye on this if you do keep the foreign keys.

services/QuillLMS/spec/factories/student_response_answer_text.rb
```diff
@@ -0,0 +1,7 @@
+# frozen_string_literal: true
+
+FactoryBot.define do
+  factory :student_response_answer_text, class: 'StudentResponseAnswerText' do
+    sequence(:text) { |n| "This a student response answer #{n}." }
```
dandrabik: I think the naming of this field even tricked yourself (you have this as a `jsonb` column). Again, let's change that name.

services/QuillLMS/app/models/student_response.rb
```diff
@@ -0,0 +1,257 @@
...
+  def self.create_from_concept_result(concept_result)
+    return if concept_result.student_response
+
+    student_response = StudentResponse.find_by(
+      activity_session: concept_result.activity_session,
+      attempt_number: concept_result.metadata['attemptNumber'],
+      question_number: concept_result.metadata['questionNumber']
+    )
+
+    StudentResponse.transaction do
```
dandrabik: I'd be curious with this transaction and the logic, how long with it take to run 1.7 billion of these?

services/QuillLMS/app/models/student_response.rb
```diff
@@ -0,0 +1,257 @@
...
+
+    metadata = data_hash[:metadata]
+
+    question = calculate_question_from_hash(data_hash)
+
+    # Make sure that singular concept_id gets converted to array of
+    # conept_ids if that hasn't been done by an earlier step
+    data_hash[:concept_ids] = [data_hash[:concept_id]] unless data_hash.key?(:concept_ids) && !data_hash[:concept_ids].empty?
+
+    extra_metadata = extract_extra_metadata(metadata)
```
dandrabik: There's a mix of class methods/variables, that likely should be instance variables/methods, e.g. you might consider setting an `attr_accessor` of `metadata` on the student_response model, then you could set everyone on the instance from that.

dandrabik (changes requested):
Thanks for all of your work on this. I left a good amount of comments, so it might be worth talking through them together.

Some of the major items:
1. Renaming the tables to be a little more succinct.
2. Remove some abstraction from a small mixin
3. Remove some metaprogramming in favor of slightly duplicative, but simple code.

Plus a bunch of smaller comments. It's worth reading through first, then we should chat.

services/QuillLMS/app/models/student_response.rb
anathomical: I think you might be missing that on line 219 we pull `value` into the rollup value.  This behavior is covered in a test (`student_response_spec.rb:157`).

services/QuillLMS/app/workers/copy_concept_results_to_student_responses_worker.rb
anathomical: It may be reasonable to rework this from a batch executor to an individual executor just to keep things simple.  It sort of felt to me like putting fewer jobs on the queue might be a benefit, but on second thought I'm not totally sure that's true.

As for idempotency, `StudentResponse.create_from_concept_result` itself should be idempotent (there are tests around that), so if some items in a batch fail, rerunning the whole batch should be safe.

services/QuillLMS/app/models/concerns/is_student_response_normalized_text.rb
anathomical: I am (perhaps incorrectly) given to understand that this is how you validate non-null values while allowing an empty string.

services/QuillLMS/app/models/concerns/is_student_response_normalized_text.rb
anathomical: I was on the fence about it myself.  I'm totally glad to just put the two lines in the model files.  Especially with the comments about `json` fields you left elsewhere.  If we're not going to use the abstract in all six models, we might as well just be explicit in them instead.

services/QuillLMS/app/models/student_response.rb
anathomical: This seems like a totally reasonable suggestion.  I was uncomfortable with the long names, but already had it in my head, so couldn't come up with a better alternative.  I'm going to implement this suggestion since I think you're right that it's more readable.

services/QuillLMS/app/models/student_response.rb
anathomical: Seems reasonable.  Will make that change.

services/QuillLMS/app/models/student_response.rb
anathomical: Seems reasonable.  Will change.

services/QuillLMS/app/models/student_response.rb
anathomical: Not entirely obvious given the repetition, but while we pull 5 of our six values out of `metadata`, the value for `question_type` is expected to come from `data_hash` directly.

services/QuillLMS/app/models/student_response.rb
anathomical: I did want to guard against `nil` values, but I can just do that with guard statements.  I'll update to your suggested approach.

services/QuillLMS/app/models/student_response.rb
anathomical: Good catch.  Removed.

services/QuillLMS/app/models/student_response.rb
anathomical: Seems reasonable.  Will update.

services/QuillLMS/app/models/student_response.rb
anathomical: That said, I think that `group_by` is a much cleaner approach that the stuff I'm doing with `key`, so I think I can convert the code I have here to something using that approach.  Good suggestion.  (I wasn't aware of `group_by`, but it's definitely much cleaner for this application.)

services/QuillLMS/db/migrate/20220505154357_create_student_response_answer_text.rb
anathomical: Went back to calling this column `json`

services/QuillLMS/spec/factories/student_response_answer_text.rb
anathomical: Name updated.

services/QuillLMS/app/models/response.rb
```diff
@@ -0,0 +1,145 @@
...
+
+  validates_exclusion_of :correct, in: [nil]
+
+  # This is a list of keys from the old ConceptResults records
+  # that this model knows how to normalize.  Any keys that are
+  # fed into create_from_json that aren't from this list,
+  # will be stashed in a related ResponseExtraMetadata
+  # record
+  KNOWN_METADATA_KEYS = [
+    :answer,
```
happythenewsad: Were any of these keys always present, during your sampling? if so, we can make our data shape stricter by enforcing `null: false` in the migration file. Even if some are not always present, or if we're not 100% confident since we didn't sample the entire ConceptResult table, we could add sane defaults.

services/QuillLMS/app/models/response_instructions.rb
```diff
@@ -0,0 +1,19 @@
+# frozen_string_literal: true
+
+# == Schema Information
+#
+# Table name: response_instructions
+#
+#  id         :integer          not null, primary key
+#  text       :text             not null
```
happythenewsad: Does a not null text column default to `''`?

services/QuillLMS/app/workers/enqueue_concept_results_migration_worker.rb
```diff
@@ -0,0 +1,19 @@
+# frozen_string_literal: true
+
+class EnqueueConceptResultsMigrationWorker
+  include Sidekiq::Worker
+  sidekiq_options queue: SidekiqQueue::MIGRATION
+
+  BATCH_SIZE=100000
+
+  def perform(start, finish)
```
happythenewsad, line 9: This could have been a rake task. Personal preference, I guess.

services/QuillLMS/app/workers/copy_concept_results_to_responses_worker.rb
```diff
@@ -0,0 +1,54 @@
+# frozen_string_literal: true
+
+class CopyConceptResultsToResponsesWorker
```
happythenewsad: Please add a note in your rollout strategy to actively monitor the Sidekiq UI for job failures. We'll need to notice these and also capture the args for the failed job, in order to know what to re-run.

happythenewsad (approved):
Approving, assuming that my comments about null and default column values have already been anticipated.

services/QuillLMS/db/migrate/20220505154951_create_response_prompt.rb
```diff
@@ -0,0 +1,15 @@
+# frozen_string_literal: true
+
+class CreateResponsePrompt < ActiveRecord::Migration[5.1]
+  def change
+    create_table :response_prompts, id: false do |t|
+      t.serial :id, limit: 4, primary_key: true
```
brendanshean: What's the reason for using `t.serial` here?

brendanshean: Nevermind. I use `create_table :response_prompts, id: :integer` for `limit: 4` and then `:bigint` for `limit: 8`. I had not seen `t.serial` before.

services/QuillLMS/app/workers/copy_concept_results_to_responses_worker.rb
```diff
@@ -0,0 +1,54 @@
...
+    Response.bulk_insert(ignore: true) do |worker|
+      ConceptResult.where(id: start..finish).each do |concept_result|
+
+        directions = concept_result.metadata['directions']
+        instructions = concept_result.metadata['instructions']
+        previous_feedback = concept_result.metadata['lastFeedback']
+        prompt = concept_result.metadata['prompt']
+        question_type = concept_result.question_type
+
+        directions_cache[directions] ||= ResponseDirections.find_or_create_by(text: directions)&.id if directions
```
brendanshean: Are we concerned with any `find_or_create_by` silently failing?  I realize the data is already in the database, but the unknown unknowns always lurk, especially with so many records..  Maybe `add a find_or create_by!`  and a corresponding `rescue` clause that documents the failed creation?

services/QuillLMS/spec/workers/copy_concept_results_to_responses_worker_spec.rb
```diff
@@ -0,0 +1,80 @@
...
+      expect(response.response_prompt.text).to eq(metadata[:prompt])
+      expect(response.response_question_type.text).to eq(concept_result.question_type)
+    end
+
+    it 'should be idempotent so that if the same ID is present twice, only one new record is created' do
+      expect do
+        subject.perform(concept_result.id, concept_result.id)
+        subject.perform(concept_result.id, concept_result.id)
+      end.to change(Response, :count).by(1)
+         .and change(Response, :count).by(1)
```
brendanshean: What's the purpose of this conjunct?

brendanshean (approved):
Great work on this!   Nice clear testing and separation of new concepts.  I left a few comments but nothing major that should hold up approval.

services/QuillLMS/app/models/response_instructions.rb
anathomical: Not unless specified.  Basically there must be a specified string here, never an inferred one.  If a `Response` doesn't have (in this case) an `instructions` value, no `ResponseInstructions` will be created, and the foreign key reference in the `Response` will be `NULL`, not pointing to a non-existent value.

services/QuillLMS/app/models/response.rb
anathomical: All of the text normalization handles nulling sort of automatically via foreign keys, but do you have thoughts on `NOT NULL`ing any of the other columns?

services/QuillLMS/db/migrate/20220505154951_create_response_prompt.rb
anathomical: I hadn't either until I was trying to figure out how to specify auto incremented IDs directly.  The internet taught me.

services/QuillLMS/app/workers/copy_concept_results_to_responses_worker.rb
anathomical: What would a failure look like?  Are you imagining that it somehow fails to find, tries to do a create, but then the `create` fails?  That seems like a reasonable thing to worry about.  I think we could just get away with the just making it `find_or_create_by!` and not rescuing.  As long as we're monitoring the retry queue, we should be able to keep things cleaned up.

services/QuillLMS/spec/workers/copy_concept_results_to_responses_worker_spec.rb
anathomical: Good catch.  I think that may have been a different value in some other version of the approach, but it is definitely unnecessary now.

services/QuillLMS/app/workers/copy_concept_results_to_responses_worker.rb
brendanshean: That sounds good.  I just realized that `rescue` clauses add a performance penalty, and given the size of this migration, it's probably better to just monitor the retry queue like you suggest.

services/QuillLMS/app/models/response.rb
```diff
@@ -0,0 +1,145 @@
...
+    extra_metadata = self.class.parse_extra_metadata(metadata)
+
+    return if extra_metadata.empty?
+
+    self.extra_metadata = extra_metadata
+  end
+
+  private def legacy_format_metadata
+    legacy_format_base_metadata
+      .merge!(extra_metadata || {})
```
dandrabik: This should use the non-destructive merge to maintain immutability
```suggestion
      .merge(extra_metadata || {})
```

services/QuillLMS/app/models/response.rb
```diff
@@ -0,0 +1,145 @@
...
+  belongs_to :activity_session
+  belongs_to :concept
+  belongs_to :concept_result
+  belongs_to :response_directions
+  belongs_to :response_instructions
+  belongs_to :response_previous_feedback
+  belongs_to :response_prompt
+  belongs_to :response_question_type
+
+  validates_exclusion_of :correct, in: [nil]
```
dandrabik: Since this is a boolean, would it be more explicit to validate this way?:
```ruby
validates :correct, inclusion: {in: [true, false]}

File: services/QuillLMS/app/models/response_directions.rb [hunk 30]
```diff
@@ -0,0 +1,19 @@
...
+#  created_at :datetime         not null
+#
+# Indexes
+#
+#  index_response_directions_on_text  (text) UNIQUE
+#
+class ResponseDirections < ApplicationRecord
+  has_many :responses
+
+  validates :text, uniqueness: true, length: { minimum: 1, allow_nil: false }
```
dandrabik: I believe your length validations are equivalent to the standard `presence: true`, e.g.
```suggestion
  validates :text, uniqueness: true, presence: true
```

services/QuillLMS/app/models/response_instructions.rb
```diff
@@ -0,0 +1,19 @@
...
+#  created_at :datetime         not null
+#
+# Indexes
+#
+#  index_response_instructions_on_text  (text) UNIQUE
+#
+class ResponseInstructions < ApplicationRecord
+  has_many :responses
+
+  validates :text, uniqueness: true, length: { minimum: 1, allow_nil: false }
```
dandrabik: Note the `presence: true` applies to all of these

services/QuillLMS/app/workers/copy_concept_results_to_responses_worker.rb
```diff
@@ -0,0 +1,54 @@
...
+
+        directions_cache[directions] ||= ResponseDirections.find_or_create_by!(text: directions)&.id if directions
+        instructions_cache[instructions] ||= ResponseInstructions.find_or_create_by!(text: instructions)&.id if instructions
+        previous_feedbacks_cache[previous_feedback] ||= ResponsePreviousFeedback.find_or_create_by!(text: previous_feedback)&.id if previous_feedback
+        prompts_cache[prompt] ||= ResponsePrompt.find_or_create_by!(text: prompt)&.id if prompt
+        question_types_cache[question_type] ||= ResponseQuestionType.find_or_create_by!(text: question_type)&.id if question_type
+
+        extra_metadata = Response.parse_extra_metadata(concept_result.metadata)
+
+        worker.add({
```
dandrabik: Let's name this something other than `worker`, since it's slightly confusing within a file named `_worker`. Maybe `bulk_insert`?

services/QuillLMS/app/workers/copy_concept_results_to_responses_worker.rb
```diff
@@ -0,0 +1,54 @@
...
+  # rubocop:disable Metrics/CyclomaticComplexity
+  def perform(start, finish)
+    directions_cache = {}
+    instructions_cache = {}
+    previous_feedbacks_cache = {}
+    prompts_cache = {}
+    question_types_cache = {}
+
+    Response.bulk_insert(ignore: true) do |worker|
+      ConceptResult.where(id: start..finish).each do |concept_result|
```
dandrabik: This will load 100,000 records into memory, but only working on one at a time. We should use [find_each](https://apidock.com/rails/ActiveRecord/Batches/find_each) to find this in smaller batches (even something like 10,000 records). Or you could adjust the batch size.

services/QuillLMS/app/workers/copy_concept_results_to_responses_worker.rb
```diff
@@ -0,0 +1,54 @@
+# frozen_string_literal: true
+
+class CopyConceptResultsToResponsesWorker
+  include Sidekiq::Worker
+  sidekiq_options queue: SidekiqQueue::MIGRATION
+
+  BATCH_SIZE=100000
```
dandrabik: Small style nit, you can use the `_` symbol in Ruby integers to make large ones more readable, e.g.
```suggestion
  BATCH_SIZE = 100_000
```

services/QuillLMS/app/models/response.rb
```diff
@@ -0,0 +1,145 @@
...
+#  response_instructions_id      :integer
+#  response_previous_feedback_id :integer
+#  response_prompt_id            :integer
+#  response_question_type_id     :integer
+#
+# Indexes
+#
+#  index_responses_on_concept_result_id  (concept_result_id) UNIQUE
+#
+class Response < ApplicationRecord
```
dandrabik: Naming-wise, does response still make sense given how the scope has changed? Should this be something like `concept_results`?

dandrabik (changes requested):
Overall this looks good. A bunch of small comments for tweaks to make.

The one item I'd like to discuss is the naming of these tables. I believe with this new shape we discussed naming them something along the lines of `concept_results` (and naming the old table `concept_results_legacy` or something). Given it affects so much code, we should lock that down before rolling this out. Let's chat tomorrow morning.

services/QuillLMS/spec/rails_helper.rb
```diff
@@ -34,6 +34,10 @@
   end
 end
 
+# Define any negatated matchers we need
+# https://relishapp.com/rspec/rspec-expectations/v/3-8/docs/define-negated-matcher
+RSpec::Matchers.define_negated_matcher :not_change, :change
+
```
brendanshean: @anathomical I'm really happy you found this.  I've started using it elsewhere.

services/QuillLMS/app/models/concept_result.rb
```diff
@@ -100,19 +100,19 @@ def legacy_format
...
 
   def assign_normalized_text(data_hash)
     metadata = data_hash[:metadata]
 
-    self.response_directions = ResponseDirections.find_or_create_by(text: metadata[:directions])
-    self.response_instructions = ResponseInstructions.find_or_create_by(text: metadata[:instructions])
-    self.response_previous_feedback = ResponsePreviousFeedback.find_or_create_by(text: metadata[:lastFeedback])
-    self.response_prompt = ResponsePrompt.find_or_create_by(text: metadata[:prompt])
-    self.response_question_type = ResponseQuestionType.find_or_create_by(text: data_hash[:question_type])
+    self.concept_result_directions = ConceptResultDirections.find_or_create_by(text: metadata[:directions])
```
brendanshean, line 114: Probably too late with the upcoming Rails 6 upgrade, but something to be aware of in the future; 
 https://www.bigbinary.com/blog/rails-6-adds-create_or_find_by

services/QuillLMS/spec/factories/concept_result.rb
```diff
@@ -0,0 +1,16 @@
+# frozen_string_literal: true
+
+FactoryBot.define do
+  factory :concept_result, class: 'ConceptResult' do
```
brendanshean: In all of the subsequent factory definitions, I don't think it's necessary to explicitly name `class: [ClassName]`.  Factory bot should infer it [automatically](https://github.com/thoughtbot/factory_bot/blob/master/GETTING_STARTED.md#specifying-the-class-explicitly).

brendanshean (approved):

services/QuillLMS/spec/factories/concept_result.rb
anathomical: This probably is true, but there are, like, 15 or so examples in our Factory files that use this pattern.  I was trying to maintain that pattern.  However, another 60 or so factories don't follow that pattern.  Maybe it's worth creating a card to just generally clean this up to unify the approach?

services/QuillLMS/app/models/concept_result.rb
```diff
@@ -0,0 +1,145 @@
...
+  end
+
+  def self.find_or_create_from_old_concept_result(old_concept_result)
+    return old_concept_result.concept_result if old_concept_result.concept_result
+
+    create_from_json(old_concept_result.as_json)
+  end
+
+  def self.parse_extra_metadata(metadata)
+    metadata.deep_symbolize_keys.except(*KNOWN_METADATA_KEYS).reject{ |_,v| v.nil? || (v.is_a?(Enumerable) && v.empty?) }
```
dandrabik: Let's put code like this on multiple lines for readability
```suggestion
    metadata
      .deep_symbolize_keys
      .except(*KNOWN_METADATA_KEYS)
      .reject{ |_,v| v.nil? || (v.is_a?(Enumerable) && v.empty?) }
```

services/QuillLMS/spec/factories/concept_result.rb
dandrabik: If the factory is the same name snakecase as the class, then `class:` is extraneous. Let's remove the `class:` flags from this PR. 

Note, there are plenty of non-optimal patterns in the repo, but PRs are the proper place to ensure we are following proper patterns for new code.

dandrabik (approved):
A few small suggestions, but overall :+1:. Thanks for putting this together.

services/QuillLMS/spec/factories/concept_result.rb
brendanshean: @anathomical I have a side [branch](https://github.com/empirical-org/Empirical-Core/tree/bs-standardize-factory-names) for factory cleanup so I can remove those other ones there.

PR #9191, opened 2022-05-19
evidence/remove-extra-feedback-controller-endpoints

anathomical:
# NOTE: We currently plan to hold off on merging this until we have a replacement for specifying checks that we want to limit ourselves to.  That coming in a separate PR (#9281).

## WHAT
Remove API-specific Feedback controller actions
## WHY
Since we now use `create` to run an iterative check on all different APIs rather than using an external service (like Go) to hit each one separately.
## HOW
Remove routing for the extra actions, remove the action definitions in the controller, and remove all the old tests that pointed at this endpoints.

### Notion Card Links
https://www.notion.so/quill/Remove-extra-Evidence-FeedbackController-endpoints-for-individual-APIs-4dd3c3d6608f4410820bf7f6ef78b89f

PR Checklist | Your Answer
------------ | -------------


happythenewsad (approved):
good to go, but could you also add [this work ](https://quill.slack.com/archives/C03UVPPNH/p1651171747051439?thread_ts=1651167173.571719&cid=C03UVPPNH)to the PR, or make a follow-on ticket? thanks

anathomical:
> good to go, but could you also add [this work ](https://quill.slack.com/archives/C03UVPPNH/p1651171747051439?thread_ts=1651167173.571719&cid=C03UVPPNH)to the PR, or make a follow-on ticket? thanks

Good question.  I was thinking to do it separately, but that may not be a good idea if we're actively using Postman for testing specific APIs right now.  Is that something you're doing?  Should we make sure that we don't remove these endpoints until we have a replacement method for testing?

happythenewsad:
> > good to go, but could you also add [this work ](https://quill.slack.com/archives/C03UVPPNH/p1651171747051439?thread_ts=1651167173.571719&cid=C03UVPPNH)to the PR, or make a follow-on ticket? thanks
> 
> Good question. I was thinking to do it separately, but that may not be a good idea if we're actively using Postman for testing specific APIs right now. Is that something you're doing? Should we make sure that we don't remove these endpoints until we have a replacement method for testing?

I have used the individual endpoints often in the past couple sprints. That said, I don't have an immediate need. As long as it's addressed within the next couple weeks, things should be fine.

anathomical:
> I have used the individual endpoints often in the past couple sprints. That said, I don't have an immediate need. As long as it's addressed within the next couple weeks, things should be fine.

I think it makes sense to add some sort of solution here rather than risk stranding us with no method for testing.  This isn't urgent.  I'll get something figured out and added.

services/QuillLMS/engines/evidence/spec/controllers/evidence/feedback_controller_spec.rb
```diff
@@ -321,327 +320,5 @@ module Evidence
...
-        parsed_response = JSON.parse(response.body)
-        expect(parsed_response["optimal"]).to be true
-      end
-
-      it 'should return 404 if prompt id does not exist' do
-        post("plagiarism", :params => ({ :entry => "No plagiarism here.", :prompt_id => 100, :session_id => 1, :previous_feedback => ([]) }), :as => :json)
-        expect(response.status).to(eq(404))
-      end
-
-      it 'should return successfully when there is plagiarism that matches the first plagiarism text string' do
```
dandrabik, line 525: I think these tests should be ported to the `create` endpoint. You can isolate plagiarism by stubbing the check to only run plagiarism. `stub_const("Evidence::Check::ALL_CHECKS", [Check::Plagiarism])`

dandrabik (changes requested):
So overall this looks good. 

'Requesting changes' since I think we should port some of these tests to the `create` endpoint, rather than delete them. This is something I probably should have done when I created the `create` endpoint, but I only ported the happy path tests, since the code was identical and I wasn't 100% sure we were going to use the `create` endpoint. 

Now that we are committed to the `create` endpoint and since we are deleting the other endpoints, let's port all of the relevant tests over so that coverage is not lost. 

Note, you can use `stub_const("Evidence::Check::ALL_CHECKS", [Check::Plagiarism])` to isolate a check type.

anathomical:
@happythenewsad I've added a mechanism to specify specific algorithms to run on the main action now.  Let me know if you think that'll work well for us, or if we should consider another approach.  @dandrabik I refactored the old tests to work with the new endpoint.  This involved removing a couple of cases that no longer apply in the new approach.

happythenewsad:
> @happythenewsad I've added a mechanism to specify specific algorithms to run on the main action now. Let me know if you think that'll work well for us, or if we should consider another approach. @dandrabik I refactored the old tests to work with the new endpoint. This involved removing a couple of cases that no longer apply in the new approach.

Yeah, your new solution looks great, thanks

services/QuillLMS/engines/evidence/app/controllers/evidence/feedback_controller.rb
```diff
@@ -4,75 +4,17 @@ module Evidence
   require 'json'
 
   class FeedbackController < ApiController
-    before_action :set_params, only: [:automl, :plagiarism, :regex, :spelling, :create]
+    before_action :set_params, only: [:create]
```
happythenewsad, line 7: nit: we no longer need the `only` specifier

services/QuillLMS/engines/evidence/lib/evidence/evidence/check.rb
```diff
@@ -59,5 +61,16 @@ def self.fallback_feedback
...
       FALLBACK_RESPONSE
     end
+
+    def self.checks_to_run(feedback_types)
+      return ALL_CHECKS if feedback_types.nil? || feedback_types.empty?
+
+      qualified_feedback_types = feedback_types.map { |t| "Evidence::Check::#{t}" }
+      filtered_checks = ALL_CHECKS.select { |check| qualified_feedback_types.include?(check.name) }
+
+      raise NoMatchedFeedbackTypesError if filtered_checks.empty?
```
happythenewsad: This changes the business logic beyond the scope of the Notion spec. I suggest sticking with the original behavior of returning FALLBACK_FEEDBACK rather than raise an error in this case.  

If there's a good reason to fail here, let's mention it in the spec, and raise it earlier at the controller params level.

happythenewsad (changes requested):
Looks great except for my comment on `NoMatchedFeedbackTypesError`

services/QuillLMS/engines/evidence/lib/evidence/evidence/check.rb
anathomical: We could do that, but returning default feedback will, generally speaking, make it impossible for you to tell the difference between "the response is optimal for the specified algorithms" and "the algorithms specified are invalid".  You'll get exactly the same response in both cases.  That ambiguity strikes me as a pretty bad outcome.

I'm also not clear on what the risk of 500-ing sometimes is, but maybe I'm missing something obvious?

services/QuillLMS/engines/evidence/app/controllers/evidence/feedback_controller.rb
anathomical: Agree that it's not strictly necessary, but I generally prefer to specify `only` for `before_actions` even if the intended list is all of them so someone doesn't run into issues if they add a new action later.  Could be convinced that this is overly verbose, though.

services/QuillLMS/engines/evidence/lib/evidence/evidence/check.rb
happythenewsad: Screen Shot 2022-06-15 at 2 11 58 PM

If a user ever got into this state, the UX would fail hard, for example. The evidence flow doesn't currently handle API failures gracefully.

happythenewsad: I agree with your concern about different errors getting compressed into FALLBACK_FEEDBACK in a confusing way, so that should be addressed one way or another.

anathomical: I'm torn.  On the one hand, you're right, the front-end can't really handle errors well.  On the other, would we ever want to use this for anything other than Postman-style API testing?  That is, would the front-end ever send the params necessary to trigger this code branch?  I suppose it's dangerous to assert that something will "never" happen in the future.

Maybe we should just send the error via `ErrorNotifier` and then swallow it?  The downside there is that if you're just looking at HTTP responses, you still can't tell the difference, and need to be in Sentry and/or NewRelic, but perhaps that's acceptable?

happythenewsad: We could avoid a hard error, and also improve messaging, but having the FALLBACK_FEEDBACK display different texts, depending on the code path of the error. Just an idea

services/QuillLMS/engines/evidence/app/controllers/evidence/feedback_controller.rb
```diff
@@ -4,75 +4,17 @@ module Evidence
...
 
     def create
-      feedback = Check.get_feedback(@entry, @prompt, @previous_feedback)
+      feedback = Check.get_feedback(@entry, @prompt, @previous_feedback, @feedback_types)
```
dandrabik: Reading through this notion card and PR description and surprised to see existing functionality change. Maybe I'm missing something?

dandrabik: Seeing the PR comments and commits where this is mentioned. I think it would have been better to isolate this into its own PR, considering it's out-of-scope of the clean-up/refactor PR and it changes the API protocol and functionality.

dandrabik (approved):
:+1: This looks good!

happythenewsad (approved):

PR #9192, opened 2022-05-19
Send a Segment event when a school or user subscription is 30 days from expiring

cissyxyu:
## WHAT
Run a nightly worker that checks to find all subscriptions that are 30 days away from expiring and marked as `recurring: true` and send an event to Sentry that alerts the subscription is going to expire. 

## WHY
This is part of the Subscriptions workflow we're trying to set up in Sentry to allow Peter Sharkey to hook up automated emails to our users with upcoming subscription renewals.

## HOW
Create a worker that fetches all subscriptions that will expire in 30 days and triggers Sentry tracking events.

### Screenshots

### Notion Card Links
https://www.notion.so/quill/Send-renewal-reminders-to-users-with-a-direct-subscription-automatic-renewal-off-1-9158f5cd10bd44e58bf37bf066e71f20
https://www.notion.so/quill/Send-renewal-reminders-to-users-with-a-direct-subscription-automatic-renewal-on-1-cb39bb9cbd80482b968d3e58cccd920e

PR Checklist | Your Answer
------------ | -------------

### DIFF
--- a/services/QuillLMS/app/models/cron.rb
+++ b/services/QuillLMS/app/models/cron.rb
@@ -26,6 +26,7 @@ class Cron
     # subscriptions
     RenewExpiringRecurringSubscriptionsWorker.perform_async
     UpdateExpiringSchoolSubscriptionsWorker.perform_async
+    AlertSoonToExpireSubscriptionsWorker.perform_async
 
     # demo
     ResetDemoAccountWorker.perform_async
--- a/services/QuillLMS/app/models/cron.rb
+++ b/services/QuillLMS/app/models/cron.rb
@@ -19,16 +19,24 @@ class Cron
     # pass yesterday's date for stats email queries and labels
     date = Time.current.getlocal('-05:00').yesterday
 
+    # internal analytics and security
     DailyStatsEmailJob.perform_async(date)
     QuillStaffAccountsChangedWorker.perform_async
+
+    # subscriptions
     RenewExpiringRecurringSubscriptionsWorker.perform_async
     UpdateExpiringSchoolSubscriptionsWorker.perform_async
+
+    # demo
     ResetDemoAccountWorker.perform_async
+
+    # third party analytics
     SyncVitallyWorker.perform_async
+
+    # caching
     MaterializedViewRefreshWorker.perform_async
     RematchUpdatedQuestionsWorker.perform_async(date.beginning_of_day, date.end_of_day)
     PreCacheAdminDashboardsWorker.perform_async
-
     Question::TYPES.each { |type| RefreshQuestionCacheWorker.perform_async(type) }
   end
 
--- a/services/QuillLMS/app/models/subscription.rb
+++ b/services/QuillLMS/app/models/subscription.rb
@@ -118,6 +118,7 @@ class Subscription < ApplicationRecord
   scope :not_recurring, -> { where(recurring: false) }
   scope :not_stripe, -> { where(stripe_invoice_id: nil) }
   scope :started, -> { where("start_date <= ?", Date.current) }
+  scope :paid_with_card, -> { where.not(stripe_invoice_id: nil).or(where(payment_method: 'Credit Card')) }
 
   def is_trial?
     account_type && TRIAL_TYPES.include?(account_type)
--- a/services/QuillLMS/app/services/analytics/segment_analytics.rb
+++ b/services/QuillLMS/app/services/analytics/segment_analytics.rb
@@ -148,24 +148,22 @@ class SegmentAnalytics
 
   end
 
-  def trigger_teacher_subscription_will_expire(subscription_id)
+  def track_teacher_subscription(subscription_id, event)
     subscription = Subscription.find(subscription_id)
     teacher_id = subscription.users.first.id
-    event = subscription.recurring ? SegmentIo::BackgroundEvents::RECURRING_TEACHER_SUB_WILL_EXPIRE : SegmentIo::BackgroundEvents::NON_RECURRING_TEACHER_SUB_WILL_EXPIRE
 
     track({
       user_id: teacher_id,
       event: event,
       properties: {
-        subscription_id: subscription.id
+        subscription_id: subscription_id
       }
     })
   end
 
-  def trigger_school_subscription_will_expire(subscription_id)
+  def track_school_subscription(subscription_id, event)
     subscription = Subscription.find(subscription_id)
     school_id = subscription.schools.first.id
-    event = subscription.recurring ? SegmentIo::BackgroundEvents::RECURRING_SCHOOL_SUB_WILL_EXPIRE : SegmentIo::BackgroundEvents::NON_RECURRING_SCHOOL_SUB_WILL_EXPIRE
 
     if subscription.purchaser_id.present?
       track({
--- a/services/QuillLMS/app/services/analytics/segment_analytics.rb
+++ b/services/QuillLMS/app/services/analytics/segment_analytics.rb
@@ -151,10 +151,11 @@ class SegmentAnalytics
   def trigger_teacher_subscription_will_expire(subscription_id)
     subscription = Subscription.find(subscription_id)
     teacher_id = subscription.users.first.id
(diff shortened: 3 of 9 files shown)


services/QuillLMS/app/models/cron.rb
```diff
@@ -19,16 +19,25 @@ def self.interval_1_day
...
     UpdateExpiringSchoolSubscriptionsWorker.perform_async
+    AlertSoonToExpireSubscriptionsWorker.perform_async
+
+    # demo
     ResetDemoAccountWorker.perform_async
+
+    # third party analytics
     SyncVitallyWorker.perform_async
+
+    # caching
```
emilia-friedberg, line 37: Thanks for adding these comments!

services/QuillLMS/app/services/analytics/segment_analytics.rb
```diff
@@ -148,6 +148,35 @@ def track_previewed_activity(user_id, activity_id)
...
+    })
+  end
+
+  def trigger_school_subscription_will_expire(subscription_id)
+    subscription = Subscription.find(subscription_id)
+    school_id = subscription.schools.first.id
+
+    track({
+      # Segment requires us to send a unique User ID or Anonymous ID for every event
+      # generate a random UUID here because we don't want the School Subscription event to be associated to any real user
```
emilia-friedberg: We have the `purchaser_id` on the subscription itself - maybe it makes sense to use that where we have it? It isn't present on all subscriptions but it seems like it might be useful when it is?

emilia-friedberg (commented):
You have both Notion cards linked here, but I only see the cases in [this card ](https://www.notion.so/quill/Send-renewal-reminders-to-users-with-a-direct-subscription-automatic-renewal-on-1-cb39bb9cbd80482b968d3e58cccd920e) handled. Totally fine by me to do them separately, but since we are going to have to handle the cases where they're not recurring (ie, going to expire and not renew), I think the naming here might be confusing because these are subscriptions that are _not_ going to expire. We might want to call these `will_renew` instead of `will_expire` and save `will_expire` for the events in the other card.

services/QuillLMS/app/services/analytics/segment_analytics.rb
cissyxyu: That makes sense, I'll add this field as user ID when we have the value.

cissyxyu:
> You have both Notion cards linked here, but I only see the cases in [this card ](https://www.notion.so/quill/Send-renewal-reminders-to-users-with-a-direct-subscription-automatic-renewal-on-1-cb39bb9cbd80482b968d3e58cccd920e) handled. Totally fine by me to do them separately, but since we are going to have to handle the cases where they're not recurring (ie, going to expire and not renew), I think the naming here might be confusing because these are subscriptions that are _not_ going to expire. We might want to call these `will_renew` instead of `will_expire` and save `will_expire` for the events in the other card.

Thanks for calling this out -- I had misunderstood the spec when reading it this first time. I've now changed the code to send different events for recurring and non-recurring subscriptions!

services/QuillLMS/app/services/analytics/segment_io.rb
```diff
@@ -47,6 +47,10 @@ module BackgroundEvents
     VIEWED_DEMO = 'User logged into Demo account'
     PREVIEWED_ACTIVITY = 'Teacher previewed an activity'
     TEACHER_COMPLETED_ONBOARDING_CHECKLIST = 'Teacher completed onboarding checklist'
+    RECURRING_TEACHER_SUB_WILL_EXPIRE = 'Teacher Premium Renews in 30 Days | Automatic Renewal On'
+    RECURRING_SCHOOL_SUB_WILL_EXPIRE = 'School Premium Renews in 30 Days | Automatic Renewal On'
+    NON_RECURRING_TEACHER_SUB_WILL_EXPIRE = 'Teacher Premium Renews in 30 Days | Automatic Renewal Off'
+    NON_RECURRING_SCHOOL_SUB_WILL_EXPIRE = 'School Premium Renews in 30 Days | Automatic Renewal Off'
```
emilia-friedberg: I think the language is supposed to be a little different for the ones that are non-renewing: https://www.notion.so/quill/Send-renewal-reminders-to-users-with-a-direct-subscription-automatic-renewal-off-1-9158f5cd10bd44e58bf37bf066e71f20#abb52119c8f24cbd984fd430b5dd3311. I also think we need one for School Premium expiring in 14 days?

services/QuillLMS/app/models/cron.rb
brendanshean: Agreed

services/QuillLMS/app/services/analytics/segment_io.rb
cissyxyu: My bad on this -- completely dropped the ball on this Spec. I went back and made sure to implement it point by point this time, sorry about this! We have 6 different subscription types to account for, and I've changed the logic somewhat to account most efficiently for all of them.

services/QuillLMS/app/models/subscription.rb
```diff
@@ -118,6 +118,7 @@ class RenewalNilStripeCustomer < StandardError; end
   scope :not_recurring, -> { where(recurring: false) }
   scope :not_stripe, -> { where(stripe_invoice_id: nil) }
   scope :started, -> { where("start_date <= ?", Date.current) }
+  scope :paid_with_card, -> { where.not(stripe_invoice_id: nil).or(where(payment_method: 'Credit Card')) }
```
brendanshean, line 121: Nice use of disjunction in the scope!

emilia-friedberg (approved):
Looks great! No worries - lots of cases to parse out and sometimes it's harder when things are spread out over a couple of different cards.

services/QuillLMS/app/workers/alert_soon_to_expire_subscriptions_worker.rb
```diff
@@ -0,0 +1,26 @@
...
+    teacher_subs_renewing_in_thirty = subs_expiring_in_thirty.where(account_type: Subscription::OFFICIAL_TEACHER_TYPES, recurring: true)
+    school_subs_renewing_in_thirty = subs_expiring_in_thirty.where(account_type: Subscription::OFFICIAL_SCHOOL_TYPES, recurring: true)
+    teacher_subs_expiring_in_thirty = subs_expiring_in_thirty.where(account_type: Subscription::OFFICIAL_TEACHER_TYPES, recurring: false)
+    school_subs_expiring_in_thirty = subs_expiring_in_thirty.where(account_type: Subscription::OFFICIAL_SCHOOL_TYPES, recurring: false)
+    teacher_subs_expiring_in_fourteen = Subscription.paid_with_card.where(account_type: Subscription::OFFICIAL_TEACHER_TYPES, expiration: current_time + 14.days, recurring: false)
+    school_subs_expiring_in_fourteen = Subscription.paid_with_card.where(account_type: Subscription::OFFICIAL_SCHOOL_TYPES, expiration: current_time + 14.days, recurring: false)
+
+    analytics = SegmentAnalytics.new
+    teacher_subs_renewing_in_thirty.each { |sub| analytics.track_teacher_subscription(sub.id, SegmentIo::BackgroundEvents::TEACHER_SUB_WILL_RENEW) }
```
brendanshean: Since the `track_school_subscription` and `track_teacher_subscription` methods end up calling `Subscription.find(subscription_id)`, what about changing the interface, and just passing the `sub` object itself? (i.e. `def track_school_subscription(subscription, event)`).

services/QuillLMS/spec/workers/alert_soon_to_expire_subscriptions_worker_spec.rb
```diff
@@ -0,0 +1,55 @@
+# frozen_string_literal: true
+
+require 'rails_helper'
+
+describe AlertSoonToExpireSubscriptionsWorker do
+  subject { described_class.new }
+
+  describe '#perform' do
+    let(:analytics) { double(:analytics).as_null_object }
```
brendanshean, line 9: Nice use of `as_null_object`!

brendanshean (approved):
There are lots of details to keep straight in this one. Nice clear tests to sort them out.  I left a comment on changing the interface, but once that's done this should be good to go!

PR #9195, opened 2022-05-20
fix bug with highlighting

emilia-friedberg:
## WHAT
Fix issue where highlights with `<br/>` tags in them cause code to break.

## WHY
We don't want that to happen.

## HOW
Update the processing to return nothing rather than calling `.data` on something broken. 

### Screenshots

### Notion Card Links
https://www.notion.so/quill/Skateboarding-and-judo-crashing-with-2nd-layer-feedback-08fdef24f5954f538d482b28f071668d

PR Checklist | Your Answer
------------ | -------------


happythenewsad (approved):

anathomical (approved):
Long term we might want to make this logic into a highlight, but this should work for now.

PR #9222, opened 2022-05-31
Add back in user.last_four for stripe legacy subscriptions

brendanshean:
## WHAT
Add back in `last_four` method on the `User` model.

## WHY
Previously, we charged subscriptions through stripe without them actually being subscriptions on stripe, but instead just a purchase.  However, when we display our subscriptions page, we currently pull `last_four` for subscriptions via `stripe_invoice_id`.  Since older subscriptions do not have a `stripe_invoice_id`, we need to pull the `last_four` from the user model itself using `stripe_customer_id`.

## HOW
Revert this [method](https://github.com/empirical-org/Empirical-Core/blob/59566e8958834aecd0d6833219be14f0d1108cb3/services/QuillLMS/app/models/user.rb#L304) 

### Screenshots

### Notion Card Links
https://www.notion.so/quill/Credit-card-last-four-showing-null-for-users-on-an-old-style-Stripe-subscription-903fa69f11894fd582f759026d250f63

PR Checklist | Your Answer
------------ | -------------

### DIFF
--- a/services/QuillLMS/app/models/subscription.rb
+++ b/services/QuillLMS/app/models/subscription.rb
@@ -112,7 +112,7 @@ class Subscription < ApplicationRecord
 
   validates :stripe_invoice_id, allow_blank: true, stripe_uid: { prefix: :in }
 
-  delegate :stripe_cancel_at_period_end, :last_four, :stripe_subscription_id, :stripe_subscription_url,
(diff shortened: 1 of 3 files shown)


dandrabik (approved):
:+1: Thanks for fixing this!

PR #9228, opened 2022-06-01
Datepicker

happythenewsad:
## WHAT
By default, set the 'start time' filter in Rules Analysis to the time the automl model was last updated.

## WHY
Curriculum team request

### Screenshots

### Notion Card Links
https://www.notion.so/quill/Product-Board-30f97e4eb01246dbb8264253fb385073?p=40d447da79b94edf9ac4a9e213b30961

PR Checklist | Your Answer
------------ | -------------

### DIFF
--- a/services/QuillLMS/client/app/bundles/Staff/components/evidence/rulesAnalysis/rulesAnalysis.tsx
+++ b/services/QuillLMS/client/app/bundles/Staff/components/evidence/rulesAnalysis/rulesAnalysis.tsx
@@ -14,6 +14,7 @@ import { fetchActivity } from '../../../utils/evidence/activityAPIs';
 import { fetchRuleFeedbackHistories } from '../../../utils/evidence/ruleFeedbackHistoryAPIs';
 import { DropdownInput, Spinner, ReactTable, expanderColumn, } from '../../../../Shared/index';
 import { RULES_ANALYSIS } from '../../../../../constants/evidence';
+import { fetchModels } from "../../../utils/evidence/modelAPIs";
 
 const DEFAULT_RULE_TYPE = 'All Rules'
 
@@ -55,11 +56,18 @@ const MoreInfo = (row) => {
   )
 }
 
+const getDateFromLatestAutoMLModel = (models) => {
+  const latestModel = models?.find( model => model.state === 'active')
+  console.log("getDateFromLatestAutoMLModel called, latest model:", latestModel)
+  if (!latestModel) { return '' }
+  return latestModel.created_at
+}
+
 const RulesAnalysis: React.FC<RouteComponentProps<ActivityRouteProps>> = ({ history, match }) => {
   const { params } = match;
   const { activityId, promptConjunction, } = params;
   const today = new Date();
-  const thirtyDaysAgo = new Date().setDate(today.getDate()-30);
+  const thirtyDaysAgo = new Date().setDate(today.getDate()); // TODO: REVERT
 
   const ruleTypeValues = [DEFAULT_RULE_TYPE].concat(Object.keys(apiOrderLookup))
   const ruleTypeOptions = ruleTypeValues.map(val => ({ label: val, value: val, }))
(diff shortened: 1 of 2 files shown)


services/QuillLMS/client/app/bundles/Staff/components/evidence/rulesAnalysis/rulesAnalysis.tsx
```diff
@@ -104,6 +117,12 @@ const RulesAnalysis: React.FC<RouteComponentProps<ActivityRouteProps>> = ({ hist
     queryFn: fetchActivity
   });
 
+  React.useEffect(() => {
```
happythenewsad, line 120: I'm using this hook because we want this to happen at component load time, but also react to the async fetchModels request. Please suggest another technique if this is not ideal.

services/QuillLMS/client/app/bundles/Staff/components/evidence/rulesAnalysis/rulesAnalysis.tsx
```diff
@@ -104,6 +117,12 @@ const RulesAnalysis: React.FC<RouteComponentProps<ActivityRouteProps>> = ({ hist
...
+    onStartDateChange(new Date(getDateFromLatestAutoMLModel(modelsData?.models)))
+    setStartDate(getDateFromLatestAutoMLModel(modelsData?.models))
```
happythenewsad: We could clean up this code a bit by collapsing the `startDate` and `startDateForQuery` state hooks into a single hook. I *think* the only reason why we have both is that one is string format, and the other is date format, which we could address by adding a conversion function between the two (since downstream components expect both different formats).

happythenewsad:
Ignore the lint errors. I am addressing them in a separate PR.

services/QuillLMS/client/app/bundles/Staff/components/evidence/rulesAnalysis/rulesAnalysis.tsx
eadams17: These were originally two separate pieces of state because we didn't want to trigger a request on first load until after the filter button was clicked. The date could be changed multiple times in the datepicker, stored to `startDate` and then `startDateForQuery` is set once the user clicks filter, triggering the request and bypassing the early return [here ](https://github.com/empirical-org/Empirical-Core/blob/develop/services/QuillLMS/client/app/bundles/Staff/utils/evidence/ruleFeedbackHistoryAPIs.ts#L6)

If that's no longer the case and we want to trigger this request on load, then you should be fine to collapse them into one.

eadams17: These were originally two separate pieces of state because we didn't want to trigger a request on first load until after the filter button was clicked. The date could be changed multiple times in the datepicker, stored to `startDate` and then `startDateForQuery` is set once the user clicks filter, triggering the request and bypassing the early return [here ](https://github.com/empirical-org/Empirical-Core/blob/develop/services/QuillLMS/client/app/bundles/Staff/utils/evidence/ruleFeedbackHistoryAPIs.ts#L6)

If that's no longer the case and we want to trigger this request on load, then you should be fine to collapse them into one.

eadams17: also I believe you can just wrap these two lines in an if statement checking for the existence of `modelsData`, and if you pull the `getDateFromLatestAutoMLModel` function inside `RulesAnalysis`, you won't need to pass these arguments since you'll have access to `modelsData` within:

```
  React.useEffect(() => {
    if(modelsData && modelsData.models) {
      onStartDateChange(new Date(getDateFromLatestAutoMLModel()))
      setStartDate(getDateFromLatestAutoMLModel())
    }

  }, [modelsData])

  function getDateFromLatestAutoMLModel() {
    const { models } = modelsData
    const latestModel = models.find(model => model.state === 'active')
    if (!latestModel) { return null }
    return latestModel.created_at
  }
  ```

Comment (happythenewsad) [reply services/QuillLMS/client/app/bundles/Staff/components/evidence/rulesAnalysis/rulesAnalysis.tsx]: I've added the if statement check - good idea. I'm keeping `getDateFromLatestAutoMLModel` outside of the RulesAnalysis component, so that it can remain a pure function.

File: services/QuillLMS/client/app/bundles/Staff/components/evidence/rulesAnalysis/rulesAnalysis.tsx [hunk 3]
```diff
@@ -69,24 +77,29 @@ const RulesAnalysis: React.FC<RouteComponentProps<ActivityRouteProps>> = ({ hist
...
   const [turkSessionIDForQuery, setTurkSessionIDForQuery] = React.useState<string>(initialTurkSessionId);
   const [formattedRows, setFormattedRows] = React.useState<any[]>(null);
+  const [startDate, onStartDateChange] = React.useState<Date>(initialStartDate);
 
   const selectedConjunction = selectedPrompt ? selectedPrompt.conjunction : promptConjunction
 
+  const { data: modelsData } = useQuery({
+    queryKey: [`models-${selectedPrompt?.id}`, selectedPrompt?.id],
+    queryFn: fetchModels,
+    enabled: selectedPrompt !== null
```
emilia-friedberg: Unless you have a reason for the more specific case, I would just do `!!selectedPrompt` here.

emilia-friedberg (approved):
One small tweak I might make but LGTM overall.

PR #9231, opened 2022-06-01
Backend code for Enable 'Renew Subscription' 

brendanshean:
## WHAT
When a user attempts to renew a school subscription through stripe associate the relevant schools to that subscription

## WHY
We would like to enable school subscriptions through stripe.

## HOW
When a user attempts to renew a school subscription, add the relevant school_id to the `stripe_subscription.metadata` object under the key `school_ids`.  Upon successful checkout on stripe, the relevant [school_ids](https://github.com/empirical-org/Empirical-Core/pull/9231/files#diff-59492826a34540b6a1a39aaeeae1d87fca0385e29154c3e0d8e4ca3114a79cadR59) can then be pulled from the metadata and saved in our database as new `SchoolSubscription` records.

Note, this code also sends the ground work for new purchases, but those flows will be implemented in another PR

### Screenshots

### Notion Card Links

PR Checklist | Your Answer
------------ | -------------

### DIFF
--- a/services/QuillLMS/app/controllers/stripe_integration/subscription_checkout_sessions_controller.rb
+++ b/services/QuillLMS/app/controllers/stripe_integration/subscription_checkout_sessions_controller.rb
@@ -11,13 +11,14 @@ module StripeIntegration
 
     private def subscription_checkout_session_args
       {
-        success_url: success_url,
         cancel_url: cancel_url,
-        mode: SUBSCRIPTION_MODE,
         line_items: [{
-          quantity: 1,
-          price: stripe_price_id
-        }]
+          price: stripe_price_id,
+          quantity: 1
+        }],
+        mode: SUBSCRIPTION_MODE,
+        subscription_data: subscription_data,
+        success_url: success_url
       }.merge(customer_arg)
     end
 
@@ -37,6 +38,10 @@ module StripeIntegration
       params[:customer_email]
     end
 
+    private def school_ids
+      params[:school_id]
+    end
+
     private def stripe_price_id
       params[:stripe_price_id]
     end
@@ -47,6 +52,12 @@ module StripeIntegration
 
     private def success_url
       "#{subscriptions_url}?checkout_session_id={CHECKOUT_SESSION_ID}"
+    end
+
+    private def subscription_data
+      return {} unless School.exists?(id: school_ids)
+
+      { metadata: { school_ids: school_ids } }
     end
   end
 end
--- a/services/QuillLMS/app/controllers/stripe_integration/subscription_checkout_sessions_controller.rb
+++ b/services/QuillLMS/app/controllers/stripe_integration/subscription_checkout_sessions_controller.rb
@@ -65,9 +65,11 @@ module StripeIntegration
     end
 
     private def success_url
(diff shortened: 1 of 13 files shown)


services/QuillLMS/app/controllers/stripe_integration/subscription_checkout_sessions_controller.rb
```diff
@@ -55,7 +65,15 @@ def create
     end
 
     private def success_url
-      "#{subscriptions_url}?checkout_session_id={CHECKOUT_SESSION_ID}"
+      return "#{subscriptions_url}?checkout_session_id={CHECKOUT_SESSION_ID}" if schools.empty?
+
+      "#{teacher_admin_subscriptions_url(school_id: school_ids.first)}&checkout_session_id={CHECKOUT_SESSION_ID}"
+    end
```
emilia-friedberg: If a teacher associated with a school, but who is not an admin for that school, renews that school's subscription from their My  Subscriptions page, I think this code is going to send them to the school subscriptions page, which will error because it's only available to admins. Let me know if I'm reading this incorrectly, but otherwise I think we'll have to handle this differently - maybe we can just check to see if they are an admin?

emilia-friedberg (commented):
I think this looks really good. Just one question about the redirect logic.

services/QuillLMS/app/controllers/stripe_integration/subscription_checkout_sessions_controller.rb
brendanshean: Good catch.  I will fix that to check for admin.

services/QuillLMS/app/controllers/stripe_integration/subscription_checkout_sessions_controller.rb
```diff
@@ -55,7 +65,17 @@ def create
     end
 
     private def success_url
-      "#{subscriptions_url}?checkout_session_id={CHECKOUT_SESSION_ID}"
+      if schools.empty? || !customer.admin?
```
emilia-friedberg, line 68: It looks to me like `.admin?` checks that the user role is an admin (I think that's defunct) and not that they have schools that they administrate.

emilia-friedberg (changes requested):

services/QuillLMS/app/controllers/stripe_integration/subscription_checkout_sessions_controller.rb
brendanshean: Ironically, that's what I assumed `admin?` was checking, but it looks like it's pulling from `SchoolsAdmins`: https://github.com/empirical-org/Empirical-Core/blob/develop/services/QuillLMS/app/models/user.rb#L369

brendanshean: I think is a gotcha, since the other role booleans (student?, teacher?, staff?) behave as expected.

emilia-friedberg: Oops, you're right. I was looking at the [authentication function](https://github.com/empirical-org/Empirical-Core/blob/feat/admin-reports/services/QuillLMS/app/lib/quill_authentication.rb#L149).

emilia-friedberg (approved):

PR #9234, opened 2022-06-02
bump typescript

happythenewsad:
## WHAT
- Bumps top-level typescript package by a minor version 
- fixes eslint errors that resulted from this version bump

## WHY
So that our commit hook can parse es2021 syntax. Notably: [optional chaining](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Optional_chaining), among other things.
### Screenshots

### Notion Card Links
none

PR Checklist | Your Answer
------------ | -------------
Have you added and/or updated tests? |  no - third party lib only.
Self-Review: Have you done an initial self-review of the code below on Github? | n/a

### DIFF
--- a/services/QuillLMS/client/app/bundles/Teacher/components/teacher_guide/getting_started_mini.jsx
+++ b/services/QuillLMS/client/app/bundles/Teacher/components/teacher_guide/getting_started_mini.jsx
@@ -7,13 +7,13 @@ import LoadingIndicator from '../shared/loading_indicator'
 
 export default class GettingStartedMini extends React.Component {
   checklistSection() {
-    const props = this.props
-    if (props.checkboxData.loading) {
+    const { checkboxData} = this.props
+    if (checkboxData.loading) {
       return <LoadingIndicator />;
     }
     return (
       <CheckboxSection
-        checkboxes={props.checkboxData}
+        checkboxes={checkboxData}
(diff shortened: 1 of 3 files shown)


services/QuillLMS/client/app/bundles/Teacher/components/teacher_guide/getting_started_mini.jsx
```diff
@@ -1,38 +1,43 @@
...
 import PercentageGraph from './percentage_graph'
 import CheckboxSection from './checkbox_sections'
+
 import LoadingIndicator from '../shared/loading_indicator'
 
 export default class GettingStartedMini extends React.Component {
   checklistSection() {
-    if (this.props.checkboxData.loading) {
+    const props = this.props
+    if (props.checkboxData.loading) {
```
emilia-friedberg: I would use deconstruction (`const { checkBoxData, } = this.props`) here rather than just saving the `props` variable (same for the other instances in this file).

emilia-friedberg (approved):
Small syntax change requested but nothing I need to see again. Thanks for taking care of this version bump!

dandrabik (approved):
Agreed with the deconstruction comment, but otherwise looks :+1:. Thanks for bumping this.

PR #9239, opened 2022-06-03
fix bug with find admin users

emilia-friedberg:
## WHAT
Fix bug where `FindAdminUsersWorker` was failing because I'd inadvertently made the scope of a query too narrow, causing a call later to fail because the query was missing rows.

## WHY
We want this to work so admins can load their dashboard.

## HOW
Use the query limiter in the left join instead of the `where` clause, making sure that teacher rows will populate even if they don't have active classrooms. That way the later function should always be able to find the right key in the hash, but just in case, I also added a nil guard there.

### Screenshots

### Notion Card Links

PR Checklist | Your Answer
------------ | -------------


brendanshean (approved):
Thanks for the quick fix!

PR #9241, opened 2022-06-03
Fix various stripe integration bugs

brendanshean:
## WHAT
Address a few bugs from stripe integration
https://sentry.io/organizations/quillorg-5s/issues/3321246859/?referrer=slack
https://sentry.io/organizations/quillorg-5s/issues/3321252188/?referrer=slack
https://sentry.io/organizations/quillorg-5s/issues/3321260200/?referrer=slack

Also, fix `/subscription_payment_methods_controller_spec.rb`, which was a copy-paste error of a different spec.  I've updated it with the actual spec.

## WHY
We want our webhooks from stripe to be processed properly.

## HOW
1.  Fix Stripe::Object access of a `status`.  Sometimes this object does not have a `status` method so we need to check for it first before accessing it.
2. Fix the interface for CustomerSubscriptionDeletedEventHandler.  It should be taking a Stripe::Subscription object but i was wrongly passing it the Stripe::Event wrapper
3. Check to make sure that a Stripe Subscription isn't canceled before trying to update it.

### Screenshots

### Notion Card Links

PR Checklist | Your Answer
------------ | -------------

### DIFF
--- a/services/QuillLMS/app/controllers/stripe_integration/subscription_renewals_controller.rb
+++ b/services/QuillLMS/app/controllers/stripe_integration/subscription_renewals_controller.rb
@@ -2,10 +2,14 @@
 
 module StripeIntegration
   class SubscriptionRenewalsController < ApplicationController
+    class UpdatingCanceledSubscriptionError < StandardError; end
+
     def create
+      raise UpdatingCanceledSubscriptionError if stripe_subscription_canceled?
+
       Stripe::Subscription.update(stripe_subscription_id, cancel_at_period_end: cancel_at_period_end)
       render json: {}, status: 200
-    rescue Stripe::InvalidRequestError => e
+    rescue UpdatingCanceledSubscriptionError, Stripe::InvalidRequestError => e
       ErrorNotifier.report(e, stripe_subscription_id: stripe_subscription_id)
       render json: {}, status: 500
     end
@@ -16,6 +20,10 @@ module StripeIntegration
 
     private def stripe_subscription_id
       params[:stripe_subscription_id]
+    end
+
+    private def stripe_subscription_canceled?
+      Stripe::Subscription.retrieve(stripe_subscription_id).canceled_at.present?
     end
   end
 end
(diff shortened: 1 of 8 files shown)


services/QuillLMS/spec/support/shared_contexts/stripe_integration/stripe_subscription.rb
```diff
@@ -35,7 +37,7 @@
       discount: nil,
       ended_at: nil,
       items: {
-        object: 'list',
+        object: 'lisk',
```
dandrabik, line 40: Not sure if this matters, but is this a typo?

services/QuillLMS/app/controllers/stripe_integration/subscription_renewals_controller.rb
```diff
@@ -2,10 +2,14 @@
...
   class SubscriptionRenewalsController < ApplicationController
+    class UpdatingCanceledSubscriptionError < StandardError; end
+
     def create
+      raise UpdatingCanceledSubscriptionError if stripe_subscription_canceled?
+
       Stripe::Subscription.update(stripe_subscription_id, cancel_at_period_end: cancel_at_period_end)
       render json: {}, status: 200
-    rescue Stripe::InvalidRequestError => e
+    rescue UpdatingCanceledSubscriptionError, Stripe::InvalidRequestError => e
```
dandrabik, line 12: Is this just to track these? Since the outcome (a returned `500` error)

dandrabik (approved):

services/QuillLMS/spec/support/shared_contexts/stripe_integration/stripe_subscription.rb
brendanshean: Whoops.  Vim navigation typo.

services/QuillLMS/app/controllers/stripe_integration/subscription_renewals_controller.rb
brendanshean: This should be an error, but I would be surprised if this ever fires since it's a very weird edge case (user tries to update renewal settings on a subscription after canceling the subscription on stripe UI).   It happened previously since it was dependent on another webhook that wasn't functioning properly.

PR #9244, opened 2022-06-03
add nil guard

happythenewsad:
## WHAT
Add a nil guard so that a serializer in Standards Report does not raise an exception.

## WHY
So the Standards Report page can load correctly when `average_score` is nil. 

### Screenshots

### Notion Card Links
https://www.notion.so/quill/Product-Board-30f97e4eb01246dbb8264253fb385073?p=d0394e4bc7784d559757ce366f4e8594

PR Checklist | Your Answer
------------ | -------------

### DIFF
--- a/services/QuillLMS/app/serializers/progress_reports/standards/student_serializer.rb
+++ b/services/QuillLMS/app/serializers/progress_reports/standards/student_serializer.rb
@@ -18,7 +18,7 @@ class ProgressReports::Standards::StudentSerializer < ActiveModel::Serializer
              :mastery_status
 
   def average_score
-    object.average_score.round(2)
+    object.average_score&.round(2) || 0
   end
 
   def student_standards_href
--- a/services/QuillLMS/spec/serializers/progress_reports/standards/student_serializer_spec.rb
+++ b/services/QuillLMS/spec/serializers/progress_reports/standards/student_serializer_spec.rb
@@ -61,11 +61,8 @@ describe ProgressReports::Standards::StudentSerializer, type: :serializer do
     end
 
     it 'should not raise error when average_score is nil' do
-      mock_object = double.tap { |mock| allow(mock).to receive(:average_score).and_return(nil) }
-      serializer = described_class.new(mock_object)
-      expect do
-        serializer.average_score
-      end.to_not raise_error
+      serializer = described_class.new(double(average_score: nil))
+      expect { serializer.average_score }.to_not raise_error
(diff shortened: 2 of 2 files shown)


services/QuillLMS/spec/serializers/progress_reports/standards/student_serializer_spec.rb
```diff
@@ -58,5 +59,15 @@
     it 'includes properly rounded scores' do
       expect(parsed_student['average_score']).to eq(0.75)
     end
+
+    it 'should not raise error when average_score is nil' do
+      mock_object = double.tap { |mock| allow(mock).to receive(:average_score).and_return(nil) }
+      serializer = described_class.new(mock_object)
+      expect do
```
brendanshean: A couple of stylistic suggestions:
1) Since we're not using `mock_object` elsewhere, you could just pass the double in directly `serializer = described_class.new(double(average_score: nil))` and remove the mock_object assignment altogether.  The `tap` construction here seems verbose, but I might be missing something.

2) Make the expectation a one-liner since it's a one-line block:  `expect { serializer.average_score }.not_to raise_error`

services/QuillLMS/app/serializers/progress_reports/standards/student_serializer.rb
```diff
@@ -18,7 +18,7 @@ class ProgressReports::Standards::StudentSerializer < ActiveModel::Serializer
              :mastery_status
 
   def average_score
-    object.average_score.round(2)
+    object.average_score&.round(2) || 0
```
anathomical, line 21: Are we sure we want to default to 0 for a "score" value?  There's definitely a difference between "you have no score" and "you scored 0%" in some contexts.  Is this one of them?

anathomical (approved):
Nil guard seems reasonable.  Just want to double-check on whether we want our fallback value to be `0`, or if it should just return `nil` if there is no score.

services/QuillLMS/app/serializers/progress_reports/standards/student_serializer.rb
happythenewsad: Good question - I'm currently working on a related ticket that will change display logic: https://www.notion.so/quill/Product-Board-30f97e4eb01246dbb8264253fb385073?p=2d5f2d4bf8874dfd8d854fdb5ef546b7

brendanshean (approved):
Looks good.  Thanks for catching this bug.

PR #9247, opened 2022-06-07
Prevent duplicate subscription charges

brendanshean:
## WHAT
Clean up some edge cases involving users purchasing duplicate subscriptions.

## WHY
The presence of duplicate subscriptions causes a headache for support as well as our backend.

## HOW
In the original design, `Stripe::Checkout::Session` objects were [created](https://github.com/empirical-org/Empirical-Core/blob/develop/services/QuillLMS/app/controllers/stripe_integration/subscription_checkout_sessions_controller.rb#L8) with each request to the controller.  However, this leads to an edge cases where a user opens multiple tabs and attempts to purchase two subscriptions at separate times.  

To fix this, I'm introducing a new database model, `StripeCheckoutSession` that will store a few pieces of information that will point each user to a sole url for purchase of a particular subscription.
1. `external_checkout_session_id` is the id used for the checkout session on stripe.  This value is used with the webhooks `checkout.sesssion.completed` and `checkout.session.expired`.
2. `stripe_price_id` identifies the type of subscription being purchased
3. `url` is the redirect_url that stripe produces that will direct users to the purchasing page.
4.  `user_id`

With this change, if a user clicks 'Buy Now / Renew Subscription' for the same product in different tabs, they will redirected to the the same URL which refers to the same `StripeCheckoutSession` (i.e (NB Stripe will block duplicate purchases from the same session).

Upon completion of purchase (i.e. webhook 'checkout.session.completed'), or expiration of the checkout session (i.e. webhook 'checkout.session.expired'), the `StripeCheckoutSession` will be deleted.

### Screenshots

### Notion Card Links
https://www.notion.so/quill/Handle-duplicate-charges-336f63d940da4988aeb1d761b9755302

PR Checklist | Your Answer
------------ | -------------

### DIFF
--- a/services/QuillLMS/app/models/stripe_checkout_session.rb
+++ b/services/QuillLMS/app/models/stripe_checkout_session.rb
@@ -1,5 +1,27 @@
 # frozen_string_literal: true
 
+# == Schema Information
+#
+# Table name: stripe_checkout_sessions
+#
+#  id                           :bigint           not null, primary key
+#  expiration                   :datetime         not null
+#  url                          :string           not null
+#  created_at                   :datetime         not null
+#  updated_at                   :datetime         not null
+#  external_checkout_session_id :string           not null
+#  stripe_price_id              :string           not null
+#  user_id                      :bigint
+#
+# Indexes
+#
+#  index_stripe_checkout_sessions_on_external_checkout_session_id  (external_checkout_session_id)
+#  index_stripe_checkout_sessions_on_user_id                       (user_id)
+#
+# Foreign Keys
+#
+#  fk_rails_...  (user_id => users.id)
+#
 class StripeCheckoutSession < ApplicationRecord
   belongs_to :user
 
--- a/services/QuillLMS/app/models/stripe_checkout_session.rb
+++ b/services/QuillLMS/app/models/stripe_checkout_session.rb
@@ -5,7 +5,6 @@
 # Table name: stripe_checkout_sessions
 #
 #  id                           :bigint           not null, primary key
-#  expiration                   :datetime         not null
 #  url                          :string           not null
 #  created_at                   :datetime         not null
 #  updated_at                   :datetime         not null
@@ -25,16 +24,13 @@
 class StripeCheckoutSession < ApplicationRecord
   belongs_to :user
 
-  scope :not_expired, -> { where('expiration > ?', DateTime.now.utc) }
-
   def self.custom_find_or_create_by!(external_checkout_session_args, stripe_price_id, user_id)
-    stripe_checkout_session = StripeCheckoutSession.not_expired.find_by(stripe_price_id: stripe_price_id, user_id: user_id)
+    stripe_checkout_session = StripeCheckoutSession.find_by(stripe_price_id: stripe_price_id, user_id: user_id)
     return stripe_checkout_session if stripe_checkout_session.present?
 
     external_checkout_session = Stripe::Checkout::Session.create(external_checkout_session_args)
 
     StripeCheckoutSession.create!(
-      expiration: Time.at(external_checkout_session.expires_at).utc.to_datetime,
       external_checkout_session_id: external_checkout_session.id,
       stripe_price_id: stripe_price_id,
       url: external_checkout_session.url,
--- a/services/QuillLMS/app/models/stripe_checkout_session.rb
+++ b/services/QuillLMS/app/models/stripe_checkout_session.rb
@@ -1,0 +1,22 @@
+# frozen_string_literal: true
+
+class StripeCheckoutSession < ApplicationRecord
+  belongs_to :user
+
+  scope :not_expired, -> { where('expiration > ?', DateTime.now.utc) }
+
+  def self.custom_find_or_create_by!(external_checkout_session_args, stripe_price_id, user_id)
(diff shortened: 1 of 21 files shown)


services/QuillLMS/app/models/stripe_checkout_session.rb
```diff
@@ -0,0 +1,22 @@
+# frozen_string_literal: true
+
```
anathomical, line 2: Should we have the auto-generated annotation code here from running `rake db:migrate`?

services/QuillLMS/app/services/stripe_integration/webhooks/checkout_session_completed_event_handler.rb
```diff
@@ -0,0 +1,22 @@
+# frozen_string_literal: true
+
+module StripeIntegration
+  module Webhooks
+    class CheckoutSessionCompletedEventHandler < EventHandler
+      def run
+        stripe_checkout_session&.update!(expiration: DateTime.now.utc)
```
anathomical: Should we consider just deleting this record instead of keeping it around with an expiration assigned?  Do we actually want this for record-keeping purposes?  Or am I wrong in thinking that deleting the record would give us the same result in terms of duplicate-protection?

services/QuillLMS/app/services/stripe_integration/webhooks/checkout_session_expired_event_handler.rb
```diff
@@ -0,0 +1,22 @@
+# frozen_string_literal: true
+
+module StripeIntegration
+  module Webhooks
+    class CheckoutSessionExpiredEventHandler < EventHandler
+      def run
+        stripe_checkout_session&.update!(expiration: DateTime.now.utc)
```
anathomical: Same note as my previous about deleting this record instead of updating expiration.

anathomical (approved):
This looks fine to me.  One question: should we keep the expired items in our DB, or does it make sense to just delete them?  Do we get ongoing value from retaining the records given that we weren't even writing them before?

services/QuillLMS/app/services/stripe_integration/webhooks/checkout_session_completed_event_handler.rb
brendanshean: @anathomical Good question.  I wrestled with this decision myself.  My first pass at this, I just deleted the record like you suggested `stripe_checkout_session`.  However, when troubleshooting a specific edge case (and potentially future ones), I found it more useful to keep the object around.  

That being said, I do agree though that these probably add little value long term.  Maybe a cron job  to clean these out maybe weekly/monthly?

brendanshean: On second thought.  Maybe I should just trust that the code works and just revert back to deleting.  That would be much simpler and then we can get rid of expiration.

brendanshean: @anathomical I've moved it back to just `destroy!`.  Thank you for the nudge

services/QuillLMS/app/services/stripe_integration/webhooks/checkout_session_expired_event_handler.rb
```diff
@@ -0,0 +1,22 @@
+# frozen_string_literal: true
+
+module StripeIntegration
+  module Webhooks
+    class CheckoutSessionExpiredEventHandler < EventHandler
```
dandrabik, line 5: This is identical to the handler above, is there a reason to not reuse it?

services/QuillLMS/app/services/stripe_integration/webhooks/checkout_session_completed_event_handler.rb
```diff
@@ -0,0 +1,22 @@
+# frozen_string_literal: true
+
+module StripeIntegration
+  module Webhooks
+    class CheckoutSessionCompletedEventHandler < EventHandler
+      def run
+        stripe_checkout_session&.destroy!
```
dandrabik, line 7: Is there any value in saving this record and doing a safe delete (e.g. setting a `finished_at`/`expired_at` or a `processed/expired` state)? For example, would it be helpful for debugging checkout issues or giving a full timeline/understanding of the checkout process? It may not be that helpful, so we can delete, but it feels worth asking that question.

dandrabik (approved):
Overall this looks good. I had some comments that are worth looking at around some duplicate code and whether we might want to safe delete the checkout record (since it might be useful in the future)

services/QuillLMS/app/services/stripe_integration/webhooks/checkout_session_completed_event_handler.rb
brendanshean: I've kind of been going back and forth on [this](https://github.com/empirical-org/Empirical-Core/pull/9247#discussion_r892589050), but I think deleting them is the most practical approach.  We can recover some of the information via `checkout.session.expired`, but if we end up needing more, I might add the `expiration` back in.

services/QuillLMS/app/services/stripe_integration/webhooks/checkout_session_expired_event_handler.rb
brendanshean: I thought about it, but since it's only two examples, I'll leave as-is for now.

PR #9252, opened 2022-06-08
update premium page to allow direct purchases of school premium

emilia-friedberg:
## WHAT
Update the premium page to allow direct purchases of school premium via a modal. Once the modal is opened, there are a couple of branching user flows which work as follows:

1. If the user is not signed in, they get routed to the sign-in page, and then back to the appropriate stage of step 2.
2.
a. The user is only associated with one school (associated here meaning either via the schools_users table or the schools_admins table) and they get sent directly to Stripe to complete the purchase.
b. The user is not associated with any schools (or only ones that are placeholder schools, like "not listed" or "homeschool") and they see the school selection component we use in the sign-up and account settings flows. Once they select a school, they are sent to Stripe. If they select "skip for now", the modal closes and they see a snackbar that says "Sorry, you must select a school to purchase school premium."
c. The user is associated with more than one school, and they see a dropdown prompting them to pick a school to purchase premium for.

## WHY
So that new school premium accounts can be purchased via credit card.

## HOW
Just React changes mostly, with some minor backend edits to pass in the data we need.

### Screenshots
Screen Shot 2022-06-08 at 12 03 24 PM
Screen Shot 2022-06-08 at 12 03 51 PM
Screen Shot 2022-06-08 at 12 04 47 PM
Screen Shot 2022-06-08 at 12 09 40 PM
Screen Shot 2022-06-08 at 12 11 58 PM
Screen Shot 2022-06-08 at 12 12 25 PM
Screen Shot 2022-06-08 at 12 12 37 PM
Screen Shot 2022-06-08 at 12 13 18 PM
Screen Shot 2022-06-08 at 12 13 31 PM

### Notion Card Links
https://www.notion.so/quill/User-flow-for-School-Premium-purchase-cd15f9ccabae4e75848235f7cd2988a6

PR Checklist | Your Answer
------------ | -------------


services/QuillLMS/client/app/bundles/Teacher/components/premium/school_and_district_premium_modal.tsx
```diff
@@ -0,0 +1,153 @@
...
+    const path = '/stripe_integration/subscription_checkout_sessions'
+    const data = {
+      cancel_path: 'premium',
+      customer_email: customerEmail,
+      ...(schoolIds && { school_ids: JSON.stringify(schoolIds) }),
+      stripe_price_id: stripeSchoolPlan.plan.stripe_price_id
+    }
+
+    requestPost(path, data, body => { window.location.replace(body.redirect_url) })
+  }
```
emilia-friedberg, line 82: It wasn't possible to reuse the existing button component here, because we need to be able to activate this logic indirectly (like when a user with one school is signed out, pressing "Buy Now" needs to redirect them to sign in and then send them directly to Stripe, or when a user with no schools selects one from the school selection component there isn't an additional button to press).

services/QuillLMS/app/controllers/pages_controller.rb
```diff
@@ -409,10 +408,9 @@ def premium
     @user_has_school = !!current_user&.school && ['home school', 'us higher ed', 'international', 'other', 'not listed'].exclude?(current_user&.school&.name)
     @user_belongs_to_school_that_has_paid = current_user&.school ? Subscription.school_or_user_has_ever_paid?(current_user&.school) : false
     @customer_email = current_user&.email
-    @school_ids = [current_user&.school&.id].to_json
+    @associated_schools = current_user ? [current_user.school].concat(current_user.administered_schools).uniq.select { |s| School::ALTERNATIVE_SCHOOL_NAMES.exclude?(s.name) } : []
```
brendanshean: The first branch feels like a reusable method.  What about extracting it into the user model:

```
def associated_schools
  [school].concat(administered_schools).uniq.select { |s| School::ALTERNATIVE_SCHOOL_NAMES.exclude?(s.name) }
end
```

Leaving
` @associated_schools = current_user ? current_user.associated_schools : [] `

Not sure on the flow with UI, but if there's a possibility `current_user.school` is nil, you'll probably need a `compact`:  `[school].concat(administered_schools).uniq.compact`

services/QuillLMS/app/views/pages/shared/_premium_main.html.erb
```diff
@@ -13,12 +13,11 @@ document.addEventListener('click', function(e) {
 
 <div id="first-preview-section">
   <%= react_component('PremiumPricingGuideApp', props: {
+      associatedSchools: @associated_schools,
       customerEmail: @customer_email,
       diagnosticActivityCount: @diagnostic_activity_count,
       independentPracticeActivityCount: @independent_practice_activity_count,
       lessonsActivityCount: @lessons_activity_count,
-      schoolIds: @school_ids,
-      showSchoolBuyNow: @show_school_buy_now,
```
brendanshean, line 21: Thanks for cleaning up the SchoolBuyNow stuff.

services/QuillLMS/client/app/bundles/Teacher/components/premium/school_and_district_premium_modal.tsx
brendanshean: That makes sense.  Looking at the StripeSubscriptionCheckoutSessionButton,  if we were to fix it, maybe refactoring the [alerts](https://github.com/empirical-org/Empirical-Core/blob/develop/services/QuillLMS/client/app/bundles/Teacher/components/shared/StripeSubscriptionCheckoutSessionButton.jsx#L17) out of the component as a variable passed into the component, but that might still be messy.

services/QuillLMS/client/app/bundles/Teacher/components/premium/school_and_district_premium_modal.tsx
```diff
@@ -0,0 +1,153 @@
...
+import { closeIcon, DropdownInput, Spinner, } from '../../../Shared/index'
+
+export const SCHOOL_PURCHASE_STAGE_TWO = 'school_purchase_stage_two'
+
+const SchoolAndDistrictPremiumModal = ({ stripeSchoolPlan, associatedSchools, userIsSignedIn, startAtStageTwo, customerEmail, closeModal, handleNotListedSelection, }) => {
+  const [stage, setStage] = React.useState(startAtStageTwo ? 2 : 1)
+  const [selectedSchool, setSelectedSchool] = React.useState(null)
+  const [loading, setLoading] = React.useState(false)
+
+  React.useEffect(handleStageTwo, [])
```
brendanshean: Is this needed?

services/QuillLMS/client/app/bundles/Teacher/components/premium/school_and_district_premium_modal.tsx
```diff
@@ -0,0 +1,153 @@
...
+      ...(schoolIds && { school_ids: JSON.stringify(schoolIds) }),
+      stripe_price_id: stripeSchoolPlan.plan.stripe_price_id
+    }
+
+    requestPost(path, data, body => { window.location.replace(body.redirect_url) })
+  }
+
+  function goToStripeWithSelectedSchool() { goToStripe([selectedSchool.value]) }
+
+  function selectSchool(idOrType) {
```
brendanshean, line 86: Unless I'm missing something, I feel like the variable `idOrType` should just be `schoolId` since we are never passing in a schoolType to `goToStripe`.

services/QuillLMS/client/app/bundles/Teacher/components/premium/school_and_district_premium_modal.tsx
```diff
@@ -0,0 +1,153 @@
+import * as React from 'react';
+
+import SchoolSelector, { NOT_LISTED, } from '../shared/school_selector'
+import { requestPost, requestPut, } from '../../../../modules/request';
+import { closeIcon, DropdownInput, Spinner, } from '../../../Shared/index'
+
+export const SCHOOL_PURCHASE_STAGE_TWO = 'school_purchase_stage_two'
```
brendanshean: I think renaming usages of the concepts stage_one and stage_two with something more descriptive (e.g. 
stage_1 =>  plan_selection_stage, stage_2 => school_selection_stage) would greatly improve readability of the code as well as facilitating refactoring later.

While you and I know what the stages are, another developer (or even myself in a couple months) looking at the code might find the labels stage_1 and stage_2 opaque. 

You can also replace the magic numbers 2 and 1 here
`const [stage, setStage] = React.useState(startAtStageTwo ? 2 : 1)` 

with constants like PLAN_SELECTION_STAGE = 1 and SCHOOL_SELECTION_STAGE = 2.

An example of why this helps with refactoring later is as follows.  Let's say we wanted to add another stage in between stage_1 and stage_2.  We'd first have to go through the error-prone strategy of renaming all stage_2 to stage_3 before adding in the new stage_2.  If we have plan_selection_stage and school_selection_stage, this is no longer a worry.

brendanshean (approved):
A herculean effort on this piece of the subscription!  I do have a stylistic comment concerning the StageOne and StageTwo language, but I think once that's addressed, this is good to go, no need for re-review.

services/QuillLMS/client/app/bundles/Teacher/components/premium/school_and_district_premium_modal.tsx
emilia-friedberg: It is true that the schoolType never makes it to `goToStripe`, but `not listed` can come through as the type here.

services/QuillLMS/client/app/bundles/Teacher/components/premium/school_and_district_premium_modal.tsx
emilia-friedberg: Yes - this handles when we are stage two as of initial load (because the user was sent to login and then redirected back here with the query param). The following effect will only run when the `stage` has been changed. Unfortunately there isn't a more succinct way to say "run this hook both at initial load and when the stage changes".

services/QuillLMS/client/app/bundles/Teacher/components/premium/school_and_district_premium_modal.tsx
brendanshean: Good point

services/QuillLMS/client/app/bundles/Teacher/components/premium/premium_pricing_minis_row.jsx
```diff
@@ -19,7 +21,7 @@ export default class PremiumPricingMinisRow extends React.Component {
 
   listenScrollEvent = e => {
     const { isScrolled, } = this.state
-    if (window.scrollY > 337 && !isScrolled) {
+    if (window.scrollY > 337 && !isScrolled && window.innerWidth > MOBILE_WIDTH) {
```
eadams17: I think we'd also want this 337 value to be a const as well?

services/QuillLMS/client/app/bundles/Teacher/components/premium/school_and_district_premium_modal.tsx
```diff
@@ -0,0 +1,157 @@
...
+        <div className="modal-background" />
+        <div className="school-and-district-premium-modal stage-two quill-modal modal-body">
+          <Spinner />
+        </div>
+      </div>
+    )
+  }
+
+  if (stage === SCHOOL_SELECTION_STAGE_NUMBER) {
+    if (eligibleSchools.length) {
```
eadams17: we could just combine these into a single if statement

eadams17 (approved):
Thanks for knocking all of this out! Looks great. I just left two small stylistic comments.

services/QuillLMS/client/app/bundles/Teacher/components/premium/school_and_district_premium_modal.tsx
emilia-friedberg: We can't, because the `else` for the second `if` should only happen if the first conditional is true (ie, if the stage is SCHOOL_SELECTION_STAGE_NUMBER but there are zero eligible schools). Your comment made me realize that these nested conditionals are confusing, so I moved out the inner conditional block into a separate function.

PR #9258, opened 2022-06-09
Prevent duplicate stripe school subscriptions

brendanshean:
## WHAT
Add checks in subscription creation to prevent duplicate stripe school subscriptions.

## WHY
We would like to have the same duplicate prevention with school subscriptions that we have with teacher subscriptions.

## HOW
Add an array of integers attribute `school_ids` to the `StripeCheckoutSession` object.  This will ensure that the session is unique to the school[s] a user is attempting to purchase subscriptions for.

Add logic in the duplicate_subscription check with `SubscriptionCreator` that evaluates duplicate subscriptions based on the type of plan.

### Screenshots

### Notion Card Links

PR Checklist | Your Answer
------------ | -------------

### DIFF
--- a/services/QuillLMS/app/services/stripe_integration/webhooks/subscription_creator.rb
+++ b/services/QuillLMS/app/services/stripe_integration/webhooks/subscription_creator.rb
@@ -36,7 +36,11 @@ module StripeIntegration
         when Plan.stripe_teacher_plan
           purchaser.subscription&.plan == plan
         when Plan.stripe_school_plan
-          purchaser.subscriptions.active.any? { |subscription| subscription.schools.pluck(:id).sort == school_ids }
+          purchaser.associated_schools.any? do |school|
+            school.subscriptions.active.any?  do |subscription|
+              subscription.schools.pluck(:id).sort == school_ids
+            end
+          end
         end
       end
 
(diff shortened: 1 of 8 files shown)


services/QuillLMS/app/services/stripe_integration/webhooks/subscription_creator.rb
```diff
@@ -24,13 +24,22 @@ def run
...
         save_stripe_customer_id
         run_plan_custom_tasks
       end
 
+      private def duplicate_subscription?
+        case plan
+        when Plan.stripe_teacher_plan
+          purchaser.subscription&.plan == plan
+        when Plan.stripe_school_plan
+          purchaser.subscriptions.active.any? { |subscription| subscription.schools.pluck(:id).sort == school_ids }
```
emilia-friedberg: If a user is an admin for a school, but does not belong to that school, I don't think that school's subscription will come up under `purchaser.subscriptions`, will it?

brendanshean: Good point.  I will fix that.

brendanshean: @emilia-friedberg I'm going to refactor and add tests tomorrow but I'm thinking something like this?

```
        when Plan.stripe_school_plan
          purchaser.subscriptions.active.any? { |subscription| subscription.schools.pluck(:id).sort == school_ids } ||
            purchaser.associated_schools.any? do |school| 
              school.subscriptions.active.any?  { |subscription| subscription.schools.pluck(:id).sort == school_ids }
            end
        end
```

emilia-friedberg: That makes sense to me. I'm not sure you need the first part of the `or` - I think the second case will always cover the first (because if a user can only have a school subscription if they are associated with a school with a subscription).

emilia-friedberg (approved):

PR #9260, opened 2022-06-13
Create and manage district subscriptions

brendanshean:
## WHAT
Allow staff to create and edit district-level subscriptions

Addendum:
Instead of automatically creating SchoolSubscriptions for all schools in a given district, let staff choose which schools for a given district should receive SchoolSubscriptions.

## WHY
This will helps integration with Streak and Vitally.  See Notion card for more discussion.

## HOW
1.  Create a new model `DistrictSubscriptions`.  
2. Add a new concern `Subscriber` meant be a highest most object (i.e. User, School, or District) that the subscription applies to.  This concern is meant to replace logic that uses `school_or_user` references.  On the frontend, pass `subscriberType` to distinguish what kind of subcriber `class` that we are dealing with.
3.  Within the District Manager, add new buttons 'Create Subscription' and when applicable, 'Edit Subscription. 
4. The UI flows to district subscription creation and editing follows closely with SchoolSubscription.
5. Within the UI, staff selects / unselects which schools should receive SchoolSubscriptions.
6.  The listing and checkboxes for the UI is populated via `Cms::DistrictSchoolsAndSubscriptionStatus` service object. 
7.  Upon creating/updating the subscription, the `Cms::SchoolSubscriptionsUpdater` is used to determine which schools should have new SchoolSubscription records created as well as existing SchoolSubscription records delete.

### Screenshots

### Notion Card Links
https://www.notion.so/quill/Make-it-possible-to-create-and-manage-subscriptions-at-the-district-level-71fd1efe79a6488a9de0d2863b1b4cc7

PR Checklist | Your Answer
------------ | -------------
Screen Shot 2022-06-14 at 12 24 45 PM
Screen Shot 2022-06-17 at 12 24 36 PM


services/QuillLMS/app/models/concerns/subscriber.rb
```diff
@@ -0,0 +1,39 @@
+# frozen_string_literal: true
+
+module Subscriber
```
emilia-friedberg, line 3: This was a really good call. Much cleaner and more legible.

services/QuillLMS/app/models/concerns/teacher.rb
```diff
@@ -408,23 +408,21 @@ def update_teacher params
   end
   # rubocop:enable Metrics/CyclomaticComplexity
 
-  # rubocop:disable Metrics/CyclomaticComplexity
   def updated_school(school_id)
     school = School.find_by(id: school_id)
-    if subscription && subscription.school_subscriptions.any? && !has_matching_subscription?(id, school&.subscription&.id)
+    if subscription && subscription.school_subscriptions.any? && !has_matching_subscription?(self, school&.subscription)
```
emilia-friedberg: I think this could be a little shorter if it were `if subscription&.school_subscriptions&.any? && ...`

services/QuillLMS/app/models/school.rb
```diff
@@ -90,26 +92,16 @@ class School < ApplicationRecord
...
-      .limit(1)
-      .first
+  def join_subscription(subscription)
+    school_subscriptions.create(subscription: subscription)
   end
 
-  def present_and_future_subscriptions
-    subscriptions.active
+  def not_alternative?
+    !ALTERNATIVE_SCHOOL_NAMES.include?(name)
```
emilia-friedberg: There is an `.exclude?` method you can use instead of negating the `.include?`

emilia-friedberg (approved):
LGTM. The new subscriber concern and all the associated cleanup is really great. Thank you for doing all that.

services/QuillLMS/app/models/bonus_days_calculator.rb
```diff
@@ -15,17 +15,13 @@ def initialize(user, start: Date.current)
   end
 
   def run
-    return 0 if user.nil? || school.nil? || school_has_paid_before?
+    return 0 if user.nil? || school.nil? || school.ever_paid_for_subscription?
```
dandrabik, line 18: I like that you moved this method to the school model. :+1:

services/QuillLMS/app/models/concerns/subscriber.rb
dandrabik: Agreed

services/QuillLMS/app/models/concerns/subscriber.rb
```diff
@@ -0,0 +1,39 @@
...
+    is_a?(School)|| is_a?(District)
+  end
+
+  def subscription
+    subscriptions
+      .started
+      .not_expired
+      .not_de_activated
+      .order(expiration: :desc)
+      .limit(1)
```
dandrabik: `limit(1)` is redundant if you call .`first` immediately after.
```suggestion
```

services/QuillLMS/app/models/district.rb
```diff
@@ -24,21 +24,28 @@
...
+  has_many :subscriptions, through: :district_subscriptions
 
   scope :by_name, ->(name) { where('lower(name) LIKE ?', "%#{name.downcase}%") }
   scope :by_city, ->(city) { where('lower(city) LIKE ?', "%#{city.downcase}%") }
   scope :by_state, ->(state) { where(:state => state.upcase) }
   scope :by_zipcode, ->(zipcode) { where(:zipcode => zipcode) }
   scope :by_nces_id, ->(nces_id) { where(:nces_id => nces_id) }
 
+
+  def join_subscription(subscription)
```
dandrabik: Consider using a different word than `join` here, since my quick read thought this was a SQL finder helper, e.g. `.joins(:subscription)`. I was surprised to see a `create`

services/QuillLMS/app/models/school.rb
```diff
@@ -90,26 +92,16 @@ class School < ApplicationRecord
...
-      .order(expiration: :desc)
-      .limit(1)
-      .first
+  def join_subscription(subscription)
+    school_subscriptions.create(subscription: subscription)
   end
 
-  def present_and_future_subscriptions
-    subscriptions.active
+  def not_alternative?
```
dandrabik: You might consider making this a `alternative?` and using `!school.alternative?` 👨‍🎤

services/QuillLMS/app/controllers/cms/schools_controller.rb
```diff
@@ -354,4 +357,13 @@ def unlink
   def fallback_location
     cms_school_path(params[:id].to_i)
   end
+
+  private def subscription_data
+    @premium_types = Subscription::OFFICIAL_SCHOOL_TYPES
+    @subscription_payment_methods = Subscription::CMS_PAYMENT_METHODS
+
+    return unless @school&.not_alternative?
```
dandrabik: Double negative. Consider equivalent:
```ruby
 return if @school.nil? || @school.alternative?

File: services/QuillLMS/app/controllers/cms/subscriptions_controller.rb [hunk 9]
```diff
@@ -33,22 +33,32 @@ def destroy
     @subscription = Subscription.find(params[:id])
   end
 
+  private def subscription_data
+    @district = @subscription.districts&.first
+    @school = @subscription.schools&.first
+    @premium_types = @subscription.premium_types
+    @subscription_payment_methods = Subscription::CMS_PAYMENT_METHODS
+
+    return unless @school&.not_alternative?
```
dandrabik: double negative.

services/QuillLMS/app/models/concerns/subscriber.rb
```diff
@@ -0,0 +1,39 @@
...
+
+  def ever_paid_for_subscription?
+    (Subscription::OFFICIAL_PAID_TYPES & subscriptions.pluck(:account_type)).present?
+  end
+
+  def last_expired_subscription
+    subscriptions
+      .expired
+      .order(expiration: :desc)
+      .limit(1)
```
dandrabik: `limit(1)` is redundant if you call `.first` immediately after.
```suggestion

```

services/QuillLMS/app/models/district_subscription.rb
```diff
@@ -0,0 +1,32 @@
...
+#  fk_rails_...  (subscription_id => subscriptions.id)
+#
+class DistrictSubscription < ApplicationRecord
+  belongs_to :district, optional: false
+  belongs_to :subscription, optional: false
+
+  after_create :create_school_subscriptions
+
+  def create_school_subscriptions
+    district.schools.each { |school| school.school_subscriptions.create!(subscription: subscription) }
```
dandrabik: Flagging a raisable error in a loop. 

Should this be in a transaction so they are either all created or none are created?  How this is currently coded, if the 3rd of 10 schools errors, the first 2 subscriptions would be created and the last 8 would not.

services/QuillLMS/app/models/subscription.rb
```diff
@@ -125,6 +142,42 @@ class RenewalNilStripeCustomer < StandardError; end
...
+        subscription_attrs = subscription_attrs.merge(Subscription.set_trial_expiration_and_start_date(subscriber))
+      when CB_LIFETIME_SUBSCRIPTION_TYPE
+        subscription_attrs = subscription_attrs.merge(Subscription.set_cb_lifetime_expiration_and_start_date)
+      else
+        subscription_attrs = subscription_attrs.merge(Subscription.set_premium_expiration_and_start_date(subscriber))
+      end
+    end
+
+    subscriber.subscriptions.each do |existing_subscription|
+      existing_subscription.update!(recurring: false)
```
dandrabik: Possible raise in a loop. Is a transaction needed here?

services/QuillLMS/app/models/subscription.rb
```diff
@@ -125,6 +142,42 @@ class RenewalNilStripeCustomer < StandardError; end
...
+      existing_subscription.stripe_cancel_at_period_end
+    end
+
+    subscription = Subscription.create!(subscription_attrs)
```
dandrabik: There are multiple places that this method can raise while partially completed. Maybe a transaction is needed for this whole method?

services/QuillLMS/app/models/subscription.rb
```diff
@@ -125,6 +142,42 @@ class RenewalNilStripeCustomer < StandardError; end
...
+    subscription
+  end
+
+  def audience
+    return DISTRICT if district_subscriptions.exists?
+    return SCHOOL if school_subscriptions.exists?
+    return TEACHER if user_subscriptions.exists?
+  end
+
+  def premium_types
```
dandrabik, line 156: Just noting that this doesn't have a default, so could return `nil`. That might be unexpected downstream. Might be better to default to `[]` at least.

services/QuillLMS/app/models/subscription.rb
```diff
@@ -125,6 +142,42 @@ class RenewalNilStripeCustomer < StandardError; end
...
+      existing_subscription.update!(recurring: false)
+      existing_subscription.stripe_cancel_at_period_end
+    end
+
+    subscription = Subscription.create!(subscription_attrs)
+    subscriber.join_subscription(subscription)
+    subscription
+  end
+
+  def audience
```
dandrabik: I think `audience` is only used for the method below. If that's the case, consider combining for simplicity, e.g.
```ruby
def premium_types
  return OFFICIAL_DISTRICT_TYPES if district_subscriptions.exists?
  return OFFICIAL_SCHOOL_TYPES if school_subscriptions.exists?
  ...

**Review by dandrabik — CHANGES_REQUESTED** [review]
Overall this looks really good! 

I'm requesting changes since there are two places in which errors could be raised within a loop (so associations could be left in a partially created state, some created, some not). I think those sections should be revisited and we should consider using transactions there, i.e. create everything if all valid, otherwise create nothing. Happy to talk through those decisions, since they can be tricky.

File: services/QuillLMS/app/models/district_subscription.rb (hunk 11 above)
Comment (brendanshean) [reply services/QuillLMS/app/models/district_subscription.rb]: Great point.  I will update.

File: services/QuillLMS/app/models/district.rb (hunk 6 above)
Comment (brendanshean) [reply services/QuillLMS/app/models/district.rb]: Good point.  Renaming `attach_subscription`

File: services/QuillLMS/app/controllers/cms/schools_controller.rb (hunk 8 above)
Comment (brendanshean) [reply services/QuillLMS/app/controllers/cms/schools_controller.rb]: That's much cleaner.

**Comment by brendanshean** [comment]
@emilia-friedberg Re-requesting review since I've added in a the UI element for selecting individual Schools that will receive SchoolSubscriptions.

**Review by dandrabik — APPROVED** [review]
:+1: Looks good, thanks for making these changes!

File: services/QuillLMS/app/assets/stylesheets/cms/subscription.scss [hunk 16]
```diff
@@ -9,6 +9,11 @@
   button {
     margin-top: 25px;
   }
+  .data-table {
+    button {
+      margin-top: 0px;
+    }
+  }
```
emilia-friedberg, line 16: Yay!

services/QuillLMS/client/app/bundles/Teacher/containers/EditOrCreateSubscription.jsx
```diff
@@ -21,33 +22,11 @@ export default class EditOrCreateSubscription extends React.Component {
     this.state = {
       subscription: defaultSubscription,
       firstFocused: false,
-      secondFocused: false
+      secondFocused: false,
+      schools: this.props.schools,
```
emilia-friedberg: Not important but ideally this would be deconstructed (you'd add `schools` to line 14 above and then just do:

```
  secondFocused: false,
  schools
}
```

here.

services/QuillLMS/client/app/bundles/Teacher/containers/EditOrCreateSubscription.jsx
```diff
@@ -72,6 +51,70 @@ export default class EditOrCreateSubscription extends React.Component {
...
+
+    if (!created_at) { return }
+
+    return (
+      <React.Fragment>
+        <label>Created At:</label>
+        <span>{created_at}</span>
+        <br />
+      </React.Fragment>
+    )
```
emilia-friedberg, line 66: Great use of fragment.

emilia-friedberg (approved):

services/QuillLMS/app/assets/stylesheets/cms/subscription.scss
brendanshean: Thanks again for the help!

PR #9274, opened 2022-06-15
Handle refunded invoice events

brendanshean:
## WHAT
Fix an edge case involving invoice.paid events that are actually refunds.

There are also a couple of commits that remove dead code:
Remove dead code from school subscriptions
PurchaseModal and SchoolPremiumModal dead code

## WHY
When an `invoice.paid` webhook events is received, our system creates a new subscription.  We shouldn't do this for invoices that are reporting a refund.

## HOW
Check if the amount_paid on the stripe invoice is $0.  This is a refund and we should skip SubscriptionCreation in this case.
I've also added a check that compares actual amount paid with plan price.

### Screenshots

### Notion Card Links

PR Checklist | Your Answer
------------ | -------------


services/QuillLMS/app/services/stripe_integration/webhooks/invoice_paid_event_handler.rb
```diff
@@ -21,6 +21,10 @@ def run
         stripe_invoice.subscription.nil?
       end
 
+      private def refund_invoice?
+        stripe_invoice.amount_paid.zero?
```
anathomical: Is it possible for a purchase to have a `0` amount paid if they had some discounts or something?  Just wondering if there's a less ambiguous way to identify refunds on this hook.

anathomical (approved):
One question about how we're determining if it's a refund, but it's likely that we're just at the mercy of Stripe data models there, so there may be nothing we can do that's better.  I also really appreciate the removal of dead routes.  We have so many routes.

services/QuillLMS/app/services/stripe_integration/webhooks/invoice_paid_event_handler.rb
brendanshean: That's a good question.  I really can't say since this has only happened once and after giving instructions to Support about how to cancel a subscription with the new Stripe integration, this really shouldn't happen at all.  I will change this to an exception for now and then that will give me more information on how to handle this in the future.

emilia-friedberg (approved):
Thanks for all the cleanup work!

services/QuillLMS/app/services/stripe_integration/webhooks/invoice_paid_event_handler.rb
```diff
@@ -3,15 +3,22 @@
...
 
       def run
+        check_for_zero_paid
         create_subscription unless manual_invoice?
         stripe_webhook_event.processed!
       rescue => e
         stripe_webhook_event.log_error(e)
       end
 
+      private def check_for_zero_paid
```
dandrabik: You might be able to check the `subscription.status` == 'active' to confirm this: https://stripe.com/docs/billing/subscriptions/webhooks#events

Screen Shot 2022-06-16 at 9 43 16 AM

dandrabik (approved):
Approving since this is better than the current state, but it might be worth testing the refund detection a little more (I think there is a way to determine if it's a refund, I left a comment).

By the way, love PRs like this: ➖ :100:

Screen Shot 2022-06-16 at 9 37 21 AM

services/QuillLMS/app/services/stripe_integration/webhooks/invoice_paid_event_handler.rb
brendanshean: Good idea.  I will investigate.

PR #9278, opened 2022-06-16
upgrade LMS to ruby 2.7

happythenewsad:
## WHAT
- upgrade Heroku environments to Ruby 2.7.6
- upgrade CircleCI environment to Ruby 2.7.6
- patch existing code that is deprecated in Ruby 2.7.6
- pin existing gems to minor versions, so that meaningful changes in Gemfile.lock are more easily noticeable 

## WHY
We need to keep our ruby versions up to date. 

## HOW
- mostly just change config files 

## Rollout and QA plans 
...are in the notion task 

### Notion Card Links
https://www.notion.so/quill/Product-Board-30f97e4eb01246dbb8264253fb385073?p=1a5e06502aa142e3a8ccf7779d96032c

PR Checklist | Your Answer
------------ | -------------
Have you added and/or updated tests? |  Yes. Also: a dedicated QA plan exists in the notion ticket.


services/QuillLMS/app/views/pages/home_new.html.erb
```diff
@@ -20,7 +20,7 @@
   </div>
   <div class="q-hero-footer">
     <p class="text-white">
-      <span class="hero-footer-underline"><%= number_to_human(@number_of_students).downcase %></span> students have written <span class="hero-footer-underline"><%= number_to_human(@number_of_sentences).downcase %></span> sentences on Quill.
```
happythenewsad, line 23: `number_to_human` throws an exception after the ruby bump, due to the internally bundled BigDecimal library's API changing. Since we're only using `number_to_human` in this one file, I just wrote a custom helper to replace `number_to_human`. 

Adding `bigdecimal` to the Gemfile and telling Rails to prefer use of this BigDecimal instance did not fix the issue. That approach requires a new gem anyway.

brendanshean: Do you mean after the ruby bump?

happythenewsad: Fixed it, thanks

brendanshean: Also, what is the specific exception that you get?  I'm not seeing any exception with `number_to_human` locally on this branch.

happythenewsad: undefined method `new' for BigDecimal:Class

brendanshean (approved):
Looks good to me.  I like your solution of not adding another gem.

emilia-friedberg (approved):

anathomical (approved):
I especially love pinning the minor versions of our Gems.  Something we probably should have done already, but now's a good time to do it.

brendanshean:
While working on Rails 6 upgrade, I noticed these two files:
https://github.com/empirical-org/Empirical-Core/blob/develop/services/.ruby-version
https://github.com/empirical-org/Empirical-Core/blob/develop/services/rubocop.sh

It might be worth upgrading in this PR.

happythenewsad:
> 

~Good idea; addressed~ 

I'm addressing this separately, as it's more complex than it looks.

PR #9281, opened 2022-06-17
Evidence/specify specific checks to run

anathomical:
Follow-on to #9191

## WHAT
Add a new way to specify a limited set of checks to run when asking for feedback.  This isn't really intended for use by normal API consumers, but is more a debugging/testing tool.  As such, it may really belong on a separate controller action.  Open to discussion on that front.  (Basically, I don't think we want non-test/non-debugging API calls to use this functionality.  Specific common-use limits to checks should belong on their own endpoints for sure.)
## WHY
We want to be able to test specific Checks via curl or Postman during development and testing.
## HOW
- Add a new optional param that the feedback controller can consume that specifies an array of checks to limit the feedback request to
- If the param is not supplied, default to running all configured checks
- Add a `debug` key to the Fallback feedback that contains the `message` of any errors in cases where an error is raised during the check process
- If the param is supplied, but none of the provided items in the array are valid Check names, return the fallback feedback with a `debug` message explaining that this happened

### Notion Card Links
https://www.notion.so/quill/Remove-extra-Evidence-FeedbackController-endpoints-for-individual-APIs-4dd3c3d6608f4410820bf7f6ef78b89f

PR Checklist | Your Answer
------------ | -------------


services/QuillLMS/engines/evidence/lib/evidence/evidence/check.rb
```diff
@@ -20,44 +22,61 @@ module Check
...
-    def self.get_feedback(entry, prompt, previous_feedback)
-      triggered_check = find_triggered_check(entry, prompt, previous_feedback)
+    def self.get_feedback(entry, prompt, previous_feedback, feedback_types=nil)
+      triggered_check = find_triggered_check(entry, prompt, previous_feedback, feedback_types)
 
-      triggered_check&.response || fallback_feedback
+      triggered_check || fallback_feedback
     end
 
     # returns first nonoptimal feedback, and if all are optimal, returns automl feedback
```
happythenewsad, line 31: Please mention in the comment that this function can also return `nil`, as that's an important code path.

services/QuillLMS/engines/evidence/lib/evidence/evidence/check.rb
```diff
@@ -20,44 +22,61 @@ module Check
...
 
+      return FALLBACK_RESPONSE.merge({debug: debug}) if debug
+
       FALLBACK_RESPONSE
     end
+
+    def self.checks_to_run(feedback_types)
+      return ALL_CHECKS if feedback_types.nil? || feedback_types.empty?
+
+      qualified_feedback_types = feedback_types.map { |t| "Evidence::Check::#{t}" }
```
happythenewsad, line 74: This will raise an  `uninitialized constant`  when a member of `feedback_types` isn't a member of ALL_CHECKS. Cleaner to compare  `feedback_types`  against `ALL_CHECKS.map(&:to_s)` IMO.

happythenewsad (changes requested):
Looks great, apart from a few minor things.

You raised a valid issue re: exposing a debug API to the public. I counter that keeping the debug API in a separate code path is worse, since the implementation of the debug API could drift from the public API over time.

happythenewsad (approved):

services/QuillLMS/engines/evidence/lib/evidence/evidence/check.rb
anathomical: What conditions result in a `nil` return?  I think that all code paths should return the fallback feedback if something unexpected happens.

PR #9283, opened 2022-06-20
Fix bug with stripe_setup_intent

brendanshean:
## WHAT
Fix an [error](https://sentry.io/organizations/quillorg-5s/issues/3360924957/?referrer=slack
) involving the webhook event `setup_intent.succeeded`

## WHY
We would like relevant webhooks to be handled properly.  

The typical use case for this webhook is a user changing their credit card on a subscription.  In this case, the user changed had no underlying stripe subscription, but updated their credit card on stripe.

## HOW
The `event.metadata` call returns a `Stripe::StripeObject` instead of a hash.  Safe navigation with `metadata&.subscription_id` returns an error here, so use `metadata.respond_to?(:subscription_id)` instead.

### Screenshots

### Notion Card Links

PR Checklist | Your Answer
------------ | -------------

### DIFF
--- a/services/QuillLMS/app/services/stripe_integration/webhooks/setup_intent_succeeded_event_handler.rb
+++ b/services/QuillLMS/app/services/stripe_integration/webhooks/setup_intent_succeeded_event_handler.rb
@@ -12,6 +12,10 @@ module StripeIntegration
         stripe_webhook_event.log_error(e)
       end
 
+      private def metadata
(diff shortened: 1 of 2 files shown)


dandrabik (approved):
:+1: Thanks for knocking out these edge cases as they come in!

emilia-friedberg (approved):

PR #9287, opened 2022-06-21
Update secure-headers-spec

brendanshean:
## WHAT
Update a spec in anticipation of Rails 5.2 upgrade.

## WHY
In [Rails 5.2](https://github.com/rails/rails/issues/32557#issuecomment-382864622) the headers for 'Cache-Control' the will be changed.

## HOW
Remove a couple of keys from the spec.

### Screenshots

### Notion Card Links

PR Checklist | Your Answer
------------ | -------------
Spec Review: Have you reviewed the spec and ensured this meets requirements and/or matches design mockups? | YS

### DIFF
--- a/services/QuillLMS/spec/requests/secure_headers_spec.rb
+++ b/services/QuillLMS/spec/requests/secure_headers_spec.rb
@@ -27,7 +27,7 @@ describe DummyController, type: :request do
 
   it 'should set cache control headers' do
     get '/dummy'
-    expect(response.header['Cache-Control']).to match('no-cache, no-store, max-age=0, must-revalidate')
+    expect(response.header['Cache-Control']).to match('no-cache, no-store')
     expect(response.header['Pragma']).to match('no-cache')
   end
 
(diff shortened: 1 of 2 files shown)


services/QuillLMS/spec/requests/secure_headers_spec.rb
```diff
@@ -27,7 +27,7 @@ def index
 
   it 'should set cache control headers' do
     get '/dummy'
-    expect(response.header['Cache-Control']).to match('no-cache, no-store, max-age=0, must-revalidate')
+    expect(response.header['Cache-Control']).to match('no-cache, no-store')
```
dandrabik, line 30: Note, we are setting this [here](https://github.com/empirical-org/Empirical-Core/blob/develop/services/QuillLMS/app/controllers/application_controller.rb#L124). So Rails is stripping it out, but we should also not set it to avoid confusion in the future. This might be worth pinging @happythenewsad to give a once over since he worked on this.

dandrabik (approved):
:+1: This looks good to me. Just an FYI, this line came up in an external security review we did last year, but this change complies with their recommendation (setting to `'no-cache, no-store'`). I left a comment about updating our code that sets this setting (that Rails now strips out).

happythenewsad (approved):
LGTM. 

My only comment is that we should add a "inspect headers" item to your QA / rollout checklist as part of the larger project.

PR #9313, opened 2022-06-30
Update active model serializers

brendanshean:
## WHAT
Upgrade the active model serializers to 0.10.13

## WHY
There's a deprecation related to the current version of the gem:
```
DEPRECATION WARNING: `Module#parent` has been renamed to `module_parent`. 
`parent` is deprecated and will be removed in Rails 6.1. 
```
referenced [here](https://github.com/rails-api/active_model_serializers/issues/2316#issuecomment-493128817)

## HOW
1.  Change the version and run bundle update.

2. The bump from v0.9 -> v0.10 adds a breaking a change that removes root keys by default.  The docs [suggest](https://github.com/QultureRocks/active_model_serializers/blob/master/docs/howto/add_root_key.md) configuring an json adapter to configure root keys, however this doesn't work for serializers used [directly](https://github.com/QultureRocks/active_model_serializers/blob/master/docs/general/adapters.md).  The key point is highlighted here:

> IMPORTANT: Adapter configuration has no effect on a serializer instance being used directly. That is, UserSerializer.new(user).as_json will always behave as if the adapter were the 'Attributes' adapter. See [Outside Controller Usage](https://github.com/rails-api/active_model_serializers/blob/0-10-stable/docs/howto/outside_controller_use.md) for more details on recommended usage.

To solve this I'm abstracting out a solution via [ApplicationSerializer](https://github.com/empirical-org/Empirical-Core/pull/9313/files#diff-5e3ecc225e00ef6c4ebead7c5713903fb9bd7c40d28a5ab3ffb54c83d30a6076R3).  This super class checks for `root: false` being passed in and if so, use the v0.10 behavior (i.e. `super`).  Otherwise, following the code [here](https://github.com/rails-api/active_model_serializers/blob/0-10-stable/docs/howto/outside_controller_use.md) of first create a serializable resource, then call `as_json` to get root keys included.

### Screenshots

### Notion Card Links

PR Checklist | Your Answer
------------ | -------------


brendanshean:
@anathomical @happythenewsad I'll resubmit once I clean up something Peter pointed out.

brendanshean:
@happythenewsad @anathomical Should be ready for review now

anathomical (approved):
Looks good to me.  I like creating the intervening `Serializer` class to standardize our custom behavior on.

happythenewsad:
@brendanshean  Going forward, should we always use `root: null` when invoking serializes, if we don't explicitly specify the top level json property name?

brendanshean:
> @brendanshean Going forward, should we always use `root: null` when invoking serializes, if we don't explicitly specify the top level json property name?

@happythenewsad I would use `root: false` if we _do_ specify the top level property name. For example, the following three forms would be equivalent:

1.  `WidgetSerializer.new(widget).as_json` 
2.  `{ widget: WidgetSerializer.new(widget).as_json(root: false)}` 
3.  `{ widget: { widget_attr1: val1, ... , widget_attrN: valN } }`

services/QuillLMS/app/serializers/application_serializer.rb
```diff
@@ -0,0 +1,21 @@
...
+  end
+
+  private def options
+    { serializer: self.class, adapter: :json }
+  end
+
+  private def serializable_resource
+    return nil if object.nil?
+
+    ActiveModelSerializers::SerializableResource.new(object, options)
```
happythenewsad, line 19: In this PR's execution context:


```
[1] pry(main)> ActiveModelSerializers::SerializableResource.new(nil)                                                                                                          
=> #<ActiveModelSerializers::SerializableResource:0x00007fb828115e58 @adapter_opts={}, @resource=nil, @serializer_opts={}>
[3] pry(main)> ActiveModelSerializers::SerializableResource.new(nil).as_json                                                                                                  
Rendered ActiveModel::Serializer::Null with Class (0.04ms)
=> nil
```

Looks like the early return is a no-op

services/QuillLMS/app/serializers/application_serializer.rb
```diff
@@ -0,0 +1,21 @@
+# frozen_string_literal: true
+
+class ApplicationSerializer < ActiveModel::Serializer
```
happythenewsad, line 3: How can we avoid developers accidentally subclassing from ActiveModel::Serializer instead of this new class?

One way would be to monkeypatch `ActiveModel::Serializer` and include a `self.inherited` hook that raises an exception or prints a warning when `ActiveModel::Serializer` is subclassed. I don't think this is ideal, however, because it breaks the abstraction layer between the gem and LMS code.

happythenewsad (approved):
I have a nit you can fix, but otherwise looks good.

services/QuillLMS/app/serializers/application_serializer.rb
brendanshean: I think it's still necessary though since the options has will break the creation of this resource:

```
[1] pry(main)> options = { serializer: UserSerializer, adapter: :json }                                                       
=> {:serializer=>UserSerializer, :adapter=>:json}
[2] pry(main)> object = nil                                                                                                   
=> nil
[3] pry(main)> ActiveModelSerializers::SerializableResource.new(object, options).as_json                                      
Rendered UserSerializer with ActiveModelSerializers::Adapter::Json (3.61ms)
NoMethodError: undefined method `model_name' for NilClass:Class
Did you mean?  module_parent
from /Users/brendanshean/.asdf/installs/ruby/2.7.6/lib/ruby/gems/2.7.0/gems/active_model_serializers-0.10.13/lib/active_model/serializer.rb:388:in `json_key'
```

services/QuillLMS/app/serializers/application_serializer.rb
brendanshean: It's an interesting question but one that we can probably leave up to code review since this would be more about enforcing a coding convention than the code not working.  Analogous issues exists not inheriting from `ApplicationRecord` for models, `ApplicationController` and controllers, etc.  In all of these cases, there's probably some missing functionality that will nudge the developer to inherit from the correct superclass.

services/QuillLMS/app/serializers/application_serializer.rb
happythenewsad: Gotcha

dandrabik (approved):
This looks good to me. I like this pattern, feels like a standard Rails approach.

Also, nice that if you use a generator (e.g. `rails g serializer student`, new serializers will descend from `ApplicationSerializer`

Screen Shot 2022-07-11 at 11 43 37 AM

brendanshean:
> This looks good to me. I like this pattern, feels like a standard Rails approach.
> 
> Also, nice that if you use a generator (e.g. `rails g serializer student`, new serializers will descend from `ApplicationSerializer`

Cool.  I didn't even think to check that.

PR #9315, opened 2022-06-30
Lms/rename concept results table

anathomical:
## WHAT
Rename `ConceptResult` model to `OldConceptResult` and the `concept_results` database table to `old_concept_results`

Note: In addition to manual testing on staging, I've run staging Ghost Inspector tests, too.  Everything looks good to me.
## WHY
Given that the new model that we're going to replace the current `ConceptResult` with is going to be a row-for-row replacement, we decided it made sense to call the new model `ConceptResult`.  In preparation for that, we want to rename the old table so that the table name is available when we migrate.
## HOW
- Add a migration to rename `concept_results` to `old_concept_results`
- Change the name of the model from `ConceptResult` to `OldConceptResult`
- Update model associations to point at the renamed model
- Update code to reference the new model name
- Update all raw SQL that references `concept_results` to reference `old_concept_results` instead
- Update all ActiveRecord association references from `concept_results` to the renamed `old_concept_results` association name

### Notion Card Links
https://www.notion.so/quill/Rename-ConceptResult-to-OldConceptResult-in-preparation-for-replacing-it-with-the-new-normalized-f6002e476c0943d7ae2d27ef3e655583

PR Checklist | Your Answer
------------ | -------------

### DIFF
--- a/services/QuillLMS/app/services/demo/concept_results.rb
+++ b/services/QuillLMS/app/services/demo/concept_results.rb
@@ -29,7 +29,7 @@ module Demo::OldConceptResults
   end
 
   def self.handle_scores(activity_session)
-    crs = activity_session.reload.concept_results
+    crs = activity_session.reload.old_concept_results
     percentage = activity_session.percentage
     num_correct = (percentage*crs.count).ceil
     shuffled = crs.shuffle
--- a/services/QuillLMS/app/services/demo/concept_results.rb
+++ b/services/QuillLMS/app/services/demo/concept_results.rb
@@ -1,6 +1,6 @@
 # frozen_string_literal: true
 
-module Demo::ConceptResults
+module Demo::OldConceptResults
 
   def self.create_from_classrooms(classrooms)
     activity_sessions = classrooms
@@ -54,7 +54,7 @@ module Demo::ConceptResults
     concept = Concept.find_or_create_by(name: concept_name)
     crs = []
     number_of_concept_results.to_i.times do |i|
-      cr = ConceptResultOld.find_or_create_by(concept: concept, activity_session: activity_session)
+      cr = OldConceptResult.find_or_create_by(concept: concept, activity_session: activity_session)
       cr.update(metadata: {})
       crs.push(cr)
     end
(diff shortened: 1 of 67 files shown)


services/QuillLMS/app/services/demo/concept_results.rb
```diff
@@ -1,6 +1,6 @@
 # frozen_string_literal: true
 
-module Demo::ConceptResults
```
brendanshean, line 3: It probably doesn't really matter, but maybe rename file as old_concept_results.rb?

brendanshean (approved):
It's nice to see these stages being completed!  Just a few notes but nothing that should hold up approval.

We should probably add 'OldConceptResults' to the exclude [list](https://github.com/empirical-org/Empirical-Core/blob/develop/services/QuillLMS/config/initializers/rails_admin.rb#L174) for RailsAdmin.

I noticed a rake this rake [task](https://github.com/empirical-org/Empirical-Core/blob/develop/services/QuillLMS/lib/tasks/temporary/concept_results.rake#L5) has a ConceptResult reference and then I left one other comment on a file name.  Both probably aren't required but figured I would point them out anyway.

While I did notice some other references to concept_results, they seemed to all be temporary variables and while I would rename of those all just so I have less to juggle, I think that's a judgment call for you to make since you'll be handling the subsequent stages.  There was one in [particular](https://github.com/empirical-org/Empirical-Core/blob/a0f1ad97c3d574297cf1a49e829519891d14bf7f/services/QuillLMS/app/controllers/api/v1/activity_sessions_controller.rb#L130) concerning params where we want to make sure that the value in the params isn't being inferred upstream by the model name (i.e. old_concept_results instead of concept_results).  Our testing probably catches this anyway.

services/QuillLMS/app/services/demo/concept_results.rb
anathomical: This is an interesting point.  I think that for a lot of these cases, my gut feeling is to leave these local variables and function params unchanged because the hope is that within a month or two we'll be rotating all references to `old_concept_results` back to the replacement `concept_results` records when the new model is in place.  So while I agree that there's some mental juggling to be done in the short term, it'll simplify things in the longer-term.

anathomical: (Note, I'm also trying not to touch front-end code, so as far as HTTP payloads are concerned, this is still `concept_result`.  Agree that, as with the above, this can be confusing in the short-term, but I'm hoping that the simplification in the long-term will pay off.

PR #9317, opened 2022-07-01
check in latest lockfile

happythenewsad:
## WHAT
Forgot to generate and check in lockfile after rebasing.


emilia-friedberg (approved):

happythenewsad:
The failing spec is `spec/services/vitally_integration/serialize_vitally_sales_user_spec.rb`, which seems unrelated to checking in the Gemfile.lock

PR #9322, opened 2022-07-05
make sure to include activity id in the key for diagnostic results summary

emilia-friedberg:
## WHAT
Remove caching for the results_summary for the diagnostic pages.

## WHY
This cache doesn't care about the activity itself, so in cases where multiple diagnostics were assigned but have the same classroom_unit, teachers can only ever see the data for the most recently completed diagnostic. We don't want that to happen.

## HOW
Just remove the caching wrapper (we already weren't caching the `growth_results_summary`, maybe for the same reason).

### Screenshots

### Notion Card Links
https://www.notion.so/quill/Incorrect-results-showing-up-for-ELL-Intermediate-and-Advanced-Pre-test-Diagnostics-5768fecfa3a547239edd813dccd8cf0f

PR Checklist | Your Answer
------------ | -------------

### DIFF
--- a/services/QuillLMS/app/controllers/teachers/progress_reports/diagnostic_reports_controller.rb
+++ b/services/QuillLMS/app/controllers/teachers/progress_reports/diagnostic_reports_controller.rb
@@ -272,7 +272,20 @@ class Teachers::ProgressReports::DiagnosticReportsController < Teachers::Progres
   end
 
   def diagnostic_results_summary
-    render json: ResultsSummary.results_summary(results_summary_params[:activity_id], results_summary_params[:classroom_id], results_summary_params[:unit_id])
+    render json: fetch_diagnostic_results_summary_cache
+  end
+
+  private def fetch_diagnostic_results_summary_cache
+    groups = { activity_id: params[:activity_id] }
+    current_user.classroom_unit_by_ids_cache(
+      classroom_id: params[:classroom_id],
+      unit_id: params[:unit_id],
+      activity_id: params[:activity_id],
+      key: 'teachers.progress_reports.diagnostic_reports.diagnostic_results_summary',
+      groups: groups
+    ) do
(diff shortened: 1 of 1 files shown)


anathomical (approved):
This does seem like the only solution available to this problem as we don't have a more specific level of caching (something at the `activity` level, perhaps).  I'll write up a task card to add a more specific caching level and add it back here, but that may not be prioritized for a while.

In the meantime, this seems like a good solution.

services/QuillLMS/app/controllers/teachers/progress_reports/diagnostic_reports_controller.rb
```diff
@@ -272,18 +272,7 @@ def diagnostic_status
...
-    render json: fetch_diagnostic_results_summary_cache
-  end
-
-  private def fetch_diagnostic_results_summary_cache
-    current_user.classroom_unit_by_ids_cache(
-      classroom_id: params[:classroom_id],
-      unit_id: params[:unit_id],
-      activity_id: params[:activity_id],
-      key: 'teachers.progress_reports.diagnostic_reports.diagnostic_results_summary'
-    ) do
```
anathomical: I think that we can actually fix this by adding a `groups` param to the cache call, like this.
```
groups = {
  activity_id: params[:activity_id]
}
current_user.classroom_unit_by_ids_cache(
  classroom_id: params[:classroom_id],
  unit_id: params[:unit_id],
  activity_id: params[:activity_id],
  key: 'teachers.progress_reports.diagnostic_reports.diagnostic_results_summary',
  groups: groups
) do
```

anathomical (changes requested):
Actually, on second thought, I think there is a solution to this problem.  We can pass a `groups` value into the cache call after the `key` value.  `groups` is a hash value that's used to validate further uniqueness.  So if we create a group hash using `activity_id`, it should build and use separate caches for different activities even if they're part of the same `classroom_unit`.

anathomical (approved):

PR #9324, opened 2022-07-05
add grade levels and support for curriculum to change them

emilia-friedberg:
## WHAT
Add a migration to add `minimum_grade_level` and `maximum_grade_level` to the `activity` table, then add support to auto-populate these values based on the readability grade level both as a callback on the model and also from the frontend for the curriculum team.

## WHY
We want to use these broader grade ranges widely. 

## HOW
After reviewing everyone's feedback on the RFC, I decided to move forward with the minimum and maximum integer approach but on the activity table itself, since as Dan pointed out it's really just a property of that table and not an independent entity. Hannah expressed that she would like for these values to be autopopulated based on the readability grade level, but for the curriculum team to be able to change them, so I enabled that both in a callback step on save and also in the Activity Editor (so the curriculum team doesn't actually have to press save to see the values).

### Screenshots

### Notion Card Links
https://www.notion.so/quill/Grade-ranges-bcb7769a2e55498eb89a6184230e6a64

PR Checklist | Your Answer
------------ | -------------


eadams17 (approved):

services/QuillLMS/app/models/activity.rb
```diff
@@ -66,6 +68,7 @@ class Activity < ApplicationRecord
   has_many :activity_topics, dependent: :destroy
   has_many :topics, through: :activity_topics
   before_create :flag_as_beta, unless: :flags?
+  before_save :set_minimum_and_maximum_grade_levels_to_default_values, unless: :minimum_grade_level
```
anathomical, line 71: This isn't strictly necessary since we don't really have much in the way of validations on these older models, but it might be worth adding min/max validations for grade levels:
```ruby
validates :minimum_grade_level, numericality: { greater_than_or_equal_to: 4, less_than_or_equal_to: 12 }
validates :maximum_grade_level, numericality: { greater_than_or_equal_to: 4, less_than_or_equal_to: 12 }

```
Or whatever the appropriate ranges are.

services/QuillLMS/db/migrate/20220705143703_add_minimum_and_maximum_grade_levels_to_activities.rb
```diff
@@ -0,0 +1,8 @@
+# frozen_string_literal: true
+
+class AddMinimumAndMaximumGradeLevelsToActivities < ActiveRecord::Migration[5.2]
+  def change
+    add_column :activities, :minimum_grade_level, :integer
+    add_column :activities, :maximum_grade_level, :integer
```
anathomical: Another super minor thing that almost certainly doesn't matter for our purposes, but these columns could be type `:tinyint` rather than `:integer`.  `tinyint` has a maximum value of 127, which is way larger than what we'd ever use here, and it takes up less space on disk to store.  Given how few `activities` rows we have, it probably doesn't matter, but I figured I'd flag it as something to consider when adding numerical columns in the future.

anathomical (approved):
I flagged a couple of very minor enhancements that could be done on the backend, but I don't think either of them really have to be done here considering the scope of the work.  So, more for your edification than anything else. 

This looks good to me.

brendanshean (commented):
The `minimum_grade_level` and `maximum_grade_level` attributes feel coupled to me which would point to extracting a GradeBand object with its own table.  I think it would at least be worth exploring that redesign since it would move reorganize some of the logic into one place.  This would keep the activity model and raw_score models lighter.

emilia-friedberg:
> The `minimum_grade_level` and `maximum_grade_level` attributes feel coupled to me which would point to extracting a GradeBand object with its own table. I think it would at least be worth exploring that redesign since it would move reorganize some of the logic into one place. This would keep the activity model and raw_score models lighter.

Hey @brendanshean, I appreciate the feedback. I should probably have talked about this more in the PR summary, but in addition to Dan's feedback about grade bands as really properties of the `activity` table, it occurred to me while I was planning that most of the logic relevant to the grade bands would still properly belong to the activity model, since the default values are based on the activity's raw score (which the grade bands, were they independent, shouldn't really need to know anything about). Given that, as well as the fact that we would need to add a `grade_band_id` to the activity table (meaning separating it out into its own table only saves us one additional column on the activity table), and the fact that having it in one table will make it easier for us to query, I feel pretty confident that they are where they belong for now. Let me know if you still have strong reservations and we can talk about it more.

services/QuillLMS/db/migrate/20220705143703_add_minimum_and_maximum_grade_levels_to_activities.rb
emilia-friedberg: Done! It looks like it is :smallint in postgres, just fyi.

services/QuillLMS/app/models/activity.rb
```diff
@@ -227,9 +243,24 @@ def add_question(question)
...
+  def default_maximum_grade_level
+    return nil if readability_grade_level.nil?
+
+    DEFAULT_MAX_GRADE_LEVEL
+  end
 
-    raw_score.readability_grade_level(activity_classification_id)
+  def set_minimum_and_maximum_grade_levels_to_default_values
+    self.minimum_grade_level = default_minimum_grade_level
+    self.maximum_grade_level = default_minimum_grade_level ? default_maximum_grade_level : nil
```
brendanshean, line 263: self.maximum_grade_level = default_maximum_grade_level if default_minimum_grade_level

services/QuillLMS/app/models/activity.rb
```diff
@@ -101,6 +107,16 @@ class Activity < ApplicationRecord
   ELL_ADVANCED_DIAGNOSTIC_ACTIVITY_ID = 1590
   PRE_TEST_DIAGNOSTIC_IDS = [STARTER_DIAGNOSTIC_ACTIVITY_ID, INTERMEDIATE_DIAGNOSTIC_ACTIVITY_ID, ADVANCED_DIAGNOSTIC_ACTIVITY_ID, ELL_STARTER_DIAGNOSTIC_ACTIVITY_ID, ELL_INTERMEDIATE_DIAGNOSTIC_ACTIVITY_ID, ELL_ADVANCED_DIAGNOSTIC_ACTIVITY_ID]
 
+  READABILITY_GRADE_LEVEL_TO_MINIMUM_GRADE_LEVEL_MAP = {
+    RawScore::SECOND_THROUGH_THIRD => 4,
```
brendanshean, line 111: Just double checking this should be 4 and not 2.

services/QuillLMS/app/controllers/cms/activities_controller.rb
```diff
@@ -83,6 +85,7 @@ def destroy
     @raw_score_options = RawScore.order_by_name
     @grade_band_hash = {}
     @raw_score_options.each { |rs| @grade_band_hash[rs.name] = rs.readability_grade_level(@activity_classification.id) }
+    @readability_grade_band_to_minimum_grade_level_map = Activity::READABILITY_GRADE_LEVEL_TO_MINIMUM_GRADE_LEVEL_MAP
```
brendanshean: I think `_map` suffix at the end of the variable name: `@readability_grade_band_to_minimum_grade_level_map` and the constant: `Activity::READABILITY_GRADE_LEVEL_TO_MINIMUM_GRADE_LEVEL_MAP` is redundant since both have the word "to" in the middle.

brendanshean: I realize it's older code, but I think it would be worth renaming `@grade_band_hash` to something consistent with the naming of  `@readability_grade_band_to_minimum_grade_level`.

brendanshean (approved):
Nice work on this!  A couple of very minor comments, but nothing holding up approval.

services/QuillLMS/app/models/activity.rb
emilia-friedberg: So that's actually what I had initially, but I think in the case where for whatever reason there already was a `maximum_grade_level` set here, you would want to explicitly reset it to nil.

services/QuillLMS/app/models/activity.rb
emilia-friedberg: Yes, thanks for double checking! We start at 4.

services/QuillLMS/app/controllers/cms/activities_controller.rb
```diff
@@ -81,11 +81,11 @@ def destroy
     @activity_classification = ActivityClassification.find_by_id!(params[:activity_classification_id])
   end
 
-  protected def set_raw_score_options_and_grade_band_hash
+  protected def set_raw_score_options_and_raw_score_to_readability_grade_band
     @raw_score_options = RawScore.order_by_name
-    @grade_band_hash = {}
-    @raw_score_options.each { |rs| @grade_band_hash[rs.name] = rs.readability_grade_level(@activity_classification.id) }
-    @readability_grade_band_to_minimum_grade_level_map = Activity::READABILITY_GRADE_LEVEL_TO_MINIMUM_GRADE_LEVEL_MAP
+    @raw_score_to_readability_grade_band = {}
```
brendanshean, line 86: @emilia-friedberg Much better!

PR #9326, opened 2022-07-05
Numeric lib should handle strings more gracefully

happythenewsad:
## WHAT
Numeric lib should handle string input more gracefully.

## WHY
Redis values are string by default.

### Screenshots

### Notion Card Links
none

PR Checklist | Your Answer
------------ | -------------
Self-Review: Have you done an initial self-review of the code below on Github? | na

### DIFF
--- a/services/QuillLMS/app/lib/utils/numeric.rb
+++ b/services/QuillLMS/app/lib/utils/numeric.rb
@@ -2,7 +2,7 @@
 
 module Utils::Numeric
   def self.to_human_string(numeric)
-    raise ArgumentError, "Not a numeric" unless numeric.is_a?(Numeric)
+    return numeric.to_s unless numeric.is_a?(Numeric)
 
     millions = ->(x) { x >= 1_000_000 && x < 1_000_000_000 }
     billions = ->(x) { x >= 1_000_000_000 }
--- a/services/QuillLMS/app/lib/utils/numeric.rb
+++ b/services/QuillLMS/app/lib/utils/numeric.rb
@@ -2,16 +2,16 @@
 
 module Utils::Numeric
   def self.to_human_string(numeric)
-    return numeric.to_s unless numeric.is_a?(Numeric)
+    numeric = numeric.to_f
 
     millions = ->(x) { x >= 1_000_000 && x < 1_000_000_000 }
     billions = ->(x) { x >= 1_000_000_000 }
 
     case numeric
     when millions
-      "#{(numeric.to_f / 1_000_000).round} Million"
+      "#{(numeric / 1_000_000).round} Million"
     when billions
-      "#{(numeric.to_f / 1_000_000_000).round} Billion"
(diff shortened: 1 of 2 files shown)


services/QuillLMS/app/lib/utils/numeric.rb
```diff
@@ -2,7 +2,7 @@
 
 module Utils::Numeric
   def self.to_human_string(numeric)
-    raise ArgumentError, "Not a numeric" unless numeric.is_a?(Numeric)
+    return numeric.to_s unless numeric.is_a?(Numeric)
```
anathomical: If something can be converted to a number, like these values in Redis, don't we actually want the utility of the `Numeric` conversion here?  Like, if we pull `"5700000"` out of Redis, don't we want this to actually return "6 million" or whatever?

anathomical (changes requested):
Do we actually want to just throw strings back?  Shouldn't we actually process numbers that just happen to be stored as strings?

services/QuillLMS/app/lib/utils/numeric.rb
happythenewsad: Yes, that's a nice enhancement; I've added it. I leveraged the fact the `to_f` will never raise an exception based on it's input. 

`"asdf".to_f => 0.0`

anathomical (commented):
Nice.  Do we need to deal with the possibility of someone passing `nil` or `"hello"` into the function mistakenly, or should we just deal with that if/when it happens?

brendanshean:
> Nice. Do we need to deal with the possibility of someone passing `nil` or `"hello"` into the function mistakenly, or should we just deal with that if/when it happens?

I'm guessing this probably related: I pushed a temporary [fix](https://github.com/empirical-org/Empirical-Core/pull/9332) due to this [error](https://one.newrelic.com/nr1-core/errors/overview/MjYzOTExM3xBUE18QVBQTElDQVRJT058NTQ4ODU2ODc1?account=2639113&duration=10800000&state=e3a4cfcf-c418-926a-1d72-f2718b48341d)

happythenewsad:
> 

Both of these cases are currently handled. This helper function is only called in one place in the code, FYI.

anathomical (approved):

happythenewsad:
> > Nice. Do we need to deal with the possibility of someone passing `nil` or `"hello"` into the function mistakenly, or should we just deal with that if/when it happens?
> 
> I'm guessing this probably related: I pushed a temporary [fix](https://github.com/empirical-org/Empirical-Core/pull/9332) due to this [error](https://one.newrelic.com/nr1-core/errors/overview/MjYzOTExM3xBUE18QVBQTElDQVRJT058NTQ4ODU2ODc1?account=2639113&duration=10800000&state=e3a4cfcf-c418-926a-1d72-f2718b48341d)

That's right - you can revert that commit once this one deploys.

PR #9336, opened 2022-07-06
lms/clean-up-duplicate-schools

anathomical:
## WHAT
Add a rake task to clean up duplicate school entries
## WHY
When we renewed our `School` data from NCES sources this year (the last time we'd updated this data was in 2014) we did not realize that while the NCES IDs for schools in the two datasets were the same, the original 2014 data set left-padded NCES IDs with "0"s while the new 2022 data set did not.  This meant that instead of recognizing that an existing record needed to be updated from the data, a new school was created instead.  We want to correct that mistake.
## HOW
- Identify all School records that have the same NCES ID as an older "original" school except that the older one has left-padded 0s
- For each of those "duplicate" schools, find any Users or Admins attached to them, and move those attachments over to the "original" school
- Update the "original" school columns with data from the "duplicate" school if the "duplicate" school has something set.  (Things like mailing addresses, phone numbers, student population, and FRPL status.)
- Delete the "duplicate" school from the database

### Notion Card Links
https://www.notion.so/quill/Clean-up-accidental-duplicate-records-in-Schools-87d040bfc1894314babd563665771ac1

PR Checklist | Your Answer
------------ | -------------

### DIFF
--- a/services/QuillLMS/lib/tasks/schools.rake
+++ b/services/QuillLMS/lib/tasks/schools.rake
@@ -172,7 +172,7 @@ namespace :schools do
     ActiveRecord::Base.connection.execute(schools_with_duplicates).each do |row|
       ActiveRecord::Base.transaction do
         duplicate = School.find(row['duplicate_school_id'])
-        update_hash = duplicate.as_json.except('id', 'nces_id').select { |_,v| v.present? }
+        update_hash = duplicate.as_json.except('id', 'nces_id', 'created_at').select { |_,v| v.present? }
         School.update(row['original_school_id'], update_hash)
 
         duplicate.destroy!
--- a/services/QuillLMS/lib/tasks/schools.rake
+++ b/services/QuillLMS/lib/tasks/schools.rake
@@ -175,6 +175,7 @@ namespace :schools do
         update_hash = duplicate.as_json.except('id', 'nces_id', 'created_at').select { |_,v| v.present? }
         School.update(row['original_school_id'], update_hash)
 
+        puts duplicate.as_json
         duplicate.destroy!
         duplicate_schools_deleted += 1
       end
--- a/services/QuillLMS/lib/tasks/schools.rake
+++ b/services/QuillLMS/lib/tasks/schools.rake
@@ -119,17 +119,15 @@ namespace :schools do
   desc 'Clean up duplicate schools based on 0-left-padded NCES ID'
   task :clean_up_duplicates => :environment do
 
-    # First we want to retrieve a list of all `schools_users` and
-    # `schools_admins` records where someone is assigned to a duplicate
-    # school and re-assign them to the correct school
     teachers_assigned_to_duplicates = <<-SQL
       SELECT original.id AS original_school_id,
+            duplicate.id AS duplicate_school_id,
             schools_users.id AS schools_users_id,
             schools_admins.id AS schools_admins_id
           FROM schools AS original
           JOIN schools AS duplicate
-              ON CONCAT('0', original.nces_id) = duplicate.nces_id
-                  OR CONCAT('00', original.nces_id) = duplicate.nces_id
+              ON original.nces_id = CONCAT('0', duplicate.nces_id)
+                  OR original.nces_id = CONCAT('0', duplicate.nces_id)
           LEFT OUTER JOIN schools_admins
               ON duplicate.id = schools_admins.school_id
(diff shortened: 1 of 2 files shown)


services/QuillLMS/lib/tasks/schools.rake
```diff
@@ -107,4 +107,61 @@ namespace :schools do
...
+  task :clean_up_duplicates => :environment do
+
+    teachers_assigned_to_duplicates = <<-SQL
+      SELECT original.id AS original_school_id,
+            duplicate.id AS duplicate_school_id,
+            schools_users.id AS schools_users_id,
+            schools_admins.id AS schools_admins_id
+          FROM schools AS original
+          JOIN schools AS duplicate
+              ON original.nces_id = CONCAT('0', duplicate.nces_id)
```
happythenewsad, line 127: Lines 129 and 130 have identical expressions. 

Furthermore, you should be able to accomplish the original-to-duplicate matching without an OR statement, since each side of the join will have , for example, id `012` and id `12`.

services/QuillLMS/lib/tasks/schools.rake
```diff
@@ -107,4 +107,61 @@ namespace :schools do
...
+          LEFT OUTER JOIN schools_admins
+              ON duplicate.id = schools_admins.school_id
+          LEFT OUTER JOIN schools_users
+              ON duplicate.id = schools_users.school_id
+    SQL
+
+    users_updated = 0
+    admins_updated = 0
+    duplicate_schools_deleted = 0
+    ActiveRecord::Base.connection.execute(teachers_assigned_to_duplicates).each do |row|
```
happythenewsad, line 137: This is more personal preference / risk tolerance, but I would migrate all data from the duplicates to the originals in one task, then do the deletions in a separate task, allowing for inspection between the two tasks.

services/QuillLMS/lib/tasks/schools.rake
```diff
@@ -107,4 +107,61 @@ namespace :schools do
...
+      ActiveRecord::Base.transaction do
+        if row['schools_users_id']
+          update_me = SchoolsUsers.find(row['schools_users_id'])
+          update_me.update(school_id: row['original_school_id'])
+          users_updated += 1
+        end
+
+        if row['schools_admins_id']
+          update_me = SchoolsAdmins.find(row['schools_admins_id'])
```
happythenewsad: Avoid repeating the same variable name `update_me`. Also, they should be named `schools_user` and `schools_admin` respectively.

services/QuillLMS/lib/tasks/schools.rake
happythenewsad: Also, with this query

`select count(*) from schools where nces_id::varchar(255) LIKE '0%'` 

we observe that only 19k, or ~12% of all schools, have an `nces_id` with a leading zero. Operating over this subset of schools should cause a significant speed up.

happythenewsad (changes requested):

services/QuillLMS/lib/tasks/schools.rake
brendanshean: I like the idea of separate tasks or logging the to-deleted-records to file before actually destroying.

brendanshean (approved):
Once Peter's comments are addressed I think this should be good to go!

services/QuillLMS/lib/tasks/schools.rake
anathomical: @happythenewsad Can you expand on what you mean in this second comment?  Are you suggesting adding a `WHERE` clause to the overall query, or are you suggesting something else?

happythenewsad: Up to you... I won't hold up the PR because of this, since the rake task may complete in a timely manner without it.

happythenewsad (approved):
Thanks for making the changes. 

Not sure if your terminal will buffer all of the output that 
`puts duplicate.as_json` will generate. Try piping to a file when executing. ✌️

anathomical:
> Thanks for making the changes.
> 
> Not sure if your terminal will buffer all of the output that `puts duplicate.as_json` will generate. Try piping to a file when executing. ✌️

That was exactly my plans: `heroku run 'rake schools:clean_up_duplicates' -a empirical-grammar > clean_up_duplicates.log`

PR #9339, opened 2022-07-07
Remove self.formats calls

brendanshean:
## WHAT
In the PagesController, remove attempts to force content headers to html for `home` and `home_new` actions.

## WHY
The `self.formats` calls were a [mechanism](https://github.com/empirical-org/Empirical-Core/pull/5390) for handling exceptions for invalid accept headers.  In Rails 6, this functionality is broken due to this [PR](https://github.com/rails/rails/pull/40353).  Basically, an accept header with invalid mime type like the spec below (i.e.         `headers = { Accept: "RandomGobbledyguck"}` will now trigger a `ActionDispatch::Http::MimeNegotiation::InvalidType` exception further upstream preventing it from reaching the controller method altogether.  This [thread](https://github.com/rails/rails/issues/37620#issuecomment-598221401) discusses ways of handling he error, but most of the discussion leads to updates in the various third-party libraries that log the exception reporting (e.g. honeybadger, sentry).

While, the long-term solution is to upgrade `sentry-raven` to the newer `sentry-ruby` since it will [ignore](https://github.com/getsentry/sentry-ruby/pull/1215) these exceptions,  such an upgrade will require [migration](    https://docs.sentry.io/platforms/ruby/migration) to sentry-ruby entailing more investigation.

## HOW
Remove the self.formats calls and the spec which handled the invalid accept headers.  This might result in new exceptions on sentry until we upgrade the gem.

### Screenshots

### Notion Card Links

PR Checklist | Your Answer
------------ | -------------

### DIFF
--- a/services/QuillLMS/app/controllers/pages_controller.rb
+++ b/services/QuillLMS/app/controllers/pages_controller.rb
@@ -23,7 +23,6 @@ class PagesController < ApplicationController
 
     @body_class = 'home-page'
     @activity = Activity.with_classification.find_by_uid(ENVr.fetch('HOMEPAGE_ACTIVITY_UID', ''))
-    self.formats = ['html']
   end
 
   # rubocop:disable Metrics/CyclomaticComplexity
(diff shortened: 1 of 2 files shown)


happythenewsad (approved):

anathomical (approved):
I think it's reasonable to deal with the extra error reporting in Sentry, especially since it should be just a single error that gets added, no mater what the volume ends up being.

PR #9343, opened 2022-07-07
Upgrade LMS Rails version to 6.0.5

brendanshean:
## WHAT
Upgrade LMS from 5.2.8 to 6.0.5

## WHY
Newer versions of Rails provide new features as well as security enhancements.

## HOW
Update Gemfile with new Rails version.  Then, after lots of flipping switches with Gemfile.lock, run `bundle`.

A couple of specs had to be fixed:
    1. Fixing module_parent calls:  `parent` will be deprecated in Rails 6 so update those calls to `module_parent`
    2. Fix time comparison: `as_json` calls in Rails 6 converts `ActiveSupport::TimeWithZone` values to strings whereas it currently leaves the object as a `ActiveSupport::TimeWithZone`.  (NB:  Comparisons with these objects are tricky, see [here](    https://tddium.wordpress.com/2011/08/07/rails-time-comparisons-devil-details-etc/) and [here](    https://stackoverflow.com/questions/20403063/trouble-comparing-time-with-rspec))

Addendum 1:
Cookie serialization is updated to `hybrid` per this [post](https://www.botreetechnologies.com/blog/rails-5-and-cookie-serialization)

Addendum 2:
Removed [livingstyleguide](https://github.com/livingstyleguide/livingstyleguide/issues/234) gem from development group since it breaks development build.  This was missed on heroku since the gem doesn't run.

### Screenshots

### Notion Card Links

PR Checklist | Your Answer
------------ | -------------


services/QuillLMS/Gemfile
```diff
@@ -142,28 +139,36 @@ gem 'rack-heartbeat'
...
+# Memory profiling
+gem 'puma_worker_killer', '~> 0.1.0'
+
+# temp for migrations
+gem 'paperclip'
+
+# TBD
+gem 'rb-readline'
+gem 'awesome_print'
+
```
anathomical, line 151: I like the consolidation you did here for items we should do later analysis on regarding whether we use these or not.  It may be worth going ahead and creating a Task card in Notion so that we remember to come back and do this.

services/QuillLMS/spec/models/feedback_history_spec.rb
```diff
@@ -129,12 +129,13 @@
...
       expect(json_hash['feedback_text']).to eq(@feedback_history.feedback_text)
       expect(json_hash['feedback_type']).to eq(@feedback_history.feedback_type)
-      expect(json_hash['time']).to eq(@feedback_history.time)
       expect(json_hash['metadata']).to eq(@feedback_history.metadata)
       expect(json_hash['rule_uid']).to eq(@feedback_history.rule_uid)
 
       expect(json_hash['prompt']).to eq(@feedback_history.prompt.as_json)
       expect(json_hash['prompt']['text']).to eq(@prompt.text)
+
+      expect(Time.iso8601(json_hash['time'])).to be_within(1.second).of @feedback_history.time
```
anathomical, line 138: This seems to be a reasonable solution to this problem.  I've generally fallen back to coercing string representations, but this feel cleaner.

anathomical (approved):

services/QuillLMS/Gemfile
brendanshean: I'm planning on doing an audit after the 6.1 upgrade.

services/QuillLMS/Gemfile
```diff
@@ -5,7 +5,7 @@ source 'https://rubygems.org'
 ruby '2.7.6'
 
 # CORE DEPS
-gem 'rails', '5.2.8'
+gem 'rails', '~> 6.0.5'
```
dandrabik: I'm curious the about why the change away from a locked patch version. I would think even a patch upgrade of Rails would be a manually planned/tested endeavor, so curious what we gain from this change.

services/QuillLMS/config/initializers/content_security_policy.rb
```diff
@@ -0,0 +1,30 @@
+# frozen_string_literal: true
```
dandrabik: Just to clarify, does this empty file mean this functionality is off (that's what we would want)? Wondering since we are using the [secure_headers](https://github.com/github/secure_headers) gem to set CSP [here](https://github.com/empirical-org/Empirical-Core/blob/develop/services/QuillLMS/config/initializers/secure_headers.rb). We'd want to ensure that these don't conflict.

services/QuillLMS/config/initializers/cookies_serializer.rb
```diff
@@ -0,0 +1,7 @@
+# frozen_string_literal: true
+
+# Be sure to restart your server when you modify this file.
+
+# Specify a serializer for the signed and encrypted cookie jars.
+# Valid options are :json, :marshal, and :hybrid.
+Rails.application.config.action_dispatch.cookies_serializer = :hybrid
```
dandrabik, line 7: Nice that they provide a migration path. I don't think this is an issue for how we use cookies, but just flagging it in case it comes up:
> SPOILER! When using the :json or :hybrid serializer, you should be aware that not all [Ruby](https://www.botreetechnologies.com/blog/interesting-methods-to-work-with-ruby-hashes-in-a-better-way) objects can be serialized as JSON. For example, Date and Time objects will be serialized as strings, and Hashes will have their keys stringified.

dandrabik (approved):
:+1: Thanks for your work on this. I left a few small comments on things to double check, but nothing is blocking.

services/QuillLMS/config/initializers/content_security_policy.rb
brendanshean: This was autogenerated, but I agree that it's confusing.  I will remove it.

services/QuillLMS/Gemfile
brendanshean: I've viewed patch upgrades even with Rails as safe updates, but I will change it back to the locked version since I don't really have a strong preference.

PR #9344, opened 2022-07-08
simplify and fix database.yml

happythenewsad:
## WHAT
1. Makes the `development` database config inherit from `development_shared`, thereby including the `database` yaml property. Without the `database` yaml property, this line of active_record will throw an unhandled exception, preventing local schema dumps: `activerecord-5.2.8/lib/active_record/tasks/postgresql_database_tasks.rb:116`

Here's the output of the error:
```
rake aborted!
TypeError: no implicit conversion of nil into String
/Users/pk/.rbenv/versions/2.6.6/bin/bundle:23:in `load'
/Users/pk/.rbenv/versions/2.6.6/bin/bundle:23:in `<main>'
Tasks: TOP => db:structure:dump
```

2. I also made database.yml a bit more DRY and readable.

## WHY
So that all developer environments can make schema dumps. 

### Screenshots

### Notion Card Links
none

PR Checklist | Your Answer
------------ | -------------
Have you added and/or updated tests? |  no - config change only

### DIFF
--- a/services/QuillLMS/config/database.yml
+++ b/services/QuillLMS/config/database.yml
@@ -1,17 +1,17 @@
 default: &default
   pool: <%= ENV.fetch("MAX_THREADS") { 5 } %>
   prepared_statements: false
+  adapter: postgresql
   timeout: 5000
 
-development_env: &development_env
-  adapter: postgresql
+development_shared: &development_shared
   encoding: unicode
   database: emp_gr_development
   host:     localhost
   port: 5432
 
 test_env: &test_env
-  <<: *development_env
+  <<: *development_shared
   database: emp_gr_test
 
 # While we don't actually connect to any follower DBs, they
@@ -24,7 +24,7 @@ leader_db_env: &leader_db_env
 
 development: &development
   <<: *default
-  adapter: postgresql
+  <<: *development_shared
   <<: *leader_db_env
 
 # Warning: The database defined as "test" will be erased and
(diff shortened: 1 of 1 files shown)


happythenewsad:
@emilia-friedberg Given that you mentioned in slack that you were able to dump schema without this PR, we may want to sync on this and compare environments.

happythenewsad:
Additionally, I notice that we started checking in `database.yml` in Jan 2020, making database.yml.example obsolete. Can I get a 👍 to remove all references to `database.yml.example` in a consecutive PR?

emilia-friedberg (approved):
I don't think this is quite the problem that I had (mine was attempting to make more significant structural changes to the structure.sql, but could write to it fine) but this change looks good to me regardless. I ended up just working around it by only checking in the relevant part of the structure.sql change (my new table).

services/QuillLMS/config/database.yml
```diff
@@ -1,17 +1,17 @@
...
   pool: <%= ENV.fetch("MAX_THREADS") { 5 } %>
   prepared_statements: false
+  adapter: postgresql
   timeout: 5000
 
-development_env: &development_env
-  adapter: postgresql
+development_shared: &development_shared
   encoding: unicode
   database: emp_gr_development
```
brendanshean: I think it's a little confusing to have `database: emp_gr_development` set and then overwritten with test_env.  What about moving it to the development group itself on line 25?

brendanshean (approved):
One minor suggestion, otherwise looks good.  Thank you for tracking down this error.

PR #9346, opened 2022-07-08
add ortto into head embed codes

emilia-friedberg:
## WHAT
Add Ortto tracking into head embed codes.

## WHY
We are moving to use Ortto instead of Intercom + Heap for product messaging and metrics.

## HOW
Just embed Ortto's code into our `head_embed_codes` partial, using the same logic we do for our other tracking platforms (ie the same rules as the load_intercom? method).

### Screenshots

### Notion Card Links
https://www.notion.so/quill/Add-Ortto-tracking-code-to-all-pages-website-and-application-b680c292587e45a8a03392bd39822415

PR Checklist | Your Answer
------------ | -------------
Have you deployed to Staging? |Not yet - deploying now!

### DIFF
--- a/services/QuillLMS/app/views/application/_head_embed_codes.html.erb
+++ b/services/QuillLMS/app/views/application/_head_embed_codes.html.erb
@@ -7,11 +7,11 @@
         integrations: {
           all: true
 
-          <%#= We're using load_intercom? as a proxy for Vitally
+          <%#= We're using user_is_trackable_teacher? as a proxy for Vitally
                since the conditions are basically the same (teachers only).
                If we shouldn't load Intercom, then we shouldn't send any
                events to Vitally, either. %>
-          <% unless load_intercom? %>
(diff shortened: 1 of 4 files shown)


services/QuillLMS/app/views/application/_head_embed_codes.html.erb
```diff
@@ -72,4 +72,5 @@
   </script>
   <!-- end Segment.io -->
   <%= render partial: 'application/coview' if load_intercom? %>
+  <%= render partial: 'application/ortto' if load_intercom? %>
```
anathomical: Is it worth renaming this function now that we use the same logic to gate multiple things?  Obviously not a high priority, but something like `user_is_trackable_teacher?` or something might reduce confusion about the exact stuff this condition is checking for.

anathomical (approved):
Minor thought that you could pursue, but this looks good to merge as-is to me.

happythenewsad (approved):

PR #9347, opened 2022-07-08
Suppress puts statements in rspec on circleci

brendanshean:
## WHAT
Prevent specs from printing to stdout.

## WHY
While puts statements are helpful when running rake tasks in development or production, while running tests on circleci they add unnecessary information which should be reserved for errors, warnings and deprecations.

## HOW
Add a configuration that stubs out these calls.
Various versions where proposed in this [thread](https://stackoverflow.com/a/57538689)

### Screenshots

### Notion Card Links

PR Checklist | Your Answer
------------ | -------------

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -56,6 +56,7 @@ jobs:
       - run:
           name: Run tests
           command: |
+            SUPPRESS_PUTS="true"
             cd services/QuillLMS
             TESTFILES=$(circleci tests glob "spec/**/*_spec.rb" | grep -v "spec/system/" | grep -v "/skip_ci/" | circleci tests split)
             bundle exec rspec -- ${TESTFILES}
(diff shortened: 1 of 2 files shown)


anathomical (approved):
Good solution to this.  I will note that this also suppresses deprecation warnings that might pop up during test runs.  Is that desirable?

happythenewsad (commented):
I'd only want this if it applied exclusively to circleCI. Otherwise, we'd have to  touch this line of code every time we're using print statements to debug in rspec.

brendanshean:
> Good solution to this. I will note that this also suppresses deprecation warnings that might pop up during test runs. Is that desirable?

@anathomical The deprecation warnings still show on this branch.  I was under the assumption they went to $stderr

brendanshean:
> I'd only want this if it applied exclusively to circleCI. Otherwise, we'd have to touch this line of code every time we're using print statements to debug in rspec.

@happythenewsad Good point.  I'll add something for circleci only.

happythenewsad (approved):

PR #9348, opened 2022-07-08
Address Rails 6.1 deprecations

brendanshean:
## WHAT
Fix some deprecations:

1.  Rails 6.1 will return Content-Type header without modification. If you want just the MIME type, please use `#media_type` instead. 
2.   Class level methods will no longer inherit [scoping](https://github.com/rails/rails/pull/32380) from `create` in Rails 6.1.
3.   `update_attributes` is deprecated and will be removed in Rails 6.1:  
4.  Shoulda [warns](https://github.com/thoughtbot/shoulda-matchers/blob/master/lib/shoulda/matchers/active_model/validate_inclusion_of_matcher.rb#L56) of` validates_inclusion_of` matching on booleans

Also, Set the parallelism on rails_evidence_build to 1.

## WHY
Deprecation messages pollute the circleci logs.  It's better to clean them up now if possible.

Also, evidence_rails_build has four nodes but they all test the same specs.  Since the suite is already really fast, it probably makes sense to just set parallelism in the config to 1 and save cycles in our quota for other builds.

Here's a screen shot of the four parallel runs running the same [specs](https://app.circleci.com/pipelines/github/empirical-org/Empirical-Core/16681/workflows/eb6ce1e2-f98e-4972-8730-4d5f8c1ca885/jobs/226645/parallel-runs/1?filterBy=ALL):
Screen Shot 2022-07-08 at 3 44 58 PM

## HOW
1.   In specs, use `media_type` instead of `content_type`.  Also, turbolinks needed to be bumped due to this [bug](https://github.com/rails/rails/issues/36989) 
2.  Add unscoped to the relevant queries except ones where default scope is expected.  For those, add `.default_scope`  
3.  This one is just a simple change from `update_attributes` to `update`
4.  Change the spec to check for appropriate values

### Screenshots

### Notion Card Links

PR Checklist | Your Answer
------------ | -------------

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -245,7 +245,7 @@ jobs:
             go test
   evidence_rails_build:
     working_directory: ~/Empirical-Core
-    parallelism: 4
+    parallelism: 1
     docker:
       - image: cimg/ruby:2.7.6
         environment:
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -56,10 +56,11 @@ jobs:
       - run:
           name: Run tests
           command: |
-            SUPPRESS_PUTS="true"
             cd services/QuillLMS
             TESTFILES=$(circleci tests glob "spec/**/*_spec.rb" | grep -v "spec/system/" | grep -v "/skip_ci/" | circleci tests split)
             bundle exec rspec -- ${TESTFILES}
+          environment:
+            SUPPRESS_PUTS: "true"
   lms_integration_build:
     working_directory: ~/Empirical-Core
     parallelism: 1
--- a/services/QuillLMS/app/models/invitation.rb
+++ b/services/QuillLMS/app/models/invitation.rb
@@ -18,7 +18,6 @@
 #  index_invitations_on_inviter_id     (inviter_id)
 #
 class Invitation < ApplicationRecord
-
   belongs_to :inviter, class_name: 'User', foreign_key: 'inviter_id'
   has_many :coteacher_classroom_invitations, dependent: :destroy
 
@@ -30,8 +29,6 @@ class Invitation < ApplicationRecord
   MAX_COTEACHER_INVITATIONS_PER_TIME = 50
   MAX_COTEACHER_INVITATIONS_PER_TIME_LIMIT_RESET_HOURS = 24
 
-
-
   def downcase_fields
     invitee_email.downcase!
   end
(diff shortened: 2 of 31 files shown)


.circleci/config.yml
```diff
@@ -245,7 +245,7 @@ jobs:
             go test
   evidence_rails_build:
     working_directory: ~/Empirical-Core
-    parallelism: 4
+    parallelism: 1
```
anathomical, line 249: Can you drop a note in the main PR description expanding on why we want to do this?  I don't object or anything, as it's a pretty small test suite, so I doubt it will have much real world impact on our development cycles, but I'd like to know why.

anathomical (approved):
One request for more information, but as you said in the description: this is pretty straight-forward.

.circleci/config.yml
brendanshean: @anathomical Sure thing, I've added a screenshot for more clarity.

happythenewsad (approved):

.circleci/config.yml
dandrabik, line 249: Nice catch, this should definitely just be 1.

services/QuillLMS/app/models/invitation.rb
```diff
@@ -40,7 +37,7 @@ def validate_invitation_limit
     # In order to avoid letting people use our platform to spam folks,
     # we want to put some limits on the number of invitations a user can issue.
     # One of those limits is a cap on invitations per user per time period
-    recent_invitation_count = Invitation.where(inviter_id: inviter_id, invitation_type: TYPES[:coteacher], created_at: MAX_COTEACHER_INVITATIONS_PER_TIME_LIMIT_RESET_HOURS.hours.ago..Time.current).count()
+    recent_invitation_count = Invitation.unscoped.where(inviter_id: inviter_id, invitation_type: TYPES[:coteacher], created_at: MAX_COTEACHER_INVITATIONS_PER_TIME_LIMIT_RESET_HOURS.hours.ago..Time.current).count()
```
dandrabik, line 40: I don't think I fully understand how this scope is polluted (no `default_scope`, no named scopes). Is this from a warning?

dandrabik (approved):

dandrabik (approved):
I don't think I fully understand why some of the `scope` changes are necessary, so I'd be curious to learn that (and why there aren't more changes needed in other models if the ones here are needed), but otherwise, looks good to me. Thanks for cleaning these up.

brendanshean:
> I don't think I fully understand why some of the `scope` changes are necessary, so I'd be curious to learn that (and why there aren't more changes needed in other models if the ones here are needed), but otherwise, looks good to me. Thanks for cleaning these up.

I spoke with @happythenewsad about this earlier.  I think the docs could probably be a little clearer.  What I gathered from the example at the top of this [PR](https://github.com/rails/rails/pull/32380) is the violations are only flagged when there's a Class level querying within another method that includes scope chaining.  There might not be any actual scoping leaking in, but more the potential for it (as is the case in the Invitation example).

services/QuillLMS/app/models/invitation.rb
brendanshean: This one is from a warning.   I'm guessing it's a false positive.

brendanshean:
@happythenewsad @dandrabik I forgot about this [thread](https://github.com/rails/rails/issues/38881#issuecomment-610004085) where they mentioned: "The reason why we deprecated was to avoid callbacks to inherit the scope".  

All four cases addressed in this PR, including Invitation rely on a callback that triggers this deprecation.  For example, If the  `validate :validate_invitation_limit` is removed and one just calls invitation.validate_invitation_limit, there warning goes away.

PR #9352, opened 2022-07-11
adding grade levels into activity library and other updates

emilia-friedberg:
## WHAT
Add grade levels into Activity Library and make other UI changes (adding ELL filters, updating how we display topics and readability information, updating mobile views).

## WHY
This is a new attribute we are using as a primary way for teachers to find content that is relevant for their students, so we want to showcase that (and make some additional tweaks).

## HOW
Update the activity search query to pass up grade level data, and then just a lot of React and a lot of CSS.

### Screenshots
Screen Shot 2022-07-12 at 2 28 57 PM
Screen Shot 2022-07-12 at 2 28 36 PM
Screen Shot 2022-07-12 at 12 17 11 PM
Screen Shot 2022-07-12 at 12 15 08 PM
Screen Shot 2022-07-12 at 12 14 52 PM

### Notion Card Links
https://www.notion.so/quill/Activity-library-updates-4da4250d44dc44169e9f0fff815e6cb1

PR Checklist | Your Answer
------------ | -------------


anathomical (approved):
Love this.

services/QuillLMS/client/app/bundles/Teacher/components/assignment_flow/create_unit/custom_activity_pack/activity_row.tsx
```diff
@@ -102,49 +98,76 @@ const ActivityRowConcept = ({ conceptName, }: { conceptName?: string }) => {
...
-    )
-  }
+const ActivityRowGradeRange = ({ minimumGradeLevel, maximumGradeLevel, gradeLevelFilters, }: { minimumGradeLevel?: number, maximumGradeLevel?: number, gradeLevelFilters: number[] }) => {
+  const className = "attribute-section grade-range"
 
-  return <span className={className} />
+  const lowestGradeLevelFilter = gradeLevelFilters[0]
+  const highestGradeLevelFilter = gradeLevelFilters[gradeLevelFilters.length - 1]
+
+  const gradeBands = ["4 - 5", "6 - 7", "8 - 9", "10 - 12"].map(gradeBand => {
```
eadams17, line 145: don't we have these grade bands constantized somewhere?

services/QuillLMS/client/app/bundles/Teacher/components/assignment_flow/create_unit/custom_activity_pack/grade_level_filters.tsx
```diff
@@ -0,0 +1,80 @@
...
+
+import NumberSuffixBuilder from '../../../../components/modules/numberSuffixBuilder'
+import { TwoThumbSlider, Tooltip, helpIcon, } from '../../../../../Shared/index'
+
+interface GradeLevelFiltersProps {
+  gradeLevelFilters: number[],
+  handleGradeLevelFilterChange: (gradeLevelFilters: number[]) => void,
+}
+
+const GRADE_LEVEL_LABELS = ['2-3', '4-5', '6-7', '8-9', '10-12']
```
eadams17, line 13: (In reference to my earlier comment) I think if we don't have these constantized elsewhere, it would be nice to have them defined since we do use them in multiple places

eadams17 (approved):
Looks great! Thanks for knocking this out. I just left one small question/comment about the grade band definitions

services/QuillLMS/client/app/bundles/Teacher/components/assignment_flow/create_unit/custom_activity_pack/grade_level_filters.tsx
emilia-friedberg: Yeah, so the readability grade bands (which are in fact the same as these, but only incidentally, which is to say they could be different so we don't necessarily want to share a constant), are defined as a constant in the shared file. The grade level labels we use here are not quite the same as the ones we use for display in the activity row itself (there are spaces around the hyphen and we omit 2-3), so we can't use the same constant for those. Since the labels aren't actually shared and since in both places they are used for display purposes, not business logic, I actually think it makes sense to not abstract them further than necessary. Let me know if you feel strongly though and we can talk about it more.

eadams17: That makes sense! Thanks for clarifying. No need to worry about it then

PR #9361, opened 2022-07-13
Activity versioning migrations

happythenewsad:
## WHAT
Migrations for the Evidence Activity Versioning project. 

## WHY
- Deploying the migrations first allows for a rollback-able final deploy, minimizing risk 
- avoids having to add throwaway backwards-compatability logic to the app code

## HOW
migrations 

### Screenshots

### Notion Card Links
https://www.notion.so/quill/Product-Board-30f97e4eb01246dbb8264253fb385073?v=388007d5669c4347b4215e24770ebbcd&p=f3e8454cd5f14b2ebda15dddd796fdf5

PR Checklist | Your Answer
------------ | -------------

### DIFF
--- a/services/QuillLMS/db/migrate/20220623214342_add_activity_version_to_feedback_history.rb
+++ b/services/QuillLMS/db/migrate/20220623214342_add_activity_version_to_feedback_history.rb
@@ -1,0 +1,5 @@
+class AddActivityVersionToFeedbackHistory < ActiveRecord::Migration[5.1]
+  def change
+    add_column :feedback_histories, :activity_version, :smallint, null: false, default: 0
+  end
+end
--- a/services/QuillLMS/db/migrate/20220623214342_add_activity_version_to_feedback_history.rb
+++ b/services/QuillLMS/db/migrate/20220623214342_add_activity_version_to_feedback_history.rb
@@ -1,6 +1,6 @@
 # frozen_string_literal: true
 
-class AddActivityVersionToFeedbackHistory < ActiveRecord::Migration[5.1]
+class AddActivityVersionToFeedbackHistory < ActiveRecord::Migration[6.0]
   def change
     add_column :feedback_histories, :activity_version, :smallint, null: false, default: 0
   end
--- a/services/QuillLMS/db/migrate/20220623214342_add_activity_version_to_feedback_history.rb
+++ b/services/QuillLMS/db/migrate/20220623214342_add_activity_version_to_feedback_history.rb
@@ -1,3 +1,5 @@
+# frozen_string_literal: true
+
 class AddActivityVersionToFeedbackHistory < ActiveRecord::Migration[5.1]
   def change
     add_column :feedback_histories, :activity_version, :smallint, null: false, default: 0
--- a/services/QuillLMS/db/migrate/20220708201219_add_version_to_evidence_activity.evidence.rb
+++ b/services/QuillLMS/db/migrate/20220708201219_add_version_to_evidence_activity.evidence.rb
@@ -1,0 +1,6 @@
+# This migration comes from evidence (originally 20220623205532)
+class AddVersionToEvidenceActivity < ActiveRecord::Migration[5.1]
+  def change
+    add_column :comprehension_activities, :version, :smallint, null: false, default: 0
+  end
+end
--- a/services/QuillLMS/db/migrate/20220708201219_add_version_to_evidence_activity.evidence.rb
+++ b/services/QuillLMS/db/migrate/20220708201219_add_version_to_evidence_activity.evidence.rb
@@ -1,3 +1,5 @@
+# frozen_string_literal: true
+
 # This migration comes from evidence (originally 20220623205532)
 class AddVersionToEvidenceActivity < ActiveRecord::Migration[5.1]
   def change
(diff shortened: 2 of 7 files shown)


services/QuillLMS/engines/evidence/spec/models/evidence/activity_spec.rb
```diff
@@ -32,6 +32,21 @@ module Evidence
...
+        end
+
+        it 'a non-monotonically increasing version should be invalid' do
+          activity = create(:evidence_activity)
+          expect do
+            activity.update!({ version: activity.version + 2 })
+          end.to raise_error(ActiveRecord::RecordInvalid)
+        end
+      end
+
```
anathomical: These tests look great.  Is it worth having some kind of test around the `before_validation` hook?

services/QuillLMS/engines/evidence/app/models/evidence/activity.rb
```diff
@@ -18,6 +18,7 @@ class Activity < ApplicationRecord
 
     before_destroy :expire_turking_rounds
     before_validation :set_parent_activity, on: :create
+    before_validation :increment_version, on: :update
```
anathomical: Is there any worry around this causing version changes when it's not a meaningful change?  Like, tweaking the text of the title or changing the value in the `notes` attribute?  I can imagine a scenario where Curriculum changes the ML model (and we want a version update), then a week later tweaks the title, and suddenly we have a discontinuity where the grading data before and after the title change should be grouped together, but aren't.

It's possible that the rest of the code that will come after this migration addresses the issue, in which case you can ignore this concern, but I wanted to raise it just in case.

anathomical (changes requested):
A couple of minor questions about approach here, but I suspect they can be worked out via discussion.

services/QuillLMS/db/migrate/20220623214342_add_activity_version_to_feedback_history.rb
```diff
@@ -0,0 +1,5 @@
+class AddActivityVersionToFeedbackHistory < ActiveRecord::Migration[5.1]
```
brendanshean: We should probably update these from 5.1 -> 6.0

services/QuillLMS/db/migrate/20220708201219_add_version_to_evidence_activity.evidence.rb
```diff
@@ -0,0 +1,6 @@
+# This migration comes from evidence (originally 20220623205532)
+class AddVersionToEvidenceActivity < ActiveRecord::Migration[5.1]
```
brendanshean: We should probably update these from 5.1 -> 6.0

services/QuillLMS/engines/evidence/db/migrate/20220623205532_add_version_to_evidence_activity.rb
```diff
@@ -0,0 +1,5 @@
+class AddVersionToEvidenceActivity < ActiveRecord::Migration[5.1]
```
brendanshean: We should probably update these from 5.1 -> 6.0

services/QuillLMS/db/migrate/20220623214342_add_activity_version_to_feedback_history.rb
```diff
@@ -0,0 +1,7 @@
+# frozen_string_literal: true
+
+class AddActivityVersionToFeedbackHistory < ActiveRecord::Migration[5.1]
+  def change
+    add_column :feedback_histories, :activity_version, :smallint, null: false, default: 0
```
brendanshean, line 5: Really just a preference thing, but I notice all of the `default: 0`.  Would `default: 1` make more sense semantically to curriculum?

brendanshean (commented):
I left some minor comments.  So far, I like the lightweight approach here as opposed to reaching for papertrail.

services/QuillLMS/engines/evidence/app/models/evidence/activity.rb
brendanshean: I'd like to hear more about this too.

services/QuillLMS/db/migrate/20220623214342_add_activity_version_to_feedback_history.rb
happythenewsad: This was a requirement in the spec. `0` denotes "all versions before activity versioning was rolled out."

services/QuillLMS/db/migrate/20220623214342_add_activity_version_to_feedback_history.rb
happythenewsad: Addressed

services/QuillLMS/engines/evidence/app/models/evidence/activity.rb
happythenewsad: Good point. I removed this logic from the `update` callback. We will increment via a single UX and API endpoint, so every version bump occurs when a human decides.

services/QuillLMS/engines/evidence/spec/models/evidence/activity_spec.rb
happythenewsad: n/a - I removed this hook

anathomical (approved):
Thanks for the quick turn-around.

services/QuillLMS/engines/evidence/app/models/evidence/activity.rb
brendanshean: Awesome. I was going to suggest the same thing.

brendanshean (approved):

PR #9377, opened 2022-07-19
Lms/concept result migration feedback

anathomical:
## WHAT
Refactor the batch migration job for `ConceptResult` records so that failing batches fan out into individual migration jobs
## WHY
These batches are failing because specific rows have problematic data in them.  By un-batching the migrations, we can work our way down to the specific failing rows and address those issues on an individual basis without worrying about the surrounding records.
## HOW
- Create a new worker that migrates individual `OldConceptResult` rows to `ConceptResult` records
- Refactor the batch-migrator to enqueue its records into this new worker

PR Checklist | Your Answer
------------ | -------------
Have you deployed to Staging? | There's no locked-up data on Staging, so I'm deploying now to make sure it still works, but can't easily test this in staging

### DIFF
--- a/services/QuillLMS/app/workers/copy_old_concept_results_to_concept_results_worker.rb
+++ b/services/QuillLMS/app/workers/copy_old_concept_results_to_concept_results_worker.rb
@@ -4,55 +4,9 @@ class CopyOldConceptResultsToConceptResultsWorker
   include Sidekiq::Worker
   sidekiq_options queue: SidekiqQueue::MIGRATION
 
-  class ConceptResultMigrationDeadlocked < StandardError; end
-
-  BATCH_SIZE = 1_000
-
-  # rubocop:disable Metrics/CyclomaticComplexity
   def perform(start, finish)
-    directions_cache = {}
-    instructions_cache = {}
-    previous_feedbacks_cache = {}
-    prompts_cache = {}
-    question_types_cache = {}
-
-    ConceptResult.bulk_insert(ignore: true) do |bulk_inserter|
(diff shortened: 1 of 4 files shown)


brendanshean (approved):
Looks good to me.

services/QuillLMS/app/workers/copy_old_concept_results_to_concept_results_worker.rb
```diff
@@ -4,55 +4,9 @@ class CopyOldConceptResultsToConceptResultsWorker
...
-          concept_result_instructions_id: instructions_cache[instructions],
-          concept_result_previous_feedback_id: previous_feedbacks_cache[previous_feedback],
-          concept_result_prompt_id: prompts_cache[prompt],
-          concept_result_question_type_id: question_types_cache[question_type]
-        })
-      end
+    (start..finish).each do |old_concept_result_id|
+      CopySingleConceptResultWorker.perform_async(old_concept_result_id)
     end
-  rescue ActiveRecord::Deadlocked => e
```
brendanshean, line 54: Do we need to rescue anything in the individual case or were deadlocks more of an issue due to the multiple record inserts?

anathomical: The latter.  While I'm not a total expert, my understanding is this: the old way of doing this involved doing bulk inserts, and those inserts took place within explicit transaction blocks.  Deadlocks occur when two transactions make changes that the database doesn't know how to reconcile with each other (often when it's trying to assign large numbers of `id` values).  Because this approach only does single row inserts (via `create!`), and doesn't invoke transactions within the database, we should be able to avoid deadlocks.

PR #9378, opened 2022-07-19
Vitally organizations

anathomical:
## WHAT
- Sync data to Vitally when `SalesFormSubmission` records get created
- Allow Intercom to send sales opportunities to our system in the form of `SalesFormSubmission` records
- Sync `District` records to Vitally as `Organizations`
## WHY
This should allow us to do almost all customer journey management in Vitally
## HOW
CISSY MAGIC (that is to say, Cissy wrote all the code and most of the tests before she left, I finished about a dozen `NotImplemented` tests).
- Modify the `SalesFormSubmission` model to do a bunch of Vitally API calls `after_save` so that those opportunities sync to Vitally
- Add a new webhook controller that accepts data from Intercom, validates that it's cryptographically signed with our shared key, and then creates `SalesFormSubmission` records based on the payload
- Add logic to the `SyncVitallyWorker` to figure out which `Districts` we should sync to Vitally and then enqueue those `Districts` for upload to Vitally via the `SyncVitallyOrganizations` worker.  (We call them Districts, and Vitally calls them Organizations.)

### Notion Card Links
https://www.notion.so/quill/Move-the-sales-opportunity-pipeline-from-Streak-to-Vitally-project-8f04ef91406844c5a3746add11304e0c

PR Checklist | Your Answer
------------ | -------------


services/QuillLMS/app/services/vitally_integration/vitally_rest_api.rb
```diff
@@ -0,0 +1,46 @@
...
+    post(type, payload)
+  end
+
+  def exists?(type, id)
+    get(type, id).try(:error).blank?
+  end
+
+  def get(type, id)
+    HTTParty.get("#{VITALLY_REST_API_BASE_URL}/#{type}/#{id}",
+      headers: {
```
brendanshean: I would refactor headers for the three methods

```
private def headers
   {  Authorization: "Basic #{@api_key}",  "Content-Type": "application/json" }
end
```

services/QuillLMS/app/views/pages/home_new.html.erb
```diff
@@ -415,7 +415,7 @@ Quill Grammar has over 150 sentence writing activities to help your students. Ou
     <p class="premium-header">Quill Premium</p>
     <p class="premium-text">Quill Premium's advanced reporting features are the best way to support teachers at the school or district level.</p>
     <div class="premium-button-container">
-      <a class="quill-button medium primary contained sales-button" href="https://quillpremium.wufoo.com/forms/quill-premium-quote/" rel="noopener noreferrer" target="_blank">Contact sales</a>
+      <a class="quill-button medium primary contained sales-button" href="https://quill.org/request_quote" rel="noopener noreferrer" target="_blank">Contact sales</a>
```
brendanshean: I would use path helpers here instead of hardcoded urls:

```
      <a class="quill-button medium primary contained sales-button" href=<%= request_quote_path %>  noreferrer" target="_blank">Contact sales</a>
      <a class="quill-button medium secondary outlined" href=<%= premium_path %> target="_blank">Learn more</a>
```

services/QuillLMS/app/services/vitally_integration/vitally_rest_api.rb
```diff
@@ -0,0 +1,46 @@
...
+  def initialize
+    @api_key = ENV['VITALLY_REST_API_KEY']
+  end
+
+  def create(type, payload)
+    post(type, payload)
+  end
+
+  def exists?(type, id)
+    get(type, id).try(:error).blank?
```
brendanshean: I tend to use safe navigation over `try` due to [performance](https://jetrockets.com/blog/nswsheigth-safe-navigation-vs-try-in-rails-part-2-performance).

I realize they are not quite equivalent but just wondering.

services/QuillLMS/app/services/vitally_integration/vitally_rest_api.rb
```diff
@@ -0,0 +1,46 @@
...
+
+class VitallyRestApi
+  VITALLY_REST_API_BASE_URL = 'https://rest.vitally.io/resources'
+
+  def initialize
+    @api_key = ENV['VITALLY_REST_API_KEY']
+  end
+
+  def create(type, payload)
+    post(type, payload)
```
brendanshean: what about removing the private post method and just moving it into create?

services/QuillLMS/app/workers/sync_vitally_worker.rb
```diff
@@ -28,10 +31,14 @@ def perform
   # rubocop:enable Metrics/CyclomaticComplexity
 
   def schools_to_sync
-    School.select(:id).distinct.joins(:users).where('users.role = ?', 'teacher')
+    @schools_to_sync ||= School.select(:id, :district_id).distinct.joins(:users).where('users.role = ?', 'teacher').where.not(district_id: nil)
```
brendanshean: Do we want the `where.not(district_id: nil)`?  That makes sense for the district case, but  `.compact` will handle that case in the `districts_to_sync`.

services/QuillLMS/client/app/bundles/Teacher/components/subscriptions/current_subscription.jsx
```diff
@@ -40,7 +40,7 @@ export default class CurrentSubscription extends React.Component {
       return <span>{baseText} To prevent your subscription from expiring, contact the purchaser at <a href={`mailto:${purchaser_email}`}>{purchaser_email}</a> and ask them to turn on automatic renewal.</span>
     }
     if (payment_method !== CREDIT_CARD && (authorityLevel || !purchaser_email)) {
-      return <span>{baseText} To renew your subscription for next year, <a href="mailto:sales@quill.org">contact us</a>.</span>
+      return <span>{baseText} To renew your subscription for next year, <a href="https://quill.org/request_renewal">contact us</a>.</span>
```
brendanshean: I would extract `https://quill.org/request_renewal` into a prop and specify `requestQuotePath: request_quote_path` [here](https://github.com/empirical-org/Empirical-Core/blob/fd7c4a8a9d5840e61aba12273337b108475adcdc/services/QuillLMS/app/views/pages/shared/_premium_main.html.erb#L22)

services/QuillLMS/client/app/bundles/Teacher/components/modules/questionsAndAnswers/premium.jsx
```diff
@@ -15,7 +15,7 @@ const premium = () => [
   },
   {
     question: 'How much does premium cost?',
-    answer: <p>Quill Premium is provided as a teacher license, school license, or district license. The teacher license covers all of the students for one teacher and costs $80 per year. The school license is a site-wide license that covers all teachers and students in the school. For districts, we provide custom pricing, on-site training, and district dashboards. <a href="https://quillpremium.wufoo.com/forms/quill-premium-quote/">Contact us to receive a demo.</a></p>
+    answer: <p>Quill Premium is provided as a teacher license, school license, or district license. The teacher license covers all of the students for one teacher and costs $80 per year. The school license is a site-wide license that covers all teachers and students in the school. For districts, we provide custom pricing, on-site training, and district dashboards. <a href="https://quill.org/request_quote">Contact us to receive a demo.</a></p>
```
brendanshean, line 18: I would extract `https://quill.org/request_quote` into a prop and specify `requestQuotePath: request_quote_path` [here](https://github.com/empirical-org/Empirical-Core/blob/fd7c4a8a9d5840e61aba12273337b108475adcdc/services/QuillLMS/app/views/pages/shared/_premium_main.html.erb#L22)

services/QuillLMS/client/app/bundles/Teacher/components/premium/premium_minis/school_pricing_mini.jsx
```diff
@@ -16,6 +16,7 @@ const SchoolPricingMini = ({ plan, premiumFeatureData, showBadges, handleClickPu
       <div className="premium-button-container">
         <button
           className="quill-button contained medium primary focus-on-light"
+          href="https://quill.org/request_quote"
```
brendanshean, line 19: I would use the prop specified earlier.

services/QuillLMS/app/models/user.rb
```diff
@@ -176,6 +176,26 @@ class User < ApplicationRecord
 
   validate :validate_flags
 
+  TEACHER = 'teacher'
```
brendanshean: There's some duplication here with constants.  They are defined higher up in the file.

services/QuillLMS/app/controllers/cms/districts_controller.rb
```diff
@@ -48,10 +48,13 @@ def edit_subscription
 
   def create
     new_district = District.new(district_params)
-    if new_district.save
+    begin
```
brendanshean: I feel like begin/rescue exception handling is being [overused](https://dalibornasevic.com/posts/52-don-t-overuse-exceptions) here.  Some if/else conditionals would probably suffice.

services/QuillLMS/app/controllers/intercom_integration/webhooks_controller.rb
```diff
@@ -0,0 +1,94 @@
...
+    private def verify_signature
+      client_secret = ENV.fetch('INTERCOM_APP_SECRET', '')
+      hexdigest = OpenSSL::HMAC.hexdigest('sha1', client_secret, payload)
+      return if request.headers['X-Hub-Signature'] == "sha1=#{hexdigest}"
+
+      raise "unauthorized call of Intercom webhook"
+    end
+
+    private def find_or_create_user
+      user = User.find_by(id: user_payload["user_id"]) if !user_payload["anonymous"]
```
brendanshean: `unless` might be easier to read.

services/QuillLMS/app/controllers/intercom_integration/webhooks_controller.rb
```diff
@@ -0,0 +1,94 @@
...
+    private def user_payload
+      @user_payload ||= parsed_payload["data"]["item"]["user"] || parsed_payload["data"]["item"]["contact"]
+    end
+
+    private def verify_signature
+      client_secret = ENV.fetch('INTERCOM_APP_SECRET', '')
+      hexdigest = OpenSSL::HMAC.hexdigest('sha1', client_secret, payload)
+      return if request.headers['X-Hub-Signature'] == "sha1=#{hexdigest}"
+
+      raise "unauthorized call of Intercom webhook"
```
brendanshean: Should this be extracted into an Error class?

services/QuillLMS/spec/workers/sync_vitally_worker_spec.rb
```diff
@@ -8,12 +8,39 @@
   describe "#perform" do
     it "make queries for schools and users and enqueue them for further jobs" do
       ENV['SYNC_TO_VITALLY'] = 'true'
```
brendanshean: All the `ENV['SYNC_TO_VITALLY']` calls in these tests should be removed since they can leak state.
I would opt a single  `before { stub_const('ENV', {'SYNC_TO_VITALLY' => 'true'}) }` that applies to all of the specs instead.

brendanshean (commented):
I've left some comments, but I'm still only halfway through the PR.

services/QuillLMS/app/workers/sync_vitally_worker.rb
anathomical: I didn't write that addition, but based on my review of the specs, I think that you're right.  I'll remove the condition.

services/QuillLMS/app/controllers/cms/districts_controller.rb
anathomical: Yeah, that makes sense to me.  I'll update.

services/QuillLMS/app/controllers/intercom_integration/webhooks_controller.rb
anathomical: Yeah, that seems smart.

services/QuillLMS/app/services/vitally_integration/vitally_rest_api.rb
anathomical: So it turned out that the `try` was attempting to cover an edge case where `HTTParty` might sometimes have access to an `error` attribute and sometimes not.  But it turns out that's not how `HTTParty` works, so I ended up reworking this line.

anathomical:
@brendanshean I think I've addressed all of the stuff you called out in you pass so far.

brendanshean:
> @brendanshean I think I've addressed all of the stuff you called out in you pass so far.

Thank you @anathomical.  Looks great.  I will take a look at the rest of the PR.

services/QuillLMS/app/services/vitally_integration/vitally_rest_api.rb
brendanshean: Thank you for doing a deeper investigation.

services/QuillLMS/app/models/district.rb
```diff
@@ -27,6 +27,7 @@ class District < ApplicationRecord
   include Subscriber
 
   validates :name, presence: true
+  validates :nces_id, uniqueness: true
```
brendanshean: Good call on validation here.

services/QuillLMS/spec/controllers/sales_form_submission_controller_spec.rb
```diff
@@ -5,9 +5,21 @@
 describe SalesFormSubmissionController, type: :controller do
 
   describe '#create' do
-    let!(:sales_form_submission) { create(:sales_form_submission) }
+    let!(:school) { create(:school) }
+
+    before do
```
brendanshean: Are these three stubs necessary?

services/QuillLMS/spec/controllers/sales_form_submission_controller_spec.rb
```diff
@@ -5,9 +5,21 @@
...
+      allow(HTTParty).to receive(:get)
+      allow(HTTParty).to receive(:put)
+      allow(HTTParty).to receive(:post)
+    end
 
     it 'should create sales_form_submission entry' do
+      sales_form_submission = double
+      allow(SalesFormSubmission).to receive(:new).and_return(sales_form_submission)
+      allow(sales_form_submission).to receive(:save!).and_return(true)
+      allow(sales_form_submission).to receive(:send_opportunituy_to_vitally)
```
brendanshean: Looks like there's a typo `:send_opportunituy_to_vitally`

services/QuillLMS/spec/controllers/sales_form_submission_controller_spec.rb
```diff
@@ -5,9 +5,21 @@
...
+
+      expect(sales_form_submission).to receive(:save!)
```
brendanshean: Were lines 19 and 22 what we discussed at the last engineering meeting?

services/QuillLMS/spec/controllers/intercom_integration/webhooks_controller_spec.rb
```diff
@@ -0,0 +1,96 @@
...
+    end
+
+    context 'authentication not valid' do
+
+      # Basically, since we're not providing signed headers, this will be invalid
+      # by default
+      it 'raises error' do
+        expect { post :create }
+          .to raise_error("unauthorized call of Intercom webhook")
+      end
```
brendanshean: I would use the exception class that you have now defined:

`it { expect { post :create }.to raise_error(described_class::UnauthorizedIntercomWebhookCallError) }`

services/QuillLMS/app/models/sales_form_submission.rb
```diff
@@ -30,18 +32,192 @@ class SalesFormSubmission < ApplicationRecord
...
-  validates :zipcode, presence: true
-  validates :school_name, presence: true, unless: :district_name
-  validates :district_name, presence: true, unless: :school_name
   validates :school_premium_count_estimate, presence: true, numericality: { greater_than_or_equal_to: 0 }
   validates :teacher_premium_count_estimate, presence: true, numericality: { greater_than_or_equal_to: 0 }
   validates :student_premium_count_estimate, presence: true, numericality: { greater_than_or_equal_to: 0 }
   validates :collection_type, presence: true, inclusion: { in: COLLECTION_TYPES }
   validates :submission_type, presence: true, inclusion: { in: SUBMISSION_TYPES }
+
+  after_save :vitally_callbacks
```
brendanshean, line 61: I tend to avoid after_save callbacks except those managing state internal to the object.  (e.g. see this [post](https://samuelmullen.com/2013/05/the-problem-with-rails-callbacks/)).  Maybe discuss offline pros/cons?

brendanshean (commented):
Some more comments

services/QuillLMS/client/app/bundles/Teacher/components/subscriptions/current_subscription.jsx
eadams17: we'd want `rel="noopener noreferrer" target="_blank"` for this link

services/QuillLMS/app/models/sales_form_submission.rb
anathomical: For posterity: Brendan and I had a call about this and basically decided to compromise.  We want the `after_save` because we want to guarantee that this data syncs to Vitally no matter how we end up creating these records.  However, we also definitely don't want to tightly couple the ability to save models to external APIs (which is what happens if you do this synchronously).  So we're keeping `after_save`, but moving the sync work into a Sidekiq worker that's triggered by said `after_save` hook.

services/QuillLMS/client/app/bundles/Teacher/components/subscriptions/current_subscription.jsx
```diff
@@ -72,7 +72,7 @@ export default class CurrentSubscription extends React.Component {
     return (
       <span>
         To renew your subscription for next year, contact us now
-        at <a href="mailto:sales@quill.org">sales@quill.org</a>.
+        at <a href="https://quill.org/request_renewal">sales@quill.org</a>.
       </span>
```
eadams17, line 76: I think we'd probably want to change this to maybe have "contact us now" be the displayed link text (and also add `rel="noopener noreferrer" target="_blank"`)

eadams17 (approved):
looks good! just a few small comments around links

services/QuillLMS/db/migrate/20220609175032_change_sales_form_submission_columns.rb
```diff
@@ -0,0 +1,10 @@
+# frozen_string_literal: true
+
+class ChangeSalesFormSubmissionColumns < ActiveRecord::Migration[5.1]
+  def change
+    change_column_null :sales_form_submissions, :phone_number, true
+    change_column_null :sales_form_submissions, :zipcode, true
+    change_column_null :sales_form_submissions, :school_name, true
+    change_column_null :sales_form_submissions, :district_name, true
```
happythenewsad: school_name and district_name don't have the non-null constraint, so lines 7,8 should be removed because they are redundant.

services/QuillLMS/spec/models/sales_form_submission_spec.rb
```diff
@@ -11,48 +11,77 @@
...
 
-  context 'school name validation with blank district name' do
-    let(:sales_form_submission) { build(:sales_form_submission, district_name: '') }
+  context 'after_save, should perform this behavior on internal User records' do
+    let(:school) { create(:school)}
 
-    it { expect(sales_form_submission).to be_valid }
-    it { should validate_presence_of(:school_name) }
-    it { should validate_presence_of(:district_name) }
+    #before do
```
happythenewsad: remove comments

services/QuillLMS/spec/models/user_spec.rb
```diff
@@ -346,8 +346,8 @@
 
   describe 'constants' do
     it "should give the correct value for all the contstants" do
```
happythenewsad: typo

happythenewsad (changes requested):

services/QuillLMS/spec/models/sales_form_submission_spec.rb
```diff
@@ -11,48 +11,77 @@
...
+    it 'should enqueue jobs after save' do
+      sales_form_submission = create(:sales_form_submission)
 
-    it { expect(sales_form_submission).to be_valid }
-    it { should validate_presence_of(:school_name) }
-    it { should validate_presence_of(:district_name) }
+      expect(SyncSalesFormSubmissionToVitallyWorker).to receive(:perform_async).with(sales_form_submission.id)
+      expect(SendSalesFormSubmissionToSlackWorker).to receive(:perform_async).with(sales_form_submission.id)
+
+      sales_form_submission.created_at = Time.now
```
brendanshean: I'm not sure why the linter didn't catch this but we should use Time.current over Time.now.   Also, what's the reasoning behind updating the created_at here.

brendanshean (approved):
I think the refactoring after_save to use workers looks much better.  Thanks for finishing the last mile of this large PR.

services/QuillLMS/app/controllers/cms/districts_controller.rb
```diff
@@ -48,10 +48,11 @@ def edit_subscription
 
   def create
     new_district = District.new(district_params)
-    if new_district.save
+    if new_district.valid?
```
dandrabik: I think this was actually better as `if new_district.save` vs. `.valid?` and then `save!`

services/QuillLMS/app/controllers/cms/districts_controller.rb
```diff
@@ -48,10 +48,11 @@ def edit_subscription
...
+      new_district.save!
       redirect_to cms_district_path(new_district)
     else
-      redirect_to cms_districts_path, error: new_district.errors, flash: { error: new_district.errors }
+      redirect_to cms_districts_path, flash: { error: "A district with this NCES ID already exists." }
```
dandrabik: Are we sure this is what the error would be? Seems like the previous approach gave more specific errors.

services/QuillLMS/app/controllers/intercom_integration/webhooks_controller.rb
```diff
@@ -0,0 +1,96 @@
...
+  class WebhooksController < ApplicationController
+    UNKNOWN_NAME = "Unknown User"
+
+    class UnauthorizedIntercomWebhookCallError < StandardError; end
+
+    skip_before_action :verify_authenticity_token
+    before_action :verify_signature, only: [:create]
+
+    def create
+      topic = parsed_payload["topic"]
```
dandrabik: Is there a reason we're not using standard Rails `params` here?

services/QuillLMS/app/controllers/intercom_integration/webhooks_controller.rb
```diff
@@ -0,0 +1,96 @@
...
+          email: user.email,
+          phone_number: user_payload["phone"],
+          zipcode: user_payload["location_data"]["postal_code"],
+          school_name: user&.school&.name,
+          district_name: user&.school&.district&.name,
+          source: SalesFormSubmission::INTERCOM_SOURCE,
+          intercom_link: intercom_link(user_payload["id"])
+        )
+        sales_form_submission.collection_type = collection_type(tag)
+        sales_form_submission.submission_type = submission_type(tag)
```
dandrabik: Could we move these two lines inside the `.new`, . e.g `(collection_type: collection_type(tag), submission_type = submission_type(tag))`

services/QuillLMS/app/controllers/intercom_integration/webhooks_controller.rb
```diff
@@ -0,0 +1,96 @@
...
+
+    skip_before_action :verify_authenticity_token
+    before_action :verify_signature, only: [:create]
+
+    def create
+      topic = parsed_payload["topic"]
+      if ["user.tag.created", "contact.tag.created"].include?(topic)
+        tag = parsed_payload["data"]["item"]["tag"]["name"]
+        user = find_or_create_user
+        sales_form_submission = SalesFormSubmission.new(
```
dandrabik: If you add `collection_type` and `submission_type` as params, you can change this to `.create!` and remove the `.save!`. Also, `sales_form_submission` isn't used so you don't need to assign.

services/QuillLMS/app/controllers/intercom_integration/webhooks_controller.rb
```diff
@@ -0,0 +1,96 @@
...
+    UNKNOWN_NAME = "Unknown User"
+
+    class UnauthorizedIntercomWebhookCallError < StandardError; end
+
+    skip_before_action :verify_authenticity_token
+    before_action :verify_signature, only: [:create]
+
+    def create
+      topic = parsed_payload["topic"]
+      if ["user.tag.created", "contact.tag.created"].include?(topic)
```
dandrabik: Maybe make this this a constant to help give context, e.g. `ACCEPTED_WEBHOOK_TOPICS.include?(topic)` or whatever it should be called.

services/QuillLMS/app/controllers/intercom_integration/webhooks_controller.rb
```diff
@@ -0,0 +1,96 @@
...
+        sales_form_submission.save!
+      end
+      head 200
+    end
+
+    private def intercom_link(intercom_user_id)
+      "https://app.intercom.com/a/apps/v2ms5bl3/users/#{intercom_user_id}/all-conversations"
+    end
+
+    private def payload
```
dandrabik, line 61: Yeah, let's definitely use Rails `params` instead of this custom code.

services/QuillLMS/app/controllers/intercom_integration/webhooks_controller.rb
```diff
@@ -0,0 +1,96 @@
...
+      return user if user.present?
+
+      user = User.find_by(email: user_payload["email"])
+      return user if user.present?
+
+      User.create!(
+        email: user_payload["email"],
+        role: User::SALES_CONTACT,
+        name: user_payload["name"] || UNKNOWN_NAME,
+        password: SecureRandom.uuid
```
dandrabik, line 90: So we are generating a password for them that they won't know?

services/QuillLMS/app/controllers/intercom_integration/webhooks_controller.rb
```diff
@@ -0,0 +1,96 @@
...
+      )
+    end
+
+    private def collection_type(tag)
+      case tag
+      when "Quote Request School", "Renewal Request School"
```
dandrabik: Strings like `"Quote Request School"` are used multiple times for logic, let's make them constants.

services/QuillLMS/app/controllers/intercom_integration/webhooks_controller.rb
```diff
@@ -0,0 +1,96 @@
...
+        SalesFormSubmission::SCHOOL_COLLECTION_TYPE
+      when "Quote Request District", "Renewal Request District"
```
dandrabik: For these type of lookups, consider using a hash constant and using `.fetch(key, default)`, e.g.

```ruby
COLLECTION_MAPPING = {
   QUOTE_REQUEST => SalesFormSubmission::DISTRICT_COLLECTION_TYPE,
   RENEWAL_REQUEST => SalesFormSubmission::DISTRICT_COLLECTION_TYPE,
...
}

File: services/QuillLMS/app/services/vitally_integration/vitally_rest_api.rb [hunk 35]
```diff
@@ -0,0 +1,40 @@
+# frozen_string_literal: true
+
+class VitallyRestApi
+  VITALLY_REST_API_BASE_URL = 'https://rest.vitally.io/resources'
+
+  def initialize
+    @api_key = ENV['VITALLY_REST_API_KEY']
```
dandrabik: This can be a constant and you can drop the `initialize` method. e.g. `API_KEY = ENV['VITALLY_REST_API_KEY']`

services/QuillLMS/app/workers/sync_sales_form_submission_to_vitally_worker.rb
```diff
@@ -0,0 +1,107 @@
...
+  def create_records_in_vitally
+    create_school_or_district_if_none_exist
+    create_vitally_user_if_none_exists
+    send_opportunity_to_vitally
+  end
+
+  def create_school_or_district_if_none_exist
+    if (@sales_form_submission.district_collection? &&
+        @sales_form_submission.district.present? &&
+        !api.exists?(SalesFormSubmission::VITALLY_DISTRICTS_TYPE, @sales_form_submission.district.id))
```
dandrabik: Consider a `create_unless_exists` method for the API wrapper, e.g.

```ruby
def create_unless_exists(type, data)
  return if exists?(type, data)

  create(type, data)
end

File: services/QuillLMS/app/workers/sync_sales_form_submission_to_vitally_worker.rb [hunk 37]
```diff
@@ -0,0 +1,107 @@
...
+  end
+
+  def create_records_in_vitally
+    create_school_or_district_if_none_exist
+    create_vitally_user_if_none_exists
+    send_opportunity_to_vitally
+  end
+
+  def create_school_or_district_if_none_exist
+    if (@sales_form_submission.district_collection? &&
```
dandrabik: I think the logic here is slightly off and the `api.exists?` shouldn't be in the branching checks.

Code as is:
- If it has district data AND that district IS in vitally, it will try next to create a school
- If it has district data AND that district IS NOT in vitally, it will create a district, but NOT try to create a school

services/QuillLMS/app/workers/sync_sales_form_submission_to_vitally_worker.rb
```diff
@@ -0,0 +1,107 @@
...
+    user = @sales_form_submission.find_or_create_user
+
+    if api.exists?(SalesFormSubmission::VITALLY_USERS_TYPE, user.id)
+      api.update(SalesFormSubmission::VITALLY_USERS_TYPE, user.id, vitally_user_update_data)
+    else
+      api.create(SalesFormSubmission::VITALLY_USERS_TYPE, vitally_user_create_data)
+    end
+  end
+
+  def send_opportunity_to_vitally
```
dandrabik, line 40: Should this be a `create_unless_exists`?

services/QuillLMS/spec/controllers/accounts_controller_spec.rb
```diff
@@ -115,6 +115,36 @@
           expect(response.body).to eq({errors: {email: ["Enter a valid email"]}}.to_json)
         end
       end
+
+      context 'when user already exists in db as a sales contact' do
+        let!(:user) { create(:user, role: User::SALES_CONTACT) }
+
+        it 'should use that same user record but update all the fields' do
+          name = "Test Name"
```
dandrabik: Consider making these `let` vars since they are in multiple tests.

services/QuillLMS/spec/services/vitally_integration/vitally_rest_api_spec.rb
```diff
@@ -0,0 +1,90 @@
+# frozen_string_literal: true
+
+require 'rails_helper'
+
+describe VitallyRestApi do
+  let(:api)  { VitallyRestApi.new }
+
+  before do
+    stub_const('ENV', {'VITALLY_REST_API_KEY' => 'test api key'})
```
dandrabik: When we make API_KEY a class constant, let's stub that instead to avoid touching the global space.

services/QuillLMS/spec/services/vitally_integration/vitally_rest_api_spec.rb
```diff
@@ -0,0 +1,90 @@
...
+  end
+
+  describe '#create' do
+    it 'should make a POST call to the Vitally API with the specified command and payload' do
+      payload = 'test payload'
+      expect(HTTParty).to receive(:post).with("#{VitallyRestApi::VITALLY_REST_API_BASE_URL}/type",
+        headers: {
+          Authorization: "Basic #{ENV['VITALLY_REST_API_KEY']}",
```
dandrabik: I think you'd want to test that this is `Authorization: "Basic 'test api key'"`, i.e. if you are stubbing it, you should test that the expected stub is coming through.

services/QuillLMS/spec/services/vitally_integration/vitally_rest_api_spec.rb
```diff
@@ -0,0 +1,90 @@
...
+        },
+        body: payload.to_json
+      )
+      api.create('type', payload)
+    end
+  end
+
+  describe '#get' do
+    it 'should make a GET call to the Vitally API with the specified type and ID' do
+      id = 1
```
dandrabik: Vars that could be reusable `let`s.

services/QuillLMS/spec/workers/send_sales_form_submission_to_slack_worker_spec.rb
```diff
@@ -0,0 +1,21 @@
...
+  subject { described_class.new }
+
+  let(:sales_form_submission) { create(:sales_form_submission) }
+
+  describe '#perform' do
+    it 'sends a Slack message' do
+      stub_url = 'fake_slack_url'
+      # We only actually try to send messages if RAILS_ENV is production
+      stub_const('ENV', {'SLACK_API_WEBHOOK_SALES' => stub_url,
+                         'RAILS_ENV' => 'production'})
```
dandrabik: Let's avoid changing the `RAILS_ENV` to 'production' for a test. Using a constant in the class that you can stub.

services/QuillLMS/spec/controllers/concerns/slack_tasks_spec.rb
```diff
@@ -5,10 +5,14 @@
 describe SlackTasks do
   include SlackTasks
 
-  let(:sales_form_submission) { create(:sales_form_submission) }
+  let(:sales_form_submission) { build(:sales_form_submission) }
 
   before do
     stub_const('ENV', {'RAILS_ENV' => 'production', 'SLACK_API_WEBHOOK_SALES' => 'slack-test.com'})
```
dandrabik: Outside this PR, but let's avoid changing the `RAILS_ENV` to `production`.

Let's add a local constant to SlackTasks and stub that if needed:

```ruby
Class Slack
  LIVE = Rails.env.production?

dandrabik (changes requested):
Thanks @anathomical for moving this along. I have a bunch of really small clean-up asks in my comments. None of them are particularly large, but since there is a good number of them, I'm marking it as "request changes" and will re-review when prompted.

Thanks for pitching and helping to carry this code along.

dandrabik (approved):
:+1: Thanks for making all of these changes and pushing this code forward.

happythenewsad (approved):
LGTM

PR #9307, opened 2022-06-29
Evidence internal tool/ fix rule analysis bug

eadams17:
## WHAT
fix issue where page filters were causing Evidence rule analysis page to crash

## WHY
this page shouldn't crash when updating filters

## HOW
make sure that we pass a `defaultPageSize` prop to the `ReactTable` instance that is greater than 0

### Screenshots
(If applicable. Also, please censor any sensitive data)

### Notion Card Links
(Please provide links to any relevant Notion card(s) relevant to this PR.)
https://www.notion.so/quill/Evidence-Rules-Analysis-Page-sometimes-crashing-out-16346bcfddc04196979f0010ccf804a4

PR Checklist | Your Answer
------------ | -------------
Have you added and/or updated tests? | small change, manually tested
Have you deployed to Staging? | yes
Self-Review: Have you done an initial self-review of the code below on Github? | yes
Spec Review: Have you reviewed the spec and ensured this meets requirements and/or matches design mockups? | N/A

PR #9382, opened 2022-07-20
add correct fallback value for react table page size

happythenewsad:
## WHAT
Use a workable default for ReactTable defaultPageSize. 

## WHY
This has arisen on multiple occasions, because we pass a value of `0` by default, which ReactTable chokes on (it uses this value to create an array, resulting in a RangeError). Time to add a generalized solution. 

- https://www.notion.so/quill/b417e5b89fe04af89cba62d4198f7031?v=77f294a5b0df466e98633244de7ebfef&p=9128aa61bbe44b3a9ce339c606269792
- https://github.com/empirical-org/Empirical-Core/pull/9307
### Screenshots

### Notion Card Links
https://www.notion.so/quill/b417e5b89fe04af89cba62d4198f7031?v=77f294a5b0df466e98633244de7ebfef&p=9128aa61bbe44b3a9ce339c606269792

PR Checklist | Your Answer
------------ | -------------
Have you added and/or updated tests? |  no - config only

### DIFF
--- a/services/QuillLMS/client/app/bundles/Shared/components/reactTable/reactTable.tsx
+++ b/services/QuillLMS/client/app/bundles/Shared/components/reactTable/reactTable.tsx
@@ -3,6 +3,8 @@ import { useTable, useSortBy, usePagination, useFilters, useExpanded, useGroupBy
 
 import ReactTablePagination from './reactTablePagination'
 
+const DEFAULT_PAGE_SIZE = 100
+
 function columnClassName(isSorted, isSortedDesc) {
   const defaultClassName = 'rt-th -cursor-pointer'
   if (!isSorted) { return defaultClassName }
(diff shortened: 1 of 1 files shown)


services/QuillLMS/client/app/bundles/Shared/components/reactTable/reactTable.tsx
```diff
@@ -3,6 +3,8 @@ import { useTable, useSortBy, usePagination, useFilters, useExpanded, useGroupBy
 
 import ReactTablePagination from './reactTablePagination'
 
+const DEFAULT_PAGE_SIZE = 100
```
emilia-friedberg, line 6: Maybe we should just do 1 here? Since it seems like this will never come into play unless there is no data at all.

emilia-friedberg (approved):

eadams17 (approved):

PR #9383, opened 2022-07-20
lms/fix-concept-result-migration-type

anathomical:
## WHAT
Fix a typo bug and update tests to make sure it doesn't happen again
## WHY
So that these jobs will complete
## HOW
Add an `=` where it should be, and also make sure that our tests cover setting the value (the tests failed because input data bypassed this line due to its guard statement).

PR Checklist | Your Answer
------------ | -------------

### DIFF
--- a/services/QuillLMS/app/workers/copy_single_concept_result_worker.rb
+++ b/services/QuillLMS/app/workers/copy_single_concept_result_worker.rb
@@ -16,7 +16,7 @@ class CopySingleConceptResultWorker
 
     ActiveRecord::Base.transaction do
       directions = ConceptResultDirections.find_or_create_by(text: directions)&.id if directions
-      instructions ConceptResultInstructions.find_or_create_by(text: instructions)&.id if instructions
+      instructions = ConceptResultInstructions.find_or_create_by(text: instructions)&.id if instructions
       previous_feedbacks = ConceptResultPreviousFeedback.find_or_create_by(text: previous_feedback)&.id if previous_feedback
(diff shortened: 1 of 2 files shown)


emilia-friedberg (approved):

services/QuillLMS/app/workers/copy_single_concept_result_worker.rb
```diff
@@ -16,7 +16,7 @@ def perform(old_concept_result_id)
 
     ActiveRecord::Base.transaction do
       directions = ConceptResultDirections.find_or_create_by(text: directions)&.id if directions
-      instructions ConceptResultInstructions.find_or_create_by(text: instructions)&.id if instructions
+      instructions = ConceptResultInstructions.find_or_create_by(text: instructions)&.id if instructions
```
brendanshean, line 19: Even with the guard clause, I'm not sure I understand why the compiler allowed that.

anathomical: I was wondering the same thing.  I guess maybe it's the Ruby ability to parse function arguments without parens?  I think it parses that as `instructions(ConceptResultInstructions.find_or_create_by(text: instructions)&.id) if instructions`, but there is no function `instructions`...  Though, again, the parser should probably have noticed?

brendanshean: Good point on the parentheses.  Since the ruby type checking happens at runtime instead of compile time, the parser doesn't know that instructions isn't a function and therefore allows the possibility that it's a function call.

PR #9393, opened 2022-07-21
Lms/updated parallel write responses

anathomical:
## WHAT
Begin writing `ConceptResult` records whenever we write `OldConceptResult` records so that both tables keep up to date with any new data from student activities
## WHY
We eventually want to replace `OldConceptResult` with `ConceptResult`, so we need to make sure we can write to both at the same time.  Once we're doing that, we can confirm that the data being written is complete in both places to build confidence for a full switch-over
## HOW
Add a worker that writes `ConceptResult` records, and enqueue that job during the `ActivitySessionController`'s `handle_concept_results` function which is the place where that data gets written.

### Notion Card Links
https://www.notion.so/quill/Write-Response-records-in-parallel-with-ConceptResults-977206a533da442c9c3cd9fa9ad97be7

PR Checklist | Your Answer
------------ | -------------

### DIFF
--- a/services/QuillLMS/app/models/concept_result.rb
+++ b/services/QuillLMS/app/models/concept_result.rb
@@ -55,6 +55,7 @@ class ConceptResult < ApplicationRecord
 
   def self.create_from_json(data_hash)
     response = init_from_json(data_hash)
+    puts response.as_json
     response.save!
     response
   end
@@ -81,13 +82,16 @@ class ConceptResult < ApplicationRecord
   end
 
   def self.bulk_create_from_json(data_hash_array)
-    data_hash_array.map { |data_hash| create_from_json(data_hash) }
+    data_hash_array.map { |data_hash| create_from_json(data_hash.clone) }
   end
 
   def self.find_or_create_from_old_concept_result(old_concept_result)
     return old_concept_result.concept_result if old_concept_result.concept_result
 
-    create_from_json(old_concept_result.as_json)
+    json = old_concept_result.as_json
+    json['old_concept_result_id'] = old_concept_result.id
+
+    create_from_json(json)
   end
 
   def self.parse_extra_metadata(metadata)
--- a/services/QuillLMS/app/models/concept_result.rb
+++ b/services/QuillLMS/app/models/concept_result.rb
@@ -82,7 +82,7 @@ class ConceptResult < ApplicationRecord
   end
 
   def self.bulk_create_from_json(data_hash_array)
-    data_hash_array.map { |data_hash| create_from_json(data_hash.clone) }
+    data_hash_array.map { |data_hash| create_from_json(data_hash) }
   end
 
   def self.find_or_create_from_old_concept_result(old_concept_result)
--- a/services/QuillLMS/app/models/concept_result.rb
+++ b/services/QuillLMS/app/models/concept_result.rb
@@ -88,10 +88,10 @@ class ConceptResult < ApplicationRecord
   def self.find_or_create_from_old_concept_result(old_concept_result)
     return old_concept_result.concept_result if old_concept_result.concept_result
 
-    json = old_concept_result.as_json
-    json['old_concept_result_id'] = old_concept_result.id
+    data_hash = old_concept_result.as_json
+    data_hash['old_concept_result_id'] = old_concept_result.id
 
-    create_from_json(json)
+    create_from_json(data_hash)
   end
 
   def self.parse_extra_metadata(metadata)
--- a/services/QuillLMS/app/models/concept_result.rb
+++ b/services/QuillLMS/app/models/concept_result.rb
@@ -55,7 +55,6 @@ class ConceptResult < ApplicationRecord
 
   def self.create_from_json(data_hash)
     response = init_from_json(data_hash)
-    puts response.as_json
     response.save!
     response
   end
@@ -71,6 +70,7 @@ class ConceptResult < ApplicationRecord
       concept_id: data_hash[:concept_id],
       attempt_number: metadata[:attemptNumber],
       correct: ActiveModel::Type::Boolean.new.cast(metadata[:correct]),
+      old_concept_result_id: data_hash[:old_concept_result_id],
       question_number: metadata[:questionNumber],
       question_score: metadata[:questionScore]
     )
--- a/services/QuillLMS/app/models/old_concept_result.rb
+++ b/services/QuillLMS/app/models/old_concept_result.rb
@@ -46,7 +46,7 @@ class OldConceptResult < ApplicationRecord
   # as documented when passed `return_primary_keys: true`.  Minimal modifications from
   # the original code found here, specifically around return values in if blocks:
   # https://github.com/jamis/bulk_insert/blob/master/lib/bulk_insert.rb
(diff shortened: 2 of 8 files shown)


services/QuillLMS/spec/controllers/api/v1/activity_sessions_controller_spec.rb
```diff
@@ -74,23 +75,29 @@
...
       it 'stores the concept results' do
-        activity_session.reload
-        expect(activity_session.old_concept_results.size).to eq 7
+        # Run Sidekiq jobs immediately instead of queuing them
+        Sidekiq::Testing.inline! do
+          expect do
+            put :update, params: { id: activity_session.uid, concept_results: concept_results }, as: :json
+          end.to change { activity_session.reload.old_concept_results.size }.by(3)
+             .and change { activity_session.reload.concept_results.size }.by(3)
+        end
```
anathomical, line 90: The original version of those code with `expect ... to eq 7` was counting the items that were created as part of setup as well as the ones created by the function call.  This cleans things up slightly to only count items new from the call.

services/QuillLMS/spec/workers/save_activity_session_concept_results_worker_spec.rb
```diff
@@ -0,0 +1,22 @@
...
+
+describe SaveActivitySessionConceptResultsWorker, type: :worker do
+
+  context '#perform' do
+    let(:old_concept_result) { create(:old_concept_result) }
+
+    it 'should save new ConceptResult records based on an array of OldConceptResult ids' do
+      expect { subject.perform(old_concept_result.id) }
+        .to change(ConceptResult, :count).by(1)
+      expect(old_concept_result.reload.concept_result).to be
```
brendanshean: I bet it's one of those rspec wrinkles but the need for `reload` in line 13 but not in line 19 is odd.

services/QuillLMS/app/models/old_concept_result.rb
```diff
@@ -40,4 +40,27 @@ def correct?
   def concept_uid=(concept_uid)
     self.concept = Concept.where(uid: concept_uid).first
   end
+
+  # Over-riding the `bulk_insert` classmethod from the bulk_insert gem because the
+  # behavior documented in the gem doesn't actually work.  This change makes it behave
+  # as documented when passed `return_primary_keys: true`.  Minimal modifications from
+  # the original code found here, specifically around return values in if blocks:
+  # https://github.com/jamis/bulk_insert/blob/master/lib/bulk_insert.rb
```
brendanshean, line 48: For future reference, I would add a note here about the open issue:  https://github.com/jamis/bulk_insert/issues/66#issuecomment-938355641

services/QuillLMS/app/models/concept_result.rb
```diff
@@ -87,7 +88,10 @@ def self.bulk_create_from_json(data_hash_array)
   def self.find_or_create_from_old_concept_result(old_concept_result)
     return old_concept_result.concept_result if old_concept_result.concept_result
 
-    create_from_json(old_concept_result.as_json)
+    json = old_concept_result.as_json
```
brendanshean: I think `data_hash` In lines 91-92, instead of `json` for the variable name would be more clear.

services/QuillLMS/app/models/concept_result.rb
```diff
@@ -70,6 +70,7 @@ def self.init_from_json(data_hash)
       concept_id: data_hash[:concept_id],
       attempt_number: metadata[:attemptNumber],
       correct: ActiveModel::Type::Boolean.new.cast(metadata[:correct]),
+      old_concept_result_id: data_hash[:old_concept_result_id],
```
brendanshean, line 73: I'm trying to make sense of this addition.  Do concept_results (i.e. the newer ones) from the large migration last week last lack this value?  Or were they added somewhere else.

services/QuillLMS/spec/controllers/api/v1/activity_sessions_controller_spec.rb
```diff
@@ -47,24 +47,25 @@
         create(:old_concept_result,
           activity_session_id: activity_session.id,
           concept: writing_concept,
-          metadata: { foo: 'bar' }
+          metadata: { foo: 'bar', correct: true }
```
brendanshean, line 50: Does this `correct: true` have some semantics behind it or is it just another key-value pair?

brendanshean (approved):
A couple of minor comments/questions but nothing holding up approval.  Nice sleuthing on the `bulk_insert` bug.

emilia-friedberg (approved):

services/QuillLMS/app/models/concept_result.rb
anathomical: Good question!  Actually, the migration took place entirely outside of the class structure (well, not totally, it did use `parse_extra_metadata`) in order to allow for bulk inserts.  You can see that code original migration code (which has already been replaced as I worked through the edge cases in the migration) here: https://github.com/empirical-org/Empirical-Core/blob/dd7ea7d7557/services/QuillLMS/app/workers/copy_old_concept_results_to_concept_results_worker.rb

services/QuillLMS/app/models/concept_result.rb
anathomical: That seems like a reasonable change.  I'll make the update.

services/QuillLMS/app/models/old_concept_result.rb
anathomical: Good suggestion.  I'll add a link to the comment.

services/QuillLMS/spec/controllers/api/v1/activity_sessions_controller_spec.rb
anathomical: This is a side effect of slightly more strict validation on the new model.  Specifically, SOME value for `correct` is required to build the new-shaped model.  So in this case, the `correct` key is required, but the value could be `true` or `false` (or even truth-y or false-y).

services/QuillLMS/app/models/old_concept_result.rb
```diff
@@ -40,4 +40,27 @@ def correct?
...
+  def self.bulk_insert(*columns, values: nil, set_size: 500, ignore: false, update_duplicates: false, return_primary_keys: false)
```
dandrabik: Maybe we should use a different method name here, rather than overwriting the gem's api (this could be surprising to someone later)?

dandrabik (approved):
One note about naming, but looks good to me.

services/QuillLMS/spec/workers/save_activity_session_concept_results_worker_spec.rb
anathomical: Actually, the inconsistency was just something that happened without planning.  Good call out since it turns out the `reload` is unnecessary.  Removed.

services/QuillLMS/app/models/old_concept_result.rb
anathomical: Yeah, that seems super-reasonable.  I'll change it to `custom_bulk_insert` unless you have a something else you prefer.

PR #9395, opened 2022-07-21
Unify google-oauth2 access

brendanshean:
## WHAT
Add a new `AuthGoogleAccessForm` component for google login/sign up.

## WHY
Having a single source of truth for which route to take makes our code more legible.

## HOW
Our omniauth configuration has two access [types](https://github.com/empirical-org/Empirical-Core/pull/9395/files#diff-73b9e84ca74d4d010f73cc756cb0557167455ca12d3eb3d64eceb9ecc5246f80R14) for google_oauth2, online and offline.  Each configuration has corresponding endpoints `/auth/google/online_access` and `/auth/google/offline_access` which were previously labeled `authentication_only` and `authentication_and_authorization` respectively.  Upon closer examination, I realized that `authentication_only` actually did authorization too, it just lacked the offline access, hence the renaming.

The distinction between online and offline access is tied to [refresh tokens](https://developers.google.com/identity/protocols/oauth2/openid-connect#refresh-tokens) but the overview is that the online access mode is preferred if our system already has a valid refresh token for a user.  If a user's refresh token is missing or expired, then the offline access flow is utilized to get a new one.

The one new piece of UI/UX is the `redirect to new_session_path(google_offline_access_expired: true)`.  Basically, if the check for a user's offline access fails, they are redirected to a sessions/new page which asks for re-authorization (see screenshot below).  @petersharkeyquill approved the copy and design for this new transition page.

### Screenshots
Screen Shot 2022-07-21 at 1 45 50 PM

### Notion Card Links
https://www.notion.so/quill/Fix-Google-oauth2-issue-c7615d7078c94c848317ca339bd19934

PR Checklist | Your Answer
------------ | -------------


anathomical (approved):
Looks good to me.

services/QuillLMS/app/controllers/auth/google_controller.rb
```diff
@@ -37,11 +40,11 @@ def authentication
...
   end
 
-  private def verify_authorization_for_offline_access
-    user = User.where('google_id = ? OR email = ?', @profile.google_id&.to_s, @profile.email&.downcase).first
+  private def check_for_authorization
+    user = User.find_by('google_id = ? OR email = ?', @profile.google_id&.to_s, @profile.email&.downcase)
     return if user.nil?
 
-    redirect_to GoogleIntegration::AUTHORIZATION_AND_AUTHENTICATION_PATH unless user.google_authorized?
+    redirect_to new_session_path(google_offline_access_expired: true) unless user.google_authorized?
```
dandrabik: Consider making this `return if user.google_authorized?` above

dandrabik (approved):
:+1: LGTM. Let's do a good amount of smoke testing, then 🚢 .

Thanks for putting this in place.

services/QuillLMS/app/controllers/auth/google_controller.rb
brendanshean: Good idea

services/QuillLMS/client/app/bundles/Teacher/components/accounts/AuthGoogleAccessForm.tsx
```diff
@@ -0,0 +1,33 @@
...
+export const AUTH_GOOGLE_ONLINE_ACCESS_PATH = "/auth/google/online_access"
+export const AUTH_GOOGLE_OFFLINE_ACCESS_PATH = "/auth/google/offline_access"
+
+const AuthGoogleAccessForm = ({
+  buttonClass = '',
+  formClass = '',
+  offlineAccess = false,
+  showIcon = true,
+  spanClass = '',
+  text
```
emilia-friedberg, line 14: Nice use of default parameters for default props.

emilia-friedberg (approved):
This looks good to me. I like your use of default props for that shared component.

services/QuillLMS/client/app/bundles/Teacher/components/accounts/AuthGoogleAccessForm.tsx
brendanshean: @emilia-friedberg I meant to ask you about this.  I chose empty string but wondered if `null` makes more sense here?

brendanshean:
FFR, smoke testing: https://www.notion.so/quill/Fix-Google-oauth2-issue-c7615d7078c94c848317ca339bd19934

PR #9397, opened 2022-07-21
lms/handle-null-characters-in-concept-result-migration

anathomical:
## WHAT
Add error handling for a couple of super rare examples
## WHY
Fewer than 30 records from OldConceptResult run into the errors being handled here when migrated in production, and those records are, as far as I can tell, the only ones left in the retry queue for the migration
## HOW
- Strip `\u0000` characters if they've somehow made it into `answer` since they're illegal in the table we're moving them to
- Check to see if `metadata` is a string instead of a JSON object, and if it is parse it into a JSON object.  (Not sure how this happened, but one or two records JSON-formatted strings instead of JSON objects, and need to be parsed on their way out of the database)

### Notion Card Links

PR Checklist | Your Answer
------------ | -------------

### DIFF
--- a/services/QuillLMS/app/workers/copy_single_concept_result_worker.rb
+++ b/services/QuillLMS/app/workers/copy_single_concept_result_worker.rb
@@ -10,10 +10,13 @@ class CopySingleConceptResultWorker
     return unless old_concept_result
     return if old_concept_result.concept_result.present?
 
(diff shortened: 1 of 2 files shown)


brendanshean (approved):
Looks good to me.  Nice work on sorting out this edge case.

happythenewsad (approved):

PR #9403, opened 2022-07-22
add low confidence feedback rules

emilia-friedberg:
## WHAT
Add low confidence feedback rules that are matched when a matched automl rule is suboptimal and has a confidence value of less than 0.6.

## WHY
We want to make sure that students are getting helpful feedback. This will help us not serve confusing feedback to students, and also more easily track when our models need training.

## HOW
Add the new rule type where necessary, and update the `automl` check to return this matched rule if the logic calls for it and the rule exists. Then update the Evidence CMS to allow curriculum team members to create and edit these rules. modeled closely on the existing interface we use for Plagiarism rules, as well as see some low-confidence-specific data in the Rules Analysis pages.

### Screenshots
Screen Shot 2022-07-22 at 2 15 20 PM
Screen Shot 2022-07-22 at 2 09 16 PM

### Notion Card Links
https://www.notion.so/quill/Custom-Low-confidence-Sub-optimal-Evidence-Feedback-Per-Prompt-9f1e268b6d2646108e6b2760445680b0

PR Checklist | Your Answer
------------ | -------------


services/QuillLMS/client/app/bundles/Staff/components/evidence/configureRules/ruleViewFormWrapper.tsx
```diff
@@ -23,15 +23,20 @@ const RuleViewFormWrapper = ({ activityData, isSemantic, isUniversal, location,
   let rule: RuleInterface;
 
   if(!ruleId) {
-    const ruleType = state && state.ruleType;
-    const promptIds = state && state.promptIds;
+    const { ruleType, promptIds, conceptUid, name, } = state || {}
```
anathomical, line 26: I dig this pattern for conditional destructuring.

services/QuillLMS/engines/evidence/lib/evidence/evidence/automl_check.rb
```diff
@@ -14,37 +16,73 @@ def initialize(entry, prompt, previous_feedback=[])
 
     # rubocop:disable Metrics/CyclomaticComplexity
     def feedback_object
-      return unless matched_rule
+      return unless matched_automl_rule
 
-      feedback = matched_rule.determine_feedback_from_history(@previous_feedback)
-      highlight = feedback.highlights.map do |h|
+      if matched_low_confidence_rule
```
anathomical: There's a lot of repetition between the `if` and `else` block here that I think is unnecessary.  The only differences are `feedback_type` which I think we can pull from `matched_rule.rule_type`, and the values in `api`.  Speaking of, I think it's pretty weird to save `@confidence_score` as two different values.  We should probably aim to be consistent and just store it as `confidence` in any `FeedbackHistory` that has access to that number so that we can do comparisons between them.  (See my notes below.)

services/QuillLMS/engines/evidence/lib/evidence/evidence/automl_check.rb
```diff
@@ -14,37 +16,73 @@ def initialize(entry, prompt, previous_feedback=[])
...
+    private def fetch_matched_low_confidence_rule
+      @prompt.rules&.find_by(rule_type: Evidence::Rule::TYPE_LOW_CONFIDENCE)
+    end
+
+    private def matched_automl_rule
+      @matched_automl_rule ||= fetch_matched_automl_rule
     end
 
-    private def fetch_matched_rule
+    private def fetch_matched_automl_rule
```
anathomical: I think that most of this can actually be consolidated into a `fetch_matched_rule` function that does the work to figure out whether the "matched" rule is of type AutoML or of type LowConfidence.  Something like:
```
private def fetch_matched_automl_rule
  return unless @automl_model

  @google_automl_label, @confidence_score = @automl_model.fetch_automl_label(@entry)
  automl_rule = @prompt.rules.joins(:label).find_by(comprehension_labels: {name: @google_automl_label})
  if automl_rule.optimal || @confidence_score >= LOW_CONFIDENCE_THRESHOLD
    automl_rule
  else
    @prompt.rules&.find_by(rule_type: Evidence::Rule::TYPE_LOW_CONFIDENCE)
  end
end
```
Then you just have a single `matched_rule` value that you can use to build your payload above, and you can figure out which `api` values based on the `matched_rule.rule_type` and set `rule_type` by `matched_rule.rule_type`.

services/QuillLMS/engines/evidence/spec/models/evidence/rule_spec.rb
```diff
@@ -220,6 +220,31 @@ module Evidence
...
+      let!(:prompt2) { create(:evidence_prompt) }
+      let!(:low_confidence_rule) { create(:evidence_rule, :rule_type => (Rule::TYPE_LOW_CONFIDENCE), :prompt_ids => ([prompt1.id])) }
+
+      it 'should creates low_confidence rule if first rule for prompt' do
+        expect(low_confidence_rule.valid?).to(eq(true))
+      end
+
+      it 'should does not create low_confidence rule if low_confidence rule already exists for prompt' do
+        invalid_low_confidence_rule = build(:evidence_rule, :rule_type => (Rule::TYPE_LOW_CONFIDENCE), :prompt_ids => ([prompt1.id]))
+        expect((!invalid_low_confidence_rule.valid?)).to(be_truthy)
```
anathomical: The negation here reads a little odd.  I think I'd write this to expect false to match your other tests: `expect(invalid_low_confidence_rule.valid?).to(eq(false))`

services/QuillLMS/app/lib/rule_feedback_history.rb
```diff
@@ -42,13 +42,24 @@ def self.exec_query(conjunction:, activity_id:, start_date:, end_date:, turk_ses
...
+
     {
         response_id: f_h.id,
         datetime: f_h.updated_at,
         entry: f_h.entry,
         highlight: f_h.metadata.instance_of?(Hash) ? f_h.metadata['highlight'] : '',
         session_uid: f_h.feedback_session_uid,
-        strength: f_h.feedback_history_ratings.max_by(&:updated_at)&.rating
+        strength: f_h.feedback_history_ratings.max_by(&:updated_at)&.rating,
+        low_confidence_predicted_rule: low_confidence_predicted_rule
```
anathomical: This approach seems just fine, but any reason not to just naively include `api: f_h.metadata['api']` on the JSON body and let the front-end piece together name, uid, and confidence without doing the work here?  And like you fallback in the `else` above, this could just be nil (or you could do `|| {}` if you need an empty hash as a default).

services/QuillLMS/spec/lib/rule_feedback_history_spec.rb
```diff
@@ -296,6 +296,7 @@ def feedback_factory(&hash_block)
         datetime: f_h.updated_at,
         entry: f_h.entry,
         highlight: f_h.metadata.instance_of?(Hash) ? f_h.metadata['highlight'] : '',
+        low_confidence_predicted_rule: {},
```
anathomical: Note that if you replace this key with a naive pass-through of `api` then this will also need to be updated.

services/QuillLMS/engines/evidence/spec/lib/automl_check_spec.rb
```diff
@@ -43,6 +45,14 @@ module Evidence
         end
       end
 
+      it 'should return the low confidence feedback payload when there is a label match but the confidence is below the threshold' do
+        AutomlModel.stub_any_instance(:fetch_automl_label, [label.name, 0.5]) do
+          entry = "entry"
+          automl_check = Evidence::AutomlCheck.new(entry, prompt)
+          expect(:feedback => low_confidence_feedback.text, :feedback_type => "low-confidence", :optimal => low_confidence_rule.optimal,  :entry => entry, :concept_uid => ((low_confidence_rule&.concept_uid or "")), :rule_uid => (low_confidence_rule&.uid), :highlight => ([]), :hint => nil, :api => {:original_rule_confidence => 0.5, :original_rule_name => rule.name, :original_rule_uid => rule.uid}).to(eq(automl_check.feedback_object))
```
anathomical: I think this will be more readable if you split the expected hash into multiple lines:
```
expect({
  feedback: low_confidence_feedback.text,
  ...
  original_rule_uid: rule.uid
}).to(eq(automl_check.feedback_object))
```
(Also, minor personal preference, but I generally like to `expect` the value of function returns or other variables, and `to eq` constants.  So I'd actually `expect(automl_check.feedback_object).to eq({ ... })`, but it doesn't really matter, so no need to change it.)

anathomical (changes requested):
This general approach looks solid to me.  I think you can consolidate some code in the `automl_check` file so that it's more readable and repeats itself less.  You don't have to do everything I suggest there, but I'd like to see some amount of clean-up of that code.

services/QuillLMS/engines/evidence/lib/evidence/evidence/automl_check.rb
emilia-friedberg: @anathomical that makes sense, I'll get to consolidating this on Monday. To your second point, I actually originally had just `confidence` for both, but I think that's a little misleading because it means different things in these cases, since the `@confidence_score` is always from the original model. With just `confidence` for both I'd be worried that someone would read the value as an evaluation of how sure we are about applying the `low_confidence` rule, which it isn't really. I'm open to changing it back but wanted to flag that there is a little bit of disjunction in how we would be understanding the same value in both cases. Let me know what you think.

anathomical: @emilia-friedberg That's a totally valid concern to have.  I think that we probably want comparable values so that, at the very least, someone could do a query of "all `FeedbackHistory` records with confidence = X" and get both items we flagged low confidence and those we didn't.  Especially since, and I just realized this, the spec calls for only using "low confidence" for non-optimal responses.  So we may have a non-trivial number of `AutoML`-type `FeedbackHistory` records that have confidence below the specified threshold.  (I don't think the current code covers this, I'll modify my comment below with a sample I think handles the case pretty simply.)

Basically, you're definitely right that this may cause confusion, but I think we want to be able to compare confidence levels no matter what type of feedback a student sees.

services/QuillLMS/engines/evidence/lib/evidence/evidence/automl_check.rb
emilia-friedberg: Hey @anathomical  - I started trying to implement this and then realized that it is a little complicated by the fact that we do need the `automl` values for the `low_confidence` rule for the api values, so I think it still makes sense to have separate methods that define both. I did refactor the top function in the way you suggested - let me know if you think there's more cleanup we should do here.

anathomical (approved):
Looks good to me.  Thanks for the clean up.

services/QuillLMS/client/app/bundles/Staff/components/evidence/configureRules/ruleViewForm.tsx
```diff
@@ -268,7 +270,7 @@ const RuleViewForm = ({
           rulePromptsDisabled={rulePromptsDisabled}
           setRulePrompts={setRulePrompts}
         />}
-        {(isUniversal || isSemantic) && <RuleUniversalAttributes
+        {(isUniversal || isSemantic || isLowConfidence) && <RuleUniversalAttributes
```
eadams17: I feel like we can probably pull this out into a const, something like `showUniversalAttributes`

eadams17 (approved):
looks great! thanks for knocking this out

services/QuillLMS/engines/evidence/lib/evidence/evidence/automl_check.rb
```diff
@@ -24,27 +28,45 @@ def feedback_object
...
       }
     end
     # rubocop:enable Metrics/CyclomaticComplexity
 
-    private def matched_rule
-      @matched_rule ||= fetch_matched_rule
+    private def matched_low_confidence_rule
+      return unless matched_automl_rule && !matched_automl_rule.optimal && @confidence_score < LOW_CONFIDENCE_THRESHOLD
+
+      @matched_low_confidence_rule ||= fetch_matched_low_confidence_rule
```
happythenewsad: Looks like this instance var only exists for testing. It would be cleaner to stub the method `matched_low_confidence_rule` and just return a value on this line, rather than making an instance variable.

happythenewsad (changes requested):
I did a quick review as requested - looks good except for one syntax thing.

happythenewsad (approved):

PR #9408, opened 2022-07-25
Remove rails_12_factor gem

brendanshean:
## WHAT
Remove the rails_12_factor gem from the LMS.

## WHY
The gem is no longer maintained.

There is also a deprecation in Rails 6.1: 
```
DEPRECATION WARNING: Including LoggerSilence is deprecated and will be removed in Rails 6.1. 
Please use `ActiveSupport::LoggerSilence` instead 
(called from <top (required)> at /tmp/build_48c3bc04/config/application.rb:15)
```

## HOW
Follow the [instructions](https://github.com/heroku/rails_12factor#migrating-to-rails-5) for removing  the gem.

Note the setting of ENV variables:
Screen Shot 2022-07-25 at 10 55 15 AM

### Screenshots

### Notion Card Links

PR Checklist | Your Answer
------------ | -------------

### DIFF
--- a/services/QuillLMS/Gemfile
+++ b/services/QuillLMS/Gemfile
@@ -151,7 +151,6 @@ gem 'awesome_print'
 
 group :production, :staging do
   gem 'rails_autoscale_agent'
-  gem 'rails_12factor'
   gem 'lograge'
   gem 'sqreen'
 end
--- a/services/QuillLMS/config/environments/production.rb
+++ b/services/QuillLMS/config/environments/production.rb
@@ -115,4 +115,14 @@ EmpiricalGrammar::Application.configure do
     'www.quill.org.' => 'www.quill.org',
     'quill.org' => 'www.quill.org'
   }
+
+  # Disable serving static files from the `/public` folder by default since
+  # Apache or NGINX already handles this.
+  config.public_file_server.enabled = ENV['RAILS_SERVE_STATIC_FILES'].present?
+
+  if ENV["RAILS_LOG_TO_STDOUT"].present?
+    logger = ActiveSupport::Logger.new($stdout)
+    logger.formatter = config.log_formatter
+    config.logger = ActiveSupport::TaggedLogging.new(logger)
+  end
 end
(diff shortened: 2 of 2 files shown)


anathomical (approved):
Should individual developers expect to need to provide these ENV variables for their staging environments?  What about our shared staging environments?  Just making sure we know what the plan is for all of our different environments.

dandrabik (changes requested):
Thanks for cleaning this up. Should we add this to `.env-sample` for the bootstrap script? Not sure if this is preferred to have set locally.

Requesting these small changes: 

`ENV` vars evaluate to strings. It's probably better to check for `var.present? && var == "true"` to make the `ENV` var work as expected (setting it to `false` currently still evaluates to `true`, see screens). (Note, I'm aware this is migration code copied directly from the gem, but I think we should make it better):
Screen Shot 2022-07-25 at 11 16 23 AM
Screen Shot 2022-07-25 at 11 12 57 AM

brendanshean:
> Thanks for cleaning this up. Should we add this to `.env-sample` for the bootstrap script? Not sure if this is preferred to have set locally.
> 
> Requesting these small changes:
> 
> `ENV` vars evaluate to strings. It's probably better to check for `var.present? && var == "true"` to make the `ENV` var work as expected (setting it to `false` currently still evaluates to `true`, see screens). (Note, I'm aware this is migration code copied directly from the gem, but I think we should make it better): Screen Shot 2022-07-25 at 11 16 23 AM Screen Shot 2022-07-25 at 11 12 57 AM

Good point.  I will update.

brendanshean:
> Should individual developers expect to need to provide these ENV variables for their staging environments? What about our shared staging environments? Just making sure we know what the plan is for all of our different environments.

@anathomical Good question.  I'm planning on updating all the relevant heroku environments once this PR is approved.

brendanshean:
> Thanks for cleaning this up. Should we add this to `.env-sample` for the bootstrap script? Not sure if this is preferred to have set locally.

@dandrabik I don't think it's necessary for `.env-sample` since this configuration is for the production environment.

happythenewsad (approved):

dandrabik (approved):
:+1: Thanks for making these changes and cleaning this up.

PR #9417, opened 2022-07-25
lms/create-concept-result-analyzer

anathomical:
## WHAT
Add a background worker to figure out which old_concept_ids didn't migrate
## WHY
So that we can get a list of items to try migrating again in order to hopefully fill in all of our gaps.  We want all of our `OldConceptResult` records to be successfully migrated.
## HOW
Do a series of LEFT OUTER JOINS to figure out which `OldConceptResult`s don't have a corresponding `ConceptResult` record, then email a list of all of those ids to Thomas for later use.

### Notion Card Links

PR Checklist | Your Answer
------------ | -------------

### DIFF
--- a/services/QuillLMS/app/mailers/application_mailer.rb
+++ b/services/QuillLMS/app/mailers/application_mailer.rb
@@ -1,0 +1,5 @@
+# frozen_string_literal: true
+
+class ApplicationMailer < ActionMailer::Base
+
+end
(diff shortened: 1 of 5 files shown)


services/QuillLMS/app/mailers/application_mailer.rb
```diff
@@ -0,0 +1,5 @@
+# frozen_string_literal: true
+
+class ApplicationMailer < ActionMailer::Base
```
anathomical, line 3: We have a linting rule not to allow new classes to inherit from `Action____::Base`, so we need to create intermediate `Application____` models to inherit from.  This is the first `Mailer` that we've written since that rule got put into place.

brendanshean (approved):
Looks good!  Nice clear testing.

PR #9419, opened 2022-07-26
update readability grade level mapping

emilia-friedberg:
## WHAT
Update readability grade level mapping.

## WHY
This was a request from the curriculum team.

## HOW
Just update this method in accordance with the new mapping (https://docs.google.com/spreadsheets/d/1Xu0MemcC4ArLO2lnxVa8yjXy8Uir5h5OvX6bVs636LY/edit#gid=1190603534) which no longer uses activity classifications. We could also just make this into a hash now that we don't need to take the extra parameter into account - if anyone feels strongly about that let me know. Once this is live, I'm going to run `Activity.all.each { |a| a.set_minimum_and_maximum_grade_levels_to_default_values }` to make sure the new grade ranges (aka `minimum_grade_level` and `maximum_grade_level` match the new readability values).

### Screenshots

### Notion Card Links
https://www.notion.so/quill/Update-lexile-score-to-readability-grade-level-mapping-804103f5feae4fd49fb8c95cb6ec39db

PR Checklist | Your Answer
------------ | -------------
Have you added and/or updated tests? |  YES (just removed the test around how the grade levels should change if the activity classification is different)

### DIFF
--- a/services/QuillLMS/app/controllers/cms/activities_controller.rb
+++ b/services/QuillLMS/app/controllers/cms/activities_controller.rb
@@ -84,7 +84,7 @@ class Cms::ActivitiesController < Cms::CmsController
   protected def set_raw_score_options_and_raw_score_to_readability_grade_band
     @raw_score_options = RawScore.order_by_name
     @raw_score_to_readability_grade_band = {}
-    @raw_score_options.each { |rs| @raw_score_to_readability_grade_band[rs.name] = rs.readability_grade_level(@activity_classification.id) }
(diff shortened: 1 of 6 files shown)


brendanshean (approved):
Looks good to me.  I really like that you have reduced complexity of the code.

In this instance, I like the case statement over a hash since it communicates intentional grouping of labels. If there were just one label and one value, I would probably recommend a hash.

anathomical (approved):
I like that this is more simplified so that we don't have a weird exception for Evidence.

happythenewsad (approved):

PR #9420, opened 2022-07-26
Handle invalid authenticity token errors

brendanshean:
## WHAT
Address invalid authenticity token exceptions.

Addendum: Restrict scope to just SessionsController.  In a later PR, extend to ApplicationController

## WHY
This [post](https://abletech.nz/resource/rails-invalidauthenticitytoken-around-the-edges/) describes some use cases for invalid authenticity tokens.  Currently, if a user encountered this state where their session had expired, we are not showing any errors, but then also getting reports in New Relic and Sentry.

## HOW
I added a `rescue_from ActionController::InvalidAuthenticityToken` clause that will reload the page for a user when this invalid authenticity token error is triggered thereby getting the csrf-token in sync with the current session.  Subsequent post and put requests should then go through as expected.

NB.
To test the redirection manually, I added some randomization for invalid tokens:
```
authenticity_token: Math.random() < 0.5 ? 'invalidAuthenticityToken' : getAuthToken()
```

### Screenshots
Screen Shot 2022-07-26 at 1 09 16 PM

### Notion Card Links
https://www.notion.so/quill/Address-invalid-authenticity-token-errors-0be45c9bc9af4f848c8d132516a9d5ad

PR Checklist | Your Answer
------------ | -------------


anathomical (approved):
Looks good to me.  Stylistically, do we want to generally prefer creating concerns for this sort of thing even when we're only intending to include it in a single place?  No objections, just figured it was worth a brief discussion.

brendanshean:
> Looks good to me. Stylistically, do we want to generally prefer creating concerns for this sort of thing even when we're only intending to include it in a single place? No objections, just figured it was worth a brief discussion.

@anathomical Initially I extracted this out into a concern since I was using it multiple places, but after moving into `ApplicationController` only, the concern doesn't make as much sense.

NB. After discussing with Dan, I've scaled back scope to just `SessionController` for now to see if this is working there before I extend to the rest of the app.

brendanshean:
@dandrabik You asked earlier about GET requests and infinite loops.  This was the aforementioned line about [idempotency](https://github.com/rails/rails/blob/433115554d3ea0b4dab9df99548c47707cadb20d/actionpack/lib/action_controller/metal/request_forgery_protection.rb#L290)

PR #9421, opened 2022-07-26
change beta and gamma flag logic

happythenewsad:
## WHAT
Changing activity flagging logic. 
- Gamma now includes gamma + beta. 
- Beta no longer includes gamma. 
- made existing rspecs more readable. 

## WHY
Product request. 

## HOW
Update Activity scope logic. 

### Screenshots

### Notion Card Links
https://www.notion.so/quill/Product-Board-30f97e4eb01246dbb8264253fb385073?p=a8906a7321c6439586a3b3aef14f9d54&pm=s

PR Checklist | Your Answer
------------ | -------------

### DIFF
--- a/services/QuillLMS/app/queries/activity_search.rb
+++ b/services/QuillLMS/app/queries/activity_search.rb
@@ -12,9 +12,9 @@ class ActivitySearch
     when 'alpha'
       flags = "'alpha', 'beta', 'gamma', 'production'"
     when 'beta'
-      flags = "'beta', 'gamma', 'production'"
+      flags = "'beta', 'production'"
     when 'gamma'
-      flags = "'gamma', 'production'"
+      flags = "'gamma', 'production', 'beta'"
     else
       flags = "'production'"
     end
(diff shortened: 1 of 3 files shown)


anathomical (approved):
This accomplishes the stated goal, so I'm approving this.  That said, I think that this reveals that the way we've been using "gamma" is probably wrong.  We use it to mean "College Board teachers", and this change makes it so that "beta" users don't see College Board content when they're not supposed to.  However, it now gives every College Board teacher access to our beta activities, even if that doesn't make sense.

So we should merge this as-is to solve the problem short-term, but ultimately we should do something more targeted for College Board content.

services/QuillLMS/app/queries/activity_search.rb
```diff
@@ -12,9 +12,9 @@ def self.search(flag)
     when 'alpha'
       flags = "'alpha', 'beta', 'gamma', 'production'"
     when 'beta'
-      flags = "'beta', 'gamma', 'production'"
+      flags = "'beta', 'production'"
```
happythenewsad, line 15: @anathomical FYI I added this change to the PR, since ActivitySearch uses a separate code path.

anathomical: 👍

PR #9422, opened 2022-07-26
Segment/ update client side events

eadams17:
## WHAT
update client side calls for Segment

## WHY
we want these to be accessible in Ortto

## HOW
add an API for fetching the `user_id` of a student and the teacher for that assigned activity to pass to the Segment event, add an `isTrackableStudentEvent` function to only allow event calls to happen for student users; update naming from Comprehension to Evidence

Note: I had to remove the function that appended `custom` to event property keys because we need the keys to be `user_id` and `properties: { student_id: ### }` in order for these events to be captured in Ortto.
### Screenshots

### Notion Card Links
https://www.notion.so/quill/Add-User-ID-to-client-side-Segment-events-or-convert-them-to-server-side-af4b5293c59d409f9c3b563ed5aecca2

PR Checklist | Your Answer
------------ | -------------


services/QuillLMS/app/controllers/api/v1/users_controller.rb
```diff
@@ -17,6 +17,14 @@ def current_user_role
     render json: { role: current_user&.role }
   end
 
+  def student_and_teacher_ids_for_session
+    uid = params[:activity_session_uid]
+    student_id = nil
+    student_id = current_user.id if current_user&.student?
```
emilia-friedberg: Couldn't we just do `student_id = current_user&.student? current_user.id : nil` here?

services/QuillLMS/app/controllers/api/v1/users_controller.rb
```diff
@@ -17,6 +17,14 @@ def current_user_role
...
+    teacher_id = ActivitySession.where(uid: uid).first&.classroom&.teachers&.first&.id
```
emilia-friedberg: I think this could be `teacher_id = ActivitySession.find_by(uid: uid)&.classroom&.owner&.id`

services/QuillLMS/client/app/bundles/Diagnostic/modules/analytics/index.ts
```diff
@@ -41,7 +41,7 @@ class SegmentAnalytics {
       return false;
     }
 
-    const eventProperties = Object.assign(this.formatCustomProperties(properties), this.getDefaultProperties());
+    const eventProperties = Object.assign(properties, this.getDefaultProperties());
```
emilia-friedberg: I don't know if this is going to cause a bug here, but in general we should use the spread syntax and either do `Object.assign(...properties, this.getDefaultProperties)` or just `{ ...properties, ...this.getDefaultProperties }` to avoid directly transforming the target object. This wasn't an issue with the code before because we were translating the output of `this.formatCustomProperties` instead of `properties` itself, but can result in wonky behavior.

services/QuillLMS/client/app/bundles/Evidence/components/studentView/container.tsx
```diff
@@ -335,7 +365,7 @@ export const StudentViewContainer = ({ dispatch, session, isTurk, location, acti
   function completeStep(stepNumber: number) {
     const newCompletedSteps = completedSteps.concat(stepNumber)
     const uniqueCompletedSteps = Array.from(new Set(newCompletedSteps))
-    trackCurrentPromptCompletedEvent()
+    isTrackableStudentEvent(idData) && trackCurrentPromptCompletedEvent()
```
emilia-friedberg: I think it would probably cleaner to move the check about whether or not we should track into the tracking function itself.

services/QuillLMS/client/app/bundles/Evidence/components/studentView/container.tsx
```diff
@@ -361,7 +391,7 @@ export const StudentViewContainer = ({ dispatch, session, isTurk, location, acti
     completeStep(READ_PASSAGE_STEP_NUMBER)
     const scrollContainer = document.getElementsByClassName("read-passage-container")[0]
     scrollContainer.scrollTo(0, 0)
-    trackPassageReadEvent();
+    isTrackableStudentEvent(idData) && trackPassageReadEvent();
```
emilia-friedberg: See above.

services/QuillLMS/client/app/bundles/Evidence/modules/analytics/index.ts
```diff
@@ -26,7 +26,7 @@ class SegmentAnalytics {
       return false;
     }
 
-    const eventProperties = Object.assign(this.formatCustomProperties(properties), this.getDefaultProperties());
+    const eventProperties = Object.assign(properties, this.getDefaultProperties());
```
emilia-friedberg: Spread thing again.

emilia-friedberg (changes requested):
I left some specific comments, but overall I think this approach feels a little verbose to me - we are passing around the same parameters through a lot of child components because the `analytics` function always needs them, which makes me feel like we should just be doing the actual retrieval of that data in the analytics function itself. We could do both the `isTrackableStudentEvent` and `idData` retrieval there, and then you wouldn't have to update any of the existing tracking code or pass `idData` around as a prop.

emilia-friedberg (approved):
Looks great! Thanks for making those changes!

services/QuillLMS/client/app/bundles/Diagnostic/modules/analytics/index.ts
```diff
@@ -61,17 +76,6 @@ class SegmentAnalytics {
...
-  formatCustomProperties(properties: object): object {
-    if (typeof properties != 'object') {
-      properties = {};
-    }
-    return Object.keys(properties).reduce((accumulator, key) => {
-      let customKeyName = `custom_${key}`;
-      accumulator[customKeyName] = properties[key];
-      return accumulator;
-    }, {});
-  }
```
anathomical: As long as this is working in testing on staging, I don't think there'll be a problem, but now that I look at this code again I vaguely remember that we had to prepend `custom_` to avoid colliding with some platform's reserved keys list.  (Like, some platform wouldn't let you provide `user` or something.  But, like I said, if this is working in testing, I wouldn't worry about it.)

services/QuillLMS/client/app/bundles/Evidence/modules/analytics/index.ts
```diff
@@ -1,21 +1,37 @@
 import { Event } from './event_definitions';
 import Events from './events';
 
+import getParameterByName from '../../helpers/getParameterByName';
```
anathomical, line 4: I noticed that you updated this file in the `Diagnostic` modules, too.  Are they actually identical?  Should we move this to `Shared`?

anathomical (approved):
One comment about moving a pair of potentially duplicate files to `Shared`, but this looks fine to me.

services/QuillLMS/client/app/bundles/Evidence/modules/analytics/index.ts
eadams17: yeah good catch, we actually have this same function defined in all of the bundles. it's used in a lot of different places, so I'll make a card to consolidate it into the Shared bundle and replace all of its instances

services/QuillLMS/client/app/bundles/Diagnostic/modules/analytics/index.ts
eadams17: I'll add this back in and just add a check to skip the `user_id` and `properties` keys

PR #9437, opened 2022-07-28
Segment/ omit trait properties with nil values

eadams17:
## WHAT
omit trait properties with nil values from Segment identify calls

## WHY
they cause 0 values to be stored in Ortto which is not ideal

## HOW
just add a `.reject` call onto the properties object that checks for `nil` values

### Screenshots

### Notion Card Links
https://www.notion.so/quill/Add-1-school-2-admin-status-3-email-to-Segment-Identify-call-08e66be5f62b4b2baab94170f92f344f

PR Checklist | Your Answer
------------ | -------------

### DIFF
--- a/services/QuillLMS/app/services/analytics/segment_analytics.rb
+++ b/services/QuillLMS/app/services/analytics/segment_analytics.rb
@@ -227,7 +227,7 @@ class SegmentAnalytics
         school_name: user.school&.name,
         school_id: user.school&.id,
         district: user.school&.district&.name
-      },
+      }.reject {|_,v| v.nil? },
       integrations: integration_rules(user.id)
     }
   end
--- a/services/QuillLMS/spec/services/analytics/segment_analytics_spec.rb
+++ b/services/QuillLMS/spec/services/analytics/segment_analytics_spec.rb
@@ -10,221 +10,221 @@ describe 'SegmentAnalytics' do
   let(:track_calls) { analytics.backend.track_calls }
   let(:identify_calls) { analytics.backend.identify_calls }
 
-  # context 'tracking classroom creation' do
-  #   let(:classroom) { create(:classroom) }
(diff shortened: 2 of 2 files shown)


services/QuillLMS/app/services/analytics/segment_analytics.rb
```diff
@@ -227,7 +227,7 @@ def identify(user)
         school_name: user.school&.name,
         school_id: user.school&.id,
         district: user.school&.district&.name
-      },
+      }.reject {|_,v| v.nil? },
```
brendanshean, line 230: Take or leave comment:  I realize this is older code, but since you are in there, a suggested refactor.

The traits method is highly coupled to `user`, and looks like it's violating the law of demeter.  To fix this extract the traits into the user model:

```
def segment_identify_traits
  {
    premium_state: premium_state,
    premium_type: subscription&.account_type,
    auditor: auditor?,
    is_admin: admin?,
    email: email,
    school_name: school&.name,
    school_id: school&.id,
    district: school&.district&.name
    }.reject {|_,v| v.nil? }
end
```

Then you have a much cleaner:

```
private def identify_params(user)
  {
    user_id: user.id,
    traits: user.segment_identify_traits
    integrations: integration_rules(user.id)
  }
end
```

brendanshean: You could do the same thing with integration_rules

brendanshean (approved):
One minor refactoring comment, but otherwise looks good to go.

anathomical (approved):

services/QuillLMS/app/services/analytics/segment_analytics.rb
eadams17: thanks for explaining this refactor! I'll add it in with the additional Segment updates that are slated for this sprint

PR #9439, opened 2022-07-29
Evidence Seed Data via OpenAI

dandrabik:
## WHAT
Add some classes and interfaces for generating synthetic data using OpenAI's completion API (the one powered by GPT-3). 

The goal is to generate initial seed data for an activity so the team can start labeling. This is an initial attempt to replace Mechanical Turk for that part of the process.

This will be followed by code to upload or email these csvs, and a UI to enter nouns and hit generate.

Note, the algorithm for generation will likely change a bit as we experiment (Curriculum is reviewing the first output of the script).

## WHY
Mechanical Turk is slow, expensive, and not particularly good for generating data for our needs. This is a major pain point for Curriculum.
## HOW
Basically, these are the forms that we send to the 
- full passage plus prompt
- full passage plus prompt plus common starter noun for that passage
- A chunk of the passage plus prompt

### Screenshots
Sample output
## Instant replay can make NFL games more accurate, so
Screen Shot 2022-07-29 at 3 53 50 PM

### Notion Card Links

PR Checklist | Your Answer
------------ | -------------
Have you deployed to Staging? | Not yet - deploying shortly


services/QuillLMS/engines/evidence/lib/evidence/evidence/open_ai.rb
```diff
@@ -0,0 +1,110 @@
...
+        input: prompt,
+        instruction: PARAPHRASE_INSTRUCTION,
+        n: count,
+      }.merge(options_hash)
+    end
+
+    def edit_request
+      return response if response.present?
+
+      @response = self.class.post(EDITS_ENDPOINT, body: edits_request_body.to_json, headers: self.class.header_options)
```
brendanshean: Where is `header options` coming from?

services/QuillLMS/engines/evidence/lib/evidence/evidence/open_ai.rb
```diff
@@ -0,0 +1,110 @@
...
+      @response = nil
+    end
+
+    # Edit endpoint code still in experimentation phase
+    EDITS_ENDPOINT = '/edits'
+    EDIT_MODELS = {
+      ada: 'text-ada-edit-001',
+      babbage: 'text-babbage-edit-001',
+      curie: 'text-curie-edit-001',
+      davinci: 'text-davinci-edit-001'
```
brendanshean: Why is there a diff with 'text-davinci-edit-002' and the COMPLETION_MODELS 'text-davinci-edit-001' but not the other models?

services/QuillLMS/engines/evidence/lib/evidence/evidence/open_ai.rb
```diff
@@ -0,0 +1,110 @@
+# frozen_string_literal: true
+
+module Evidence
+  class OpenAI < Evidence::ApplicationService
```
brendanshean: `OpenAI` feels more like a namespace descriptor.  Since this object is coupled to `/completion` and `COMPLETION_MODELS` it might be more transparent to have the class name be: `OpenAICompletion`

brendanshean (approved):
This looks great!  It was really fun to go through this one.

services/QuillLMS/engines/evidence/lib/evidence/evidence/open_ai.rb
dandrabik: Good Point. I'm going to change/move this to `Evidence::OpenAI::Completion` making `OpenAI` a proper module/namespace

services/QuillLMS/engines/evidence/lib/evidence/evidence/open_ai.rb
dandrabik: Note, I'm going to remove the edit code for now (stash it in a branch) since it's still untested. 

I just started experimenting with it, and according to the docs the `edit` models are "different" than the `completion` ones (note the 'edit' in the name), but for some reason it doesn't document all of their names. The davinci one is the only one I got to work (that name was in an example), but I was guessing at the rest of the names for now.

services/QuillLMS/engines/evidence/lib/evidence/evidence/open_ai.rb
dandrabik: Good catch, that was a typo from refactoring (but I removed this line anyway)

services/QuillLMS/engines/evidence/lib/evidence/evidence/synthetic/initial_data.rb
```diff
@@ -0,0 +1,138 @@
...
+          if index == 0
+            passage_csv_name = "#{short_name}_passage_chunks"
+            csvs[passage_csv_name] = data.generate_passages_csv
+          end
+        end
+
+        csvs
+      end
+
+      # passage = Evidence::Synthetic::TestConstants::NFL_PASSAGE
```
happythenewsad: delete?

services/QuillLMS/engines/evidence/lib/evidence/evidence/synthetic/initial_data.rb
```diff
@@ -0,0 +1,138 @@
...
+      # returns a hash of the form {'csv name' => CSVString, 'csv name2' =>...}
+      def self.generate_csvs(activity_id:, nouns: [])
+        activity = Activity.find(activity_id)
+        passage = activity.passages.first.text
+        prompts = activity.prompts
+        short_name = activity.title.first(20).gsub(' ', '_')
+
+        csvs = {}
+
+        prompts.each.with_index do |prompt, index|
```
happythenewsad: `each_with_index` is available in Ruby 2.7

services/QuillLMS/engines/evidence/lib/evidence/evidence/synthetic/initial_data.rb
```diff
@@ -0,0 +1,138 @@
+# frozen_string_literal: true
+
+module Evidence
+  module Synthetic
+    class InitialData
```
happythenewsad: This class is being used as a runner script, so naming it something like `%Runner%` would be clearer.  This also allows you to rename the `data` variable to `runner` on line 34.

services/QuillLMS/engines/evidence/lib/evidence/evidence/synthetic/initial_data.rb
happythenewsad: Or is this documentation? If so, just preface it with 'How to run' or something.

services/QuillLMS/engines/evidence/lib/evidence/evidence/synthetic/initial_data.rb
```diff
@@ -0,0 +1,138 @@
...
+        csvs = {}
+
+        prompts.each.with_index do |prompt, index|
+          csv_name = "#{short_name}_#{prompt.conjunction}"
+
+          data = new(passage: passage, stem: prompt.text, nouns: nouns)
+          data.run
+
+          csvs[csv_name] = data.generate_csv
+          if index == 0
```
happythenewsad: It's not clear to me why we're using index 0 for this code branch.

happythenewsad (commented):
Looks fine for an MVP. I left some refinement suggestions that you can implement or punt to the next version. 

Also: Have you considered executing this within Jupyter notebook? It would allow users to iterate and view data faster (no need to open individual CSV files, for example).

services/QuillLMS/engines/evidence/lib/evidence/evidence/open_ai/completion.rb
```diff
@@ -0,0 +1,80 @@
...
+        return response if response.present?
+
+        @response = self.class.post(ENDPOINT, body: request_body.to_json, headers: headers)
+
+        cleaned_results
+      end
+
+      def cleaned_results
+        result_texts
+          .map{|r| r.gsub(/^(\n|-|\s)+/, BLANK)} # strip all leading \n, -, or whitespace
```
anathomical, line 60: I don't think this is really a problem, but is the `\s` part of this `gsub` redundant with the `strip` a few steps after?

services/QuillLMS/engines/evidence/lib/evidence/evidence/synthetic/initial_data.rb
anathomical: Should we just move this block to before the `each` loop using `prompts.first.text` when initializing the `data` object?

services/QuillLMS/engines/evidence/lib/evidence/evidence/synthetic/initial_data.rb
```diff
@@ -0,0 +1,138 @@
...
+          .map{|s| s.join(SPACE)}
+          .map{|s| s.last == PERIOD ? s : (s + PERIOD)} # end it in a period, so stem is new sentence.
+      end
+
+      def to_csv(file_path)
+        CSV.open(file_path, "w") do |csv|
+          csv << ['Text', 'Seed']
+          results.each {|r| csv << [r.text, r.seed]}
+        end
+      end
```
anathomical: Rather than generating a CSV object independelty here and in the `generate_csv` function below, can we re-use the other function and make this:
```ruby
def to_csv(file_path)
  File.write(file_path, generate_csv)
end
```

services/QuillLMS/engines/evidence/lib/evidence/evidence/synthetic/initial_data.rb
dandrabik: That's fair. I'm going to rename it to `SeedDataGenerator`. I think that's the most descriptive term for this class and what I'm referring to it in meetings. Thanks for the feedback.

services/QuillLMS/engines/evidence/lib/evidence/evidence/synthetic/initial_data.rb
```diff
@@ -0,0 +1,138 @@
...
+
+      def passages_to_csv(file_path)
+        CSV.open(file_path, "w") do |csv|
+          csv << ['Index', 'Passage Chunk']
+          split_passage.each.with_index {|s,i| csv << [i + 1, s]}
+        end
+      end
```
anathomical: Same as my note above about `to_csv`, can we just write the results of `generate_passage_csv` rather than duplicating the CSV generation loop?

services/QuillLMS/engines/evidence/lib/evidence/evidence/synthetic/initial_data.rb
dandrabik: It's equivalent code. This is just a style thing, but I do like the composable `.with_index` since you can use it with anything, e.g. you can do `map.with_index` but there is no `map_with_index`

anathomical (approved):
Minor thoughts on simplifying the code in a couple of spots, but this generally seems like a solid pattern to establish for synthetic data generation.

services/QuillLMS/engines/evidence/lib/evidence/evidence/synthetic/initial_data.rb
dandrabik: I'll make a comment in the code. This isn't the most elegant code, but I want to generate the passage guide, it's a csv with of the text_chunk indexes and text (for help in analysis). I only need to generate this once, so I'm doing it the first time.

services/QuillLMS/engines/evidence/lib/evidence/evidence/synthetic/initial_data.rb
dandrabik: Fair. I think I'll just remove it and keep it for myself locally.

dandrabik:
> Looks fine for an MVP. I left some refinement suggestions that you can implement or punt to the next version.
> 
> Also: Have you considered executing this within Jupyter notebook? It would allow users to iterate and view data faster (no need to open individual CSV files, for example).

So in a follow-up PR, this is going to plug into a button in the Evidence Admin panel, so I want to keep it in the app. In the future though, I'd be interested to experiment with doing something like this in an jupiter notebook.

services/QuillLMS/engines/evidence/lib/evidence/evidence/open_ai/completion.rb
dandrabik: It's similar, but it has a different effect, so both are needed. 

The first regex is removing starting sequences of newlines,spaces,or dashes, so if the space is removed, strings starting with `" \n"` won't be cleaned properly (especially because of the split on remaining `\n` later:

```ruby
input = "   \n\n this is a sentence \n this is a second one"
# current cleaning 
output = "this is a sentence"
# removing the \s
output = ""

```
The second strip is because some of the things removed can cause extra whitespace.

services/QuillLMS/engines/evidence/lib/evidence/evidence/synthetic/initial_data.rb
dandrabik: That's a good point. I might test that and use that. There's a chance I'm removing that `file_path` code, so will hold off on refactoring now.

services/QuillLMS/engines/evidence/lib/evidence/evidence/synthetic/initial_data.rb
dandrabik: @anathomical Yeah, I did something similar to make it less confusing. Thanks for the feedback.

PR #9440, opened 2022-07-29
Fix bug involving reset_session and some general refactoring

brendanshean:
## WHAT

Note:  I'm tagging everyone on this PR since it the code is in ApplicationController which affects a large segment of our site.  but, specifically note the new `before_action hook :check_staff_for_extended_session` since that will affect all of us.

This PR fixes a [bug](https://www.notion.so/quill/Stuck-in-a-loop-when-attempting-to-sign-into-teacher-account-c67531bea440471ab67b4dd1071b4d31) where staff are unable to sign_in for users that had been inactive for too long.

In the process of uncovering this edge case, I refactored some of the logic, but also noticed a [comment](https://github.com/empirical-org/Empirical-Core/blob/be48add06ab38897bf9df928d8247293ec50697b/services/QuillLMS/app/controllers/application_controller.rb#L150) about staff sign out.

NB.  Unrelated, I've removed the time_difference gem as it appears it is no longer used.

## WHY
If staff tries to login to a user that has not been active in awhile, the staff's session gets reset (not the user they are attempting to impersonate).  This logs the staff out and then confusingly redirects to the sessions/new page.

## HOW
To fix this, we need to a [guard clause](https://github.com/empirical-org/Empirical-Core/blob/0fbe35716f3118af309a8bf911885127e67eb2b1/services/QuillLMS/app/controllers/application_controller.rb#L155) to check for `session[:staff_id]` before resetting the session.

### Screenshots

### Notion Card Links
https://www.notion.so/quill/Stuck-in-a-loop-when-attempting-to-sign-into-teacher-account-c67531bea440471ab67b4dd1071b4d31

PR Checklist | Your Answer
------------ | -------------

### DIFF
--- a/services/QuillLMS/client/webpack.client.dev.base.config.js
+++ b/services/QuillLMS/client/webpack.client.dev.base.config.js
@@ -8,8 +8,8 @@ const railsEnv = process.env.RAILS_ENV || process.env.NODE_ENV
 
 const goFanoutUrl = process.env.GOLANG_FANOUT_URL;
 const pusherKey = process.env.PUSHER_KEY;
-const defaultUrl = process.env.DEFAULT_URL;
-const cdnUrl = process.env.CDN_URL;
+const defaultUrl = process.env.DEFAULT_UR || 'http://localhost:3000'
+const cdnUrl = process.env.CDN_URL || 'https://assets.quill.org'
 const grammarUrl = process.env.QUILL_GRAMMAR_URL || 'http://localhost:3000/grammar/#';
 const lessonsWebsocketsUrl = process.env.LESSONS_WEBSOCKETS_URL || 'http://localhost:3200';
 const quillCmsUrl = process.env.QUILL_CMS || 'http://localhost:3100';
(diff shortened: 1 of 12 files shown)


dandrabik (approved):
:+1: Good sleuthing on this, glad you were able to track this down.

LGTM!

services/QuillLMS/client/webpack.client.dev.base.config.js
```diff
@@ -8,8 +8,8 @@ const railsEnv = process.env.RAILS_ENV || process.env.NODE_ENV
 
 const goFanoutUrl = process.env.GOLANG_FANOUT_URL;
 const pusherKey = process.env.PUSHER_KEY;
-const defaultUrl = process.env.DEFAULT_URL;
-const cdnUrl = process.env.CDN_URL;
+const defaultUrl = process.env.DEFAULT_UR || 'http://localhost:3000'
```
emilia-friedberg: looks like you're missing the final letter of `DEFAULT_URL`

```suggestion
const defaultUrl = process.env.DEFAULT_URL || 'http://localhost:3000'
```

emilia-friedberg (approved):
Looks great. I appreciate all the cleanup - code is way more readable now. One suggestion made where I think you missed a letter.

services/QuillLMS/client/webpack.client.dev.base.config.js
brendanshean: Thank you.  Great eye on that one.

happythenewsad (approved):

eadams17 (approved):

PR #9445, opened 2022-08-01
Fix stripe PurchaserNotFound bug

brendanshean:
## WHAT
Fix an error with StripeIntegration::SubscriptionCreator where PurchaserNotFound exception is raised

## WHY
We'd like subscription creation to go through so that users can use their subscription.

## HOW
Users sometimes have a different email associated with their stripe_customer_id on the stripe website from the email found in our User table.  If we add in a disjunction to check for `stripe_customer_id`, this edge case should be handled.

The first half of this handled already [here](https://github.com/empirical-org/Empirical-Core/blob/develop/services/QuillLMS/app/controllers/stripe_integration/subscription_checkout_sessions_controller.rb#L32) but we need to have  the same call when the webhook is received.

### Screenshots

### Notion Card Links
https://sentry.io/organizations/quillorg-5s/issues/3469766303/?referrer=slack

PR Checklist | Your Answer
------------ | -------------

### DIFF
--- a/services/QuillLMS/app/controllers/stripe_integration/subscription_checkout_sessions_controller.rb
+++ b/services/QuillLMS/app/controllers/stripe_integration/subscription_checkout_sessions_controller.rb
@@ -2,6 +2,8 @@
 
 module StripeIntegration
   class SubscriptionCheckoutSessionsController < ApplicationController
+    class CustomerNotFoundError < StandardError; end
+
     SUBSCRIPTION_MODE = 'subscription'
 
     def create
@@ -30,6 +32,8 @@ module StripeIntegration
 
     private def customer
       User.find_by_stripe_customer_id_or_email!(stripe_customer_id, customer_email)
+    rescue ActiveRecord::RecordNotFound
+      raise CustomerNotFoundError
     end
 
     private def customer_arg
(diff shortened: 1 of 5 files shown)


services/QuillLMS/app/controllers/stripe_integration/subscription_checkout_sessions_controller.rb
```diff
@@ -29,7 +29,7 @@ def create
     end
 
     private def customer
-      User.find_by_stripe_customer_id_or_email(stripe_customer_id, customer_email)
+      User.find_by_stripe_customer_id_or_email!(stripe_customer_id, customer_email)
```
dandrabik, line 34: You might consider using a specific `PurchaserNotFoundError` here as well for tracking (although I don't know if it's really possible to raise that error).

dandrabik (approved):
:+1: Looks good to me. I added a note to take or leave, but looks good either way. Thanks for the quick fix.

services/QuillLMS/app/controllers/stripe_integration/subscription_checkout_sessions_controller.rb
brendanshean: Good idea.

brendanshean: The more I think about it, this condition should be more or less symmetric with call later, so we should actually see this exception here first.

PR #9458, opened 2022-08-02
remove incorrect caching to fix bug

happythenewsad:
## WHAT
Remove caching that was overly greedy, causing the same report to display for different students.

Note: We could also address the issue by caching every single student key. I'd like to see a compelling reason for this though. It would result in this controller action generating a multiplicative number of new keys, which, in a our current caching strategy, would result in many other keys getting evacuated from the cache store. CC @anathomical 

## WHY
Report data should be tailored to individual students, not just the first student that was cached. 

### Screenshots

### Notion Card Links
https://www.notion.so/quill/b417e5b89fe04af89cba62d4198f7031?v=77f294a5b0df466e98633244de7ebfef&p=450b080433ad4a628be360d6044a478e&pm=s

PR Checklist | Your Answer
------------ | -------------
Have you added and/or updated tests? |  Tested manually. This PR is removing logic, not adding it.

### DIFF
--- a/services/QuillLMS/app/controllers/teachers/progress_reports/concepts/concepts_controller.rb
+++ b/services/QuillLMS/app/controllers/teachers/progress_reports/concepts/concepts_controller.rb
@@ -13,11 +13,8 @@ class Teachers::ProgressReports::Concepts::ConceptsController < Teachers::Progre
   end
 
   private def json_payload
-    concepts_json = current_user.all_classrooms_cache(key: 'teachers.progress_reports.concepts.concepts') do
-      concepts_as_json
-    end
     {
-      concepts: concepts_json,
+      concepts: concepts_as_json,
       student: {name: student.name}
     }
   end


services/QuillLMS/app/controllers/teachers/progress_reports/concepts/concepts_controller.rb
```diff
@@ -13,11 +13,8 @@ def index
   end
 
   private def json_payload
-    concepts_json = current_user.all_classrooms_cache(key: 'teachers.progress_reports.concepts.concepts') do
-      concepts_as_json
-    end
     {
-      concepts: concepts_json,
+      concepts: concepts_as_json,
```
happythenewsad, line 17: Note: concepts_as_json is a little confusing. It should be called concepts_as_json_for_student. I will adjust, after review.

anathomical (approved):
Good catch on the overly-broad caching logic.  I think I'd actually be inclined to accept the proliferation of keys and add create a `group` attribute on the `params[:student_id]` value.  I don't think this report is used heavily enough to significantly impact most of the rest of our caching regime, and our cache eviction policy should be good about evicting the least-likely-to-be-used-again values.  I could be wrong, but that's my initial thought.

We could just stop caching this report, but I think, in general, we want all of our reports cached until/unless we have pressing proof that we shouldn't.  So I'm approving this in case we want to get it out fast as a fix, but we should try to get caching at the student level back by the end of the sprint, I think.

happythenewsad:
> Good catch on the overly-broad caching logic. I think I'd actually be inclined to accept the proliferation of keys and add create a `group` attribute on the `params[:student_id]` value. I don't think this report is used heavily enough to significantly impact most of the rest of our caching regime, and our cache eviction policy should be good about evicting the least-likely-to-be-used-again values. I could be wrong, but that's my initial thought.
> 
> We could just stop caching this report, but I think, in general, we want all of our reports cached until/unless we have pressing proof that we shouldn't. So I'm approving this in case we want to get it out fast as a fix, but we should try to get caching at the student level back by the end of the sprint, I think.

Good idea - I'm deploying now but keeping the card open for the additions you mention.

PR #9461, opened 2022-08-02
Lms/write old concept results in workers

anathomical:
## WHAT
Refactor the way we write `OldConceptResult` records and `ConceptResult` records whenever we mark an `ActivitySession` as complete
## WHY
It appears that the old code that was intended to use `bulk_insert` to create a bunch of `OldConceptResult` records and then get their IDs to use in the subsequent creation of `ConceptResult` records was sometimes failing to actually provide IDs.  Without the IDs, the subsequent attempt to create `ConceptResult` records fails.

By moving the code into a worker we can keep the load on the controller small while allowing more flexible ways of handling the insertions than mucking around with `bulk_insert`.
## HOW
- Create a new Worker to handle the creation of `OldConceptResult` records when an `ActivitySession` is completed
- Use that new worker instead of `bulk_insert` to create the `OldConceptResult` records
- Once the `OldConceptResult` records are persisted, pass their IDs into a Worker to create `ConceptResults` for them

### Notion Card Links

PR Checklist | Your Answer
------------ | -------------

### DIFF
--- a/services/QuillLMS/app/workers/save_activity_session_old_concept_results_worker.rb
+++ b/services/QuillLMS/app/workers/save_activity_session_old_concept_results_worker.rb
@@ -1,0 +1,15 @@
+# frozen_string_literal: true
+
+class SaveActivitySessionOldConceptResultsWorker
+  include Sidekiq::Worker
+
+  def perform(concept_results_hashes)
+    ocr_ids = []
+    OldConceptResult.transaction do
+      concept_results_hashes.each do |concept_result_hash|
+        ocr_ids.append(OldConceptResult.create!(concept_result_hash).id)
+      end
+    end
+    SaveActivitySessionConceptResultsWorker.perform_async(ocr_ids)
+  end
+end
--- a/services/QuillLMS/app/controllers/api/v1/activity_sessions_controller.rb
+++ b/services/QuillLMS/app/controllers/api/v1/activity_sessions_controller.rb
@@ -85,12 +85,7 @@ class Api::V1::ActivitySessionsController < Api::ApiController
     concept_results_to_save = @concept_results.map { |c| concept_results_hash(c) }.reject(&:empty?)
     return if concept_results_to_save.empty?
 
-    bulk_inserter = OldConceptResult.custom_bulk_insert(values: concept_results_to_save, return_primary_keys: true)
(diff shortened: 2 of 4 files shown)


emilia-friedberg (approved):

services/QuillLMS/app/workers/save_activity_session_old_concept_results_worker.rb
```diff
@@ -0,0 +1,15 @@
+# frozen_string_literal: true
+
+class SaveActivitySessionOldConceptResultsWorker
+  include Sidekiq::Worker
+
+  def perform(concept_results_hashes)
+    ocr_ids = []
+    OldConceptResult.transaction do
+      concept_results_hashes.each do |concept_result_hash|
+        ocr_ids.append(OldConceptResult.create!(concept_result_hash).id)
```
brendanshean, line 10: I think it would be useful to have a spec that has create! raise an error so that your expectations on     `SaveActivitySessionConceptResultsWorker.perform_async(ocr_ids)` being called are transparent.

brendanshean: Which also kind of raises the question, should failed create! be sent to ErrorNotifier like in    SaveActivitySessionConceptResultsWorker?

brendanshean (commented):
I have a general question about error handling, but I think once that's addressed, this should be good to go.

services/QuillLMS/app/workers/save_activity_session_old_concept_results_worker.rb
anathomical: I've added a spec, so let me know what you think.  As to your follow-up question, no, actually.  Errors are automatically sent to Sentry and NewRelic when they're raised.  In the other worker, we're not actually raising the error (because we wanted to allow some weird retry logic that I suspect won't be necessary.  That is: I don't think that code path ever triggers.  But, short version: `ErrorNotifier` isn't necessary here because we're actually raising the error.

brendanshean: Got it.  Thanks for the explanation.

brendanshean (approved):
Definitely an improvement on the code and nice clear testing.

PR #9462, opened 2022-08-02
Evidence/ add home page materials

eadams17:
## WHAT
update homepage and add Reading for Evidence to tools section

## WHY
we're getting ready to do a full, public launch!

## HOW
various new React components, updates to rails page views/partials, CSS

### Screenshots
Screen Shot 2022-08-02 at 4 26 00 PM
Screen Shot 2022-08-02 at 4 26 14 PM
Screen Shot 2022-08-02 at 4 26 24 PM
Screen Shot 2022-08-02 at 4 26 38 PM

### Notion Card Links
https://www.notion.so/quill/Create-Evidence-Landing-Page-71dbca3375c44031b1b0a0ecca1623be

PR Checklist | Your Answer
------------ | -------------


emilia-friedberg:
@eadams17 This isn't a part of your code, but I was looking at the illustrations of how the activity works and some of the feedback doesn't make sense with what's displayed (the first example on the homepage says "now check the bolded word" but only the stem is bolded; same with the "3rd attempt" on the Evidence tool page). We probably want to fix that before this goes live.

services/QuillLMS/app/views/pages/home_new.html.erb
```diff
@@ -8,15 +8,16 @@
...
-    <h1 class="q-h1 text-white">Free tools to make your students better writers.</h1>
-    <p class="text-white">
-      Quill provides free writing and grammar activities
-      for elementary, middle, and high school students.
-    </p>
+    <section class="header-section">
+      <h1 class="q-h1 text-white">Free tools to make your students better <u>writers</u> and <u>readers</u>.</h1>
+      <p class="new-tag">NEW</p>
+    </section>
+    <p class="text-white" id="first">Quill.org, a non-profit, provides free literacy activities that build reading comprehension, writing, and language skills for elementary, middle, and high school students.</p>
```
emilia-friedberg: I don't think I have when to use ids vs classes and naming strategies for both 100% down pat myself, but an id like `first` feels like a codesmell to me. Even a class like "first" could probably be accomplished with css selectors like `:first-of-type` rather than applying an actual attribute, and beyond that it just feels like too generic of a name to be a unique identifier.

Probably part of a larger discussion but in general, I think my preference would be to move towards not using ids for styling at all, and instead to only use them for things like the "#diagnosticCarousel" below when there is JS or other behavior that relies on them. I find myself thinking along the lines of [this person](https://dev.to/clairecodes/reasons-not-to-use-ids-in-css-4ni4). Maybe you and I can talk about that sometime soon.

services/QuillLMS/app/views/evidence/_first_topic.html.erb
```diff
@@ -0,0 +1,6 @@
+<div class="topic-section">
+  <p class="topic">Culture & Society Topics</p>
+  <img alt="photograph of a football" src="https://assets.quill.org/images/evidence/home_page/EvidenceLandingPage_Image_Football.png"/>
+  <p class="activity-title">"Should Schools Have Grade Requirements for Student Athletes?"</p>
+  <a href='https://www.quill.org/evidence/#/play?uid=180&skipToPrompts=true' target="_blank" class="q-button text-quillteal bg-white">View a Sample Activity</a>
```
emilia-friedberg: This is true for all the "View a Sample Activity" buttons, but they're title-cased while the rest of the buttons on the page are sentence-cased - maybe worth checking with Jack to see if we can make that consistent.

services/QuillLMS/client/app/bundles/Tools/components/evidenceHomeSection.tsx
```diff
@@ -0,0 +1,74 @@
...
+          <img alt="A screenshot of an example Reading for Evidence prompt" className={`evidence-image ${evidenceImageUrl === IMG_URLS[5] ? '' : INACTIVE}`} src={IMG_URLS[5]} />
+          <img alt="A screenshot of an example Reading for Evidence prompt" className={`evidence-image ${evidenceImageUrl === IMG_URLS[4] ? '' : INACTIVE}`} src={IMG_URLS[4]} />
+          <img alt="A screenshot of an example Reading for Evidence prompt" className={`evidence-image ${evidenceImageUrl === IMG_URLS[3] ? '' : INACTIVE}`} src={IMG_URLS[3]} />
+          <img alt="A screenshot of an example Reading for Evidence prompt" className={`evidence-image ${evidenceImageUrl === IMG_URLS[2] ? '' : INACTIVE}`} src={IMG_URLS[2]} />
+          <img alt="A screenshot of an example Reading for Evidence prompt" className={`evidence-image ${evidenceImageUrl === IMG_URLS[1] ? '' : INACTIVE}`} src={IMG_URLS[1]} />
+          <img alt="A screenshot of an example Reading for Evidence prompt" className={`evidence-image ${evidenceImageUrl === IMG_URLS[0] ? '' : INACTIVE}`} src={IMG_URLS[0]} />
```
emilia-friedberg: I think since these are all the same except their index, it would probably be best just to iterate through an array of numbers and have it map to a shared component - something like:

```
IMG_URLS.map(url => (
  <img alt="A screenshot of an example Reading for Evidence prompt" className={`evidence-image ${evidenceImageUrl === url ? '' : INACTIVE}`} src={url} />
)
```

Also, as far as the img alt text goes - I think we should probably either describe these in more detail (with the text of the student's submission and feedback) or just leave it as an empty string.

emilia-friedberg (commented):
Looks great overall. I appreciate the componentization of some of this shared code and I know how onerous it is just to get alllllllll that text in there! Just some clean-up comments and thoughts on ids-vs-classes.

services/QuillLMS/app/views/pages/home_new.html.erb
eadams17: thanks for writing this up. I think I agree-- especially re avoiding `!important` declarations

services/QuillLMS/client/app/bundles/Tools/components/evidenceHomeSection.tsx
eadams17: good catch, I'll make that update

emilia-friedberg (approved):

services/QuillLMS/app/views/pages/evidence_tool.erb
```diff
@@ -0,0 +1,14 @@
+<section class="tool-header">
+  <%= render partial: 'pages/shared/navbar' %>
```
brendanshean, line 2: When rendering partials, I usually prefer the [shorthand](https://guides.rubyonrails.org/layouts_and_rendering.html#using-partials): `<%= render 'pages/shared/navbar' %>` instead of `render partial` which I only use when passing a layout argument.   

Note the section "Rendering the default case" in https://api.rubyonrails.org/classes/ActionView/PartialRenderer.html

brendanshean (approved):
Tremendous work on this!  Excited to see this go out.

PR #9467, opened 2022-08-03
Segment/ add first and last name properties to identify

eadams17:
## WHAT
add first and last name properties to Segment identify calls; this also extracts some user methods and removes part of a check for the `diagnosticLanguageSelected` event that was causing it to not fire

## WHY
this is part of various Segment updates needed to optimize Ortto use

## HOW
extract user specific methods into model, add new properties

### Screenshots

### Notion Card Links
https://www.notion.so/quill/Add-first-name-and-last-name-to-Segment-Identify-call-cb28b2b0a3ec4a75a451dc1ce1ce17fc

PR Checklist | Your Answer
------------ | -------------


services/QuillLMS/app/models/user.rb
```diff
@@ -656,6 +656,28 @@ def last_active_too_long_ago?
...
+      school_name: school&.name,
+      school_id: school&.id,
+      district: school&.district&.name
+    }.reject {|_,v| v.nil? }
+  end
+
+  def segment_integration_rules
+    {
+     all: true,
+     Intercom: (role == 'teacher')
```
brendanshean: very minor nit:  `Intercom: teacher?`

brendanshean (approved):
Looks good and I like the refactoring.

If after some time there ended up being lots of these segment hashes that were coupled to user, you could extract out into a decorator class using SimpleDelegator (e.g. I use that pattern where I want to extract out some methods that apply only to Stripe Subscriptions [StripeIntegration::Subscription](https://github.com/empirical-org/Empirical-Core/blob/develop/services/QuillLMS/app/models/stripe_integration/subscription.rb)).

services/QuillLMS/app/services/analytics/segment_analytics.rb
```diff
@@ -189,7 +189,8 @@ def track_school_subscription(subscription, event)
   def track(options)
     return unless backend.present?
 
-    options[:integrations] = integration_rules(options[:user_id])
+    user = User.find(options[:user_id])
+    options[:integrations] = user&.segment_integration_rules
```
anathomical: I think this is actually different behavior.  In the removed code if `user` is nil, we set a hash for `options[:integrations]` with `{ all: true, Intercom: false }`, but in the new code, if there is no user, this sets `options[:integrations]` to `nil`.  Basically, you're changing the fallback logic for when there is no valid `user` here.  I don't know how often that comes up, but I'd be a bit worried about it being more common than we expect.  Probably solvable with just an `|| default_integration_rules` thing here, where you pre-defined the `default_integration_rules with `Intercom: false`.

services/QuillLMS/client/app/bundles/Diagnostic/components/eslDiagnostic/languagePage.jsx
```diff
@@ -9,7 +9,7 @@ export class LanguagePage extends React.Component {
   handleClickLanguage = (e) => {
     const language = e.currentTarget.value;
     const { dispatch, setLanguage, previewMode, begin } = this.props;
-    if(process.env.NODE_ENV === 'production' && language !== ENGLISH) {
+    if(language !== ENGLISH) {
```
anathomical, line 12: Good catch on the issue here.  We probably actually DO want to be setting `NODE_ENV` on our environments, so in the long run we should probably revisit this, but no point in doing a whole bunch of work around that now.

anathomical (changes requested):
I'd like to see a minor change around integration rules default behavior, but this looks good to me otherwise.

anathomical (approved):
Looks good to me.  Thanks for making the change.

PR #9468, opened 2022-08-03
add onboarding tour to demo account

emilia-friedberg:
## WHAT
Add an onboarding tour to the demo account using `react-joyride`.

## WHY
This is a feature we are hoping will help with new user retention.

## HOW
I added the `react-joyride` library, and a controller action to determine whether a user is currently in a demo account or not. Then I created a central `demo_onboarding_tour` component which uses `localStorage` and that controller action to determine whether or not to display on the pages where it is included.

### Screenshots
Screen Shot 2022-08-03 at 12 04 27 PM
Screen Shot 2022-08-03 at 12 04 35 PM
Screen Shot 2022-08-03 at 12 04 42 PM
Screen Shot 2022-08-03 at 12 04 48 PM
Screen Shot 2022-08-03 at 12 04 57 PM
Screen Shot 2022-08-03 at 12 05 03 PM
Screen Shot 2022-08-03 at 12 05 12 PM

### Notion Card Links
https://www.notion.so/quill/Demo-Account-Guide-Tour-Built-by-Quill-defc79da9ed040ccb01aec26c5eb5568

PR Checklist | Your Answer
------------ | -------------


services/QuillLMS/app/controllers/teachers/classroom_manager_controller.rb
```diff
@@ -244,6 +244,12 @@ def unset_view_demo
     redirect_to '/profile'
   end
 
+  # rubocop:disable Naming/PredicateName
+  def is_demo
```
brendanshean, line 248: I'm kind of late to the party here, but can you maybe say a little bit about what the difference between `view_demo` and `is_demo` are?  (My gut says this method/label should be `demo` and this is kind of what rubocop is advocating as well)

brendanshean: In general, the Naming/PredicateName is a pretty solid cop.  There's some discussion [here](https://github.com/rubocop/rubocop/issues/4479) concerning it's application to Rails controllers, but I would agree with their conclusion.  In this case,  `is_demo` conveys boolean semantics but all that's being passed is a hash; the actual boolean check is in the [component](https://github.com/empirical-org/Empirical-Core/pull/9468/files#diff-ef59189626d7a911f9682a2800e3f0e94968aea1b92a8dbbedc38e4820bba37bR147) itself.   What about `def current_user_demo_id` instead?

brendanshean (approved):
This looks really cool.  I have a minor comment about the naming of a method.  Given that there's quite a bit of front-end code, I'll would recommend at least one more approval.

services/QuillLMS/app/controllers/teachers/classroom_manager_controller.rb
anathomical: What if we just called this action `demo_id` or something?

services/QuillLMS/client/app/bundles/Teacher/components/shared/demo_onboarding_tour.tsx
```diff
@@ -0,0 +1,177 @@
...
+
+import useWindowSize from '../../../Shared/hooks/useWindowSize'
+import { requestGet } from '../../../../modules/request/index';
+
+export const DASHBOARD = 'demo-onboarding-dashboard'
+export const STUDENT_REPORTS_LANDING_PAGE = 'demo-onboarding-student-reports-landing-page'
+export const DIAGNOSTIC_ACTIVITY_PACKS_INDEX = 'demo-onboarding-diagnostic-activity-packs-index'
+export const DIAGNOSTIC_RESULTS_SUMMARY = 'demo-onboarding-diagnostic-results-summary'
+export const DIAGNOSTIC_RECOMMENDATIONS = 'demo-onboarding-diagnostic-recommendations'
+export const DIAGNOSTIC_GROWTH_SUMMARY = 'demo-onboarding-diagnostic-growth-summary'
```
anathomical: This probably isn't actually a problem since it becomes clear what these values are when you actually import them into another file, but I wanted to call out that if you do import one of these values, the name stops being clearly-related to the demo.  (You can just end up referencing `DIAGNOSTIC_GROWTH_SUMMARY` in a component without it being obvious that the value is related to the demo.)

So this can likely just be left as-is, but I wanted to flag it in case you felt it was a valid concern.

anathomical (approved):
I've flagged a minor item that is probably just fine.  Other than that, I wanted to say that this looks great!

eadams17 (approved):

PR #9499, opened 2022-08-09
Segment/ add teacher signed in properties

eadams17:
## WHAT
add teacher signed in properties to Segment track calls

## WHY
we need these traits for tracking in Orrto

## HOW
update how we handle Segment track calls for when a teacher signs in; this also adds a SimpleDelegator to help keep the User model skinnier 

### Screenshots

### Notion Card Links
https://www.notion.so/quill/Add-properties-to-Teacher-signed-in-Segment-track-event-618106297ba347a2bd090ca78a3e9fda

PR Checklist | Your Answer
------------ | -------------

### DIFF
--- a/services/QuillLMS/app/models/user.rb
+++ b/services/QuillLMS/app/models/user.rb
@@ -658,8 +658,8 @@ class User < ApplicationRecord
     last_active < USER_INACTIVITY_DURATION.ago
   end
 
-  def segment_properties
-    SegmentProperties::User.new(self)
+  def segment_integration
+    SegmentIntegration::User.new(self)
   end
 
   private def validate_flags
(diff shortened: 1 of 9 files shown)


anathomical (approved):
Looks good.  I like pulling the Segment stuff into a delegator.  `User` is already too big.

services/QuillLMS/app/models/user.rb
```diff
@@ -658,26 +658,8 @@ def last_active_too_long_ago?
...
-      district: school&.district&.name
-    }.reject {|_,v| v.nil? }
-  end
-
-  def segment_integration_rules
-    {
-     all: true,
-     Intercom: (teacher?)
-    }
+  def segment_integration
```
brendanshean: Naming is always a hard problem.  I would name this `segment_user` instead of the namespace `segment_integration` since this would pose a naming problem if in the future another SegmentIntegration delegator were introduced to the user model.  For example,

```
def segment_foo
  SegmentIntegration::Foo.new(self)
end
```

brendanshean (approved):
One minor comment on naming, but otherwise looks good!

PR #9513, opened 2022-08-11
Add stripe webhook handler for disputed.charged.created event

brendanshean:
## WHAT
Add a stripe webhook event handler for `disputed.charge.created` event types

## WHY
We'd like to communicate to sales/support when these disputes arise.

## HOW
Add a handler and then send email to hello@quill.org

### Screenshots

### Notion Card Links

PR Checklist | Your Answer
------------ | -------------


services/QuillLMS/app/mailers/application_mailer.rb
```diff
@@ -1,5 +1,5 @@
 # frozen_string_literal: true
 
 class ApplicationMailer < ActionMailer::Base
-
+  QUILL_TEAM_EMAIL_ADDRESS = 'hello@quill.org'
```
anathomical, line 4: Probably not worth futzing with in this PR, but we may want to set up some specific email address for this kind of issue.

anathomical (approved):
Minor comment about sending every email to `hello@quill.org`, but this looks good to me.

services/QuillLMS/app/mailers/application_mailer.rb
brendanshean: While I do agree with the smell of using the same email everywhere, Stripe subscriptions has been on since May and we've only had one of these errors, so the volume should be very low.

PR #9527, opened 2022-08-12
Lms/use concept results in raw sql

anathomical:
## WHAT
Update our code that reference `old_concept_results` via raw SQL queries to reference the new `concept_results` table and the correct location for the queried data in the new structure.
## WHY
We want to remove the `old_concept_results` table when we can, and this is a good first step
## HOW
- Update the various raw SQL references to `old_concept_results` to point to `concept_results` instead
- Tweak any column references to look at where the data is in the new schema over the old schema
- Find all the failing tests and update their setup conditions to instantiate `ConceptResult` models instead of `OldConceptResult` models

### Notion Card Links
https://www.notion.so/quill/Add-code-to-read-ConceptResult-data-from-Responses-records-d03eaa99d4a545cfb9e4d9a8d18bc4c8

PR Checklist | Your Answer
------------ | -------------

### DIFF
--- a/services/QuillLMS/app/models/activity_session.rb
+++ b/services/QuillLMS/app/models/activity_session.rb
@@ -69,7 +69,7 @@ class ActivitySession < ApplicationRecord
   has_many :concept_results
   has_many :old_concept_results
   has_many :teachers, through: :classroom
-  has_many :concepts, -> { distinct }, through: :old_concept_results
+  has_many :concepts, -> { distinct }, through: :concept_results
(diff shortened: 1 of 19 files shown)


brendanshean (approved):
Nice work on this. The newer model seems like a much cleaner interface which is reflected in the thorough testing.

emilia-friedberg (approved):
Looks great. So nice to not have to do all that json casting.

happythenewsad (approved):
LGTM

PR #9530, opened 2022-08-15
A few stripe integration updates

brendanshean:
## WHAT
1.  Remove guard clause for nil purchaser email
2. Add some webhook event types to ignore list
3. Add ENV configuration for stripe notifications mailer

## WHY
1.  This should have been removed in an earlier PR. 
2. This will result in less noise in our handling of webhooks
3. We might want this email to go somewhere other than hello@quill.org 

## HOW
1.  Delete the associated guard clause
2. Add the items to the IGNORED_EVENT_TYPES array
3. Set the to: field equal to ENV['STRIPE_NOTIFICATIONS_EMAIL'

### Screenshots

### Notion Card Links

PR Checklist | Your Answer
------------ | -------------

### DIFF
--- a/services/QuillLMS/app/mailers/stripe_integration/mailer.rb
+++ b/services/QuillLMS/app/mailers/stripe_integration/mailer.rb
@@ -8,7 +8,7 @@ module StripeIntegration
     CHARGE_DISPUTE_CREATED_URL = "#{STRIPE_DASHBOARD_URL}/disputes?statuses[0]=needs_response"
 
     def charge_dispute_created
-      mail(to: QUILL_TEAM_EMAIL_ADDRESS, subject: 'Charge Dispute Created') do |format|
+      mail(to: ENV.fetch('STRIPE_NOTIFICATIONS_EMAIL'), subject: 'Charge Dispute Created') do |format|
         format.text do
           render plain: "A response is needed on a disupted stripe subscription charge: #{CHARGE_DISPUTE_CREATED_URL}"
         end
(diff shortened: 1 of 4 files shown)


services/QuillLMS/app/mailers/stripe_integration/mailer.rb
```diff
@@ -8,7 +8,7 @@ class Mailer < ::ApplicationMailer
     CHARGE_DISPUTE_CREATED_URL = "#{STRIPE_DASHBOARD_URL}/disputes?statuses[0]=needs_response"
 
     def charge_dispute_created
-      mail(to: QUILL_TEAM_EMAIL_ADDRESS, subject: 'Charge Dispute Created') do |format|
+      mail(to: ENV.fetch('STRIPE_NOTIFICATIONS_EMAIL'), subject: 'Charge Dispute Created') do |format|
```
anathomical: This isn't a huge deal, but I think we've been moving to sort of proxy `ENV` variables as class constants.  I don't know that it's better, but consistency is good, so how about we just create a `STRIPE_NOTIFICATION_EMAIL = ENV.fetch('STRIPE_NOTIFICATION_EMAIL')` constant.  Then the spec can stub the class constant rather than ENV.

anathomical (approved):
Minor note about style, but this looks good to me as-is.

services/QuillLMS/app/mailers/stripe_integration/mailer.rb
brendanshean: Good point.  I will update.

PR #9538, opened 2022-08-17
Segment/ add flags property

eadams17:
## WHAT
add flags property to Segment identify calls

## WHY
we want to track this property for our users

## HOW
just add this property to the `identify_params` method

### Screenshots

### Notion Card Links
https://www.notion.so/quill/Add-new-flags-trait-to-Segment-Identify-call-6432214d04ac4c61b5e9eed569d19143

PR Checklist | Your Answer
------------ | -------------
Back-to-school 2022: Have you checked the [webinar schedule](https://www.notion.so/quill/Back-to-school-webinar-banners-2022-a75a89cfad9f434899ef6be3eb184733) to avoid for downtime/risky deploys? | yes

### DIFF
--- a/services/QuillLMS/spec/models/segment_integration/user_spec.rb
+++ b/services/QuillLMS/spec/models/segment_integration/user_spec.rb
@@ -3,23 +3,29 @@
 require 'rails_helper'
 
 RSpec.describe SegmentIntegration::User do
-  let(:teacher) { create(:teacher) }
+  let(:teacher1) { create(:teacher) }
+  let(:teacher2) { create(:teacher, flags: ["private", "beta"]) }
 
   context '#identify_params' do
 
     it 'returns the expected params hash' do
       params = {
-        user_id: teacher.id,
+        user_id: teacher1.id,
         traits: {
-          **teacher.segment_user.common_params,
-          auditor: teacher.auditor?,
-          first_name: teacher.first_name,
-          last_name: teacher.last_name,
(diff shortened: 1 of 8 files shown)


brendanshean (approved):

services/QuillLMS/spec/models/segment_integration/user_spec.rb
```diff
@@ -3,45 +3,51 @@
 require 'rails_helper'
 
 RSpec.describe SegmentIntegration::User do
-  let(:teacher) { create(:teacher) }
+  let(:teacher1) { create(:teacher) }
+  let(:teacher2) { create(:teacher, flags: ["private", "beta"]) }
```
anathomical: Super-minor, but do you need to create a second teacher instead of just setting `flags` on the original model?  It doesn't look like there are any `it` cases below where you want to check both models and see different behavior, so you could probably just make this one model.  It doesn't really matter, though, so no need to clean it up.

anathomical (approved):
One note, but no reason to change anything unless you want to.

services/QuillLMS/spec/models/segment_integration/user_spec.rb
brendanshean: I thought about suggesting this too.  On second thought, I think it should be cleaned up.

Having teacher1 and teacher2 is frequently an indicator that you'll have a test that uses both teachers in the same spec.

PR #9542, opened 2022-08-17
lms/allow-recommendations-for-one-name-students

anathomical:
## WHAT
Allow recommendations to generate for students with only one name
## WHY
The original code crashed when trying to generate recommendations for a classroom that contained any students with only one name (where `.split[1]` produced `nil`), but since this sometimes happens in real classrooms, we should make sure this doesn't crash instead.
## HOW
In cases where we can't naively identify a last name via `name.split[1]`, use `""` as the sort key instead to sort single-name students to the beginning of the list.

### Notion Card Links
https://www.notion.so/quill/Spinning-wheel-on-diagnostic-recommendations-page-e0091fd3cf154959bb32ed60e85a6940

PR Checklist | Your Answer
------------ | -------------

### DIFF
--- a/services/QuillLMS/app/models/concerns/public_progress_reports.rb
+++ b/services/QuillLMS/app/models/concerns/public_progress_reports.rb
@@ -282,7 +282,7 @@ module PublicProgressReports
       {id: s&.id, name: s&.name || "Unknown Student", completed: completed }
     end
 
-    sorted_students = students.compact.sort_by {|stud| stud[:name].split()[1] || ''}
+    sorted_students = students.compact.sort_by {|stud| stud[:name].split().second || ''}
 
     recommendations = RecommendationsQuery.new(diagnostic.id).activity_recommendations.map do |activity_pack_recommendation|
       students = []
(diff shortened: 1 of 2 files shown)


services/QuillLMS/app/models/concerns/public_progress_reports.rb
```diff
@@ -282,7 +282,7 @@ def generate_recommendations_for_classroom(current_user, unit_id, classroom_id,
       {id: s&.id, name: s&.name || "Unknown Student", completed: completed }
     end
 
-    sorted_students = students.compact.sort_by {|stud| stud[:name].split()[1]}
+    sorted_students = students.compact.sort_by {|stud| stud[:name].split()[1] || ''}
```
brendanshean: I think `split()[1]` could be rewritten `split.second`

brendanshean (approved):
Looks good.  Thanks for cleaning this up

eadams17 (approved):

emilia-friedberg (approved):

PR #9545, opened 2022-08-17
remove app setting conditionals from blogs views

happythenewsad:
## WHAT
remove Evidence conditionals in teacher center static pages 

## WHY
Everyone should see Evidence-related content. Even if an AppSetting is enabled for all, it will still return false if a user does not exist (i.e. is not logged in).

### Screenshots

### Notion Card Links
no - rush order from PG

PR Checklist | Your Answer
------------ | -------------
Back-to-school 2022: Have you checked the [webinar schedule](https://www.notion.so/quill/Back-to-school-webinar-banners-2022-a75a89cfad9f434899ef6be3eb184733) to avoid for downtime/risky deploys? | (Yes or N/A)

### DIFF
--- a/services/QuillLMS/app/helpers/teacher_center_helper.rb
+++ b/services/QuillLMS/app/helpers/teacher_center_helper.rb
@@ -8,17 +8,17 @@ module TeacherCenterHelper
   ALL = 'All'
 
   def teacher_center_tabs(large: true)
-    comprehension_tab = {
-      id: BlogPost::USING_QUILL_FOR_READING_COMPREHENSION,
-      name: COMPREHENSION,
-      url: 'teacher-center/topic/using-quill-for-reading-comprehension'
-    }
     premium_tab = {
       id: PREMIUM,
       name: PREMIUM,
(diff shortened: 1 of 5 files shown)


services/QuillLMS/app/helpers/teacher_center_helper.rb
```diff
@@ -46,7 +45,7 @@ def teacher_center_tabs(large: true)
         url: 'faq'
       }
     ]
-    tabs.insert(1, comprehension_tab) if is_comprehension_user
+    tabs.insert(1, comprehension_tab)
```
emilia-friedberg: Since there is no longer a conditional here it could just be added to the array directly.

emilia-friedberg (approved):
One small nit but LGTM overall.

services/QuillLMS/app/helpers/teacher_center_helper.rb
happythenewsad: Good catch; done

PR #9574, opened 2022-08-22
lms/sort-classrooms-by-order

anathomical:
## WHAT
Tweak sort order to work in cases where some classrooms are deleted
## WHY
We were running into issues in production where some teachers had classrooms that had been deleted (marked visible: false) before they ordered their classrooms.  This resulted in a mix of ordered and unordered classrooms which caused order to be ignored for sorting.
## HOW
Instead of using a ternary to sort two different ways, just combine the two sorting systems so that cases where there are mixed systems, it still sorts as expected.

### Notion Card Links
https://www.notion.so/quill/Reordering-classes-doesn-t-save-ea53232924f74743a7a7ead5700a4074

PR Checklist | Your Answer
------------ | -------------


dandrabik (approved):
:+1: This seems like a reasonable fix.

PR #9575, opened 2022-08-22
Add rake tasks to copy staging configs to all staging environments.

dandrabik:
## WHAT
Add a rake task that allows you to:
1. Set a config and value for all staging environments
2. Remove a config from all staging environments
3. Copy a config on main staging to the rest of the staging environments
## WHY
We have a lot of staging environments and are changing configs a good amount. That can lead to broken environments, manual fixes, or just drift from the correct settings.
## HOW
Simple wrapper around the heroku commands.
### Screenshots
Screen Shot 2022-08-22 at 12 04 39 PM

PR Checklist | Your Answer
------------ | -------------
Have you added and/or updated tests? |  No, just going to write rake tasks from now on and live on the edge.
Have you deployed to Staging? |  No deploy, but I set and unset staging vars

### DIFF
--- a/services/QuillLMS/lib/tasks/staging_config.rake
+++ b/services/QuillLMS/lib/tasks/staging_config.rake
@@ -6,7 +6,7 @@ namespace :staging_config do
 
   # bundle exec rake staging_config:copy_from_staging_to_personals\[SOME_KEY\]
   desc "copy a config from main staging to all other staging envs"
-  task :copy_from_staging_to_personals, [:config] do |t,args|
+  task :copy_from_staging_to_personals, [:config] => :environment do |t,args|
     include StagingConfigCommands
 
     config = args[:config]
@@ -21,7 +21,7 @@ namespace :staging_config do
 
   # bundle exec rake staging_config:set_all\[TEST_KEY='some_value'\]
   desc "set a config key/value on all staging envs"
-  task :set_all, [:config_setting] do |t,args|
+  task :set_all, [:config_setting] => :environment do |t,args|
     include StagingConfigCommands
     config_setting = args[:config_setting]
 
@@ -32,7 +32,7 @@ namespace :staging_config do
 
   # bundle exec rake staging_config:remove_all\[TEST_KEY\]
   desc "delete a config from all staging envs"
-  task :remove_all, [:config] do |t,args|
+  task :remove_all, [:config] => :environment do |t,args|
     include StagingConfigCommands
     config = args[:config]
 
@@ -58,10 +58,10 @@ namespace :staging_config do
     end
 
     def run_cmd(command)
-      stdout_str, stderr_str, _ = Open3.capture3(command)
+      stdout_str, stderr_str, = Open3.capture3(command)
 
-      puts "#{stdout_str}"
-      puts "#{stderr_str}"
+      puts stdout_str
+      puts stderr_str
 
       stdout_str
     end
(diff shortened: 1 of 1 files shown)


anathomical (approved):
This looks good to me, and I agree that we're doing this often enough that we should have a tool like this.  One super-minor thing to note: the screenshot looks like it's the signature of `set_all`, but it says it's a call to `set_all`.  Probably just a run of a different version of the code, so no big deal.

dandrabik:
> This looks good to me, and I agree that we're doing this often enough that we should have a tool like this. One super-minor thing to note: the screenshot looks like it's the signature of `set_all`, but it says it's a call to `set_all`. Probably just a run of a different version of the code, so no big deal.

I'm not sure I understand your comment (maybe there is a typo?). The screenshot should be accurate for running the rake task to copy from main staging to the rest of the staging apps.

services/QuillLMS/lib/tasks/staging_config.rake
```diff
@@ -0,0 +1,82 @@
...
+    STAGING_APP = 'empirical-grammar-staging'
+    STAGING_PERSONAL_APPS = %w(
+      quill-lms-brendan
+      quill-lms-cissy
+      quill-lms-dan
+      quill-lms-emilia
+      quill-lms-eric
+      quill-lms-pkong
+      quill-lms-thomas
+      empirica-grammar-security
```
emilia-friedberg, line 78: I think this is a typo (should be `empirical-grammar-security`)

emilia-friedberg (approved):
This is great. One small note on what I think is a typo but otherwise LGTM. Thanks for doing that.

anathomical:
> > This looks good to me, and I agree that we're doing this often enough that we should have a tool like this. One super-minor thing to note: the screenshot looks like it's the signature of `set_all`, but it says it's a call to `set_all`. Probably just a run of a different version of the code, so no big deal.
> 
> I'm not sure I understand your comment (maybe there is a typo?). The screenshot should be accurate for running the rake task to copy from main staging to the rest of the staging apps.

I think I figured it out.  The last character is, I suspect, cut off in the screenshot.  So your example command ends with `\[ZZZ_TEST_KEY\]`, and that final slash is not a new-line enabler of some sort.  In turn, `dancedance123` is output from the script, not part of the input.  Sorry about the confusion.

services/QuillLMS/lib/tasks/staging_config.rake
dandrabik: Good catch! 

Is is a typo, but it's a typo in the name of the app (I copied the list from a CLI command)
Screen Shot 2022-08-22 at 12 53 52 PM

dandrabik:
> > > This looks good to me, and I agree that we're doing this often enough that we should have a tool like this. One super-minor thing to note: the screenshot looks like it's the signature of `set_all`, but it says it's a call to `set_all`. Probably just a run of a different version of the code, so no big deal.
> > 
> > 
> > I'm not sure I understand your comment (maybe there is a typo?). The screenshot should be accurate for running the rake task to copy from main staging to the rest of the staging apps.
> 
> I think I figured it out. The last character is, I suspect, cut off in the screenshot. So your example command ends with `\[ZZZ_TEST_KEY\]`, and that final slash is not a new-line enabler of some sort. In turn, `dancedance123` is output from the script, not part of the input. Sorry about the confusion.

Ah, I see, yeah, I should have not cropped so close to the edge. And for rake tasks using parameters the form is:

`rake namespace:command[param1, param2]`
 
but in `.zsh` you have to escape the `[` and `]`, so you end up with this less pretty version:

`rake namespace:command\[param1, param2\]`

PR #9579, opened 2022-08-22
lms/show-evidence-completed-in-activity-feed

anathomical:
## WHAT
Ensure that Evidence activities show as 'completed' in activity feed
## WHY
We want the Activity Feed to accurately show Evidence activities as 'completed' the same way we do for Diagnostic activities
## HOW
There was an early return to protect this code from errors with calculating stuff around percentages, but it was accidentally early-returning on Evidence activities (which we explicitly don't record a `percentage` value for).  This wasn't caught in existing specs because the factory to create Evidence activity sessions wasn't explicitly using `nil` for percentage.  Updated the factory to confirm existing specs would catch this condition, then moved the early return to after the `unscored?` check.

### Notion Card Links
https://www.notion.so/quill/Show-Completed-in-the-activity-feed-for-Evidence-activities-39a3be9a81ea436885b9610b3ae9fc25

PR Checklist | Your Answer
------------ | -------------

### DIFF
--- a/services/QuillLMS/spec/factories/activity_sessions.rb
+++ b/services/QuillLMS/spec/factories/activity_sessions.rb
@@ -106,6 +106,8 @@ FactoryBot.define do
 
     factory :evidence_activity_session do
       activity { create(:evidence_activity) }
+      # We explicitly don't record percentraes for Evidence sessions
+      percentage { nil }
     end
 
     factory :proofreader_activity_session do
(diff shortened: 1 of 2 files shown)


brendanshean (approved):
Thanks for fixing this!

services/QuillLMS/spec/factories/activity_sessions.rb
```diff
@@ -106,6 +106,8 @@
 
     factory :evidence_activity_session do
       activity { create(:evidence_activity) }
+      # We explicitly don't record percentraes for Evidence sessions
```
brendanshean: There's a typo with percentages

emilia-friedberg (approved):

PR #9585, opened 2022-08-23
lms/handle-vitally-sync-rate-limit

anathomical:
## WHAT
Allow SyncOrganization worker to handle rate-limiting from the API
## WHY
Organizations (`District` records in our database) are synced to Vitally via a different API than users or schools.  This different API has rate-limiting built in rate limiting.  Our limited rate is 1000 items per minute, and resets 60 seconds after the first request, but the API is fast and we're currently attempting to sync 12k items in this 60 second window without checking to see if they succeed.
## HOW
The new code examines the result of our API call, and if it is rate limited (as indicated by a 429 response code, according to Vitally documentation: https://docs.vitally.io/pushing-data-to-vitally/rest-api#rate-limiting) we schedule a retry to happen randomly between 2 and 20 minutes in the future.  The two minute floor should ensure we don't retry until after our current rate limit window passes, and allowing reschedules out to 20 minutes should, on average, allow us to submit around 19~20k records that don't trigger a second rate-limit response.  Given that we only have around 12k records to sync at the moment, that should give us plenty of overhead.

### Notion Card Links
https://www.notion.so/quill/Modify-SyncVitallyOrganizationsWorker-to-gracefully-handle-rate-limiting-from-Vitally-s-new-API-fe0fd9f61d26402786109da6aa697bf6

PR Checklist | Your Answer
------------ | -------------

### DIFF
--- a/services/QuillLMS/app/workers/sync_vitally_organization_worker.rb
+++ b/services/QuillLMS/app/workers/sync_vitally_organization_worker.rb
@@ -10,9 +10,10 @@ class SyncVitallyOrganizationWorker
     district = District.find(district_id)
     api = VitallyRestApi.new
     response = api.create("organizations", district.vitally_data)
-    if request_rate_limited?(response)
-      requeue_after_rate_limit(district_id)
-    end
+
+    return unless request_rate_limited?(response)
+
+    requeue_after_rate_limit(district_id)
   end
 
   private def request_rate_limited?(response)
--- a/services/QuillLMS/app/workers/sync_vitally_organization_worker.rb
+++ b/services/QuillLMS/app/workers/sync_vitally_organization_worker.rb
@@ -3,6 +3,8 @@
 class SyncVitallyOrganizationWorker
   include Sidekiq::Worker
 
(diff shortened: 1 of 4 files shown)


services/QuillLMS/app/workers/sync_vitally_organization_worker.rb
```diff
@@ -3,9 +3,25 @@
 class SyncVitallyOrganizationWorker
```
happythenewsad, line 3: Looks good. There's a chance that even with this enhancement, jobs will continue to "fail forward" for a long time. We won't be able to diagnose this from the job retry count, since the soft failure path in this PR leads to the creation of a fresh job. 

I'd like to consider:
- widening the requeue time to the max acceptable by stakeholders
- adding New Relic or some other alerting, so that we can track "long-failing" jobs

happythenewsad (commented):
Very reasonable solution. However, it allows the possibility for jobs to get rolled forward indefinitely due to 429 issues

happythenewsad (approved):
Great additions, LGTM!

PR #9586, opened 2022-08-23
fix all but one package.json critical/high security issue

emilia-friedberg:
## WHAT
Fix all but one of the [high and critical security issues](https://www.notion.so/quill/Javascript-security-updates-c0590a4d12884ae6b85a05f4bf8dae12#94cf6a7765d047cca348fe6f04c7f0b1) in our `package.json` files.

## WHY
We don't want major security issues in our code.

## HOW
Just bump the versions of the affected libraries where possible. The only one that wasn't possible to do was `scss-tokenizer`, which is a dependency of our installed library `node-sass`. `node-sass` [has been deprecated](https://sass-lang.com/blog/libsass-is-deprecated), and the last released version still relies on an unpatched version of `scss-tokenizer`, so we either need to replace it or use the override available in the latest version of `npm` once we've set that up. I'm making a card to do the former, but if it proves very complicated for some reason we may want to fall back to the latter.

### Screenshots

### Notion Card Links
https://www.notion.so/quill/Javascript-security-updates-c0590a4d12884ae6b85a05f4bf8dae12

PR Checklist | Your Answer
------------ | -------------


happythenewsad (approved):
LGTM

anathomical (approved):

PR #9594, opened 2022-08-25
Flagsets - backend

happythenewsad:
## WHAT
- adds Flagset concern on User. Flagsets are sets of flags.
- When deployed, this code won't be active until the front end dropdowns are updated (subsequent PR).
- 
## WHY
- required for Flagset project 

## HOW
- add Concern
- add migration

### Screenshots

### Notion Card Links
https://www.notion.so/quill/Remove-Flag-Hierarchy-Create-New-Beta-Flags-to-Enable-Reading-for-Evidence-Testing-afd471842d74402d9a616545f8917d9b

PR Checklist | Your Answer
------------ | -------------
Back-to-school 2022: Have you checked the [webinar schedule](https://www.notion.so/quill/Back-to-school-webinar-banners-2022-a75a89cfad9f434899ef6be3eb184733) to avoid for downtime/risky deploys? |  N/A)


services/QuillLMS/app/models/firebase_app.rb
```diff
@@ -8,6 +8,7 @@
 #  name       :string
 #  pkey       :text
 #  secret     :string
+#  throwaway  :text             default("lorem")
```
happythenewsad, line 11: Unrelated annotation

services/QuillLMS/app/models/user_activity_classification.rb
```diff
@@ -13,7 +13,7 @@
 #
 #  index_user_activity_classifications_on_classifications  (activity_classification_id)
 #  index_user_activity_classifications_on_user_id          (user_id)
-#  user_activity_classification_unique_index               (user_id,activity_classification_id)
+#  user_activity_classification_unique_index               (user_id,activity_classification_id) UNIQUE
```
happythenewsad, line 16: unrelated annotation

services/QuillLMS/app/models/concerns/flagset.rb
```diff
@@ -0,0 +1,78 @@
...
+        display_name: 'production',
+        flags: {
+          'production' =>        { display_name: 'Production' }
+        }
+      },
+
+      alpha: {
+        display_name: 'Alpha',
+        flags: {
+          'alpha' =>             { display_name: 'Alpha' },
```
happythenewsad: Using string hash keys here to avoid `to_s` elsewhere

services/QuillLMS/app/models/concerns/flagset.rb
```diff
@@ -0,0 +1,78 @@
...
+
+# This concern is only intended to be included in the User model
+module Flagset
+  #FLAGS = %w(production archived alpha beta gamma private)
+  extend ActiveSupport::Concern
+
+
+  included do
+    FLAGSETS = {
+
```
brendanshean: I think it would be worth adding some constants for the Flagset keys.

```
PRODUCTION_FLAGSET = 'production'
ALPHA_FLAGSET = 'alpha'
EVIDENCE_BETA1_FLAGSET = 'evidence_beta1'
...
```

Also, I would introduce the following constants in flag.rb and reference them in flagset.rb:
```
PRODUCTION_FLAG = 'production'
ALPHA_FLAG = 'alpha'
COLLEGE_BOARD = 'college_board'
...
```

that way there's a single source of truth in the code for these flag and flagset names.

services/QuillLMS/app/models/concerns/flagset.rb
```diff
@@ -0,0 +1,78 @@
...
+      }
+    }
+
+    validates :flagset, inclusion: { in: FLAGSETS.keys.map(&:to_s) }
+  end
+
+  def activity_viewable?(activity)
+    return false unless activity.is_a?(::Activity)
+
+    user_flagset = FLAGSETS.fetch(flagset.to_sym){ return false }
```
brendanshean: The variable `user_flagset` indicates a coupling to the user model.  If the intent is only to use this as a user concern, I would rename it `UserFlagset`.  If this model is meant to be included elsewhere, I would rename the `user_flagset` variable to something more generic.

On a related note is `user_flagset` necessary at all since there is     
`validates :flagset, inclusion: { in: FLAGSETS.keys.map(&:to_s) }`.  In other words shouldn't the fetch always succeed?

brendanshean (approved):
Nice work on this.  I left a couple of minor comments but this doesn't need a re-review after addressing.

services/QuillLMS/app/models/concerns/flagset.rb
happythenewsad: > On a related note is `user_flagset` necessary at all since there is `validates :flagset, inclusion: { in: FLAGSETS.keys.map(&:to_s) }`. In other words shouldn't the fetch always succeed?

The flagsets will change in the future, I prefer to keep the guard here.

services/QuillLMS/app/models/concerns/user_flagset.rb
```diff
@@ -0,0 +1,77 @@
...
+        }
+      },
+
+      evidence_beta2: {
+        display_name: 'Evidence Beta 2',
+        flags: {
+          Flags::EVIDENCE_BETA1 =>    { display_name: 'Evidence Beta 1' },
+          Flags::PRODUCTION =>        { display_name: 'Production' }
+        }
+      },
```
anathomical: This one is confusingly-named if it's accurate.  The flagset is called `beta2`, but it gives access to `beta1`?  We either need a better name, or this may be supposed to give access to `beta2`?

anathomical (approved):
One note on a confusingly-named/configured flagset which is probably worth cleaning up, but other than that this looks good to me.

services/QuillLMS/app/models/concerns/user_flagset.rb
happythenewsad: Address, thanks.

PR #9600, opened 2022-08-25
Resolve existing users with both google_id and clever_id

brendanshean:
## WHAT
There are some users with both a `google_id` and a `clever_id`.  We'd like to sensibly set one of those values to nil.

## WHY
There's an issue with the `reset_session` [logic](https://github.com/empirical-org/Empirical-Core/blob/develop/services/QuillLMS/app/controllers/application_controller.rb#L149) and google authorization where a teacher has switched over a classroom of students to clever but the students have an existing auth_credential pointing to google.  This results in a loop where students cannot login.

## HOW
There are three strategies for resolving the two terms
1) account_type:  this was set to `User::CLEVER_ACCOUNT` or `User::GOOGLE_CLASSROOM_ACCOUNT` in a recent update of the google clever code.  Having an account_type of one of these two values represents the most recent update to the user and therefore points to whether google_id or clever_id should be removed
2) auth_credential: if there's a recent authorization grant, we can use this to infer the current account status
3) last_classroom:  the last updated classroom can have a value for `google_classroom_id`, `clever_id`, or neither.  (I've verified there are no classrooms with both `google_classroom_id` and `clever_id set`)

### Screenshots

### Notion Card Links

PR Checklist | Your Answer
------------ | -------------

### DIFF
--- a/services/QuillLMS/app/services/dual_google_id_and_clever_id_resolver.rb
+++ b/services/QuillLMS/app/services/dual_google_id_and_clever_id_resolver.rb
@@ -15,7 +15,7 @@ class DualGoogleIdAndCleverIdResolver < ApplicationService
   end
 
   def run
-    RESOLVERS.detect { |resolver| send(resolver) }
+    RESOLVERS.find { |resolver| send(resolver) }
   end
 
   private def log_account_type_change(changed_attribute, previous_value, action)
@@ -27,6 +27,7 @@ class DualGoogleIdAndCleverIdResolver < ApplicationService
       previous_value: previous_value
     )
   end
+
   private def resolve_by_account_type
     case user.account_type
     when User::GOOGLE_CLASSROOM_ACCOUNT then set_user_account_type_google
--- a/services/QuillLMS/app/services/dual_google_id_and_clever_id_resolver.rb
+++ b/services/QuillLMS/app/services/dual_google_id_and_clever_id_resolver.rb
@@ -1,0 +1,63 @@
+# frozen_string_literal: true
+
+class DualGoogleIdAndCleverIdResolver < ApplicationService
+  attr_reader :user
+
+  RESOLVERS = %i[
+    resolve_by_account_type
+    resolve_by_auth_credential
+    resolve_by_last_classroom
+  ].freeze
+
+
+  def initialize(user)
+    @user = user
+  end
+
+  def run
+    RESOLVERS.detect { |resolver| send(resolver) }
+  end
+
+  private def log_account_type_change(changed_attribute, previous_value, action)
+    ChangeLog.create(
+      action: action,
+      changed_attribute: changed_attribute,
+      changed_record: user,
+      explanation: caller_locations[0].to_s,
+      previous_value: previous_value
+    )
+  end
+  private def resolve_by_account_type
+    case user.account_type
+    when User::GOOGLE_CLASSROOM_ACCOUNT then set_user_account_type_google
+    when User::CLEVER_ACCOUNT then set_user_account_type_clever
+    end
+  end
+
+  private def resolve_by_auth_credential
+    return set_user_account_type_google if user.google_authorized?
+    return set_user_account_type_clever if user.clever_authorized?
+  end
+
+  private def resolve_by_last_classroom
+    last_classroom = user&.classrooms&.order(:updated_at)&.last
+
+    return false unless last_classroom.present?
+    return set_user_account_type_google if last_classroom.google_classroom_id.present?
+    return set_user_account_type_clever if last_classroom.clever_id.present?
+  end
+
+  private def set_user_account_type_clever
+    google_id = user.google_id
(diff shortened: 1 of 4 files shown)


services/QuillLMS/app/services/dual_google_id_and_clever_id_resolver.rb
```diff
@@ -0,0 +1,63 @@
...
+  private def log_account_type_change(changed_attribute, previous_value, action)
+    ChangeLog.create(
+      action: action,
+      changed_attribute: changed_attribute,
+      changed_record: user,
+      explanation: caller_locations[0].to_s,
+      previous_value: previous_value
+    )
+  end
+  private def resolve_by_account_type
```
anathomical, line 31: Not that it matters, but missing newline space separating this method from the previous one?

services/QuillLMS/lib/tasks/temporary/users.rake
```diff
@@ -20,4 +20,13 @@ namespace :users do
 
     puts "\nUpdated #{num_updates} user emails"
   end
-end
\ No newline at end of file
+
+  task resolve_dual_google_id_and_clever_id: :environment do
+    puts "Going to resolve #{users.count} google_id and clever_id"
```
anathomical: Where does `users` come from here?  Does this actually work?

services/QuillLMS/app/services/dual_google_id_and_clever_id_resolver.rb
```diff
@@ -0,0 +1,63 @@
...
+
+    return false unless last_classroom.present?
+    return set_user_account_type_google if last_classroom.google_classroom_id.present?
+    return set_user_account_type_clever if last_classroom.clever_id.present?
+  end
+
+  private def set_user_account_type_clever
+    google_id = user.google_id
+    user.update!(account_type: User::CLEVER_ACCOUNT, google_id: nil)
+    log_account_type_change(:google_id, google_id, ChangeLog::SET_USER_ACCOUNT_TYPE_CLEVER)
```
anathomical, line 54: Strongly approve of you adding `ChangeLog` events for these.

anathomical (approved):
Looks good.  I really love the test coverage.

services/QuillLMS/app/services/dual_google_id_and_clever_id_resolver.rb
```diff
@@ -0,0 +1,63 @@
...
+  def initialize(user)
+    @user = user
+  end
+
+  def run
+    RESOLVERS.detect { |resolver| send(resolver) }
+  end
+
+  private def log_account_type_change(changed_attribute, previous_value, action)
+    ChangeLog.create(
```
dandrabik, line 22: I really like that you are recording a Changelog.

services/QuillLMS/app/services/dual_google_id_and_clever_id_resolver.rb
```diff
@@ -0,0 +1,63 @@
...
+    resolve_by_last_classroom
+  ].freeze
+
+
+  def initialize(user)
+    @user = user
+  end
+
+  def run
+    RESOLVERS.detect { |resolver| send(resolver) }
```
dandrabik: Consider using `.find` since it's the more common synonym of `detect`.

services/QuillLMS/lib/tasks/temporary/users.rake
```diff
@@ -20,4 +20,13 @@ namespace :users do
...
+    ActiveRecord::Base.transaction do
```
dandrabik: Putting all of these in one transaction (80K rows, several writes each, plus code logic runtime), might lock rows long enough to cause some errors if these users hit the site (it might be fine, I'm not sure how fast this will be). We might consider putting a transaction around smaller sections (e.g. with each `find_in_batches` instead of `find_each`).

It might also be worth asking about whether we need a transaction in general.

dandrabik (approved):
Overall looks good! Thanks for putting this together. Small note about the transactions, but will defer to your judgement on that.

services/QuillLMS/lib/tasks/temporary/users.rake
anathomical: Actually, given that resolved issues won't be found with the `where` clauses in subsequent runs, this should be idempotent without the transaction block.

brendanshean: @dandrabik Good point about transactions and whether they are necessary.  @anathomical good call on this being safe for running again if it fails at some point.  I'll remove the transaction block.

services/QuillLMS/app/services/dual_google_id_and_clever_id_resolver.rb
brendanshean: That's really ugly.  Thanks for flagging this.

services/QuillLMS/lib/tasks/temporary/users.rake
brendanshean: Good point.  That will not work.

PR #9604, opened 2022-08-26
Lms/reset activity feed on classroom archive

anathomical:
## WHAT
Reset and rebuild the ActivityFeed whenever a teacher's classroom is changed between visible and non-visible
## WHY
When a classroom is archived we don't want its events to show up in the activity feed anymore, and if a classroom was then subsequently unarchived, we want its data to appear in the activity feed again
## HOW
- Add an `after_save` hook to `Classroom` to trigger the `ActivityFeedRefillWorker` for related teachers when its `visible` value changes.
- Update the `ActivityFeedRefillWorker` so that it calls `reset!` on a given ActivityFeed even in cases where it isn't going to have any data to replace that feed with.  If a teacher archives all their classes, we expect their feed to become empty.

### Notion Card Links
https://www.notion.so/quill/Last-Year-s-Activities-showing-up-in-activity-feed-for-eacher-2c85fa52b10b44d6b11bcbf502b269ab

PR Checklist | Your Answer
------------ | -------------

### DIFF
--- a/services/QuillLMS/app/models/classroom.rb
+++ b/services/QuillLMS/app/models/classroom.rb
@@ -41,6 +41,8 @@ class Classroom < ApplicationRecord
   after_commit :hide_appropriate_classroom_units
   after_commit :trigger_analytics_for_classroom_creation, on: :create
 
+  after_save :reset_teacher_activity_feed, if: :saved_change_to_visible?
+
   has_many :classroom_units, dependent: :destroy
   has_many :units, through: :classroom_units
   has_many :unit_activities, through: :units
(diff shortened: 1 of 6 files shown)


dandrabik (approved):
Looks good to me from a code perspective, thanks for fixing this edge case and adding tests.

The only question I have is around the Product decision of whether we want to clear the feed or not. Might be worth clarifying that with Peter Sharkey, since I'm not 100% sure where that discussion landed.

happythenewsad (approved):

anathomical:
@dandrabik Per discussion on the Notion ticket here (https://www.notion.so/quill/Last-Year-s-Activities-showing-up-in-activity-feed-for-eacher-2c85fa52b10b44d6b11bcbf502b269ab), I think we're going to proceed to make this change and monitor to see if we should keep it.

PR #9605, opened 2022-08-26
lms/only-prohibit-unarchived-classroom-unit-names

anathomical:
## WHAT
Do not prohibit names from units assigned to archived classrooms
## WHY
Some teachers intend to archive their classrooms from last year and then reuse the activity pack names from then this year
## HOW
Modify the code that generates prohibited names to check whether the classroom a unit is attached to is `visible` or not, and allow users to re-use unit names if their classrooms are `visible: false`

### Notion Card Links
https://www.notion.so/quill/Teacher-Is-Prompted-to-Rename-Activity-Packs-That-Have-Not-Been-Assigned-Yet-44eefd4648544f56bdb56f9a6ee9e43c

PR Checklist | Your Answer
------------ | -------------

### DIFF
--- a/services/QuillLMS/app/controllers/teachers/units_controller.rb
+++ b/services/QuillLMS/app/controllers/teachers/units_controller.rb
@@ -27,7 +27,7 @@ class Teachers::UnitsController < ApplicationController
   end
 
   def prohibited_unit_names
-    unit_names = current_user.units.pluck(:name).map(&:downcase)
+    unit_names = current_user.units.joins(:classrooms).where(classrooms: {visible: true}).pluck(:name).map(&:downcase)
     render json: { prohibitedUnitNames: unit_names }.to_json
   end
 
--- a/services/QuillLMS/spec/controllers/teachers/units_controller_spec.rb
+++ b/services/QuillLMS/spec/controllers/teachers/units_controller_spec.rb
@@ -18,6 +18,14 @@ describe Teachers::UnitsController, type: :controller do
    )
   end
 
+  let!(:classroom_unit2) do
+    create(:classroom_unit,
+     unit: unit2,
+     classroom: classroom,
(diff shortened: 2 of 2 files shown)


brendanshean:
Are we using this classroom unit name uniqueness elsewhere or is this just something to help teachers not confuse their classes?  If so, what about:

Teacher has 
classroom_unit(id: 1, name: 'Best Class', visible: true)

archives it:
classroom_unit(id: 1, name: 'Best Class', visible: false)

then imports new classroom with same name:
classroom_unit(id: 2, name: 'Best Class', visible: true)

then un-archives original class
classroom_unit(id: 1, name: 'Best Class', visible: true)

brendanshean (approved):
I left a general question about effects elsewhere, but looks good otherwise.  Thanks for adding this.

emilia-friedberg (approved):

anathomical:
> Are we using this classroom unit name uniqueness elsewhere or is this just something to help teachers not confuse their classes? If so, what about:
> 
> Teacher has classroom_unit(id: 1, name: 'Best Class', visible: true)
> 
> archives it: classroom_unit(id: 1, name: 'Best Class', visible: false)
> 
> then imports new classroom with same name: classroom_unit(id: 2, name: 'Best Class', visible: true)
> 
> then un-archives original class classroom_unit(id: 1, name: 'Best Class', visible: true)

@brendanshean As far as I know, this is explicitly to help teachers not confuse themselves.  I don't think there's anywhere else where it matters.  (Basically we don't want teachers with two active packs with the same name because it'd explicitly be confusing on the report page.)  You are correct that this opens an edge case where someone archives a class, creates packs, and unarchives a class.  I think we should let that be possible in lieu of doing something more complicated until/unless it becomes an issue.

PR #9608, opened 2022-08-29
fix bug with admins logging into teacher accounts

emilia-friedberg:
## WHAT
Fix bug with school admins getting a "premium expired" message when they try to log into teacher accounts.

## WHY
We want admins of premium schools to be able to access those teachers' accounts.

## HOW
This code was flawed in that it was looking at the school administrator's subscription status, which could be a falsy value because we don't actually require that school admins "belong" to the school they administrate, meaning they don't actually themselves have the premium subscription. Further, it is possible to have different schools you administer have different subscriptions, meaning some could have active premium status while some don't, so it made more sense to move this check to the teacher.

We will have to run this: 

```
user_ids = SchoolsAdmins.all.map(&:user_id)
user_ids.each { |id| $redis.del("SERIALIZED_ADMIN_USERS_FOR_#{id}") }
```

when it is deployed to make sure cached data is up to date.

### Screenshots

### Notion Card Links
https://www.notion.so/quill/Admin-does-not-have-access-to-Teacher-Access-feature-c8ffef9428684e448da0eeac55abe07b

PR Checklist | Your Answer
------------ | -------------
Have you deployed to Staging? | No, but tested locally with staging data

### DIFF
--- a/services/QuillLMS/app/queries/teachers_data.rb
+++ b/services/QuillLMS/app/queries/teachers_data.rb
@@ -85,13 +85,19 @@ module TeachersData
         email: hash_value[:email],
         school: hash_value[:school]
       )
+
       user.define_singleton_method(:number_of_students) do
         hash_value[:number_of_students]
       end
+
       user.define_singleton_method(:number_of_activities_completed) do
         hash_value[:number_of_activities_completed]
       end
+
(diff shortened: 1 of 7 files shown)


anathomical (approved):

brendanshean (approved):
Thanks for fixing this.  It looks like a tricky bug to sort out.

PR #9620, opened 2022-08-30
Add backend objects for Staggered Release

brendanshean:
## WHAT
Add `ActivityPackSequence` and `ActivityPackSequenceActivityPack` models to facilitate the staggered release feature.

## WHY
These objects help keep track of what ActivityPacks are using the staggered release method

## HOW
`r g model ActivityPackSequence classroom:references diagnostic_activity:references release_method` and 
`r g model ActivityPackSequenceActivityPack unit:references activity_pack_sequence:references order:integer`

### Screenshots

### Notion Card Links
https://www.notion.so/quill/Staggered-Release-RFC-73333d8777234daa927fece5690d237e

PR Checklist | Your Answer
------------ | -------------


services/QuillLMS/spec/factories/activity_pack_release_units.rb
```diff
@@ -0,0 +1,9 @@
+# frozen_string_literal: true
+
+FactoryBot.define do
```
emilia-friedberg: I don't think there's a model file for this one and it doesn't seem to show up anywhere else, is this supposed to be here?

services/QuillLMS/app/models/activity_pack_sequence.rb
```diff
@@ -0,0 +1,34 @@
...
+#  fk_rails_...  (classroom_id => classrooms.id)
+#
+class ActivityPackSequence < ApplicationRecord
+  RELEASE_METHODS = [
+    IMMEDIATE_RELEASE = 'immediate',
+    STAGGERED_RELEASE = 'staggered'
+  ]
+
+  belongs_to :classroom
+  belongs_to :source, class_name: 'Activity', foreign_key: 'activity_id'
```
emilia-friedberg: I'd defer to everyone else's judgment on this, but I thought we were saying that the id on the table itself would be `source_id`, rather than the id being `activity_id` and just setting up `source` as the active record name. Having two different ways of calling the same thing here just seems a little ripe for confusion to me, but I'm the least versed in Rails architecture.

emilia-friedberg (approved):
Small comments but LGTM overall! Excited to have this moving. I'm wondering if the change in table names changes the params we decided to pass up to the frontend, but I should have a PR up for that pretty soon so maybe it'll be easier to talk about that when we can just look at the code together.

services/QuillLMS/app/models/activity_pack_sequence.rb
anathomical: I don't have strong feelings about what we use for `source`, but it definitely feels weird to preemptively de-couple the column name from the relation.  The current spec for the next round of scheduling-related stuff looks like it won't have a `source` at all, for instance.  It appears that the next thing we're likely to do i this space is put date-scheduling for assignment and due date on individual Activities in a single Activity Pack (unit).  I suppose that, in that universe, the `source` is a `unit_id` or something?  Might be worth deciding now if we think that the code in this PR is going to be reused in any way for the next scheduling feature set.

services/QuillLMS/spec/factories/activity_pack_release_units.rb
anathomical: I suspect that this is leftover from an earlier draft of the code.  Note that it also reference `activity_pack_release` as an attribute, and I don't think that's a thing at the moment.  Good catch.

anathomical (approved):
Basically: what Emilia said.  This looks good, some minor concerns about naming things, but it probably won't be a big deal whatever we decide.  There may be room for a more complex discussion that tries to incorporate these changes with features we want to build in the future, but we could also just decide to deal with that when we get to it.

services/QuillLMS/app/models/activity_pack_sequence.rb
dandrabik: I agree that naming-wise, we should use the same in the DB and the relation. So either one of these (I'll defer to you how much we want to abstract now, we can always migrate later):
1. The specific: `activity` and `activity_id` 
2. The generic: `source` and `source_id`

I will add, looking through the code, I did find reading `activity_id` in the DB a little confusing, e.g. "How is this sequence of activities bound to one single activity?" which caused me to re-read the RFC.

services/QuillLMS/db/migrate/20220830205005_create_activity_pack_sequence_units.rb
```diff
@@ -0,0 +1,13 @@
+# frozen_string_literal: true
+
+class CreateActivityPackSequenceUnits < ActiveRecord::Migration[6.1]
+  def change
+    create_table :activity_pack_sequence_units do |t|
```
dandrabik: Some naming considerations here. If we rename `units` to `activity_packs` (which we should eventually), this will be `activity_pack_sequence_activity_packs`. It still may be the best name, just noting it.

services/QuillLMS/app/models/activity_pack_sequence.rb
```diff
@@ -0,0 +1,34 @@
+# frozen_string_literal: true
+
+# == Schema Information
+#
+# Table name: activity_pack_sequences
+#
+#  id             :bigint           not null, primary key
+#  release_method :string           default("immediate")
```
dandrabik: `release_method` is new since I viewed the RFC. Is the plan to create a record for all assigned activity packs (both staggered and immediate)? 

If one with `immediate` is created, are we creating `ActivityPackSequenceUnit`s for them as well?

services/QuillLMS/app/models/activity_pack_sequence.rb
```diff
@@ -0,0 +1,34 @@
...
+class ActivityPackSequence < ApplicationRecord
+  RELEASE_METHODS = [
+    IMMEDIATE_RELEASE = 'immediate',
+    STAGGERED_RELEASE = 'staggered'
+  ]
+
+  belongs_to :classroom
+  belongs_to :source, class_name: 'Activity', foreign_key: 'activity_id'
+
+  validates :release_method, inclusion: { in: RELEASE_METHODS }
```
dandrabik, line 32: Just a note, but if we end up having a lot of branching logic on this string, you could change this to `type` (or overwrite `inheritance_column`) and use `STI` to encapsulate the differing logic in separate classes.

dandrabik (approved):
Overall this looks good. I left a few comments, but nothing is blocking.

services/QuillLMS/app/models/activity_pack_sequence.rb
brendanshean: After talking with @anathomical I'm going to go with `diagnostic_activity_id` since that's the current use case.   If at some point we add some more uses of ActivityPackSequence we can abstract then.

services/QuillLMS/db/migrate/20220830205005_create_activity_pack_sequence_units.rb
brendanshean: I debated doing this but I found the name really ugly and long.  However, on the second thought it does get us moving in the right direction of more clarity in our codebase so I will rename it.

services/QuillLMS/app/models/activity_pack_sequence.rb
brendanshean: I was thinking about doing this on all, but I think to keep scope small we'll just create staggered ones for now.

PR #9621, opened 2022-08-30
Lms/enable new sales contacts

anathomical:
## WHAT
- Re-enable the routing for the creation of Sales Contact Forms
- Update user auth and password reset logic to account for new `User` records with `sales-contact` roles.
## WHY
We want to use this new workflow for collecting Sales Contact data for incoming sales opportunities, but also want to guarantee two things:
1. If a person submits a sales form request to us and later becomes a user, we want to make sure that the form and user are associated with each other when we sync data to Vitally
2. No one should experience any weirdness with authentication or sign up given that we're using the `User` table in an unconventional way
## HOW
- Stop redirecting sales form requests to the old form
- Confirm that Clever auth acts as it always does even if there is a sales-contact user
- Make sure that password login and password reset treat `sales-contact` `User`s as if they don't exist
- Update Google auth logic to handle `sales-contact` `User`s as if they don't exist unless someone's signing up

### Notion Card Links
https://www.notion.so/quill/Address-edge-cases-in-the-sales-contact-implementation-6bbc884385f348c3af58c4496f6bb3f5

PR Checklist | Your Answer
------------ | -------------

### DIFF
--- a/services/QuillLMS/app/controllers/auth/google_controller.rb
+++ b/services/QuillLMS/app/controllers/auth/google_controller.rb
@@ -85,17 +85,28 @@ class Auth::GoogleController < ApplicationController
     end
     @user = GoogleIntegration::User.new(@profile).update_or_initialize
 
+    update_role_if_necessary
+    show_user_not_found_if_necessary
+  end
+
+  private def in_sign_up_flow?
     # session[:role] is only set during the account creation workflow, checking for it
     # lets us differentiate between sign in and sign up
-    if session[:role]
-      @user.update(role: session[:role]) if @user.non_authenticating?
-    elsif @user.new_record? || @user.non_authenticating?
-      flash[:error] = user_not_found_error_message
-      flash.keep(:error)
-      redirect_to(new_session_path, status: :see_other)
-    end
-    # if neither special condition above is true, then setting @user is the last
-    # thing we need to do, just return nil
+    session[:role].present?
+  end
+
+  private def update_role_if_necessary
+    return unless in_sign_up_flow? && @user.non_authenticating?
+
+    @user.update(role: session[:role])
+  end
+
+  private def show_user_not_found_if_necessary
+    return unless (@user.new_record? || @user.non_authenticating?) && !in_sign_up_flow?
+
+    flash[:error] = user_not_found_error_message
+    flash.keep(:error)
+    redirect_to(new_session_path, status: :see_other)
   end
 
   private def user_not_found_error_message
--- a/services/QuillLMS/app/controllers/auth/google_controller.rb
+++ b/services/QuillLMS/app/controllers/auth/google_controller.rb
@@ -102,7 +102,11 @@ class Auth::GoogleController < ApplicationController
   end
 
   private def show_user_not_found_if_necessary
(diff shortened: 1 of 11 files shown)


eadams17 (approved):

services/QuillLMS/app/models/user.rb
```diff
@@ -662,6 +663,15 @@ def segment_user
     SegmentIntegration::User.new(self)
   end
 
+  # With the introduction of the SALES_CONTACT we now have a sort of
+  # "prospective user" type of user.  These people haven't signed up
+  # through our onboarding flow, but are given a User record so that we
+  # can sync their data to Vitally.  We need to treat these users specially
+  # during auth flows because they haven't actually signed up.
+  def non_authenticating?
```
dandrabik: I think using a positive name here, e.g.  `.prospective?` would make some of the code easier to read and avoid double/triple negatives, e.g. `!@user.non_authenticating` or `return unless @user.non_authenticating` 

And I understand what you are trying to do here with the naming and this comment, but you should also consider just naming this `sales_contact?`. We only have one type in the role, it might be better to keep the naming specific until we have a second case.

services/QuillLMS/app/controllers/sessions_controller.rb
```diff
@@ -46,7 +46,7 @@ def create
   def login_through_ajax
     email_or_username = params[:user][:email].downcase.strip unless params[:user][:email].nil?
     @user =  User.find_by_username_or_email(email_or_username)
-    if @user.nil?
+    if @user.nil? || @user&.non_authenticating?
```
dandrabik: You don't need a second `&.` nil check here.

services/QuillLMS/app/controllers/auth/google_controller.rb
```diff
@@ -85,7 +85,28 @@ def online_access_callback
...
+    show_user_not_found_if_necessary
+  end
+
+  private def in_sign_up_flow?
+    # session[:role] is only set during the account creation workflow, checking for it
+    # lets us differentiate between sign in and sign up
+    session[:role].present?
+  end
+
+  private def update_role_if_necessary
```
dandrabik: This is a very general name with a specific purpose. Renaming/ordering might make this easier to read, e.g.:
```ruby
def update_sales_contact_role
  return unless @user.sales_contact? && in_sign_up_flow?

dandrabik (changes requested):
Thanks for pushing this along. I think some re-naming here might make some of the code easier to read, so I'm requesting that, e.g. something like `sales_contact?` instead of `non_authenticating?` 

I'm also a little worried that we are adding another edge-case in code that has a lot of branching logic, but I understand this is kinda a means-to-an-end for now. We'll just have to monitor and be on top of potential bugs.

services/QuillLMS/app/models/user.rb
anathomical: I went back and forth on what to call this, and sort of went with the broader version "just in case", but I think you're right that we might as well keep the scope of this restricted until we actually know we have another example.  There's a good chance we'd want to refactor this if we had additional use cases anyway.

dandrabik (approved):
Thanks for making these changes!

PR #9626, opened 2022-08-31
Lms/refactor tooltip to use concept results

anathomical:
## WHAT
Update the Tooltip raw SQL query to use the new `concept_results` table instead of the `old_concept_results` table.  We didn't include this update in the last SQL update because it also involved some front-end tweaks.
## WHY
We want to pull all of our Concept Result data from the new table eventually, so this is a step along that path
## HOW
- Update the query in the raw SQL in the controller
- Tweak the front-end code so that instead of looking for the `correct` value inside a `metadata` key, just look for it on the `concept_result` directly
- Tweak the front-end code so that it expect `correct` to be a boolean value rather than an integer
- Update tests to more strictly ensure that we adhere to these formats

### Notion Card Links

PR Checklist | Your Answer
------------ | -------------

### DIFF
--- a/services/QuillLMS/app/controllers/grades_controller.rb
+++ b/services/QuillLMS/app/controllers/grades_controller.rb
@@ -23,16 +23,16 @@ class GradesController < ApplicationController
     RawSqlRunner.execute(
       <<-SQL
         SELECT
-          old_concept_results.metadata,
+          concept_results.correct,
           activities.description,
           concepts.name,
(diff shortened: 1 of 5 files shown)


emilia-friedberg (approved):

dandrabik (approved):
:+1: Thanks for pushing this forward!

happythenewsad (approved):

PR #9630, opened 2022-09-01
Remove unit id query string from Diagnostic Report url

cissyxyu:
## WHAT
Solve a bug where teachers would get a 500 error when they click through this path:
1. From Activity Analysis page, select any Diagnostic.
2. From that Diagnostic's summary page, select a different classroom that was assigned the same diagnostic but through a different Activity Pack.
3. Hit a 500 error instead of seeing the appropriate Diagnostic Report.

## WHY
A small portion of users are hitting this bug and seeing a 500 error page or forever loading screen, which is bad for the user experience.

## HOW
Currently, when teachers click any Diagnostic's name from their Activity Analysis page, we redirect them to the url for that Diagnostic's Diagnostic Report page, for example: https://www.quill.org/teachers/progress_reports/diagnostic_reports#/diagnostics/1663/classroom/1205406/summary?unit=2153487

Notice that at the end of the url, we attach a unit query string like "?unit=2153487". 

This unit query string tells the front end to pass that unit ID as a payload param for any further clicks or actions the teacher makes on that page. But this causes a bug - when the teacher clicks the classroom dropdown to select a different classroom, for example:
Screen Shot 2022-09-01 at 2 46 26 PM

they may run into a 500 error. It's because the front end passes the unit ID through as a param, even though other classrooms that have this Diagnostic assigned may not have the same unit ID attached to that diagnostic. The back end then looks for a ClassroomUnit with that particular Unit ID, which does not always exist.

We can fix this by removing the unit query string from the end of the URL when the teacher selects a new classroom the classroom dropdown. This is something we already do when the user accesses Diagnostic Reports through the "Diagnostics" button in the menu. We generate a URL without any unit ID attached which then tells the back end to disregard unit ID when searching for the ClassroomUnit, instead using only the classroom ID and activity ID.

### Screenshots
Screen Shot 2022-09-01 at 2 53 10 PM

### Notion Card Links
https://www.notion.so/quill/Sentry-Error-NilError-lesson_recommendations_for_classroom-d2b84917eafb4b33a5e68a059628b476 
https://www.notion.so/quill/Sentry-Error-NilError-recommendations_for_classroom-3835c6f1ed69478a88232548707fef74

PR Checklist | Your Answer
------------ | -------------

### DIFF
--- a/services/QuillLMS/app/controllers/teachers/progress_reports/diagnostic_reports_controller.rb
+++ b/services/QuillLMS/app/controllers/teachers/progress_reports/diagnostic_reports_controller.rb
@@ -183,8 +183,7 @@ class Teachers::ProgressReports::DiagnosticReportsController < Teachers::Progres
     elsif Activity.diagnostic_activity_ids.include?(activity_id.to_i)
       activity_is_a_post_test = Activity.find_by(follow_up_activity_id: activity_id).present?
       summary_or_growth_summary = activity_is_a_post_test ? 'growth_summary' : 'summary'
-      unit_query_string = "?unit=#{unit_id}"
-      render json: { url: "/teachers/progress_reports/diagnostic_reports#/diagnostics/#{activity_id}/classroom/#{classroom_id}/#{summary_or_growth_summary}#{unit_query_string}" }
+      render json: { url: "/teachers/progress_reports/diagnostic_reports#/diagnostics/#{activity_id}/classroom/#{classroom_id}/#{summary_or_growth_summary}" }
     else
       render json: { url: "/teachers/progress_reports/diagnostic_reports#/u/#{unit_id}/a/#{activity_id}/c/#{classroom_id}/students" }
     end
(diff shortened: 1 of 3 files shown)


emilia-friedberg (changes requested):
Right now, accessing the reports with and without the unit query string will give you two different sets of data: one with aggregated results for students in that class for that diagnostic, which can be across multiple units, and one with just one specific to the passed unit id. Having the unit id when they follow the link from the activity summary was a product decision, because it was confusing for teachers to think they were going to see results for one specific pack their students had completed and then not getting the data they expected. This was probably especially an issue last year, when we had already had a lot of teachers assign the same diagnostic as both a pre- and post-test before official post-tests were released, but I would still be wary of just removing the code without a formal product decision. I think you could fix this bug by updating the frontend so that the unit query is removed when you change classrooms.

anathomical:
I agree with @emilia-friedberg that we should tweak the way the front-end behaves, at least as a bug-fix.

cissyxyu:
> Right now, accessing the reports with and without the unit query string will give you two different sets of data: one with aggregated results for students in that class for that diagnostic, which can be across multiple units, and one with just one specific to the passed unit id. Having the unit id when they follow the link from the activity summary was a product decision, because it was confusing for teachers to think they were going to see results for one specific pack their students had completed and then not getting the data they expected. This was probably especially an issue last year, when we had already had a lot of teachers assign the same diagnostic as both a pre- and post-test before official post-tests were released, but I would still be wary of just removing the code without a formal product decision. I think you could fix this bug by updating the frontend so that the unit query is removed when you change classrooms.

Understood! I changed the fix in this PR to implement your idea instead -- now this PR will just change the front end code so that when a teacher selects a new classroom from the Summary page, it will remove the Unit ID from the URL. Let me know what you think!

emilia-friedberg (approved):

anathomical (approved):

happythenewsad (approved):

PR #9631, opened 2022-09-01
lower minimum for actually performing search on school selector

emilia-friedberg:
## WHAT
Lower the minimum for actually performing a search from the school selector, and add debounce.

## WHY
Previously, we weren't showing results if a school had fewer than five characters in its name, which was confusing in some instances. This lowers that threshold to two characters, but adds a debounce so we aren't making unnecessary long-running queries.

## HOW
I rewrote the school selector component as a functional component to use our debounce hook, and dropped the minimum threshold on the front and backend to accommodate this search.

### Screenshots

### Notion Card Links
https://www.notion.so/quill/School-search-requires-a-user-to-hit-space-to-return-results-8b84e1aa09e548c7b1e660f6e410f2f0

PR Checklist | Your Answer
------------ | -------------


services/QuillLMS/app/controllers/schools_controller.rb
```diff
@@ -6,7 +6,7 @@ class SchoolsController < ApplicationController
   before_action :require_user, only: [:select_school]
 
   include CheckboxCallback
-  MIN_PREFIX_LENGHT_WHEN_LAT_LON_NOT_PRESENT = 4
+  MIN_PREFIX_LENGTH_WHEN_LAT_LON_NOT_PRESENT = 2
```
eadams17: I think maybe it's worth dropping the `WHEN_LAT_LON_NOT_PRESENT` part of this variable name because it doesn't seem to be relevant anymore?

eadams17 (approved):

PR #9638, opened 2022-09-01
rake task for flagset csvs

happythenewsad:
## WHAT
Convenience rake task for uploading flagsets from CSVs.

## WHY
We need to do this on a recurring basis. 

### Screenshots

### Notion Card Links
https://www.notion.so/quill/Product-Board-30f97e4eb01246dbb8264253fb385073?p=afd471842d74402d9a616545f8917d9b&pm=s

PR Checklist | Your Answer
------------ | -------------
Have you added and/or updated tests? |  no - simple rake task
Back-to-school 2022: Have you checked the [webinar schedule](https://www.notion.so/quill/Back-to-school-webinar-banners-2022-a75a89cfad9f434899ef6be3eb184733) to avoid for downtime/risky deploys? | (Yes or N/A)

### DIFF
--- a/services/QuillLMS/lib/tasks/flagsets.rake
+++ b/services/QuillLMS/lib/tasks/flagsets.rake
@@ -1,0 +1,27 @@
+# frozen_string_literal: true
+
+require 'csv'
+
+namespace :flagsets do
+  namespace :users do
+    # Example usage: rake 'flagsets:users:update_flagsets_from_csv[filepath, flagset]'
+    desc 'Add flagsets to users from csv'
+    task :update_flagsets_from_csv, [:filepath, :flagset] => :environment do |t, args|
+      iostream = File.read(args[:filepath])
+      if (CSV.parse(iostream, headers: true).headers & ["email", "flagset"]).count != 2
+        puts "Invalid headers. Exiting."
+        exit 1
+      end
+
+      CSV.parse(iostream, headers: true) do |row|
+        user = User.find_by_email(row['email'])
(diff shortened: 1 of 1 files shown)


services/QuillLMS/lib/tasks/flagsets.rake
```diff
@@ -0,0 +1,27 @@
...
+
+require 'csv'
+
+namespace :flagsets do
+  namespace :users do
+    # Example usage: rake 'flagsets:users:update_flagsets_from_csv[filepath, flagset]'
+    desc 'Add flagsets to users from csv'
+    task :update_flagsets_from_csv, [:filepath, :flagset] => :environment do |t, args|
+      iostream = File.read(args[:filepath])
+      if (CSV.parse(iostream, headers: true).headers & ["email", "flagset"]).count != 2
```
anathomical, line 11: Seems a little odd to require a `flagset` header if we also override that value if we pass in `flagset` as an arg, but as long as we document the behavior it should be fine since this is a purely internal tool.

anathomical (approved):
Looks good to me.

services/QuillLMS/lib/tasks/flagsets.rake
happythenewsad: It's for flexibility - sometimes Curriculum will want a mix of flagsets uploaded. Sometimes they just give a list of emails. In the latter case, it's nice to have the parameter override, so you don't need to populate a CSV column of repetitive values.

PR #9659, opened 2022-09-07
Lms/stop referencing old concept results

anathomical:
## WHAT
Refactor our code, and especially our tests, to stop referencing `OldConceptResult` records and reference `ConceptResult` records instead.  Highlights include:
- Refactoring the way we save `ConceptResults` into a single worker that saves an `OldConceptResult` and a new `ConceptResult` at the same time so there's only one place where this code happens and it's easy to clean up later
- Refactor any models with an ActiveRecord relationship to `old_concept_results` to not have those relationships anymore
- Refactor demo creation logic to use the new `ConceptResult` models instead of the old ones
- Clean up a ton of tests, mostly around initial data set-up, so that they pass as expected with the new models
## WHY
We want to phase `OldConceptResults` out completely, and this update should hopefully result in no active code reading from the old table anymore.  If we get this into production and don't have any bugs or errors, we should feel confident that we can safely drop the old table.
## HOW
- Replace all the infrastructure for parallel writing of `OldConceptResult` and `ConceptResult` with a single worker
- Change the `let` calls in a bunch of tests to create `ActivitySession` records without corresponding `ConceptResult` records.  The base `ActivitySession` factory creates attached `ConceptResult` records by default, but that throws off a bunch of calculations and metrics, so use the version of the factory that doesn't create the `ConceptResult` in those test situations.

### Notion Card Links

PR Checklist | Your Answer
------------ | -------------


services/QuillLMS/app/services/demo/report_demo_creator.rb
```diff
@@ -378,15 +378,14 @@ def self.create_replayed_activity_session(student, classroom_unit)
...
     act_session = ActivitySession.create({activity_id: REPLAYED_ACTIVITY_ID, classroom_unit_id: classroom_unit.id, user_id: student.id, state: "finished", percentage: replayed_session&.percentage})
-    replayed_session&.old_concept_results&.each do |cr|
+    replayed_session&.concept_results&.each do |cr|
       values = {
         activity_session_id: act_session.id,
         concept_id: cr.concept_id,
-        metadata: cr.metadata,
-        question_type: cr.question_type
+        metadata: cr.legacy_format[:metadata],
+        question_type: cr.concept_result_question_type&.text
```
brendanshean, line 386: Is this expected that `question_type` uses `&.text` whereas the report_demo_ap_creator and create_admin_report do not.

services/QuillLMS/app/services/demo/report_demo_creator.rb
```diff
@@ -403,15 +402,14 @@ def self.create_activity_sessions(students, classroom)
...
 
-          temp.old_concept_results.each do |cr|
+          temp.concept_results.each do |cr|
             values = {
               activity_session_id: act_session.id,
               concept_id: cr.concept_id,
-              metadata: cr.metadata,
-              question_type: cr.question_type
+              metadata: cr.legacy_format[:metadata],
+              question_type: cr.concept_result_question_type&.text
```
brendanshean, line 410: Is this expected that question_type uses &.text whereas the report_demo_ap_creator and create_admin_report do not.

services/QuillLMS/app/workers/save_activity_session_concept_results_worker.rb
```diff
@@ -5,13 +5,14 @@ class SaveActivitySessionConceptResultsWorker
...
 
-  def perform(old_concept_result_ids)
-    OldConceptResult.where(id: old_concept_result_ids).each do |old_concept_result|
-      concept_result = ConceptResult.find_or_create_from_old_concept_result(old_concept_result)
-      unless concept_result&.id
-        SaveActivitySessionConceptResultsWorker.perform_async([old_concept_result.id])
-        ErrorNotifier.report(ConceptResultCopyFailedException.new("Failed to copy OldConceptResult #{old_concept_result.id}.  Retrying."))
-      end
+  def perform(json_payload)
+    # TODO: When we stop writing OldConceptResult records, update this to
```
brendanshean, line 9: Good idea with the TODO here.

services/QuillLMS/spec/models/activity_session_spec.rb
```diff
@@ -829,16 +829,20 @@
...
-      expect(OldConceptResult).to receive(:create).with({
-        activity_session_id: activity_session.id,
-        concept_id: concept.id,
-        metadata: metadata,
-        question_type: 'lessons-slide'
-      }).and_call_original
-      ActivitySession.save_concept_results([activity_session], concept_results)
+      Sidekiq::Testing.inline! do
+        expect do
+          expect(ConceptResult).to receive(:init_from_json).with({
```
brendanshean: Is there a reason the spy `expect(ConceptResults)` is inside expect change matcher (instead of before it)?

services/QuillLMS/spec/controllers/api/v1/activity_sessions_controller_spec.rb
```diff
@@ -42,37 +42,45 @@
...
       let(:writing_concept) { create(:concept, name: 'Creative Writing') }
+      let(:another_concept) { create(:concept, name: 'Creative Writing') }
 
       let(:concept_result1) do
-        create(:old_concept_result,
-          activity_session_id: activity_session.id,
-          concept: writing_concept,
-          metadata: { foo: 'bar', correct: true }
-        )
+        {
```
brendanshean, line 48: Is there a reason we're not using the factory to generate these concept results?

brendanshean (approved):
A few very minor comments/questions, but overall looks great.  One step closer to getting rid of the old model.

services/QuillLMS/app/services/demo/report_demo_creator.rb
anathomical: Good catch on this and the next comment.  Looks like there's no spec coverage on the additional files, so they didn't get caught by automatic testing.  I'll make sure to do manual testing around it on staging.

services/QuillLMS/spec/controllers/api/v1/activity_sessions_controller_spec.rb
anathomical: Perhaps these should have different names, but these aren't actually `ConceptResult`-shaped hashes.  These are the shape that the front-end uses when it reports concept result data.  So what we're actually testing here is that the payloads do what we want.  Would renaming these variables be enough to make that clear, do you think?

services/QuillLMS/spec/models/activity_session_spec.rb
anathomical: No particular reason.  I'll pull the expect out to make things clearer.

services/QuillLMS/spec/controllers/api/v1/activity_sessions_controller_spec.rb
brendanshean: I think renaming would help with readability.  Otherwise, I would look at this code in a few months and ask the same question why a factory is not used here.  A different name can also help identify a possibly new object that could be abstracted later.

anathomical: Done.

services/QuillLMS/app/controllers/api/v1/activity_sessions_controller.rb
```diff
@@ -85,7 +85,9 @@ def destroy
     concept_results_to_save = @concept_results.map { |c| concept_results_hash(c) }.reject(&:empty?)
     return if concept_results_to_save.empty?
```
dandrabik: You can remove the `return` above since `each` does that check. You could also consider removing the interim variable and just chaining it all (take or leave this)

```ruby
@concept_results
  .map {|cr| concept_results_hash(cr) }
  .reject(&:empty?)
  .each {|crh| SaveActivitySessionConceptResultsWorker.perform_async(crh) }

dandrabik (approved):
Looks good to me, thanks for making these changes and pushing this forward.

services/QuillLMS/app/controllers/api/v1/activity_sessions_controller.rb
anathomical: Making it a single chain seems reasonable.  I'll make that update.

PR #9661, opened 2022-09-08
Flagset bugfix

happythenewsad:
## WHAT
Due to a hardcoded flag list, activity search caches were not being refreshed properly, resulting in newly-created activities not showing up in search. 

This fix removes the hardcoded logic, preventing future instances of this bug if/when flagsets change. 

## WHY
Activity search cache should be up to date ( < 5 mins). 

### Screenshots

### Notion Card Links
https://www.notion.so/quill/Product-Board-30f97e4eb01246dbb8264253fb385073?p=68bd96b3151947b3964b5f69358cf1f7&pm=s

PR Checklist | Your Answer
------------ | -------------
Back-to-school 2022: Have you checked the [webinar schedule](https://www.notion.so/quill/Back-to-school-webinar-banners-2022-a75a89cfad9f434899ef6be3eb184733) to avoid for downtime/risky deploys? | N/A)

### DIFF
--- a/services/QuillLMS/spec/queries/activity_search_spec.rb
+++ b/services/QuillLMS/spec/queries/activity_search_spec.rb
@@ -8,6 +8,7 @@ describe ActivitySearch do
     let!(:standard) { create(:standard) }
     let!(:standard_level) { create(:standard_level) }
     let!(:activity) { create(:activity, activity_categories: [], flags: %w{beta production}, activity_classification_id: activity_classification.id, standard: standard, standard_level: standard_level) }
+    let!(:beta_flagset_activity) { create(:activity, activity_categories: [], name: 'special activity', flags: %w{evidence_beta1}, activity_classification_id: activity_classification.id, standard: standard, standard_level: standard_level) }
     let!(:activity_category) { create(:activity_category) }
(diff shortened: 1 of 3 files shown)


services/QuillLMS/spec/queries/activity_search_spec.rb
```diff
@@ -8,40 +8,52 @@
...
+            "content_partner_name" => content_partner.name,
+            "order_number" => activity_category_activity.order_number,
+            "classification_id" => activity_classification.id,
+            "classification_key" => activity_classification.key
+          }
+        )
+      end
     end
+
+    context 'production flagset input' do
```
happythenewsad, line 50: This spec does not directly deal with the bug. I added it while investigating the cause of the bug, and think we should keep it.

anathomical (approved):

services/QuillLMS/spec/queries/activity_search_spec.rb
anathomical: Agreed.  There are tons of gaps  in our spec coverage, so any new specs that might catch something in the future, I'm in favor of.

dandrabik (approved):
:+1: Thanks for fixing this up!

PR #9664, opened 2022-09-08
add constraint for no spaces in emails and add rake task to clean up existing users

emilia-friedberg:
## WHAT
Add constraint against whitespace in emails and add rake task to clean up existing users with bad emails. 

## WHY
This is a constraint that already should have been in place. Valid emails don't have whitespace characters in them.

## HOW
Add constraint and spec, then a rake task to clean up all existing users with this issue (4403 currently on prod). On staging, which was replicated a few weeks ago, we had 4397. After running this rake task, it was down to 270 (1547 after just removing spaces). By spot-checking, most of these users seem inactive - a lot of them don't appear to have ever logged into their account after creating it, probably because the space in it was a typo and they couldn't get back in. I think we'll have to address it on a case-by-case basis if we have support issues come up as a result of this down the line.

### Screenshots

### Notion Card Links
Discovered in the course of investigating this card: https://www.notion.so/quill/Capital-City-Charter-Schools-Clever-issues-3c203cf4832641a2b2c86ec72fe0e560

PR Checklist | Your Answer
------------ | -------------
Have you deployed to Staging? | NO - non-app change, but ran rake task on staging db

### DIFF
--- a/services/QuillLMS/app/models/user.rb
+++ b/services/QuillLMS/app/models/user.rb
@@ -157,6 +157,7 @@ class User < ApplicationRecord
   validates :email,
     presence: { if: :email_required? },
     uniqueness:  { message: :taken, if: :email_required_or_present? },
+    format: { without: /\s/, message: :no_spaces_allowed },
     length: { maximum: CHAR_FIELD_MAX_LENGTH }
 
   validate :username_cannot_be_an_email
--- a/services/QuillLMS/spec/models/user_spec.rb
+++ b/services/QuillLMS/spec/models/user_spec.rb
@@ -901,6 +901,11 @@ describe User, type: :model do
         expect(user).to be_valid
       end
 
+      it 'is invalid when there is a space in it' do
+        user = build(:user,  email: 'test@test.lan ')
+        expect(user).to_not be_valid
+      end
+
       context 'when role requires email' do
         it 'is invalid without email' do
           user.safe_role_assignment 'teacher'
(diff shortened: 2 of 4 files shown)


services/QuillLMS/app/models/user.rb
```diff
@@ -157,6 +157,7 @@ class User < ApplicationRecord
   validates :email,
     presence: { if: :email_required? },
     uniqueness:  { message: :taken, if: :email_required_or_present? },
+    format: { without: /\s/, message: :no_spaces_allowed },
```
brendanshean, line 160: I haven't seen this `without` configuration before.  Cool.

brendanshean (approved):
Thanks for fixing this very subtle bug.

services/QuillLMS/spec/models/user_spec.rb
```diff
@@ -901,6 +901,11 @@
         expect(user).to be_valid
       end
 
+      it 'is invalid when there is a space in it' do
+        user = build(:user,  email: 'test@test.lan ')
+        expect(user).to_not be_valid
```
anathomical, line 906: It may also be worth confirming that `user.errors[:email]` includes our specific error for this class just in case sometime in the future we do something crazy like make `.lan` email addresses invalid and then break the spacing invalidation.  (Very unlikely, but maybe worth covering anyway.)

anathomical (approved):
I've got a suggested tweak to the spec, but this looks good to me.  Per our discussion on Slack, I agree that the users who fall through the rake task (have an email with a space, but already have a different account using their non-space email) are items we can just clean up later if people report issues.

PR #9668, opened 2022-09-08
Ortto

happythenewsad:
## WHAT
Ortto newsletter logic:
- delete old sengrid logic
- create new newsletter worker
- add callbacks when `User.send_newsletter` property changes
- add webhook so we can toggle this value from Ortto 
- rake task to set `send_newsletter: true` for all teachers

## WHY
Product request 

### Screenshots

### Notion Card Links
https://www.notion.so/quill/Product-Board-30f97e4eb01246dbb8264253fb385073?p=0fd1b419312f4991a9d33beee5c3a5d0&pm=s

PR Checklist | Your Answer
------------ | -------------
Back-to-school 2022: Have you checked the [webinar schedule](https://www.notion.so/quill/Back-to-school-webinar-banners-2022-a75a89cfad9f434899ef6be3eb184733) to avoid for downtime/risky deploys? |  N/A)


services/QuillLMS/app/workers/ortto_integration/newsletter_worker.rb
```diff
@@ -0,0 +1,35 @@
+# frozen_string_literal: true
+
+module OrttoIntegration
+  class NewsletterWorker
```
brendanshean: Based on the name `NewsletterWorker`, I'm not really sure what this method is doing.  For readability, I would rename it to something that communicates what it is doing with the API (e.g. `UpdateNewsletterSubscriptionStatusWorker`)

services/QuillLMS/app/models/concerns/teacher.rb
```diff
@@ -26,6 +28,12 @@ def scope
     end
   end
 
+  def ortto_newsletter_callback
```
brendanshean: Depending on what the worker is renamed to, I would in turn rename this callback method (e.g. `update_ortto_newsletter_subscription_status`). Otherwise, it's too vague and my initial guess is that it's a worker that sends a newsletter.

services/QuillLMS/app/controllers/ortto_integration/webhooks_controller.rb
```diff
@@ -0,0 +1,38 @@
...
+    before_action :valid_params?, only: [:create]
+
+    class OrttoWebhookBadRequestException < StandardError; end
+
+    VALID_ACTIONS = [
+      SUBSCRIBE   = 'subscribe',
+      UNSUBSCRIBE = 'unsubscribe'
+    ]
+
+    def create
```
brendanshean, line 15: With webhooks, there's usually some security signature verification (e.g. [stripe](https://github.com/empirical-org/Empirical-Core/blob/develop/services/QuillLMS/app/controllers/stripe_integration/webhooks_controller.rb#L20) or [intercom](https://github.com/empirical-org/Empirical-Core/blob/develop/services/QuillLMS/app/controllers/intercom_integration/webhooks_controller.rb#L70).  There probably should be some kind of security check for incoming requests.

brendanshean (commented):
I think this is pretty close to being ready but I have a couple comments.

services/QuillLMS/spec/models/concerns/teacher_spec.rb
```diff
@@ -551,6 +551,22 @@
     end
   end
 
+  context 'callbacks' do
+    describe '#ortto_newsletter_callback' do
+      let!(:teacher) { create(:teacher, send_newsletter: false) }
+
+      it 'should call NewsletterWorker when User.send_newsletter is changed' do
```
brendanshean: I think a separate spec for testing changing of `send_newsletter` from true to false would be helpful.  Otherwise this feels like the worker is initiated more with a switch to true.

services/QuillLMS/app/controllers/ortto_integration/webhooks_controller.rb
happythenewsad: Forgot to mention - there's been some issues configuring the webhook within Ortto's UI. The webhook won't be used immediately upon launch, so Peter Sharkey is fine with configuring the webhook after the rest of the code is deployed.

brendanshean (approved):
Looks good.  Nice work on this!

emilia-friedberg (approved):
Just want to confirm the sign up page doesn't look crazy with the newspaper row removed (it was already a little wonky with the button placement so I have a different PR to address that). Otherwise all looks good to me.

happythenewsad:
@emilia-friedberg 

Screen Shot 2022-09-09 at 9 57 29 AM

PR #9673, opened 2022-09-08
Lms/provide old concept result worker

anathomical:
## WHAT
Add flexibility to the ConceptResult creation worker so that it can handle arrays of JSON objects as inputs, not just single items
## WHY
Some of the old jobs that we moved over to the new job were enqueued with arrays instead of single items
## HOW
Add some logic to check if the input value is an array and iterate over it if so, or just save a single record if not

PR Checklist | Your Answer
------------ | -------------


brendanshean (approved):
This looks good and I think should address the overnight errors.

PR #9678, opened 2022-09-09
A Passage-specific spelling error synthetic generator

dandrabik:
## WHAT
Add a generator to make spelling errors that are specific to the passage. If an entry has a repeated word from the passage, generate a spelling error by dropping a middle character from the word.
## WHY
We know that misspellings can caused the ML models to mislabel entries if the training / seed data lacks misspellings, especially if the word was crucial to the meaning of the sentence. Our current spelling error generator works off of a static list. This is an attempt to generate spelling errors specific to "important words" in the entries.
## HOW

There are two decisions here:
1. **How to flag "important" words**: I defined "important" as repeated in the passage 5 times and having a length of 5 or more (to avoid 'the', 'and', 'but', etc.).
2. **What type of spelling error to create**: I created omission spelling errors (dropped a letter). These are common and easy to generate

We can iterate on both of these, but this felt like a reasonable place to start.
(Note, an earlier version of this used word length for "important", but more testing showed that wasn't ideal)
### Screenshots
Smallish sample, and of college students, but spelling error proportions
Screen Shot 2022-09-09 at 12 10 33 PM

### Notion Card Links

PR Checklist | Your Answer
------------ | -------------
Back-to-school 2022: Have you checked the [webinar schedule](https://www.notion.so/quill/Back-to-school-webinar-banners-2022-a75a89cfad9f434899ef6be3eb184733) to avoid for downtime/risky deploys? | (N/A)

### DIFF
--- a/services/QuillLMS/engines/evidence/lib/evidence/evidence/html_tag_remover.rb
+++ b/services/QuillLMS/engines/evidence/lib/evidence/evidence/html_tag_remover.rb
@@ -14,7 +14,9 @@ module Evidence
       html
         .gsub(HTML_TAG_REGEX, " ") # remove html tags
         .gsub("&#x27;", "'") # replace html single quotes
+        .gsub(/(\‘|\’)/,"'") # replace html single quotes
         .gsub("&quot;","\"") # replace html double quotes
+        .gsub(/(\“|\”)/,"\"") # replace html double quotes
         .gsub(/\s+/," ") # replace multiple spaces with single space
         .strip
     end
--- a/services/QuillLMS/engines/evidence/lib/evidence/evidence/html_tag_remover.rb
+++ b/services/QuillLMS/engines/evidence/lib/evidence/evidence/html_tag_remover.rb
@@ -1,0 +1,22 @@
+# frozen_string_literal: true
+
+module Evidence
+  class HTMLTagRemover < ApplicationService
+    HTML_TAG_REGEX = /<("[^"]*"|'[^']*'|[^'">])*>/
+
+    attr_reader :html
+
+    def initialize(html)
+      @html = html
+    end
+
+    def run
+      html
+        .gsub(HTML_TAG_REGEX, " ") # remove html tags
+        .gsub("&#x27;", "'") # replace html single quotes
+        .gsub("&quot;","\"") # replace html double quotes
+        .gsub(/\s+/," ") # replace multiple spaces with single space
+        .strip
+    end
+  end
+end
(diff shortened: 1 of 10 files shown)


services/QuillLMS/engines/evidence/lib/evidence/evidence/synthetic/seed_data_generator.rb
```diff
@@ -46,14 +45,8 @@ def self.csvs_for_activity(activity_id:, nouns: [])
...
 
-
       def initialize(passage:, stem:, nouns: [])
-        @passage = passage
-          .gsub(HTML_TAG_REGEX, " ") # remove html tags
-          .gsub("&#x27;", "'") # replace html single quotes
-          .gsub("&quot;","\"") # replace html double quotes
-          .gsub(/\s+/," ") # replace multiple spaces with single space
-          .strip
+        @passage = Evidence::HTMLTagRemover.run(passage)
```
anathomical, line 49: Love moving this to a generalized utility.

anathomical (approved):
I agree with your thought that this is a good starting place.  As you say, we'll want to refine both how to identify the right words, and how to misspell them most usefully, but doing it at all should let us get toward better models now, and we can continue to improve later.

services/QuillLMS/engines/evidence/lib/evidence/evidence/html_tag_remover.rb
```diff
@@ -0,0 +1,24 @@
...
+
+    attr_reader :html
+
+    def initialize(html)
+      @html = html
+    end
+
+    def run
+      html
+        .gsub(HTML_TAG_REGEX, " ") # remove html tags
```
brendanshean: For some HTML structures*, regexes will actually not [work](https://stackoverflow.com/a/11230103) for stripping tags.  Fortunately Rails has a [strip_tags](https://api.rubyonrails.org/classes/ActionView/Helpers/SanitizeHelper.html#method-i-strip_tags) method.

*HTML is a context free (type 2) grammar whereas regexes are a regular (type 3) grammar.

happythenewsad (approved):

services/QuillLMS/engines/evidence/lib/evidence/evidence/html_tag_remover.rb
dandrabik: Oh, good call, going to use `strip_tags` instead.

dandrabik: Actually, scratch that, I'll need to improve on this in a future PR. `strip_tags` replaces with a `blank` and we have a lot of stretches with no spaces in our passages, e.g. `Living</strong></p><p><mark>Electricity`, which results in `LivingElectricity` being pulled as a word. I don't have a good way to reliably pull those apart.

brendanshean: [loofah](https://github.com/flavorjones/loofah) `to_text` is probably close to what you want:

```
doc = Loofah.fragment("<h1>Title</h1><div>Content<br>Next line</div>")
doc.text    # => "TitleContentNext line"            # probably not what you want
doc.to_text # => "\nTitle\n\nContent\nNext line\n"  # better
```

Then you can just gsub("\n")

PR #9683, opened 2022-09-12
Add 300px margin to the website footer

cissyxyu:
## WHAT
Add a 300px margin to the footer. This looks awkward on some pages because of the beige body background, so on those pages, I changed the body background color to white for a more uniform appearance.

## WHY
It makes our website look more professional to have a good margin between the footer and the website content.

## HOW
Add a 300px margin to the footer styling.
On the pages where this looks awkward, override the body background styling to make the body background white.

### Scr
Screen Shot 2022-09-08 at 5 18 51 PM
eenshots
Screen Shot 2022-09-08 at 5 18 41 PM

### Notion Card Links
https://www.notion.so/quill/Design-Add-padding-between-footer-and-primary-content-area-2917350f1ca148268a3efce2cb4bb850

PR Checklist | Your Answer
------------ | -------------
Back-to-school 2022: Have you checked the [webinar schedule](https://www.notion.so/quill/Back-to-school-webinar-banners-2022-a75a89cfad9f434899ef6be3eb184733) to avoid for downtime/risky deploys? | Yes

### DIFF
--- a/services/QuillLMS/app/views/pages/premium.html.erb
+++ b/services/QuillLMS/app/views/pages/premium.html.erb
@@ -1,9 +1,3 @@
-<style>
-  body {
-    background-color: white;
-  }
-</style>
-
 <% if (current_user && current_user.role == 'teacher') %>
   <%= render 'teachers/shared/scorebook_tabs' %>
 <% else %>
--- a/services/QuillLMS/app/views/pages/premium.html.erb
+++ b/services/QuillLMS/app/views/pages/premium.html.erb
@@ -1,3 +1,9 @@
+<style>
+  body {
+    background-color: white;
+  }
+</style>
+
 <% if (current_user && current_user.role == 'teacher') %>
   <%= render 'teachers/shared/scorebook_tabs' %>
 <% else %>
--- a/services/QuillLMS/app/assets/stylesheets/shared/footer.scss
+++ b/services/QuillLMS/app/assets/stylesheets/shared/footer.scss
@@ -6,7 +6,7 @@
   display: flex;
   justify-content: center;
   box-shadow: 0 100vh 0 100vh #fff;
-  margin: auto;
+  margin: 300px auto auto auto;
   .footer-content {
     display: flex;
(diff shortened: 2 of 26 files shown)


services/QuillLMS/app/views/pages/premium.html.erb
```diff
@@ -1,3 +1,9 @@
+<style>
+  body {
+    background-color: white;
+  }
+</style>
```
emilia-friedberg: This should be in a CSS file.

emilia-friedberg (changes requested):
I think if we want to change the background color universally, which it seems like is effectively what you're doing (since we use the footer near-universally), I would just add a new rule to the `application.scss` file to that effect. I'd definitely run that by Jack if you haven't already, but that seems like the cleanest way to solve the problem for me. Otherwise, if we wanted to keep all the pages with the background color they have now, I think you'd have to do something like this:

- find all the erb files where the footer partial is included (from a quick global search it looks to me like there are ten)
- wrap the rest of the content of the page (probably excluding a navbar if there is one) in a new html tag, maybe `<main>` if it works or a div with a unique className (I'd test both, `main` would be the most semantic html if we can use it)
- add a css rule to your new `main` element that adds 300px of padding to the bottom of that, rather than 300px margin to the top of the footer

I *think* that would let you add that spacing in a way where the spacing would just be the background color of the existing page, if we do want to keep that gray background in some places. It at least gives us more flexibility to have differently colored pages if we want to (if we add the padding to the footer itself, we're committing ourselves to white everything from here on out). But the code is a little messier, so it's a trade off.

happythenewsad (commented):
Please re-request a review once Emilia's comment is resolved; thanks Cissy!

emilia-friedberg (approved):
Approving once tests pass! Thanks for sticking with this through multiple go-rounds. I think the solution we came to is a good one.

happythenewsad (approved):

eadams17 (approved):
looks good! thanks for tackling this

PR #9686, opened 2022-09-12
Lms/bulk rename activities

anathomical:
## WHAT
Add a rake task that can be used to upload a Curriculum-provided CSV and rename a large number of Activities based on that data
## WHY
Curriculum wants to provide better names for a large swath of activities, and we shouldn't make them do it all by hand with a ton of clicks
## HOW
Write a rake task that takes a CSV input and parses it for `update!` calls to the specified Activities

### Notion Card Links
https://www.notion.so/quill/Update-Activity-Names-Descriptions-and-Theme-Tags-via-a-Script-360c2816197b444494245da12651e1e4

PR Checklist | Your Answer
------------ | -------------
Have you added and/or updated tests? |  No tests on rake tasks

### DIFF
--- a/services/QuillLMS/lib/tasks/activities.rake
+++ b/services/QuillLMS/lib/tasks/activities.rake
@@ -61,7 +61,6 @@ namespace :activities do
     CSV.parse(pipe_data, headers: true) do |row|
       activity = Activity.find(row['Activity ID'])
       activity.update!(name: row['New Name'])
- 
     rescue
       puts "Failed to update for activity with id '#{row['Activity ID']}'"
     else
--- a/services/QuillLMS/lib/tasks/activities.rake
+++ b/services/QuillLMS/lib/tasks/activities.rake
@@ -44,4 +44,25 @@ namespace :activities do
       puts "Failed to update for activity named '#{row['name']}'"
     end
   end
+
+  desc 'Take a pipe of CSV with activity IDs and new names, set the specified Activities to their new name'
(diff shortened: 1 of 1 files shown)


services/QuillLMS/lib/tasks/activities.rake
```diff
@@ -44,4 +44,27 @@ namespace :activities do
...
+      puts '  rake activities:bulk_update_names < path/to/local/file.csv'
+      puts ''
+      puts 'If you are piping data into Heroku, you need to include the --no-tty flag:'
+      puts '  heroku run rake activities:bulk_update_names -a empirical-grammar --no-tty < path/to/local/file.csv'
+      exit 1
+    end
+
+    CSV.parse(pipe_data, headers: true) do |row|
+      activity = Activity.find(row['Activity ID'])
+      activity.update!(name: row['New Name'])
```
happythenewsad: In the past, I've found that CSVs can contain unexpected values. `Activity.name` is not a no null column. Also, values like `\t` or `\n` will cause unpleasant behavior. So I recommend some validation here.

happythenewsad (approved):
Approving, but I suggest implementing my comment.

emilia-friedberg (approved):

services/QuillLMS/lib/tasks/activities.rake
anathomical: Done.

PR #9701, opened 2022-09-14
Lms/fix lesson concept results race condition

anathomical:
## WHAT
This PR does two things:
1. It addresses a potential race condition (which I haven't been able to validate is actually happening in production, but might be) in which `ActivitySessions` for completed `Lesson` activities might get deleted out from under `ConceptResults` during the process of marking all completion data.
2. Adds some error reporting that aims to identify what's coming from the front-end that's causing some `Lesson` activities get at least some of their `ConceptResults` stuck in the Sidekiq retry queue with errors related to not having an `ActivitySession` to connect the `ConceptResult` to.  (I've traced the code from front-end to back-end, and can't see anywhere where this discrepancy might be getting added, so hopefully starting to report this issue in NewRelic before it gets to Sidekiq will be revelatory.)
## WHY
1. We don't want to erroneously delete data that we're going to use.
2. We want to understand what's going on with `ConceptResult` recording so that we can make sure we don't lose student data.
## HOW
1. Refactor the way we clean up "unused" `ActivitySessions` to use known input data rather than relying on the shifting state of data throughout the processing of that data.  (Basically, the old code relied on everything being synchronous and individual, apparently unrelated class methods not changing behavior while the new code just reads the incoming payload to figure out what to do.)
2. After we enqueue the `ConceptResult`s that have been provided from the Lessons front-end (even if they have bad data we don't want to just throw them away until we understand what's going on), we check to see if any of the payloads we just submitted are mal-formed, and if so, we call `ErrorNotifier.report` so that get alerted in NewRelic and Sentry of the problem.

### Notion Card Links

PR Checklist | Your Answer
------------ | -------------

### DIFF
--- a/services/QuillLMS/app/controllers/api/v1/classroom_units_controller.rb
+++ b/services/QuillLMS/app/controllers/api/v1/classroom_units_controller.rb
@@ -43,13 +43,21 @@ class Api::V1::ClassroomUnitsController < Api::ApiController
 
     states.update_all(locked: true, pinned: false, completed: true)
 
+    concept_results = concept_result_params[:concept_results].map(&:to_h)
+
+    # The incoming ActivitySession payload that reaches this controller from the
+    # front-end includes ActivitySessions for students assigned to take the Lesson
+    # who didn't show up.  For students who "miss" a Lesson in this manner, we mark
+    # that by deleting their related ActivitySession.  So if a student submitted no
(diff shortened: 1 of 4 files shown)


emilia-friedberg (approved):

dandrabik (approved):
Thanks for cleaning this up!

PR #9714, opened 2022-09-15
automatically set timezones for users

emilia-friedberg:
## WHAT
Automatically set timezones for users on save and on school selection (from the select school page, not from the my account page, because they can already set timezone manually there).

## WHY
This is a prerequisite for activity pack scheduling, since we don't want teachers to be surprised by when their packs are available and unavailable for students.

## HOW
Find gems that will a) translate a zip code into a time zone (for schools), and b) translate an IP address into a time zone if the user hasn't set a school. Then use these gems in a setter function on the user model, to be called as a `before_save` callback if the `time_zone` isn't already set and in the `select_school` controller.

### Screenshots

### Notion Card Links
](https://www.notion.so/quill/Implement-automatic-timezone-setting-cd57b15b81d6430a83569fe1d5605fe7)

PR Checklist | Your Answer
------------ | -------------


services/QuillLMS/app/models/user.rb
```diff
@@ -183,6 +183,7 @@ class User < ApplicationRecord
   before_validation :generate_student_username_if_absent
   before_validation :prep_authentication_terms
   before_save :capitalize_name
+  before_save :set_time_zone, { if: ->(u) { u.time_zone.nil? } }
```
anathomical: This might also work, for our purposes, with the slightly more readable `before_save :set_time_zone, unless: :time_zone`.  The logic isn't an exact mirror, but might be good enough?

services/QuillLMS/spec/spec_helper.rb
```diff
@@ -17,6 +19,14 @@
...
 end
 
+Geocoder::Lookup::Test.set_default_stub(
+  [
+    {
+      timezone: 'America/New_York'
+    }
+  ]
+)
+
```
anathomical: Nice job configuring this.

anathomical (approved):
Looks good to me!  Very clean implementation.  One thought on the save hook condition, but you can take it or leave it.

services/QuillLMS/spec/models/user_spec.rb
```diff
@@ -56,6 +56,7 @@
 #  users_to_tsvector_idx5             (to_tsvector('english'::regconfig, split_part((ip_address)::text, '/'::text, 1))) USING gin
 #
 require 'rails_helper'
+require("webmock/minitest")
```
brendanshean: minor nit.  `require 'webmock/minitest'`

services/QuillLMS/spec/spec_helper.rb
brendanshean: Agreed.  Sometimes the hardest part is getting the spec configured properly. 

Some notes on conventions:

`spec_helper.rb` is normally reserved for configuring specs that are not rails dependent (more info [here](https://stackoverflow.com/a/24145760))

For rspec configuration other than general configuration that applies to every spec, a convention is to place gem configuration in a new file within spec/support (e.g. [carrierwave.rb](https://github.com/empirical-org/Empirical-Core/blob/develop/services/QuillLMS/spec/support/carrierwave.rb)

So for this just create a new file `spec/support/geocoder.rb`: 
``` 
Geocoder.configure(lookup: :test, ip_lookup: :test)
Geocoder::Lookup::Test.set_default_stub([{ timezone: 'America/New_York' }])
```

This will automatically get pulled via rails_helper.rb using this [line](https://github.com/empirical-org/Empirical-Core/blob/cbcc5133d404bca494601add99a7d546aeab2271/services/QuillLMS/spec/rails_helper.rb#L43)

services/QuillLMS/app/models/user.rb
```diff
@@ -655,6 +656,18 @@ def sales_contact?
...
   end
 
+  def set_time_zone
+    z = ::Ziptz.new
+    school_timezone = z.time_zone_name(school&.zipcode || school&.mail_zipcode)
+
+    if school_timezone.present?
+      self.time_zone = school_timezone
+    else
+      geocoder_results = Geocoder.search(ip_address)
```
brendanshean, line 666: Are we guaranteed to have an ip_address for every user?  If not, we might want to test what happens with bad input to Geocoder.

brendanshean (approved):
Some minor comments, but nothing requiring re-review.

Nice work on getting this setup!

services/QuillLMS/app/models/user.rb
happythenewsad: I'm also curious about this. What about VPNs? 

It might be better to have the user self-configure, and just message this issue louder in various places.

happythenewsad (commented):
See comment.

services/QuillLMS/app/models/user.rb
emilia-friedberg: We don't necessarily have an `ip_address` for every user (though we will for everyone who goes through the signup flow, so it's only like imported students who won't immediately, and then they will get one on sign-in). `Geocoder.search(nil)` returns an empty array, though, so we'll set the time_zone for those users to nil. If it's a VPN, it's true that we'll set the time_zone to an inaccurate location (I think this was the idea behind trying it based on the school first). I think the decision to set it automatically was intentional, [based on competing product offerings](https://www.notion.so/quill/PS-Activity-pack-scheduling-01e66b545385448a8e086daeb62deafa#2b2aee57a24c432aa602b50542b79de6), and users are still able to change it in our system if they want to, but I did flag your concern on the product card @happythenewsad.

emilia-friedberg: @happythenewsad see response from PG [here](https://www.notion.so/Implement-automatic-timezone-setting-cd57b15b81d6430a83569fe1d5605fe7?d=eaa9aa84b46048fd994e306ac117d8e5)

happythenewsad (approved):

PR #9715, opened 2022-09-15
lms/sometimes-hide-vitally-districts

anathomical:
## WHAT
Only sync Vitally School district.id if there are active subscriptions
## WHY
We only want to organize Schools under Districts in Vitally if there's someone paying for something.  So if a district doesn't have at least one school with a subscription, don't attach it to any schools during the sync process.
## HOW
When determining what `organizationId` to attach to a school, first check to see if the district in question has any schools with active subscriptions.  If not, send `""` as the `organizationId` so that the school doesn't get assigned to a District since we only use District-level organization in Vitally for sales management.

### Notion Card Links
https://www.notion.so/quill/Overview-Only-Send-Vitally-Districts-for-Schools-in-Districts-that-Contain-Paying-Schools-10f355b54d494679b9e1510192bd34e5

PR Checklist | Your Answer
------------ | -------------


dandrabik (approved):
This looks good to me, thanks for making the change.

PR #9718, opened 2022-09-16
Rubocop - Ignore change for ambiguous block.

dandrabik:
## WHAT
Tiny PR for Rubocop config adjustment. For [ambiguous block linter](https://docs.rubocop.org/rubocop/cops_lint.html#lintambiguousblockassociation),  ignore rspec's `change` method 
## WHY
I'm stuck in an unfixable linter error loop in a branch with the `change` method (if I change the code, it either breaks or hits another Cop for empty params). The documentation of the Cop uses `change` as an example of a method you might ignore, so I'm assuming this is common.
## HOW
Add `IgnoreMethods`

This also allows us to remove 3 spec files being excluded from this rule (they also had this issue).
### Screenshots
Screen Shot 2022-09-16 at 10 51 55 AM

Screen Shot 2022-09-16 at 10 57 17 AM

### Notion Card Links

PR Checklist | Your Answer
------------ | -------------
Have you deployed to Staging? | NO - non-app change, NO - tiny change
Back-to-school 2022: Have you checked the [webinar schedule](https://www.notion.so/quill/Back-to-school-webinar-banners-2022-a75a89cfad9f434899ef6be3eb184733) to avoid for downtime/risky deploys? |  N/A

### DIFF
--- a/services/.rubocop.yml
+++ b/services/.rubocop.yml
@@ -132,10 +132,6 @@ Lint/AmbiguousBlockAssociation:
   Enabled: true
   IgnoredMethods:
     - change
-  Exclude:
-    - "QuillLMS/spec/models/activity_session_spec.rb"
-    - "QuillLMS/spec/models/user_spec.rb"
-    - "QuillLMS/spec/controllers/pages_controller_spec.rb"
 
 Lint/AmbiguousOperatorPrecedence:
   Enabled: true
--- a/services/.rubocop.yml
+++ b/services/.rubocop.yml
@@ -130,6 +130,8 @@ Lint/AmbiguousAssignment:
 
 Lint/AmbiguousBlockAssociation:
   Enabled: true
+  IgnoredMethods:
+    - change
   Exclude:
     - "QuillLMS/spec/models/activity_session_spec.rb"
     - "QuillLMS/spec/models/user_spec.rb"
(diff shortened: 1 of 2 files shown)


anathomical (approved):

brendanshean:
@dandrabik One thing I tend to do with the block form is specify a `.from(...)` and/or `.to(...)` and that seems to fix it.  I like this approach since it gives even more transparency about assumptions for the spec.

For example:
https://github.com/empirical-org/Empirical-Core/blob/develop/services/QuillLMS/spec/services/analytics/classroom_unit_updater_spec.rb#L54

dandrabik:
> @dandrabik One thing I tend to do with the block form is specify a `.from(...)` and/or `.to(...)` and that seems to fix it. I like this approach since it gives even more transparency about assumptions for the spec.
> 
> For example: https://github.com/empirical-org/Empirical-Core/blob/develop/services/QuillLMS/spec/services/analytics/classroom_unit_updater_spec.rb#L54

That's helpful to know, thanks for sharing. 

I think, though, if a linter rule requires a somewhat unrelated change to 'fix' it (and one that can't be understood from its message, `Parenthesize the param change to make sure that the block will be associated with the change method call`), we should remove or fix the rule (that very specific message is quite unfortunate in this case, because it's very clear, but impossible to implement, which is pretty frustrating)

brendanshean:
I think it's possible to fix it via their suggestion, but then that triggers a different linter error which nudges you to the different change format of `change(object, :attribute)` which I only like to use with a simple methods:

https://user-images.githubusercontent.com/2057805/190684957-e555491c-0821-4596-97cd-789bb122dfa7.mov

That being said, I agree with this change.

PR #9723, opened 2022-09-16
Lms/maintain access to assigned archived activities

anathomical:
## WHAT
Ensure that Students and Teachers maintain access to assigned activities even in rare cases where Curriculum later marks that activity as "archived".
## WHY
It's confusing to both students and teachers to have activities that they've seen, and know that they are working on, just disappear from their dashboards with no explanation.  We'd prefer not to confuse our users, and it's been decided that letting a small handful of users access an activity that's been archived for whatever reason is better.  If we, for some reason, absolutely need to prevent access to an activity or question, there are other methods (removing questions from the Activity or just deleting the record altogether).
## HOW
Remove `WHERE` conditions from raw SQL queries that were excluding results where `activities.flags` included "archived" in three different places (one student-facing API endpoint, and two teacher-facing API endpoints).

### Notion Card Links
https://www.notion.so/quill/Archiving-Activities-that-are-already-assigned-Only-Hide-in-Assignment-Flow-for-New-Teachers-Not-06034161a4f2420983680ba3c9cb8d48

PR Checklist | Your Answer
------------ | -------------

### DIFF
--- a/services/QuillLMS/app/controllers/teachers/units_controller.rb
+++ b/services/QuillLMS/app/controllers/teachers/units_controller.rb
@@ -215,13 +215,9 @@ class Teachers::UnitsController < ApplicationController
 
     return [] if teach_own_or_coteach_classrooms_array.empty?
 
-    scores, completed, archived_activities = ''
+    scores, completed = ''
 
-    if report
-      completed = lessons ? "HAVING ca.completed" : "HAVING SUM(CASE WHEN act_sesh.visible = true AND act_sesh.state = 'finished' THEN 1 ELSE 0 END) > 0"
-    else
-      archived_activities = "AND 'archived' != ANY(activities.flags)"
-    end
+    completed = lessons ? "HAVING ca.completed" : "HAVING SUM(CASE WHEN act_sesh.visible = true AND act_sesh.state = 'finished' THEN 1 ELSE 0 END) > 0" if report
 
     if lessons
       lessons = "AND activities.activity_classification_id = 6"
@@ -291,7 +287,6 @@ class Teachers::UnitsController < ApplicationController
           AND units.visible = true
           AND cu.visible = true
           AND ua.visible = true
-          #{archived_activities}
           #{lessons}
         GROUP BY
           units.name,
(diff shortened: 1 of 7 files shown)


services/QuillLMS/app/controllers/teachers/units_controller.rb
```diff
@@ -215,13 +215,9 @@ def score_info
...
 
-    scores, completed, archived_activities = ''
+    scores, completed = ''
 
-    if report
-      completed = lessons ? "HAVING ca.completed" : "HAVING SUM(CASE WHEN act_sesh.visible = true AND act_sesh.state = 'finished' THEN 1 ELSE 0 END) > 0"
-    else
-      archived_activities = "AND 'archived' != ANY(activities.flags)"
-    end
+    completed = lessons ? "HAVING ca.completed" : "HAVING SUM(CASE WHEN act_sesh.visible = true AND act_sesh.state = 'finished' THEN 1 ELSE 0 END) > 0" if report
```
brendanshean: While I normally prefer guard clauses,  I think a conditional block might be more readable given the ternary as well as expression length:

```
if report
  completed = lessons ? "HAVING ca.completed" : "HAVING SUM(CASE WHEN act_sesh.visible = true AND act_sesh.state = 'finished' THEN 1 ELSE 0 END) > 0"
end
```

services/QuillLMS/spec/queries/scorebook/query_spec.rb
```diff
@@ -35,6 +35,13 @@
     expect(results.map{|res| res["id"]}).to include(activity_session2.id)
   end
 
+  it 'returns activities even if the underlying activity has been archived' do
+    activity2.update(flags: [Activity::ARCHIVED])
+
+    results = Scorebook::Query.run(classroom.id)
+    expect(results.map{|res| res["id"]}).to include(activity_session2.id)
```
brendanshean: I think you can use pluck here:
`expect(results.pluck('id')).to include(activity_session2.id)`

brendanshean (approved):
This seems like a change in the positive direction.

happythenewsad (approved):

PR #9731, opened 2022-09-19
LMS/ fix schools with missing 0 in zipcode

eadams17:
## WHAT
add a check for 5 digit zipcode length for new schools and rake task to append missing lead 0 for School's with 4 digit zipcodes in our database

## WHY
it was causing these schools to be unsearchable in the school search on teacher accounts

## HOW
add rake task to append 0 for these Schools and add a validation that checks for a zipcode of length 5 when zipcode is present

### Screenshots

### Notion Card Links
https://www.notion.so/quill/Zip-codes-on-School-Selection-screen-for-teachers-are-missing-initial-0-when-applicable-31c6bf39e73746498d747ea2a2b05bf8

PR Checklist | Your Answer
------------ | -------------
Back-to-school 2022: Have you checked the [webinar schedule](https://www.notion.so/quill/Back-to-school-webinar-banners-2022-a75a89cfad9f434899ef6be3eb184733) to avoid for downtime/risky deploys? | yes

### DIFF
--- a/services/QuillLMS/app/controllers/cms/schools_controller.rb
+++ b/services/QuillLMS/app/controllers/cms/schools_controller.rb
@@ -94,8 +94,7 @@ class Cms::SchoolsController < Cms::CmsController
     if new_school.save
       redirect_to cms_school_path(new_school.id)
     else
-      flash[:error] = new_school.errors
-      redirect_to new_cms_school_path
+      redirect_to new_cms_school_path, flash: { error: new_school.errors }
     end
   end
 
--- a/services/QuillLMS/app/controllers/cms/schools_controller.rb
+++ b/services/QuillLMS/app/controllers/cms/schools_controller.rb
@@ -94,7 +94,8 @@ class Cms::SchoolsController < Cms::CmsController
     if new_school.save
       redirect_to cms_school_path(new_school.id)
     else
-      render :new
+      flash[:error] = new_school.errors
+      redirect_to new_cms_school_path
     end
   end
 
--- a/services/QuillLMS/app/models/school.rb
+++ b/services/QuillLMS/app/models/school.rb
@@ -66,7 +66,7 @@ class School < ApplicationRecord
   after_save :attach_new_district_school_admins, if: :saved_change_to_district_id?
 
   validate :lower_grade_within_bounds, :upper_grade_within_bounds,
-           :lower_grade_greater_than_upper_grade
+           :lower_grade_greater_than_upper_grade, :zipcode_length
 
   ALTERNATIVE_SCHOOL_NAMES = [
     HOME_SCHOOL_SCHOOL_NAME = 'home school',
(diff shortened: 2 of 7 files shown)


services/QuillLMS/app/models/school.rb
```diff
@@ -209,4 +209,10 @@ def subscription_status
       .where(user: old_district_admins)
       .destroy_all
   end
+
+  private def zipcode_length
```
anathomical, line 214: Two thoughts:
1) we may want to allow for 9-digit zip codes, or for non-US style zipcodes.  Maybe just make this a minimum rather than a strict length requirement?
2) We can use a Rails helper rather than writing a validation function.  This can be done with: `validates :zipcode, length: {min: 5}, allow_blank: true`

anathomical (approved):
General approach seems fine.  I'd like to see a bit of clean-up on how we do the validation (let's use a helper, not write a whole function).

Also, before we deploy this, we should do at least a little thinking (probably worth talking to Coaches) about non-US zipcodes and whether/how much we care about them.

services/QuillLMS/app/controllers/cms/schools_controller.rb
```diff
@@ -94,7 +94,8 @@ def create
     if new_school.save
       redirect_to cms_school_path(new_school.id)
     else
-      render :new
+      flash[:error] = new_school.errors
+      redirect_to new_cms_school_path
```
brendanshean: I think you can include flash keys in the redirect itself:

`redirect_to new_cms_school_path, flash: { error: new_school.errors }`

brendanshean (approved):
I think this looks good to go after Thomas' comments are addressed.

services/QuillLMS/app/models/school.rb
eadams17: This makes sense to me, thanks for the suggestion! As per your comment above, I checked in with Shannon about non-US zipcodes and it doesn't seem like they're much of a concern-- they're often left blank, so I think this `length: {min: 5}, allow_blank: true` validation makes the most sense

services/QuillLMS/app/controllers/cms/schools_controller.rb
eadams17: oh cool, I didn't know that-- thanks for the suggestion!

anathomical: Neither did I!  TIL.

brendanshean: Somewhat [confusingly](https://github.com/rails/rails/blob/83217025a171593547d1268651b446d3533e2019/actionpack/lib/action_controller/metal/flash.rb#L11), if you are using `flash[:alert]` or `flash[:notice]`, you don't even need to use the flash parent key:

```
redirect_to path, notice: 'some notice'
redirect_to path, alert: 'some alert'
```

anathomical: Rails maaaagic!

PR #9736, opened 2022-09-20
Ortto webhook

happythenewsad:
## WHAT
Addresses integrations issues with the Ortto webhook. 
- sending basic auth via an HTTP header was causing Heroku to 401 at the router level, for unknown reasons. Ortto support suggested passing the basic other in the query string, which worked. 
- we don't need 'subscribe' functionality, only 'unsubscribe', so I removed logic pertaining to that
- relaxed param checking, since Ortto issues a 'test connection' request, whose payload is different than the standard webhook request.

## WHY
So Peter Sharkey can use this webhook for Ortto funnel sequences. 

### Screenshots

### Notion Card Links
https://www.notion.so/quill/Product-Board-30f97e4eb01246dbb8264253fb385073?p=0d07dee76d7842a695882af4e719bddc&pm=s

PR Checklist | Your Answer
------------ | -------------
Back-to-school 2022: Have you checked the [webinar schedule](https://www.notion.so/quill/Back-to-school-webinar-banners-2022-a75a89cfad9f434899ef6be3eb184733) to avoid for downtime/risky deploys? | N/A)

### DIFF
--- a/services/QuillLMS/app/controllers/ortto_integration/webhooks_controller.rb
+++ b/services/QuillLMS/app/controllers/ortto_integration/webhooks_controller.rb
@@ -5,31 +5,31 @@ module OrttoIntegration
     protect_from_forgery except: :create
     http_basic_authenticate_with name: 'ortto_webhook_user', password: ENV['ORTTO_WEBHOOK_PASSWORD']
 
-    before_action :valid_params?, only: [:create]
-
     class OrttoWebhookBadRequestException < StandardError; end
 
     def create
       puts "\n\n --- PARAMS: #{params.inspect}"
-      unless valid_params?
-        ErrorNotifier.report(
-          OrttoWebhookBadRequestException.new("Bad request with params: #{params}")
-        )
-        return head(:bad_request)
-      end
+      puts "\nactivity: #{params.dig(:activity).class} #{params.dig(:activity)}"
+      puts "\ncontact: #{params.dig(:activity, :contact).class} #{params.dig(:activity, :contact)}"
 
       email = params.dig(:activity, :contact, :email)
       user = User.find_by_email(email)
-      return head :accepted unless user
+
+      if email.nil? || user.nil?
+        ErrorNotifier.report(
+          OrttoWebhookBadRequestException.new("Bad request with params: #{params}")
+        )
+        return head :accepted
+      end
 
       user.update!(send_newsletter: false)
       head :ok
     end
 
-    private def valid_params?
-      return false unless params.dig(:activity, :contact, :email)
-      true
-    end
+    # private def valid_params?
+    #   return false unless params.dig(:activity, :contact, :email)
+    #   true
+    # end
 
   end
 end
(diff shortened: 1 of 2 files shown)


services/QuillLMS/app/controllers/ortto_integration/webhooks_controller.rb
```diff
@@ -3,35 +3,27 @@
...
-      unless valid_params?
+      email = params[:email]
+      user = User.find_by_email(email)
+
+      if email.nil? || user.nil?
         ErrorNotifier.report(
           OrttoWebhookBadRequestException.new("Bad request with params: #{params}")
         )
-        return head(:bad_request)
+        return head :accepted
```
happythenewsad: Using HTTP 202 here will allow the initial 'test connection', and helps developers disambiguate between the test request and a normal request to this API.

services/QuillLMS/app/controllers/ortto_integration/webhooks_controller.rb
```diff
@@ -3,35 +3,27 @@
...
     class OrttoWebhookBadRequestException < StandardError; end
 
-    VALID_ACTIONS = [
-      SUBSCRIBE   = 'subscribe',
-      UNSUBSCRIBE = 'unsubscribe'
-    ]
-
     def create
-      unless valid_params?
+      email = params[:email]
```
brendanshean, line 11: Since it doesn't seem necessary to query User if email is nil, what about adding a guard clause early on?

```
email = params[:email]
return bad_request if email.nil?

user = User.find_by(email: email)
return bad_request if user.nil?
...

private def bad_request
  ErrorNotifier.report(OrttoWebhookBadRequestException.new("Bad request with params: #{params}"))
  head :accepted
end
```

services/QuillLMS/spec/controllers/ortto_integration/webhooks_controller_spec.rb
```diff
@@ -5,30 +5,36 @@
...
+      end
+
+      context 'user not found' do
+        it 'should return 202 and trigger an error report' do
+          stub_const('ENV', { 'ORTTO_WEBHOOK_PASSWORD' => ortto_secret } )
+          expect(ErrorNotifier).to receive(:report)
+          post :create, params: { email: 'an email', secret: ortto_secret }
 
-        expect(response.status).to eq 400
+          expect(response.status).to eq 202
```
brendanshean, line 36: I don't really have a preference on 202 or :accepted, but I think it should be consistent with the app file.

services/QuillLMS/spec/controllers/ortto_integration/webhooks_controller_spec.rb
```diff
@@ -5,30 +5,36 @@
...
+        expect(User.find(subscribed_user.id).send_newsletter).to eq false
+        expect(response.status).to eq 200
       end
     end
 
     context 'invalid payload' do
-      it "handles the JSON format error and reports to new relic" do
-        post :create, params: {invalid_param: 'xyz' }
+      context 'bad authentication' do
+        it 'should return 401' do
```
brendanshean: 403?

brendanshean (approved):
A few minor refactoring comments, but looks good otherwise.  No need for a re-review.

anathomical (approved):
Seems reasonable to me.  I also approve of @brendanshean's suggestion about a guard clause to avoid going to the db if we don't have to.

PR #9737, opened 2022-09-20
lms/clear-bad-concept-results-from-queue

anathomical:
## WHAT
Add a temporary early return to clear the Sidekiq queue of bad data
## WHY
As far as I can tell, we're no longer generating any new jobs on the Sidekiq queue with this particular data corruption, but there are a bunch in the retry queue that we want to clear out.
## HOW
Just check the incoming payload and don't try to save anything if that payload is corrupted

### Notion Card Links

PR Checklist | Your Answer
------------ | -------------


brendanshean (commented):
I don't have an issue with clearing the results in the queue.  I wonder though if it would make more sense just running this as a rake task or in the console.  The idea of committing it permanently leaves open the possibility of obscuring future corrupted lesson slide data from coming in.

brendanshean (approved):

happythenewsad (approved):

brendanshean (approved):
Looks good.  We also now have a nice template for future clearing.

PR #9744, opened 2022-09-21
Fix bug involving admins impersonating teachers

brendanshean:
## WHAT
Fix a bug where admins attempting to sign in as a teacher are redirected to sessions/new

## WHY
This happens when the teacher being impersonated is a google user that has expired credentials.  When the admin attempts to sign_in as the teacher, the session is labeled invalid which calls `reset_session_and_redirect_to_sign_in`.

While investigating this bug, I noticed the nested factory `:admin` under the `:user` factory which assigns a role: 'admin'.
I believe this is a vestigial factory (there are no users on production with role 'admin') since the admin relationship is currently modeled via `SchoolsAdmins` and `DistrictsAdmins` tables.  As a result, uses of `{ create(:admin) }` were replaced with `{ create(:teacher) }` since that more accurately represents how our system currently operates.

## HOW
Add a clause to the `confirm_valid_session` that checks to see if an admin is attempting to impersonate a teacher.

### Screenshots

### Notion Card Links
https://www.notion.so/quill/Google-loop-on-admin-dashboard-8f1034020df44942ac0a379a4007a879

PR Checklist | Your Answer
------------ | -------------

### DIFF
--- a/services/QuillLMS/spec/support/system.rb
+++ b/services/QuillLMS/spec/support/system.rb
@@ -4,7 +4,7 @@ RSpec.configure do |config|
   Capybara.register_driver :local_selenium_chrome_headless do |app|
     options = Selenium::WebDriver::Chrome::Options.new(
       args: [
-        'headless',
+        # 'headless',
         'window-size=1920x1280'
       ]
     )
--- a/services/QuillLMS/spec/support/system.rb
+++ b/services/QuillLMS/spec/support/system.rb
@@ -4,7 +4,7 @@ RSpec.configure do |config|
   Capybara.register_driver :local_selenium_chrome_headless do |app|
     options = Selenium::WebDriver::Chrome::Options.new(
       args: [
-        # 'headless',
+        'headless',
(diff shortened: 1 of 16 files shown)


services/QuillLMS/spec/support/system.rb
```diff
@@ -4,7 +4,7 @@
   Capybara.register_driver :local_selenium_chrome_headless do |app|
     options = Selenium::WebDriver::Chrome::Options.new(
       args: [
-        'headless',
+        # 'headless',
```
anathomical: Should we just remove this line altogether if we're commenting it out?

anathomical (approved):
Looks good to me.  One question about removing a line, but no other feedback.

services/QuillLMS/spec/support/system.rb
brendanshean: Whoops.  That's a mistake.  I will add that back in.  Thank you for spotting

cissyxyu (approved):

PR #9745, opened 2022-09-21
Lms/remove old concept results

anathomical:
## WHAT
This PR removes the `OldConceptResult` model and drops the `old_concept_results` table.  As such, this won't be merged or deployed immediately after approval.  We'll definitely want to do some addition validation and planning, but it made sense to me to go ahead and vet the code.
## WHY
We want to fully migrate to the new smaller-footprint of `ConceptResults` complete, and thus don't need to preserve the old models or data.
## HOW
- Remove references to `OldConceptResults` in the few places they remained (mostly to provide parallel writes to ensure we didn't lose data if we needed to revert)
- Remove all the files involved in the migration from old to new concept results
- Clean up a single test that had references to the old model's factory

### Notion Card Links
https://www.notion.so/quill/Test-drop-of-OldConceptResult-records-in-Staging-in-prep-for-doing-it-in-Prod-5aefde9aa374498090f491c71d92e45e

PR Checklist | Your Answer
------------ | -------------


services/QuillLMS/db/migrate/20220920190724_drop_old_concept_results.rb
```diff
@@ -0,0 +1,12 @@
+# frozen_string_literal: true
+
+class DropOldConceptResults < ActiveRecord::Migration[6.1]
+  def up
+    remove_column :concept_results, :old_concept_result_id
+    drop_table :old_concept_results
+  end
+
+  def down
+    raise ActiveRecord::IrreversibleMigration
```
brendanshean: 1. I would recommend doing this migration in a subsequent commit.  That way you are likely to avoid any issues with stale code referencing a table that is no longer there.

2. In general I try to avoid irreversible migrations even with drop table.  Rails will complain about irreversible migration with methods like `drop_table` and `remove_{column|reference}` etc, but in most cases this is just a matter of adding type information along with reference/index commands:

```
  def change
    remove_reference :concept_results, :old_concept_result_id, :int 
    remove_reference :old_concept_results, :activity_classification, foreign_key: true
    remove_index :old_concept_results, :concept_id

    drop_table :old_concept_results do |t|
      t.integer :activity_session_id
      t.integer :concept_id, null: false
      t.string :question_type
      t.json :metadata
    end
  end
```

services/QuillLMS/db/migrate/20220920190724_drop_old_concept_results.rb
```diff
@@ -0,0 +1,12 @@
+# frozen_string_literal: true
+
+class DropOldConceptResults < ActiveRecord::Migration[6.1]
+  def up
+    remove_column :concept_results, :old_concept_result_id
```
brendanshean: This one should be remove_reference vs remove_column since we want to undo this [operation](https://github.com/empirical-org/Empirical-Core/blob/develop/services/QuillLMS/db/migrate/20220707155016_create_concept_result.rb#L10)

brendanshean (commented):
I think this looks good overall, just a couple of comments.  I'd like to hear Peter and Dan's thoughts on Irreversible migrations before approval.

services/QuillLMS/db/migrate/20220920190724_drop_old_concept_results.rb
happythenewsad: My thinking is: eventually we want to drop the column, so might as well run `remove_column` now, to remove both the column and the reference in one fell swoop.

happythenewsad (commented):
This looks pretty good, but I'm not clear on why you're asking for a review if you don't want to merge and migrate this code right away. 

Should we be treating this as a draft, but review it as a regular PR?

Regarding @brendanshean 's comment on irreversible migrations: I don't like them, but they already exist in this code base, so the marginal 'code damage' is basically zero.

anathomical:
> This looks pretty good, but I'm not clear on why you're asking for a review if you don't want to merge and migrate this code right away.
> 
> Should we be treating this as a draft, but review it as a regular PR?
> 
> Regarding @brendanshean 's comment on irreversible migrations: I don't like them, but they already exist in this code base, so the marginal 'code damage' is basically zero.

anathomical:
Accidentally closed.

anathomical:
> This looks pretty good, but I'm not clear on why you're asking for a review if you don't want to merge and migrate this code right away.
> 

My thought here is that this is a high-stakes change to make, and we want to merge and deploy it sometime in the next sprint (most likely), so I wanted to ensure plenty of time to discuss things.  Like irreversible migrations.

Speaking of, this is probably worth having a bigger discussion.  As @happythenewsad points out, we have some of these in the codebase, but I feel like we should settle on a team approach we like.  @brendanshean my thought is that, conceptually, this migration is NOT reversible.  Re-creating the table can't restore the data, so a rollback won't actually put us in the unmigrated state.  I believe that's been our rationale in the past when we introduce irrerversible migrations: we don't want to lead people to think that they can "undo" what is being done (contrast with adding a column or table), but I'm totally open to being convinced otherwise.

services/QuillLMS/db/migrate/20220920190724_drop_old_concept_results.rb
brendanshean: If `remove_column` runs both then ignore my comment.

brendanshean:
@anathomical While one _could_ restore the data via a backup file, that seems like overkill here.  I'm fine with keeping the irreversible migration.

services/QuillLMS/spec/queries/teachers_data_spec.rb
```diff
@@ -40,8 +40,8 @@
 
   let!(:concept1) { create(:concept) }
   let!(:concept2) { create(:concept) }
-  let!(:concept_result1) { create(:old_concept_result, concept: concept1, activity_session: activity_session1) }
-  let!(:concept_result2) { create(:old_concept_result, concept: concept2, activity_session: activity_session2) }
+  let!(:concept_result1) { create(:concept_result, concept: concept1, activity_session: activity_session1) }
```
dandrabik: I'm surprised this test works by changing these from `old` to `new` concept_results. I'm guessing these aren't used.

services/QuillLMS/spec/models/concept_result_spec.rb
```diff
@@ -249,45 +215,17 @@
           "questionScore": 0.8
         }
       end
-      let(:old_concept_result) { create(:sentence_combining, concept: concept, activity_session: activity_session, concept_uid: concept.uid, metadata: metadata, activity_classification_id: activity.activity_classification_id) }
+      let(:concept_result) { ConceptResult.create_from_json({concept_id: concept.id, activity_session_id: activity_session.id, metadata: metadata, activity_classification_id: activity.activity_classification_id, question_type: 'sentence-combining'}) }
 
       it 'should return data in the same shape as a ConceptResult' do
```
dandrabik: I think we want this test name to reference "should return data in the same shape as a legacy ConceptResult'"

services/QuillLMS/db/structure.sql
```diff
@@ -9,7 +9,6 @@ SET xmloption = content;
 SET client_min_messages = warning;
 SET row_security = off;
```
dandrabik: This file feels like it doesn't belong in this PR.

anathomical:
@brendanshean @happythenewsad Just a heads-up that I've pulled the migration file out of this PR and will open a separate PR with that.

dandrabik (approved):
So it looks like you pulled the migrations out of this branch, but the `db/structure.sql` changes are still here. I think you need to move those into the migration branch.

I left a few other comments of small changes, but overall looks good (once the changes above are made).

Also, I don't have strong feelings about reversible vs. irreversible migrations in this particular case, with a slight preference towards reversible ones. One thing I will say is, it's not difficult to write a reversible migration (Brendan outlined it), and they tend to be easier to work with in development/test environments (if you have to re-run a migration or it errors), so it might be worth using these PRs to learn/practice how to write the reversible migration.

services/QuillLMS/spec/queries/teachers_data_spec.rb
anathomical: You were right.  I'm removing these lines.

happythenewsad (approved):

PR #9748, opened 2022-09-21
Remove unused code coverage libraries from test suite.

dandrabik:
## WHAT
We have a few code coverage libraries that make running tests slow, I'm proposing we remove them since I don't think anyone is looking at these reports (we can always add them back). **You will definitely notice the speed difference.**

To be specific, this is when the tests finish and rspec is still running for some reason, then you hit `Ctrl + c` and it says `RSpec is shutting down and will print the summary report... Interrupt again to force quit.`, then you hit `Ctrl + c` again and it finally finishes . . . all of that is the Code Coverage gems trying to generate/update reports. Without it, rspec ends almost instantly. 

Note, the reports gets slower the more tests you run, since some of these update previous reports, so the reports grow and grow, and eventually writing to them is very slow.
## WHY
Fast tests are very nice and I don't think anyone is using these reports.

Note, we could fix the setup to prevent large files (the slow part) or only run them on full suite runs, but it's easier to just remove if they aren't actively used.
## HOW
Removing gems.
### Screenshots
Screen Shot 2022-09-21 at 6 13 19 PM

### Notion Card Links

PR Checklist | Your Answer
------------ | -------------
Self-Review: Have you done an initial self-review of the code below on Github? | Yup

### DIFF
--- a/services/QuillLMS/Gemfile
+++ b/services/QuillLMS/Gemfile
@@ -198,16 +198,12 @@ end
 group :test do
   gem 'capybara'
   gem 'capybara-screenshot'
-  gem 'codeclimate-test-reporter', require: nil
-  gem 'codecov'
   gem 'mini_racer', '= 0.4.0'
   gem 'fakeredis', '~> 0.7.0'
   gem 'rails-controller-testing'
   gem 'selenium-webdriver', '~> 3'
   gem 'shoulda-matchers', '~> 4'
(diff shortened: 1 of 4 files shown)


happythenewsad (approved):
👍

brendanshean (approved):
Good idea.

anathomical (approved):

PR #9749, opened 2022-09-22
Require an Activity's topics to be level one, Rake Task for attaching new topics

cissyxyu:
## WHAT
We're changing the way we store topics on Activities through the ActivityTopics table. 
Instead of saving an activity's level zero, level one, level two, and level three topics, we now want to only store an Activity's level 1 topics (multiple level 1 topics can now be stored per activity).
Also included a .csv with new topics to attach to Activities and a rake task to attach them.

## WHY
This will allow us to give one activity multiple level 1 topics. We also don't need to store an Activity's level 2/level 3 topics anymore because we can find those topics by backtracing the `parent_id `of the level 1 topic, now that we've required all level 1 topics to have associated level 2 and level 3 topics.

## HOW
Add a validation to the `ActivityTopic` relationship to only save if the topic is level 1.
Modify places where we send an activities' level 1, level 2, and level 3 topic to instead send up to 3 level 1 topics. This will be more informative for Curriculum and reporting going forward.

### Screenshots

### Notion Card Links
https://www.notion.so/quill/Activities-can-have-3-tags-733866911ca14340bb5b34eccd1c2dd0

PR Checklist | Your Answer
------------ | -------------


services/QuillLMS/app/models/activity_topic.rb
```diff
@@ -23,4 +23,10 @@ class ActivityTopic < ApplicationRecord
   belongs_to :topic
 
   validates :topic_id, presence: true
+
```
brendanshean, line 26: Rails has a [validates_with](https://guides.rubyonrails.org/active_record_validations.html#validates-with) helper that would be more conventional here.  This also establishes a pattern of adding validation error messages that can be useful later.

```
class ActivityTopicLevelOneValidator < ActiveModel::Validator
  def validate(record)
    record.errors.add :base, "Some descriptive error message here" unless record.topic.level_one?
  end
end

class ActivityTopic < ApplicationRecord
  validates_with ActivityTopicLevelOneValidator
end
```

services/QuillLMS/lib/tasks/attach_new_topics_to_activities.rake
```diff
@@ -0,0 +1,18 @@
...
+namespace :topics do
+  desc 'Attach topics from new_topics_to_activities csv'
+  task :attach => :environment do
+    topics_table = CSV.parse(File.read("lib/data/new_topics_to_activities.csv"), headers: true)
+    topics_table.each do |row|
+      activity = Activity.find_by(id: row["Activity ID"])
+      next if activity.blank?
+
+      topic_names = [row["Topic 1"], row["Topic 2"], row["Topic 3"]]
+      activity.topic_ids = topic_names.map { |topic_name| Topic.find_by(name: topic_name)&.id }
```
brendanshean, line 12: Adding `compact` at the end will get rid of potential nils:
`topic_names.map { |topic_name| Topic.find_by(name: topic_name)&.id }.compact`

brendanshean (commented):
Retracting this comment on refactoring the topic factory.  Please ignore.

PR #9753, opened 2022-09-22
upgrade npm

happythenewsad:
## WHAT
Upgrade npm to 8.19.2 everywhere 

## WHY
- We get new syntax with 8.3.x (see Notion)
- it's good to keep tools up to date

## HOW
- update package.json to include npm specifier
- update circleci docs 

### Screenshots

### Notion Card Links
https://www.notion.so/quill/Product-Board-30f97e4eb01246dbb8264253fb385073?p=24e5e128a5ca46a492d04304453bfc18&pm=s

PR Checklist | Your Answer
------------ | -------------

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -153,10 +153,9 @@ jobs:
     steps:
       - checkout
       - run:
-        name: Install npm 8
-        command: |
-          sudo apt-cache search npm
-          sudo apt install -y npm=8.19.2
+          name: Install npm 8
+          command: |
+            sudo npm install npm@8.19.2 -g
 
       - restore_cache:
           keys:
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -152,6 +152,11 @@ jobs:
       - image: circleci/node:erbium
     steps:
       - checkout
+      - run:
+        name: Install npm 8
+        command: |
+          sudo apt-cache search npm
+          sudo apt install -y npm=8.19.2
 
       - restore_cache:
           keys:
--- a/services/QuillLMS/client/package.json
+++ b/services/QuillLMS/client/package.json
@@ -6,8 +6,10 @@
     "url": "https://github.com/empirical-org/Empirical-Core.git"
   },
   "license": "AGPL-3.0",
+  "engineStrict" : true,
   "engines": {
-    "node": "12.x"
+    "node": "12.x",
+    "npm": "8.x"
   },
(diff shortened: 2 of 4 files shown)


services/QuillLMS/client/package.json
```diff
@@ -6,8 +6,10 @@
     "url": "https://github.com/empirical-org/Empirical-Core.git"
   },
   "license": "AGPL-3.0",
+  "engineStrict" : true,
```
happythenewsad, line 9: Setting this flag to `true` will cause npm to throw exceptions when `engines` directives are not met. This line isn't necessary, but I found it gave me peace of mind while testing this PR to know that my intended version of npm was being used. Feel free to chime in about keeping this flag or not.

.circleci/config.yml
```diff
@@ -152,6 +152,10 @@ jobs:
       - image: circleci/node:erbium
     steps:
       - checkout
+      - run:
+          name: Install npm 8
+          command: |
+            sudo npm install npm@8.19.2 -g
```
anathomical, line 158: It feels like there should be an available docker image that has npm8 pre-installed so we don't need a `run` block here, but I think that our version of Node may be too far out of date (we're pinned to 12.x, it looks like) for images to have updated npm in them, so I'm okay with this approach until/unless we upgrade our Node version.

services/QuillLMS/client/package.json
anathomical: This seems reasonable to me, assuming that it doesn't cause issues during the build process.

anathomical (approved):
Not super happy about using `run` to install this in CI instead of just having a Docker image with everything installed already, but given how far behind our Node version is this may be the cleanest solution.

.circleci/config.yml
happythenewsad: Yeah - I researched this. Even if we were to bump to node 16.x, which is out of scope of this PR, node 16.x only supports npm 7.x

More discussion here: https://github.com/nodejs/node/discussions/37689

anathomical: I hadn't seen that.  Even with that being true, CircleCI (as an example) provides some convenience images that included `npm` 8 along with other versions of `node` (https://circleci.com/developer/images/image/cimg/node), but I don't see any images with `node` 12.x that do that.  And I agree that bumping the `node` version is well beyond the scope of this PR, so this work-around seem reasonable.

PR #9754, opened 2022-09-22
Demo Account Reset

dandrabik:
## WHAT
Reset changes within the demo account with a teacher clicks `view_demo` and when they click `exit demo` (do this one async). This 'reset' leaves the demo teacher in place, and edits/deletes the data associated with them to prevent people in the demo account from being logged out (the previous demo account `reset` which I've renamed `recreate` would delete the teacher).

Also, there is some refactoring/renaming within here to make this implementation easier to code and follow.
## WHY
Teachers can play with our demo account, which semi-frequently leaves the account in a weird state that is less helpful to the next teacher that sees it, e.g. with added classes or with the demo classroom deleted (the one with all of the reports).
## HOW
There are multiple ways we can prevent changes, and we are just trying this light-weight approach (looking for changes to the demo account and removing them) to start (it's only called in two places, and doesn't require us to edit other endpoints or frontend code).  It ensures that teachers will see a clean account when they land on the demo account. It's likely we'll iterate on this approach in the future with a more robust solution, e.g. a `demo` role than can edit anything.

PR Checklist | Your Answer
------------ | -------------
Have you added and/or updated tests? |  'YES'. (one failing test I'm working on)


services/QuillLMS/app/controllers/teachers/classroom_manager_controller.rb
```diff
@@ -8,7 +8,7 @@ class Teachers::ClassroomManagerController < ApplicationController
 
   respond_to :json, :html
 
-  around_action :force_writer_db_role, only: [:assign, :dashboard, :lesson_planner]
+  around_action :force_writer_db_role, only: [:assign, :dashboard, :lesson_planner, :view_demo]
```
dandrabik, line 11: Note, this adds a `get` link that has writes in it (to clean-up the demo account before the user gets there). We could make that fire off an async worker (that's what the leave demo account link does), but that wouldn't guarantee the demo account is in a good place when the teacher lands there(in all likelihood, it would not be reset by the time they land). Also, I believe, we want this to be a url that is clickable, shareable, rather than `post`.

This feels like we should make an exception to our "no new GET writes" policy for this one link, but open to feedback.

anathomical: I'm on board with this being an intentional exception.  We shouldn't be dogmatic about rules like that, and I think you're right about sharability.  Also, I can easily imagine Coaches or other staff keeping bookmarks for the demo, where it will again make sense to be a GET.  I feel like the alternative would be some sort of Pusher-based solution where you land on a page that starts a background job and gives you a spinner until the job completes (and sends a Pusher notification) back to the front-end client.  It's not clear to me that it's worth the extra complexity just to keep all GETs from having writes in them.

services/QuillLMS/spec/controllers/students_controller_spec.rb
```diff
@@ -67,8 +67,7 @@
 
     context 'when angie thomas does not exist' do
```
happythenewsad, line 68: Optional nit: rename context block to something better

happythenewsad (approved):
Looks good. Codebase is looking a bit more trim

anathomical (approved):
This looks good to me.  I seem to recall in discussions about doing this in the past we were concerned that doing tons of data reset on the demo would create tons of "bad" data for our analytics around things like activity rates and stuff.  Is that no longer a concern, or did we just decide this was worth it?

dandrabik:
> This looks good to me. I seem to recall in discussions about doing this in the past we were concerned that doing tons of data reset on the demo would create tons of "bad" data for our analytics around things like activity rates and stuff. Is that no longer a concern, or did we just decide this was worth it?

So that is still a concern. We currently generate this data once a day (overnight in a cron job). This PR ONLY re-generates the activity data if a teacher has specifically edited or deleted the `Quill Demo` classroom (because of the data concern you mention). I'm going to monitor this, but the hope (and my initial look at the data) is that this will only need to be re-generated maybe once-a-day. Also, if this works well enough, we might be able to remove the overnight reset.

PR #9756, opened 2022-09-23
Lms/fix concept result instructions

anathomical:
## WHAT
Add a Worker to process `ConceptResult` records and remove the `instructions` key from their `extra_metadata` if the value in that key is already stored in the related `concept_result_instructions` record.
## WHY
A bug that we fixed a while ago (#9593) was causing this data to be written to `extra_metadata` even though we were normalizing it.  While we no longer produce records with this duplication, we should clean up the places where we did it to reduce redunancy.
## HOW
Just cycle across `ConceptResult` records, confirm that the value in `extra_metadata['instructions']` matches the value in `concept_result_instructions.text`, and if so, remove the value in `extra_metadata`.

### Notion Card Links
https://www.notion.so/quill/Write-a-script-to-strip-instructions-from-ConceptResult-extra_metadata-if-a-related-instructions--2e23118c244e4bdcbebc93df54ae0ce0

PR Checklist | Your Answer
------------ | -------------

### DIFF
--- a/services/QuillLMS/app/workers/clean_concept_result_instructions_worker.rb
+++ b/services/QuillLMS/app/workers/clean_concept_result_instructions_worker.rb
@@ -5,8 +5,6 @@ class CleanConceptResultInstructionsWorker
   sidekiq_options queue: SidekiqQueue::MIGRATION
 
   def perform(start, stop)
-    stop ||= ConceptResult.maximum(:id)
-
     ConceptResult.includes(:concept_result_instructions).where(id: start..stop).find_each do |concept_result|
       next unless concept_result.concept_result_instructions&.text
       next unless concept_result.concept_result_instructions.text == concept_result.extra_metadata['instructions']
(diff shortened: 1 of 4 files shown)


services/QuillLMS/app/workers/clean_concept_result_instructions_worker.rb
```diff
@@ -0,0 +1,17 @@
+# frozen_string_literal: true
+
+class CleanConceptResultInstructionsWorker
+  include Sidekiq::Worker
+  sidekiq_options queue: SidekiqQueue::MIGRATION
+
+  def perform(start, stop)
+    stop ||= ConceptResult.maximum(:id)
```
happythenewsad: `stop` should never be nil, so the conditional assignment is not necessary

happythenewsad (approved):

services/QuillLMS/app/workers/clean_concept_result_instructions_worker.rb
anathomical: I've been using the ability to pass `nil` as the value for `stop` in testing, but you're probably right that it's not necessary to keep in for the final version.

dandrabik (approved):
Looks good to me.

eadams17 (approved):

PR #9757, opened 2022-09-23
lms/bugfix-activity-session-teacher-fix

anathomical:
## WHAT
Update teacher fix so that it maintains assigned students on ClassroomUnits, even for students who haven't started the activity yet.
## WHY
The old logic replaced the assignment list with the users who has ActivitySessions.  But only users who have started the activity have ActivitySessions.  This meant that students who hadn't started the activity when the teacher fix was run were being unassigned, which is not the intended behavior.
## HOW
Just use a `union` of the existing list of assignments and the list of assignments recovered from ActivitySessions rather than stomping the former with the latter.

### Notion Card Links
https://www.notion.so/quill/Adding-students-to-an-activity-pack-unchecks-other-students-names-a5ee8958fde94619ac59e4ccc079bdd8

PR Checklist | Your Answer
------------ | -------------

### DIFF
--- a/services/QuillLMS/app/controllers/teacher_fix_controller.rb
+++ b/services/QuillLMS/app/controllers/teacher_fix_controller.rb
@@ -68,7 +68,9 @@ class TeacherFixController < ApplicationController
       if unit
         ClassroomUnit.unscoped.where(unit_id: unit.id).each do |cu|
           activity_sessions = ActivitySession.unscoped.where(classroom_unit_id: cu.id)
-          cu.update(visible: true, assigned_student_ids: activity_sessions.map(&:user_id))
+          recovered_assigned_students = activity_sessions.map(&:user_id)
+          cu.update(visible: true, assigned_student_ids: cu.assigned_student_ids.union(recovered_assigned_students))
+          #cu.update(visible: true, assigned_student_ids: activity_sessions.map(&:user_id))
           activity_sessions.update_all(visible: true)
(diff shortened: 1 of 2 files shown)


services/QuillLMS/app/controllers/teacher_fix_controller.rb
```diff
@@ -68,7 +68,9 @@ def recover_activity_sessions
       if unit
         ClassroomUnit.unscoped.where(unit_id: unit.id).each do |cu|
           activity_sessions = ActivitySession.unscoped.where(classroom_unit_id: cu.id)
-          cu.update(visible: true, assigned_student_ids: activity_sessions.map(&:user_id))
+          recovered_assigned_students = activity_sessions.map(&:user_id)
+          cu.update(visible: true, assigned_student_ids: cu.assigned_student_ids.union(recovered_assigned_students))
+          #cu.update(visible: true, assigned_student_ids: activity_sessions.map(&:user_id))
```
eadams17: we want to remove this, right?

eadams17 (approved):
great catch! thanks for putting out this fix

services/QuillLMS/app/controllers/teacher_fix_controller.rb
anathomical: Good catch!

emilia-friedberg (approved):

PR #9758, opened 2022-09-23
Faster Rspec via Spring

dandrabik:
## WHAT
Ensure that Spring is set up properly for rspec.
## WHY
This cuts down on test loading time. It's sooo much faster to use spring to preload the app, but we also need to change the `cache_classes` config so we don't have two competing cache sources (spring and rails).
### Screenshots
This is also a PSA that you should be running `bin/rspec` and not `bundle exec rspec` (For me, it's `11 seconds` vs. `0.5 seconds` to load!)
Screen Shot 2022-09-23 at 3 03 47 PM

Screen Shot 2022-09-23 at 3 04 21 PM

PR Checklist | Your Answer
------------ | -------------
Have you added and/or updated tests? | No, config change


services/QuillLMS/spec/system/activity_pack_assignment_spec.rb
```diff
@@ -69,6 +69,9 @@
     # the following line checks for the presence of an element that only loads once activity data has loaded, ensuring that the assign click will be successful
     first('.review-activities-data-table-section').click
     click_on 'Assign pack to classes'
+
+    # TODO: Fix bug in which you can click too fast and break this flow (remove sleep when that is done)
```
dandrabik, line 73: I believe this is an unrelated to my code failure. I'll create a card to fix this.

services/QuillLMS/bin/spring
```diff
@@ -5,12 +5,14 @@
 # It gets overwritten when you run the `spring binstub` command.
 
 unless defined?(Spring)
-  require "rubygems"
-  require "bundler"
+  require 'rubygems'
```
happythenewsad, line 8: Have you confirmed that this code will not run spring in the prod environment? 

There are special instructions for avoiding loading spring in prod, here: https://github.com/rails/spring#deployment

happythenewsad (commented):
I'm excited for this to merge

services/QuillLMS/bin/spring
dandrabik: Note, this code isn't adding `spring`. It's already in the `Gemfile`. This code just edits the config for caching classes in `test`, and updates CircleCI to use `bin/rspec`.

Heroku already skips loading `development` and `test` gems (but again, nothing is changing here):


Screen Shot 2022-09-26 at 1 59 00 PM

services/QuillLMS/bin/rspec
```diff
@@ -2,9 +2,9 @@
 # frozen_string_literal: true
 
 begin
-  load File.expand_path("../spring", __FILE__)
-rescue LoadError
-  nil
+  load File.expand_path('../spring', __FILE__)
+rescue LoadError => e
+  raise unless e.message.include?('spring')
```
happythenewsad, line 7: Is this because the error message will include the string 'spring' when rails is in develop mode, but not production?

happythenewsad (approved):

anathomical (approved):
Looks good to me.  Excited about faster tests.

services/QuillLMS/bin/rspec
dandrabik: So this is the generated code written by the [Spring config script](https://github.com/rails/spring#setup). I believe the intention is to swallow errors related to loading spring (spring is additive for caching purpose, but doesn't affect the code, so we shouldn't error for it), but not other non-spring errors.

Note, this is an `rspec` file, so it's not related to development or production.

PR #9763, opened 2022-09-27
Require an Activity's topics to be level one; Rake task for attaching new topics

cissyxyu:
## WHAT
We're changing the way we store topics on Activities through the ActivityTopics table. 
Instead of saving an activity's level zero, level one, level two, and level three topics, we now want to only store an Activity's level 1 topics (multiple level 1 topics can now be stored per activity).
Also included a .csv with new topics to attach to Activities and a rake task to attach them.

## WHY
This will allow us to give one activity multiple level 1 topics. We also don't need to store an Activity's level 2/level 3 topics anymore because we can find those topics by backtracing the `parent_id `of the level 1 topic, now that we've required all level 1 topics to have associated level 2 and level 3 topics.

## HOW
Add a validation to the `ActivityTopic` relationship to only save if the topic is level 1.
Modify places where we send an activities' level 1, level 2, and level 3 topic to instead send up to 3 level 1 topics. This will be more informative for Curriculum and reporting going forward.

### Screenshots

### Notion Card Links
https://www.notion.so/quill/Activities-can-have-3-tags-733866911ca14340bb5b34eccd1c2dd0

PR Checklist | Your Answer
------------ | -------------


services/QuillLMS/app/models/activity_topic.rb
```diff
@@ -18,9 +18,15 @@
...
+    record.errors.add :base, "Topics associated with Activities must be level one." unless record.topic.present? && record.topic.level_one?
+  end
+end
+
 class ActivityTopic < ApplicationRecord
   belongs_to :activity
   belongs_to :topic
 
-  validates :topic_id, presence: true
+  validates_with ActivityTopicLevelOneValidator
```
anathomical: I think this is slightly cleaner if it's a function on the `ActivityTopic` model rather than a separate class:
```
validate :activity_topic_level_one_validator

def activity_topic_level_one_validator
   errors.add(:base, "Topics associated with Activities must be level one." unless topic.present? && topic.level_one?
end
```

services/QuillLMS/app/models/topic.rb
```diff
@@ -64,6 +68,9 @@ def level_three?
   end
 
   private def valid_parent_structure?
-    level_two? ? level_three_parent? : no_parent?
+    return level_two_parent? if level_one?
+    return level_three_parent? if level_two?
+
+    no_parent?
```
anathomical: Is it the case that `no_parent?` `Topic`s are only valid if they're `level_three?`?  Should we include that as part of this validation?

services/QuillLMS/lib/tasks/attach_new_topics_to_activities.rake
```diff
@@ -0,0 +1,18 @@
+# frozen_string_literal: true
```
anathomical: Not a huge deal, but you could combine these two `rake` tasks into a single file given that you're already sharing the namespace.  A single `topics.rake` file just makes things slightly more compact, but it's no big deal if you don't do so.

anathomical (approved):
A couple of minor comments, but this looks good to me overall.

happythenewsad (approved):

brendanshean (approved):
Nice work on this!

PR #9766, opened 2022-09-27
add publish date to unit activities and update student query to use them

emilia-friedberg:
## WHAT
Add `publish_date` to unit activities and update the student query to use them to filter activities for display.

## WHY
This is pre-work for the scheduled activities feature, which will rely on this background logic.

## HOW
Add a migration and update the SQL query used to generate the data for the student profile to filter out activities that have a publish date that is after the current time in their teacher's timezone. Then update tests (including some test file cleanup).

### Screenshots

### Notion Card Links
https://www.notion.so/quill/Add-publish-date-to-unit-activities-9a2cd31390694496afad5d3d4acce991
https://www.notion.so/quill/Update-student-query-to-take-publish-date-into-account-with-teacher-s-timezone-0371a4638eb549fc8a6c8a139408b299

PR Checklist | Your Answer
------------ | -------------

### DIFF
--- a/services/QuillLMS/app/models/unit_activity.rb
+++ b/services/QuillLMS/app/models/unit_activity.rb
@@ -166,7 +166,12 @@ class UnitActivity < ApplicationRecord
           AND cu.visible = true
           AND unit.visible = true
           AND ua.visible = true
-          AND (ua.publish_date IS NULL OR ua.publish_date <= NOW() at time zone '#{supply_timezone(teachers.time_zone)}')
+          AND
+          (
+            ua.publish_date IS NULL
+            OR (teachers.time_zone IS NOT NULL AND ua.publish_date <= NOW() at time zone teachers.time_zone)
+            OR (teachers.time_zone IS NULL AND ua.publish_date <= NOW())
+          )
         GROUP BY
           unit.id,
           unit.name,
@@ -177,6 +182,7 @@ class UnitActivity < ApplicationRecord
           activity.id,
           activity.uid,
           ua.due_date,
+          ua.publish_date,
           ua.created_at,
           unit_activity_id,
           cuas.completed,
@@ -195,10 +201,6 @@ class UnitActivity < ApplicationRecord
           ua.id ASC
       SQL
     ).to_a
-  end
-
-  def supply_timezone(teachers_timezone)
-    teachers_timezone || DEFAULT_TIMEZONE
   end
 
 end
--- a/services/QuillLMS/app/models/unit_activity.rb
+++ b/services/QuillLMS/app/models/unit_activity.rb
@@ -124,8 +124,8 @@ class UnitActivity < ApplicationRecord
           ua.activity_id,
           MAX(acts.updated_at) AS act_sesh_updated_at,
           ua.order_number,
-          ua.due_date,
-          ua.publish_date,
+          CASE WHEN teachers.time_zone IS NOT NULL THEN ua.due_date at time zone teachers.time_zone ELSE ua.due_date END AS due_date,
+          CASE WHEN teachers.time_zone IS NOT NULL THEN ua.publish_date at time zone teachers.time_zone ELSE ua.publish_date END AS publish_date,
           pre_activity.id AS pre_activity_id,
           cu.created_at AS unit_activity_created_at,
           COALESCE(cuas.locked, false) AS locked,
@@ -166,12 +166,7 @@ class UnitActivity < ApplicationRecord
           AND cu.visible = true
           AND unit.visible = true
           AND ua.visible = true
-          AND
-          (
-            ua.publish_date IS NULL
-            OR (teachers.time_zone IS NOT NULL AND ua.publish_date <= NOW() at time zone teachers.time_zone)
-            OR (teachers.time_zone IS NULL AND ua.publish_date <= NOW())
-          )
+          AND (ua.publish_date IS NULL OR ua.publish_date <= NOW())
         GROUP BY
           unit.id,
           unit.name,
@@ -190,7 +185,8 @@ class UnitActivity < ApplicationRecord
           cuas.pinned,
           ua.id,
           activity_classifications.key,
-          pre_activity.id
+          pre_activity.id,
+          teachers.time_zone
         ORDER BY
           pinned DESC,
           locked ASC,
--- a/services/QuillLMS/app/models/unit_activity.rb
+++ b/services/QuillLMS/app/models/unit_activity.rb
@@ -36,6 +36,8 @@ class UnitActivity < ApplicationRecord
 
   after_save :hide_appropriate_activity_sessions, :teacher_checkbox
 
+  DEFAULT_TIMEZONE = "UTC"
+
   def teacher_checkbox
     return unless unit
     return unless unit.user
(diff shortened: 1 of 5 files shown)


services/QuillLMS/app/models/unit_activity.rb
```diff
@@ -157,11 +160,18 @@ def self.get_classroom_user_profile(classroom_id, user_id)
...
         WHERE #{user_id.to_i} = ANY (cu.assigned_student_ids::int[])
           AND cu.classroom_id = #{classroom_id.to_i}
           AND cu.visible = true
           AND unit.visible = true
           AND ua.visible = true
+          AND
+          (
+            ua.publish_date IS NULL
+            OR (teachers.time_zone IS NOT NULL AND ua.publish_date <= NOW() at time zone teachers.time_zone)
+            OR (teachers.time_zone IS NULL AND ua.publish_date <= NOW())
```
anathomical: PostgreSQL actually does some magic to simplify this for you.  All timestamps are stored in the db as UTC and converted during input and output.  That means that you don't have to check the teacher timezone and do the conversion yourself.  You should be able to simply always check `ua_publish_date <= NOW()` without any of the timezone stuff because `NOW()` will default to UTC, and the timestamp is already UTC in the db.

services/QuillLMS/db/structure.sql
```diff
@@ -138,11 +125,11 @@ CREATE FUNCTION public.timespent_question(act_sess integer, question character v
                       'epoch' FROM (activity_sessions.completed_at - activity_sessions.started_at)
                     )
                 END) INTO time_spent FROM activity_sessions WHERE id = act_sess AND state='finished';
-
+
```
anathomical, line 128: I know this file is auto-generated, but it's weird that it's making these empty lines with spaces in them.

services/QuillLMS/spec/models/unit_activity_spec.rb
```diff
@@ -146,19 +142,66 @@
...
+      expect(unit_activities.count).to eq(2)
     end
+
+    describe 'publish dates' do
+
+      it 'includes unit activities that do not have a publish date' do
+        unit_activity.update(publish_date: nil)
+        lessons_unit_activity.update(publish_date: nil)
+        unit_activities = UnitActivity.get_classroom_user_profile(classroom.id, student.id)
+        expect(unit_activities.count).to eq(2)
```
anathomical: The `count` is a fine thing to check here, but it may also be worth checking that the expected values are in the results.  That's a bit tricky here because of the use of raw SQL, but you could maybe do something like:
```
expect(unit_activities.map{ |ua| ua['ua_id'] }).to include(unit_activity.id, lessons_unit_activity.id)
```
If that makes sense, it may be worth adding similar conditions to the follow-on cases as well.

services/QuillLMS/spec/models/unit_activity_spec.rb
```diff
@@ -146,19 +142,66 @@
...
+      end
+
+      it 'includes unit activities that have a publish date that have already passed in the timezone of the teacher' do
```
anathomical: Per my note on timezones in the query, I don't think you actually need to do anything related to timezones in the specs for the query.  The one thing we do need to do is make sure that if the teacher does have a timezone set, that we apply said timezone when we write the `publish_date` to the database, but I think that's actually going to be in a different PR since there's no code in this one to handle that writing.

anathomical (approved):
This works as-is, but I've made a couple of notes about how timezones work in PostgreSQL that should allow you to make both the code and the test cases a little simpl,er.

services/QuillLMS/app/models/unit_activity.rb
happythenewsad: Interesting!

happythenewsad (approved):

services/QuillLMS/spec/models/unit_activity_spec.rb
emilia-friedberg: @anathomical That's an interesting point, and one I hadn't really considered (saving the publish date itself in the teacher's timezone, instead of in UTC). Newer versions of postgresql do have a data type for date times with time zones, but we'd have to do a funky migration to store it that way (by default the `datetime` columns are saved as `timestamp without time zone` data types). I think this is the first and only place in our database we'd be introducing that. I think it's likely, but I'm not sure, that the new data type would translate cleanly into our frontend without having to make additional adjustments. Saving the timezone on the publish date itself also means that adjusting the set time zone wouldn't automatically adjust the publish dates, which may or may not be behavior we want. Due dates are not currently saved with timezones, though we could update that. I'm not sure whether the simplification of this query is worth introducing the new data type or not - what do you think?

anathomical: @emilia-friedberg Sorry, I think I was unclear.  What I meant was that we need to make sure that whatever payload we pass from the front-end with datetime data in it includes a timezone so that PostgreSQL can accurately know how to convert that value to UTC for storage in the database.  I think that you are totally correct that we shouldn't store timezone offsets directly on the column, even if PostgreSQL makes that possible as that would be a big change in how we handle all other datetimes.

Shorter version: I don't think we need timezone-related specs on this PR since we're just going to rely on PostgreSQL to manage UTC conversion.  We should have some timezone-related specs on whatever code we write that sets `publish_date`, but that's going to be in a different PR as I understand it.

emilia-friedberg:
Re-requesting review after talking to @anathomical - we arrived at the conclusion that, while we do not currently adjust any datetimes in our database to account for the user's timezone, it probably makes sense to do so for both due dates and publish dates (ie, all the datetimes that are explicitly set by the user) so that the datetimes in the database are internally consistent (you don't have to convert anything to understand a timeline around when an assignment was published, started, completed, and due). I've updated the query here to convert the output of the `publish_date` and `due_date`, but given that we intend to save it this way we no longer need the funky logic for comparison to `NOW()`. In a future PR I'll be ensuring that we save both publish date and due date with the appropriate offset.

services/QuillLMS/app/models/unit_activity.rb
```diff
@@ -36,6 +36,8 @@ class UnitActivity < ApplicationRecord
 
   after_save :hide_appropriate_activity_sessions, :teacher_checkbox
 
+  DEFAULT_TIMEZONE = "UTC"
```
anathomical: Is this actually being used anywhere?

services/QuillLMS/app/models/unit_activity.rb
```diff
@@ -122,7 +124,8 @@ def self.get_classroom_user_profile(classroom_id, user_id)
           ua.activity_id,
           MAX(acts.updated_at) AS act_sesh_updated_at,
           ua.order_number,
-          ua.due_date,
+          CASE WHEN teachers.time_zone IS NOT NULL THEN ua.due_date at time zone teachers.time_zone ELSE ua.due_date END AS due_date,
+          CASE WHEN teachers.time_zone IS NOT NULL THEN ua.publish_date at time zone teachers.time_zone ELSE ua.publish_date END AS publish_date,
```
anathomical, line 126: It feels like this particular pattern could get cumbersome, but I think that there's probably not a simpler approach in cases where we're writing raw SQL rather than going through ActiveRecord.  So let's go with this for now, but keep an eye out for how often we have to do this sort of thing.

anathomical (approved):
Looks good to me.

PR #9770, opened 2022-09-28
New Staff UI to add up to three topics to Activities

cissyxyu:
## WHAT
Created a new React UI for staff to add up to three level one topics to Activities.

## WHY
We want staff to be able to set topics on activities and create new topics.

## HOW
Create a React class TopicGroup that contains all 3 selector elements you would need to select a new level one topic (the level 3 topic selector, level 2 topic selector, and level 1 topic selector). Make the parent class Topic create three of these TopicGroups instead of three TopicColumns like it was doing before.

### Screenshots
Screen Shot 2022-09-28 at 3 50 46 PM

### Notion Card Links
https://www.notion.so/quill/Activity-UI-Supports-3-tags-163deee555324ebbb7baa1b6e08a8af3

PR Checklist | Your Answer
------------ | -------------

### DIFF
--- a/services/QuillLMS/client/app/bundles/Staff/components/activityForm/topicColumn.tsx
+++ b/services/QuillLMS/client/app/bundles/Staff/components/activityForm/topicColumn.tsx
@@ -1,10 +1,9 @@
 import * as React from 'react'
-import { ReactTable, TextFilter, } from '../../../Shared/index'
 
-const TopicColumn = ({ createNewTopicNew, levelNumber, selectTopicNew, getFilteredOptionsForLevelNew, topic, removeTopic }) => {
+const TopicColumn = ({ createNewTopic, levelNumber, selectTopic, getFilteredOptionsForLevel, topic, removeTopic }) => {
   const [newTopicName, setNewTopicName] = React.useState('')
 
-  const options = getFilteredOptionsForLevelNew(levelNumber, topic.id)
+  const options = getFilteredOptionsForLevel(levelNumber, topic.id)
 
   function handleNewTopicNameChange(e) {
     setNewTopicName(e.target.value)
@@ -16,26 +15,17 @@ const TopicColumn = ({ createNewTopicNew, levelNumber, selectTopicNew, getFilter
       level: 1,
       name: newTopicName
     }
-    createNewTopicNew(newTopic)
+    createNewTopic(newTopic)
     setNewTopicName('')
   }
 
+  const handleClickSelectTopic = (id) => {
+    selectTopic(id, levelNumber)
+  }
 
-  const columns = (selectedOption) => ([
-    {
-      Header: 'Name',
-      accessor: 'name',
-      key: 'name',
-      Cell: ({row}) => (<div className={`record-cell ${selectedOption && selectedOption.id === row.original.id  ? 'selected' : ''}`} onClick={() => selectTopicNew(row.original.id, levelNumber)}>{row.original.name}</div>),
-      sortType:  (a, b) => (a && b ? a.original.name.localeCompare(b.original.name) : 0),
-      Filter: TextFilter,
-      filter: (rows, idArray, filterValue) => {
-        return rows.filter(row => {
-          return row.original.name ? row.original.name.toLowerCase().includes(filterValue.toLowerCase()) : ''
-        })
-      },
-    }
-  ]);
+  function handleClickRemoveTopic(e) {
+    removeTopic(levelNumber)
+  }
 
   const selectedOption = topic
 
@@ -45,7 +35,7 @@ const TopicColumn = ({ createNewTopicNew, levelNumber, selectTopicNew, getFilter
   if (selectedOption) {
     selectedOptionElement = (<div className="selected-option" >
       <span>{selectedOption.name}</span>
-      <button className="interactive-wrapper" onClick={(e) => removeTopic(levelNumber)}><i className="fas fa-times" /></button>
+      <button aria-label="remove topic" className="interactive-wrapper" onClick={handleClickRemoveTopic} type="button"><i className="fas fa-times" /></button>
     </div>)
   }
 
@@ -55,7 +45,7 @@ const TopicColumn = ({ createNewTopicNew, levelNumber, selectTopicNew, getFilter
       buttonClassName+= ' disabled'
     }
     newTopicField = (<div className="new-topic-field">
-      <label>New Topic Level 1</label>
+      <span>New Topic Level 1</span>
       <input disabled={!options.length} onChange={handleNewTopicNameChange} value={newTopicName} />
       <button className={buttonClassName} disabled={!options.length} onClick={handleClickAddTopic}>Add</button>
     </div>)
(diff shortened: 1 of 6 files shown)


services/QuillLMS/client/app/bundles/Staff/components/activityForm/topicColumn.tsx
```diff
@@ -65,12 +55,20 @@ const TopicColumn = ({ createNewTopic, levelNumber, getFilteredOptionsForLevel,
...
-        filterable
-      />
+      <div>
+        {
+          options.map((option, i) => {
+            let clickEventHelper = () => {
+              handleClickSelectTopic(option.id)
+            };
+            return (
+              <button className="record-cell" key={i} onClick={clickEventHelper} onKeyDown={clickEventHelper} type="button">
```
emilia-friedberg, line 65: You shouldn't need an `onKeyDown` for a `button` element (it will automatically be handled by the `onClick`, but that isn't true for interactive elements).

services/QuillLMS/client/app/bundles/Staff/components/activityForm/topicColumn.tsx
```diff
@@ -65,12 +55,20 @@ const TopicColumn = ({ createNewTopic, levelNumber, getFilteredOptionsForLevel,
...
-      <ReactTable
-        columns={columns(selectedOption)}
-        data={options}
-        defaultFilterMethod={(filter, row) => row._original.name ? row._original.name.toLowerCase().includes(filter.value.toLowerCase()) : ''}
-        filterable
-      />
+      <div>
+        {
+          options.map((option, i) => {
+            let clickEventHelper = () => {
```
emilia-friedberg, line 61: This doesn't get reset so it should just be a const.

services/QuillLMS/client/app/bundles/Staff/components/activityForm/topicColumn.tsx
```diff
@@ -65,12 +55,20 @@ const TopicColumn = ({ createNewTopic, levelNumber, getFilteredOptionsForLevel,
...
+              handleClickSelectTopic(option.id)
+            };
+            return (
+              <button className="record-cell" key={i} onClick={clickEventHelper} onKeyDown={clickEventHelper} type="button">
+                {option.name}
+              </button>
+            )
+          })
+        }
```
emilia-friedberg: For code cleanliness I would probably pull this out into a constant instead of doing the iteration inside of the render.

services/QuillLMS/client/app/bundles/Staff/components/activityForm/topicColumn.tsx
emilia-friedberg: You shouldn't need an onKeyDown for a button element (it will automatically be handled by the onClick, but that isn't true for interactive elements).

services/QuillLMS/client/app/bundles/Staff/components/activityForm/topicColumn.tsx
emilia-friedberg: This doesn't get reset so it should just be a const.

services/QuillLMS/client/app/bundles/Staff/components/activityForm/topicGroup.tsx
```diff
@@ -0,0 +1,80 @@
+import * as React from 'react'
+
+import TopicColumn from './topicColumn'
+
+const TopicGroup = ({ createNewTopic, findParentTopic, topic, topicOptions, handleLevelOneChange }) => {
+
+  const [levelOneTopic, setLevelOneTopic] = React.useState(topic)
+  const [levelTwoTopic, setLevelTwoTopic] = levelOneTopic.id ? React.useState(findParentTopic(topic)) : React.useState({})
+  const [levelThreeTopic, setLevelThreeTopic] = levelOneTopic.id ? React.useState(findParentTopic(levelTwoTopic)) : React.useState({})
```
emilia-friedberg: Should this be `levelTwoTopic.id ? ...` ?

services/QuillLMS/client/app/bundles/Staff/components/activityForm/topicGroup.tsx
```diff
@@ -0,0 +1,80 @@
...
+
+  React.useEffect(() => {
+    if (levelOneTopic.parent_id !== levelTwoTopic.id) {
+      handleLevelOneChange(levelOneTopic.id, null)
```
emilia-friedberg: It feels like a little bit of a code smell to explicitly call `handleLevelOneChange` before you set it when we could use an effect hook to do that when the value changes. That's a little complicated by the fact that you need the old value to pass up to the function, but you could do something like [this](https://stackoverflow.com/questions/53446020/how-to-compare-oldvalues-and-newvalues-on-react-hooks-useeffect) which I think would still be cleaner. Up to you, just wanted to flag it as a little unparadigmatic.

services/QuillLMS/client/app/bundles/Staff/components/activityForm/topicGroup.tsx
```diff
@@ -0,0 +1,80 @@
...
+  function selectTopic(topicId, level) {
+    const newTopic = topicOptions.find(t => t.id === topicId)
+    if (level === 1) {
+      handleLevelOneChange(levelOneTopic.id, newTopic.id)
+      setLevelOneTopic(newTopic)
+    } else if (level === 2) {
+      setLevelTwoTopic(newTopic)
+    } else if (level === 3) {
+      setLevelThreeTopic(newTopic)
+    }
```
emilia-friedberg: Since this `if` block basically gets repeated in the `removeTopic` function below, I wonder if it's worth setting up a const like:

```
const levelNumberToFunction = {
  1: setLevelOneTopic,
  2: setLevelTwoTopic,
  3: setLevelThreeTopic
}
```

And then here you could just call `levelNumberToTopic[level](newTopic)` and below you would call `levelNumberToTopic[level]({})`.

services/QuillLMS/client/app/bundles/Staff/components/activityForm/topics.tsx
```diff
@@ -1,6 +1,8 @@
 import * as React from 'react'
 
-import TopicColumn from './topicColumn'
+import TopicGroup from './topicGroup'
+
+import '../../../../../../app/assets/stylesheets/cms/activities.scss'
```
emilia-friedberg: Let's import this into a style file rather than into the component itself - that isn't really a pattern we use elsewhere (arguably it's one we should move towards, but I think better to be consistent for now).

emilia-friedberg (approved):
Mostly nits - feel free to re-request a review if you want eyes on it again but otherwise LGTM once those changes are made. Nice work!

happythenewsad (approved):

emilia-friedberg (approved):

emilia-friedberg:
Thanks for making those changes! Looks great.

brendanshean (approved):
This looks very cool!

PR #9778, opened 2022-09-29
remove ip lookup code for now

emilia-friedberg:
## WHAT
Fix bug where we were calling `.timezone` on all objects passed back by Geocoder, but not all of them have this method available.

## WHY
We don't want this to error.

## HOW
The original bug was a result of me not reading docs closely enough, and it turns out that not every object Geocoder passes back will [have the same methods](https://github.com/alexreisner/geocoder#custom-result-handling). Calling `.timezone` worked on the instances I was testing before production, but we have had several thousand instances where this is not the case. According to the docs, extracting the key directly from the `data` object should be the safer way to handle this, and testing in the rails console bears out that it works with some of the ids that we were seeing fail with the earlier version.

### Screenshots

### Notion Card Links

PR Checklist | Your Answer
------------ | -------------
Have you added and/or updated tests? |  YES (updated Geocoder stub to reflect new data expectations)

### DIFF
--- a/services/QuillLMS/app/models/user.rb
+++ b/services/QuillLMS/app/models/user.rb
@@ -664,7 +664,7 @@ class User < ApplicationRecord
       self.time_zone = school_timezone
     else
       geocoder_results = Geocoder.search(ip_address)
-      self.time_zone = geocoder_results&.first&.timezone
+      self.time_zone = geocoder_results&.first&.data ? geocoder_results.first.data['timezone'] : nil
     end
   end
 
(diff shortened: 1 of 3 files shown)


eadams17 (approved):

anathomical (approved):
It is crazy that it works this way, but good job finding that.

emilia-friedberg:
> It is crazy that it works this way, but good job finding that.

Thank you! I think it's because it's actually just aggregating a bunch of different apis. It does seem really strange to me too.

PR #9780, opened 2022-09-29
lms/add-organization-diagnostics-to-vitally

anathomical:
## WHAT
Add diagnostic roll-up stats to Vitally payloads for districts
## WHY
Partnerships wants to be able to generate reports about this data directly from Vitally instead of having to use Metabase and copying that data into emails.
## HOW
Write some functions to do the roll-up calculations and add those to the `vitally_data` function on the `District` model

### Notion Card Links
https://www.notion.so/quill/PS-Add-diagnostic-assignment-traits-to-accounts-in-Vitally-953745c1d1a74996b108e54a9e75f235

PR Checklist | Your Answer
------------ | -------------

### DIFF
--- a/services/QuillLMS/app/models/district.rb
+++ b/services/QuillLMS/app/models/district.rb
@@ -62,8 +62,9 @@ class District < ApplicationRecord
         zipcode: zipcode,
         phone: phone,
         total_students: total_students,
-        total_schools: total_schools
-      }.merge(vitally_diagnostic_rollups)
+        total_schools: total_schools,
+        **vitally_diagnostic_rollups
+      }
     }
   end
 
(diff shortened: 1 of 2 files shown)


services/QuillLMS/app/models/district.rb
```diff
@@ -63,7 +63,57 @@ def vitally_data
         phone: phone,
         total_students: total_students,
         total_schools: total_schools
-      }
+      }.merge(vitally_diagnostic_rollups)
```
happythenewsad: Since the elements of `vitally_diagnostic_rollups` are conceptual siblings to the other members of this hash, it makes sense to use the double asterisk expansion syntax within this hash literal, rather than `merge`. Open to other opinions if you have them, though.

services/QuillLMS/app/models/district.rb
```diff
@@ -63,7 +63,57 @@ def vitally_data
...
+    }
+  end
+
+  def vitally_diagnostic_rollups
```
happythenewsad, line 71: Curious to see what the performance hit is in prod.

happythenewsad (approved):

brendanshean (approved):
Looks good to me.

PR #9781, opened 2022-09-29
Demo Reset - Fix Race Condition

dandrabik:
## WHAT
The Demo `reset_account` method deletes a `classroom` and recreates it. There is a race condition in which the demo reset gets in a conflicting state when it's triggered again before finishing, i.e. the `classroom` being populated with data gets deleted. This causes `activity_sessions` being created to fail (`null` id) because the `student` is no longer in the `classroom_unit`. This validation on [activity_session](https://github.com/empirical-org/Empirical-Core/blob/develop/services/QuillLMS/app/models/activity_session.rb#L518-L529) is the error being thrown..

This PR wraps the `reset_account` and `create_demo` methods transactions to prevent the data from changing while it's being created.

Also, I moved the call to reset_account to a worker for all calls for now, just in case there are other issues (they won't be user facing).

## WHY
We don't want users to get an error while going to the demo account, and we don't want to leave the demo in a weird state.

Given that `reset_account` takes 10 seconds+ or so to run, this race condition is easy to trigger using two rails consoles (I don't have a great way to write a test for it though).
## HOW
Wrap the methods in transactions to ensure only one set is committed, and use the ActiveRecord method calls that raise errors, e.g. `create!`, and rescue `ActiveRecord::RecordInvalid` so it halts execution at that point and does not re-run. 

### Screenshots
**Activity Session Error**
Screen Shot 2022-09-29 at 5 10 54 PM
**Code triggering it**
Screen Shot 2022-09-29 at 5 34 29 PM
**Subsequent Errored job with a null activity session**
Screen Shot 2022-09-29 at 5 35 22 PM

PR Checklist | Your Answer
------------ | -------------

### DIFF
--- a/services/QuillLMS/app/controllers/teachers/classroom_manager_controller.rb
+++ b/services/QuillLMS/app/controllers/teachers/classroom_manager_controller.rb
@@ -8,7 +8,7 @@ class Teachers::ClassroomManagerController < ApplicationController
 
   respond_to :json, :html
 
-  around_action :force_writer_db_role, only: [:assign, :dashboard, :lesson_planner, :view_demo]
+  around_action :force_writer_db_role, only: [:assign, :dashboard, :lesson_planner]
 
   before_action :teacher_or_public_activity_packs, except: [:unset_preview_as_student, :unset_view_demo]
   # WARNING: these filter methods check against classroom_id, not id.
--- a/services/QuillLMS/app/controllers/teachers/classroom_manager_controller.rb
+++ b/services/QuillLMS/app/controllers/teachers/classroom_manager_controller.rb
@@ -8,7 +8,7 @@ class Teachers::ClassroomManagerController < ApplicationController
 
   respond_to :json, :html
 
-  around_action :force_writer_db_role, only: [:assign, :dashboard, :lesson_planner]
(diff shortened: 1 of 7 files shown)


anathomical (approved):
This looks good to me.  I'm not sure if we actually want to re-use this pattern, but I found an instance where we used `Thread` to implement a race-condition spec in the code-base: https://github.com/empirical-org/Empirical-Core/blob/a710c85a2c2c3db921d5c192ad2cc19c157fa8d4/services/QuillLMS/spec/models/user_activity_classification_spec.rb#L66-L92

dandrabik:
> This looks good to me. I'm not sure if we actually want to re-use this pattern, but I found an instance where we used `Thread` to implement a race-condition spec in the code-base:
> 
> https://github.com/empirical-org/Empirical-Core/blob/a710c85a2c2c3db921d5c192ad2cc19c157fa8d4/services/QuillLMS/spec/models/user_activity_classification_spec.rb#L66-L92

Good Flag. I've added a race condition test here: https://github.com/empirical-org/Empirical-Core/pull/9781/commits/89ab7b01ab096719d429e349b6361e2c8b76b05b

PR #9783, opened 2022-09-30
lms/normalize-evidence-responses-for-feedback

anathomical:
## WHAT
Add a string-normalization step before sending entries for feedback
## WHY
We want to make sure that we process out any special characters (like "smart" quotes) that might make it into a student's response before we send the response to our various algorithms for feedback.  This lets us avoid the work of making each of our algorithms special-character tolerant separately.
## HOW
- Copy the replacement logic we use in the front-end string normalizer (https://github.com/empirical-org/quill-string-normalizer/blob/master/src/main.ts)
- Add some additional normalization from the spec in Notion

### Notion Card Links
https://www.notion.so/quill/Normalize-common-UTF-8-characters-in-Evidence-student-responses-063b5c5c29214e9991252dd438fd5c3d

PR Checklist | Your Answer
------------ | -------------

### DIFF
--- a/services/QuillLMS/engines/evidence/lib/evidence/evidence/check.rb
+++ b/services/QuillLMS/engines/evidence/lib/evidence/evidence/check.rb
@@ -23,9 +23,9 @@ module Evidence
     }
 
     def self.get_feedback(entry, prompt, previous_feedback, feedback_types=nil)
-      entry = normalize_entry_text(entry)
+      normalized_entry = StringNormalizer.new(entry).run
 
-      triggered_check = find_triggered_check(entry, prompt, previous_feedback, feedback_types)
+      triggered_check = find_triggered_check(normalized_entry, prompt, previous_feedback, feedback_types)
 
       triggered_check || fallback_feedback
     end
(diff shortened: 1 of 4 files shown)


services/QuillLMS/engines/evidence/lib/evidence/evidence/check.rb
```diff
@@ -23,6 +23,8 @@ class NoMatchedFeedbackTypesError < StandardError; end
     }
 
     def self.get_feedback(entry, prompt, previous_feedback, feedback_types=nil)
+      entry = normalize_entry_text(entry)
```
dandrabik: let's assign this to a new variable rather that overwriting `entry`.

services/QuillLMS/engines/evidence/lib/evidence/evidence/check.rb
```diff
@@ -78,5 +80,16 @@ def self.checks_to_run(feedback_types)
 
       filtered_checks
     end
+
+    def self.normalize_entry_text(entry)
```
dandrabik: Take or leave. You might consider making this is own small `ApplicationService` class ([example)](https://github.com/empirical-org/Empirical-Core/blob/develop/services/QuillLMS/engines/evidence/lib/evidence/evidence/html_tag_remover.rb), since I could see this being reused.

services/QuillLMS/engines/evidence/spec/lib/check_spec.rb
```diff
@@ -191,5 +191,64 @@ module Evidence
         end.to raise_error(Check::NoMatchedFeedbackTypesError)
       end
     end
+
+    context 'normalize_entry_text' do
```
dandrabik: It would be nice to see an additional test for a string with all of these `utf-8` characters and not the codes. That would be a nice full process test that would also verify that the codes are the correct ones for the characters you expect.

dandrabik (approved):
Overall, looks good. I left a few comments you should look at and I'm requesting an additional test (I don't need to re-review though).

Thanks for working on this.

services/QuillLMS/engines/evidence/lib/evidence/evidence/check.rb
anathomical: Good call.  Updating.

services/QuillLMS/engines/evidence/lib/evidence/evidence/check.rb
anathomical: Seems reasonable.  I've gone ahead and done this.

services/QuillLMS/engines/evidence/spec/lib/check_spec.rb
anathomical: That seems reasonable.  I've added a case that explicitly checks that normalizing a string of all of the intended characters works.

emilia-friedberg (approved):

happythenewsad (approved):

PR #9784, opened 2022-09-30
Bug fix for invitations controller.

dandrabik:
## WHAT
Fixing two validation bug in the invitations controller that I'm seeing show up a lot in our dead sidekiq jobs for unsendable emails:
1. Emails with spaces (fix these)
2. Two emails entered at once (return an error, so the user fixes it).

We could try to parse the second one into `n` number of emails, but that's a larger refactor (I'm making a card to explore it), so we should least give the user a chance to fix it (and not have it fail silently).
## WHY
Invites are important to our growth. These fail silently on send.
## HOW
Add failing tests, fix them. Strip whitespace and use a different regex validator (`User::VALID_EMAIL_REGEX`).
### Screenshots

### Notion Card Links

PR Checklist | Your Answer
------------ | -------------

### DIFF
--- a/services/QuillLMS/app/controllers/invitations_controller.rb
+++ b/services/QuillLMS/app/controllers/invitations_controller.rb
@@ -59,14 +59,14 @@ class InvitationsController < ApplicationController
   end
 
   private def validate_email_format
-    return if @invitee_email =~ /.+@.+\..+/i
+    return if @invitee_email =~ User::VALID_EMAIL_REGEX
 
     raise StandardError, "Please make sure you've entered a valid email."
   end
 
   private def set_classroom_ids_and_inviteee_email
     @classroom_ids = params[:classroom_ids]
-    @invitee_email = params[:invitee_email]
+    @invitee_email = params[:invitee_email].gsub(/\s/, '') #strip whitespace
   end
 
 
--- a/services/QuillLMS/app/controllers/invitations_controller.rb
+++ b/services/QuillLMS/app/controllers/invitations_controller.rb
@@ -66,7 +66,7 @@ class InvitationsController < ApplicationController
 
   private def set_classroom_ids_and_inviteee_email
     @classroom_ids = params[:classroom_ids]
-    @invitee_email = params[:invitee_email]&.gsub(/\s/, '') #strip whitespace
+    @invitee_email = params[:invitee_email]&.gsub(/\s/, '') # strip whitespace
   end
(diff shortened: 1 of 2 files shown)


services/QuillLMS/app/controllers/invitations_controller.rb
```diff
@@ -59,14 +59,14 @@ def destroy_pending_invitations_from_specific_inviter
   end
 
   private def validate_email_format
-    return if @invitee_email =~ /.+@.+\..+/i
+    return if @invitee_email =~ User::VALID_EMAIL_REGEX
```
happythenewsad, line 62: 👍

services/QuillLMS/app/controllers/invitations_controller.rb
```diff
@@ -59,14 +59,14 @@ def destroy_pending_invitations_from_specific_inviter
...
 
     raise StandardError, "Please make sure you've entered a valid email."
   end
 
   private def set_classroom_ids_and_inviteee_email
     @classroom_ids = params[:classroom_ids]
-    @invitee_email = params[:invitee_email]
+    @invitee_email = params[:invitee_email]&.gsub(/\s/, '') #strip whitespace
```
anathomical: Should we just use `.strip` here and rely on the validation regex to catch instances where there is whitespace in the email address itself?  Probably doesn't matter either way.

happythenewsad: I'm surprised the linter doesn't complain about the lack of a space after the comment hash symbol

anathomical (approved):
Looks good to me.

happythenewsad (approved):

services/QuillLMS/app/controllers/invitations_controller.rb
dandrabik: I don't think it's a big deal either way. Looking at the [failed invites](https://data.quill.org/question/958-invalid-invites), there are several that would be valid if we stripped the space and seem like obvious typos (e.g. `email@gmail. com`). Given emails can't have whitespace in the middle, I think it's ok to attempt to fix that for the user, then validate, rather than sending it back to them to fix (I could be convinced otherwise).

anathomical: That seems reasonable to me.  I don't object to us fixing something for users rather than making them do it.

PR #9788, opened 2022-10-03
scheduled activities

emilia-friedberg:
## WHAT
This PR comprises the bulk of the work for the scheduled activities feature (the migrations to add publish dates and some work to set user timezones automatically came previously). It consists of frontend updates to the Review Activities section of the assignment flow, the My Activities page, the Activity Summary, and the student profile, and backend updates to the queries that provide data to those views as well as a `before_save` callback on unit activities that adjusts submitted due dates and publish dates to the timezone of the assigning user.

## WHY
This is a long-requested feature, and one that Peter Sharkey did a bunch of UX research to plan.

## HOW
On the backend, mostly adding the `before_save` hook and updating queries to pass the necessary information to the frontend. On the frontend, I introduced a new datepicker component that could handle times as well as dates (a feature requirement), and implemented it in all the places we'd previously been setting due dates, as well as adding the publish dates. The bulk of this project was spent trying to get my brain to translate UTC to EST times correctly in order to test that what I was doing was working (I think we got there).

### Screenshots
Review Activities section
review_activities

Datepicker open
calendar_open

My activities page - activity pack
my_activities

My activities page - copy modal
copy_modal

My activities page - invalid dates selected
invalid_dates

Activity summary page - open tooltip for scheduled activity
activity_summary_scheduled

Activity summary page - open tooltip for published activity
activity_summary_published

Student profile activity with due date
student_profile

### Notion Card Links
https://www.notion.so/quill/PS-Activity-pack-scheduling-01e66b545385448a8e086daeb62deafa

PR Checklist | Your Answer
------------ | -------------


eadams17 (approved):
this looks great-- I don't think I noticed anything off. great job knocking this all out!

services/QuillLMS/app/controllers/teachers/unit_activities_controller.rb
```diff
@@ -23,10 +23,8 @@ def hide
     render json: {}
   end
 
-  def update_multiple_due_dates
-    base_unit_activities = UnitActivity.where(id: params[:unit_activity_ids])
-    activity_ids = base_unit_activities.map(&:activity_id)
-    UnitActivity.where(activity_id: activity_ids, unit_id: base_unit_activities.first.unit_id).update_all(due_date: params[:due_date])
+  def update_multiple_dates
+    UnitActivity.where(id: params[:unit_activity_ids]).each { |ua| ua.update(params[:date_attribute] => params[:date]) }
```
anathomical: I was going to suggest replacing the `each` call here with `update_all`, but then remembered that `update_all` bypasses before and after hooks, so doing `each ... update` is probably the safer thing to do here.

services/QuillLMS/app/controllers/teachers/units_controller.rb
```diff
@@ -252,7 +254,16 @@ def score_info
           cu.unit_id AS unit_id,
           array_to_json(cu.assigned_student_ids) AS assigned_student_ids,
           COUNT(DISTINCT students_classrooms.id) AS class_size,
-          ua.due_date,
+          ua.due_date #{user_timezone_offset_string} AS due_date,
+          ua.publish_date #{user_timezone_offset_string} AS publish_date,
+          CASE WHEN unit_owner.time_zone IS NOT NULL THEN ua.publish_date at time zone unit_owner.time_zone >= NOW() at time zone unit_owner.time_zone ELSE ua.publish_date >= NOW() END AS scheduled_in_user_time_zone,
```
anathomical: If we're properly converting from user's timezone to UTC when we write datetimes to the database, I don't know that we need this `CASE` value.  I think we can just do `ua.publish_date > NOW()`?  Maybe?  (Per our Zoom discussion, timezones are really confusing, so I could be wrong.)  My thinking is that if the line below (260) works without messing with timezone conversion, this one should, too.

services/QuillLMS/app/controllers/teachers/units_controller.rb
```diff
@@ -252,7 +254,16 @@ def score_info
...
+          ua.publish_date >= NOW() AS scheduled,
+          NOW() as now_with_no_time_zone,
+          NOW() at time zone 'utc' AS now_in_utc,
+          NOW() at time zone unit_owner.time_zone AS now_in_est,
```
anathomical: Should we be calling this `now_in_est`?  For a lot of cases, it will be Eastern time, but it kind of explicitly might not be here, right?

services/QuillLMS/app/controllers/teachers/units_controller.rb
```diff
@@ -252,7 +254,16 @@ def score_info
...
+          ua.publish_date AS publish_date_with_no_time_zone,
+          ua.publish_date at time zone 'utc' AS publish_date_in_utc,
+          ua.publish_date at time zone unit_owner.time_zone AS publish_date_in_est,
```
anathomical: Ditto here about `est` in the name.

services/QuillLMS/app/controllers/teachers/units_controller.rb
```diff
@@ -252,7 +254,16 @@ def score_info
           cu.unit_id AS unit_id,
           array_to_json(cu.assigned_student_ids) AS assigned_student_ids,
           COUNT(DISTINCT students_classrooms.id) AS class_size,
-          ua.due_date,
+          ua.due_date #{user_timezone_offset_string} AS due_date,
+          ua.publish_date #{user_timezone_offset_string} AS publish_date,
+          CASE WHEN unit_owner.time_zone IS NOT NULL THEN ua.publish_date at time zone unit_owner.time_zone >= NOW() at time zone unit_owner.time_zone ELSE ua.publish_date >= NOW() END AS scheduled_in_user_time_zone,
+          ua.publish_date >= NOW() AS scheduled,
+          NOW() as now_with_no_time_zone,
+          NOW() at time zone 'utc' AS now_in_utc,
```
anathomical: No objection to having both of these keys, but will they always be the same?

services/QuillLMS/app/models/unit_activity.rb
```diff
@@ -82,6 +84,18 @@ def from_valid_date_for_activity_analysis?
...
+    return unless due_date.present? && unit&.user&.time_zone&.present?
+
+    self.due_date = due_date - unit.user.utc_offset
+  end
+
+  def adjust_publish_date_for_timezone
+    return unless publish_date.present? && unit&.user&.time_zone&.present?
+
+    self.publish_date = publish_date - unit.user.utc_offset
+  end
```
anathomical: Do we run into issues here if the front-end passes timezone aware values for time?  Like `"12:00:00+04:00"` or something?  It's possible that we're comfortable that this won't ever happen because of the way that we're building the UX, but I wanted to double-check just in case.

I think I'm worried that we're a little to naive in how we're approaching this.  This code makes a hard assumption that if a user has a timezone assigned, the incoming values will be in that timezone.  (And not in UTC or something, which I think might be the case if we do something automatic like an endpoint that sets all due dates to `1.day.from_now` or something.)  I wonder if, before we apply these offsets, we don't need to do some Timezone conversions to make sure that any timestamps are in the user's timezone if they weren't already (or we assign the user's timezone if the timestamps weren't timezone aware).  That might also be confusing though.

Ugh.  I hate trying to figure out timezone stuff.

services/QuillLMS/app/queries/scorebook/query.rb
```diff
@@ -21,6 +21,12 @@ def self.run(classroom_id, current_page=1, unit_id=nil, begin_date=nil, end_date
           activity.description AS activity_description,
           MAX(acts.updated_at) #{user_timezone_offset} AS updated_at,
           MIN(acts.started_at) #{user_timezone_offset} AS started_at,
+          unit_activities.due_date #{user_timezone_offset} AS due_date,
+          unit_activities.publish_date #{user_timezone_offset} AS publish_date,
+          unit_activities.created_at #{user_timezone_offset} AS unit_activity_created_at,
+          CASE WHEN unit_owner.time_zone IS NOT NULL THEN unit_activities.publish_date at time zone unit_owner.time_zone >= NOW() at time zone unit_owner.time_zone ELSE unit_activities.publish_date >= NOW() END AS scheduled_in_user_time_zone,
```
anathomical: Same as my earlier note in the `unit_activities_controller` regarding possibly not needing this `CASE` condition.

services/QuillLMS/client/app/bundles/Teacher/components/assignment_flow/create_unit/stage2/review_activities.jsx
```diff
@@ -23,51 +40,130 @@ const tableHeaders = [
...
+    const dates = this.props[dateAttributeKey]
+    const activity = activities.find(act => act.id === id)
+    const formattedDate = date || null
+    this.setState({ erroredActivityIds: [], })
+    assignActivityDate(activity, formattedDate, dateAttributeKey);
+  }
+
+  triggerSnackbar() {
+    this.setState({ snackbarVisible: true }, () => {
+      setTimeout(() => this.setState({ snackbarVisible: false, }), defaultSnackbarTimeout)
```
anathomical, line 92: Is this a place to use `useSnackbarMonitor`?

services/QuillLMS/client/app/bundles/Teacher/components/assignment_flow/create_unit/stage2/review_activities.jsx
```diff
@@ -23,51 +40,130 @@ const tableHeaders = [
...
+      }, this.triggerSnackbar)
+    } else {
+      this.handleDateChange(date, id, PUBLISH_DATE_ATTRIBUTE_KEY)
+    }
+  }
+
+  copyPublishDateToAll = () => {
+    const { publishDates, dueDates, activities, } = this.props
+    const publishDate = publishDates[activities[0].id]
+    const activitiesWithEarlierDueDates = activities.filter(act => dueDates[act.id] && (publishDate >= dueDates[act.id]))
```
anathomical, line 123: This is only repeated twice, so no big deal, but it might make sense for this to be a component function since this const is defined both here and in `copyDueDateToAll`?

services/QuillLMS/client/app/bundles/Teacher/components/assignment_flow/manage_units/manage_units.jsx
```diff
@@ -120,6 +120,15 @@ export default class ManageUnits extends React.Component {
       activityClassificationId: u.activity_classification_id,
       classroomId: u.classroom_id,
       dueDate: u.due_date ? u.due_date.replace(' ', 'T') : null,
+      publishDate: u.publish_date ? u.publish_date.replace(' ', 'T') : null,
+      compare_now_with_no_time_zone: u.now_with_no_time_zone ? u.now_with_no_time_zone.replace(' ', 'T') : null,
+      compare_now_in_utc: u.now_in_utc ? u.now_in_utc.replace(' ', 'T') : null,
+      compare_now_in_est: u.now_in_est ? u.now_in_est.replace(' ', 'T') : null,
+      compare_publish_date_with_no_time_zone: u.publish_date_with_no_time_zone ? u.publish_date_with_no_time_zone.replace(' ', 'T') : null,
+      compare_publish_date_in_utc: u.publish_date_in_utc ? u.publish_date_in_utc.replace(' ', 'T') : null,
+      compare_publish_date_in_est: u.publish_date_in_est ? u.publish_date_in_est.replace(' ', 'T') : null,
```
anathomical: This makes me think we need some sort of utility function.  `normalize_timestamp`, maybe?  This ternary logic appears 7 times here, it makes more sense to me that we just pass the values to a function rather than repeat the logic.

services/QuillLMS/client/app/bundles/Teacher/components/assignment_flow/manage_units/manage_units.jsx
```diff
@@ -151,6 +160,15 @@ export default class ManageUnits extends React.Component {
             classroomId: u.classroom_id,
             createdAt: u.ca_created_at,
             dueDate: u.due_date ? u.due_date.replace(' ', 'T') : null,
+            publishDate: u.publish_date ? u.publish_date.replace(' ', 'T') : null,
```
anathomical: If this is a function, you can re-use it again for these.

services/QuillLMS/client/app/bundles/Teacher/components/general_components/tooltip/activity_details.jsx
```diff
@@ -39,19 +39,21 @@ export const ActivityDetails = ({ data }) => {
...
     }
 
+    const publishDateForDisplay = publishDate || unitActivityCreatedAt
     const objectiveSection = activity_description ? <p><strong>Objectives:</strong>{` ${activity_description}`}</p> : <span />
-    const dateSection = dateTitle ? <p><strong>{`${dateTitle}: `}</strong>{`${moment(dateBody).format('MMMM D, YYYY')}`}</p> : <span />
-    const completedSection = completedTitle ? <p><strong>{`${completedTitle}: `}</strong>{`${moment(completedBody).format('MMMM D, YYYY')}`}</p> : <span />
+    const publishDateSection = <p><strong>{scheduled ? 'Scheduled For: ' : 'Published: '}</strong>{`${moment.utc(publishDateForDisplay).format('MMMM D, YYYY [at] h:mm a')}`}</p>
+    const dueDateSection = <p><strong>Due: </strong>{dueDate ? `${moment.utc(dueDate).format('MMMM D, YYYY [at] h:mm a')}` : 'N/A'}</p>
+    const dateSection = dateTitle ? <p><strong>{`${dateTitle}: `}</strong>{`${moment.utc(dateBody).format('MMMM D, YYYY')}`}</p> : <span />
+    const completedSection = completedTitle ? <p><strong>{`${completedTitle}: `}</strong>{`${moment.utc(completedBody).format('MMMM D, YYYY')}`}</p> : <span />
```
anathomical, line 50: Should we really be coercing these timestamps to UTC?  Am I mis-remembering how we were thinking to do timezones?

anathomical (approved):
Overall this looks good to me.  I have some concerns about timezone conversions, but can't articulate them well enough to make concrete suggestions.  Maybe some other folks can chime in on the topic.

services/QuillLMS/app/controllers/teachers/unit_activities_controller.rb
brendanshean: I would add a check here that 
```
if params[:date_attribute].in?('publish_date', 'due_date')
  UnitActivity
    .where(id: params[:unit_activity_ids])
    .each { |ua| ua.update(params[:date_attribute] => params[:date]) }
end
```
 to ensure that only the columns we're expecting to be updated are updated.

For example, we wouldn't want this to be allowed: `put :update_multiple_dates, params: { unit_activity_ids: [unit_activity.id, unit_activity2.id, unit_activity3.id], date_attribute: 'activity_id', date: 5 }`

brendanshean: I think there's also a larger question about error states here.  Currently, if the update fails, it fails silently.  For that reason I tend to prefer `update!` over `update`, in which case you'll need a transaction block since there are multiple objects being created:

```
  UnitActivity
      .where(id: params[:unit_activity_ids])
      .each { |ua| ua.update!(params[:date_attribute] => params[:date]) }
end
```

services/QuillLMS/spec/controllers/teachers/units_controller_spec.rb
```diff
@@ -232,6 +233,54 @@
       expect(parsed_response.length).to eq(1)
     end
 
+    describe 'when the teacher has a time zone' do
+      it 'should return activities with publish dates in the future as scheduled' do
+        tz_string = 'America/New_York'
+        teacher.update(time_zone: tz_string)
+        publish_date = Time.now.utc + 1.hour
+        # have to do update_columns here because otherwise the publish date is offset by a callback
```
brendanshean, line 241: Appropriate use of comments in these specs.

services/QuillLMS/spec/models/unit_activity_spec.rb
```diff
@@ -140,6 +142,62 @@
...
   end
 
+  describe 'adjust_due_date_for_timezone' do
+    it 'adjusts the submitted time to the utc equivalent of that time in the timezone of the teacher' do
+      tz_string = 'America/New_York'
+      teacher.update(time_zone: tz_string)
+      due_date = Time.now.utc - 1.hour
+      unit_activity.due_date = due_date
+      unit_activity.adjust_due_date_for_timezone
+      expect(unit_activity.due_date).to eq(due_date + 4.hours)
```
brendanshean: 4.hours here is a magic number.  I would define a new variable like 
`let(:et_offset) { 4.hours }`

However, I think UTC does not observe daylight savings so this test will probably start failing in a month.  We probably need some way of calculating this offset.

services/QuillLMS/spec/models/unit_activity_spec.rb
```diff
@@ -186,9 +244,10 @@
         tz_string = 'America/New_York'
         teacher.update(time_zone: tz_string)
         publish_date = Time.now.utc - 1.hour
-        unit_activity.update(publish_date: publish_date)
+        # have to do update_columns here because otherwise the publish date is offset by a callback
+        unit_activity.update_columns(publish_date: publish_date)
         unit_activities = UnitActivity.get_classroom_user_profile(classroom.id, student.id)
-        expect(unit_activities.find{ |ua| ua['ua_id'].to_i == unit_activity.id}['publish_date'].to_time.strftime('%a %b %d %H:%M:%S %Z %Y')).to eq((publish_date - teacher.utc_offset).to_time.strftime('%a %b %d %H:%M:%S %Z %Y'))
+        expect(unit_activities.find{ |ua| ua['ua_id'].to_i == unit_activity.id}['publish_date'].to_time.strftime('%a %b %d %H:%M:%S %Z %Y')).to eq((publish_date + teacher.utc_offset).to_time.strftime('%a %b %d %H:%M:%S %Z %Y'))
```
brendanshean: This line is really long.  Consider breaking it up into two:
```
  expect(unit_activities.find{ |ua| ua['ua_id'].to_i == unit_activity.id}['publish_date'].to_time.strftime('%a %b %d %H:%M:%S %Z %Y'))
    .to eq((publish_date + teacher.utc_offset).to_time.strftime('%a %b %d %H:%M:%S %Z %Y'))
```

services/QuillLMS/app/controllers/teachers/units_controller.rb
emilia-friedberg: @anathomical you're right, this line isn't actually in use here (I was testing some equivalencies and forgot to remove it, it's `scheduled` that gets used on the frontend). I'll pull this out.

services/QuillLMS/app/controllers/teachers/units_controller.rb
emilia-friedberg: The last several comments are still me forgetting to do cleanup! These were keys that were unused that I was logging for testing. Sorry for the confusion.

services/QuillLMS/client/app/bundles/Teacher/components/assignment_flow/create_unit/stage2/review_activities.jsx
emilia-friedberg: No, because this is a class component and not a functional component. `useSnackbarMonitor` is a hook, which can't be used in a class component.

services/QuillLMS/app/models/unit_activity.rb
emilia-friedberg: This code is making almost the inverse assumption - that the passed value is coming in UTC (that's why it's possible to subtract the offset and get the equivalent of the UTC time in the user's timezone). The frontend is set up to only ever pass utc objects back (all conversions are in `moment.utc`, which handles that - we can't rely on the browser time because it doesn't necessarily match the user's saved timezone on the backend). I think you're right that we shouldn't rely on the frontend and should instead confirm that the time is in utc before the subtraction. I'll update it to reflect that.

brendanshean (approved):
This is an impressive PR.  I left a few comments and I do agree with Thomas' concerns about timezone conversions.  Maybe have a few people on the team try it out that don't live in Eastern time zone for smoke testing?  Other than that, great work on this.

services/QuillLMS/client/app/bundles/Teacher/components/assignment_flow/create_unit/stage2/review_activities.jsx
emilia-friedberg: They are very similar, but they aren't identical (the variables change).

services/QuillLMS/client/app/bundles/Teacher/components/general_components/tooltip/activity_details.jsx
emilia-friedberg: The backend sends up timezone-agnostic times (because that's what's in the database). If you don't convert them to UTC, they are automatically converted to the browser time (which may or may not be what the user has saved as their timezone in the backend). So the coercion here is really just saying "don't convert me" - because what the backend is sending up is the correct time based on the user's saved timezone.

services/QuillLMS/client/app/bundles/Teacher/components/assignment_flow/manage_units/manage_units.jsx
emilia-friedberg: More code that was actually just being used for testing! Sorry!

services/QuillLMS/client/app/bundles/Teacher/components/assignment_flow/manage_units/manage_units.jsx
emilia-friedberg: It actually looks like we don't need this anymore (I had copied it from what we were doing with due date) so I just removed it from both.

services/QuillLMS/spec/models/unit_activity_spec.rb
brendanshean: I'm not sure this is fixed yet.

emilia-friedberg: @brendanshean it was fixed in [this commit](https://github.com/empirical-org/Empirical-Core/pull/9788/commits/391ebba9c8ef69d1b56f1230bf3069d446967c33) (I updated it to use a calculated offset rather than a constant).

brendanshean: Perfect.  For some reason the UI was still linking me to the original `4.hours` code

PR #9790, opened 2022-10-03
Bulk update subscriptions

happythenewsad:
## WHAT
Script for bulk updating premium credits from CSV file. 

## WHY
Curriculum ask (for TAC users) 

## HOW
Rake file 

### Screenshots

### Notion Card Links
https://www.notion.so/quill/Need-253-Ts-to-be-given-free-Teacher-Premium-for-5-yrs-in-addition-to-whatever-they-have-currently-985fe99b6b724d3dac9686d1d9a3747c

PR Checklist | Your Answer
------------ | -------------
Have you added and/or updated tests? |  tested manually on unsubscribed and subscribed users

### DIFF
--- a/services/QuillLMS/app/models/subscription.rb
+++ b/services/QuillLMS/app/models/subscription.rb
@@ -42,7 +42,6 @@ class Subscription < ApplicationRecord
   has_many :user_subscriptions
   has_many :users, through: :user_subscriptions
   belongs_to :purchaser, class_name: "User"
-  belongs_to :subscription_type
   belongs_to :plan, optional: true
 
   validates :expiration, presence: true
--- a/services/QuillLMS/lib/tasks/subscriptions.rake
+++ b/services/QuillLMS/lib/tasks/subscriptions.rake
@@ -20,8 +20,8 @@ namespace :subscriptions do
       end
 
       begin
-        CreditTransaction.create!(user_id: user.id, amount: args[:amount_in_days].to_i, source: user)
-        puts "CreditTransaction created for #{user.email}"
+        credit_transaction = CreditTransaction.create!(user_id: user.id, amount: args[:amount_in_days].to_i, source: user)
+        puts "CreditTransaction with id #{credit_transaction.id} created for #{user.email}"
 
         user.redeem_credit unless user.subscription
       rescue => e
--- a/services/QuillLMS/lib/tasks/subscriptions.rake
+++ b/services/QuillLMS/lib/tasks/subscriptions.rake
@@ -1,0 +1,32 @@
+# frozen_string_literal: true
+require 'csv'
+
+namespace :subscriptions do
+  desc 'Add credit to users from a CSV file'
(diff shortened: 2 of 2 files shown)


services/QuillLMS/app/models/subscription.rb
```diff
@@ -42,7 +42,6 @@ class RenewalNilStripeCustomer < StandardError; end
   has_many :user_subscriptions
   has_many :users, through: :user_subscriptions
   belongs_to :purchaser, class_name: "User"
-  belongs_to :subscription_type
```
happythenewsad, line 45: This change is not germane to the PR, but I discovered it was unused so I'm cleaning it up.

services/QuillLMS/lib/tasks/subscriptions.rake
```diff
@@ -0,0 +1,33 @@
...
+    CSV.parse(iostream, headers: true) do |row|
+      email = row['email']
+      user = User.find_by_email(email)
+      if user.nil? || user.role != 'teacher'
+        puts "Unable to locate teacher with email #{email}"
+        next
+      end
+
+      begin
+        CreditTransaction.create!(user_id: user.id, amount: args[:amount_in_days].to_i, source: user)
```
anathomical: Not sure if there's an easy way to resolve this, but this line probably makes this script non-idempotent.  If it crashes halfway through, you can't just run it again with the same input file unless it's safe (and intended) to double-credit some users.

anathomical (approved):
This is probably fine?  I have some concerns about idempotency, but if this is really a one-off request, it probably won't matter?

services/QuillLMS/lib/tasks/subscriptions.rake
happythenewsad: Right - it's hard to make a script involving `CreditTransaction` idempotent, since it behaves like a changelog. My solution was to log liberally, so that in the worst case we can create additional `CreditTransactions` on failed iterations. However, the script did not fail on staging, so I'm not expecting this to happen.

brendanshean: You could create an ids array of created CreditTransactions and return them at the end of the rake task.  Then if you had to re-run the rake task cross check for user ids and remove those in the csv of already created credit transactions.

brendanshean: I'm not sure about the `source: user` assignment in the CreditTransaction creation:  https://github.com/empirical-org/Empirical-Core/blob/develop/services/QuillLMS/app/models/credit_transaction.rb#L25

The code as-is seems to be creating a new source type without a corresponding action in the model file.  Not sure this matters but figured I would point it out.

happythenewsad: I added the transaction id logging. 

Regarding `source: user` - I believe it's the most logical thing, since the premium credit originates from the user's status as a TAC; they were not referred, and there is no pre-existing subscription. Also, we don't do much with this value anyway.

brendanshean (approved):

PR #9796, opened 2022-10-04
Demo - Hide classroom rather than destroy.

dandrabik:
## WHAT
There seems to be some callback issues with destroying a classroom, so instead, I'm going to hide the demo classrooms in the callback. Also, don't retry this job if it fails.
## WHY
The [reset-job is falling](https://sentry.io/organizations/quillorg-5s/issues/3639965126/?project=11238&query=is%3Aunresolved) for a dependent-destroy foreign-key conflict. 
## HOW
`.update(visible: false)` instead of `.destroy`

There are likely some `dependent: destroy`'s that are needed on the `classroom.destroy`, but I ran into issues making that work (it doesn't seem to be common to destroy a classroom in our codebase). Instead, I mimicked the 'archive' function which teachers use, which marks a `classroom` as `visible=false`.

Note, I also refactored a few slow tests (call the slow setup once, and chain within test with `.and`)

PR Checklist | Your Answer
------------ | -------------

### DIFF
--- a/services/.rubocop.yml
+++ b/services/.rubocop.yml
@@ -132,6 +132,7 @@ Lint/AmbiguousBlockAssociation:
   Enabled: true
   IgnoredMethods:
     - change
+    - not_change
 
 Lint/AmbiguousOperatorPrecedence:
   Enabled: true
--- a/services/QuillLMS/app/services/demo/report_demo_creator.rb
+++ b/services/QuillLMS/app/services/demo/report_demo_creator.rb
@@ -315,7 +315,7 @@ module Demo::ReportDemoCreator
 
       # Note, you can't early return within a transaction in Rails 6.1+
       if teacher && demo_classroom_modified?(teacher)
-        demo_classroom(teacher)&.destroy
+        demo_classroom(teacher)&.update(visible: false)
         create_demo_classroom_data(teacher, teacher_demo: true)
       end
     end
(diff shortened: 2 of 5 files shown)


services/.rubocop.yml
```diff
@@ -132,6 +132,7 @@ Lint/AmbiguousBlockAssociation:
   Enabled: true
   IgnoredMethods:
     - change
+    - not_change
```
dandrabik, line 135: This is in the same spirit as `change` since it has identical syntax.

brendanshean (approved):
The logic for reset_account makes sense. 
 
On a sidekiq note, does _not_ retrying failed jobs leave the demo account in some weird state indefinitely or is the expectation that it will be cleaned up in subsequent jobs?

anathomical (approved):

dandrabik:
> The logic for reset_account makes sense.
> 
> On a sidekiq note, does _not_ retrying failed jobs leave the demo account in some weird state indefinitely or is the expectation that it will be cleaned up in subsequent jobs?

Well, I removed retries because if it doesn't work on the first try, it's not going to work on subsequent tries (there's nothing really gained by this being retried).

Since the bulk of the action is in a transaction, the account shouldn't be left in a weird state (it would all be rolled back). If the clean-up is still needed, it would be cleaned up on the next job (although, to be clear, a retry is more a sign of a bug, than something that needs to be re-run).

PR #9800, opened 2022-10-05
Fix display bugs caused by new Topic structures in Explore Activities

cissyxyu:
## WHAT
We deployed a restructuring of topics and it broke the teacher display for Activities in the "Create your own activity pack" page. This PR adjusts the front end display so that it now accommodates multiple level one topics on an activity.

## WHY
We want teachers to be able to see all the level one topics associated with an activity, as well as those topics' parent topics.

## HOW
1. Add an /index endpoint in the back end for Topics.
2. For the activity row React component, call /topics to get all the topics and then trace back the parent relationship of each level one topic to display.
3. Instead of showing one topic and its parent in the Activity Row collapsed view, show all of the activity's topics.

### Screenshots
Screen Shot 2022-10-05 at 1 45 52 PM

### Notion Card Links
https://www.notion.so/quill/Activity-Filter-displays-3-tags-eee251fae3c54199a4df46b4d5a95d05

PR Checklist | Your Answer
------------ | -------------


emilia-friedberg (commented):
I could be talked out of this, but it seems to me like it probably makes more sense to bundle all the topic stuff up into the same place we get all the rest of this data for this page (the `ActivitySearch` query). Then I don't think you'd have to make any changes to the frontend at all. Is there a benefit to having the two different ajax calls here?

services/QuillLMS/spec/controllers/topics_controller_spec.rb
```diff
@@ -0,0 +1,22 @@
...
+  let!(:topic) { create(:topic) }
+  let!(:level_two_topic) { create(:topic, level: 2) }
+  let!(:level_three_topic) { create(:topic, level: 3) }
+
+  describe '#index' do
+    it 'should return all the topics' do
+      get :index
+      parsed_response = JSON.parse(response.body)
+      topic_results = parsed_response["topics"]
+      Topic.all.each do |t|
```
brendanshean: Since you're only comparing ids, you could rewrite lines 15-17 as `expect(topic_results.pluck("id")).to match_array Topic.pluck(:id)`

brendanshean (approved):
Thanks for fixing this.

brendanshean:
> I could be talked out of this, but it seems to me like it probably makes more sense to bundle all the topic stuff up into the same place we get all the rest of this data for this page (the `ActivitySearch` query). Then I don't think you'd have to make any changes to the frontend at all. Is there a benefit to having the two different ajax calls here?

Just saw this and I agree.  The topics payload is pretty small so it seems like you could just get it all in there initially.

services/QuillLMS/client/app/bundles/Teacher/components/assignment_flow/create_unit/custom_activity_pack/activity_row.tsx
```diff
@@ -104,38 +105,49 @@ const ActivityRowConcept = ({ conceptName, }: { conceptName?: string }) => {
...
+        <span className="attribute-section extended-topic">
+          <span><img alt="Globe icon" src={topicSrc} />{thirdLevelTopic?.name}</span>
+          {diagonalDivider}
+          <span>{secondLevelTopic?.name}</span>
+          {diagonalDivider}
+          <span className="lowest-level">{topic.name}</span>
+        </span>
+      )
+    })
+    return topicsWithParentBadges
```
anathomical: This probably doesn't matter, but you could probably just return the results of `topics.map` rather than saving it to a `const` and returning the `const`.  Same thing below.

services/QuillLMS/client/app/bundles/Teacher/components/assignment_flow/create_unit/custom_activity_pack/index.tsx
```diff
@@ -122,6 +123,14 @@ const CustomActivityPack = ({
     updateFilteredActivities()
   }
 
+  function getTopics() {
```
anathomical: Is this getting called anywhere?

services/QuillLMS/client/app/bundles/Teacher/components/assignment_flow/create_unit/custom_activity_pack/shared.ts
```diff
@@ -155,12 +155,7 @@ export const filters = {
 }
 
 export const stringifyLowerLevelTopics = (topics) => {
```
anathomical, line 157: Is this still an accurate name for this function?  It might be if we're only passing in arrays of lower level topics, so I figured I'd double-check.

anathomical (approved):
A couple of minor notes on style and naming, but this looks fine to me.

emilia-friedberg (approved):
I think my original comment was ambiguous - I was actually suggesting that you update the `ActivitySearch` query [itself](https://github.com/empirical-org/Empirical-Core/blob/develop/services/QuillLMS/app/queries/activity_search.rb#L9) to contain the associated topics for the given activity (through a self-join to the topics table). You'd probably then have to update [this function](https://github.com/empirical-org/Empirical-Core/blob/develop/services/QuillLMS/app/services/activity_search_wrapper.rb#L50), but you wouldn't have to change anything on the frontend to handle the data shape, and it would work the same way as all the other attributes. I think your new solution is good too, though, and both approaches have tradeoffs.

PR #9809, opened 2022-10-07
Evidence/re use annotated examples

anathomical:
## WHAT
Change the way that we handle "Annotated Examples" or "Hints" for Rules in Evidence.  This PR creates a new interface to create, edit, and delete Hints from a central place, rather than having to hunt down what Rule they're attached to.  In addition, instead of a set of form fields on Rules that allow you to define Hint data, it's been replaced with a drop-down selector of all hints, and a set of display data to show you exactly what that Hint contains when you select it.
## WHY
We want to be able to use the same Hint on multiple Rules (rather than the current one-to-one relationship set we maintain).  We also want to be able to make broad tweaks to those hints from a central interface.
## HOW
- Create a new Evidence CRUD resource for `Hint` objects so that they can be managed independently of `Rules`
- Modify the `Rule` model so that it can no longer accept nested `Hint` data, preventing us from accidentally impacting a `Hint` with broader implications than is obvious
- Add a migration and update `ActiveRecord` relationships to make it so that one `Hint` can be related to many `Rules` rather than the old one-to-one setup
- Create a new UI/UX for managing Hints in the Evidence dashboard
- Update Rule detail UI/UX that used to let staff create Hints in-place on those pages with a drop-down that lets staff select existing Hint objects to assign them to a Rule

### Screenshots
image
image
image

### Notion Card Links
https://www.notion.so/quill/Save-Re-use-annotated-examples-13ff1dbd4a5340d38639dd0412440644

PR Checklist | Your Answer
------------ | -------------


services/QuillLMS/engines/evidence/app/controllers/evidence/hints_controller.rb
```diff
@@ -0,0 +1,60 @@
+# frozen_string_literal: true
+
+module Evidence
+  class HintsController < ApiController
+    before_action :set_hint, only: [:create, :show, :update, :destroy]
+
+    # GET /hints.json
+    def index
+      @hints = Evidence::Hint
```
brendanshean, line 9: I think this assignment would be more consistent / conventional:

```
@hints = Evidence::Hint.all
render json: @hints
```

services/QuillLMS/client/app/bundles/Staff/components/evidence/configureRules/ruleHintPicker.tsx
```diff
@@ -0,0 +1,29 @@
+import * as React from "react";
+
+import RuleHintDropdown from './ruleHintDropdown';
+
+const RuleHintPicker = ({
+  hint,
+  onHintChange,
+}) => {
+
+  const hintId = (hint && hint.id) || null
```
brendanshean: What about using the safe navigation (elvis) operator for these assignments:
(e.g. `const hintId = hint?.id`)

services/QuillLMS/client/app/bundles/Staff/helpers/evidence/ruleHelpers.tsx
```diff
@@ -320,7 +320,7 @@ export const buildRule = ({
     rule_type: ruleType.value,
     suborder: suborder ? suborder : order,
     universal: universal,
-    hint_attributes: ruleHint,
+    hint_id: (ruleHint && ruleHint.id) || null,
```
brendanshean: Consider using safe navigation operator.

services/QuillLMS/client/app/bundles/Staff/utils/evidence/hintAPIs.ts
```diff
@@ -0,0 +1,70 @@
+import { HintInterface } from '../../interfaces/evidenceInterfaces';
+import { handleApiError, apiFetch, handleRequestErrors, requestFailed } from '../../helpers/evidence/routingHelpers';
+
+export const fetchHints = async () => {
+  const response = await apiFetch(`hints`);
+  let hints = await response.json();
+  if(hints && hints.hints) {
```
brendanshean: Consider `hints?.hints`

services/QuillLMS/engines/evidence/spec/controllers/evidence/hints_controller_spec.rb
```diff
@@ -0,0 +1,115 @@
...
+  RSpec.describe(HintsController, :type => :controller) do
+    before { @routes = Engine.routes }
+
+    context 'should index' do
+
+      it 'should return successfully with no records' do
+        get(:index)
+        parsed_response = JSON.parse(response.body)
+        expect(response.status).to eq(200)
+        expect(parsed_response.class).to(eq(Array))
```
brendanshean: You could collapse 15-16 into `expect(parsed_reponse).to match_array([])`

services/QuillLMS/engines/evidence/spec/controllers/evidence/rules_controller_spec.rb
```diff
@@ -306,6 +306,14 @@ module Evidence
         expect(RegexRule.count).to(eq(1))
       end
 
+      it 'should connect a Hint when a hint_id is passed' do
+        hint = create(:evidence_hint)
+
+        post(:create, :params => ({ :rule => ({ :concept_uid => rule.concept_uid, :note => rule.note, :name => rule.name, :optimal => rule.optimal, :state => rule.state, :suborder => rule.suborder, :rule_type => rule.rule_type, :universal => rule.universal, :hint_id => hint.id }) }))
```
brendanshean: For readability, consider breaking this line up into multiple lines.

brendanshean (approved):
Very nice work on this, The code was very readable both on frontend and backend.  I'm still learning react-query so I'm bookmarking this PR as a reference for checking understanding later.  For that reason, I would recommend at least one more approval given the amount of frontend work here.

services/QuillLMS/client/app/bundles/Staff/components/evidence/EvidenceLanding.tsx
```diff
@@ -14,6 +15,7 @@ const EvidenceLanding = () => (
       <Route component={UniversalRule} path='/universal-rules/:ruleId' />
       <Route component={UniversalRulesIndex} path='/universal-rules' />
       <Route component={Activities} path='/activities' />
+      <Route component={Hints} path='/hints/:hintId?' />
```
emilia-friedberg, line 18: Should that question mark be at the end there?

services/QuillLMS/client/app/bundles/Staff/components/evidence/__tests__/activity.test.tsx
```diff
@@ -12,7 +12,7 @@ describe('Activity component', () => {
     </MemoryRouter>
   );
 
-  it('should render 2 NavLinks', () => {
-    expect(container.find(NavLink).length).toEqual(3);
```
emilia-friedberg, line 16: lol at this existing mismatch

services/QuillLMS/client/app/bundles/Staff/components/evidence/configureRules/ruleHint.tsx
```diff
@@ -69,19 +63,18 @@ const RuleHint = ({
...
       </React.Fragment>
     )
   }
 
   return(
     // eslint-disable-next-line
     <React.Fragment>
-      {hint && !hint._destroy ? renderHint() : <button className="add-feedback-layer quill-button small primary outlined" onClick={handleHintCreation} type="button">Add Hint</button>}
+      {renderHint()}
     </React.Fragment>
```
emilia-friedberg: NBD but this doesn't need to be nested in a Fragment anymore, and I don't think that `eslint-disable` is necessary any longer.

services/QuillLMS/client/app/bundles/Staff/components/evidence/configureRules/ruleHintDropdown.tsx
```diff
@@ -0,0 +1,52 @@
...
+
+  const selectedHintOption = hintOptions.find((hintOption) => hintOption.value === selectedHintId) || hintOptions[0]
+
+  const handleOnHintChange = (selection) => {
+    const selectedHint = hintsData.hints.find((hint) => hint.id === selection.value)
+    onHintChange(selectedHint)
+  }
+
+  return(
+    // eslint-disable-next-line
```
emilia-friedberg: Why is this necessary?

services/QuillLMS/client/app/bundles/Staff/components/evidence/configureRules/ruleHintPicker.tsx
```diff
@@ -0,0 +1,29 @@
...
+  const hintExplanation = (hint && hint.explanation) || null
+  const hintImageLink = (hint && hint.image_link) || null
+  const hintImageAltText = (hint && hint.image_alt_text) || null
```
emilia-friedberg: I think for all of these, the tidiest way to do it would just be:

`{ id, explanation, image_link, image_alt_text } = hint || {}`

You'd have to change the variable names below, but the file is so small I don't think you're at any risk of getting confused about what those attributes refer to.

services/QuillLMS/client/app/bundles/Staff/components/evidence/configureRules/ruleHintPicker.tsx
```diff
@@ -0,0 +1,29 @@
...
+
+  return(
+    // eslint-disable-next-line
```
emilia-friedberg: I don't think you need this here either.

services/QuillLMS/client/app/bundles/Staff/components/evidence/hints.tsx
```diff
@@ -0,0 +1,153 @@
...
+
+    flashSnackbar('Hint deleted.')
+    deleteHint(hint.id).then(handleSaveResponse)
+      .then(() => setHint(newHint))
+  }
+
+  const flashSnackbar = (text: string) => {
+    setSnackbarText(text)
+    setSnackbarVisible(true)
+    setTimeout(() => setSnackbarVisible(false), defaultSnackbarTimeout)
```
emilia-friedberg: Just FYI, in hooks-based React components, you can use the `useSnackbarMonitor` hook to avoid some of this boilerplate.

services/QuillLMS/client/app/bundles/Staff/components/evidence/hints.tsx
```diff
@@ -0,0 +1,153 @@
...
+  const handleSaveHint = () => {
+    if (hint.id == '') {
+      createHint(hint).then(handleSaveResponse)
+    } else {
+      updateHint(hint.id, hint).then(handleSaveResponse)
+    }
+  }
+
+  const handleSaveResponse = (response) => {
+    if (response.errors.length > 0) setErrors(result.errors)
```
emilia-friedberg: You can just do `if (response.errors.length)`

services/QuillLMS/client/app/bundles/Staff/components/evidence/hints.tsx
```diff
@@ -0,0 +1,153 @@
...
+      return setHint(newHint)
+    }
+    setHint(selectedHint)
+    history.push({
+      pathname: `/hints/${selectedHint.id}`
+    })
+  }
+
+  const handleSaveHint = () => {
+    if (hint.id == '') {
```
emilia-friedberg: The repetition of checking against an empty string for logic makes me a little nervous - I might either set that empty string to be a constant (ex: `const NO_VALUE = ''` at the top and then `hint.id === NO_VALUE` throughout the file) or check against `hint.id.length` instead.

emilia-friedberg (approved):
This looks great. I made some clean-up suggestions on the frontend but nothing major - really nice work.

happythenewsad (approved):

services/QuillLMS/client/app/bundles/Staff/components/evidence/EvidenceLanding.tsx
anathomical: So according to this StackOverflow article (https://stackoverflow.com/questions/35604617/react-router-with-optional-path-parameter), React Router doesn't document this particularly well, but the `?` after a param in a route means that the param is optional.  Doing it this way means that `#/hints` and `#/hints/123` both load the same component `Hints`.  There might be another way to do this that's more in line with our existing approaches, but I didn't see anything obvious in my survey of routes.

services/QuillLMS/client/app/bundles/Staff/components/evidence/__tests__/activity.test.tsx
anathomical: I did the same thing.  It was one of those "clearly was 2 at one point" situations.

services/QuillLMS/client/app/bundles/Staff/components/evidence/hints.tsx
```diff
@@ -0,0 +1,153 @@
...
+  const { data: hintsData } = useQuery("hints", fetchHints);
+  const [errors, setErrors] = React.useState<string[]>([])
+  const [hint, setHint] = React.useState<HintInterface>(newHint)
+  const [snackbarText, setSnackbarText] = React.useState<string[]>('')
+  const [snackbarVisible, setSnackbarVisible] = React.useState<Boolean>(false)
+
+  const queryClient = useQueryClient()
+
+  if (hintId && hintsData && hint.id != hintId) {
+    const hintFromUrl = hintsData.hints.find((hint) => hint.id == hintId)
```
eadams17: line 37 should be `!==` and line 38 should be `===`

services/QuillLMS/client/app/bundles/Staff/components/evidence/hints.tsx
```diff
@@ -0,0 +1,153 @@
...
+import RuleHint from "./configureRules/ruleHint";
+import RuleHintDropdown from "./configureRules/ruleHintDropdown";
+
+const Hints = ({ location, match }) => {
+  const { params } = match;
+  const { hintId } = params;
+
+  const history = useHistory();
+
+  const newHint: HintInterface = {
```
eadams17: nbd, but this could be pulled out and exported from the Evidence constants file where we have similar constants for a new activity, new rule, etc: https://github.com/empirical-org/Empirical-Core/blob/develop/services/QuillLMS/client/app/constants/evidence.ts#L184

services/QuillLMS/client/app/bundles/Staff/components/evidence/hints.tsx
```diff
@@ -0,0 +1,153 @@
...
+    )
+  }
+
+  const saveButton = () => {
+    const buttonText = hint.id == '' ? 'Create Hint' : 'Update Hint'
+    return (<button className="quill-button medium primary outlined save-hint" onClick={handleSaveHint} type="button">{buttonText}</button>)
+  }
+
+  const deleteButton = () => {
+    if (hint.id == '') return;
```
eadams17: lines 123 and 128 should be `===`

eadams17 (approved):
looks good! just a few, small comments

services/QuillLMS/client/app/bundles/Staff/components/evidence/configureRules/ruleHintPicker.tsx
anathomical: That seems reasonable.

services/QuillLMS/client/app/bundles/Staff/components/evidence/configureRules/ruleHintPicker.tsx
anathomical: I think you're right that safe navigation is better here, but Emilia made an even better suggestion to replace all four of these lines with an expansion call, which I think is cleaner.

services/QuillLMS/client/app/bundles/Staff/components/evidence/hints.tsx
anathomical: Good call-out.  I like that approach better anyway.

services/QuillLMS/client/app/bundles/Staff/components/evidence/configureRules/ruleHint.tsx
anathomical: @emilia-friedberg I tried removing this wrapper, and started getting errors on render: `Uncaught Error: Objects are not valid as a React child (found: object with keys {}). If you meant to render a collection of children, use an array instead.`.  There may be a better way to get that error to stop happening besides wrapping the return value in a `React.Fragment>` element, but that did work.  Any suggestions for a better solution?

PR #9818, opened 2022-10-11
Labeled Synthetic - Add spelling errors to validation/test data.

dandrabik:
## WHAT
We currently generate spelling errors for `TRAINING` data, but not `TEST` and `VALIDATION` data. This adds passage-specific spelling errors to the `TEST` and `VALIDATION` data.
## WHY

We've been restricting the synthetic data to only the TRAINING set for now, out of concern that if the synthetic data shifts too much from the label, it might have an outsized impact on the smaller `TEST` and `VALIDATION` sets (5-10% of the data each). The spelling synthetic data is a little more controlled (we're basically 100% confident the label is accurate), so we can be more confident validating the model with it.

Also, generally speaking, the student data will have spelling errors in it, so our `TEST` and `VALIDATION` sets should have it to make our model stronger and out scoring more accurate.

## HOW

Make the `run` code a little more generic and specific `TEST_GENERATORS` for a second pass

PR Checklist | Your Answer
------------ | -------------
Have you added and/or updated tests? | 'YES'

### DIFF
--- a/services/QuillLMS/engines/evidence/lib/evidence/evidence/synthetic/labeled_data_generator.rb
+++ b/services/QuillLMS/engines/evidence/lib/evidence/evidence/synthetic/labeled_data_generator.rb
@@ -54,18 +54,22 @@ module Evidence
       end
 
       def run
-        generators.each do |type, generator|
-          results_hash = generator.run(results_to_train.map(&:text), languages: languages, passage: passage)
-
-          results_to_train.each do |result|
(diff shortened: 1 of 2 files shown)


anathomical (approved):
Seems reasonable to me.  I also like the generalization approach making it easy to add other generators for TEST and VALIDATION.

happythenewsad (approved):

PR #9820, opened 2022-10-11
lms/clear-cache-on-activity-session-teacher-fix

anathomical:
## WHAT
Manually touch the ActivitySessions that we update to clear cache
## WHY
The `update_all` call we use in this code bypasses normal ActiveRecord callbacks so the `touch` call that should bubble up to Classroom to clear report caching doesn't happen and it's not clear that the teacher fix works (even though it does)
## HOW
Add a `.each(&:touch)` after the `.update_all` call

### Notion Card Links
https://www.notion.so/quill/Can-t-recover-student-data-with-the-restore-activity-sessions-fix-5fb30953412b48e39145cf41c17799b9

PR Checklist | Your Answer
------------ | -------------

### DIFF
--- a/services/QuillLMS/spec/controllers/teacher_fix_controller_spec.rb
+++ b/services/QuillLMS/spec/controllers/teacher_fix_controller_spec.rb
@@ -161,6 +161,11 @@ describe TeacherFixController do
           post :recover_activity_sessions, params: { email: user.email, unit_name: "some name" }
           expect(classroom_unit.reload.assigned_student_ids).to include(another_user.id, other_student.id)
         end
+
+        it 'should touch ActivitySessions to bubble up touch for cache invalidation purposes' do
+          expect_any_instance_of(ClassroomUnit).to receive(:touch) 
+          post :recover_activity_sessions, params: { email: user.email, unit_name: "some name" }
+        end
       end
 
       context 'when unit does not exist' do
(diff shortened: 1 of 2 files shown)


emilia-friedberg (approved):

services/QuillLMS/spec/controllers/teacher_fix_controller_spec.rb
```diff
@@ -161,6 +161,11 @@
           post :recover_activity_sessions, params: { email: user.email, unit_name: "some name" }
           expect(classroom_unit.reload.assigned_student_ids).to include(another_user.id, other_student.id)
         end
+
+        it 'should touch ActivitySessions to bubble up touch for cache invalidation purposes' do
+          expect_any_instance_of(ClassroomUnit).to receive(:touch)
```
emilia-friedberg, line 167: Should this be `ActivitySession` instead of `ClassroomUnit`?

emilia-friedberg (approved):

cissyxyu (approved):

dandrabik (approved):

services/QuillLMS/spec/controllers/teacher_fix_controller_spec.rb
anathomical: This is intentional.  The reason is that we cache based on the `touch` value on `Classroom` and `ClassroomUnit`, and `ActivitySession` is configured to "bubble up" touch through those models.  Basically, what we're testing for here, is that it's bubbling up as intended.  It may make sense to add a comment here, though, so I'll do that.

PR #9823, opened 2022-10-12
Staggered Release Endpoints Part 1

brendanshean:
## WHAT
In preparation for staggered release, update the endpoints:
1)  `teachers/progress_reports/previously_assigned_recommendations` 
2) `teachers/progress_reports/assign_selected_packs`

## WHY
These endpoints track changes from staggered to immediate release.

## HOW
1) Add releaseMethod to the payload of `public_progress_reports#get_previously_assigned_recommendations_by_classroom`.  This will ensure that any previously set release method on a set of recommendations will be assigned to an update of those same recommendations.

2) Break up controller action `assign_selected_packs`  into two separate routes and controller actions:
* assign_independent_practice_activity_packs
* assign_whole_class_instruction_activity_packs (NB staggered release does not apply to whole class instruction)

When a new set of recommended activity packs are assigned with release_method: staggered release, create the corresponding activity_pack_sequence and activity_pack_sequence_activity_pack objects if they don't already exist

When a new set of recommended activity packs are assigned with any other release method: destroy any corresponding activity_pack_sequence and activity_pack_sequence_activity_pack objects.

There's also some semi-related refactoring with unit_activities_saver and classroom_unit_saver.rb code. @dandrabik pointed out that the bulk creation doesn't give any performance gain so I've removed those calls for better readability.

### Screenshots

### Notion Card Links

PR Checklist | Your Answer
------------ | -------------


services/QuillLMS/spec/workers/fast_assign_worker_spec.rb
```diff
@@ -5,7 +5,6 @@
 
 describe FastAssignWorker, type: :worker do
   include_context "Unit Assignments Variables"
-  let!(:unit) {create(:unit, name: unit_template1.name, user_id: teacher.id)}
```
brendanshean, line 8: This `let!` assignment was in the wrong place.  I moved it inside the updates context.  Otherwise, the creates context doesn't actually create a unit (i.e. call Units::Creator)

services/QuillLMS/app/models/concerns/public_progress_reports.rb
```diff
@@ -319,29 +320,49 @@ def get_previously_assigned_recommendations_by_classroom(classroom_id, activity_
...
+        .where(unit_template_id: recommended_lesson_activity_ids, user_id: associated_teacher_ids)
+        .joins(:classroom_units)
+        .where(classroom_units: { classroom_id: classroom_id })
+        .pluck(:unit_template_id)
+
     {
       previouslyAssignedIndependentRecommendations: assigned_recommendations,
-      previouslyAssignedLessonsRecommendations: assigned_lesson_ids
+      previouslyAssignedLessonsRecommendations: assigned_lesson_ids,
+      releaseMethod: release_method
```
emilia-friedberg, line 365: I might be missing something, and it doesn't look like it's documented in Notion (unfortunately) but I think we had talked in a huddle and on Slack (last reference [here](https://quill.slack.com/archives/D01LD5AUQH2/p1661359033026189)) about having `releaseMethod` send up either `immediate` or `staggered` if the recommendations had already been assigned, or `null` if it hadn't. I think the way you have it here, it will always either send up `staggered` or `null`. I can update the frontend to make it work with this setup if you'd prefer, but I wanted to call that out.

emilia-friedberg (approved):
This was really easy to follow and represents a major improvement on the underlying code as well as the new features. Thank you for doing that. I did have one question about a place where I think the backend is going to be sending up something other than what we'd talked about (in the `previouslyAssignedRecommendations` call) - I can change the frontend to match but let me know how you'd like to proceed there.

services/QuillLMS/app/models/concerns/public_progress_reports.rb
brendanshean: My apologies, I forgot to update you on this.  Basically, if `immediate` or `nil` comes from the frontend, the backend will delete any previous `activity_pack_sequence` and `activity_pack_sequence_activity_packs`.

I decided that keeping activity_pack_sequence objects in the db for `immediate_release` would require backfilling previous recommendations without really adding any value to the system.  I could be wrong, but I think most of the `immediate` state can be inferred.

services/QuillLMS/app/services/activity_pack_sequence_getter.rb
```diff
@@ -0,0 +1,38 @@
...
+  attr_reader :classroom_id, :diagnostic_activity_id, :release_method
+
+  def initialize(classroom_id, diagnostic_activity_id, release_method)
+    @classroom_id = classroom_id
+    @diagnostic_activity_id = diagnostic_activity_id
+    @release_method = release_method
+  end
+
+  def run
+    destroy_existing_activity_pack_sequences
```
dandrabik: I found it surprising to see a `destroy_all` as the first call in `ActivityPackSequenceGetter`, kinda like seeing a write in GET request. Consider a different name (or making the `destroy_all` separate).

services/QuillLMS/app/controllers/teachers/progress_reports/diagnostic_reports_controller.rb
```diff
@@ -192,15 +201,39 @@ def redirect_to_report_for_most_recent_activity_session_associated_with_activity
   end
   # rubocop:enable Metrics/CyclomaticComplexity
 
-  def assign_selected_packs
-    if params[:selections]
-      params[:selections].map { |s| s['id']}.each do |ut_id|
-        ut = UnitTemplate.find(ut_id)
-        Unit.unscoped.find_or_create_by(unit_template_id: ut.id, name: ut.name, user_id: current_user.id)
-      end
+  def assign_independent_practice_activity_packs
```
dandrabik: Are we concerned about these two types of assignment (independent vs whole class) drifting from each other with regards to params, what is saved, run, etc.? I see in the frontend code, that these requests are sent independently (so this wouldn't require a refactor there).


I'm not against this approach and I don't see this as a blocker, just wondering aloud about the long-term ramifications of this (e.g. any assignment code must hit multiple independent endpoints to assign).

services/QuillLMS/app/services/independent_practice_activity_packs_assigner.rb
```diff
@@ -0,0 +1,80 @@
...
+        student_ids: selection[:classrooms][0][:student_ids].compact,
+        unit_template_id: selection[:id]
+      )
+    end
+  end
+
+  private def find_or_create_activity_packs
+    selections
+      .map { |selection| UnitTemplate.find(selection['id']) }
+      .each { |ut| Unit.unscoped.find_or_create_by!(unit_template: ut, name: ut.name, user: user) }
```
dandrabik: Bang method in a loop. What do we want to happen here if the 3rd of 20 `units` fails. If they all should fail, we likely need a transaction somewhere.

dandrabik (approved):
Some comments that you should consider, but overall, this looks good to me (all of the comments are take or leave).

Thanks for pushing this forward!

services/QuillLMS/app/models/concerns/public_progress_reports.rb
```diff
@@ -319,29 +320,49 @@ def get_previously_assigned_recommendations_by_classroom(classroom_id, activity_
     classroom = Classroom.find(classroom_id)
     teacher_id = classroom.owner.id
     diagnostic = Activity.find(activity_id)
-    assigned_recommendations = RecommendationsQuery.new(diagnostic.id).activity_recommendations.map do |rec|
+
+    release_method = ActivityPackSequence.find_by(classroom: classroom, diagnostic_activity: diagnostic)&.release_method
+
+    assigned_recommendations = RecommendationsQuery.new(diagnostic.id).activity_recommendations.map do |recommendation|
+      unit_template_id = recommendation[:activityPackId]
```
anathomical, line 327: I very much appreciate the explicit equating of these two ways of referring to the same item.

services/QuillLMS/app/workers/assign_recommendations_worker.rb
```diff
@@ -4,53 +4,68 @@ class AssignRecommendationsWorker
...
-    options = options.with_indifferent_access
-    unit_template_id = options["unit_template_id"]
-    classroom_id = options["classroom_id"]
-    student_ids = options["student_ids"]
-    last = options["last"]
-    lesson = options["lesson"]
-    assign_on_join = options["assign_on_join"] || false
-    assigning_all_recommended_packs = options["assigning_all_recommended_packs"]
+  # rubocop:disable Metrics/ParameterLists
+  def perform(
```
anathomical, line 8: I like moving to named params with default values over the weird `options` shuffling of the original.

services/QuillLMS/app/workers/assign_recommendations_worker.rb
```diff
@@ -4,53 +4,68 @@ class AssignRecommendationsWorker
...
     classroom_data = {
       id: classroom_id,
       student_ids: student_ids,
       assign_on_join: assign_on_join
     }
-    unit ||= nil
+
     assign_unit_to_one_class(unit, classroom_id, classroom_data, unit_template_id, teacher.id)
+
+    unit = find_unit(find_units(unit_template_id, teacher.id)) if unit.nil?
```
anathomical, line 38: Is there a reason to do this here instead of line 23?  Looks like this is identical to the value set on 26, so we might as well just set this exact value earlier in the workflow rather than having the possibility of setting it in the `if` block on line 21 or set here to the same value if it's not set in the `if` block.

services/QuillLMS/spec/support/shared/unit_assignments_variables.rb
```diff
@@ -37,12 +37,18 @@
   end
 
   def unit_templates_have_a_corresponding_unit?(unit_template_ids)
-    names_from_templates = UnitTemplate.where(id: unit_template_ids).pluck(:name)
-    (Unit.all.map(&:name).flatten & names_from_templates).length == names_from_templates.length
+    UnitTemplate
+      .where(id: unit_template_ids)
+      .pluck(:name)
+      .difference(Unit.pluck(:name))
+      .none?
```
anathomical, line 44: This is so much more readable than the old forms of these functions.  Nicely done.

anathomical (approved):
Looks good to me.  One note about a possible duplicate line, but the code should work as-written.

services/QuillLMS/app/controllers/teachers/progress_reports/diagnostic_reports_controller.rb
brendanshean: I think based on how much logic is different between these two endpoints already these should be separate paths, but I think the general point is worth discussing with the team.

services/QuillLMS/app/services/activity_pack_sequence_getter.rb
brendanshean: I would agree.  I'm not happy with this service object name either.  I meant to solicit feedback from group about this for names/patterns.

services/QuillLMS/app/workers/assign_recommendations_worker.rb
brendanshean: Before explaining this logic, I'd like to say I didn't fully refactor this code as much I wanted, so there's more refactoring opportunity here, specifically the clunky logic flow of `find_unit` and `find_units` that you are inquiring about.

Let's say were at line 38.  There are two cases

Case 1:  `unit != nil`.  
Then line 36 doesn't matter doesn't get called and we move along.

Case 2:  `unit == nil` 
Then Units::Creator will be called on line 55 from the `assign_unit_to_one_class` on line 36.  This creates a new unit that will be assigned in line 36.  We can't make this assignment earlier since the unit doesn't exist yet and we need to pass it to `save_activity_pack_sequence_activity_pack` immediately following.

services/QuillLMS/app/services/independent_practice_activity_packs_assigner.rb
brendanshean: Good call!  Updated.

services/QuillLMS/app/services/activity_pack_sequence_getter.rb
brendanshean: On second thought, I've [folded](https://github.com/empirical-org/Empirical-Core/pull/9823/commits/99c5f1df852ae170352340b2a1cd6e7a67f0c71e) the service back into the caller and separated out the destroy call from getter.

services/QuillLMS/app/workers/assign_recommendations_worker.rb
brendanshean: FYI, @anathomical I just realized that sidekiq doesn't [support](https://github.com/mperham/sidekiq/issues/2372) keyword arguments.

anathomical: Oh, yeah.  I should have remembered that because I got into trouble trying to do something similar a while ago.  Whoops!

PR #9827, opened 2022-10-12
lms/restrict-vitally-district-sync

anathomical:
## WHAT
Do not include districts with no paid schools for sync to Vitally
## WHY
We only want to manage districts ("Organizations") in Vitally if we have some sort of paid relationship with them because Vitally charges us for every Organization we manage with them.  So we should stop sending them data about districts we don't have paid relationships with.
## HOW
Just add an extra `select` to the code that retrieves the list of districts to sync to filter out any districts without paid schools in them (which is the vast majority).

### Notion Card Links
https://www.notion.so/quill/Overview-Only-Send-Vitally-Districts-for-Schools-in-Districts-that-Contain-Paying-Schools-10f355b54d494679b9e1510192bd34e5

PR Checklist | Your Answer
------------ | -------------

### DIFF
--- a/services/QuillLMS/app/workers/sync_vitally_worker.rb
+++ b/services/QuillLMS/app/workers/sync_vitally_worker.rb
@@ -47,6 +47,6 @@ class SyncVitallyWorker
   end
 
   def districts_to_sync
-    schools_to_sync.map(&:district).compact.uniq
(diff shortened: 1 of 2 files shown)


dandrabik (approved):
LGTM, thanks for fixing this up.

PR #9838, opened 2022-10-14
OpenAI Edit API

dandrabik:
## WHAT
Add an interface to use the OpenAI [edit](https://beta.openai.com/docs/api-reference/edits) API . You submit a piece of text (`input`) and how you want to edit that text (`instruction`) and the API returns the edited text.
## WHY
I'll be using this API for some of the synthetic data experimentation, e.g. "paraphrase this sentence". Even thought it's not called by other code, this felt like an isolated chunk of code that I can test and release.
## HOW
Refactor out two concerns from `OpenAI::Completion` and follow a similar pattern to that file.

PR Checklist | Your Answer
------------ | -------------
Have you deployed to Staging? | Yes, deploying now.


services/QuillLMS/engines/evidence/lib/evidence/evidence/open_ai/concerns/api.rb
```diff
@@ -0,0 +1,46 @@
...
+
+      included do
+        include HTTParty
+        base_uri 'https://api.openai.com/v1'
+
+        attr_accessor :response
+      end
+
+      # classes using this concern require these methods
+      prepended do
```
happythenewsad: Interesting new syntax - I just read the docs. 

However, according to the [docs](https://ruby-doc.org/core-2.7.0/Module.html#method-i-prepend), prepended will only be called when the module is included via `prepend`, but we aren't using that directive anywhere in this PR or the codebase. Am I missing something?

happythenewsad (approved):

services/QuillLMS/engines/evidence/lib/evidence/evidence/open_ai/concerns/api.rb
dandrabik: Ah, good catch, I was interpreting prepended incorrectly. You are right that it shouldn't be in this block. I've pulled it out and added a generic set of tests for the concern. Thanks for the review!

anathomical (approved):
Seems like a reasonable abstraction to me.

services/QuillLMS/engines/evidence/spec/lib/open_ai/concerns/api_spec.rb
```diff
@@ -0,0 +1,54 @@
...
+    let(:sample_request) {{'key1' => 'value'}}
+    let(:sample_response_body) do
+      {
+        "key"=>"value",
+      }
+    end
+    # include headers in response for proper parsing by HTTParty
+    let(:sample_response) { {body: sample_response_body.to_json, headers: {content_type: 'application/json'}} }
+
+    let(:class_without_defined_methods) do
```
brendanshean, line 14: Cool.  I like this pattern for testing for implementation of the concern's methods.

brendanshean (approved):
Nice work on this.

PR #9839, opened 2022-10-14
Staggered Release Endpoints Part 2

brendanshean:
## WHAT
In preparation for staggered release, update the endpoint `student_profile_data`

## WHY
This update will add PackSequenceItem status:  [`locked`, `unlocked`, `unlocked`] to each result of the student profile query.

## HOW
Add some joins to the student profile query to pull in PackSequences as well as PackSequenceItems. 

1. Pass results into `PackSequenceItemStatusMerger`
2. This in turn iterates over each PackSequence and calls `PackSequenceItemStatusMapper` which constructs a mapping of status to individual PackSequenceItems.
3. Once this map is constructed for each PackSequence, merge in the status to the original `results` array.

### Screenshots

### Notion Card Links

PR Checklist | Your Answer
------------ | -------------


services/QuillLMS/app/services/pack_sequence_item_status_merger.rb
```diff
@@ -0,0 +1,35 @@
...
+    results
+      .select { |result| result[PackSequence::RELEASE_METHOD_KEY] == PackSequence::STAGGERED_RELEASE }
+      .reject { |result| result[PackSequence::ID_KEY].nil? }
+      .group_by { |result| result[PackSequence::ID_KEY] }
+      .values
+  end
+
+  private def map_statuses
+    all_pack_sequence_item_results.each do |pack_sequence_item_results|
+      pack_sequence_item_statuses.merge!(PackSequenceItemStatusMapper.run(pack_sequence_item_results))
```
dandrabik: Not to constantly bring up naming discussions, but I thought this function call was recursive and couldn't figure out how this could work . . . it took me a minute to see the name difference.
`PackSequenceItemStatusMapper`
`PackSequenceItemStatusMerger`

Take or leave, but maybe a different shaped word for one? I don't have a good suggestion, maybe `Combiner` or something.

services/QuillLMS/app/services/pack_sequence_item_status_mapper.rb
```diff
@@ -0,0 +1,40 @@
...
+    map_statuses
+    pack_sequence_item_statuses
+  end
+
+  private def map_statuses
+    previous_packs_unfinished = false
+    previous_id = nil
+    status = UNLOCKED
+
+    ordered_pack_sequence_item_results.each do |pack_sequence_item_result|
```
dandrabik, line 25: So the `locked/unlocked` `status` of an `item` isn't stored in the DB and is live calculated on each call. I mentioned this on the RFC, but this might have performance issues long-term. It will likely be fine for an individual student, but it does make it a more complicated calculation for other uses, e.g. a teacher with 30 students needs to see which activities for their class are `unlocked` at a given moment, or a school wanted to know that, or we wanted to run some analytics queries just in `SQL`. It does have the benefit of not getting out of sync, since it's always calculated live, so there is an upside to this approach (and `state` is always tricky to deal with).

So I'm ok with trying this approach, just with the caveat that if we do hit issues, we may have to pivot to a different approach down the road.

dandrabik (approved):
Overall, looks good to me. I left a comment about how we might need to store state in the DB if we build other features using this data, but I'm also interested to see how this approach without a stored state fairs.

Thanks for pushing this along, very excited for this feature to go out.

anathomical (approved):

services/QuillLMS/app/services/pack_sequence_item_status_merger.rb
brendanshean: I think naming discussions are really important, so I'm happy for the call out.

I would agree they are pretty close.  I think `Combiner` is an improvement.  I initially had this as Mapper and Reducer but didn't want to confuse with the actual MapReduce

services/QuillLMS/app/services/pack_sequence_item_status_mapper.rb
brendanshean: Given the complexity of this individual case, it did make me wonder about moving the value into the database.  There are however, various edge cases that would add even more complexity for such an implementation.  

Another option would be adding a raw SQL solution involving a lateral join or subquery, but I felt at this point that would be overkill.

That being said, I'm see no issue with pivoting if this current approach is not performant.

services/QuillLMS/app/services/pack_sequence_item_status_combiner.rb
```diff
@@ -0,0 +1,35 @@
+# frozen_string_literal: true
+
+class PackSequenceItemStatusCombiner < ApplicationService
+  attr_reader :pack_sequence_item_statuses, :results
+
+  def initialize(results)
+    @results = results
+    @pack_sequence_item_statuses = Hash.new(PackSequenceItem::NO_LOCK)
```
emilia-friedberg, line 8: I've never seen this syntax before! Cool.

services/QuillLMS/app/services/pack_sequence_item_status_mapper.rb
brendanshean: Also, this would most likely require a new database object since the `pack_sequence_item` is not coupled to a user and the locked/unlocked status is a per user designation.

emilia-friedberg (approved):
LGTM and should merge beautifully with frontend (with the exception of the key change that we discussed). We talked about this on Slack but there is a [change we'll have to make](https://www.notion.so/quill/Blocked-Changes-to-the-activity-summary-report-to-represent-scheduled-and-locked-activities-1b6ddaad6539442b87d0483ad7c771f0) to [this query](https://github.com/empirical-org/Empirical-Core/blob/feat/scheduled-release-frontend/services/QuillLMS/app/queries/scorebook/query.rb#L2), but the queries are similar enough that we should be able to reuse some of this code.

PR #9840, opened 2022-10-14
Evidence activity versioning/ create initial version change logs for activities

eadams17:
## WHAT
create initial version change logs for Evidence activities

## WHY
Curriculum would like activities to be created with an initial version 1

## HOW
add logic to create a change log entry after activity creation and add a rake task to create version change logs for all activities that are currently without a version

### Screenshots

### Notion Card Links

PR Checklist | Your Answer
------------ | -------------

### DIFF
--- a/services/QuillLMS/lib/tasks/create_initial_change_log_for_evidence_activities.rake
+++ b/services/QuillLMS/lib/tasks/create_initial_change_log_for_evidence_activities.rake
@@ -3,24 +3,33 @@
 namespace :create_initial_change_log_for_evidence_activities do
   desc 'create change log for all Evidence activities with no initial version'
   task :run => :environment do
+    rake_task_user = User.find_or_create_by(name: 'Rake Robot', role: 'staff')
     Evidence::Activity.all.each do |activity|
-      initial_version = Evidence.change_log_class.where(changed_record_id: activity.id, changed_attribute: 'version', changed_record_type: 'Evidence::Activity').empty?
-      if initial_version
-        changelog_params = {
-          action: Evidence.change_log_class::EVIDENCE_ACTIONS[:create],
-          changed_record_type: 'Evidence::Activity',
-          changed_record_id: activity.id,
-          # this is the user_id associated with our hello@quill.org staff account
-          user_id: 188146,
-          explanation: "Activity Created",
-          changed_attribute: 'version',
-          previous_value: "0",
-          new_value: "1",
-          created_at: activity.created_at,
-          updated_at: activity.created_at
-        }
-        Evidence.change_log_class.create!(changelog_params)
+      changelog_params = {
+        action: Evidence.change_log_class::EVIDENCE_ACTIONS[:create],
+        changed_record_type: 'Evidence::Activity',
+        changed_record_id: activity.id,
+        user_id: rake_task_user.id,
+        explanation: "Activity Created",
+        changed_attribute: 'version',
+        previous_value: "0",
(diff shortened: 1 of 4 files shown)


services/QuillLMS/lib/tasks/create_initial_change_log_for_evidence_activities.rake
```diff
@@ -0,0 +1,24 @@
...
+        changelog_params = {
+          action: Evidence.change_log_class::EVIDENCE_ACTIONS[:create],
+          changed_record_type: 'Evidence::Activity',
+          changed_record_id: activity.id,
+          explanation: "Activity Created",
+          changed_attribute: 'version',
+          previous_value: "0",
+          new_value: "1",
+          created_at: activity.created_at,
+          updated_at: activity.created_at
```
anathomical: Feels like we should try to pass some sort of `user_id` in on this.  I vaguely remember that we had a regular user_id value for a "this was a script" changes in ChangeLog, but I could be mis-remembering.  Either way, I think we want to record something in that column if we can.

anathomical (approved):
Could be run as-is, but I think it'd be better to supply some sort of `user_id` for the rake task.

services/QuillLMS/lib/tasks/create_initial_change_log_for_evidence_activities.rake
brendanshean: maybe something like:  `User.find_or_create_by(name: 'Rake Robot', role: 'staff')`?

brendanshean (approved):

services/QuillLMS/lib/tasks/create_initial_change_log_for_evidence_activities.rake
eadams17: @anathomical @brendanshean looks like we have a "Quill Staff" user with the hello@quill.org email, so maybe that account makes sense?

anathomical: @eadams17 @brendanshean I think I'd mildly prefer having some specific account that isn't used for anything else, but I think `hello@quill.org` would be fine if there's no easy way to do that.

eadams17: @anathomical @brendanshean I searched the codebase and I'm not seeing any kind of reference to a user_id value for that use case, so I suppose I could create a new user like Brendan suggested?

eadams17: @anathomical @brendanshean so I think doing a `User.find_or_create_by` in this rake task won't actually work because it won't save due to the validations on the User model (password, email, etc).

I did see we also have an account for devtools@quill.org that seems like it makes sense to use (I don't think it's used in any other capacity)? 

I could also register a new email account and manually create a "Rake Robot" account associated with that on both staging and prod, then do a check for `User.find_by(name: "Rake Robot")`

what would y'all prefer ?

brendanshean: @eadams17 I would agree with Thomas' preference for having an account that just does this, so I would pick creating a new user on both staging and production.

anathomical: @eadams17 Just using whatever account has `devtools@quill.org` as an email seems reasonable to me.

PR #9845, opened 2022-10-17
Update Synthetic Model, Stop Tokens, and add rake task

dandrabik:
## WHAT
The major change here is changing the default OpenAI model from `babbage` to `curie`. I've also updated the stop words (when to stop generating), and a rake task to make local iterative testing earlier.
## WHY
During some of my experimentation in a different branch, I found that this more powerful model (`curie`) gives better results overall for many of the texts I'm testing (and doesn't give worse results). So I broke off that change to push out while I continue to experiment.

Also, according to the docs, Curie is better at `text summarization`, which is closest to our use case.
## HOW
Change the default model
Add a rake task to generate local files which can focus on a particular activity, conjunction(s), and run number (to make comparing two runs easier)
### Screenshots

Screen Shot 2022-10-17 at 12 47 51 PM

### Notion Card Links

PR Checklist | Your Answer
------------ | -------------
Have you added and/or updated tests? | No, tiny change.

### DIFF
--- a/services/QuillLMS/engines/evidence/lib/evidence/evidence/open_ai/completion.rb
+++ b/services/QuillLMS/engines/evidence/lib/evidence/evidence/open_ai/completion.rb
@@ -22,7 +22,7 @@ module Evidence
 
       attr_accessor :response, :prompt, :temperature, :count, :model_key, :options_hash
 
-      def initialize(prompt:, temperature: 0.5, count: 1, model_key: :babbage, options_hash: {})
+      def initialize(prompt:, temperature: 0.5, count: 1, model_key: :curie, options_hash: {})
         @prompt = prompt
         @temperature = temperature
         @count = count
--- a/services/QuillLMS/engines/evidence/lib/evidence/evidence/open_ai/completion.rb
+++ b/services/QuillLMS/engines/evidence/lib/evidence/evidence/open_ai/completion.rb
@@ -17,7 +17,7 @@ module Evidence
         davinci: 'text-davinci-002'
       }
       BLANK = ''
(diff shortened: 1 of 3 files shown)


services/QuillLMS/engines/evidence/lib/evidence/evidence/open_ai/completion.rb
```diff
@@ -17,12 +17,12 @@ class Completion < Evidence::ApplicationService
         davinci: 'text-davinci-002'
       }
       BLANK = ''
-      STOP_TOKENS = [". ", ", "]
+      STOP_TOKENS = [". ", "; ", "? ", "! "] # max of 4 stop tokens
       MAX_COUNT = 128 # API has an undocument max of 128 for 'n'
 
       attr_accessor :response, :prompt, :temperature, :count, :model_key, :options_hash
```
dandrabik, line 24: This line and the additions to stop tokens are the only ones that have an effect on current production code.

happythenewsad (approved):
Nit: It may be helpful for future users to have the rake task named something more descriptive. "synthetic" could refer to New Relic synthetics, for example.

anathomical (approved):

dandrabik:
> Nit: It may be helpful for future users to have the rake task named something more descriptive. "synthetic" could refer to New Relic synthetics, for example.

I'll keep that in mind. I use `synthetic` as a namespace in multiple places in `evidence` and it is an industry term within the ML world, so I think the context is clear enough here and would be a larger change (but maybe something like `synthetic_data` would be clearer). New Relic "synthetics", in my opinion, are somewhat poorly named, but unfortunately we can't change that.

PR #9847, opened 2022-10-17
Fix reauthorization bug

brendanshean:
## WHAT
Fix a bug that forces some Google users to reauthorize constantly.

## WHY
The google refresh token flow yields a key `expires_in` that usually returns 3600, which is short for 3600 seconds or 1 hour.  

`AuthCredential` has a datetime field called `expires_at` so when this code is called, it tries to assign an integer to a datetime and ends up with a `nil`. 

When our reauthorization flow was checking to see if a [refresh token expires at](https://github.com/empirical-org/Empirical-Core/blob/develop/services/QuillLMS/app/models/auth_credential.rb#L60), it was getting nil which was in turn triggering a loop of re-authorization for the user. 

This bug has been in the code for over [4 years](https://github.com/empirical-org/Empirical-Core/commit/78851a24ac862c25837ca4830f0d3ec68a840093), but I suspect it only manifest once I added code that forced users to reauthorize upon expiration.  Before that, users were kind of in a weird state if they didn't login at least once every 6 months.

## HOW
The fix is to convert the integer to a datetime for proper storage in expires_at.
```        
attributes[:expires_at] = data['expires_in'].seconds.from_now if data['expires_in'].is_a?(Integer)
```

### Screenshots

### Notion Card Links
https://www.notion.so/quill/Google-synced-users-asked-to-reauthorize-account-every-time-they-log-in-c0420e9ec6cb4dcc9ef3067a6d380fbc

PR Checklist | Your Answer
------------ | -------------


services/QuillLMS/app/services/google_integration/refresh_access_token.rb
```diff
@@ -71,7 +71,7 @@ def refresh
     private def parse_attributes(data)
       {}.tap do |attributes|
         attributes[:access_token] = data['access_token'] if data['access_token'].present?
-        attributes[:expires_at] = data['expires_in'] if data['expires_in'].present?
+        attributes[:expires_at] = data['expires_in'].seconds.from_now if data['expires_in'].is_a?(Integer)
```
anathomical, line 74: Is there a non-Integer value we should handle here?  (That is, could this value be `present?` but not an Integer?, and do we care if that's the case?)

anathomical (approved):
I can't believe you managed to track this down.  Nice job.

services/QuillLMS/app/services/google_integration/refresh_access_token.rb
brendanshean: Good question.  I had to check, but [defined](https://www.oauth.com/oauth2-servers/making-authenticated-requests/refreshing-an-access-token/) to be the number of seconds.

brendanshean: Screen Shot 2022-10-17 at 3 47 08 PM

anathomical: In that case, do we even need to use `is_a?`?  I don't have a strong preference, to be honest, I just noticed we were creating a specific condition and not handling other cases.  There's probably not practical difference for us between `is_a?` and `present?` here.

brendanshean: The [spec](https://github.com/empirical-org/Empirical-Core/pull/9847/files#diff-e95fea07468bd8b1a33ff1c01cee661810f8d9731c2f395b5dc37455e8432b72L97) had a passing in of a string 'foo'.  Putting a `present?` would permit that, but then would trigger an error with the `data['expires_in'].seconds.from_now` call.  Using `is_a?(Integer)` ensures the `.seconds` call is made with the correct type.  I would agree it's redundant for the API, but I think it's more transparent for testing.

PR #9857, opened 2022-10-18
Evidence version control/ UI updates

eadams17:
## WHAT
various UI updates to the version history in the Evidence internal tool

## WHY
these changes were requested by Curriculum

## HOW
change the column widths, add logic to check for session count by version and pass value to frontend

### Screenshots
Screen Shot 2022-10-18 at 5 00 02 PM

### Notion Card Links

PR Checklist | Your Answer
------------ | -------------

### DIFF
--- a/services/QuillLMS/client/app/bundles/Staff/components/evidence/versionHistory/versionHistory.tsx
+++ b/services/QuillLMS/client/app/bundles/Staff/components/evidence/versionHistory/versionHistory.tsx
@@ -1,11 +1,12 @@
 import * as React from "react";
 import { useQuery, useQueryClient, } from 'react-query';
+import moment from 'moment';
 
 import { sort } from '../../../../../modules/sortingMethods.js';
 import SubmissionModal from '../shared/submissionModal';
 import { fetchActivity, fetchActivityVersions, updateActivityVersion } from '../../../utils/evidence/activityAPIs';
 import { renderHeader } from "../../../helpers/evidence/renderHelpers";
-import { Input, ReactTable, Spinner } from '../../../../Shared/index';
+import { Input, ReactTable, Spinner, TextArea } from '../../../../Shared/index';
 import { TITLE } from "../../../../../constants/evidence";
 
 const formatChangeLogRows = (activityVersionData) => {
@@ -13,14 +14,16 @@ const formatChangeLogRows = (activityVersionData) => {
   return activityVersionData.changeLogs.map(row => {
     const {
       note,
-      updated_at,
+      created_at,
+      session_count,
       new_value
     } = row
 
     return {
-      updatedAt: updated_at,
-      note,
-      version: new_value
+      createdAt: `${moment(created_at).utcOffset('-0500').format('YYYY/MM/DD h:mm a')} ET`,
+      version: new_value,
+      sessionCount: session_count,
+      notes: note,
     }
   })
 }
@@ -82,7 +85,7 @@ const VersionHistory = ({ history, match }) => {
   }
 
   const activityVersionDisplayValue = ({version}): string => {
-    if (version === 0) return 'Initial Version'
+    if (version === 1) return 'Initial Version'
     return version
   }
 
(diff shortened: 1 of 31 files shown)


services/QuillLMS/engines/evidence/lib/evidence/evidence/change_log.rb
```diff
@@ -63,12 +63,25 @@ def activity_versions
       Evidence.change_log_class.where(changed_record_id: id, changed_attribute: 'version', changed_record_type: 'Evidence::Activity').map do |row|
         {
           note: row.explanation,
-          updated_at: row.updated_at,
-          new_value: row.new_value
+          created_at: row.created_at,
+          new_value: row.new_value,
+          sessions: version_sessions(row, id).count
```
brendanshean: For better transparency, I feel like this key should be called `session_count`.  I would also rename `version_sessions` to `session_count` and have that method just return the count itself rather than the relation.

services/QuillLMS/engines/evidence/lib/evidence/evidence/change_log.rb
```diff
@@ -63,12 +63,25 @@ def activity_versions
...
       end
     end
 
+    def version_sessions(change_log, activity_id)
+      start_date = change_log.created_at
+      end_date = Time.current
+
+      if change_log.previous_value != "0"
+        value = change_log.new_value.to_i + 1
+        next_change_log = Evidence.change_log_class.where(changed_record_id: activity_id, changed_attribute: 'version', changed_record_type: 'Evidence::Activity', new_value: value.to_s).first
```
brendanshean: What about going through the changed_record directly:

```
next_change_log = 
  Evidence::Activity
    &.find_by(activity_id)
    &.change_logs
    &.where(changed_attribute: 'version', new_value: value)
    &.first
```

services/QuillLMS/client/app/bundles/Staff/components/evidence/versionHistory/versionHistory.tsx
```diff
@@ -82,33 +85,40 @@ const VersionHistory = ({ history, match }) => {
   }
 
   const activityVersionDisplayValue = ({version}): string => {
-    if (version === 0) return 'Initial Version'
+    if (version === 1) return 'Initial Version'
```
happythenewsad, line 88: Other zero-index assumptions for this code exist:
1. https://github.com/empirical-org/Empirical-Core/blob/develop/services/QuillLMS/engines/evidence/app/controllers/evidence/feedback_controller.rb#L30
2. In the column migration. Create a new data migration that changes the default value to 1. 
3. Create a backfill script that moves all 0-valued records to 1-valued.

Alternatively, since this is rather annoying, you could push back on the spec. It's not clear to me that moving from 0-index to 1-index adds a lot of value in this case.

happythenewsad (changes requested):
Looks pretty good, but implicit dependencies need to be addressed before merging.

brendanshean (commented):
I left a couple of backend comments.

services/QuillLMS/engines/evidence/lib/evidence/evidence/change_log.rb
```diff
@@ -63,12 +63,30 @@ def activity_versions
...
+          session_count: session_count(row, id)
         }
       end
     end
 
+    def session_count(change_log, activity_id)
+      start_date = change_log.created_at
+      end_date = Time.current
+
+      if change_log.previous_value != "0"
```
brendanshean: Should this be "1" now that 1 is the default?  Also, given that 1 is magic number in many pieces of this commit, I would extract it out as a constant:  DEFAULT_VERSION = 1

Then you have: `change_log.previous_values != DEFAULT_VERSION.to_s`

brendanshean (approved):
I'd highly recommend extracting out that default_version as a constant.  Then I think this should be good to go.  Nice work on this.

services/QuillLMS/engines/evidence/lib/evidence/evidence/change_log.rb
eadams17: so this should still be 0-- we initially did not have any version `ChangeLog` entries for when an Evidence activity was created. Now that we are starting at version one, version ChangeLog entries were backfilled for activity creation with `previous_value` being 0 and `new_value` being 1, so an activity that's on the first version won't step into this if statement.

good call on the default_version constant, I'll add that

happythenewsad (approved):
LGTM

PR #9864, opened 2022-10-19
Synthetic Seed Data filter out common regex.

dandrabik:
## WHAT
We want to exclude data that matches common regexes from our seed data.

## WHY
Evidence has multiple feedback algorithms/APIs that run before the ML one. One is a regex check that flags `because of` and `so that` (these responses that are always incorrect). Since, in practice, the live model will never see entries of this type, we want to exclude them from the training data.
## HOW
Add another config setup, add the two current regexes. Add tests.

PR Checklist | Your Answer
------------ | -------------

### DIFF
--- a/services/QuillLMS/engines/evidence/lib/evidence/evidence/synthetic/seed_data_generator.rb
+++ b/services/QuillLMS/engines/evidence/lib/evidence/evidence/synthetic/seed_data_generator.rb
@@ -159,7 +159,7 @@ module Evidence
       private def regex_exclude?(text)
         return false if conjunction_exclusions.empty?
 
-        conjunction_exclusions.any?{|regex| regex.match(text) }
+        conjunction_exclusions.any?{|regex| regex.match(text.strip) }
       end
 
       private def conjunction_exclusions
--- a/services/QuillLMS/engines/evidence/lib/evidence/evidence/synthetic/seed_data_generator.rb
+++ b/services/QuillLMS/engines/evidence/lib/evidence/evidence/synthetic/seed_data_generator.rb
@@ -19,19 +19,29 @@ module Evidence
       TEMP_SECTION = 0.4 # give a lower temp (creativity) when it has less info
 
       STEM_KEY = '%<stem>s'
-
+      CONJUNCTIONS = [
+        SO = 'so',
+        BUT = 'but',
+        BECAUSE = 'because'
+      ]
       # Config for Conjunction alternates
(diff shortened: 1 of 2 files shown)


services/QuillLMS/engines/evidence/lib/evidence/evidence/synthetic/seed_data_generator.rb
```diff
@@ -108,6 +118,7 @@ def run
           .map {|s| Result.new(text: noun.nil? ? s.lstrip : [noun, s].join(SPACE), seed: seed)}
           .uniq {|r| r.text }
           .reject {|r| r.text.in?(current_result_texts)}
+          .reject {|r| regex_exclude?(r.text) }
```
anathomical, line 121: Does using `reject` here (and the line above) result in the number of results in `new_results` falling below the targets we're setting in `count`?  Should we be worried about that.

anathomical (approved):
Question about maintaining target `count` values, but looks good to me generally.

happythenewsad (approved):

services/QuillLMS/engines/evidence/lib/evidence/evidence/synthetic/seed_data_generator.rb
dandrabik: Yeah, it does, and that's a genuine concern. Because we are removing duplicates in the current code, we've gotten some sets back that are well below the counts, i.e. the API didn't generate many unique answers.

The tricky part, however, about hitting an API that returns semi-random (but somewhat consistent) results and trying to ensure a specific count is that code might result in an infinite loop hitting a paid API (that would be very bad), e.g. the API endpoint keeps returning the same results, we remove them all because they are duplicates, we are below count, so we hit the API again... (We could limit retries, but that also wouldn't ensure hitting the count)

But it is an issue. For now, I've been focusing making the general text generation better and more diverse using different generation methods and/or increasing the counts to account for some loss, with the aim that it'll be less likely (but not impossible), for the resulting data set to be too small for usage by the team. It's something to keep an eye on.

anathomical: Yeah, that makes sense.  We definitely don't want to get into a position where we're paying huge prices for just one additional non-duplicate response.  It's probably a pretty complex issue to guarantee minimum counts, and deciding to solve that later makes sense to me.

PR #9866, opened 2022-10-20
Staggered Release Endpoints part 3

brendanshean:
## WHAT
Add a new model an database table `UserPackSequenceItem`.

## WHY
We need a new table to model individual user's locked status on a particular item in a pack sequence.  Otherwise, we'll be forced to dynamically recalculate the status in various places, specifically teacher facing pages, which depending on the number of students, could be a bottleneck.

## HOW
`rails g model UserPackSequenceItem`.

### Screenshots

### Notion Card Links

PR Checklist | Your Answer
------------ | -------------

### DIFF
--- a/services/QuillLMS/app/models/user_pack_sequence_item.rb
+++ b/services/QuillLMS/app/models/user_pack_sequence_item.rb
@@ -8,8 +8,8 @@
 #  status                :string           default("locked")
 #  created_at            :datetime         not null
 #  updated_at            :datetime         not null
-#  pack_sequence_item_id :bigint
-#  user_id               :bigint
+#  pack_sequence_item_id :bigint           not null
+#  user_id               :bigint           not null
 #
 # Indexes
 #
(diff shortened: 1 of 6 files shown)


anathomical (approved):

services/QuillLMS/app/models/user_pack_sequence_item.rb
```diff
@@ -0,0 +1,34 @@
...
+class UserPackSequenceItem < ApplicationRecord
+  STATUSES = [
+    LOCKED = 'locked',
+    UNLOCKED = 'unlocked'
+  ]
+
+  belongs_to :user
+  belongs_to :pack_sequence_item
+
+  validates :status, inclusion: { in: STATUSES }
```
dandrabik, line 33: Can `user_id` or `pack_sequence_item_id` ever be null for a valid saved object?

If not, consider `validates presence` for them and making the DB fields non-null.

dandrabik (approved):
Looks good. Added a question/comment about whether there should be `null` validations.

services/QuillLMS/app/models/user_pack_sequence_item.rb
brendanshean: Good point.  I mistakenly assumed `foreign_key: true` prevented nullable references.

PR #9871, opened 2022-10-20
Synthetic Seed Data - Remove Opinions

dandrabik:
## WHAT
Similar to this PR https://github.com/empirical-org/Empirical-Core/pull/9864, we want to remove items that would be flagged by the opinion API from the seed data.
## WHY
Evidence has multiple feedback algorithms/APIs that run before the ML one. One is the opinion check. Since, in practice, the live model will never see entries of this type, we want to exclude them from the training data.
## HOW
Reject responses flagged by the Opinion API.

PR Checklist | Your Answer
------------ | -------------

### DIFF
--- a/services/QuillLMS/engines/evidence/lib/evidence/evidence/synthetic/seed_data_generator.rb
+++ b/services/QuillLMS/engines/evidence/lib/evidence/evidence/synthetic/seed_data_generator.rb
@@ -112,19 +112,18 @@ module Evidence
 
       private def run_prompt(prompt:, count:, seed:, noun: nil, temperature: 1)
         api_results = Evidence::OpenAI::Completion.run(prompt: prompt, count: count, temperature: temperature)
+        current_result_texts = results.map(&:text)
 
-        new_results = parse_completion_api_results(api_results, noun: noun, seed: seed)
+        new_results = parse_completion_api_results(api_results, current_texts: current_result_texts, noun: noun, seed: seed)
 
         @results += new_results
       end
 
-      private def parse_completion_api_results(api_results, noun: nil, seed:)
-        current_result_texts = results.map(&:text)
-
+      private def parse_completion_api_results(api_results, current_texts:, seed:, noun: nil)
         api_results
-          .map {|s| Result.new(text: noun.nil? ? s.lstrip : [noun, s].join(SPACE), seed: seed)}
+          .map {|s| Result.new(text: [noun, s].join(SPACE).strip, seed: seed)}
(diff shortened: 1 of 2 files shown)


services/QuillLMS/engines/evidence/lib/evidence/evidence/synthetic/seed_data_generator.rb
```diff
@@ -119,6 +119,7 @@ def run
           .uniq {|r| r.text }
           .reject {|r| r.text.in?(current_result_texts)}
           .reject {|r| regex_exclude?(r.text) }
+          .reject {|r| opinion_api_flagged?(r.text) }
```
dandrabik, line 128: @happythenewsad This will hit the opinion API a few hundred times in rapid succession. Is there a concern with load here?

happythenewsad: @dandrabik during the JMeter benchmarking earlier this year, opinion-api held up fine under 2,000 rpm. 

I say let's try it. We can always optimize later, by:
- scaling up opinion-api when generating synthetic data (or always)
- adding a batch api to opinion-api, which will cut out the networking overhead, but not the spacy model overhead. Maybe 1 hour dev time.

happythenewsad (approved):

services/QuillLMS/engines/evidence/lib/evidence/evidence/synthetic/seed_data_generator.rb
```diff
@@ -162,6 +163,15 @@ def run
         conjunction_exclusions.any?{|regex| regex.match(text.strip) }
       end
 
+      # Use a struct since Opinion API calls .text on the prompt
+      MockPrompt = Struct.new(:text, keyword_init: true)
```
anathomical: This struct feels a little awkward to me, but I think it makes sense to do it this way until we've integrated all of the various pre-filter APIs we want to run and then see if it makes sense to refactor this only when we've seen what other patterns we might need to support.

anathomical (approved):

services/QuillLMS/engines/evidence/lib/evidence/evidence/synthetic/seed_data_generator.rb
dandrabik: Yeah, that's fair. I'll use `::Evidence::Prompt.new` instead to match the type expected by the method (rather than mocking the type). Thanks for the feedback.

PR #9881, opened 2022-10-21
Evidence version history/ add version history filter

eadams17:
## WHAT
add version history filter to Evidence internal tool views

## WHY
Curriculum wants to be able to sort this data by activity version

## HOW
add new version filter to shared `FilterWidget`, remove turk filter, update logic in various views to handle new version search params

### Screenshots
Screen Shot 2022-10-21 at 3 49 23 PM
Screen Shot 2022-10-21 at 3 49 32 PM
Screen Shot 2022-10-21 at 3 49 40 PM
Screen Shot 2022-10-21 at 3 50 17 PM

### Notion Card Links

PR Checklist | Your Answer
------------ | -------------

### DIFF
--- a/services/QuillLMS/client/app/bundles/Staff/components/evidence/activitySessions/sessionsIndex.tsx
+++ b/services/QuillLMS/client/app/bundles/Staff/components/evidence/activitySessions/sessionsIndex.tsx
@@ -5,10 +5,10 @@ import * as moment from 'moment';
 import { firstBy } from 'thenby';
 
 import FilterWidget from "../shared/filterWidget";
-import { handlePageFilterClick } from "../../../helpers/evidence/miscHelpers";
+import { getVersionOptions, handlePageFilterClick } from "../../../helpers/evidence/miscHelpers";
 import { renderHeader } from "../../../helpers/evidence/renderHelpers";
-import { Error, Spinner, DropdownInput, Input, ReactTable, } from '../../../../Shared/index';
-import { fetchActivity, fetchActivitySessions } from '../../../utils/evidence/activityAPIs';
+import { Error, Spinner, DropdownInput, ReactTable, } from '../../../../Shared/index';
+import { fetchActivity, fetchActivitySessions, fetchActivityVersions } from '../../../utils/evidence/activityAPIs';
 import { DropdownObjectInterface, ActivitySessionInterface, ActivitySessionsInterface, InputEvent } from '../../../interfaces/evidenceInterfaces';
 import { activitySessionIndexResponseHeaders, activitySessionFilterOptions, SESSION_INDEX } from '../../../../../constants/evidence';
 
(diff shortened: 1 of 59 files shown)


services/QuillLMS/client/app/bundles/Staff/components/evidence/activitySessions/sessionsIndex.tsx
```diff
@@ -47,10 +46,29 @@ const SessionsIndex = ({ match }) => {
...
     queryFn: fetchActivitySessions
   });
 
+  const { data: activityVersionData } = useQuery({
+    queryKey: [`change-logs-for-activity-versions-${activityId}`, activityId],
+    queryFn: fetchActivityVersions
+  });
+
+  React.useEffect(() => {
+    if(activityVersionData && activityVersionData.changeLogs && (!versionOption || !versionOptions.length)) {
```
emilia-friedberg, line 59: Might consider just doing an early return here.

services/QuillLMS/client/app/bundles/Staff/components/evidence/rulesAnalysis/rulesAnalysis.tsx
```diff
@@ -194,10 +195,17 @@ const RulesAnalysis: React.FC<RouteComponentProps<ActivityRouteProps>> = ({ hist
...
   }, [ruleFeedbackHistory, selectedRuleType])
 
-  function handleSetTurkSessionID(e: InputEvent){ setTurkSessionID(e.target.value) };
+  function handleVersionSelection(versionOption: DropdownObjectInterface) {
+    const { value } = versionOption;
+    const { start_date, end_date } = value;
+    setVersionOption(versionOption);
+    onStartDateChange(new Date(start_date))
+    onEndDateChange(new Date(end_date))
+    setVersionOption(versionOption);
```
emilia-friedberg, line 204: I think you're calling the same thing twice.

services/QuillLMS/client/app/bundles/Staff/components/evidence/activitySessions/sessionsIndex.tsx
```diff
@@ -47,10 +46,29 @@ const SessionsIndex = ({ match }) => {
...
+      const options = getVersionOptions(activityVersionData);
+      const defaultOption = options[0];
+      const { value } = defaultOption;
+      const { start_date, end_date } = value;
+      !versionOption && setVersionOption(defaultOption);
+      setVersionOptions(options);
+      onStartDateChange(new Date(start_date))
+      onEndDateChange(new Date(end_date))
```
emilia-friedberg: Since you're also calling this in `handleVersionSelection`, it would be cleaner to just add a hook that sets the start and end dates when the `versionOption` changes.

emilia-friedberg (approved):
LGTM overall, with some minor cleanup I think we could do. I had two specific comments in one file, but since the code is duplicated it applies everywhere the same functions are called.

On a more general note, the amount of duplication makes me wonder if there's some refactoring we could do here to DRY it up. I wonder if we could move some of the duplicated logic into the `FilterWidget` itself, or add a hook that could handle some of the effects. Not something I think we have to address in this PR, but maybe worth thinking about for a refactoring card.

services/QuillLMS/client/app/bundles/Staff/components/evidence/rulesAnalysis/rulesAnalysis.tsx
eadams17: good catch!

services/QuillLMS/client/app/bundles/Staff/components/evidence/activitySessions/sessionsIndex.tsx
eadams17: good call, I'll extract that out into a hook

PR #9887, opened 2022-10-25
all teacher fixes with update_all calls will reset cache

emilia-friedberg:
## WHAT
DRY up some logic in the teacher fixes and make sure to call `.touch` on `.update_all` affected records so our redis caches are wiped correctly.

## WHY
Scarlet noted that the `unarchive units` teacher fix had to be followed up with the `restore activity sessions` teacher fix in order to actually get all the data back in the right places, and requested either an update to the fix or better documentation thereof. When I investigated, it looked like `unarchive units` already was restoring associated records, but not calling `.touch` on the activity sessions, so the issue was probably just stemming from the cache not being reloaded. I moved the shared logic into functions on the `TeacherFix` concern module, and ensured that each piece of shared logic calls `.touch` on the records that otherwise don't get it. Note that we aren't actually basing caches on the `units` or `unit_activities` right now, but it seemed more future-proof to go ahead and add that in while I was refactoring the code, and it's a good idea to have correct `updated_at` values regardless.

## HOW
Just pull out repeated code into functions and add new tests.

### Screenshots

### Notion Card Links
https://www.notion.so/quill/Restored-data-does-not-fully-show-on-reports-unarchive-units-fix-430a867ea371460cbfc9002718fad735

PR Checklist | Your Answer
------------ | -------------
Have you deployed to Staging? | Tested locally

### DIFF
--- a/services/QuillLMS/spec/controllers/concerns/teacher_fixes_spec.rb
+++ b/services/QuillLMS/spec/controllers/concerns/teacher_fixes_spec.rb
@@ -394,5 +394,61 @@ describe TeacherFixes do
         end
       end
     end
+
   end
+
+  describe '#recover_classroom_units_and_associated_activity_sessions' do
+    it 'should unarchive the classroom units' do
+      classroom_unit_with_activity_sessions1.update(visible: false)
+      classroom_unit_with_activity_sessions2.update(visible: false)
+      TeacherFixes::recover_classroom_units_and_associated_activity_sessions([classroom_unit_with_activity_sessions1, classroom_unit_with_activity_sessions2])
+      expect(classroom_unit_with_activity_sessions1.reload.visible).to eq(true)
+      expect(classroom_unit_with_activity_sessions2.reload.visible).to eq(true)
+    end
+
+    it 'should unarchive the associated activity sessions' do
+      classroom_units = [classroom_unit_with_activity_sessions1, classroom_unit_with_activity_sessions2]
+      activity_sessions = ActivitySession.where(classroom_unit: classroom_units)
+      activity_sessions.each { |as| as.update(visible: false) }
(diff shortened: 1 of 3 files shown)


services/QuillLMS/spec/controllers/concerns/teacher_fixes_spec.rb
```diff
@@ -394,5 +394,61 @@
...
+      activity_sessions.each { |as| as.update(visible: false) }
+      TeacherFixes::recover_classroom_units_and_associated_activity_sessions(classroom_units)
+      activity_sessions.each { |as| expect(as.reload.visible).to eq(true) }
+    end
+
+    it 'should touch each activity session to bubble up touch for cache invalidation purposes' do
+      first_activity_session = ActivitySession.find_by(classroom_unit: classroom_unit_with_activity_sessions1)
+      stored_updated_at_value = first_activity_session.updated_at
+      TeacherFixes::recover_classroom_units_and_associated_activity_sessions([classroom_unit_with_activity_sessions1])
+      expect(first_activity_session.reload.updated_at).not_to equal(stored_updated_at_value)
```
anathomical, line 421: There's nothing wrong with this formulation, but another way to write this spec is with an `expect... to change` block:
```
expect do
  TeacherFixes::recover_classroom_units_and_associated_activity_sessions([classroom_unit_with_activity_sessions1])
end.to change(ActivitySession.find_by(classroom_unit: classroom_unit_with_activity_sessions1), :updated_at)
```
I don't see a reason to change what you've got, but figured I'd call out the alternate approach for more complex situations in the future.

anathomical (approved):
Nice!  Appreciate the DRY work here.  And the tests!

cissyxyu (approved):

happythenewsad (approved):

PR #9888, opened 2022-10-25
update main request module to use fetch

emilia-friedberg:
## WHAT
Update our centralized `request` module to use `fetch` instead of the deprecated `request` library.

## WHY
`request` has been deprecated. This is the first step of a project to deduplicate the similar files we have around the site, and ultimately remove the `request` package entirely.

## HOW
Just rewrite this file using fetch and do some testing to ensure everything works as expected. One small quirk is that I had to reorder the `format.html` and `format.json` in the `handle_invalid_authenticity_token` function because otherwise `fetch` was getting an actual redirect and failing.

### Screenshots

### Notion Card Links
https://www.notion.so/quill/Replace-all-direct-imports-from-request-library-with-our-request-module-3ec643554cc1461093ae08ac20afb047 (just step one - the diff was already large)

PR Checklist | Your Answer
------------ | -------------


services/QuillLMS/app/controllers/application_controller.rb
```diff
@@ -131,8 +131,8 @@ def non_standard_route_redirect?(route)
     flash[:error] = t('actioncontroller.errors.invalid_authenticity_token')
 
     respond_to do |format|
-      format.html { redirect_back(fallback_location: root_path) }
       format.json { render json: { redirect: URI.parse(request.referer).path }, status: 303 }
+      format.html { redirect_back(fallback_location: root_path) }
```
brendanshean: Can you say a little bit more about why the reordering is necessary?  I didn't think order mattered format blocks.  Maybe related, why is reordering not needed in the method [reset_session_and_redirect_to_sign_in](https://github.com/empirical-org/Empirical-Core/blob/a7eef4060bde74528a0729e6ebffb75d33c1e4a1/services/QuillLMS/app/controllers/application_controller.rb#L172)

emilia-friedberg: We discussed in Slack but just for posterity - turns out `fetch` sends a default `Accept` header of *%* in its requests, which tell Rails it doesn't care what format it is, so it just sends back whatever's first. I've changed it now so I undo the order reset and just add `application/json` as an explicit `Accept` header.

eadams17 (approved):
thanks for knocking this out! 🎉

services/QuillLMS/app/controllers/application_controller.rb
brendanshean: Nice sleuthing on this.  I like the explicit `Accept` header.

For my future self, a couple of links from offline discussion:
https://www.bigbinary.com/blog/order-of-format-matters-in-respond_to-block
https://apidock.com/rails/ActionController/MimeResponds/InstanceMethods/respond_to#500-Ordering-of-format-blocks-is-important

brendanshean (approved):
Great work on this.  I like the idea of breaking this up into pieces.

PR #9889, opened 2022-10-25
Time track max

dandrabik:
## WHAT
For time tracking, add a maximum value per section
## WHY
There still seems to be an issue with sessions having very long times on a given section, e.g. the prompts would have times of 50 seconds, 70 seconds, 3 hours, 40 seconds, etc. We originally added code to replace large outliers (40x the median), but we left out some sections (like `proofreading the passage`) and some times are still very long given this high bar of 40x the median. This adds a second pass to cap section time at `10.minutes`.

We think this will be a more accurate representation of `timespent` on activities and more helpful for teachers.

Note, Partnerships gave their feedback and felt that `10.minutes` max for a question seems reasonable.
## HOW
Do a separate `median_outlier` pass followed by a `max_outlier` pass on time tracking data.
### Screenshots
Screen Shot 2022-10-25 at 12 02 36 PM
Screen Shot 2022-10-25 at 12 02 02 PM
Screen Shot 2022-10-25 at 12 00 40 PM

### Notion Card Links

https://www.notion.so/quill/ActivitySession-LongTimeTrackingError-1dd92449d3d047b08a4708cc86355ddd

PR Checklist | Your Answer
------------ | -------------
Spec Review: Have you reviewed the spec and ensured this meets requirements and/or matches design mockups? | N/A Yes)

### DIFF
--- a/services/QuillLMS/app/lib/time_tracking_cleaner.rb
+++ b/services/QuillLMS/app/lib/time_tracking_cleaner.rb
@@ -44,13 +44,19 @@ class TimeTrackingCleaner < ApplicationService
   end
 
   private def max_outliers
-    time_tracking
-      .except(*median_outliers.keys)
-      .select {|_,v| v > MAX_TIME_SECTION }
+    @max_outliers ||= begin
+      time_tracking
+        .except(*median_outliers.keys)
+        .select {|_,v| v > MAX_TIME_SECTION }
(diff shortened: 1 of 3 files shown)


brendanshean (approved):
This seems like a reasonable update to handling these outliers.  Nice work on this.

PR #9900, opened 2022-10-26
Fix a bug involving importing of Google Classroom Students

brendanshean:
## WHAT
Fix an edge case involving importing a student where the email associated with their google_id has changed.  But this new email already has an existing user in our system.

## WHY
Currently, our system [checks](https://github.com/empirical-org/Empirical-Core/blob/develop/services/QuillLMS/app/services/google_integration/student_email_updater.rb#L15) for the existence of another user with an email associated with a particular google_id.  Unfortunately this guard clause just sidesteps the inevitable validation error.

## HOW
Treat the `google_id` and `email` pairing that the google client sends us as the source of truth.  If new data comes in linking a different email to a google_id, then transfer the google_id to the User with that different email.  `GoogleIdTransferrer` class handles this which unlinks the google_id from one user and then links it to the newer user.

Also, add some ChangeLog to these update actions in case we need to undo/audit.

### Screenshots

### Notion Card Links

PR Checklist | Your Answer
------------ | -------------


services/QuillLMS/app/services/google_integration/google_id_transferrer.rb
```diff
@@ -0,0 +1,56 @@
...
+      @from_user = from_user
+      @from_user_google_id = from_user.google_id
+      @to_user = to_user
+      @to_user_google_id = to_user.google_id
+    end
+
+    def run
+      return if from_user_google_id.nil?
+
+      unlink_from_user
```
dandrabik: Should we consider a `transaction` here? For example, would it be bad if the `unlink` worked and the `link` didn't?

dandrabik (approved):
Looks good to me. One comment about a potential `transaction` that I'll defer to your judgement on.

Thanks for fixing this up!

services/QuillLMS/app/services/google_integration/google_id_transferrer.rb
brendanshean: Good point.  This definitely has the feel of a bundled pair of actions.

PR #9902, opened 2022-10-27
replace invocations of request library with our centralized request module

emilia-friedberg:
## WHAT
Replace everywhere that we directly invoke the `request` library, which has been deprecated, with a call to our centralized request module, which relies on `fetch`. This is the third and final PR of the request module cleanup project.

## WHY
To remove a deprecated library from our codebase, and make our AJAX calls sitewide follow a standard format. Note that we do still have some calls around the site that use `fetch` directly, and this PR was already so big I didn't want to update those too, but this handles the majority of calls.

## HOW
- Do a global search for every instance of `request.`, `request(`, `from 'request'` etc, and then update every instance where we called the library to use our central module
- Remove the library from our `package.json`
- Manually test *every changed method* to ensure that it works as expected, fixing some minor bugs and deleting some unused code along the way

Note that, in the scope of testing this PR, I did uncover some bugs that appear to be present on prod as well. I fixed some small ones but the following will require more investigation (I'll add cards to the support board):

- rematch all doesn't seem to work on connect, the diagnostic, or grammar
- grammar looks for a nonexistent Grammar activity when used as proofreader followup, resulting in a console error (nothing user-facing)
- school roster upload doesn't seem to be working

### Screenshots

### Notion Card Links
https://www.notion.so/quill/Replace-all-direct-imports-from-request-library-with-our-request-module-3ec643554cc1461093ae08ac20afb047

PR Checklist | Your Answer
------------ | -------------


services/QuillLMS/client/app/bundles/Diagnostic/actions/questions.ts
```diff
@@ -222,18 +222,15 @@ function searchResponses(qid) {
...
-      {
-        url: `${process.env.QUILL_CMS}/questions/${qid}/responses/search`,
-        method: 'POST',
-        json: { search: getFormattedSearchData(getState()), },
-      },
-      (err, httpResponse, data) => {
+    requestPost(
+      `${process.env.QUILL_CMS}/questions/${qid}/responses/search`,
+      { search: getFormattedSearchData(getState()), },
+      (data) => {
```
brendanshean: For consistency, instead of `data`, any reason not to use `body`  here?

services/QuillLMS/client/app/bundles/Diagnostic/components/questions/massEditContainer.jsx
```diff
@@ -81,16 +78,16 @@ class MassEditContainer extends React.Component {
   }
 
   updateResponseFeedbackInMassEditArray() {
-    const { dispatch, massEdit, params } = this.props;
+    const { dispatch, massEdit, match } = this.props;
```
brendanshean, line 81: I'm guessing this is one of the bugs you uncovered?

services/QuillLMS/client/app/bundles/Evidence/actions/session.ts
```diff
@@ -90,20 +88,17 @@ export const fetchActiveActivitySession = ({ sessionID, activityUID, callback, }
...
-      if (r.statusCode < 200 || r.statusCode >= 300) {
+    requestGet(
+      activeActivitySessionUrl,
+      (body) => {
+        const { submittedResponses, } = body
+        dispatch({ type: ActionTypes.SET_SUBMITTED_RESPONSES, submittedResponses });
+        if (callback) callback(body)
+      },
+      (body) => {
         dispatch({ type: ActionTypes.SESION_HAS_NO_DATA })
```
brendanshean: Maybe rename `SESION_HAS_NO_DATA` to `SESSION_HAS_NO_DATA`.

services/QuillLMS/client/app/bundles/Evidence/utils/turkAPI.ts
```diff
@@ -15,14 +14,9 @@ export const postTurkSession = async (turkingRoundID: string, activitySessionUID
...
-  request.get(requestObject, (e, r, body) => {
-    if (e) {
-      alert("Failed to validate turking activity session: " + e)
-    }
-    callback(body)
-  })
+  requestGet(
+    `${process.env.DEFAULT_URL}/api/v1/evidence/turking_round_activity_sessions/validate?activity_id=${activityId}&turking_round_id=${turkingRoundId}`,
+    (body) => callback(body),
+    (body) => alert("Failed to validate turking activity session: " + body)
```
brendanshean: does the `callback(body)` need to be called in the failure case too?

services/QuillLMS/client/app/bundles/Grammar/actions/responses.ts
```diff
@@ -20,54 +21,40 @@ export function updateData(questionId: string, responses: Response[]) {
 }
 
 
-function getQuestionLoadedStatusForGroupedResponses(groupedResponses) {
```
brendanshean, line 23: Is this dead code?

services/QuillLMS/client/app/bundles/Grammar/actions/responses.ts
```diff
@@ -200,54 +176,50 @@ function makeIterator(array) {
...
-        resp.concept_results = JSON.parse(resp.concept_results);
-      }
-      for (const cr in resp.concept_results) {
-        if (resp.concept_results.hasOwnProperty(cr)) {
-          const formattedCr = {};
-          formattedCr.conceptUID = cr;
-          formattedCr.correct = resp.concept_results[cr];
-          resp.concept_results[cr] = formattedCr;
+  requestGet(`${process.env.QUILL_CMS}/questions/${questionID}/responses`)
+    .then((body) => {
```
brendanshean: Why the use of `.then` here?

services/QuillLMS/client/app/bundles/Grammar/components/questions/conceptResultList.tsx
```diff
@@ -22,7 +22,7 @@ export default class ConceptResultList extends React.Component<ConceptResultList
 
   handleConceptChange(e: { value: string }) {
     const newConceptResults = Object.assign({}, this.state.conceptResults)
-    // newConceptResults[e.value] ? null : newConceptResults[e.value] = false
+    newConceptResults[e.value] ? null : newConceptResults[e.value] = false
```
brendanshean: What about just an `if` block here since the `else` branch does nothing

services/QuillLMS/client/app/bundles/Evidence/actions/session.ts
```diff
@@ -123,49 +118,44 @@ export const saveActiveActivitySession = ({ completedSteps, timeTracking, studen
...
-    json: true,
-  }
-
-  request.post(requestObject, (e, r, body) => {
-    if (callback) callback()
-  })
+    (body) => {
+      if (callback) callback()
+    },
+    (body) => {
```
brendanshean, line 144: I noticed that other refactoring of requestPost of this form do not include the `(body) => { if(callback) callback() }` for the error case.  Is there a reason it is in this one?

services/QuillLMS/client/app/bundles/Teacher/components/blog_posts/blog_post.jsx
```diff
@@ -21,20 +22,20 @@ export default class BlogPost extends React.Component {
   }
 
   selectRatingEmoji(rating) {
-    request.post({
-      url: `${process.env.DEFAULT_URL}/rate_blog_post`,
-      json: {
+    requestPost(
+      `${process.env.DEFAULT_URL}/rate_blog_post`,
+      {
         authenticity_token: ReactOnRails.authenticityToken(),
```
brendanshean: Is this authenticity key needed?

services/QuillLMS/client/app/bundles/Teacher/containers/EditUnitActivitiesContainer.jsx
```diff
@@ -24,19 +25,19 @@ export default class extends React.Component {
 
   updateActivities = () => {
     const that = this;
-    request.put({
-      url: `${process.env.DEFAULT_URL}/teachers/units/${that.props.match.params.unitId}/update_activities`,
-      json: {
+    requestPut(
+      `${process.env.DEFAULT_URL}/teachers/units/${that.props.match.params.unitId}/update_activities`,
+      {
         authenticity_token: getAuthToken(),
```
brendanshean: I think `authenticity_token` can be removed.

services/QuillLMS/client/app/bundles/Teacher/containers/EditUnitActivitiesContainer.jsx
```diff
@@ -1,9 +1,10 @@
 import React from 'react'
+
 import UnitStage1 from '../components/assignment_flow/create_unit/select_activities_container.jsx'
-import request from 'request'
 import getAuthToken from '../components/modules/get_auth_token';
```
brendanshean: I think the import getAuthToken can be removed

services/QuillLMS/client/app/bundles/Staff/components/UploadRosters.tsx
```diff
@@ -41,19 +41,21 @@ export const UploadRosters = () => {
   }
 
   function submitRosters() {
-    request.post(`${process.env.DEFAULT_URL}/cms/rosters/upload_teachers_and_students`, {
-      json: {
+    requestPost(`${process.env.DEFAULT_URL}/cms/rosters/upload_teachers_and_students`,
+      {
         authenticity_token: getAuthToken(),
```
brendanshean: I think authenticity_token: can be removed

services/QuillLMS/client/app/bundles/Teacher/components/progress_reports/export_csv.jsx
```diff
@@ -1,8 +1,9 @@
 import React from 'react';
-import request from 'request';
+
 import auth_token from '../modules/get_auth_token';
```
brendanshean: I think `import auth_token` can be removed along with the `authenticity_token` key further down

services/QuillLMS/client/app/bundles/Teacher/containers/EditOrCreateSubscription.jsx
```diff
@@ -1,12 +1,13 @@
 import React from 'react';
 import { SingleDatePicker } from 'react-dates'
 import _ from 'lodash';
-import request from 'request';
 import { DataTable } from '../../Shared/index'
 import ItemDropdown from '../components/general_components/dropdown_selectors/item_dropdown.jsx';
 import getAuthToken from '../components/modules/get_auth_token';
```
brendanshean: I think `import getAuthToken` and authenticity_token further down can be removed.

brendanshean: There's also one more of the `import getAuthToken` and corresponding authenticity_token that I think can be removed  in:  https://github.com/empirical-org/Empirical-Core/blob/develop/services/QuillLMS/client/app/bundles/Teacher/containers/TeacherAccount.jsx#L10 but for some reason I'm not showing that in this PR.

brendanshean (approved):
Wow.  What a monumental undertaking for bug week.  I'm really impressed with the focus on getting through the literally hundreds of cases.  Some very minor comments but nothing needing re-review from me.  Thank you so much for cleaning all of these calls up.  The consistency will save everyone brain cycles moving forward.

services/QuillLMS/client/app/bundles/Diagnostic/actions/questions.ts
anathomical: I think in most of these cases it makes sense to use whatever the old code was using for the variable name, even if it's not consistent.  So in this case, the old line 231 specifies `data`, which I think means it makes sense to keep calling it `data`.  Even though other places in the code specify `body` in that spot.

services/QuillLMS/client/app/bundles/Evidence/actions/session.ts
anathomical: It's not wholly inaccurate.  The  third argument (this one) to `requestGet` is the error-handling function to run, so this is most commonly going to be a 404 where we haven't saved a record for that particular session_uid.

services/QuillLMS/client/app/bundles/Evidence/utils/turkAPI.ts
```diff
@@ -15,14 +14,9 @@ export const postTurkSession = async (turkingRoundID: string, activitySessionUID
...
-  }
-  request.get(requestObject, (e, r, body) => {
-    if (e) {
-      alert("Failed to validate turking activity session: " + e)
-    }
-    callback(body)
-  })
+  requestGet(
+    `${process.env.DEFAULT_URL}/api/v1/evidence/turking_round_activity_sessions/validate?activity_id=${activityId}&turking_round_id=${turkingRoundId}`,
+    (body) => callback(body),
```
anathomical, line 19: I was going to suggest that we just replace this with `callback` rather than adding a layer of indirection here, but after thinking about it for a bit I think this is actually more readable even if it's (very slightly) less efficient?  Probably better to keep it in the form you have.

services/QuillLMS/client/app/bundles/Grammar/actions/responses.ts
```diff
@@ -200,54 +176,50 @@ function makeIterator(array) {
...
-      delete resp.concept_results;
-    });
-    callback(bodyToObj);
-  });
+        resp.conceptResults = resp.concept_results;
+        delete resp.concept_results;
+      });
+      callback(bodyToObj);
+    })
+    .catch((error) => {
```
anathomical: As with Brendan's call out above, if we aren't planning to use `then` for these workflows, we probably don't want to use `catch`, either.

services/QuillLMS/client/app/bundles/Diagnostic/actions/questions.ts
brendanshean: Standardizing variable names for all of these endpoints establishes a predictable pattern of code use.  Using different names can add confusion when there shouldn't be any.  As far as the occurrence of `data` in the code block itself, those should be renamed as well.

services/QuillLMS/client/app/bundles/Evidence/actions/session.ts
brendanshean: I've updated my comment for clarification which had to do with the constant name.  The error handler seemed fine to me.

anathomical (approved):
Looks good to me.  This looks like it was a ton of work just making the code changes, and way more work than that testing everything.  Thanks for working through this!

services/QuillLMS/client/app/bundles/Diagnostic/components/questions/massEditContainer.jsx
emilia-friedberg: That's correct!

services/QuillLMS/client/app/bundles/Grammar/actions/responses.ts
emilia-friedberg: Yes!

services/QuillLMS/client/app/bundles/Evidence/actions/session.ts
emilia-friedberg: The simplest explanation is that originally there wasn't, and then in testing this request failed on anonymous activity sessions and the user couldn't progress, so I fixed it. So it was just the only place that I noticed that there needed to be a matching error block too. That should have triggered me to go look for where else this pattern existed - I'll do that now before this gets released in case there's a hidden error somewhere.

PR #9906, opened 2022-10-27
Add rake task to remove duplicate student accounts that have no data

brendanshean:
## WHAT
There's a [bug](https://www.notion.so/quill/GoogleStudentImporterWorker-perform-Validation-failed-Email-That-email-is-taken-Try-another-Use-b02bc2dc7d74433698f914371ded840f) with google student importing when the importer attempts to update a student account where there exist multiple users with the same email, username, etc.   This rake task goes in and removes those duplicate users.

## WHY
These duplicate users are a huge gotcha, breaking assumptions about data integrity.

## HOW
Find email addresses that occur on multiple user accounts
Use the `User.find_by(email: email)` as the canonical account and then destroy other accounts that are lacking activity sessions and student classrooms (based on the student merge [operation](https://github.com/empirical-org/Empirical-Core/blob/develop/services/QuillLMS/app/models/concerns/student.rb#L121)).

FYI:  This won't clean all of the duplicate emails (there are ~70000).
Based on a survey of staging, this script will remove 54111 accounts. 

### Screenshots

### Notion Card Links
https://www.notion.so/quill/GoogleStudentImporterWorker-perform-Validation-failed-Email-That-email-is-taken-Try-another-Use-b02bc2dc7d74433698f914371ded840f

PR Checklist | Your Answer
------------ | -------------

### DIFF
--- a/services/QuillLMS/app/models/user.rb
+++ b/services/QuillLMS/app/models/user.rb
@@ -676,6 +676,15 @@ class User < ApplicationRecord
     end
   end
 
+  def self.duplicate_empty_student_accounts(id, email)
+    return none unless exists?(id: id)
+
+    student
+      .where(email: email)
+      .where.not(id: id)
+      .where.missing(:activity_sessions, :students_classrooms)
+  end
+
   private def validate_flags
     invalid_flags = flags - VALID_FLAGS
 
--- a/services/QuillLMS/app/models/user.rb
+++ b/services/QuillLMS/app/models/user.rb
@@ -676,10 +676,9 @@ class User < ApplicationRecord
     end
   end
 
-  def self.duplicate_empty_student_accounts(id, email)
-    return none unless exists?(id: id)
-
-    student
+  def duplicate_empty_student_accounts
+    User
+      .student
       .where(email: email)
       .where.not(id: id)
       .where.missing(:activity_sessions, :students_classrooms)
--- a/services/QuillLMS/lib/tasks/temporary/remove_student_duplicates_with_same_email.rake
+++ b/services/QuillLMS/lib/tasks/temporary/remove_student_duplicates_with_same_email.rake
@@ -3,27 +3,15 @@
 require_relative '../progress_bar'
 
 namespace :students do
-  desc 'Remove duplicate student users with same email'
   task remove_users_with_same_email: :environment do
-    duplicate_emails =
-      User
-        .student
-        .where
-        .not(email: [nil, ""])
-        .group(:email)
-        .having("count(email) > 1")
-        .pluck(:email)
+
+    duplicate_emails = User.student.where.not(email: [nil, ""]).group(:email).having("count(email) > 1").pluck(:email)
 
     progress_bar = ProgessBar.new(duplicate_emails.size)
 
-    [].tap do |ids_for_deletion|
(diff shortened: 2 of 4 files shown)


services/QuillLMS/app/models/user.rb
```diff
@@ -668,6 +668,14 @@ def set_time_zone
     end
   end
 
+  def duplicate_empty_student_accounts
+    User
+      .student
+      .where(email: email)
+      .where.not(id: id)
+      .where.missing(:activity_sessions, :students_classrooms)
```
emilia-friedberg, line 684: This syntax is new to me and handy!

emilia-friedberg (approved):

services/QuillLMS/app/models/user.rb
brendanshean: Just got added in Rails [6.1](https://github.com/rails/rails/pull/34727)

anathomical: Not exactly an objection, but is this something we want to put on the `User` model when we're aiming to clean up all instances of this in this very PR?  Like, this should always be, effectively, a null operation after the rake task runs?

anathomical (approved):
Minor concern about needing to remove code in the future, but this seems like a good approach.  I especially like the "get the easy to kill stuff now, figure out the more complicated instances later" philosophy.

services/QuillLMS/app/models/user.rb
brendanshean: Good question.  Long term, I think this method should be removed, However, the rake task will clean up some, but not all of the duplicates and I see this method being useful until all duplicates are removed.  For example, consider user1 and user2 with the same email.  [Merging](https://github.com/empirical-org/Empirical-Core/blob/develop/services/QuillLMS/app/models/concerns/student.rb#L121) user2 into user1 means that user1.duplicate_empty_student_accounts returns user2.  In which case I can re-run the rake task.

dandrabik: Oh, fun, I didn't know about the `.missing` method, neat.

Very Oliver Twist.
Screen Shot 2022-10-31 at 10 51 37 AM

services/QuillLMS/lib/tasks/temporary/remove_student_duplicates_with_same_email.rake
```diff
@@ -0,0 +1,17 @@
+# frozen_string_literal: true
+
+require_relative '../progress_bar'
+
+namespace :students do
+  task remove_users_with_same_email: :environment do
+
+    duplicate_emails = User.student.where.not(email: [nil, ""]).group(:email).having("count(email) > 1").pluck(:email)
+
+    progress_bar = ProgessBar.new(duplicate_emails.size)
```
dandrabik, line 10: I like this ProgressBar feature, I have to start using that!

dandrabik (approved):
Looks good to me. Thanks for cleaning up this low-hanging fruit. This will be a huge win!

brendanshean:
Just completed run on staging.  Will run on production tomorrow morning.

`heroku run rake students:remove_users_with_same_email -r brendan  23.03s user 12.73s system 2% cpu 24:30.37 total`

PR #9907, opened 2022-10-27
update main request module to use fetch and deduplicate modules in apps (fixed)

emilia-friedberg:
## WHAT
Fix for two issues surfaced after deploy from [this branch](https://github.com/empirical-org/Empirical-Core/pull/9903): specifically, [new Grammar sessions wouldn't start](https://www.notion.so/quill/Some-students-cannot-open-activities-spinning-wheel-1c8dd0ebcfc24df9bc0584b55add5a65), and we had a high volume of "Unexpected end of JSON input" [errors](https://sentry.io/organizations/quillorg-5s/issues/3701003941/?project=210579&query=is%3Aunresolved&referrer=issue-stream&statsPeriod=24h).

Since this PR is large (it includes the reversion from the initial PR, which involved a lot of file changes though most of them were import paths), it might be easiest to just review the changes since the reversion of the reversion: https://github.com/empirical-org/Empirical-Core/pull/9907/files/703375300fba7adc68d0068d6caa3ec836556dab..88080edd2edbe29cdce92e4e8edf0dbd42f8a357.

## WHY
We don't want any behavior to be different than it was before the original PR went out.

## HOW
I updated the central `request` module to a) append the error status (ex: 404) to the returned JSON response if an error occurred, ensuring that places in the code where we explicitly check the status will still work, and b) to check that the API is returning a non-empty response before calling `.json()` on it in order to avoid the high volume error.

I opted, in places I'd replaced app-specific request modules with the centralized one, to add additional arguments to the API calls to ensure that errors were thrown, which was the old behavior by the app-specific request modules. That way instances like [this](https://github.com/empirical-org/Empirical-Core/blob/feat/update-main-request-module-to-use-fetch/services/QuillLMS/client/app/bundles/Grammar/actions/session.ts#L40), where we rely on an error being caught to handle the session correctly, will still work as expected. This is a little bit of a verbose solution, since it involved updating each call with the additional arg or args. Some alternatives include:

1. Re-introducing app-specific request modules, which just add the additional arguments and then call the central request module
2. Going through each individual method and only adding the extra arguments where somewhere in the codebase explicitly calls that method with a `.catch` block (from an initial survey, most but not all of these functions are invoked that way, so some of these changes are unnecessary from that perspective).

I went with the approach I did because it seemed more straightforward to me to pass the extra args directly rather than relying on an additional intermediary function (as in 1), and because option 2 seemed more likely to cause an additional bug by oversight than just updating every method to work as it did previously, regardless of whether the thrown error is handled or not. I'm open to changing it if people feel strongly otherwise, though.

### Screenshots

### Notion Card Links
https://www.notion.so/quill/Some-students-cannot-open-activities-spinning-wheel-1c8dd0ebcfc24df9bc0584b55add5a65

PR Checklist | Your Answer
------------ | -------------


eadams17 (approved):
I agree with your approach, it seems like the best option out of the options that you laid out. I'm not seeing anything that looks off, so LGTM!

brendanshean (approved):
While this does feel rather boilerplate-y, your explanation for it makes sense.  I reminds me of  engineering meeting discussion on [abstracting vs branching](https://www.notion.so/quill/Engineering-Meeting-187fbbd0975c415a8d40804ab17bb870#9a6d77cd17114f4281b8733e0602a223).  I don't think there are "right" answers for this decision making, more a case of what path do you think will be helpful in the future for extensibility while not introducing more errors.  I really appreciate the work on getting this all cleaned up.

anathomical (approved):
I sort of feel that going to promise async chains as a general rule would be a good idea in the future, but I also think it'd be a lot more risky at this time to change the shared `requestGet` call to be properly async and figuring out all the places that breaks than to do what you've done here.

Basically, I think you made the right decision for maximum immediate value, and in the future we should just re-examine the `request` module and make it better when we have time to clean up all the places that depend on it.

PR #9908, opened 2022-10-27
lms/student-classrooms-race-conditions

anathomical:
## WHAT
Refactor to handle a race condition and add specs for it
## WHY
While this isn't really impacting user experience (or I don't think it is?) since the error only occurs if the data we care about has already been saved, we don't want to clutter up our logs with errors that are easily preventable.
## HOW
I first explored trying more holistic fixes like using `create_or_find_by` instead of this weird logic here, but then realized that this Associator gets called in an `after_save` callback on the model (which, funnily enough, means that the associator calls itself if it has to create a new record), and that enough other code paths create models that it would have taken considerable effort to trace through all of the code paths and ensure that the new logic did what each of them needed.

So I went with the simplest possible solution: wrap the `find_of_initialize_by ... save` code in a block with a `rescue ... retry` on `RecordNotUnique` errors.

### Notion Card Links
https://www.notion.so/quill/StudentClassrooms-ActiveRecord-RecordNotUnique-42cbec4b457542a28e8e646b576fcde1

PR Checklist | Your Answer
------------ | -------------

### DIFF
--- a/services/QuillLMS/app/controllers/students_classrooms_controller.rb
+++ b/services/QuillLMS/app/controllers/students_classrooms_controller.rb
@@ -18,7 +18,7 @@ class StudentsClassroomsController < ApplicationController
         # rubocop:enable Style/GuardClause
       end
       Associators::StudentsToClassrooms.run(@user, classroom)
-      render json: classroom.attributes
+      render json: classroom.reload.attributes
     else
       render status: 403, json: {error: "Student not logged in."}, text: "Student not logged in."
     end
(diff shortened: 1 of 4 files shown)


happythenewsad (approved):
Good sleuthing

brendanshean (approved):
Great job on sorting out the bug here.  Long term I'd advocate removing the associator call in the method itself and then explicitly find_or_create the object outright which would then trigger the associator via the after_save callback. Nonetheless, I think this commit is an improvement to the codebase.

PR #9914, opened 2022-10-28
finish_lesson endpoint speedup 

happythenewsad:
## WHAT
Adds index to speed up certain ActivitySession queries

## WHY
So that the `finish_lesson` endpoint is faster.  

Benchmark script: 
```
Benchmark.realtime { 1000.times { ActivitySession.unscoped.where(classroom_unit_id: rand(highest_c_u_id), activity_id: rand(highest_a_s_id)) } }
```
- Time (before): .11, .098
- Time (after): .075, .085

Note: I'm considering Sidekiq-ifying the meat of this endpoint in a separate PR, depending on post-deployment performance. 

### Screenshots

### Notion Card Links

PR Checklist | Your Answer
------------ | -------------
Have you added and/or updated tests? |  no - no logic change

### DIFF
--- a/services/QuillLMS/db/structure.sql
+++ b/services/QuillLMS/db/structure.sql
@@ -56,6 +56,18 @@ CREATE FUNCTION public.blog_posts_search_trigger() RETURNS trigger
 
 
 --
+-- Name: my_jsonb_to_hstore(jsonb); Type: FUNCTION; Schema: public; Owner: -
+--
+
+CREATE FUNCTION public.my_jsonb_to_hstore(jsonb) RETURNS public.hstore
+    LANGUAGE sql IMMUTABLE STRICT
+    AS $_$
+            SELECT hstore(array_agg(key), array_agg(value))
+            FROM   jsonb_each_text($1)
+          $_$;
+
+
+--
 -- Name: old_timespent_teacher(integer); Type: FUNCTION; Schema: public; Owner: -
 --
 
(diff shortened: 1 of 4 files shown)


services/QuillLMS/db/structure.sql
```diff
@@ -55,6 +55,18 @@ CREATE FUNCTION public.blog_posts_search_trigger() RETURNS trigger
       $$;
 
 
+--
+-- Name: my_jsonb_to_hstore(jsonb); Type: FUNCTION; Schema: public; Owner: -
+--
+
+CREATE FUNCTION public.my_jsonb_to_hstore(jsonb) RETURNS public.hstore
```
happythenewsad, line 62: I believe the extra artifacts are due to the previous person dumping schema using postgres 14, rather than postgres 13.

anathomical (approved):

dandrabik (commented):
So `activity_sessions` is a relatively large table with a lot of indexes on it. If I'm reading the benchmarks correctly (running a 1,000 times, gaining a ~ tenth of a second), the performance gain seems marginal.

Should we consider not rolling this out? There is some cost to having another index, and there isn't a lot gained here (but there is some). I'm curious your thoughts.
Screen Shot 2022-11-04 at 3 55 08 PM



Screen Shot 2022-11-04 at 3 53 42 PM

happythenewsad:
> Should we consider not rolling this out? 

This endpoint isn't called super often, so I think it would be ok as-is for the medium term. 

This is the controller method:
https://github.com/empirical-org/Empirical-Core/blob/develop/services/QuillLMS/app/controllers/api/v1/classroom_units_controller.rb#L22

As you can see, it does many discrete DB operations that are not easily reduced. So any improvements we do deploy will likely be small and incremental, like this one.

happythenewsad:
@dandrabik How would you like to move forward on this?

happythenewsad:
Closing for now; we can deploy at a later time if we want.

PR #9919, opened 2022-11-01
Add Confidence level, time spent, and completion rate to Evidence Activity Stats

cissyxyu:
## WHAT
Add a few more stats to the Activity Stats page, including overall activity completion rate, overall activity time spent, and average AutoML confidence and average time spent per prompt.

## WHY
This data will help Curriculum staff make decisions about which activities and prompts to focus work on.

## HOW
Add new SQL queries to calculate and display this data from the back end. Add a new endpoint `activity_health` to fetch overall activity stats separately from prompt stats. Display this data in the front end above the prompt health table.

### Screenshots
Screen Shot 2022-11-01 at 5 33 19 PM

### Notion Card Links
https://www.notion.so/quill/Product-Board-30f97e4eb01246dbb8264253fb385073?p=455d494e4b8b4df28c4e235b887ebbad&pm=c

PR Checklist | Your Answer
------------ | -------------


services/QuillLMS/app/controllers/api/v1/rule_feedback_histories_controller.rb
```diff
@@ -24,4 +24,12 @@ def prompt_health
     report = PromptFeedbackHistory.run(**options)
     render json: report
   end
+
+  def activity_health
+    raise ArgumentError unless params.include?('activity_id')
+
+    options = params.permit(:activity_id).to_h.symbolize_keys
```
happythenewsad: I think `permit` is unnecessary because of the guard clause 2 lines above.

services/QuillLMS/app/controllers/api/v1/rule_feedback_histories_controller.rb
```diff
@@ -24,4 +24,12 @@ def prompt_health
...
+    report = ActivityFeedbackHistory.run(**options)
```
happythenewsad: The arg should be passed in as `activity_id: <value>`, rather than `**options`, because `activity_id` is the only keyword arg that ActivityFeedbackHistory currently handles. To pass in `**options` suggests that there are other options to handle.

Line 31 can also be removed, so the final code is:

`report = ActivityFeedbackHistory.run(activity_id: params['activity_id'))`

services/QuillLMS/app/lib/utils/numeric.rb
```diff
@@ -16,4 +16,8 @@ def self.to_human_string(numeric)
       numeric.to_s
     end
   end
+
+  def self.seconds_to_human_readable_time(seconds)
```
happythenewsad, line 20: 👍

services/QuillLMS/app/queries/activity_feedback_history.rb
```diff
@@ -0,0 +1,35 @@
...
+
+class ActivityFeedbackHistory
+  def self.run(activity_id:)
+    activity_stats_query(activity_id: activity_id)
+  end
+
+  def self.activity_stats_query(activity_id:)
+    activity_sessions_agg = RawSqlRunner.execute(
+      <<-SQL
+        SELECT
```
happythenewsad, line 11: I haven't tested this, but it might be more performant to do the outer query in AREL, so that the current subquery is elevated to a vanilla, non sub-query.

services/QuillLMS/app/queries/activity_feedback_history.rb
```diff
@@ -0,0 +1,35 @@
...
+          activity_sessions
+          INNER JOIN feedback_sessions ON feedback_sessions.activity_session_uid = activity_sessions.uid
+          LEFT JOIN feedback_histories ON feedback_sessions.uid = feedback_histories.feedback_session_uid
+          LEFT JOIN comprehension_prompts ON feedback_histories.prompt_id = comprehension_prompts.id
+          WHERE comprehension_prompts.activity_id = #{activity_id}
+        ) inner_query
+      SQL
+    ).to_a
+
+    average_time_spent = activity_sessions_agg[0]['average_time_spent']
```
happythenewsad, line 29: Guard against `activity_sessions_agg` being an empty array here.

happythenewsad (changes requested):
Looking good; very close

services/QuillLMS/app/queries/activity_feedback_history.rb
cissyxyu: I thought about this, but I wasn't quite sure how to use AREL with a query that's not selecting from a table. So instead of `Users.select(.....)` I wasn't sure what to put on the left side of that `.select`, since I'm selecting from a subquery and not a specific table?

services/QuillLMS/app/controllers/api/v1/rule_feedback_histories_controller.rb
brendanshean: FYI.  ActionController::Parameters (i.e. `params`) [prevents](https://github.com/rails/rails/pull/28734) calls `to_h` on  without a `permit`.

happythenewsad (approved):
LGTM

services/QuillLMS/app/queries/prompt_feedback_history.rb
```diff
@@ -30,17 +39,22 @@ def self.prompt_health_query(activity_id:, start_date: nil, end_date: nil)
     query
   end
 
+  # rubocop:disable Metrics/CyclomaticComplexity
```
brendanshean: Rather than disable the cop, you could refactor this out:

```    
class PromptFeedbackHistory
  ...

  def self.serialize_results(results)
    results.to_h { |result| [result.prompt_id, ResultSerializer.new(result).run] }
  end

  class ResultSerializer
    SERIALIZED_ATTRIBUTES = [
      :avg_attempts,
      :avg_confidence,
      :avg_time_spent,
      :display_name,
      :num_final_attempt_not_optimal,
      :num_final_attempt_optimal,
      :num_first_attempt_not_optimal,
      :num_first_attempt_optimal,
      :prompt_id,
      :session_count,
      :total_responses
    ]

    attr_reader :payload, :result

    def initialize(result)
      @result = result
      @payload = result.serializable_hash(only: SERIALIZED_ATTRIBUTES, include: [])
    end

    def run
      payload.merge(
        'avg_attempts' => avg_attempts,
        'avg_confidence' => avg_confidence,
        'avg_time_spent' => avg_time_spent,
        'num_sessions_with_consecutive_repeated_rule' =>  result.num_sessions_consecutive_repeated,
        'num_sessions_with_non_consecutive_repeated_rule' => result.num_sessions_non_consecutive_repeated
      )
    end

    def avg_attempts
      payload['avg_attempts']&.round(2)&.to_f || 0
    end

    def avg_confidence
      payload['avg_confidence'] ? payload['avg_confidence'].round(2) * 100 : 0
    end

    def avg_time_spent
      payload['avg_time_spent'] ? Utils::Numeric.seconds_to_human_readable_time(payload['avg_time_spent']) : 0
    end
  end
end
```

brendanshean (approved):
There's a refactoring comment but otherwise looks good.  Nice work on this!

PR #9880, opened 2022-10-21
Synthetic Seed Data - Backend for Paraphrasing Examples

dandrabik:
## WHAT
We want to allow the curriculum team to supply a few example answers for each label and then paraphrase them a few times for seed data. This is the backend for that.
## WHY
To help ensure this is a minimum amount of data for each label.
## HOW
Lots of experimenting with the OpenAI completion and Edit endpoints, resulted in this command that works pretty well for our use case: `"rephrase with some synonyms:\n\n"`

Add some code to pass along the label configs (frontend to come).
### Screenshots
Rephrasing for `"students can easily be distracted by their phones."` (Note, we can likely also use a version of this for Labeled data generation, with a slightly lower temperature, since the changes are subtle).

Screen Shot 2022-10-21 at 3 22 02 PM

PR Checklist | Your Answer
------------ | -------------
Have you added and/or updated tests? | 'YES'.
Have you deployed to Staging? |  No - this is only the backend, will do that with the frontend code.
Self-Review: Have you done an initial self-review of the code below on Github? | Yes
Spec Review: Have you reviewed the spec and ensured this meets requirements and/or matches design mockups? | N/A

PR #9928, opened 2022-11-02
Synthetic Labels Frontend

dandrabik:
## WHAT
Front end for the [Synthetic Label examples PR](https://github.com/empirical-org/Empirical-Core/pull/9880) 

This adds a form to the Seed Data create page to add label examples for paraphrasing.

(Note, this is a PR on that PR, not `develop`. [This](https://github.com/empirical-org/Empirical-Core/compare/synthetic_labels_fe) is the diff on `develop` if you want that for reference.
## WHY
We want to make sure we have some seed data for all labels. This UI allows the team to submit some examples per label that we will paraphrase for them.
## HOW
Somewhat open-ended form for each prompt.
- No validation, to give the team complete flexibility
- Making this 2 examples per label for now consistency/simplicity (easy to expand later, it's just an array)
### Screenshots
Screen Shot 2022-11-02 at 4 14 34 PM

### Notion Card Links
https://www.notion.so/quill/Synthetic-Data-V2-6205ada84ba4445eac0cdf179435d584

PR Checklist | Your Answer
------------ | -------------

### DIFF
--- a/services/QuillLMS/client/app/bundles/Staff/components/evidence/syntheticData/seedDataForm.tsx
+++ b/services/QuillLMS/client/app/bundles/Staff/components/evidence/syntheticData/seedDataForm.tsx
@@ -43,12 +43,24 @@ const SeedDataForm = ({ history, match }) => {
     setLabelConfigs(labelConfigs => ({...data}));
   }
 
+  function onAddLabelConfigs(e) {
+    const { target } = e;
+    const { value } = target;
+    addLabelConfigs(value);
+  }
+
   function addLabelConfigs(conjunction) {
     const data = labelConfigs
     const conjunctionData = [...data[conjunction], {...blankLabelConfig}];
 
     data[conjunction] = conjunctionData
     setLabelConfigs(labelConfigs => ({...data}))
+  }
+
+  function onRemoveLabelConfig(e) {
+    const { target } = e;
+    const { id, value } = target;
+    removeLabelConfig(id, value);
   }
 
   function removeLabelConfig(index, conjunction) {
@@ -112,7 +124,9 @@ const SeedDataForm = ({ history, match }) => {
       <div className="seed-label-form" key={index}>
         <button
           className='right quill-button fun secondary outlined'
-          onClick={() => removeLabelConfig(index, conjunction)}
+          id={index}
+          onClick={onRemoveLabelConfig}
+          value={conjunction}
         >
           Remove
         </button>
@@ -139,7 +153,11 @@ const SeedDataForm = ({ history, match }) => {
           &nbsp;Label Examples
         </h4>
         {labelConfigs[conjunction].map((labelConfig, index) => renderLabelConfig(labelConfig, index, conjunction))}
-        <button className='quill-button small primary outlined' onClick={e => addLabelConfigs(conjunction)}>
+        <button
+          className='quill-button small primary outlined'
+          onClick={onAddLabelConfigs}
+          value={conjunction}
+        >
           <span className='plus'>+</span>
 
           &nbsp;Add {capitalizeConjunction} Label
--- a/services/QuillLMS/client/app/bundles/Staff/components/evidence/syntheticData/seedDataForm.tsx
+++ b/services/QuillLMS/client/app/bundles/Staff/components/evidence/syntheticData/seedDataForm.tsx
@@ -18,7 +18,6 @@ const SeedDataForm = ({ history, match }) => {
 
   const [activityNouns, setActivityNouns] = React.useState<string>('');
 
-
   const because = 'because';
   const but = 'but';
   const so = 'so';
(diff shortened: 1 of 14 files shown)


services/QuillLMS/client/app/bundles/Staff/components/evidence/syntheticData/seedDataForm.tsx
```diff
@@ -18,6 +18,52 @@ const SeedDataForm = ({ history, match }) => {
 
   const [activityNouns, setActivityNouns] = React.useState<string>('');
 
+  const because = 'because';
+  const but = 'but';
+  const so = 'so';
+
+  const blankLabelConfig = { label: '', examples: ['',''], };
+  const blankLabelConfigs = { because : [], but : [], so : [], };
```
eadams17: we already have these consts defined [here](https://github.com/empirical-org/Empirical-Core/blob/develop/services/QuillLMS/client/app/constants/evidence.ts#L4) and I think these other two variables could be exported from there as well

services/QuillLMS/client/app/bundles/Staff/components/evidence/syntheticData/seedDataForm.tsx
```diff
@@ -18,6 +18,52 @@ const SeedDataForm = ({ history, match }) => {
...
+
+    conjunctionData[index].examples[exampleIndex] = event.target.value;
+    data[conjunction] = conjunctionData;
+
+    setLabelConfigs(labelConfigs => ({...data}));
+  }
+
+  const addLabelConfigs = (conjunction) => {
+    let data = labelConfigs
+    let conjunctionData = [...data[conjunction], {...blankLabelConfig}];
```
eadams17: I think most of these lets can be consts since we're only updating their properties rather than conditionally redefining the variable

services/QuillLMS/client/app/bundles/Staff/components/evidence/syntheticData/seedDataForm.tsx
```diff
@@ -54,6 +101,57 @@ const SeedDataForm = ({ history, match }) => {
     );
   }
 
+  const renderExample = (value, index, conjunction, exampleIndex) => {
```
eadams17: this is a small nit (and something we should document), but Emilia and I have agreed for readability's sake to define a component's inner functions with `function` and reserve const for any variable declaration (i.e. a shared class name `const blueLabelClass = 'bg-quillteal label-title'`)

services/QuillLMS/client/app/bundles/Staff/components/evidence/syntheticData/seedDataForm.tsx
```diff
@@ -54,6 +101,57 @@ const SeedDataForm = ({ history, match }) => {
...
+      />
+    );
+  }
+
+  const renderLabelConfig = (labelConfig, index, conjunction) => {
+    return (
+      <div className="seed-label-form" key={index}>
+        <button
+          className='right quill-button fun secondary outlined'
+          onClick={() => removeLabelConfig(index, conjunction)}
```
eadams17: We have [a linter](https://github.com/jsx-eslint/eslint-plugin-react/blob/master/docs/rules/jsx-no-bind.md) in place that flags against arrow functions in props. It causes performance issues because it will cause a brand new function to be created on every render, so the preferred style is this way:

```
<button id={index} onClick={onRemoveLableConfig} value={conjunction} />
function onRemoveLabelConfig(e) {
  const { target } = e;
  const { id, value } = target;
  removeLabelConfig(id, value);
}
```

I think it's fine to add an exception for cases like the one below `e => handleLabelConfigsChange(e, index, conjunction, 'label')` since there are multiple arguments, but just wanted to flag how we typically handle these kinds of change events

services/QuillLMS/client/app/bundles/Staff/components/evidence/syntheticData/seedDataForm.tsx
```diff
@@ -54,6 +101,57 @@ const SeedDataForm = ({ history, match }) => {
...
+  const renderLabelSection = (conjunction) => {
+    let capitalizeConjunction = conjunction.charAt(0).toUpperCase() + conjunction.substring(1)
+    return (
+      <div className='label-section'>
+        <h4 className='bg-quillteal label-title'>
+          <span className='highlight'>{capitalizeConjunction}</span>
+          &nbsp;Label Examples
+        </h4>
+        {labelConfigs[conjunction].map((labelConfig, index) => renderLabelConfig(labelConfig, index, conjunction))}
+        <button className='quill-button small primary outlined' onClick={e => addLabelConfigs(conjunction)}>
```
eadams17: see same comment above about click handlers that only pass one or two arguments

eadams17 (approved):
Looks good to me! I left a few comments around suggested optimizations/cleanup, but nothing that warrants another review

services/QuillLMS/client/app/bundles/Staff/components/evidence/syntheticData/seedDataForm.tsx
```diff
@@ -18,28 +18,83 @@ const SeedDataForm = ({ history, match }) => {
...
+  const [labelConfigs, setLabelConfigs] = React.useState({...blankLabelConfigs});
+
+  function handleLabelConfigsChange(event, index, conjunction, key) {
+    const data = labelConfigs
+    const conjunctionData = [...data[conjunction]];
+
+    conjunctionData[index][key] = event.target.value;
+    data[conjunction] = conjunctionData;
+
+    setLabelConfigs(labelConfigs => ({...data}));
```
emilia-friedberg: This works fine, but since you aren't ever using the `labelConfigs` argument here and don't really need to worry about mutating `data` since it's only defined in the function scope, the more succinct way to write this would just be `setLabelConfigs(data)`. This is true below as well.

emilia-friedberg (approved):
Small suggestion for concision but overall LGTM!

services/QuillLMS/client/app/bundles/Staff/components/evidence/syntheticData/seedDataForm.tsx
dandrabik: Good note. I've removed the `labelConfigs` to make it `setLabelConfigs({...data})`. Note, I learned this during this project, but since the underlying `useState` item here is an object with properties I'm updating, i.e. a `{'because' : [], 'but': [], 'so' []}` object in which I edit the arrays , it seems I need to clone it (i.e. `{...data}` vs. `data`) to make the `state` update properly (this surprised me). I believe this is because of the underlying diff logic will otherwise see it as the same object and not update (if you don't clone the object, the code doesn't function).

And to just make this cleaner, I'll move the cloning up to the `data` assignment
```javascript
const data = {...labelConfigs}
setLabelConfigs(data)

PR #9929, opened 2022-11-03
Evidence session filtering/ responses for scoring filter

eadams17:
## WHAT
add a filter to Evidence sessions index for filtering by responses for scoring

## WHY
this will make it easier for Curriculum to parse session data

## HOW
update sessions query to add `responses_for_scoring` check, add frontend logic to toggle filter

### Screenshots
Screen Shot 2022-11-03 at 3 36 03 PM
Screen Shot 2022-11-03 at 3 36 13 PM

### Notion Card Links
https://www.notion.so/quill/Activity-Session-Filtering-Add-a-new-filter-for-responses-246894fb389f4ceebf69da6be7d3acc8

PR Checklist | Your Answer
------------ | -------------


services/QuillLMS/app/models/feedback_history.rb
```diff
@@ -221,6 +221,21 @@ def self.completeness_filter_query(complete)
     SQL
   end
 
+  def self.responses_for_scoring
```
happythenewsad: It's not clear to me how this code maps to the spec. 

`all prompts sum to > 5 || each prompt has at least one entry` ?

happythenewsad: If you end up keeping this nested case structure, I recommend breaking out the two substatements into separate ruby methods, then combine them through string interpolation, so that they're easier to interpret.

eadams17: @happythenewsad apologies, the logic for this was communicated over Slack with Emma. The requested logic for responses for scoring is: 6+ responses per session OR sessions with 2+ responses for each conjunction, just added that as a comment in the Notion card

emilia-friedberg (commented):
The code looks good to me but we probably want some rspec tests here for the new logic.

services/QuillLMS/app/models/feedback_history.rb
eadams17: @happythenewsad just pushed changes to break that statement out into two substatements

eadams17:
> The code looks good to me but we probably want some rspec tests here for the new logic.

@emilia-friedberg just added some tests!

emilia-friedberg (approved):

services/QuillLMS/app/models/feedback_history.rb
```diff
@@ -221,6 +221,34 @@ def self.completeness_filter_query(complete)
...
+      CASE WHEN
+        (COUNT(CASE WHEN comprehension_prompts.conjunction = '#{BECAUSE}' THEN 1 END) > 1 AND
+        COUNT(CASE WHEN comprehension_prompts.conjunction = '#{BUT}' THEN 1 END) > 1 AND
+        COUNT(CASE WHEN comprehension_prompts.conjunction = '#{SO}' THEN 1 END) > 1)
+      THEN true ELSE false END
+    ) = true
+    SQL
+  end
+
+  def self.responses_for_scoring
```
happythenewsad, line 248: Looks great!

happythenewsad (approved):

PR #9936, opened 2022-11-07
Staggered release endpoints part4

brendanshean:
## WHAT
Recalculate `UserPackSequenceItem` status when an activity session is destroyed.

## WHY
When a user has an activity session destroyed, we need to recalculate any associated `UserPackSequenceItem` status since such a change might lock a particular ActivityPack.

## HOW
1.  Add an `after_destroy` callback to ActivitySession that calls `save_user_pack_sequence_items`.
2. `UserPackSequenceItemSaver` uses results from `UserPackSequenceItemQuery` to calculate individual `UserPackSequenceItem` status.  
3. Finally, create/update the `UserPackSequenceItem` with the computed status.

Also, there's a few migrations in here.  Three of them concern uniqueness of PackSequence, PackSequenceItem, and UserPackSequenceItem records.  While updating these status on UserPackSequence, it's important to not create duplicates.  The other migration`RenamePackSequenceItemIdToUnitId` attempts to make the code and queries more readable since we are basically using `units` and the presence of item as a foreign key just adds boilerplate.

Addendum: (this was the state machine I used for this piece of [logic](https://github.com/empirical-org/Empirical-Core/pull/9936/files#diff-fcef127d69a153c852bcc8fde46b233316972c00ad4f11d60641e983357d2eedR27))
Screen Shot 2022-11-08 at 10 04 47 AM

### Screenshots

### Notion Card Links

PR Checklist | Your Answer
------------ | -------------


services/QuillLMS/app/services/user_pack_sequence_item_saver.rb
```diff
@@ -0,0 +1,52 @@
...
+      previous_pack_sequence_id = current_pack_sequence_id
+    end
+  end
+
+  private def save_statuses
+    ActiveRecord::Base.transaction do
+      user_pack_sequence_item_statuses.each_pair do |pack_sequence_item_id, status|
+        UserPackSequenceItem
+          .find_or_create_by!(user_id: user_id, pack_sequence_item_id: pack_sequence_item_id)
+          .tap { |upsi| upsi.update!(status: status) unless status == upsi.status }
```
emilia-friedberg: Now I know `.tap`!

emilia-friedberg (approved):
This looks good to me. It took a couple of read-throughs to follow the saver logic, and I'm sure it took many more iterations to think through and write it. I know there's a lot to hold in your head - thank you for all of your work on this.

services/QuillLMS/app/models/unit_activity.rb
```diff
@@ -150,10 +150,7 @@ def self.get_classroom_user_profile(classroom_id, user_id)
           ua.publish_date #{teacher_timezone_offset_string} AS publish_date,
           pre_activity.id AS pre_activity_id,
           cu.created_at AS unit_activity_created_at,
-          ps.release_method AS #{PackSequence::RELEASE_METHOD_KEY},
-          ps.id AS #{PackSequence::ID_KEY},
-          psi.id AS #{PackSequenceItem::ID_KEY},
-          psi.order AS #{PackSequenceItem::ORDER_KEY},
+          COALESCE(upsi.status, '#{UserPackSequenceItem::NO_LOCK}') AS user_pack_sequence_item_status,
```
dandrabik: Ok, so `status` is a trinary, `LOCKED`, `UNLOCKED`, `NO_LOCK`. 

Take or leave, you might consider keeping this third state as `nil`, instead of renaming to `NO_LOCK`. It's not a big deal, but you are kinda committing to doing this `COALESCE` on every query for this attribute from this point on, which may be more of a gotcha than the logic having a relatively readable `status.nil?` check (There are pros to having the explicit `NO_LOCK` so this isn't a strongly held opinion, and it's always easy to change later).

services/QuillLMS/app/queries/user_pack_sequence_item_query.rb
```diff
@@ -0,0 +1,41 @@
+# frozen_string_literal: true
+
+class UserPackSequenceItemQuery
+  FINISHED = ActivitySession::STATE_FINISHED
+  COMPLETED_KEY = 'completed'
+  PACK_SEQUENCE_ITEM_ID_KEY = 'pack_sequence_item_id'
+
+  def self.call(classroom_id, user_id)
```
dandrabik, line 8: Curious why a `self.call` here instead of the service object `run` pattern you've been using in most of your code. Do you view `queries` differently?

services/QuillLMS/app/services/user_pack_sequence_item_saver.rb
```diff
@@ -0,0 +1,52 @@
...
+  PACK_SEQUENCE_ITEM_ID_KEY = UserPackSequenceItemQuery::PACK_SEQUENCE_ITEM_ID_KEY
+
+  def initialize(classroom_id, user_id)
+    @classroom_id = classroom_id
+    @user_id = user_id
+    @user_pack_sequence_item_statuses = {}
+  end
+
+  def run
+    compute_statuses
```
dandrabik: I'm curious why you decided to do this in two passes, compute everything THEN store everything vs. compute and store in one loop?

Asking mainly because I think you want the `select`s and the `write`s both in the `transaction` to prevent race conditions (e.g. the data changes between when you read it and when you write to the DB).

dandrabik (approved):
Overall looks good, a few small comments that you should look at, but most are take or leave. Thanks for pushing this closer to release!

services/QuillLMS/app/queries/user_pack_sequence_item_query.rb
brendanshean: Conceptually, I view service objects as a generic computation with a return type that can be nil and possible side-effects, whereas queries are a SQL wrapper with a return non-nil return type (usually a ActiveRecord::Relation or PG::Result) that can be chained/filtered/composed and has no side-effects.  

As I work through the different use cases, I feel like the [query object pattern](https://thoughtbot.com/blog/a-case-for-query-objects-in-rails) will probably help, but I don't yet know what that will look like, so I'm just using the existing `self.call` pattern from other code in our codebase.

services/QuillLMS/app/services/user_pack_sequence_item_saver.rb
brendanshean: My thought for breaking this up was to limit the number of `update!(status: status)` to one per pack_sequence_item.  Otherwise, you end up doing an update per unit_activity in the pack_sequence_item.

However, you raise a good point that this leaves us open to race conditions, so I would agree on putting them inside the transaction instead and doing the redundant updates.

services/QuillLMS/app/models/unit_activity.rb
brendanshean: That makes sense.  I will just pass status instead and `nil` can be interpreted as `NO_LOCK` on the front-end

PR #9944, opened 2022-11-08
Log changes to Evidence Activity flags

cissyxyu:
## WHAT
Make a ChangeLog record every time a staff member updates an Evidence Activity's flag.

## WHY
Curriculum staff want to be able to track when Evidence Activities have changes made to flags.

## HOW
Add a callback to the `Activity` model that creates a ChangeLog record if the activity is evidence and the flag was updated.
Pass the user ID from the session to the Activity model via virtual field `lms_user_id` through both the CMS Activities Controller and the Evidence Engine Activities Controller (since there are two places staff can currently edit the flag for an Evidence Activity).

### Screenshots
Screen Shot 2022-11-08 at 6 36 53 PM

### Notion Card Links
https://www.notion.so/quill/Record-Evidence-activity-flag-changes-in-the-activity-change-log-69a8948da3714c59afef3826634696d1

PR Checklist | Your Answer
------------ | -------------


services/QuillLMS/app/controllers/cms/activities_controller.rb
```diff
@@ -5,6 +5,7 @@ class Cms::ActivitiesController < Cms::CmsController
   before_action :set_activity, only: [:update, :destroy, :edit]
   before_action :set_style_and_javascript_file, only: [:new, :edit]
   before_action :set_raw_score_options_and_raw_score_to_readability_grade_band, only: [:new, :edit]
+  append_before_action :set_lms_user_id_if_evidence, only: [:update]
```
brendanshean: Why `append_before_action` instead of just `before_action`?

services/QuillLMS/app/models/activity.rb
```diff
@@ -289,10 +292,28 @@ def segment_activity
...
+  private def update_evidence_flags?
+    is_evidence? && saved_change_to_flags? && child_activity
+  end
+
   private def update_evidence_child_title
     child_activity&.update(title: name)
   end
 
+  private def log_evidence_flag_change
+    change_log = {
```
brendanshean, line 304: I would get rid of the temp variable `change_log` and call `create` on hash directly:

```
Change.create(
  user_id: @lms_user_id,
  action: ChangeLog::EVIDENCE_ACTIONS[:update],
  ..
)
```

services/QuillLMS/engines/evidence/app/controllers/evidence/activities_controller.rb
```diff
@@ -5,7 +5,7 @@
 module Evidence
   class ActivitiesController < ApiController
     before_action :set_activity, only: [:activity_versions, :create, :show, :update, :destroy, :seed_data, :change_logs, :labeled_synthetic_data]
-    append_before_action :set_lms_user_id, only: [:create, :destroy]
+    append_before_action :set_lms_user_id, only: [:create, :destroy, :update]
```
brendanshean: Why `append_before_action` instead of just `before_action`?

services/QuillLMS/engines/evidence/spec/dummy/app/models/activity.rb
```diff
@@ -8,4 +11,26 @@ def flag(flag = nil)
...
+  def child_activity
+    Evidence::Activity.find_by(parent_activity_id: id)
+  end
+
+  private def update_evidence_flags?
+    child_activity && saved_change_to_flags?
+  end
+
+  private def log_evidence_flag_change
+    change_log = {
```
brendanshean, line 24: Same as earlier comment, remove temporary variable `change_log`.

brendanshean (approved):
Nice work on this.  I especially like the use of virtual attributes on this.

services/QuillLMS/app/controllers/cms/activities_controller.rb
happythenewsad: The former is an alias of the latter....

https://github.com/rails/rails/blob/a74041d483041345454e766c506dab1b567dec0d/actionpack/lib/abstract_controller/callbacks.rb#L219

happythenewsad (approved):
Approving, but I'm also unsure about the choice of the `append` callback

PR #9947, opened 2022-11-08
Add sidekiq-unique-jobs gem

brendanshean:
## WHAT
Add the sidekiq-unique-jobs gem to our stack.

## WHY 
The staggered release [backend](https://github.com/empirical-org/Empirical-Core/pull/9936) will call the job  `SaveUserPackSequenceItemsWorker` for various [reasons](https://www.notion.so/quill/Modeling-UserPackSequenceItem-status-in-the-db-684d74ffc765408fadcfd8aaf5007813) (15 and counting...).  While the jobs are idempotent, the query is complex and not really worth running multiple times.  This gem will give us options for handling non unique jobs from running simultaneously.

## HOW
Add the library to the Gemfile and configure the sidekiq server and client.

### Screenshots

### Notion Card Links

PR Checklist | Your Answer
------------ | -------------
Have you added and/or updated tests? |  No.  Just adding a library.

### DIFF
--- a/services/QuillLMS/Gemfile
+++ b/services/QuillLMS/Gemfile
@@ -90,6 +90,7 @@ source "https://gems.contribsys.com/" do
 end
 gem 'sidekiq', '~> 5.2.10'
 gem 'sidekiq-retries', require: false
+gem 'sidekiq-unique-jobs'
 gem 'redis', '~> 4.5'
 gem 'redis-namespace'
 gem 'redis-rails'
--- a/services/QuillLMS/Gemfile.lock
+++ b/services/QuillLMS/Gemfile.lock
@@ -139,6 +139,9 @@ GEM
       autoprefixer-rails (>= 5.2.1)
       sassc (>= 2.0.0)
     brakeman (4.1.1)
+    brpoplpush-redis_script (0.1.3)
+      concurrent-ruby (~> 1.0, >= 1.0.5)
+      redis (>= 1.0, < 6)
     builder (3.2.4)
     bulk_insert (1.9.0)
       activerecord (>= 3.2.0)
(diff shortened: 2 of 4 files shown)


dandrabik (approved):
So the gem config here looks good. I'm curious how this will work in practice (it took me a few reads of the gem docs to find that the job `params` were part of the "uniqueness" calculation. Since it's a nice to have (it's not the end of the world if one of these runs twice), this seems fine to experiment with when you use it (or you could merge this once you have a use-case).

It's also interesting the gem maintainer commented that the uniqueness part was confusing to people (under the testing section):
https://github.com/mhenrixon/sidekiq-unique-jobs#uniqueness

Screen Shot 2022-11-08 at 6 27 33 PM

brendanshean:
@dandrabik It also took me several re-reads to figure out how to use this. I originally started with v5 since I thought this aligned with the version of sidekiq (we are on v5.2.10).  I still wanted to actually test it [locally](https://github.com/mhenrixon/sidekiq-unique-jobs/issues/256#issuecomment-407149261) to see it work, which ended up not being very illuminating, which was like he intimated in the docs "leave the testing of the gem to the maintainers".  I bumped it up to v7 since it's still compatible with our sidekiq config.  

Anyways, my plan is to test this out on staging when the various use cases are complete.

happythenewsad (commented):
Are you planning to use the `reject` strategy within this gem? That will cause a lot of dead jobs. We'll need to be aware of this when debugging other dead jobs. 

Is it possible to just call `SaveUserPackSequenceItemsWorker` less often?

brendanshean:
@happythenewsad Good question.  My plan is to use [until executed](https://github.com/mhenrixon/sidekiq-unique-jobs#until-executed) configuration which doesn't use a reject strategy.  While this would not solve the issue of running duplicate jobs, it should handle the concern of running them simultaneously.

brendanshean:
@happythenewsad Thank you for flagging the dead queue concern. If end up changing the configuration to `WhileExecuting`, I would probably opt for the `reschedule` option to avoid polluting the dead queue.

PR #9957, opened 2022-11-10
Add left alignment for username column

brendanshean:
## WHAT
Fix a bug with that right aligns usernames that can be parsed as integers.

## WHY
Some usernames can be parsed as integers when in fact they should be displayed as a string.  

## HOW
Add an override for username column that returns `left` for the username column.

### Screenshots
Screen Shot 2022-11-10 at 11 26 30 AM

### Notion Card Links
https://www.notion.so/quill/Usernames-Clever-IDs-are-not-centered-on-the-username-column-for-Clever-Secure-Sync-classes-bf7be2449a4148c9bc75b39c7894ae8c

PR Checklist | Your Answer
------------ | -------------
Have you added and/or updated tests? |  No.  Manual check.


emilia-friedberg (approved):
I think this works as a temporary bug fix. I would probably try to add more general logic, since this is a Backpack component (something like passing a `textAlign` property in as part of the header object, and then checking for the presence of that value in this method and doing an early return if it's present), but I'm fine with getting this solution out and can make a follow up card to generalize the logic later.

PR #9958, opened 2022-11-10
teacher onboarding materials backend

emilia-friedberg:
## WHAT
Add migrations, models, a controller, and tests for the teacher onboarding materials project.

## WHY
This is how we'll store the data we collect during teacher onboarding (grade range and subject areas).

## HOW
Added the migrations as outlined in the RFC discussion and write out the model and controller files I'll need for the frontend, including @happythenewsad's suggestion to automatically transform the 'K' value to 0 on its way in and out of the database so we don't have to manually change it.

### Screenshots

### Notion Card Links
https://www.notion.so/quill/Collecting-Teacher-Onboarding-Backend-fa724f7b7ccd491fbad0c7de2f9c73e3

PR Checklist | Your Answer
------------ | -------------
Have you deployed to Staging? | NO - changes aren't hooked up to frontend yet (but I did test in console)

### DIFF
--- a/services/QuillLMS/app/controllers/teacher_infos_controller.rb
+++ b/services/QuillLMS/app/controllers/teacher_infos_controller.rb
@@ -1,28 +1,28 @@
 class TeacherInfosController < ApplicationController
 
   def create
-    teacher_info = TeacherInfo.create(
-      teacher: current_user,
-      minimum_grade_level: teacher_info_params[:minimum_grade_level],
-      maximum_grade_level: teacher_info_params[:maximum_grade_level]
+    teacher_info = TeacherInfo.create!(
+      user: current_user,
+      minimum_grade_level: minimum_grade_level,
+      maximum_grade_level: maximum_grade_level
     )
 
-    subject_areas = SubjectArea.where(id: teacher_info_params[:subject_areas])
+    subject_areas = SubjectArea.where(id: subject_area_ids)
     teacher_info.subject_areas.push(subject_areas)
 
     render json: {}, status: 200
   end
 
   def update
-    teacher_info = TeacherInfo.find_or_create_by(teacher: current_user)
+    teacher_info = TeacherInfo.find_or_create_by!(user: current_user)
 
-    if teacher_info_params[:minimum_grade_level] && teacher_info_params[:maximum_grade_level]
-      teacher_info.update(minimum_grade_level: teacher_info_params[:minimum_grade_level], maximum_grade_level: teacher_info_params[:maximum_grade_level])
+    if minimum_grade_level && maximum_grade_level
+      teacher_info.update!(minimum_grade_level: minimum_grade_level, maximum_grade_level: maximum_grade_level)
     end
 
-    if teacher_info_params[:subject_areas]
+    if subject_area_ids
       teacher_info.teacher_info_subject_areas.destroy_all
-      subject_areas = SubjectArea.where(id: teacher_info_params[:subject_areas])
+      subject_areas = SubjectArea.where(id: subject_area_ids)
       teacher_info.subject_areas.push(subject_areas)
     end
 
@@ -30,7 +30,20 @@ class TeacherInfosController < ApplicationController
   end
 
   private def teacher_info_params
-    params.permit(:minimum_grade_level, :maximum_grade_level, subject_areas: [])
+    params.permit(:minimum_grade_level, :maximum_grade_level, subject_area_ids: [])
   end
+
+  private def minimum_grade_level
+    teacher_info_params[:minimum_grade_level]
+  end
+
+  private def maximum_grade_level
+    teacher_info_params[:maximum_grade_level]
+  end
+
+  private def subject_area_ids
+    teacher_info_params[:subject_area_ids]
+  end
+
 
 end
--- a/services/QuillLMS/app/controllers/teacher_infos_controller.rb
+++ b/services/QuillLMS/app/controllers/teacher_infos_controller.rb
@@ -1,3 +1,5 @@
+# frozen_string_literal: true
+
 class TeacherInfosController < ApplicationController
 
   def create
--- a/services/QuillLMS/app/controllers/teacher_infos_controller.rb
+++ b/services/QuillLMS/app/controllers/teacher_infos_controller.rb
@@ -1,0 +1,36 @@
+class TeacherInfosController < ApplicationController
+
+  def create
+    teacher_info = TeacherInfo.create(
+      teacher: current_user,
+      minimum_grade_level: teacher_info_params[:minimum_grade_level],
+      maximum_grade_level: teacher_info_params[:maximum_grade_level]
+    )
+
+    subject_areas = SubjectArea.where(id: teacher_info_params[:subject_areas])
+    teacher_info.subject_areas.push(subject_areas)
+
+    render json: {}, status: 200
+  end
+
+  def update
+    teacher_info = TeacherInfo.find_or_create_by(teacher: current_user)
+
+    if teacher_info_params[:minimum_grade_level] && teacher_info_params[:maximum_grade_level]
+      teacher_info.update(minimum_grade_level: teacher_info_params[:minimum_grade_level], maximum_grade_level: teacher_info_params[:maximum_grade_level])
+    end
+
+    if teacher_info_params[:subject_areas]
+      teacher_info.teacher_info_subject_areas.destroy_all
+      subject_areas = SubjectArea.where(id: teacher_info_params[:subject_areas])
+      teacher_info.subject_areas.push(subject_areas)
+    end
+
+    render json: {}, status: 200
+  end
+
+  private def teacher_info_params
+    params.permit(:minimum_grade_level, :maximum_grade_level, subject_areas: [])
+  end
+
+end
--- a/services/QuillLMS/app/controllers/teacher_infos_controller.rb
+++ b/services/QuillLMS/app/controllers/teacher_infos_controller.rb
@@ -24,8 +24,7 @@ class TeacherInfosController < ApplicationController
 
     if subject_area_ids
       teacher_info.teacher_info_subject_areas.destroy_all
-      subject_areas = SubjectArea.where(id: subject_area_ids)
-      teacher_info.subject_areas.push(subject_areas)
+      subject_area_ids.each { |subject_area_id| TeacherInfoSubjectArea.create!(teacher_info: teacher_info, subject_area_id: subject_area_id) }
     end
 
     render json: {}, status: 200
--- a/services/QuillLMS/app/models/teacher_info.rb
+++ b/services/QuillLMS/app/models/teacher_info.rb
@@ -1,5 +1,16 @@
 # frozen_string_literal: true
 
+# == Schema Information
+#
+# Table name: teacher_infos
+#
+#  id                  :bigint           not null, primary key
+#  minimum_grade_level :integer
+#  maximum_grade_level :integer
+#  teacher_id          :bigint
+#  created_at          :datetime         not null
+#  updated_at          :datetime         not null
+#
 class TeacherInfo < ApplicationRecord
   belongs_to :teacher, :class_name => 'User'
 
@@ -8,6 +19,7 @@ class TeacherInfo < ApplicationRecord
 
   validates :minimum_grade_level, numericality: { in: 0..12 }
   validates :maximum_grade_level, numericality: { in: 0..12 }
+  validates :teacher_id, presence: true, uniqueness: true
 
   KINDERGARTEN_DISPLAY_STRING = 'K'
   KINDERGARTEN_DATABASE_INTEGER = 0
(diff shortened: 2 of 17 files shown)


services/QuillLMS/app/controllers/teacher_infos_controller.rb
```diff
@@ -0,0 +1,36 @@
+class TeacherInfosController < ApplicationController
+
+  def create
+    teacher_info = TeacherInfo.create(
```
brendanshean: Let's add `create!` to make sure that this doesn't fail silently.

services/QuillLMS/app/controllers/teacher_infos_controller.rb
```diff
@@ -0,0 +1,36 @@
...
+    )
+
+    subject_areas = SubjectArea.where(id: teacher_info_params[:subject_areas])
+    teacher_info.subject_areas.push(subject_areas)
+
+    render json: {}, status: 200
+  end
+
+  def update
+    teacher_info = TeacherInfo.find_or_create_by(teacher: current_user)
```
brendanshean: Let's use `find_or_create_by!` to ensure this doesn't fail silently.

services/QuillLMS/app/controllers/teacher_infos_controller.rb
```diff
@@ -0,0 +1,36 @@
+class TeacherInfosController < ApplicationController
+
+  def create
+    teacher_info = TeacherInfo.create(
+      teacher: current_user,
+      minimum_grade_level: teacher_info_params[:minimum_grade_level],
+      maximum_grade_level: teacher_info_params[:maximum_grade_level]
+    )
+
+    subject_areas = SubjectArea.where(id: teacher_info_params[:subject_areas])
```
brendanshean: I feel like `subject_area_ids` would more effectively communicate the data type.

services/QuillLMS/spec/controllers/teacher_infos_controller_spec.rb
```diff
@@ -0,0 +1,51 @@
...
+      teacher_info = TeacherInfo.find_by(teacher: user)
+
+      expect(teacher_info.minimum_grade_level).to eq(4)
+      expect(teacher_info.maximum_grade_level).to eq(12)
+      expect(teacher_info.subject_areas).to eq([subject_area1])
+      expect(response.status).to be(200)
+    end
+  end
+
+  describe '#update' do
```
brendanshean, line 38: A common pattern with specs is to use a `let!`  and a `before` it` block to setup the state you expect to exist outside of the the actual test:

```
describe '#update' do
  let!(:teacher_info) { create(:teacher_info, minimum_grade_level: 1, maximum_grade_level: 7) }
  
  before { teacher_info.subject_areas.push(subject_area1) }
  
  it 'should update the teacher info record with the data populated' do
      put :update, params: {minimum_grade_level: 4, maximum_grade_level: 12, subject_areas: [subject_area2.id]}
    ...
```

One reason I like this approach is it lends itself better to refactoring when you want to write similar specs that vary only in attributes.

services/QuillLMS/app/controllers/teacher_infos_controller.rb
```diff
@@ -0,0 +1,36 @@
...
+      teacher_info.subject_areas.push(subject_areas)
+    end
+
+    render json: {}, status: 200
+  end
+
+  private def teacher_info_params
+    params.permit(:minimum_grade_level, :maximum_grade_level, subject_areas: [])
+  end
+
```
brendanshean, line 48: You could add some private methods here to clean up some of the boilerplate:

```
  private def minimum_grade_level
    teacher_info_params[:minimum_grade_level]
  end

  private def maximum_grade_level
    teacher_info_params[:maximum_grade_level]
  end

  private def subject_area_ids
    teacher_info_params[:subject_area_ids]
  end
```

services/QuillLMS/app/controllers/teacher_infos_controller.rb
```diff
@@ -0,0 +1,36 @@
...
+    teacher_info.subject_areas.push(subject_areas)
+
+    render json: {}, status: 200
+  end
+
+  def update
+    teacher_info = TeacherInfo.find_or_create_by(teacher: current_user)
+
+    if teacher_info_params[:minimum_grade_level] && teacher_info_params[:maximum_grade_level]
+      teacher_info.update(minimum_grade_level: teacher_info_params[:minimum_grade_level], maximum_grade_level: teacher_info_params[:maximum_grade_level])
```
brendanshean: Let's use `update!` to ensure this doesn't fail silently

services/QuillLMS/config/routes.rb
```diff
@@ -161,6 +161,10 @@
     delete :destroy_by_activity_id_for_current_user, on: :collection
   end
 
+  resources :teacher_infos, only: [:create] do
```
brendanshean, line 164: I think you can just use one-liner `resources :teacher_infos, only: [:create, :update]`

services/QuillLMS/db/migrate/20221109182145_create_teacher_info_subject_area.rb
```diff
@@ -0,0 +1,10 @@
+class CreateTeacherInfoSubjectArea < ActiveRecord::Migration[6.1]
+  def change
+    create_table :teacher_info_subject_areas do |t|
+      t.references :teacher_info
```
brendanshean: Can teacher_info or subject_area ever be null for a valid saved object?

If not, consider validates presence for them and making the DB fields non-null.

(basically a [comment](https://github.com/empirical-org/Empirical-Core/pull/9866/files#r1001850632) Dan wrote on one of my recent PRs)

services/QuillLMS/app/models/teacher_info.rb
```diff
@@ -0,0 +1,47 @@
...
+
+  has_many :teacher_info_subject_areas, dependent: :destroy
+  has_many :subject_areas, through: :teacher_info_subject_areas
+
+  validates :minimum_grade_level, numericality: { in: 0..12 }
+  validates :maximum_grade_level, numericality: { in: 0..12 }
+
+  KINDERGARTEN_DISPLAY_STRING = 'K'
+  KINDERGARTEN_DATABASE_INTEGER = 0
+
```
brendanshean, line 26: Good suggestion on this pattern @happythenewsad.

services/QuillLMS/db/migrate/20221109181742_create_teacher_info.rb
```diff
@@ -0,0 +1,13 @@
+# frozen_string_literal: true
+
+class CreateTeacherInfo < ActiveRecord::Migration[6.1]
+  def change
+    create_table :teacher_infos do |t|
+      t.integer :minimum_grade_level
+      t.integer :maximum_grade_level
+      t.references :teacher, foreign_key: { to_table: :users }
```
brendanshean: Can teacher ever null for a valid saved object?

If not, consider validates presence for them and making the DB fields non-null.

(basically a [comment](https://github.com/empirical-org/Empirical-Core/pull/9866/files#r1001850632) Dan wrote on one of my recent PRs)

services/QuillLMS/spec/models/teacher_info_spec.rb
```diff
@@ -0,0 +1,52 @@
...
+
+describe TeacherInfo, type: :model, redis: true do
+  it { should have_many(:subject_areas) }
+  it { should have_many(:teacher_info_subject_areas) }
+  it { should belong_to(:teacher) }
+
+  it {should validate_numericality_of(:minimum_grade_level)}
+  it {should validate_numericality_of(:maximum_grade_level)}
+
+  describe 'minimum_grade_level=' do
```
brendanshean, line 30: I definitely did a double take on these specs, but I think they communicate well the subtlety of the approach.

services/QuillLMS/app/controllers/teacher_infos_controller.rb
```diff
@@ -0,0 +1,36 @@
...
+    end
+
+    if teacher_info_params[:subject_areas]
+      teacher_info.teacher_info_subject_areas.destroy_all
+      subject_areas = SubjectArea.where(id: teacher_info_params[:subject_areas])
+      teacher_info.subject_areas.push(subject_areas)
```
brendanshean: I don't really use `push` for object creation because of transparency, but instead opt for create! calls.  I would be curious to hear others thoughts on this.

brendanshean (approved):
While there are several comments on this, most are ones you can take or leave. The overall the structure looks great.  Nice work on setting up the join table stuff and the conversion of 'K' -> 0

services/QuillLMS/spec/controllers/teacher_infos_controller_spec.rb
brendanshean: NB.  I opted for `before { teacher_info.subject_areas.push(subject_area1) }`
instead of  `let!(:teacher_info_subject_area) { create(:teacher_info_subject_area, teacher_info: teacher_info, subject_area: subject_area1) }` because `teacher_info_subject_area` isn't used in the test anywhere.

services/QuillLMS/config/routes.rb
emilia-friedberg: I had that initially, but defining it that way requires the path to be `/teacher_infos/:teacher_info_id`, which isn't ideal for our use case (this action is getting called from the my account page). I can update it to pass that id to the frontend in order to send it back for this call if you think it's important we stick with the standard CRUD format here.

brendanshean: That makes sense.  Feel free to ignore my comment then.

services/QuillLMS/app/controllers/teacher_infos_controller.rb
emilia-friedberg: I did make that change because what you said makes sense to me, but I'm leaving the comment open for the sake of general discussion.

happythenewsad: Note: `push` doesn't necessarily create objects. If the pushed objects already exist, and there is no join table, `push` has the effect of updating the foreign key value. 

Now, a Socratic dialogue:

Against: `ActiveRecord#push` does a lot more than `Array#push`, which can confound future debugging/diagnosing (what Brendan said).

For: This is a valid syntax which, while not used in our codebase, is prevalent in others. We should be permissive of introducing new coding styles. If we don't, our codebase can never stylistically improve. 

IMO `Against` wins, but it's debatable.

happythenewsad (approved):
LGTM!

services/QuillLMS/db/migrate/20221109181742_create_teacher_info.rb
```diff
@@ -0,0 +1,13 @@
+# frozen_string_literal: true
+
+class CreateTeacherInfo < ActiveRecord::Migration[6.1]
+  def change
+    create_table :teacher_infos do |t|
+      t.integer :minimum_grade_level
+      t.integer :maximum_grade_level
+      t.references :user, null: false, presence: true
```
brendanshean: Do you mean `foreign_key: true` here instead of presence (which is for model files).

Also, I noticed a uniqueness validation in the model file.  To avoid [race conditions](https://thoughtbot.com/blog/validation-database-constraint-or-both), I would add a uniqueness constraint in the database too:

`add_index : teacher_infos, :user_id, unique: true`

services/QuillLMS/app/controllers/teacher_infos_controller.rb
```diff
@@ -0,0 +1,50 @@
...
+
+  def create
+    teacher_info = TeacherInfo.create!(
+      user: current_user,
+      minimum_grade_level: minimum_grade_level,
+      maximum_grade_level: maximum_grade_level
+    )
+
+    subject_areas = SubjectArea.where(id: subject_area_ids)
+    teacher_info.subject_areas.push(subject_areas)
```
brendanshean, line 13: I forgot to add the note about `push` vs `create!`.  Your analogous implementation in `update` looks good.

services/QuillLMS/db/migrate/20221109182145_create_teacher_info_subject_area.rb
```diff
@@ -0,0 +1,12 @@
+# frozen_string_literal: true
+
+class CreateTeacherInfoSubjectArea < ActiveRecord::Migration[6.1]
+  def change
+    create_table :teacher_info_subject_areas do |t|
+      t.references :teacher_info, null: false
```
brendanshean: I would add `foreign_key: true` on both of these references.

services/QuillLMS/app/controllers/teacher_infos_controller.rb
brendanshean: Just for clarification, my point about transparency had to do with silent failure (i.e. there's no corresponding `push!`)  If a push fails for whatever reason, I don't think we would know about it.

services/QuillLMS/db/migrate/20221109181742_create_teacher_info.rb
emilia-friedberg: ooh, I was trying to write `unique: true`! I'm surprised the migration ran with no issues for that.

PR #9970, opened 2022-11-15
Activity Session Filtering/ CSV download

eadams17:
## WHAT
add a CSV download button for Evidence activity sessions

## WHY
this will make it easier for Curriculum to score sessions

## HOW
add a new API for fetching feedback history data for CSV download and add CSV download button

### Screenshots
Screen Shot 2022-11-15 at 4 02 08 PM

### Notion Card Links
https://www.notion.so/quill/Activity-Session-Filtering-Download-search-results-by-CSV-fb781c4a08ae4b26975b0d384bca189b

PR Checklist | Your Answer
------------ | -------------


services/QuillLMS/client/app/bundles/Staff/helpers/evidence/routingHelpers.ts
```diff
@@ -36,6 +36,15 @@ export function getActivitySessionsUrl({ activityId, pageNumber, startDate, endD
   return url;
 }
 
+export function getActivitySessionsCSVUrl({ activityId, startDate, endDate, filterType, responsesForScoring}) {
+  let url = `session_data_for_csv?&activity_id=${activityId}`;
+  url = startDate ? url + `&start_date=${startDate}` : url;
+  url = endDate ? url + `&end_date=${endDate}` : url;
+  url = filterType ? url + `&filter_type=${filterType}` : url;
+  url = responsesForScoring ? url + '&responses_for_scoring=true' : url;
```
emilia-friedberg: Small but a slightly more succinct way to do this would be

```url += startDate ? `&start_date=${startDate}` : ''```

etc

**Review by emilia-friedberg — APPROVED** [review]

File: services/QuillLMS/client/app/bundles/Staff/helpers/evidence/miscHelpers.tsx [hunk 2]
```diff
@@ -411,3 +419,57 @@ export const activitySessionIndexResponseHeaders = [
...
+  }
+  return <p className="sub-optimal-attempt">{attemptCount}</p>
+}
+
+export function formatSessionsData(activityId, activitySessions: ActivitySessionInterface[]) {
+  return activitySessions.map(session => {
+    const { start_date, session_uid, because_attempts, because_optimal, but_attempts, but_optimal, so_attempts, so_optimal, complete } = session;
+    const dateObject = new Date(start_date);
+    const date = moment(dateObject).format("MM/DD/YY");
+    const time = moment(dateObject).format("hh:mm a");
```
emilia-friedberg: I think you can just do `moment(dateObject).format("MM/DD/YY hh:mm a")` here unless you need them to be two separate variables (it looks like you don't, below).

services/QuillLMS/client/app/bundles/Staff/components/evidence/activityStats/activityStats.tsx
```diff
@@ -80,7 +78,7 @@ const ActivityStats: React.FC<RouteComponentProps<ActivityRouteProps>> = ({ hist
 
 
   function handleFilterClick(e: React.SyntheticEvent, passedVersionOption?: DropdownObjectInterface) {
-    handlePageFilterClick({ startDate, endDate, versionOption: passedVersionOption || versionOption, setStartDate, setEndDate, setPageNumber: null, responsesForScoring, setResponsesForScoringForQuery, storageKey: ACTIVITY_STATS,  });
+    handlePageFilterClick({ startDate, endDate, versionOption: passedVersionOption || versionOption, setStartDate, setEndDate, setPageNumber: null, storageKey: ACTIVITY_STATS,  });
```
emilia-friedberg: Might break this onto multiple lines for readability.

emilia-friedberg (approved):
Some minor notes but the frontend LGTM! Thanks for all the tests on that.

services/QuillLMS/spec/models/feedback_history_spec.rb
```diff
@@ -640,5 +643,41 @@
         expect(@feedback_history.most_recent_rating).to eq true
       end
     end
+
+    context '#session_data_for_csv' do
+      it 'should take the query from #session_data_for_csv and return a shaped payload' do
```
brendanshean: Cool.  I haven't seen this `RSpec::Support::ObjectFormatter.default_instance.max_formatted_output_length ` setting before. 

One to keep in mind with changing defaults in RSpec is you want to make sure that you set them back to the original.  Otherwise, they can carry on into other specs.  (this is similar to why we use `stub_const` in before blocks).  

```
around do |example|
  default_length = RSpec::Support::ObjectFormatter.default_instance.max_formatted_output_length
  RSpec::Support::ObjectFormatter.default_instance.max_formatted_output_length = 10000
  example.run
  RSpec::Support::ObjectFormatter.default_instance.max_formatted_output_length = default_length
end

it 'should take the query from #session_data_for_csv and return a shaped payload' do
...
```

services/QuillLMS/spec/models/feedback_history_spec.rb
```diff
@@ -640,5 +643,41 @@
...
+        responses = FeedbackHistory.session_data_for_csv
+        RSpec::Support::ObjectFormatter.default_instance.max_formatted_output_length = 10000
```
brendanshean, line 661: remove `RSpec::Support::ObjectFormatter.default_instance.max_formatted_output_length = 10000`

services/QuillLMS/spec/models/feedback_history_spec.rb
```diff
@@ -640,5 +643,41 @@
...
+          expect(response[:response]).to eq(@histories[i].entry)
+          expect(response[:feedback]).to eq(@histories[i].feedback_text)
+          expect(response[:feedback_type]).to eq(@histories[i].feedback_type)
+          expect(response[:datetime]).to be_within(1.second).of @histories[i].time
+        end
+      end
+
+      it 'should take the query from #session_data_for_csv and return a shaped payload with records qualifying for scoring' do
+        responses = FeedbackHistory.session_data_for_csv(responses_for_scoring: true)
+        RSpec::Support::ObjectFormatter.default_instance.max_formatted_output_length = 10000
```
brendanshean: remove `RSpec::Support::ObjectFormatter.default_instance.max_formatted_output_length = 10000`

brendanshean (approved):
Nice work on figuring out that query!  I also appreciate the thorough testing here.

services/QuillLMS/spec/controllers/api/v1/session_feedback_histories_controller_spec.rb
```diff
@@ -215,4 +215,66 @@
       expect(response.body.include?("The resource you were looking for does not exist")).to be
     end
   end
+
+  context "session_data_for_csv" do
+    before do
+      user = create(:user)
+      @activity = Evidence::Activity.create!(notes: 'Title_1', title: 'Title 1', parent_activity_id: 1, target_level: 1)
```
dandrabik: Small note, we should convert these to `let` statements to fit `rspec` conventions, e.g.

```ruby
let(:user) { create(:user)}
# or if you need it created before the method is called use let!
let!(:user) { create(:user)}

File: services/QuillLMS/app/models/feedback_history.rb [hunk 8]
```diff
@@ -358,4 +362,45 @@ def self.serialize_detail_by_activity_session(feedback_session_uid)
     output.symbolize_keys
   end
   # rubocop:enable Metrics/CyclomaticComplexity
+
+  # rubocop:disable Metrics/CyclomaticComplexity
+  def self.session_data_for_csv(activity_id: nil, start_date: nil, end_date: nil, filter_type: nil, responses_for_scoring: false)
+    sub_query = select(
```
dandrabik: this method is worth breaking up a bit. You can make this sub_query it's own method, e.g.

```ruby
session_uids = session_data_uids(activity_id: activity_id, filter_type: filter_type, responses_for_scoring: responses_for_scoring)
...
query = query.where(feedback_session_uid: session_uids) if responses_for_scoring || filter_type


...
def self.session_data_uids(activity_id: nil, filter_type: nil, responses_for_scoring: false)
  ...

dandrabik (approved):
Overall this looks good. I left a note to refactor the tests a bit to use `let` and a small note to break up a large methods.

I mentioned in our discussion that we might want to do these filter queries differently, but looking at this and the code around it, I think that could be a separate, larger refactor. If this is working, let's 🚢 ! Nice work on this!

services/QuillLMS/client/app/bundles/Staff/helpers/evidence/miscHelpers.tsx
eadams17: good catch, I'll update

services/QuillLMS/spec/models/feedback_history_spec.rb
eadams17: @brendanshean thanks for the explanation! I actually took this from the existing `#serialize_list_by_activity_session` spec, but I'm wondering what a good default length would be? I'm looking through the docs, but not finding anything eluding to what the default is under the hood

services/QuillLMS/app/models/feedback_history.rb
eadams17: agreed, I'll break this out into two methods

PR #9972, opened 2022-11-16
Model and Migration files for Evidence Activity/Prompt Health Dashboard

cissyxyu:
## WHAT
Model and migration files in preparation for building the Evidence Activity/Prompt Health dashboard for Quill admin.

## WHY
We need to create these two tables, `evidence_activity_healths` and `evidence_prompt_healths`, to cache the data that we will generate nightly. That data will then be displayed to Curriculum staff whenever they access the Evidence Activity Health and Evidence Prompt Health dashboards.

## HOW
Create the model and migration files for these two tables. Make sure that they are related to each other through a foreign key table, so each Prompt Health entry belongs to an Activity Health entry.

### Screenshots

### Notion Card Links
https://www.notion.so/quill/1-Activity-Health-Dashboard-Back-End-7be2570c0f2341be975df35a6ec1cd5a 
https://www.notion.so/quill/3-Prompt-Health-Dashboard-Back-End-a6d3499190db431e923bb5525a632e96

PR Checklist | Your Answer
------------ | -------------


services/QuillLMS/engines/evidence/app/models/evidence/activity_health.rb
```diff
@@ -0,0 +1,43 @@
+# frozen_string_literal: true
+
+module Evidence
+  class ActivityHealth < ApplicationRecord
+    ALLOWED_TOOLS = %w(connect grammar)
+    FLAGS = [
```
happythenewsad: Please source this data from:
https://github.com/empirical-org/Empirical-Core/blob/develop/services/QuillLMS/app/models/concerns/flags.rb

Otherwise, when `flags.rb` is updated with new entries, this class will fail validation.

services/QuillLMS/db/migrate/20221110072938_create_evidence_activity_healths.evidence.rb
```diff
@@ -0,0 +1,24 @@
...
+      t.string               :flag
+      t.integer              :activity_id
+      t.integer              :version
+      t.integer              :version_plays
+      t.integer              :total_plays
+      t.integer              :completion_rate
+      t.float                :because_avg_attempts
+      t.float                :but_avg_attempts
+      t.float                :so_avg_attempts
+      t.integer              :avg_completion_time
```
happythenewsad, line 19: Most of these fields - and in EvidencePromptHealths - should be null false. Your validations like quite thorough, but the standard in this codebase is to enforce `null: false` if possible.

happythenewsad (changes requested):
I like your strategy of rolling the schema changes out first. Getting close...

happythenewsad (approved):
Thanks for making those changes

services/QuillLMS/db/migrate/20221110072939_create_evidence_prompt_healths.evidence.rb
```diff
@@ -0,0 +1,32 @@
...
+      t.integer     :percent_opinion
+      t.integer     :percent_grammar
+      t.integer     :percent_spelling
+      t.integer     :avg_time_spent_per_prompt
+
+      t.references  :evidence_activity_health
+
+      t.timestamps
+    end
+    add_foreign_key :evidence_prompt_healths, :evidence_activity_healths, column: :evidence_activity_health_id, on_delete: :cascade
```
brendanshean, line 30: Interesting, I've always just used `dependent: :destroy` on the model.  What is the reasoning for also including the `on_delete: :cascade`?

brendanshean (approved):
Looks good to me.  I also like separating the schema and bare model files first.

services/QuillLMS/db/migrate/20221110072939_create_evidence_prompt_healths.evidence.rb
cissyxyu: Hmm I've always just done it on both the migration and the model, I'm not sure if there's a plus to doing it on both honestly. From [this answer here](https://stackoverflow.com/questions/12556614/rails-delete-cascade-vs-dependent-destroy) it seems like there might be performance benefits to `on_delete: cascade`?

PR #9989, opened 2022-11-21
Featured activity packs/ previously assigned activities

eadams17:
## WHAT
flag activities in activity packs during the assignment flow that have been previously assigned by teachers

## WHY
this was a source of confusion for teachers in terms of how activities fit in other assignment flows

## HOW
add backend logic to query for units that contain activities that are in the selected activity pack and display this data via a tooltip on the frontend

### Screenshots
Screen Shot 2022-11-21 at 1 52 04 PM
Screen Shot 2022-11-21 at 1 52 13 PM

### Notion Card Links
https://www.notion.so/quill/Featured-Activity-Pack-Page-Mark-Previously-Assigned-Activities-a3a6fab71bdf4445923d295ca7e60c08

PR Checklist | Your Answer
------------ | -------------


services/QuillLMS/app/models/unit_template.rb
```diff
@@ -107,12 +107,12 @@ def self.user_scope(user_flag)
...
+  def get_cached_serialized_unit_template(flag=nil, current_user=nil)
     cache_expiration_time = 600
     cached = $redis.get("unit_template_id:#{id}_serialized")
     serialized_unit_template = cached.nil? || cached&.blank? ? nil : JSON.parse(cached)
     unless serialized_unit_template
-      serializable_unit_template = UnitTemplatePseudoSerializer.new(self, flag)
+      serializable_unit_template = UnitTemplatePseudoSerializer.new(self, flag, current_user)
       serialized_unit_template = serializable_unit_template.data
       $redis.set("unit_template_id:#{id}_serialized", serialized_unit_template.to_json, {ex: cache_expiration_time})
     end
```
emilia-friedberg: I'm pretty sure this setup is going to cache, for all users, the unit template data for any given user (because the cache key itself doesn't care about the user). My inclination is to say that we should probably split that data out into a separate query and maybe a separate controller action, so we can leave the caching logic alone, but I'd defer to a backend dev on that front.

services/QuillLMS/app/services/unit_template_pseudo_serializer.rb
```diff
@@ -129,6 +133,42 @@ def activities
     activity_hashes.uniq { |a| a[:id] }
   end
 
+  def student_counts_for_previously_assigned_activity(unit, classrooms)
+    classrooms.map do |classroom|
+      classroom_unit = unit.classroom_units.where(classroom_id: classroom[:id]).first
```
emilia-friedberg: `.where... .first` can just be `.find_by`.

services/QuillLMS/app/services/unit_template_pseudo_serializer.rb
```diff
@@ -129,6 +133,42 @@ def activities
...
+      {
+        total_students: classroom.students&.length,
+        assigned_students: classroom_unit.assigned_student_ids&.length
```
emilia-friedberg: Might want to change these keys to `total_student_count` and `assigned_student_count` for clarity.

services/QuillLMS/client/app/bundles/Teacher/components/assignment_flow/create_unit/stage2/review_activities.jsx
```diff
@@ -74,10 +82,26 @@ export default class ReviewActivities extends React.Component {
...
+    }
+  }
+
+  getProfileInfo = (id) => {
+    requestGet(`/teachers/unit_templates/profile_info?id=${id}`, (response) => {
+      const { data } = response
+      const { previously_assigned_activity_data } = data
+      this.setState({ previouslyAssignedActivityData: previously_assigned_activity_data })
+    })
+  }
```
emilia-friedberg: This is an other example of why it probably makes sense to split those queries/actions up - there is a lot of data contained in that query that we don't need for this page.

services/QuillLMS/client/app/bundles/Teacher/components/assignment_flow/unit_templates_manager/unit_template_profile/unit_template_profile_activity_table.tsx
```diff
@@ -61,21 +62,35 @@ export const UnitTemplateProfileActivityTable = ({ data }) => {
...
+        id: 'previouslyAssigned',
+        Cell: ({row}) => {
+          const { previously_assigned_activity_data } = data
+          const { original } = row;
+          const { id } = original;
+          if(previously_assigned_activity_data[id]) {
+            return(<PreviouslyAssignedTooltip previouslyAssignedActivityData={previously_assigned_activity_data[id]} />)
+          }
+          return <span />
+        },
```
emilia-friedberg: I don't know how much it really matters, but I might move this logic into the `activitiesFormattedForTable` function, just because it seems like the pattern is that the conditional logic goes in there and not in the Cell render function.

emilia-friedberg (changes requested):
The frontend for this looks great for the most part - I'm requesting changes because I think we're going to need to revamp the backend for both caching purposes and also general separation of concerns.

services/QuillLMS/app/models/unit_template.rb
eadams17: ahh ok, yeah I didn't think about the caching aspect. I agree, I'll pull the logic out into a separate query and controller action

services/QuillLMS/app/services/unit_template_pseudo_serializer.rb
brendanshean: There's actually a [subtle](https://stackoverflow.com/a/56594507) difference between `find_by` and `where(...).first`.

services/QuillLMS/app/models/unit_template.rb
```diff
@@ -140,6 +140,43 @@ def self.delete_all_caches
...
+  end
+
+  def self.previously_assigned_activity_data(activity_ids=[], current_user=nil)
+    results = {}
+    activity_ids&.map do |id|
+      units = Unit.joins(
+        " JOIN classroom_units ON classroom_units.unit_id = units.id
+        JOIN unit_activities ON unit_activities.unit_id = units.id
+        "
+        )
```
emilia-friedberg: I think you can just do `Unit.joins(:classroom_units, :unit_activities)`

services/QuillLMS/app/models/unit_template.rb
```diff
@@ -140,6 +140,43 @@ def self.delete_all_caches
     end
   end
 
+  def self.student_counts_for_previously_assigned_activity(unit=nil, classrooms=[])
+    classrooms&.map do |classroom|
```
emilia-friedberg: I don't think you need the safe navigation operator since you're explicitly asking for `[]` if `classrooms` is falsy, right?

services/QuillLMS/app/models/unit_template.rb
```diff
@@ -140,6 +140,43 @@ def self.delete_all_caches
...
+      {
+        total_student_count: classroom.students&.length,
+        assigned_student_count: classroom_unit&.assigned_student_ids&.length
+      }
+    end
+  end
+
+  def self.previously_assigned_activity_data(activity_ids=[], current_user=nil)
+    results = {}
+    activity_ids&.map do |id|
```
emilia-friedberg: See above comment about safe navigation.

services/QuillLMS/client/app/bundles/Teacher/components/assignment_flow/create_unit/stage2/review_activities.jsx
```diff
@@ -74,10 +82,30 @@ export default class ReviewActivities extends React.Component {
 
     this.state = {
       snackbarVisible: false,
-      erroredActivityIds: []
+      erroredActivityIds: [],
+      previouslyAssignedActivityData: null
+    }
+  }
+
+  componentDidUpdate() {
```
emilia-friedberg, line 91: Are we calling this here because `activities` isn't populated when the page loads? Otherwise, I'd just call this in the constructor.

emilia-friedberg (approved):
Small suggestions but overall LGTM.

services/QuillLMS/app/services/unit_template_pseudo_serializer.rb
emilia-friedberg: That's good to flag. In this instance, where there is only going to be one classroom unit per intersection of classroom and unit, it shouldn't matter, but in other cases it might.

services/QuillLMS/app/models/unit_template.rb
eadams17: yeah, you're right-- thanks for catching that!

services/QuillLMS/client/app/bundles/Teacher/components/assignment_flow/create_unit/stage2/review_activities.jsx
eadams17: yeah, initially there are no activities on page load

services/QuillLMS/spec/models/unit_template_spec.rb
```diff
@@ -204,6 +204,91 @@
...
+        {
+          total_student_count: classroom_two.students.length,
+          assigned_student_count: classroom_unit_two.assigned_student_ids.length
+        }
+      ])
+    end
+  end
+
+  describe '#previously_assigned_activity_data' do
+    let!(:current_user) { create(:teacher_with_a_couple_classrooms_with_one_student_each) }
```
brendanshean, line 241: I like to put a subject at the top of specs to describe overall what is being tested: 

So in line 239: `subject { UnitTemplate.previously_assigned_activity_data(activity_ids, current_user) }`

services/QuillLMS/spec/models/unit_template_spec.rb
```diff
@@ -204,6 +204,91 @@
...
+    let!(:activity_five) { create(:activity) }
+    let!(:classroom_unit_one) { create(:classroom_unit, unit: unit_one, classroom: classroom, assigned_student_ids: classroom.student_ids) }
+    let!(:classroom_unit_two) { create(:classroom_unit, unit: unit_two, classroom: classroom) }
+    let!(:unit_activity_one) { create(:unit_activity, unit: classroom_unit_one.unit, activity: activity_one) }
+    let!(:unit_activity_two) { create(:unit_activity, unit: classroom_unit_one.unit, activity: activity_two) }
+    let!(:unit_activity_three) { create(:unit_activity, unit: classroom_unit_two.unit, activity: activity_two) }
+    let!(:unit_activity_four) { create(:unit_activity, unit: classroom_unit_two.unit, activity: activity_three) }
+    let!(:activity_ids) { [activity_one.id, activity_two.id, activity_three.id, activity_four.id, activity_five.id] }
+    let!(:count) { classroom.student_ids.length }
+    let!(:results) { UnitTemplate.previously_assigned_activity_data(activity_ids, current_user) }
```
brendanshean: Then you could rewrite results in a way that gets rid of some of the duplication in the specs below:

`let!(:results) { subject[:previously_assigned_activity_data] }`

brendanshean (approved):
Nice work on this.  One very minor comment on refactoring specs that you can take or leave.

services/QuillLMS/spec/models/unit_template_spec.rb
```diff
@@ -204,6 +204,91 @@
...
+
+    it 'should return the expected results for the first activity' do
```
brendanshean, line 260: With refactoring from above you now get:
```
 it 'should return the expected results for the first activity' do
      expect(results[activity_one.id].length).to eq(1)
      expect(results[activity_one.id][0][:name]).to eq(unit_one.name)
      expect(results[activity_one.id][0][:assigned_date]).to be_within(1.second).of(unit_one.created_at)
      expect(results[activity_one.id][0][:classrooms]).to eq([classroom.name])
      expect(results[activity_one.id][0][:students]).to eq([{ assigned_student_count: count, total_student_count: count }])
    end
```

eadams17: oh cool, didn't know about that-- thanks for the suggestion! I'll refactor

PR #9993, opened 2022-11-22
Nightly job to calculate and cache Evidence Activity Healths

cissyxyu:
## WHAT
Create a nightly job that calculates the relevant analytics data for all Evidence Activities and saves them in the table `evidence_activity_healths` for easy future access.

## WHY
This data is time intensive to calculate, so it makes more sense to run a job at night to calculate the data for each activity, then roll it up into a table. This way admin won't have delays when using the dashboard.

## HOW
1. Create sidekiq workers that run nightly to populate the table.
2. Change the `ActivityFeedbackHistory` and `PromptFeedbackHistory` queries to run with a `version` option, because we want to limit the data to only the latest version of an activity.
3. Make the `SerializeEvidenceActivityHealth` serializer object, which runs all the queries to get the most recent version data and then saves it as a record in the `evidence_activity_healths` table.
4. A whole bunch of tests to make sure the data is calculated correctly.
 
 l### Screenshots

### Notion Card Links
https://www.notion.so/quill/1-Activity-Health-Dashboard-Back-End-7be2570c0f2341be975df35a6ec1cd5a

PR Checklist | Your Answer
------------ | -------------


services/QuillLMS/app/lib/utils/numeric.rb
```diff
@@ -20,4 +20,8 @@ def self.to_human_string(numeric)
   def self.seconds_to_human_readable_time(seconds)
     format("%<minutes>02d:%<seconds>02d", minutes: seconds / 60 % 60, seconds: seconds % 60)
   end
+
+  def self.human_readable_time_to_seconds(time)
```
happythenewsad, line 25: Please add a comment that describes the type of the param. If future users try to pass a param of class `Time`, it will result in a runtime exception.

services/QuillLMS/app/workers/populate_evidence_activity_healths_worker.rb
```diff
@@ -0,0 +1,18 @@
...
+
+  INTERVAL = 180 # 3 minutes
+
+  def perform
+    Evidence::ActivityHealth.destroy_all
+    # execute this command to start primary key indices back from 1
+    ActiveRecord::Base.connection.execute("TRUNCATE evidence_activity_healths RESTART IDENTITY CASCADE")
+
+    # spread these, to cut down on DB resource contention.
+    Evidence::Activity.all.each.with_index do |activity, index|
```
happythenewsad: You can make this more efficient by using `pluck` to get the ids.

services/QuillLMS/app/workers/populate_evidence_activity_healths_worker.rb
```diff
@@ -0,0 +1,18 @@
+# frozen_string_literal: true
+
+class PopulateEvidenceActivityHealthsWorker
```
happythenewsad: IMO having two files with only one letter - `s` - difference may be confusing. What about naming this something like `PopulateEvidenceActivityHealthAggregateWorker`?

happythenewsad (changes requested):
Getting close. Nice job on the unit tests

services/QuillLMS/app/lib/utils/numeric.rb
anathomical: Yeah, I'd love to have a note here on what shape we expect values passed in here to look like.  (I presume it's supposed to be something like `HH:MM`?)

services/QuillLMS/app/queries/activity_feedback_history.rb
```diff
@@ -1,37 +1,34 @@
...
-        ) inner_query
-      SQL
-    ).to_a
-
-    return { average_time_spent: 0, average_completion_rate: 0 } if activity_sessions_agg.empty?
+  def self.activity_stats_query(activity_id:, activity_version: nil)
+
+    inner_query = ActivitySession.unscoped
+      .joins("INNER JOIN feedback_sessions ON feedback_sessions.activity_session_uid = activity_sessions.uid")
+      .joins(:feedback_histories)
```
anathomical: It's not necessarily better, but I think you can just do:
```ruby
.joins(:feedback_session)
.joins(:feedback_history)
```
Because the `has_many :feedback_history` definition in `FeedbackSession` already specifies that it should use `uid` as the primary key, so that mapping should work with Rails magic.

anathomical: Additional note: there is a long-term plan to allow `ActivitySession` to be detached from `FeedbackHistories` for privacy reasons.  When a user gets deleted, the intent is to remove all of their `FeedbackSession` records to anonymize the `FeedbackHistory` records (basically, it'll no longer be possible to tie them back to a specific user.  This code, as written, will start to drop that data from the calculations.

This may not be a problem since we expect that to happen for older users who are no longer active, and thus won't have recent sessions to use for health metrics, but it's probably worth flagging as a risk in case we do care about the ability to keep using those sessions for metrics.

services/QuillLMS/app/services/serialize_evidence_activity_health.rb
```diff
@@ -0,0 +1,45 @@
...
+      because_final_optimal: percent_final_optimal_for_conjunction(FeedbackHistory::BECAUSE),
+      but_final_optimal: percent_final_optimal_for_conjunction(FeedbackHistory::BUT),
+      so_final_optimal: percent_final_optimal_for_conjunction(FeedbackHistory::SO),
+      avg_completion_time: activity_feedback_history[:average_time_spent] ? Utils::Numeric.human_readable_time_to_seconds(activity_feedback_history[:average_time_spent]) : nil,
+    }
+  end
+
+  def prompt_data
+    # TODO: fill this in to return serialized prompt health objects
+    []
```
anathomical, line 27: You left a TODO here.  I assume that this is intended to be part of some follow-on PR?  Might be worth deleting this function in the interim if that might be a little bit out.

services/QuillLMS/app/workers/populate_evidence_activity_healths_worker.rb
```diff
@@ -0,0 +1,18 @@
...
+  INTERVAL = 180 # 3 minutes
+
+  def perform
+    Evidence::ActivityHealth.destroy_all
+    # execute this command to start primary key indices back from 1
+    ActiveRecord::Base.connection.execute("TRUNCATE evidence_activity_healths RESTART IDENTITY CASCADE")
+
+    # spread these, to cut down on DB resource contention.
+    Evidence::Activity.all.each.with_index do |activity, index|
+      PopulateEvidenceActivityHealthWorker.perform_in(index * INTERVAL, activity.id)
```
anathomical: Just a note on this in case it hasn't been considered: if something goes wrong with the `PopulateEvidenceActivityHealthWorker` calls, we could end up truncating the old health data and then failing to re-populate it.  Given the nature of us running this every night, that might not be a big problem, but it's worth confirming that this makes sense to us.

anathomical (approved):
A couple of notes, but this generally looks good to me!

services/QuillLMS/app/services/serialize_evidence_activity_health.rb
cissyxyu: Yeah, it's a note for my next PR. I'll leave it in for now since it should be done in a week or so.

services/QuillLMS/app/queries/activity_feedback_history.rb
cissyxyu: Is the plan to only do this for deleted users? From my understanding that's fairly rare for us to delete users right? Or is the plan from now on to regularly delete users who are inactive for a certain amount of time? If that's the case then I might have to rethink how to aggregate session data.

services/QuillLMS/app/workers/populate_evidence_activity_healths_worker.rb
cissyxyu: I'm going to say that's okay for now, since this data is updated frequently enough that we could fix the issue quickly. Also the individual stats are accessible from other places on the admin panel, so if this table is somehow truncated accidentally admin will still be able to view basic stats elsewhere.

happythenewsad (approved):

PR #10001, opened 2022-11-29
Update clever-omniauth gem

brendanshean:
## WHAT
Move clever-omniauth gem to a forked repo.

## WHY
Clever is [deprecating](https://dev.clever.com/docs/api-v1-deprecation) API v1 in December 2022.

## HOW
Point the source of the git repo to the fork's url.

### Screenshots

### Notion Card Links
https://www.notion.so/quill/Update-omniauth-clever-with-a-fork-25d408640837417eba4b86d3816e9b46

PR Checklist | Your Answer
------------ | -------------
Have you added and/or updated tests? |  No.  Configuration change.

### DIFF
--- a/services/QuillLMS/Gemfile.lock
+++ b/services/QuillLMS/Gemfile.lock
@@ -1,3 +1,10 @@
+GIT
+  remote: https://github.com/cm-studios/omniauth-clever
+  revision: 9b32b15a3daca46aee1011e18e578fd61ec9c23e
+  specs:
+    omniauth-clever (1.1.1)
+      omniauth-oauth2 (~> 1.2)
+
 PATH
   remote: engines/evidence
   specs:
@@ -439,8 +446,6 @@ GEM
     omniauth (1.9.0)
       hashie (>= 3.4.6, < 3.7.0)
       rack (>= 1.6.2, < 3)
-    omniauth-clever (1.2.2)
-      omniauth-oauth2 (>= 1.1, <= 1.5)
     omniauth-google-oauth2 (0.6.0)
       jwt (>= 2.0)
       omniauth (>= 1.1.1)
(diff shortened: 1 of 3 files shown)


services/QuillLMS/Gemfile.lock
```diff
@@ -1,3 +1,10 @@
+GIT
+  remote: https://github.com/cm-studios/omniauth-clever
+  revision: 9b32b15a3daca46aee1011e18e578fd61ec9c23e
+  specs:
+    omniauth-clever (1.1.1)
```
dandrabik: So this fork is 8 commits ahead and 14 commits behind the gem we were on (this would be a change to a fork and revision to from `1.2.2` -> `1.1.1`). 

There seems to be some notable changes in urls and auth methods:
https://github.com/cm-studios/omniauth-clever/compare/master...Clever:omniauth-clever:master#diff-cd2bee9e4d8f887976e3c4942853cfbe6de360f219dcd7f8c53c1f0b3f835228

This might be worth looking at. We mind need our own fork where we incorporate this forks changes and update to `1.2.2` (there may also be a fork that already did that)

dandrabik (commented):
Left a comment about a a possible concern with the version downgrade in this fork.

services/QuillLMS/Gemfile.lock
brendanshean: There is https://github.com/Clever/omniauth-clever/compare/master...Pioneer-Valley-Books:omniauth-clever:master

dandrabik (approved):
This looks like a better gem candidate. If you ignore the added feature around `get_user_info` (which our code wouldn't use), this changeset is basically:

```ruby
# before
access_token.get('/me').parsed
# after
access_token.get('/v2.1/me').parsed
```

Looks good enough to test. If we hit issues with the other code that we aren't using here around `raw_user_info`. We could also create our own fork with just that one line change.

happythenewsad (approved):

PR #10007, opened 2022-12-01
Nightly data calculation and caching for Evidence Prompt Health

cissyxyu:
## WHAT
Similar to the logic for calculating Activity Health, we want to serialize and calculate a bunch of statistics for all Evidence Prompts and cache them in a table every night. This PR introduces a serializer class, `SerializeEvidencePromptHealth`, which aggregates all of the stats and saves them into the EvidencePromptHealths table.

## WHY
Once this logic is in place, we will be calculating and caching all the data Quill admin need for this dashboard. Then we can work on displaying it in the front end.

## HOW
Using the serializer class, pull all the necessary data from the various query classes, format them, and save them into the table.

### Screenshots

### Notion Card Links
https://www.notion.so/quill/3-Prompt-Health-Dashboard-Back-End-a6d3499190db431e923bb5525a632e96

PR Checklist | Your Answer
------------ | -------------


services/QuillLMS/app/queries/prompt_feedback_history.rb
```diff
@@ -85,7 +85,7 @@ def avg_confidence
     end
 
     def avg_time_spent
-      payload['avg_time_spent'] ? Utils::Numeric.seconds_to_human_readable_time(payload['avg_time_spent']) : 0
+      payload['avg_time_spent'] ? Utils::Numeric.seconds_to_human_readable_time(payload['avg_time_spent']) : nil
```
brendanshean: I find that `nil` in a ternary can be rewritten as a conjunction:
`payload['avg_time_spent'] && Utils::Numeric.seconds_to_human_readable_time(payload['avg_time_spent'])`

services/QuillLMS/app/services/serialize_evidence_activity_health.rb
```diff
@@ -2,8 +2,9 @@
 
 class SerializeEvidenceActivityHealth
```
brendanshean, line 5: Let's use  `attr_reader :activity, :prompt_feedback_history` for the variables initialized in `initialize`.  There are few reasons for doing this:

1) It guards against misspellings.  For example,  calling `@promp_feedback_history` returns `nil` whereas a calling `promp_feedback_history no syntax error will happen since ruby just thinks it's a new instance variable.  `promp_feedback_history` will return `NameError: undefined local variable or method `

2) While it doesn't explictly prevent all overwriting, specifically via the instance variable itself, the attr_reader prevent accidental overwrites directly.

3) it's one less character to type for each variable name.

services/QuillLMS/app/services/serialize_evidence_prompt_health.rb
```diff
@@ -0,0 +1,63 @@
+# frozen_string_literal: true
+
+class SerializeEvidencePromptHealth
+
```
brendanshean, line 4: Similar to `serialize_evidence_activity_health.rb` use `attr_reader for :prompt, :prompt_feedback_history` for the variables initialized in `initialize`.

services/QuillLMS/app/workers/populate_aggregated_evidence_activity_healths_worker.rb
```diff
@@ -10,8 +10,9 @@ def perform
     # execute this command to start primary key indices back from 1
     ActiveRecord::Base.connection.execute("TRUNCATE evidence_activity_healths RESTART IDENTITY CASCADE")
 
+    evidence_activities = Evidence::Activity.all.reject { |a| a.flag == Flags::ARCHIVED.to_sym }
```
brendanshean, line 13: I would move to a scope on Evidence::Activity model
```
scope :not_archived, -> { where.not(flag: Flags::ARCHIVED.to_sym) }
```

services/QuillLMS/app/workers/populate_aggregated_evidence_activity_healths_worker.rb
```diff
@@ -10,8 +10,9 @@ def perform
...
     # spread these, to cut down on DB resource contention.
-    Evidence::Activity.pluck(:id).each.with_index do |id, index|
+    evidence_activities.pluck(:id).each.with_index do |id, index|
```
brendanshean, line 15: `Evidence.Activity.not_archived.pluck(:id).each.with_index ...`

services/QuillLMS/app/workers/populate_evidence_activity_health_worker.rb
```diff
@@ -7,13 +7,14 @@ class PopulateEvidenceActivityHealthWorker
   def perform(id)
     @activity = Evidence::Activity.find(id)
 
-    serializer = SerializeEvidenceActivityHealth.new(@activity)
+    prompt_feedback_history = PromptFeedbackHistory.run({activity_id: @activity.id, activity_version: @activity.version})
```
brendanshean: No  need for braces if it's the last parameter in a method.

services/QuillLMS/spec/lib/rule_feedback_history_spec.rb
```diff
@@ -162,6 +162,41 @@ def feedback_factory(&hash_block)
       expect(sql_result.all.length).to eq 1
       expect(sql_result[0].rule_type).to eq 'autoML'
     end
+
+    it 'should filter by activity_version if specified' do
+      # activities
```
brendanshean: The comments in this block feel unnecessary.

brendanshean (commented):
I left some comments.  I will add some more tomorrow morning.

happythenewsad (approved):
Looks good after Brendan's comments

services/QuillLMS/app/workers/populate_aggregated_evidence_activity_healths_worker.rb
cissyxyu: So for this, `flag` is not a variable on the activity model but a method. (Because we store this variable on the parent activity class, not the evidence activity class). 

In my experience, we use scopes in cases where you essentially do an ActiveRecord query on the records in question, filtering by fields like in the example solution you gave me. I'm not sure how to do a scope where I want to filter not by field, but by the result of a certain method (`flag`) on the record?

services/QuillLMS/app/workers/populate_aggregated_evidence_activity_healths_worker.rb
cissyxyu: See above.

services/QuillLMS/app/services/serialize_evidence_activity_health.rb
```diff
@@ -23,23 +25,21 @@ def data
...
     ((prompt_feedback_history[prompt.id]["num_final_attempt_optimal"].to_f / prompt_feedback_history[prompt.id]["session_count"]) * 100).round
   end
 
   private def activity_feedback_history
-    @activity_feedback_history ||= ActivityFeedbackHistory.run({activity_id: @activity.id, activity_version: @activity.version})
-  end
-
-  private def prompt_feedback_history
-    @prompt_feedback_history ||= PromptFeedbackHistory.run({activity_id: @activity.id, activity_version: @activity.version})
+    activity_feedback_history ||= ActivityFeedbackHistory.run({activity_id: activity.id, activity_version: activity.version})
```
brendanshean: You'll want to keep the instance variable in the assignment for this private method:

`@activity_feedback_history ||= ActivityFeedbackHistory.run(activity_id: activity.id, activity_version: activity.version)` otherwise the value won't be memoized.

services/QuillLMS/app/workers/populate_aggregated_evidence_activity_healths_worker.rb
brendanshean: Oh.  My mistake.  Please ignore.

services/QuillLMS/spec/services/serialize_evidence_prompt_health_spec.rb
```diff
@@ -0,0 +1,136 @@
+# frozen_string_literal: true
+
+require 'rails_helper'
+
+describe 'SerializeEvidencePromptHealth' do
+  before do
```
brendanshean, line 7: This particular before block setups a lot of files, but each spec uses the same setup.   I would collapse all of the specs into one spec and just match the whole results hash:

```
describe 'SerializeEvidencePromptHealth' do
  let(:expected_results) do
    {
      prompt_id: @because_prompt1.id,
      activity_short_name: @activity.notes,
      text: @because_prompt1.text,
      current_version: @activity.version,
      version_responses: 10,
      first_attempt_optimal: 25,
      final_attempt_optimal: 75,
      avg_attempts: 2.5,
      confidence: 0.85,
      percent_automl_consecutive_repeated: 0,
      percent_automl: 30,
      percent_plagiarism: 10,
      percent_opinion: 20,
      percent_grammar: 30,
      percent_spelling: 10,
      avg_time_spent_per_prompt: 287
    }
  end
  
  before do
    #keep same
  end

  it 'correct basic data for that prompt' do
    prompt = @because_prompt1
    data = SerializeEvidencePromptHealth.new(prompt, @prompt_feedback_history).data
    expect(data).to eq expected_results
  end
```

If you wanted to check for specific variations on values then that would be appropriate to add more specs.

brendanshean (approved):
A couple more comments but otherwise looks good to go.  Nice work on pulling all of this data together!

anathomical (approved):

PR #10013, opened 2022-12-01
move now() query into ruby

brendanshean:
## WHAT
Move a calculation of `scheduled` out of postgres and into ruby.

## WHY
the presence of now() is preventing the query from being cached.

## HOW
Add a new select column for `ua.publish_date` and then use that value to make the comparison outside the query.  

NB `[string].time_now` without a parameter defaults to `:local`.

### Screenshots

### Notion Card Links

PR Checklist | Your Answer
------------ | -------------

### DIFF
--- a/services/QuillLMS/spec/controllers/teachers/units_controller_spec.rb
+++ b/services/QuillLMS/spec/controllers/teachers/units_controller_spec.rb
@@ -277,7 +277,6 @@ describe Teachers::UnitsController, type: :controller do
         unit_activity.update_columns(publish_date: publish_date)
         response = get :index, params: { report: false }
         parsed_response = JSON.parse(response.body)
-        # this one here
         expect(parsed_response[0]['scheduled']).to eq(false)
       end
     end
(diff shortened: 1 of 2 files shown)


services/QuillLMS/spec/controllers/teachers/units_controller_spec.rb
```diff
@@ -277,6 +277,7 @@
         unit_activity.update_columns(publish_date: publish_date)
         response = get :index, params: { report: false }
         parsed_response = JSON.parse(response.body)
+        # this one here
```
dandrabik: Stray debugging comment here.

dandrabik (approved):
This looks good, thanks for making this improvement. 

The original query is pretty dense and gnarly, so it's hard to say this will definitely fix the issue, but it will be an improvement, especially for multiple calls in a short time frame. 

Thanks for hunting this down! 🚢

happythenewsad (approved):

emilia-friedberg (approved):

PR #10014, opened 2022-12-01
Admin user management flow/ email updates

eadams17:
## WHAT
add various new email templating and workers needed for the admin user management flow updates

## WHY
we will need to send this emails as part of the new signup flow for admin users

## HOW
add a series of new templates, works and methods on the `User` model to execute these emails. Please note: there is some shared content between these emails, but I opted to keep most of these separate as the content may change in the future. The logic for when to send each email should stay the same though. 
### Screenshots

### Notion Card Links

PR Checklist | Your Answer
------------ | -------------


services/QuillLMS/app/mailers/user_mailer.rb
```diff
@@ -40,6 +45,99 @@ def account_created_email(user, temp_password, admin_name)
...
+  end
+
+  def internal_tool_made_district_admin_email(user, district_name)
+    @user = user
+    @district_name = district_name
+    @set_password_link = ENV['DEFAULT_URL']
+    mail to: user.email, subject: "#{user.first_name}, you are now a Quill admin for #{district_name}"
+  end
+
   # Sent when an admin requests an existing teacher to join one of their schools.
```
emilia-friedberg: This looks like it's going to be a duplicate of one of the emails above, is this just gonna get cleaned up later?

emilia-friedberg (approved):
This seems soooooo tedious. Thank you for doing all of that. The code looks good to me in general but I wonder if, since it's just so many functions and files, it makes sense to group the individual worker files into a `workers` subfolder (`admin_user_management_email_workers` or something), and same with the `user_mailer` erb files. And then maybe move all those functions in `user` into a helper module. That's just organizational preference but I'm worried all the new code will make it harder to scan those folders and files quickly when you're looking for something.

brendanshean:
I was actually thinking the same thing @emilia-friedberg said about reorganization.  without knowing too much about the code it looks like there might be two folders naturally:
Screenshot 2022-12-02 at 8 10 14 AM

services/QuillLMS/app/mailers/user_mailer.rb
eadams17: yeah so this is part of the existing signup flow-- this email will get removed once the last two cards are implemented and these new emails are used instead

services/QuillLMS/spec/models/user_spec.rb
```diff
@@ -57,7 +57,7 @@
 #
 require 'rails_helper'
 
-describe User, type: :model do
+describe User, type: :model do # rubocop:disable Metrics/BlockLength
```
brendanshean: Usually a good indicator to break up underlying class into new classes.

brendanshean (approved):
Other than reorganizing folder structure, I think this looks great and is ready to go.

PR #10020, opened 2022-12-02
admin user account finish setup page

emilia-friedberg:
## WHAT
Add a page for users with admin-created accounts to finish setting up their login info.

## WHY
This is a better UX than the existing flow, where we set a temporary password for the user, which is not ideal from a user experience perspective because it requires the user to log in and then change their password.

## HOW
Create the new page, based on the sign up page, and populate the edit and update methods to supply and change the necessary data. Note: these already existed, but were not being used on the site. We'll need to update the emails we are sending with the link to this page.

### Screenshots
Screen Shot 2022-12-02 at 9 59 21 AM

### Notion Card Links
https://www.notion.so/quill/Admin-user-account-setup-landing-page-32c8d3438de240f980b011574ab7d5cd

PR Checklist | Your Answer
------------ | -------------


services/QuillLMS/app/controllers/accounts_controller.rb
```diff
@@ -39,26 +38,25 @@ def create
...
+    redirect_to profile_path, notice: "Sorry, this link has expired. Please contact your Quill admin or the <a href='mailto:hello@quill.org'>Quill support team</a>".html_safe
+  end
 
-    user_params.merge! validate_username: validate_username
-    if @user.update(user_params)
-      redirect_to updated_account_path
+  def update
+    if @user.update(update_user_params)
+      sign_in @user
+      puts '@user.attributes', @user.attributes
```
brendanshean: Leftover `puts` statements

services/QuillLMS/app/controllers/accounts_controller.rb
```diff
@@ -74,20 +72,28 @@ def edit
                                  :username)
   end
 
+  protected def update_user_params
+    params.require(:user).permit(:name, :password)
+end
+
   protected def creation_json
```
brendanshean: How about a rename on `creation_json` since it's now used in `update` (e.g. `redirect_json`)

services/QuillLMS/config/routes.rb
```diff
@@ -523,9 +523,13 @@
   post '/session/set_post_auth_redirect', to: 'sessions#set_post_auth_redirect'
   resource :session
 
-  resource :account, only: [:new, :create, :edit, :update] do
+  resource :account, only: [:new, :create, :edit, :update, :show] do
     post :role, on: :member
   end
+
+  get 'account/:id/finish_set_up', to: 'accounts#edit'
```
brendanshean: Since account is a singular resource, the `:id` here is kind of confusing since it looks like a restful route.  What about using `:token`?

```
  get 'account/:token/finish_set_up', to: 'accounts#edit'
  put 'account/:token', to: 'accounts#update'
```

services/QuillLMS/app/controllers/accounts_controller.rb
```diff
@@ -39,26 +38,25 @@ def create
     end
   end
 
-  def update
-    user_params.delete(:password) unless user_params[:password].present?
-    @user = current_user
+  def edit
+    @user = User.find_by_token(params[:id])
```
brendanshean: I think this line is redundant given the before_action?

services/QuillLMS/app/controllers/accounts_controller.rb
```diff
@@ -39,26 +38,25 @@ def create
...
+    return if @user.present?
 
-    if user_params[:username] == @user.username
-      validate_username = false
-    else
-      validate_username = true
-    end
+    redirect_to profile_path, notice: "Sorry, this link has expired. Please contact your Quill admin or the <a href='mailto:hello@quill.org'>Quill support team</a>".html_safe
```
brendanshean: I like to refactor error messages into [en.yml](https://github.com/empirical-org/Empirical-Core/blob/0d83fb606899718411591c1ee25dea9d44b887fb/services/QuillLMS/config/locales/en.yml#L22) for  several reasons:  
1)  Single source of truth with the text (e.g. simplifies [testing](https://github.com/empirical-org/Empirical-Core/pull/10020/files#diff-5ef8a8db0bcd0f2523214ca8da214420110b48536a0bb4a7b1d7ce84287e4a6bR183))
2)  Sets up a nice pattern for i18n later.
3)  Keeps lines short `redirect_to profile_path, notice: t('accounts.edit.expired_link')`
4)  It's easier to keep a consistent voice/theme with messages if they're all in the same place.

services/QuillLMS/app/controllers/accounts_controller.rb
```diff
@@ -39,26 +38,25 @@ def create
...
+  end
 
-    user_params.merge! validate_username: validate_username
-    if @user.update(user_params)
-      redirect_to updated_account_path
+  def update
+    if @user.update(update_user_params)
+      sign_in @user
+      puts '@user.attributes', @user.attributes
+      puts '@user.password', @user.password
```
eadams17: I think you forgot to remove these

brendanshean (approved):
Nice work on this.  This flow seems like a definite improvement on the previous one.

eadams17 (approved):
this looks great! Thanks for knocking this out

services/QuillLMS/client/app/bundles/Session/startup/clientRegistration.js
```diff
@@ -2,5 +2,6 @@ import 'lazysizes';
 import 'lazysizes/plugins/parent-fit/ls.parent-fit';
 import ReactOnRails from 'react-on-rails';
 import NewAccountApp from './NewAccountAppClient.jsx';
+import FinishSetUpApp from './FinishSetUpAppClient';
 
-ReactOnRails.register({ NewAccountApp, });
+ReactOnRails.register({ NewAccountApp, FinishSetUpApp, });
```
anathomical, line 7: Oh, wow.  It never really occurred to me that you could pass multiple options into `ReactOnRails.register`.  Sweet!

services/QuillLMS/client/app/bundles/Teacher/components/accounts/shared/agreements.tsx
```diff
@@ -0,0 +1,7 @@
+import * as React from 'react'
+
+const Agreements = () => (
+  <p className="agreements">By signing up, you agree to our <a href="/tos">Terms of Service</a> and <a href="/privacy">Privacy&nbsp;Policy.</a></p>
+)
```
anathomical, line 5: I'm really glad you broke this out into a separate component rather than just duplicating.  Makes a lot of sense.

anathomical (approved):
Looks good to me!  I especially like the way you've broken a lot of items out into separate components where they're used in both the basic onboarding and this flow.

PR #10036, opened 2022-12-07
Staggered Release Endpoints Part 7

brendanshean:
## WHAT
Add batch processing of AssignRecommendationWorkers for independent practice activity packs.

## WHY
Recommendations are currently [assigned](https://github.com/empirical-org/Empirical-Core/blob/develop/services/QuillLMS/app/controllers/teachers/progress_reports/diagnostic_reports_controller.rb#L325) via a loop of AssignRecommendationWorkers. 

Once all of the recommendations have been assigned, (i.e. each worker is finished), the PackSequence object should to recompute all UserPackSequenceItem statuses, but not before all Recommendations Workers are finished otherwise there might be activity packs assigned to a classroom that haven't been included in the calculation of locked and unlocked.

## HOW
Sidekiq Pro has a [batch](https://github.com/mperham/sidekiq/wiki/Batches) mode that triggers actions that depend on asynchronous jobs completing via [callbacks](https://github.com/mperham/sidekiq/wiki/Batches#callbacks). 

1) Create a batch `batch = Sidekiq::Batch.new`
2) Add a callback for on success: `batch.on(:success, BatchAssignRecommendationWorker, { pack_sequence_id: pack_sequence_id })`
3) Wrap the jobs in a batch: `batch.jobs { AssignRecommendationWorker.perform... }`
4) After the jobs complete, `BatchAssignRecommendationsWorker.new.on_success(status, options)` is called.

### Screenshots

### Notion Card Links

PR Checklist | Your Answer
------------ | -------------

### DIFF
--- a/services/QuillLMS/app/services/independent_practice_packs_assigner.rb
+++ b/services/QuillLMS/app/services/independent_practice_packs_assigner.rb
@@ -73,10 +73,11 @@ class IndependentPracticePacksAssigner < ApplicationService
   end
 
   private def selections_with_students
-    @selections_with_students ||=
+    @selections_with_students ||= begin
       selections
         .select { |selection| selection[:classrooms][0][:student_ids]&.compact&.any? }
         .map { |selection| selection.permit(:id, classrooms: [:id, :order, student_ids: []]).to_h }
+    end
   end
 
   private def set_diagnostic_recommendations_start_time
--- a/services/QuillLMS/app/services/independent_practice_packs_assigner.rb
+++ b/services/QuillLMS/app/services/independent_practice_packs_assigner.rb
@@ -19,8 +19,8 @@ class IndependentPracticePacksAssigner < ApplicationService
     user:
   )
     @assigning_all_recommendations = assigning_all_recommendations
-    @classroom_id = classroom_id
-    @diagnostic_activity_id = diagnostic_activity_id
+    @classroom_id = classroom_id.to_i
+    @diagnostic_activity_id = diagnostic_activity_id.to_i
     @release_method = release_method
     @selections = selections
     @user = user
@@ -49,18 +49,11 @@ class IndependentPracticePacksAssigner < ApplicationService
   private def assign_recommendations
     pack_sequence = staggered_release? ? pack_sequence_getter : nil
 
-    selections_with_students.each_with_index do |selection, index|
-      AssignRecommendationsWorker.perform_async(
-        pack_sequence_id: pack_sequence&.id,
-        assigning_all_recommendations: assigning_all_recommendations,
-        classroom_id: classroom_id,
-        is_last_recommendation: index == last_recommendation_index,
-        lesson: false,
-        order: selection[:order],
-        student_ids: selection[:classrooms][0][:student_ids].compact,
-        unit_template_id: selection[:id]
-      )
-    end
+    BatchAssignRecommendationsWorker.perform_async(
+      assigning_all_recommendations,
+      pack_sequence&.id,
+      selections_with_students
+    )
   end
 
   private def destroy_existing_pack_sequences
@@ -79,12 +72,11 @@ class IndependentPracticePacksAssigner < ApplicationService
     end
   end
 
-  private def last_recommendation_index
-    @last_selected_recommendation_index ||= selections_with_students.length - 1
-  end
-
   private def selections_with_students
-    @selections_with_students ||= selections.select { |ut| ut[:classrooms][0][:student_ids]&.compact&.any? }
+    @selections_with_students ||=
+      selections
+        .select { |selection| selection[:classrooms][0][:student_ids]&.compact&.any? }
+        .map { |selection| selection.permit(:id, classrooms: [:id, :order, student_ids: []]).to_h }
   end
 
   private def set_diagnostic_recommendations_start_time
--- a/services/QuillLMS/app/workers/assign_recommendations_worker.rb
+++ b/services/QuillLMS/app/workers/assign_recommendations_worker.rb
@@ -6,17 +6,15 @@ class AssignRecommendationsWorker
 
   # rubocop:disable Metrics/CyclomaticComplexity
   def perform(options={})
-    options = options.with_indifferent_access
-
-    assign_on_join = options[:assign_on_join] || false
-    assigning_all_recommendations = options[:assigning_all_recommendations] || false
-    classroom_id = options[:classroom_id]
-    is_last_recommendation = options[:is_last_recommendation]
-    lesson = options[:lesson]
-    order = options[:order]
-    pack_sequence_id = options[:pack_sequence_id]
-    student_ids = options[:student_ids]
-    unit_template_id = options[:unit_template_id]
+    assign_on_join = options['assign_on_join'] || false
+    assigning_all_recommendations = options['assigning_all_recommendations'] || false
+    classroom_id = options['classroom_id']
+    is_last_recommendation = options['is_last_recommendation']
+    lesson = options['lesson']
+    order = options['order']
+    pack_sequence_id = options['pack_sequence_id']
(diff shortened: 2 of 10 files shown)


services/QuillLMS/app/services/independent_practice_packs_assigner.rb
```diff
@@ -79,12 +72,11 @@ def run
     end
   end
 
-  private def last_recommendation_index
-    @last_selected_recommendation_index ||= selections_with_students.length - 1
-  end
-
   private def selections_with_students
-    @selections_with_students ||= selections.select { |ut| ut[:classrooms][0][:student_ids]&.compact&.any? }
+    @selections_with_students ||=
```
dandrabik: Unless this has changed in Ruby, I don't think `||=` will work with multiple lines like this. I believe you need to wrap it in a block:

```ruby
@selections_with_students ||= begin
  ...
end

File: services/QuillLMS/app/workers/batch_assign_recommendations_worker.rb [hunk 2]
```diff
@@ -0,0 +1,42 @@
...
+class BatchAssignRecommendationsWorker
+  include Sidekiq::Worker
+  sidekiq_options queue: SidekiqQueue::CRITICAL
+
+  def perform(assigning_all_recommendations, pack_sequence_id, selections_with_students)
+    return if selections_with_students.empty?
+
+    last_recommendation_index = selections_with_students.length - 1
+
+    batch = Sidekiq::Batch.new
```
dandrabik: I think we'd want to add  `batch.callback_queue = SidekiqQueue::CRITICAL`

https://github.com/mperham/sidekiq/wiki/Batches#callback-details
Screenshot 2022-12-07 at 3 23 57 PM

services/QuillLMS/app/workers/batch_assign_recommendations_worker.rb
```diff
@@ -0,0 +1,42 @@
...
+    batch.on(:success, self.class, pack_sequence_id: pack_sequence_id)
+
+    batch.jobs do
```
dandrabik: The batch tracking is only needed for those with a `pack_sequence_id`. Feel free to ignore if this is an over-optimization / over-complication, but you could conditionally do this, e.g.
```ruby
batch_runner(pack_sequence_id) do
   selections_with_students.each_with_index ...
end

def batch_runner(pack_sequence_id)
  return yield if pack_sequence_id.nil?
  
  batch = Sidekiq::Batch.new
  batch.on(:success, self.class, pack_sequence_id: pack_sequence_id)
  batch.jobs do
    yield
  end
end

File: services/QuillLMS/app/workers/batch_assign_recommendations_worker.rb [hunk 4]
```diff
@@ -0,0 +1,42 @@
+# frozen_string_literal: true
+
```
dandrabik, line 2: Looks like you can monitor batches in the web UI, let's add this to the routes file:
`require 'sidekiq/pro/web'`

https://github.com/mperham/sidekiq/wiki/Batches#monitoring
Screenshot 2022-12-07 at 3 56 25 PM

services/QuillLMS/app/workers/batch_assign_recommendations_worker.rb
```diff
@@ -0,0 +1,42 @@
...
+  include Sidekiq::Worker
+  sidekiq_options queue: SidekiqQueue::CRITICAL
+
+  def perform(assigning_all_recommendations, pack_sequence_id, selections_with_students)
+    return if selections_with_students.empty?
+
+    last_recommendation_index = selections_with_students.length - 1
+
+    batch = Sidekiq::Batch.new
+    batch.on(:success, self.class, pack_sequence_id: pack_sequence_id)
```
dandrabik: Just noting that since we are using `on(:success)`, a failed `AssignRecommendationsWorker` job will prevent any sequences from being assigned until they are successful. That's probably fine, just wanted to note it in case that was a concern.

https://github.com/mperham/sidekiq/wiki/Batches#callbacks

Screenshot 2022-12-07 at 4 00 07 PM

services/QuillLMS/app/workers/assign_recommendations_worker.rb
```diff
@@ -6,17 +6,15 @@ class AssignRecommendationsWorker
...
-    is_last_recommendation = options[:is_last_recommendation]
-    lesson = options[:lesson]
-    order = options[:order]
-    pack_sequence_id = options[:pack_sequence_id]
-    student_ids = options[:student_ids]
-    unit_template_id = options[:unit_template_id]
+    assign_on_join = options['assign_on_join'] || false
+    assigning_all_recommendations = options['assigning_all_recommendations'] || false
+    classroom_id = options['classroom_id']
+    is_last_recommendation = options['is_last_recommendation']
```
dandrabik, line 12: This is likely outside of the scope of this PR, but the code in this job after `return unless is_last_recommendation` might better fit in the the batch callback you just created (i.e. it looks like the intention of that code is a jury-rigged version of this new batch code).

services/QuillLMS/app/services/independent_practice_packs_assigner.rb
```diff
@@ -19,8 +19,8 @@ def initialize(
     user:
   )
     @assigning_all_recommendations = assigning_all_recommendations
-    @classroom_id = classroom_id
-    @diagnostic_activity_id = diagnostic_activity_id
+    @classroom_id = classroom_id.to_i
+    @diagnostic_activity_id = diagnostic_activity_id.to_i
```
dandrabik, line 23: Could this be `nil` for a non-diagnostic? If so, let's do a `nil` check.

dandrabik (approved):
Overall looks good to me. A few small considerations to look at, but all are take or leave.

Getting excited this is almost ready!

anathomical (approved):
Strong +1 for Dan's recommendation of adding the ability to inspect these in the web dashboard.  I suspect that'll go a long way toward any debugging we might need to do.  Also, nice job figuring out batch processing in Sidekiq.  I imagine that this will be a useful pattern for us in the future at some point, too.

services/QuillLMS/app/workers/assign_recommendations_worker.rb
brendanshean: I actually debated doing that in this code, but I will leave that for a separate PR since it would also require updating the Whole Class Instruction [path](https://github.com/empirical-org/Empirical-Core/blob/bs-staggered-release-backend/services/QuillLMS/app/controllers/teachers/progress_reports/diagnostic_reports_controller.rb#L226) as well.  The rubocop cyclotomic complexity will be a good reminder not to forget.

services/QuillLMS/app/services/independent_practice_packs_assigner.rb
brendanshean: At this point, pack sequences are only constructed via the diagnostic activity route so this can be assumed to be non-nill.

services/QuillLMS/app/workers/batch_assign_recommendations_worker.rb
brendanshean: Thanks.  I forgot to add this in.

services/QuillLMS/app/workers/batch_assign_recommendations_worker.rb
brendanshean: Great idea.

services/QuillLMS/app/workers/batch_assign_recommendations_worker.rb
brendanshean: It's not quite clear to me if we should use the `on_complete` and `on_death` callbacks.   For the time being, I'm planning on  using the Sidekiq UI Batches tab to monitor failures.

PR #10038, opened 2022-12-07
Add more examples fields to Synthetic Data Seed data

dandrabik:
## WHAT
Increase seed label examples from 2 to 3 and add the ability to add more examples and remove examples.
## WHY
The Curriculum Team usually needs a little more data than the 2 examples provide, so we're adding a third. Plus they'd prefer controls to add more/less examples if needed.
## HOW
1. Increase default
2. Add a "Add Example button"
3. Add an `X` button to remove an example.
### Screenshots
Screenshot 2022-12-07 at 1 18 24 PM

### Notion Card Links

https://www.notion.so/quill/Synthetic-Data-Increase-Number-of-example-fields-allow-add-remove-examples-b115ac81dc1a412198d3750c24176cf4
PR Checklist | Your Answer
------------ | -------------
Have you added and/or updated tests? |   'NO', just form changes, backend didn't change

### DIFF
--- a/services/QuillLMS/client/app/bundles/Staff/components/evidence/syntheticData/seedDataForm.tsx
+++ b/services/QuillLMS/client/app/bundles/Staff/components/evidence/syntheticData/seedDataForm.tsx
@@ -151,9 +151,7 @@ const SeedDataForm = ({ history, match }) => {
         />
         <button
           className='x-button'
-          id={index}
-          onClick={e => removeExample(conjunction, index, exampleIndex}
-          value={conjunction}
+          onClick={() => removeExample(conjunction, index, exampleIndex}
         >
           <img alt="quill-circle-checkmark" src={quillCloseX} />
         </button>
--- a/services/QuillLMS/client/app/bundles/Staff/components/evidence/syntheticData/seedDataForm.tsx
+++ b/services/QuillLMS/client/app/bundles/Staff/components/evidence/syntheticData/seedDataForm.tsx
@@ -43,6 +43,19 @@ const SeedDataForm = ({ history, match }) => {
     setLabelConfigs(data)
   }
 
+  function onAddExample(e) {
+    const { target } = e;
+    const { id, value } = target;
+    addExample(value, id);
+  }
+
+  function addExample(conjunction, index) {
+    const data = {...labelConfigs}
+
+    data[conjunction][index].examples.push('')
+    setLabelConfigs(data)
+  }
+
   function onAddLabelConfigs(e) {
     const { target } = e;
     const { value } = target;
@@ -140,7 +153,16 @@ const SeedDataForm = ({ history, match }) => {
           />
         </div>
         {labelConfig.examples.map((example, exampleIndex) => renderExample(example, index, conjunction, exampleIndex))}
+        <button
+          className='quill-button fun secondary outlined'
+          id={index}
+          onClick={onAddExample}
+          value={conjunction}
+        >
+          <span className='plus'>+</span>
 
+          &nbsp;Add Example
+        </button>
       </div>
     );
   }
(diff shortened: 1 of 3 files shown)


services/QuillLMS/client/app/bundles/Staff/components/evidence/syntheticData/seedDataForm.tsx
```diff
@@ -112,11 +135,20 @@ const SeedDataForm = ({ history, match }) => {
...
-      />
+      <div className='example-container'>
+        <Input
+          className="example-input"
+          handleChange={e => handleExampleChange(e, index, conjunction, exampleIndex)}
+          label={`Example ${exampleIndex + 1}`}
+          value={value}
+        />
+        <button
+          className='x-button'
```
dandrabik, line 146: Note, I tried to get this to work through an intermediary event handler `onRemoveExample` like I did with `onAddExample`, but it didn't work consistently. Some clicks passed `undefined` props to the event handler and some worked as expected, seemingly randomly. I couldn't figure out why, so I called the function directly, which works consistently.

anathomical: Maybe some sort of weird race condition around event cascades?  I think I'm okay with a direct call like this in small cases like this.

anathomical (approved):

services/QuillLMS/client/app/bundles/Staff/components/evidence/syntheticData/seedDataForm.tsx
```diff
@@ -112,11 +135,20 @@ const SeedDataForm = ({ history, match }) => {
 
   function renderExample(value, index, conjunction, exampleIndex) {
```
emilia-friedberg, line 136: Another way to handle this would be to make this its own component, instead of just a render function. Then you could define, for example, a `handleRemoveExampleClick` function that calls `removeExample` with the props `conjunction, index, exampleIndex`, and then you can just pass `handleRemoveExampleClick` directly to the `onClick` handler. Same for the `handleExampleChange` in the `.example-input` element (which I realize is older code).

services/QuillLMS/client/app/bundles/Staff/components/evidence/syntheticData/seedDataForm.tsx
emilia-friedberg: So `e` was undefined? Weird. I think the approach from the comment above would solve that issue, but I agree with @anathomical that this case is small enough that I don't feel like it's necessary to rewrite it here (the only reason to avoid it is that using an arrow function directly like this creates a new function on every render, which could theoretically break some React optimizations, but I think it would be unnoticeable in virtually every use case and this is an internal tool. also I find them slightly harder to read.)

emilia-friedberg (approved):
Small comment about possible refactor but it's more of an FYI for the future. Overall LGTM!

services/QuillLMS/client/app/bundles/Staff/components/evidence/syntheticData/seedDataForm.tsx
dandrabik: Yeah, that's a good idea. I'm going to refactor this in a later PR.

services/QuillLMS/client/app/bundles/Staff/components/evidence/syntheticData/seedDataForm.tsx
dandrabik: Ok, thanks for the feedback on this. I'm going to deploy the code as this (so that the team has this sooner), and work on a refactor.

  And for context, `e` wasn't undefined, but the props passed via it were, i.e. what becomes `conjunction, index, exampleIndex` in the function were all `undefined`. Weirdly, it worked about 1/3 to 1/2 of the time, and failed the rest (click a bunch, sometimes it worked, sometimes it didn't).

PR #10039, opened 2022-12-07
Staggered Release Endpoints Part 8

brendanshean:
## WHAT
Add redirection to activity links for students who have not unlocked a particular activity.

## WHY
Teachers share activity links with their students, but some might be locked based on a particular student's current progress.  If it is locked, redirect the user.

## HOW
Add a check `activity.locked_user_pack_sequence_item?(user)` 

### Screenshots

### Notion Card Links
https://www.notion.so/quill/Staggered-release-for-diagnostic-recommendations-7b11a96525d847f5b7e2ec5280a32f70#1e086610d85847a09533cadefcd5b5f0

PR Checklist | Your Answer
------------ | -------------
Have you added and/or updated tests? | 'YES'

### DIFF
--- a/services/QuillLMS/config/locales/en.yml
+++ b/services/QuillLMS/config/locales/en.yml
@@ -30,6 +30,7 @@ en:
     errors:
       activity_not_assigned: 'Sorry, you do not have access to this activity because it has not been assigned to you. Please contact your teacher.'
       activity_pack_not_assigned: 'Sorry, you do not have access to this activity pack because it has not been assigned to you. Please contact your teacher.'
+      activity_locked: 'Sorry, you do not have access to this activity as it is locked.  Please contact your teacher.'
   clever:
     account_conflict: "This Clever account is already associated with another Quill account. Contact support@quill.org for further assistance."
   actioncontroller:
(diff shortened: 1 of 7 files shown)


services/QuillLMS/config/locales/en.yml
```diff
@@ -30,6 +30,7 @@ en:
     errors:
       activity_not_assigned: 'Sorry, you do not have access to this activity because it has not been assigned to you. Please contact your teacher.'
       activity_pack_not_assigned: 'Sorry, you do not have access to this activity pack because it has not been assigned to you. Please contact your teacher.'
+      activity_locked: 'Sorry, you do not have access to this activity as it is locked.  Please contact your teacher.'
```
anathomical: Do we want a more complex error here so that if a student does contact their teacher, and that teacher comes to look at their computer, the teacher can quickly figure out why the link isn't working?  Not sure exactly what that text would look like, but I think we definitely want to avoid students encountering this message, calling their teachers, and then their teachers reaching out to Support.

anathomical (approved):
I like the approach for this feature, but it may well be worth tweaking the text of the error message we show the user if they do get to a locked activity.

services/QuillLMS/config/locales/en.yml
brendanshean: That's a good point.  I've updated the copy.

PR #10040, opened 2022-12-07
admin user account creation

emilia-friedberg:
## WHAT
Update the admin user account creation UI and update the backend to handle all of the different branching paths and emails.

## WHY
We want to allow admins to create accounts for other admins, and better handle all the different possibilities of existing and new user accounts within that creation flow.

## HOW
Minor UI updates, more extensive revisions to the backend controller action which sends off the email workers. Note that I moved this method from the `teachers` controller to the `admins` controller because the route was already nested as `admins/:id/teachers`, so it felt confusing to find it in just the `teachers_controller` file, and it isn't really a straightforward create method at all. There are a lot of nested conditionals in there - very open to suggestions about how else we might approach that. 

### Screenshots
Screen Shot 2022-12-07 at 2 56 47 PM

### Notion Card Links
https://www.notion.so/quill/Admin-user-account-creation-and-linking-by-admins-via-admin-dashboard-133bd668e80e450081d37ef0aa630078

PR Checklist | Your Answer
------------ | -------------


services/QuillLMS/app/mailers/admin_dashboard_user_mailer.rb
```diff
@@ -57,4 +57,13 @@ def made_school_admin_change_school_email(user, admin_name, new_school, existing
     mail to: user.email, subject: "#{user.first_name}, you are now a Quill admin for #{new_school.name}"
   end
 
+  private def link_for_setting_password(role)
+    params = {
+      accountType: role,
+      adminFullName: @admin_name,
+      schoolName: @school_name
+    }
```
eadams17: I think rather than have this as a method in just this mailer class, we could tweak this to handle optional params and move it to the Mailer User class so it can be used for the other 3 instances where we set `@set_password_link`, ie:

```
def link_for_setting_password(role)
  params = {
    accountType: role
  }
  params[:adminFullName] = @admin_name if @admin_name
  params[:schoolName] = @school_name if @school_name
  @set_password_link = "#{ENV['DEFAULT_URL']}/account/#{self.token}/finish_set_up?#{params.to_query}"
end

...

@set_password_link = @user.mailer_user.link_for_setting_password(role: 'school admin')
```

services/QuillLMS/config/locales/en.yml
```diff
@@ -35,6 +35,17 @@ en:
...
+    existing_account:
+      admin:
+        linked: "This account already exists and is an admin of %{school_name}."
+        new: "This account already exists. They were made an admin and notified by email."
+      teacher:
+        linked: "This account already exists and is linked to %{school_name}."
+        new: "This account already exists. They were sent an email asking them to link their account to %{school_name}."
+    new_account:
+      admin: "Admin added. They were notified by email."
+      teacher: "Teacher added. They were notified by email."
```
eadams17, line 49: nice! thanks for adding this, it'll be useful for the internal tool implementation

eadams17 (approved):
LGTM! Just one comment about a suggested refactor for the password link generating method

services/QuillLMS/app/mailers/admin_dashboard_user_mailer.rb
emilia-friedberg: Sounds good, I'll move it to the `UserMailer` class. I don't think we need to worry about the params being null because `.to_query` will just stringify them as null values, so when the frontend checks for their presence it won't find anything.

emilia-friedberg:
Closing this PR because I accidentally merged in staggered release somehow and it is going to be easier to combine this with #10063 than untangle them. Also this appears to have gotten lost in a review Bermuda Triangle.

PR #10047, opened 2022-12-08
Synthetic Seed Data - Remove starting Capital letters

dandrabik:
## WHAT
The synthetic seed data we are generating often comes back starting with capital letters (especially for the paraphrasing examples code) despite the fact that this data is for the middle of the sentence.  This attempts to correct that.
## WHY
This capitalization doesn't match what student data looks like or what is correct answers, so we want to correct for it, since NLP models do factor in capitalization.
## HOW
For this solution, I'm being relatively conservative, and only lowercasing things I feel confident should be lowercased. I'm defining those as:
1. Top 100 most common English words (excluding 'I')
2. Appears in the passage in lowercase.

I think this should mitigate the issue pretty heavily while avoiding lowercasing proper nouns by accident. We could always expand on this code, but I feel this simple solution will get us 90% there.
### Screenshots

Screenshot 2022-12-08 at 5 47 18 PM

### Notion Card Links
https://www.notion.so/quill/Avoid-capitalization-in-seed-data-5daecb7653b34cfd8ec2ce23e07d15ab

PR Checklist | Your Answer
------------ | -------------


services/QuillLMS/engines/evidence/lib/evidence/evidence/open_ai/concerns/sentence_results.rb
```diff
@@ -6,13 +6,14 @@ module Concerns::SentenceResults
       extend ActiveSupport::Concern
 
       BLANK = ''
+      MIN_LENGTH = 10
```
dandrabik, line 9: This is an unrelated tiny refactor.

services/QuillLMS/engines/evidence/lib/evidence/evidence/synthetic/seed_data_generator.rb
```diff
@@ -150,13 +150,18 @@ def run
...
         api_results
+          .map {|s| lowercaser.run(s) }
           .map {|s| Result.new(text: [noun, s].join(SPACE).strip, seed: seed, label: label)}
           .uniq {|r| r.text }
           .reject {|r| r.text.in?(current_texts)}
           .reject {|r| regex_exclude?(r.text) }
           .reject {|r| opinion_api_flagged?(r.text) }
       end
 
+      private def lowercaser
```
dandrabik, line 161: Note, not using `ApplicationService` pattern, e.g. `Evidence::SafeFirstLowercaser.new(passage, text)`, since I don't want to parse the passage each time.

services/QuillLMS/engines/evidence/lib/evidence/evidence/safe_first_lowercaser.rb
```diff
@@ -0,0 +1,35 @@
...
+      @word_list = (COMMON_WORD_LIST + passage_lowercase_words).to_set
+    end
+
+    # Lowercase first word of text if it is a common word
+    # or if it appears lowercase in the passage
+    def run(text)
+      first_in_word_list(text) ? downcase_first_letter(text) : text
+    end
+
+    private def first_in_word_list(text)
```
brendanshean: I would name this `first_word_in_list?(text)` since it returns a boolean.

brendanshean (approved):
Looks good to me.  I like the idea of leveraging words occurring elsewhere in passage.

services/QuillLMS/engines/evidence/lib/evidence/evidence/safe_first_lowercaser.rb
dandrabik: Good call, I did a slight refactor based on this feedback and I think it reads better now. Thanks for the feedback!

brendanshean: I also think it read better now.

PR #10068, opened 2022-12-14
Fix checkbox callback race condition

brendanshean:
## WHAT
Properly handle a race condition with checkbox callback concern.

## WHY
This particular race condition is causing issues with StaggeredRelease.

## HOW
Wrap the `find_or_create_by` in a subtransaction following this [approach](https://github.com/rails/rails/blob/d0192e0c2daa03048a8c4d0d3b94763dbbfec4c1/activerecord/lib/active_record/relation.rb#L154-L160)

### Screenshots

### Notion Card Links

PR Checklist | Your Answer
------------ | -------------

### DIFF
--- a/services/QuillLMS/app/models/concerns/checkbox_callback.rb
+++ b/services/QuillLMS/app/models/concerns/checkbox_callback.rb
@@ -4,11 +4,17 @@ module CheckboxCallback
   extend ActiveSupport::Concern
 
   def find_or_create_checkbox(name, user, activity_id=nil)
-    return unless Objective.find_by_name(name) && user
+    objective = Objective.find_by(name: name)
 
-    Checkbox.find_or_create_by(user_id: user.id, objective_id: Objective.find_by_name(name).id)
-    CheckboxAnalyticsWorker.perform_async(user.id, activity_id) if activity_id
-  rescue => e
-    puts "Race condition"
+    return unless objective && user
+
+    begin
+      Checkbox.transaction(requires_new: true) do
+        Checkbox.find_or_create_by!(user: user, objective: objective)
+        CheckboxAnalyticsWorker.perform_async(user.id, activity_id) if activity_id
+      end
+    rescue ActiveRecord::RecordNotUnique
+      retry
+    end
   end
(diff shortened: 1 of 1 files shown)


emilia-friedberg (approved):
Yay!!!

dandrabik (approved):
This looks fine to me. Thanks for tracking this down. One note, I'm curious why we wouldn't use [create_or_find_by!](https://api.rubyonrails.org/classes/ActiveRecord/Relation.html#method-i-create_or_find_by-21) here.

services/QuillLMS/app/models/concerns/checkbox_callback.rb
```diff
@@ -4,11 +4,17 @@ module CheckboxCallback
...
 
-    Checkbox.find_or_create_by(user_id: user.id, objective_id: Objective.find_by_name(name).id)
-    CheckboxAnalyticsWorker.perform_async(user.id, activity_id) if activity_id
-  rescue => e
-    puts "Race condition"
+    return unless objective && user
+
+    begin
+      Checkbox.transaction(requires_new: true) do
+        Checkbox.find_or_create_by!(user: user, objective: objective)
```
dandrabik, line 13: Should this use [create_or_find_by](https://api.rubyonrails.org/classes/ActiveRecord/Relation.html#method-i-create_or_find_by-21). Seems to be closely mimicking that code.

Screenshot 2022-12-14 at 10 28 12 AM

brendanshean: I had looked into this yesterday, but opted to stick with `find_or_create_by!`

Some issues:
[primary keys running out](https://github.com/rails/rails/issues/35543)
[less performant](https://www.mayerdan.com/ruby/2020/10/22/ar-find_or_create)

Neither of those are deal-breakers, `create_or_find_by` would also require a rescue clause for ActiveRecord::RecordInvalid since there's a validation on the checkbox model.

I think long term I would like to default to `create_or_find_by` but I think for now I'll just leave as `find_or_create_by`

PR #10075, opened 2022-12-15
lms/clear-cache-when-hiding-activities

anathomical:
## WHAT
Touch the parent Unit when hiding UnitActivities
## WHY
So that the cache clears and the fact that these activities have been removed will reflect in reporting
## HOW
Just add a `touch` call to the `hide` controller action

### Notion Card Links
https://www.notion.so/quill/Deleted-activity-showing-in-Activity-Summary-report-7c07237133624cd5bc2cd6436e9586c1

PR Checklist | Your Answer
------------ | -------------

### DIFF
--- a/services/QuillLMS/app/controllers/teachers/unit_activities_controller.rb
+++ b/services/QuillLMS/app/controllers/teachers/unit_activities_controller.rb
@@ -21,7 +21,7 @@ class Teachers::UnitActivitiesController < ApplicationController
     @activity_sessions.update_all(visible: false)
     @unit_activities.each(&:save_user_pack_sequence_items)
     # touch the parent unit in order to clear our caches
-    @unit_activity.unit.touch
+    @unit_activity&.unit&.touch
     ResetLessonCacheWorker.new.perform(current_user.id)
     render json: {}
   end
--- a/services/QuillLMS/app/controllers/teachers/unit_activities_controller.rb
+++ b/services/QuillLMS/app/controllers/teachers/unit_activities_controller.rb
@@ -20,6 +20,8 @@ class Teachers::UnitActivitiesController < ApplicationController
     @unit_activity&.unit&.hide_if_no_visible_unit_activities
     @activity_sessions.update_all(visible: false)
     @unit_activities.each(&:save_user_pack_sequence_items)
+    # touch the parent unit in order to clear our caches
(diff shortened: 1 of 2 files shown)


services/QuillLMS/app/controllers/teachers/unit_activities_controller.rb
```diff
@@ -20,6 +20,8 @@ def hide
     @unit_activity&.unit&.hide_if_no_visible_unit_activities
     @activity_sessions.update_all(visible: false)
     @unit_activities.each(&:save_user_pack_sequence_items)
+    # touch the parent unit in order to clear our caches
+    @unit_activity.unit.touch
```
brendanshean: Line 20 has safe navigation with unit whereas line 24 does not.  I'm not sure which one makes more sense, but I think they should be consistent.

services/QuillLMS/spec/controllers/teachers/unit_activities_controller_spec.rb
```diff
@@ -24,6 +24,11 @@
       put :hide, params: { id: unit_activity.id }
       expect(unit_activity.reload.visible).to eq false
     end
+
+    it 'should touch the parent unit in order to bubble up cache clearing' do
+      expect_any_instance_of(Unit).to receive(:touch)
```
brendanshean, line 30: In general I try to avoid using `expect_any_instance` and `allow_any_instance` as they are legacy methods that the rspec maintainers [advocate](https://relishapp.com/rspec/rspec-mocks/docs/working-with-legacy-code/any-instance) not using.

For this one you could stub the class finder method that gets you the object:
```
      allow(UnitActivity).to receive(:find).with(unit_activity.id.to_s).and_return(unit_activity)
      expect(unit_activity.unit).to receive(:touch)
```
The benefit of this approach is that we're specifically targeting one instance rather than all instances.

There is a [cop](https://www.rubydoc.info/gems/rubocop-rspec/1.7.0/RuboCop/Cop/RSpec/AnyInstance) for this one, but I think we currently have too many violations so I haven't turned it on yet.

brendanshean (approved):
Tracking down caching errors can be very tricky.  Nice job on figuring this out!

eadams17 (approved):

emilia-friedberg (approved):

PR #10078, opened 2022-12-16
Evidence Activity Health Dashboard front end

cissyxyu:
## WHAT
Add the front end React components for the Evidence Activity Health table.

## WHY
This will help Curriculum team make judgments about Evidence activities to work on.

## HOW
The bulk of the work is in the new React component HealthDashboard, which will display a table of all the evidence activity health data. 
The new HealthDashboard table also has filtering and download as CSV functionalities. I edited the shared component `ReactTable` a bit to enable manual filtering (similar to the existent manual sorting), because we need to control all filters and sorts manually in order to download the data as a CSV.

### Screenshots
Screen Shot 2022-12-16 at 7 50 54 PM

### Notion Card Links
https://www.notion.so/quill/2-Activity-Health-Dashboard-Front-End-5d802fcb4c144c7985e1f4f67a0f5fe1

PR Checklist | Your Answer
------------ | -------------

### DIFF
--- a/services/QuillLMS/client/app/bundles/Staff/components/evidence/healthDashboards/healthDashboard.tsx
+++ b/services/QuillLMS/client/app/bundles/Staff/components/evidence/healthDashboards/healthDashboard.tsx
@@ -1,7 +1,6 @@
 import * as React from "react";
 import { useRef } from 'react';
 import { useQuery } from 'react-query';
-import { matchSorter } from 'match-sorter';
 import { CSVLink } from 'react-csv';
 import { firstBy } from 'thenby';
 
@@ -12,6 +11,42 @@ import { fetchAggregatedActivityHealths } from '../../../utils/evidence/ruleFeed
 import { getLinkToActivity, secondsToHumanReadableTime, addCommasToThousands } from "../../../helpers/evidence/miscHelpers";
 
 const ALL_FLAGS = "All Flags"
+
+const CustomTextFilter = ({ column, setFilter, placeholder }) => {
+  return (
+    <input
+      aria-label="text-filter"
+      onChange={event => setFilter(column.id, event.target.value)}
+      placeholder={placeholder}
+      style={{ width: "100%" }}
+      value={column.filterValue}
+    />
+  )
+}
+
+const RenderNumberFilter = ({ column, setFilter }) => (
+  <CustomTextFilter
+    column={column}
+    placeholder="1-2, <1, >1"
+    setFilter={setFilter}
+  />
+)
+
+const RenderTimeFilter = ({ column, setFilter }) => (
+  <CustomTextFilter
+    column={column}
+    placeholder="in seconds: 1-2, <1, >1"
+    setFilter={setFilter}
+  />
+)
+
+const RenderTrueFalseFilter = ({ column, setFilter }) => (
+  <CustomTextFilter
+    column={column}
+    placeholder="1 / 0"
+    setFilter={setFilter}
+  />
+)
 
 export const HealthDashboard = ({ location, match }) => {
   const [flag, setFlag] = React.useState<string>(ALL_FLAGS)
@@ -44,10 +79,7 @@ export const HealthDashboard = ({ location, match }) => {
       Header: 'Name',
       accessor: "name",
       key: "name",
-      filter: (rows, idArray, filterValue) => {
-        return matchSorter(rows, filterValue, { keys: ['original.name']})
-      },
-      Filter: TextFilter,
+      Filter: CustomTextFilter,
       filterAll: true,
       Cell: ({row}) => <span className="name"><a href={getLinkToActivity(row.original.activity_id)} rel="noopener noreferrer" target="_blank">{row.original.name}</a></span>, // eslint-disable-line react/display-name
       minWidth: 330,
@@ -58,16 +90,14 @@ export const HealthDashboard = ({ location, match }) => {
       Header: 'Version #',
       accessor: "version",
       key: "version",
-      filter: filterNumbers,
-      Filter: TextFilter,
+      Filter: RenderNumberFilter,
       minWidth: 70,
     },
     {
       Header: 'Version Plays',
       accessor: "version_plays",
       key: "versionPlays",
-      filter: filterNumbers,
-      Filter: TextFilter,
+      Filter: RenderNumberFilter,
       minWidth: 70,
       Cell: ({row}) => row.original.version_plays && <span className="versionPlays">{addCommasToThousands(row.original.version_plays)}</span>
     },
@@ -75,26 +105,23 @@ export const HealthDashboard = ({ location, match }) => {
       Header: 'Total Plays',
       accessor: "total_plays",
       key: "totalPlays",
-      filter: filterNumbers,
-      Filter: TextFilter,
+      Filter: RenderNumberFilter,
       minWidth: 70,
       Cell: ({row}) => row.original.total_plays && <span className="versionPlays">{addCommasToThousands(row.original.total_plays)}</span>
     },
     {
-      Header: 'Completion Rate',
+      Header: 'Completed Rate',
       accessor: "completion_rate",
       key: "completionRate",
-      filter: filterNumbers,
-      Filter: TextFilter,
-      minWidth: 70,
+      Filter: RenderNumberFilter,
+      minWidth: 80,
       Cell: ({row}) => row.original.completion_rate && <span className="name">{row.original.completion_rate}%</span>
     },
     {
       Header: 'Because Final Optimal',
       accessor: "because_final_optimal",
       key: "becauseFinalOptimal",
-      filter: filterNumbers,
-      Filter: TextFilter,
+      Filter: RenderNumberFilter,
       minWidth: 70,
       Cell: ({row}) => row.original.because_final_optimal && <span className={row.original.because_final_optimal <= 75 ? "poor-health" : "" + " name"}>{row.original.because_final_optimal}%</span>
     },
@@ -102,8 +129,7 @@ export const HealthDashboard = ({ location, match }) => {
       Header: 'But Final Optimal',
       accessor: "but_final_optimal",
       key: "butFinalOptimal",
-      filter: filterNumbers,
-      Filter: TextFilter,
+      Filter: RenderNumberFilter,
       minWidth: 70,
       Cell: ({row}) => row.original.but_final_optimal && <span className={row.original.but_final_optimal <= 75 ? "poor-health" : "" + " name"}>{row.original.but_final_optimal}%</span>
     },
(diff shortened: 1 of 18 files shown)


services/QuillLMS/engines/evidence/spec/factories/evidence/activity_healths.rb
```diff
@@ -0,0 +1,17 @@
+# frozen_string_literal: true
+
+FactoryBot.define do
+  factory :evidence_activity_health, class: 'Evidence::ActivityHealth' do
+    name { "test name" }
+    flag { "production" }
+    activity_id { 1 }
```
brendanshean, line 7: I noticed this `activity_id` and then checked back on the models/evidence/activity_health.rb file and noticed there was no `belongs_to :activity`.  The schema says that it can be nil but is that what we want?

Sorry of for the backend question on a frontend PR but figured I would ask.

cissyxyu: So, because the EvidenceActivityHealth records are essentially just cached data for a simple front end display, I think they don't really need to be attached to Evidence Activities. This is the pattern we use in the other Activity Health dashboard for Connect and Grammar.

We do this because there's not actually any use cases where we would want to pull an EvidenceActivityHealth record from an activity itself. When we want to pull specific data for an activity, we actually just run the queries again fresh, since the data in these EvidenceActivityHealth is one day old or more.

These EvidenceActivityHealth records essentially just exist as an "out-of-date, but useful for a quick glance" thing for front-end use, but if admin want to do any deeper work with an activity's health data they will go to the Activity's actual dashboard and see the real-time data live.

brendanshean: That makes sense.  Thank you for the thorough explanation and context.

services/QuillLMS/client/app/bundles/Staff/components/evidence/healthDashboards/healthDashboard.tsx
```diff
@@ -0,0 +1,317 @@
...
+      aria-label="text-filter"
+      onChange={event => setFilter(column.id, event.target.value)}
+      placeholder={placeholder}
+      style={{ width: "100%" }}
+      value={column.filterValue}
+    />
+  )
+}
+
+const RenderNumberFilter = ({ column, setFilter }) => (
```
emilia-friedberg: This is a little nitpicky, but since these are components and not simply functions, I wouldn't use `Render` in the name (generally verbs are more appropriate for functions and nouns for entities).

services/QuillLMS/client/app/bundles/Staff/components/evidence/healthDashboards/healthDashboard.tsx
```diff
@@ -0,0 +1,317 @@
...
+
+  const getSortedRows = (rows, sortInfo) => {
+    if(sortInfo) {
+      const { id, desc } = sortInfo;
+      // we have this reversed so that the first click will sort from highest to lowest by default
+      const directionOfSort = desc ? `asc` : 'desc';
+      return rows.sort(firstBy(id, { direction: directionOfSort }));
+    } else {
+      return rows;
+    }
```
emilia-friedberg: This could be an early return.

services/QuillLMS/client/app/bundles/Staff/components/evidence/healthDashboards/healthDashboard.tsx
```diff
@@ -0,0 +1,317 @@
...
+
+    return filteredData
+  }
+
+
+  const getSortedRows = (rows, sortInfo) => {
+    if(sortInfo) {
+      const { id, desc } = sortInfo;
+      // we have this reversed so that the first click will sort from highest to lowest by default
+      const directionOfSort = desc ? `asc` : 'desc';
```
emilia-friedberg: 'asc' here is in funny quotes.

services/QuillLMS/client/app/bundles/Staff/components/evidence/healthDashboards/healthDashboard.tsx
```diff
@@ -0,0 +1,317 @@
...
+      return rows;
+    }
+  }
+
+  const filterRowsByColumnValue = (rows, column, value) => {
+    if (column === 'name') {
+      return rows.filter(r => r.name.toLowerCase().includes(value.toLowerCase()))
+    } else if (column === 'poor_health_flag') {
+      if (value === '1') return rows.filter(r => r.poor_health_flag === true)
+      else if (value === '0') return rows.filter(r => !r.poor_health_flag)
```
emilia-friedberg: We might want to constantize these column names and values since they're being used for business logic.

services/QuillLMS/client/app/bundles/Staff/components/evidence/healthDashboards/healthDashboard.tsx
```diff
@@ -0,0 +1,317 @@
...
+      else return rows
```
emilia-friedberg: Nested else-ifs feel like a code smell to me - we should maybe break this into its own function.

services/QuillLMS/client/app/bundles/Staff/components/evidence/healthDashboards/healthDashboard.tsx
```diff
@@ -0,0 +1,317 @@
...
+
+    const columns = dataTableFields
+    let dataToDownload = []
+    for (let index = 0; index < rows.length; index++) {
+      let recordToDownload = {}
+      for(let colIndex = 0; colIndex < columns.length ; colIndex ++) {
+        recordToDownload[columns[colIndex].Header] = rows[index][columns[colIndex].accessor]
+      }
+      dataToDownload.push(recordToDownload)
+    }
```
emilia-friedberg, line 251: Since I think this data can get pretty big, we might want to avoid nested `for` loops for performance purposes. I think the approach in [this article](https://spaceofmatej.com/removing-nested-loops-in-javascript) would work for this use case. If it doesn't, it's something we could look at optimizing in the future if and when it becomes too slow.

services/QuillLMS/client/app/bundles/Staff/helpers/evidence/miscHelpers.tsx
```diff
@@ -477,3 +477,29 @@ export function renderCSVDownloadButton(csvDataLoadInitiated, handleLoadCSVDataC
...
   return <button className="quill-button fun primary contained csv-download-button" onClick={handleLoadCSVDataClick}>Load CSV Data</button>
 }
+
+
+export function addCommasToThousands(num)
+{
+  if (!num) return ""
+  let num_parts = num.toString().split(".");
+  num_parts[0] = num_parts[0].replace(/\B(?=(\d{3})+(?!\d))/g, ",");
+  return num_parts.join(".");
```
emilia-friedberg: `num_parts` should be camelCased.

services/QuillLMS/client/app/bundles/Staff/helpers/evidence/routingHelpers.ts
```diff
@@ -70,6 +70,10 @@ export const getActivityHealthUrl = ({ activityId }) => {
   return `activity_health?activity_id=${activityId}`;
 }
 
+export const getAggregatedActivityHealthsUrl = () => {
+  return `evidence/activity_healths.json`;
+}
```
emilia-friedberg: Since there are no variables this could just be a `const` and not a function, right?

services/QuillLMS/engines/evidence/app/models/evidence/activity_health.rb
```diff
@@ -24,8 +24,13 @@ def serializable_hash(options = nil)
...
           :avg_completion_time
         ],
-        include: [:prompt_healths]
+        include: [:prompt_healths],
+        methods: [:poor_health_flag]
       ))
     end
+
+    def poor_health_flag
+      return ((because_final_optimal && because_final_optimal < 75) || (but_final_optimal && but_final_optimal < 75) || (so_final_optimal && so_final_optimal < 75))
```
emilia-friedberg: 75 should probably be a const here. Also, do you need that wrapping parentheses? And the `return` is unnecessary since it's ruby.

emilia-friedberg (approved):
Some suggested refactors, mostly around making constants and nested loops, but looks great to me over all. Nice work on those updates to `ReactTable`.

services/QuillLMS/client/app/bundles/Staff/components/evidence/healthDashboards/healthDashboard.tsx
```diff
@@ -0,0 +1,317 @@
...
+  />
+)
+
+export const HealthDashboard = ({ location, match }) => {
+  const [flag, setFlag] = React.useState<string>(ALL_FLAGS)
+  const [promptSearchInput, setPromptSearchInput] = React.useState<string>("")
+  const [dataToDownload, setDataToDownload] = React.useState<Array<{}>>([])
+  const [rows, setRows] = React.useState<Array<{}>>(null)
+  const [poorHealthFlag, setPoorHealthFlag] = React.useState<boolean>(false)
+  let csvLink = useRef(null);
```
eadams17, line 59: this should be a const

services/QuillLMS/client/app/bundles/Staff/components/evidence/healthDashboards/healthDashboard.tsx
```diff
@@ -0,0 +1,317 @@
...
+    let newRows = activityHealthsData.activityHealths
+    filters.forEach((filter) => {
+      const column = filter.id
+      const value = filter.value
+
+      newRows = filterRowsByColumnValue(newRows, column, value)
+    })
+    setRows(getFilteredData(newRows))
+  }
+
```
eadams17, line 226: I think this would make more sense as follows since I'm not really seeing any reason to initially set `newRows` to the value of `activityHealthsData.activityHealths`:

```
  const handleFiltersChange = (filters) => {
    if (!rows) return;
    let newRows
    filters.forEach((filter) => {
      const column = filter.id
      const value = filter.value

      newRows = filterRowsByColumnValue(activityHealthsData.activityHealths, column, value)
    })
    setRows(getFilteredData(newRows))
  }
```

services/QuillLMS/client/app/bundles/Staff/components/evidence/healthDashboards/healthDashboard.tsx
```diff
@@ -0,0 +1,317 @@
...
+  return(
+    <React.Fragment>
+      <Navigation location={location} match={match} />
+      <div className="health-dashboards-index-container">
+        <section className="header-section">
+          <h2>Activities Health Dashboard</h2>
+          <div className="first-input-row">
+            <FlagDropdown flag={flag} handleFlagChange={handleFlagChange} isLessons={true} />
+            <div className="csv-download-button">
+              <button onClick={formatTableForCSV} style={{cursor: 'pointer'}} type="button">
```
eadams17: we generally try to avoid in-line styles so I would just add `className="quill-button (whatever backpack button style seems to fit best)"` which will already have the `cursor: pointer` style included

services/QuillLMS/client/app/bundles/Staff/styles/evidence_manager/health_dashboard.scss
```diff
@@ -0,0 +1,58 @@
...
+  }
+
+  .table-section {
+    td {
+      white-space: normal;
+    }
+    .rt-th {
+      min-width: 100px;
+      width: 100px;
+      max-width: 9.0072e15px;
```
eadams17: this seems off?

eadams17 (approved):
Thanks for tackling this! Just left a few small comments, but nothing major.

PR #10081, opened 2022-12-16
Fix Staggered Release invalid foreign key

brendanshean:
## WHAT
Fix a staggered release [bug](https://sentry.io/organizations/quillorg-5s/issues/3810073844/?project=11238&query=is%3Aunresolved&referrer=issue-stream&statsPeriod=14d) involving an invalid foreign key with Staggered release.

## WHY
When a user switches from staggered release to immediate, the underlying `pack_sequence` is destroyed.  This also destroys associated `pack_sequence_items` which in turn destroy associated `user_pack_sequence_items`.  `UserPackSequenceItem` has an after_destroy callback that is circular in that after_destroying the object, it calls a worker that effectively recreates the same object by calling SaveUserPackSequenceItemWorker.  Instead we should only be recalculating these objects through updates to ActivitySession, ClassroomUnit, etc that have actual effects on UserPackSequenceItem.  

The `after_save` callback on `UserPackSequenceItem` should also be removed since it triggered unnecessary calls to `SaveUserPackSequenceItemWorker`, namely when `after_save` is called, status is [updated](https://github.com/empirical-org/Empirical-Core/blob/develop/services/QuillLMS/app/services/user_pack_sequence_item_saver.rb#L59) which will in turn trigger the `after_save` callback again.

### Addendum
After removing the callbacks on `UserPackSequenceItem` a race condition involving `PackSequenceItem` still remained:

1.  PackSequence [destroy](https://github.com/empirical-org/Empirical-Core/blob/develop/services/QuillLMS/app/services/independent_practice_packs_assigner.rb#L62) is called
2.  Associated PackSequenceItems destroy is [called](https://github.com/empirical-org/Empirical-Core/blob/9edff3e8fcf88cf23192b1ddc261a27ed38acedc/services/QuillLMS/app/models/pack_sequence.rb#L35) due to dependent: :destroy
3.  Associated UserPackSequenceItems destroy is [called](https://github.com/empirical-org/Empirical-Core/blob/9edff3e8fcf88cf23192b1ddc261a27ed38acedc/services/QuillLMS/app/models/pack_sequence_item.rb#L29) due dependent: :destroy
 4.  PackSequenceItem after_destroy callback is called
 5.  Triggers SaveUserPackSequenceItemWorkers
 6.  This has two parts, a query and then the update of status
 7.  Since these are asynchronous, the underlying PackSequenceItem might initially exist but after the workers have been queued, will actually be destroyed.
 8.  This causes the race condition in the update of status in 6.

However, this led me to the solution which is that we don't currently need the 
`after_destroy` callback on PackSequenceItems as these objects are only destroyed via the assigner flow.

## HOW
Remove callbacks on `UserPackSequenceItem` and `PackSequenceItem`

### Screenshots

### Notion Card Links

PR Checklist | Your Answer
------------ | -------------

### DIFF
--- a/services/QuillLMS/app/models/pack_sequence.rb
+++ b/services/QuillLMS/app/models/pack_sequence.rb
@@ -41,6 +41,6 @@ class PackSequence < ApplicationRecord
   validates :release_method, inclusion: { in: RELEASE_METHODS }
 
   def save_user_pack_sequence_items
-    students.pluck(:id).each { |user_id| SaveUserPackSequenceItemsWorker.perform_async(classroom.id, user_id) }
+    classroom.students.each { |student| SaveUserPackSequenceItemsWorker.perform_async(classroom.id, student.id) }
   end
 end
--- a/services/QuillLMS/app/models/pack_sequence.rb
+++ b/services/QuillLMS/app/models/pack_sequence.rb
@@ -41,6 +41,6 @@ class PackSequence < ApplicationRecord
   validates :release_method, inclusion: { in: RELEASE_METHODS }
 
   def save_user_pack_sequence_items
-    students.pluck(:id).each { |user_id| SaveUserPackSequenceItemsWorker.perform_async(classroom&.id, user_id) }
+    students.pluck(:id).each { |user_id| SaveUserPackSequenceItemsWorker.perform_async(classroom.id, user_id) }
   end
 end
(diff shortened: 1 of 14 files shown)


dandrabik (approved):
LGTM. Thanks for tracking this down. Circular callbacks can be hard to find!

dandrabik (approved):
LGTM! Thanks for tracking this down.

PR #10086, opened 2022-12-19
Fix stripe bug with nil credit card

brendanshean:
## WHAT
Stripe stores credit card information in different places for various reasons:

1) For new subscriptions, stripe attaches the credit card info on the stripe subscription object.
2) For legacy subscriptions (i.e. subscriptions not purchased through stripe), the credit card is stored on a sources object on the stripe customer object.
3) When a user with a legacy subscription in the past changes their credit card on file _before_ the newer subscription type has been purchased.  stripe stores the credit card information on the stripe customer invoice settings.

This PR fixes a [bug](https://www.notion.so/quill/Credit-card-listed-as-null-on-My-Subscriptions-page-87900b0c36b943aeaf82e6f623100b41) with case 3 where the user's credit card listed is null

## WHY
Users should be able to see what credit card is on file.

## HOW
Add a check for invoice settings on the last_four method.

### Screenshots

### Notion Card Links
https://www.notion.so/quill/Credit-card-listed-as-null-on-My-Subscriptions-page-87900b0c36b943aeaf82e6f623100b41

PR Checklist | Your Answer
------------ | -------------

### DIFF
--- a/services/QuillLMS/app/models/stripe_integration/user.rb
+++ b/services/QuillLMS/app/models/stripe_integration/user.rb
@@ -1,0 +1,31 @@
+# frozen_string_literal: true
+
+module StripeIntegration
+  class User < SimpleDelegator
+    def last_four
+      return nil unless stripe_customer_id
+
+      stripe_default_payment_method&.card&.last4 || stripe_customer_source&.last4
+    end
+
+    private def stripe_customer
+      Stripe::Customer.retrieve(id: stripe_customer_id, expand: ['sources'])
+    rescue Stripe::InvalidRequestError
+      nil
+    end
+
+    private def stripe_customer_source
+      stripe_customer&.sources&.data&.first
+    end
+
+    private def stripe_default_payment_method
+      Stripe::PaymentMethod.retrieve(stripe_default_payment_method_id)
+    rescue Stripe::InvalidRequestError
+      nil
+    end
+
+    private def stripe_default_payment_method_id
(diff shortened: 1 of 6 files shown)


services/QuillLMS/app/models/user.rb
```diff
@@ -161,6 +161,8 @@ class User < ApplicationRecord
     allow_nil: true,
     prefix: :school
 
+  delegate :last_four, to: :stripe_user
```
anathomical, line 164: Love the use of `delegate`.  That's a pattern we're probably under-using.

services/QuillLMS/app/models/stripe_integration/user.rb
```diff
@@ -0,0 +1,31 @@
+# frozen_string_literal: true
+
+module StripeIntegration
+  class User < SimpleDelegator
```
anathomical, line 4: I hadn't used `SimpleDelegator` before.  This is sweet.

anathomical (approved):
I like sliding the stripe link logic into a delegator class like this.  Makes a lot of sense to me.

services/QuillLMS/app/models/stripe_integration/user.rb
brendanshean: There's a couple other examples:
Screenshot 2022-12-19 at 6 03 04 PM


I feel like it's a good way of carving up our large user class.

PR #10087, opened 2022-12-20
update scorebook query and view to handle locked activities

emilia-friedberg:
## WHAT
Update the scorebook query and view to handle locked activities.

## WHY
So teachers have additional context for activities that have not been completed. 

## HOW
Just update the query to use the new tables and send the data to the frontend to render the appropriate icon.

### Screenshots
Screen Shot 2022-12-20 at 11 11 22 AM

### Notion Card Links
https://www.notion.so/quill/Changes-to-the-activity-summary-report-to-represent-locked-activities-1b6ddaad6539442b87d0483ad7c771f0

PR Checklist | Your Answer
------------ | -------------

### DIFF
--- a/services/QuillLMS/app/queries/scorebook/query.rb
+++ b/services/QuillLMS/app/queries/scorebook/query.rb
@@ -29,7 +29,8 @@ class Scorebook::Query
           SUM(acts.timespent) AS timespent,
           SUM(CASE WHEN acts.state = '#{ActivitySession::STATE_FINISHED}' THEN 1 ELSE 0 END) AS completed_attempts,
           SUM(CASE WHEN acts.state = '#{ActivitySession::STATE_STARTED}' THEN 1 ELSE 0 END) AS started,
-          SUM(CASE WHEN acts.is_final_score = true THEN acts.id ELSE 0 END) AS id
+          SUM(CASE WHEN acts.is_final_score = true THEN acts.id ELSE 0 END) AS id,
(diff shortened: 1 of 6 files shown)


services/QuillLMS/app/queries/scorebook/query.rb
```diff
@@ -68,11 +74,15 @@ def self.run(classroom_id, current_page=1, unit_id=nil, begin_date=nil, end_date
           unit_activities.publish_date,
           unit_activities.due_date,
           unit_activities.created_at,
-          unit_owner.time_zone
+          unit_owner.time_zone,
+          psi.id,
+          upsi.id,
+          upsi.status
```
anathomical, line 80: Pretty sure that you don't need this `status` in `GROUP BY`.  (We haven't really been taking advantage of this, but if you `GROUP BY <table>.id` then PSQL recognizes that you're grouping by every column in that table (since ID is the most unique value in the table).  So this  means we really don't need `students.name` or `activity.name` in this `GROUP BY` either, but I don't think you need to go back and clean those up.

anathomical (approved):
A super minor PSQL thing that you could fix, but doesn't hurt anything being the way it is.  Other than that, this seems like a decent solution.

brendanshean (approved):
Looks good to me!  Nice job on figuring out the query.

eadams17 (approved):

PR #10063, opened 2022-12-13
admin user account management

emilia-friedberg:
## WHAT
Update the admin user account creation and management UIs and update the backend to handle all of the different branching paths and emails.

## WHY
We want to allow admins to create accounts for other admins, and better handle all the different possibilities of existing and new user accounts within that creation flow. We also want to make user management more straightforward within this interface.

## HOW
Minor UI updates, more extensive revisions to the backend controller action which sends off the email workers. Note that I moved this method from the `teachers` controller to the `admins` controller because the route was already nested as `admins/:id/teachers`, so it felt confusing to find it in just the `teachers_controller` file, and it isn't really a straightforward create method at all. There are a lot of nested conditionals in there - very open to suggestions about how else we might approach that. Also note that this PR contains both management work and creation work, which was originally a separate branch (#10040).

### Screenshots
Screen Shot 2022-12-07 at 2 56 47 PM
Screen Shot 2022-12-14 at 10 02 03 AM

### Notion Card Links
https://www.notion.so/quill/Admin-user-account-creation-and-linking-by-admins-via-admin-dashboard-133bd668e80e450081d37ef0aa630078
https://www.notion.so/quill/Admin-user-account-management-by-admins-via-admin-dashboard-0e57d2ee9c8a432b8006131adb77d472

PR Checklist | Your Answer
------------ | -------------
Have you added and/or updated tests? | YES
Have you deployed to Staging? | YES
Self-Review: Have you done an initial self-review of the code below on Github? | YES
Spec Review: Have you reviewed the spec and ensured this meets requirements and/or matches design mockups? | YES

PR #10092, opened 2022-12-21
Admin user management/ segment updates

eadams17:
## WHAT
add segment events for various admin user events

## WHY
we want to track when these events happen

## HOW
add two new methods to handle school and district admin tracking, and call them in the various workers we use

### Screenshots

### Notion Card Links
https://www.notion.so/quill/Admin-user-Segment-events-b10332b27a3041968c2b5e295ec9b0cd

PR Checklist | Your Answer
------------ | -------------

### DIFF
--- a/services/QuillLMS/app/workers/admin_dashboard/admin_account_created_email_worker.rb
+++ b/services/QuillLMS/app/workers/admin_dashboard/admin_account_created_email_worker.rb
@@ -8,7 +8,10 @@ class AdminDashboard::AdminAccountCreatedEmailWorker
     @admin_name = User.find_by(id: admin_user_id)&.name
     @school_name = School.find_by(id: school_id)&.name
     @is_reminder = is_reminder
-    @user&.mailer_user&.send_admin_dashboard_admin_account_created_email(@admin_name, @school_name, @is_reminder)
+
+    return unless @user && @admin_name && @school_name
+
+    @user.mailer_user.send_admin_dashboard_admin_account_created_email(@admin_name, @school_name, @is_reminder)
 
     analytics = SegmentAnalytics.new
     analytics.track_school_admin_user(
--- a/services/QuillLMS/app/workers/admin_dashboard/admin_account_created_email_worker.rb
+++ b/services/QuillLMS/app/workers/admin_dashboard/admin_account_created_email_worker.rb
@@ -9,5 +9,13 @@ class AdminDashboard::AdminAccountCreatedEmailWorker
     @school_name = School.find_by(id: school_id)&.name
     @is_reminder = is_reminder
     @user&.mailer_user&.send_admin_dashboard_admin_account_created_email(@admin_name, @school_name, @is_reminder)
+
+    analytics = Analyzer.new
+    analytics.track_school_admin_user(
+      user,
+      SegmentIo::BackgroundEvents::ADMIN_CREATED_SCHOOL_ADMIN_ACCOUNT,
+      @school_name,
+      @admin_name
+    )
   end
 end
--- a/services/QuillLMS/app/workers/admin_dashboard/admin_account_created_email_worker.rb
+++ b/services/QuillLMS/app/workers/admin_dashboard/admin_account_created_email_worker.rb
@@ -10,9 +10,9 @@ class AdminDashboard::AdminAccountCreatedEmailWorker
     @is_reminder = is_reminder
     @user&.mailer_user&.send_admin_dashboard_admin_account_created_email(@admin_name, @school_name, @is_reminder)
 
-    analytics = Analyzer.new
+    analytics = SegmentAnalytics.new
     analytics.track_school_admin_user(
-      user,
+      @user,
       SegmentIo::BackgroundEvents::ADMIN_CREATED_SCHOOL_ADMIN_ACCOUNT,
       @school_name,
       @admin_name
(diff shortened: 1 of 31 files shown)


services/QuillLMS/app/workers/admin_dashboard/admin_account_created_email_worker.rb
```diff
@@ -8,6 +8,17 @@ def perform(user_id, admin_user_id, school_id, is_reminder)
     @admin_name = User.find_by(id: admin_user_id)&.name
```
brendanshean: I use instance variables to share values across methods in an instance of a class.  Otherwise, I use regular local variables if the scope should only apply to the method scope. (see some discussion [here](https://stackoverflow.com/a/10280944)).  In most/all of these worker classes the variable assignment is local to `perform` and then sent to a mailer.   Contrast this with the mailer classes where the instance variables are picked up by the [mailer].html.erb files indirectly (i.e. we don't pass them directly to anything).

For this reason, I would change all of the instance variables in all of these worker classes to just regular local variables.  For example:

```
    def perform(user_id, admin_user_id, school_id, is_reminder)
    user = User.find_by(id: user_id)
    admin_name = User.find_by(id: admin_user_id)&.name
    school_name = School.find_by(id: school_id)&.name
    is_reminder = is_reminder

    return unless user && admin_name && school_name

    user.mailer_user.send_admin_dashboard_admin_account_created_email(admin_name, school_name, is_reminder)

    analytics = SegmentAnalytics.new
    analytics.track_school_admin_user(
      user,
      SegmentIo::BackgroundEvents::ADMIN_CREATED_SCHOOL_ADMIN_ACCOUNT,
      school_name,
      admin_name
    )
  end
```

services/QuillLMS/app/workers/admin_dashboard/admin_account_created_email_worker.rb
```diff
@@ -8,6 +8,17 @@ def perform(user_id, admin_user_id, school_id, is_reminder)
...
     @school_name = School.find_by(id: school_id)&.name
     @is_reminder = is_reminder
-    @user&.mailer_user&.send_admin_dashboard_admin_account_created_email(@admin_name, @school_name, @is_reminder)
+
+    return unless @user && @admin_name && @school_name
+
+    @user.mailer_user.send_admin_dashboard_admin_account_created_email(@admin_name, @school_name, @is_reminder)
+
+    analytics = SegmentAnalytics.new
```
brendanshean: You can collapse 16-17 into one call and get rid of the temporary variable

```
 SegmentAnalytics.new.track_school_admin_user(
      @user,
      SegmentIo::BackgroundEvents::ADMIN_CREATED_SCHOOL_ADMIN_ACCOUNT,
      @school_name,
      @admin_name
    )
```

services/QuillLMS/spec/workers/admin_dashboard/made_school_admin_link_school_email_worker_spec.rb
```diff
@@ -0,0 +1,60 @@
...
+    it 'should not send a segment.io event' do
+      expect(analytics).not_to receive(:track_school_admin_user)
+      worker.perform(nil, referring_admin.id, school.id)
+    end
+  end
+
+  describe 'user is not nil' do
+
+    before do
+      allow(User).to receive(:find_by).and_return(teacher, referring_admin)
```
brendanshean, line 43: Cool.  This is an interesting use of the multi-object return.

services/QuillLMS/spec/workers/admin_dashboard/admin_account_created_email_worker_spec.rb
```diff
@@ -0,0 +1,63 @@
+# frozen_string_literal: true
+
+require 'rails_helper'
+
+# perform(user_id, admin_user_id, school_id, is_reminder)
+
+describe AdminDashboard::AdminAccountCreatedEmailWorker, type: :worker do
+  let!(:teacher) { create(:teacher) }
```
brendanshean, line 8: Instead of the comment on line 5, use a 
`subject { perform(user_id, admin_user_id, school_id, is_reminder)`

on line 8.

This does two things:
1) it standardizes what will be tested in each spec
2) if the args for the worker change, the spec should fail, whereas the comment might go stale.

Then replaces spec with `subject`, for example:

```
  it 'should not send the mail with user mailer' do
      expect(mailer_class).not_to receive(mailer_method)
      subject
    end
```

services/QuillLMS/spec/workers/admin_dashboard/admin_account_created_email_worker_spec.rb
```diff
@@ -0,0 +1,63 @@
...
+  let!(:referring_admin) { create(:teacher) }
+  let!(:school) { create(:school) }
+  let!(:mailer_user) { Mailer::User.new(teacher) }
+  let!(:mailer_class)  { AdminDashboardUserMailer }
+  let!(:mailer_method) { :admin_account_created_email}
+  let!(:worker) { AdminDashboard::AdminAccountCreatedEmailWorker.new }
```
brendanshean: You can get rid of `let!(:worker) { AdminDashboard::AdminAccountCreatedEmailWorker.new }` after refactoring out the `subject`

brendanshean (approved):
I left some refactoring comments, but after those are addressed, I think this should be good to go.  Nice work on this.

services/QuillLMS/app/workers/admin_dashboard/admin_account_created_email_worker.rb
eadams17: That makes sense! Thanks for the explanation and link to the SO discussion 🙏

emilia-friedberg (approved):
LGTM! Will you let me know when this goes out? Once it's live I can ship #10063.

PR #10095, opened 2022-12-21
show students all work they have completed, even if some of the data has been archived

emilia-friedberg:
## WHAT
Let students see all work they have already completed, even if some of the relevant data has been archived.

## WHY
So that they have a more accurate "report" of their finished work.

## HOW
Update the query that we use for this dashboard to not filter out archived data, but denote it as such (and ultimately filter out data that is both archived and unfinished). 

### Screenshots
Screen Shot 2022-12-21 at 3 09 51 PM

### Notion Card Links
](https://www.notion.so/quill/Allow-students-to-view-completion-data-for-archived-activity-packs-169433872ac74dd98121412eb2d8681b)

PR Checklist | Your Answer
------------ | -------------


services/QuillLMS/app/models/unit_activity.rb
```diff
@@ -157,6 +157,11 @@ def self.get_classroom_user_profile(classroom_id, user_id)
           COALESCE(cuas.locked, false) AS locked,
           COALESCE(cuas.pinned, false) AS pinned,
           MAX(acts.percentage) AS max_percentage,
+          unit.visible AS unit_visible,
+          ua.visible AS ua_visible,
+          cu.visible AS cu_visible,
+          acts.visible AS acts_visible,
+          CASE WHEN unit.visible = false OR ua.visible = false OR cu.visible = false OR acts.visible = false THEN true ELSE false END as archived,
```
anathomical: I'm not sure that this is actually easier to read, but because we're always looking for the same value (`false`), this can be:
```sql
CASE WHEN false IN (unit.visible, ua.visible, cu.visible, acts.visible) THEN true ELSE false END as archived
```

services/QuillLMS/app/models/unit_activity.rb
```diff
@@ -231,5 +234,8 @@ def self.get_classroom_user_profile(classroom_id, user_id)
           ua.id ASC
      SQL
     )
+
+    # note that we could also do this with a HAVING clause, but it is pretty difficult to read because the logic for computing those values is complex
+    data.filter { |ua| ua['finished'] || !ua['archived'] }
```
anathomical, line 236: I concur with your thought of doing this filter in Ruby rather than in SQL rather than making an already complicated query more complicated.

services/QuillLMS/app/models/unit_activity.rb
```diff
@@ -157,6 +157,13 @@ def self.get_classroom_user_profile(classroom_id, user_id)
           COALESCE(cuas.locked, false) AS locked,
           COALESCE(cuas.pinned, false) AS pinned,
           MAX(acts.percentage) AS max_percentage,
+          CASE WHEN
+            unit.visible = false
+            OR ua.visible = false
+            OR cu.visible = false
+            OR acts.visible = false
```
anathomical: I don't know if this is actually simpler, but because we're looking for `false` in all of these, the four conditions here could be rewritten as:
```sql
false IN (unit.visible, ua.visible, cu.visible, acts.visible)
```

anathomical (approved):
Looks good to me.  A small syntax style note, but I don't know if my suggestion is actually more legible than what you have.

eadams17 (approved):

PR #10108, opened 2023-01-03
Fix stripe school subscription bug

brendanshean:
## WHAT
Fix a [bug](https://www.notion.so/quill/PS-When-a-user-subscribes-to-School-Premium-they-should-receive-immediate-access-to-School-Premium-33a214d965734bb1a89aac3f1829f6d7) involving school premium access.

## WHY
When a user signs up for a school premium subscription directly, stripe creates a `trialing` subscription based on the bonus days calculator.  This causes an issue because the check for AmountPaidMismatch will fail since trial plans have cost $0.

## HOW
Add a check for the trialing before raising an `AmountPaidMismatchError`

### Screenshots

### Notion Card Links

PR Checklist | Your Answer
------------ | -------------


services/QuillLMS/app/services/stripe_integration/webhooks/subscription_creator.rb
```diff
@@ -25,8 +28,9 @@ def run
         raise NilStripePriceIdError if stripe_price_id.nil?
         raise NilStripeInvoiceIdError if stripe_invoice.id.nil?
         raise DuplicateSubscriptionError if duplicate_subscription?
-        raise AmountPaidMismatchError if payment_amount != plan.price
+        raise NilSubscriptionStatusError unless stripe_subscription.respond_to?(:status)
 
+        validate_amount_paid_mismatch
```
anathomical: It's not at first glance obvious that this is intended to `raise` a specific error if certain conditions are met the way that the list of items above is.  I wonder if the syntax here isn't cleaner if you refactored `validate_amount_paid_mismatch` to just `return true` if there's not an early return from the conditions, and then `raise AmountPaidMismatchError if validate_amount_paid_mismatch`?  Or something like that?  It's not really a huge deal, but it feels like it might be worth maintaining the pattern since it's so well-established in `run` already.

anathomical (approved):
Seems like a reasonable fix.  I've got a pretty minor organization/syntax comment that might be worth pursuing, but I'm not convinced that I'm right there.

services/QuillLMS/app/services/stripe_integration/webhooks/subscription_creator.rb
brendanshean: That's a good call.  I've refactored to keep the pattern consistent.

services/QuillLMS/app/services/stripe_integration/webhooks/invoice_paid_event_handler.rb
```diff
@@ -4,6 +4,7 @@ module StripeIntegration
   module Webhooks
     class InvoicePaidEventHandler < EventHandler
       ACTIVE = 'active'
+      TRIALING = 'trialing'
```
dandrabik, line 7: Have to admit, I was thrown by the word `trialing`. Really curious why they didn't go with `trial` here.

Screenshot 2023-01-05 at 9 24 06 AM

dandrabik (approved):
Looks good to me! Thanks for fixing this up. 

Plus, as an added bonus, I've added a new word to my vocabulary, "trialing".

services/QuillLMS/app/services/stripe_integration/webhooks/invoice_paid_event_handler.rb
brendanshean: I would agree.  A couple times, I wrote the constant `TRIAL` and method `trial_subscription?` only to go back and update to trialing.

PR #10112, opened 2023-01-04
migration for closed activity packs view

emilia-friedberg:
## WHAT
Add a migration file, callback, and tests for the new `units` attribute, `active`.

## WHY
We need this column in place in order to build out the new hidden activity packs page (active is the inverse of hidden; see discussion in the comments in the Notion card).

## HOW
Just add the migration, a necessary callback (since units should never be `{ visible: false, active: true }`), and tests. After deploy and before the feature itself is added, we'll want to run the following in the console: `Unit.unscoped.where(visible: false).update(active: false)` to make sure the data is consistent for previously-archived units.

### Screenshots

### Notion Card Links
](https://www.notion.so/quill/V2-RFC-Introduce-activity-pack-hiding-4b39cd1fa869429fbbdd48cc74cbac39?d=ab13888667b34193bca1767facaa85cc)

PR Checklist | Your Answer
------------ | -------------

### DIFF
--- a/services/QuillLMS/app/models/unit.rb
+++ b/services/QuillLMS/app/models/unit.rb
@@ -47,6 +47,8 @@ class Unit < ApplicationRecord
 
   default_scope { where(visible: true)}
 
+  before_save :set_active_to_false
+
   after_save :save_user_pack_sequence_items, if: :visible
   after_save :hide_classroom_units_and_unit_activities
   after_save :create_any_new_classroom_unit_activity_states
@@ -54,6 +56,12 @@ class Unit < ApplicationRecord
   # Using an after_commit hook here because we want to trigger the callback
   # on save or touch, and touch explicitly bypasses after_save hooks
   after_commit :touch_all_classrooms_and_classroom_units
+
+  def set_active_to_false
+    return if visible
+
+    self.active = false
+  end
 
   def hide_if_no_visible_unit_activities
     return if unit_activities.exists?(visible: true)
--- a/services/QuillLMS/app/models/unit.rb
+++ b/services/QuillLMS/app/models/unit.rb
@@ -5,6 +5,7 @@
 # Table name: units
 #
 #  id               :integer          not null, primary key
+#  active           :boolean          default(TRUE), not null
 #  name             :string
 #  visible          :boolean          default(TRUE), not null
 #  created_at       :datetime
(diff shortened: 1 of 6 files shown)


services/QuillLMS/app/models/unit.rb
```diff
@@ -47,6 +48,8 @@ class Unit < ApplicationRecord
 
   default_scope { where(visible: true)}
 
+  before_save :set_active_to_false
```
anathomical: I'm a little torn on this one.  It feels more Rails-y to handle this with a validator function that marks the model as invalid if `visible == false && active == true`, which will prevent the model from saving if that illegal combination is set.  However, the nice thing about your approach is that any code that sets `visible == false`, it automatically "fixes" any values for `active` that aren't also set at the same time.

Basically, I'm not sure if the "magic" of doing this in a `before_save` call is better because it lets us fix most potential cases of conflict in one place, but it's also "magic", and could theoretically surprise someone.  That said, the "surprise" being something like "you can establish an invalid data state" probably isn't a bad one?

I'd be interested to hear other peoples' thoughts on this one.

anathomical (approved):
I've got a point of discussion, but I don't have an objection to doing this as-written.  It seems reasonable to me.  One question is whether it's worth committing the update script in the form of a `rake` task even though it's just a one-liner.  I'm inclined to say it's fine to do this as-is, but figure that's another thing we could discuss if other people have thoughts.

services/QuillLMS/app/models/unit.rb
```diff
@@ -55,6 +58,12 @@ class Unit < ApplicationRecord
   # on save or touch, and touch explicitly bypasses after_save hooks
   after_commit :touch_all_classrooms_and_classroom_units
 
+  def set_active_to_false
```
brendanshean: I think the word `hide` when used in cases not referring to `active` is confusing.  In those cases replace it with `archive`:

```
  def hide_unit_if_archived
    return if visible
  
    self.active = false
  end

  def archive_if_no_visible_unit_activities
    return if unit_activities.exists?(visible: true)

    update(visible: false)
  end

  def archive_classroom_units_and_unit_activities
    return if visible

    unit_activities.where(visible: true).update(visible: false)
    classroom_units.where(visible: true).update(visible: false)
  end

```

services/QuillLMS/app/models/unit.rb
brendanshean: Given our reliance on setting the visible flag in other callbacks, it's probably easiest to use a callback here.

brendanshean (approved):
Nice job sorting out some tricky details on this one.

dandrabik (approved):
Looks good to me. Thanks for coming up with a simple solution for this.

PR #10121, opened 2023-01-05
Fix bug with student merge account not happening properly

brendanshean:
## WHAT
There’s an issue involving schools moving from clever library to secure sync integration where duplicate users aren't being properly merged which results in duplicate users being shown in the classroom.

## WHY
Teachers should not be seeing duplicate users for cases that we can resolve.

## HOW
When attempting to [merge](https://github.com/empirical-org/Empirical-Core/blob/e2559ffb248412995512f24c4ff341e27841ab06/services/QuillLMS/app/services/clever_integration/student_updater.rb#L33) two students, there are some [cases](https://github.com/empirical-org/Empirical-Core/blob/develop/services/QuillLMS/app/models/concerns/student.rb#L123) where students cannot be merged together without the presence of the teacher_id. 

`teacher_id` is available higher upstream, so we can pass it down the chain and do the merge properly.

### Screenshots

### Notion Card Links

PR Checklist | Your Answer
------------ | -------------

### DIFF
--- a/services/QuillLMS/app/services/clever_integration/classroom_students_importer.rb
+++ b/services/QuillLMS/app/services/clever_integration/classroom_students_importer.rb
@@ -2,11 +2,12 @@
 
 module CleverIntegration
   class ClassroomStudentsImporter < ApplicationService
-    attr_reader :classroom, :students_data
+    attr_reader :classroom, :students_data, :teacher_id
 
-    def initialize(classroom, students_data)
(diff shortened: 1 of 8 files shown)


anathomical (approved):
Nice job tracing this issue down.  Seems super-reasonable to pass the teacher id value through the call-chain since we already have access to it when we run the importer.

PR #10131, opened 2023-01-09
closed activity packs

emilia-friedberg:
## WHAT
Add ability to "close" activity packs, removing them from student view though not the reports, in addition to "deleting" them. Add view for closed activity packs as well as open packs.

## WHY
This is a long-requested feature!

## HOW
Building off earlier PRs to put the migrations in place and update the student view, this code implements the actual actions of closing and reopening, as well as the frontend views for both filters and the modals to perform each action.

### Screenshots
Screen Shot 2023-01-10 at 12 17 28 PM
Screen Shot 2023-01-10 at 12 17 34 PM
Screen Shot 2023-01-10 at 12 17 52 PM
Screen Shot 2023-01-10 at 12 18 06 PM
Screen Shot 2023-01-10 at 12 18 14 PM
Screen Shot 2023-01-10 at 12 18 33 PM

### Notion Card Links
https://www.notion.so/quill/Introduce-the-ability-to-close-and-reopen-activity-packs-4b39cd1fa869429fbbdd48cc74cbac39

PR Checklist | Your Answer
------------ | -------------


services/QuillLMS/app/controllers/students_controller.rb
```diff
@@ -107,6 +107,7 @@ def join_classroom
     return redirect_to profile_path unless current_user&.student?
 
     if classroom_id && unit_id
+      flash_closed_unit_error
```
anathomical, line 110: Do we need to remove the line below this that calls `flash_missing_unit_error`?  That one looks like it'd overwrite the error we want here?

services/QuillLMS/app/controllers/teachers/units_controller.rb
```diff
@@ -131,6 +131,20 @@ def hide
...
   end
 
+  def close
+    unit = Unit.find(params[:id])
+    unit.update(open: false)
+    ResetLessonCacheWorker.new.perform(current_user.id)
+    render json: {}
+  end
+
+  def reopen
```
anathomical: Super minor point, but this could also safely just be called `open`.  Theoretically you could imagine a future where some units start `closed`, and thus are not "re" opening.  It almost certainly doesn't matter, but I figured I'd flag it.

services/QuillLMS/client/app/bundles/Teacher/components/assignment_flow/manage_units/manage_units.jsx
```diff
@@ -26,6 +26,7 @@ export default class ManageUnits extends React.Component {
       classrooms: [],
       selectedClassroomId: getParameterByName('classroom_id') || allClassroomKey,
       activityWithRecommendationsIds: [],
+      open: !window.location.pathname.includes('closed')
```
anathomical: This feels like it could be a little bit brittle.  Is there a way to get an attribute passed in through `ReactRouter`?  Maybe something like `useParams()`?  Not sure if that'd work with our current routing.

anathomical (approved):
Looks good to me.  A couple of super-minor notes that you can take or leave however you want.

eadams17 (approved):

services/QuillLMS/app/controllers/students_controller.rb
emilia-friedberg: No - they are different errors, and since `closed_unit` redirects, that line will never be called if they are supposed to see the closed unit error.

cissyxyu (approved):

PR #10133, opened 2023-01-09
Fix intermittent spec failure

brendanshean:
## WHAT
Fix an intermittent spec [failure](https://app.circleci.com/pipelines/github/empirical-org/Empirical-Core/19569/workflows/6a5e7a15-c362-4b30-be71-5e5373ab2eb1/jobs/254940/parallel-runs/1/steps/1-108) with dual_google_and_clever_id_resolver.

Also change the `classrooms` call to `classrooms_i_teach` since `classrooms` doesn't actually give us what we would expect.

## WHY
Intermittent spec failures should be fixed.

## HOW
A collection of objects was occasionally being created in a different order than specified, so stub the `last` call on the result set to ensure the right setup.

### Screenshots

### Notion Card Links

PR Checklist | Your Answer
------------ | -------------
Have you deployed to Staging? | NO

### DIFF
--- a/services/QuillLMS/app/services/dual_google_id_and_clever_id_resolver.rb
+++ b/services/QuillLMS/app/services/dual_google_id_and_clever_id_resolver.rb
@@ -18,6 +18,10 @@ class DualGoogleIdAndCleverIdResolver < ApplicationService
     RESOLVERS.find { |resolver| send(resolver) }
   end
 
+  private def last_classroom
+    @last_classroom ||= user&.classrooms_i_teach&.sort_by(&:updated_at)&.last
+  end
+
   private def log_account_type_change(changed_attribute, previous_value, action)
     ChangeLog.create(
       action: action,
@@ -41,8 +45,6 @@ class DualGoogleIdAndCleverIdResolver < ApplicationService
   end
 
   private def resolve_by_last_classroom
-    last_classroom = user&.classrooms&.order(:updated_at)&.last
-
     return false unless last_classroom.present?
     return set_user_account_type_google if last_classroom.google_classroom_id.present?
     return set_user_account_type_clever if last_classroom.clever_id.present?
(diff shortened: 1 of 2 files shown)


services/QuillLMS/app/services/dual_google_id_and_clever_id_resolver.rb
```diff
@@ -18,6 +18,10 @@ def run
     RESOLVERS.find { |resolver| send(resolver) }
   end
 
+  private def last_classroom
+    @last_classroom ||= user&.classrooms_i_teach&.sort_by(&:updated_at)&.last
```
anathomical, line 22: Would this be faster to use `ORDER BY DESC` since it basically lets us take `first`, and thus only pulls a single record from the database?  (I could be wrong, or it may not even matter much, but I figured I'd check.)
```ruby
user&.classrooms_i_teach&.order(updated_at: :desc).first
```

Obviously you'd have to change the stub logic in the spec, but I think the same idea would work with an `order` call.

anathomical (approved):
Minor suggestion, but basically looks fine.

services/QuillLMS/app/services/dual_google_id_and_clever_id_resolver.rb
brendanshean: The `classrooms_i_teach` call returns an array so we can't use SQL at that point.

The previous `classrooms` call returned an ActiveRecord::Relation which is why it was originally a `order(:updated_at)`

anathomical: Huh, that's weird.  `classrooms_i_teach` is defined as a `has_many through:` relation on the `User` model, so I'd expect that to be a relation, too.  But I guess not.  Carry on, then.

brendanshean: My guess is that is overwritten by this [one](https://github.com/empirical-org/Empirical-Core/blob/develop/services/QuillLMS/app/models/concerns/teacher.rb#L45)

anathomical: Yeah.  That makes sense.  Also, that TODO is hilarious.  I guess we probably have both because there's some code somewhere that expects one form, and other code somewhere else that expects the other.

PR #10151, opened 2023-01-13
Teacher Onboarding/ add migration for role_selected_at_signup

eadams17:
## WHAT
add a `role_selected_at_signup` column to the `teacher_infos` table

## WHY
we need this to store the role that a user selects in the teacher onboarding updates

## HOW
just add a migration to the table for this column (note: this was discussed with both Dan and Emilia, and we felt like this is the best place to store this piece of data)

### Screenshots

### Notion Card Links
https://www.notion.so/quill/Design-of-Collecting-Teacher-Onboarding-Materials-Part-II-04073c66d0d242ee95aea8be6fe34a10

PR Checklist | Your Answer
------------ | -------------
Have you added and/or updated tests? | no-- migration PR

### DIFF
--- a/services/QuillLMS/app/models/teacher_info.rb
+++ b/services/QuillLMS/app/models/teacher_info.rb
@@ -4,12 +4,13 @@
 #
 # Table name: teacher_infos
 #
-#  id                  :bigint           not null, primary key
-#  maximum_grade_level :integer
-#  minimum_grade_level :integer
-#  created_at          :datetime         not null
-#  updated_at          :datetime         not null
-#  user_id             :bigint           not null
+#  id                      :bigint           not null, primary key
+#  maximum_grade_level     :integer
+#  minimum_grade_level     :integer
+#  role_selected_at_signup :string           default("")
+#  created_at              :datetime         not null
+#  updated_at              :datetime         not null
+#  user_id                 :bigint           not null
 #
 # Indexes
 #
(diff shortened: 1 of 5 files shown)


happythenewsad (approved):
Mergable, but why not have the default value be `'user'` instead of `''`, to align with the `User.role` field?

anathomical (approved):

anathomical:
> Mergable, but why not have the default value be `'user'` instead of `''`, to align with the `User.role` field?

Alternatively, it might make sense to get a rake task that can backfill the values to existing users who need it?  Not sure how important that is.

eadams17:
> > Mergable, but why not have the default value be `'user'` instead of `''`, to align with the `User.role` field?
> 
> Alternatively, it might make sense to get a rake task that can backfill the values to existing users who need it? Not sure how important that is.

@anathomical @happythenewsad As it currently stands, we don't have a way of knowing which current teacher users identify with this new role. In the follow-up PR to this PR, we only store a value for `role_selected_at_sign_up` if the user selects this individual contributor option upon creating an account. 

The thought behind this (after I discussed with Dan) is that although we don't currently have a planned use for this value (since the actual user role functionality is the same for both teachers and individual contributors), it's better to store this value now that we're adding this option in case we do want this data further down the road (i.e. if the functionality of these roles do end up diverging). This stems off the conversation that we had at sprint wrap-up about potentially doing a larger scoped project of cleaning up how we use `role`, `account_type` etc properties on the `user` table. 

I opted for a default empty string since it currently only applies to new users that will identify with this role at signup, and we may just end up migrating this data over to the user table. I'm open to changing it to default `'user'`, I just didn't really see what the advantage would be. Thoughts?

happythenewsad:
> 

Thanks for the explanation - carry on

PR #10152, opened 2023-01-13
Teacher Onboarding/ selecting and storing signup role

eadams17:
## WHAT
update the UI to add "Individual Contributor" signup role and add logic to store it

## WHY
we want to provide this third option for users that may be a teacher, caregiver or tutor

## HOW
add card for Individual Contributor and update backend logic to handle storing selected role in `teacher_infos` as `role_selected_at_signup` property

### Screenshots
Screen Shot 2023-01-13 at 12 45 07 PM

### Notion Card Links 
https://www.notion.so/quill/Design-of-Collecting-Teacher-Onboarding-Materials-Part-II-04073c66d0d242ee95aea8be6fe34a10

PR Checklist | Your Answer
------------ | -------------


services/QuillLMS/app/models/user.rb
```diff
@@ -86,8 +86,11 @@ class User < ApplicationRecord
   STUDENT = 'student'
   STAFF = 'staff'
   SALES_CONTACT = 'sales-contact'
-  ROLES      = [TEACHER, STUDENT, STAFF, SALES_CONTACT]
-  SAFE_ROLES = [STUDENT, TEACHER, SALES_CONTACT]
+  INDIVIDUAL_CONTRIBUTOR = 'individual-contributor'
+  ONBOARDING_ROLES   = [STUDENT, TEACHER, INDIVIDUAL_CONTRIBUTOR]
+  TEACHER_INFO_ROLES = [TEACHER, INDIVIDUAL_CONTRIBUTOR]
+  ROLES              = [TEACHER, STUDENT, STAFF, SALES_CONTACT]
+  SAFE_ROLES         = [STUDENT, TEACHER, SALES_CONTACT]
```
anathomical, line 93: Should `INDIVIDUAL_CONTRIBUTOR` be here in `SAFE_RULES`?  I'm pretty sure that "safe", in this case, means that we allow users to specify this role for themselves, something we don't do for `STAFF` because it lets you into the admin tools.

services/QuillLMS/spec/controllers/teacher_infos_controller_spec.rb
```diff
@@ -14,7 +14,6 @@
 require 'rails_helper'
 
 describe TeacherInfosController do
-  let(:user) { create(:user) }
```
anathomical, line 17: Something must have come up in the tests that's not obvious from just reading the code, but I don't see a reason why it should be necessary to move the `let(:user)` from this section into each separate `describe`.  Not a big deal either way.

anathomical (approved):
A couple of minor notes, but this looks fundamentally good to me.

services/QuillLMS/client/app/bundles/Teacher/components/accounts/new/select_user_type.jsx
```diff
@@ -1,17 +1,19 @@
...
-const teacherChalkboardImg = `${process.env.CDN_URL}/images/onboarding/teacher-chalkboard.svg`
+const studentPencilImg = `${process.env.CDN_URL}/images/onboarding/student-pencil-colored.svg`
+const teacherChalkboardImg = `${process.env.CDN_URL}/images/onboarding/teacher-chalkboard-colored.svg`
+const homeSchoolImg = `${process.env.CDN_URL}/images/onboarding/home-building-colored.svg`
 
-class SelectUserType extends React.Component {
+export class SelectUserType extends React.Component {
   setRoleOnSession = (role) => {
+    const urlParam = role === INDIVIDUAL_CONTRIBUTOR ? TEACHER : role
     requestPost(`${process.env.DEFAULT_URL}/account/role`, { role, },
```
emilia-friedberg, line 13: I would take `{ role, }` out of this line here because you aren't using it now and it allows for ambiguity with the function param.

services/QuillLMS/client/app/bundles/Teacher/components/accounts/new/select_user_type.jsx
```diff
@@ -56,12 +63,19 @@ class SelectUserType extends React.Component {
...
               imgSrc={teacherChalkboardImg}
               onClick={this.handleClickTeacher}
               text="Select this option to create classes, assign activities, and view reports."
             />
+            <Card
+              header="Parent, Tutor or Caregiver"
+              imgAlt="home"
+              imgSrc={homeSchoolImg}
+              onClick={this.handleClickIndividualContributor}
+              text="Select this option to assign activities and view reports for your individual student(s)."
```
emilia-friedberg, line 76: I get that this is the spec but doesn't this make it seem like the functionality is actually different than for a K-12 teacher? Like they still have to make a class. Not saying you should do anything differently, just commenting.

emilia-friedberg (approved):
This looks good to me with one small change requested - is the part of the spec where this user type skips the school selection screen coming in a later PR?

services/QuillLMS/app/models/user.rb
eadams17: @anathomical since this isn't really a "role" in that the user still has a "teacher" role for their user entry, I didn't think it should be in the `SAFE_ROLES` array. Maybe it makes more sense to have the constants used for this feature stored on the `TeacherInfo` class since that's where this role is actually being stored?

services/QuillLMS/client/app/bundles/Teacher/components/accounts/new/select_user_type.jsx
emilia-friedberg: Nevermind! Misread that!

services/QuillLMS/spec/controllers/teacher_infos_controller_spec.rb
eadams17: I think I was anticipating an issue with how we use `current_user` in this controller and how it would affect the tests with these changes, but that's not actually the case-- I'll revert this

PR #10189, opened 2023-01-24
only cache-affecting backend updates for diagnostic progress indicators

emilia-friedberg:
## WHAT
Adding the `unit_template_id` to the standard `diagnostic_unit_record` query.

## WHY
This is data that we need for the diagnostic progress indicator project, but this query is cached, so I'm hoping to ship this change 24 hours before we launch the rest of the project so we don't have to deal with invalidating the cache manually. It is only additive, so it shouldn't be a problem to ship it early.

## HOW
Just add an extra join to the table.

### Screenshots

### Notion Card Links
https://www.notion.so/quill/Diagnostic-Progress-Indicator-919c78414c0c4da8b8c640db65102c21

PR Checklist | Your Answer
------------ | -------------


anathomical (approved):
Good solution to the cache expiration problem.

brendanshean (approved):
Looks good!  Nice solution

PR #10205, opened 2023-01-26
Sub navbar updates

eadams17:
## WHAT
update the remaining subnavbars around the Quill site to match the standard design

## WHY
we want to have design continuity for these subnavbars

## HOW
created a shared `SubNavbar` component to render for these partials as well as update/add helpers for rendering the tab data

### Screenshots
Screen Shot 2023-01-26 at 6 08 10 PM
Screen Shot 2023-01-26 at 6 08 24 PM
Screen Shot 2023-01-26 at 6 08 40 PM
Screen Shot 2023-01-26 at 6 09 08 PM

### Notion Card Links
https://www.notion.so/quill/Update-Headers-for-Other-Pages-on-Quill-cc620c0879f54ce1a5fb6f914d13116f

PR Checklist | Your Answer
------------ | -------------


services/QuillLMS/app/views/blog_posts/show.html.erb
```diff
@@ -4,8 +4,9 @@
   <% active_tab = 'All resources'%>
 <% end %>
 
-<%= render partial: 'teachers/shared/scorebook_tabs' if current_user %>
-<%= render partial: 'pages/shared/teacher_center_navbar', locals: {active_tab: active_tab} %>
+<%= render partial: 'teachers/shared/scorebook_tabs' if current_user && current_user.role != 'student' %>
+<%= render partial: 'pages/shared/teacher_center_navbar', locals: {active_tab: active_tab} if (!current_user || current_user&.role == 'teacher') %>
```
brendanshean, line 8: I think you can remove the safe navigation on this one: `(!current_user || current_user.role == 'teacher')`

services/QuillLMS/app/views/pages/shared/_teacher_center_navbar.html.erb
```diff
@@ -1,21 +1,5 @@
-<%-# required local vars: active_tab %>
```
brendanshean, line 1: Nice refactoring here!

services/QuillLMS/app/views/pages/shared/_tools_anchor_navbar.html.erb
```diff
@@ -1,82 +1,5 @@
-<div class="subnav tools-navbar">
```
brendanshean, line 1: Nice cleanup here too!

services/QuillLMS/client/app/bundles/Tools/startup/subNavbarAppClient.tsx
```diff
@@ -0,0 +1,8 @@
+import React from 'react';
+import { SubNavbar } from '../../Shared';
+
+const SubNavbarApp = (props) => (
```
brendanshean, line 4: Is the reason for the inconsistency between the name of the file `subNavbarAppClient` and the const name `subNavbarApp` a historical, conventional, required?  (It kind of breaks the symmetry with naming in other files.)

brendanshean (approved):
There's some spec failures that need to be cleaned up but I don't think they're anything requiring a re-review.  Thank you for cleaning up the code here.

emilia-friedberg (approved):
This code is much cleaner, and I like the use of the `windowSize` React hook. One downside to this approach that may be worth considering: Dan brought up in the last engineering meeting (and I don't think you were there) that we might want to minimize our use of React, at least on public-facing pages, since it takes longer to load. I think it makes sense to release this change as is for now, but we may want to revisit it if we do commit to reducing our use of React on those pages.

services/QuillLMS/client/app/bundles/Tools/startup/subNavbarAppClient.tsx
eadams17: So if you look at newer files, @emilia-friedberg and I have been following a naming convention of PascalCase for component naming and camelCase for files.

Official React docs are unopinionated, but the general consensus is that component naming should be PascalCase, whereas file naming is more ambiguous: https://stackoverflow.com/questions/43973199/file-naming-conventions-in-reactjs

I think we could maybe have a larger discussion about this an engineering meeting though since this isn't explicitly documented anywhere.

brendanshean: My question was not about the camelcasing but that the const is called `SubNavbarApp` and the file is named `subNavbarAppClient` (i.e. client is used in one but not the other).

eadams17: Oh, right-- that's just a historical convention. We could probably do some cleanup there during the summer months to make all the naming conventions consistent

PR #10214, opened 2023-01-30
lms/add-stripe-subscription-id

anathomical:
## WHAT
Add migration to create Subscription.stripe_subscription_id column
## WHY
We have historically been using `stripe_invoice_id` as a proxy for a bunch of Stripe-related values.  (It is, for instance, how we determine if someone paid for their subscription with a credit card.)  However, we're about to start using Stripe invoices in an additional way that isn't compatible with a bunch of these assumptions (specifically, we'll allow for the use of invoices without credit card payments).  In order to facilitate this logic, we want to begin storing a different value that we think represents a better proxy for our assumptions and we've settled in the Stripe `subscription_id` value.
## HOW
Just add a migration to create this column in the table.  There will be follow-up PRs to begin using this value once it is available.

### Notion Card Links
https://www.notion.so/quill/Modify-Stripe-subscription-storage-to-use-stripe_subscription_id-in-addition-to-stripe_invoice_id-64273f4887a24ccd9b81d2fbc983ef3e

PR Checklist | Your Answer
------------ | -------------
Have you added and/or updated tests? |  No test changes, but factory updated


brendanshean (approved):
Looks good to me.

You could add a validation in the model file similar to stripe_invoice_id:
`validates :stripe_subscription_id, allow_blank: true, stripe_uid: { prefix: :sub }`

eadams17 (approved):

happythenewsad (approved):

brendanshean (approved):
Nice work on this!

PR #10215, opened 2023-01-30
lms/isolate-batch-worker

anathomical:
## WHAT
Move BatchAssignRecommendationsWorker to MIGRATION queue temporarily
## WHY
So that we can test its impact on our Sidekiq work in isolation
## HOW
Update the configuration for the job so that it goes to another queue.  (Discussion with @brendanshean about the priority of this job led to us both agreeing that this should be safe to put in its own queue even if it's in the critical queue now because being by itself, no other jobs will block it.)

### Notion Card Links
https://www.notion.so/quill/Out-of-Memory-errors-for-background-workers-7fbcb78248754df5beb4ba099511b5cd

PR Checklist | Your Answer
------------ | -------------
Have you added and/or updated tests? |  No tests for Sidekiq priority

### DIFF
--- a/services/QuillLMS/app/workers/batch_assign_recommendations_worker.rb
+++ b/services/QuillLMS/app/workers/batch_assign_recommendations_worker.rb
@@ -3,7 +3,7 @@
 class BatchAssignRecommendationsWorker
   include Sidekiq::Worker
 
-  sidekiq_options queue: SidekiqQueue::CRITICAL
+  sidekiq_options queue: SidekiqQueue::MIGRATION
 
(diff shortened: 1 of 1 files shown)


happythenewsad (approved):

dandrabik (approved):
The code change looks good to me.

One thing to consider: if there are a lot of these jobs to test via isolation, It would be quicker to test them if you could just change an ENV variable to isolate a job class, since that wouldn't require multiple code changes, reviews, and deploys. That flexibility might be worth the writing that code upfront. Happy to chat through how to do that.

PR #10221, opened 2023-01-31
Log lock conflicts with SaveUserPackSequenceItemsWorker

brendanshean:
## WHAT
Add `on_conflict: :log` to the SaveUserPackSequenceItemsWorker's sidekiq-unique-jobs configuration.

## WHY
On conflict will log a conflict message when a worker is attempted that is already in the queue.

## HOW
Add to the configuration section of the worker.

### Screenshots

### Notion Card Links

PR Checklist | Your Answer
------------ | -------------
Have you added and/or updated tests? |  Manual testing on staging.

### DIFF
--- a/services/QuillLMS/app/workers/save_user_pack_sequence_items_worker.rb
+++ b/services/QuillLMS/app/workers/save_user_pack_sequence_items_worker.rb
@@ -4,7 +4,9 @@ class SaveUserPackSequenceItemsWorker
   include Sidekiq::Worker
 
   sidekiq_options queue: SidekiqQueue::DEFAULT,
-    lock: :until_executed
+    lock: :until_executed,
+    on_conflict: :log,
+    lock_ttl: 2.minutes
 
   def perform(classroom_id, user_id)
     return if classroom_id.nil? || user_id.nil?
--- a/services/QuillLMS/app/workers/save_user_pack_sequence_items_worker.rb
+++ b/services/QuillLMS/app/workers/save_user_pack_sequence_items_worker.rb
@@ -5,8 +5,7 @@ class SaveUserPackSequenceItemsWorker
 
   sidekiq_options queue: SidekiqQueue::DEFAULT,
     lock: :until_executed,
-    on_conflict: :log,
-    lock_ttl: 2.minutes
+    on_conflict: :log
 
   def perform(classroom_id, user_id)
     return if classroom_id.nil? || user_id.nil?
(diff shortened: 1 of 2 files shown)


anathomical (approved):

happythenewsad (approved):

services/QuillLMS/app/workers/save_user_pack_sequence_items_worker.rb
```diff
@@ -4,7 +4,9 @@ class SaveUserPackSequenceItemsWorker
   include Sidekiq::Worker
 
   sidekiq_options queue: SidekiqQueue::DEFAULT,
-    lock: :until_executed
+    lock: :until_executed,
```
dandrabik, line 7: Note, your config options here are valid, but some combinations of config options can be invalid, so the gem provides [built-in tests](https://github.com/mhenrixon/sidekiq-unique-jobs#validating-worker-configuration) to check for that. Probably worth seeing if these tests work as a backstop against future changes.
Screenshot 2023-01-31 at 4 35 09 PM

services/QuillLMS/app/workers/save_user_pack_sequence_items_worker.rb
```diff
@@ -4,7 +4,9 @@ class SaveUserPackSequenceItemsWorker
...
+    on_conflict: :log,
+    lock_ttl: 2.minutes
```
dandrabik: What is the reason for the lock_ttl? Is it a guard against locks that aren't cleaned up properly (for some unknown reason)?

If I'm reading the docs correctly, this `lock_ttl: 2.minutes` means that if a `SaveUserPackSequenceItemsWorker` takes 3 minutes to perform, if another `SaveUserPackSequenceItemsWorker` with the same params hits the queue at 2 minutes into processing, it will start processing as well, so two of them will be running at the same time. 

Looking at our transaction logs, we have had jobs take over 2 minutes. If this is solely a protection against uncleaned-up locks, we might want to make the `lock_ttl` longer. (Or remove it).

Screenshot 2023-01-31 at 4 25 06 PM

dandrabik (approved):
I made some general comments for your consideration, but will defer to any decisions you make :+1:

services/QuillLMS/app/workers/save_user_pack_sequence_items_worker.rb
brendanshean: Good idea.  I didn't even notice this part of the docs.   I think I was stuck on the section right below on uniqueness:  "This has been probably the most confusing part of this gem. People get really confused with how unreliable the unique jobs have been."

services/QuillLMS/app/workers/save_user_pack_sequence_items_worker.rb
brendanshean: I'm going to remove the `lock_ttl` for now.  I need to look into possibly configuring the reaper.

PR #10223, opened 2023-02-01
teacher center admin style updates

emilia-friedberg:
## WHAT
Update the design of the teacher center article admin pages.

## WHY
These are some of the most-neglected pages on the site, and deserve to be easier to use and look at.

## HOW
Mostly CSS/HTML changes. We were going to replace the Markdown text editor here with the DraftJS one we use for activity editing, but that one a) doesn't have video or image capabilities yet and b) converts to HTML, not Markdown, so we're splitting that off into a larger project to attack at a later date.

### Screenshots
Screen Shot 2023-02-01 at 10 40 37 AM
Screen Shot 2023-02-01 at 10 40 55 AM
Screen Shot 2023-02-01 at 10 41 05 AM

### Notion Card Links
https://www.notion.so/quill/Teacher-Center-Admin-Redesign-f52ad9e9226642e7869fd0eb20ad1bd7

PR Checklist | Your Answer
------------ | -------------


services/QuillLMS/config/database.yml
```diff
@@ -27,13 +27,15 @@ heroku_env: &heroku_env
 
 development:
   primary:
-    <<: *local_shared
-    database: emp_gr_development
+    # <<: *local_shared
```
cissyxyu: just curious -- what's the reasoning behind this change?

cissyxyu (approved):

services/QuillLMS/config/database.yml
emilia-friedberg: This is what I do when I want to use the staging DB instead of my local one (you have to set the `DATABASE_URL` in your .env file) but it wasn't supposed to be committed!

services/QuillLMS/client/app/bundles/Teacher/components/cms/blog_posts/featured_blog_posts.tsx
```diff
@@ -1,42 +1,48 @@
...
+    updateOrder(sortInfo)
+  }
+
+  const rows = blogPostRows(featuredBlogPosts.sort((bp1, bp2) => bp1.featured_order_number - bp2.featured_order_number), handleClickStar, false)
+
+  let saveOrderButton = (<button className="quill-button fun contained primary focus-on-light disabled" disabled type="button">Save Order</button>)
+
+  if (orderHasChanged) {
+    saveOrderButton = (<button className="quill-button fun contained primary focus-on-light" onClick={saveOrder} type="button">Save Order</button>)
+  }
```
eadams17: This is small, but since we have this same logic in the `BlogPostTable`, we could just pull this out into a shared function in the blog_posts shared file. Something like this:

```
  function saveOrderButton({ saveOrder, orderHasChanged }) {
    const className = `quill-button fun contained primary focus-on-light ${orderHasChanged ? '' : 'disabled'}`
    return <button className={className} disabled={!orderHasChanged} onClick={saveOrder} type="button">Save Order</button>
  }
  ```

eadams17 (approved):
This looks sooo much better-- thanks for putting this together!

PR #10225, opened 2023-02-01
Move feedback history session CSV download into an email job

cissyxyu:
## WHAT
Move the feedback history session CSV download into an asynchronous email rather than a synchronous endpoint on the staff page.

## WHY
As we get more and more Evidence data, we're going to have difficulty loading the data we need on this page and encountering timeouts. To prevent this, we'll move it to a job that aggregates the data and then emails a CSV to the admin who requested it.

## HOW
Create a sidekiq job that aggregates the necessary data and then generates a CSV. Send this CSV via UserMailer to the pertinent admin.

### Screenshots
Screen Shot 2023-02-02 at 2 04 40 AM

Screen Shot 2023-02-02 at 2 08 16 AM

### Notion Card Links
https://www.notion.so/quill/Product-Board-30f97e4eb01246dbb8264253fb385073?p=3fd2c1db70f045a5b44b938afe0b7899&pm=c

PR Checklist | Your Answer
------------ | -------------

### DIFF
--- a/services/QuillLMS/app/controllers/api/v1/session_feedback_histories_controller.rb
+++ b/services/QuillLMS/app/controllers/api/v1/session_feedback_histories_controller.rb
@@ -18,8 +18,14 @@ class Api::V1::SessionFeedbackHistoriesController < Api::ApiController
 
   def email_csv_data
     options = params.permit(:activity_id, :start_date, :end_date, :filter_type, :responses_for_scoring).to_h.symbolize_keys
-    email = current_user.email
-    InternalTool::EmailFeedbackHistorySessionDataWorker.perform_async(options[:activity_id], options[:start_date], options[:end_date], options[:filter_type], options[:responses_for_scoring], email)
+    InternalTool::EmailFeedbackHistorySessionDataWorker.perform_async(
+      options[:activity_id],
+      options[:start_date],
+      options[:end_date],
+      options[:filter_type],
+      options[:responses_for_scoring],
+      current_user.email
+    )
     render plain: 'OK'
   end
 
--- a/services/QuillLMS/app/controllers/api/v1/session_feedback_histories_controller.rb
+++ b/services/QuillLMS/app/controllers/api/v1/session_feedback_histories_controller.rb
@@ -19,7 +19,6 @@ class Api::V1::SessionFeedbackHistoriesController < Api::ApiController
   def email_csv_data
     options = params.permit(:activity_id, :start_date, :end_date, :filter_type, :responses_for_scoring).to_h.symbolize_keys
     email = current_user.email
-    logger.debug options
     InternalTool::EmailFeedbackHistorySessionDataWorker.perform_async(options[:activity_id], options[:start_date], options[:end_date], options[:filter_type], options[:responses_for_scoring], email)
     render plain: 'OK'
   end
--- a/services/QuillLMS/app/controllers/api/v1/session_feedback_histories_controller.rb
+++ b/services/QuillLMS/app/controllers/api/v1/session_feedback_histories_controller.rb
@@ -16,17 +16,12 @@ class Api::V1::SessionFeedbackHistoriesController < Api::ApiController
     }
   end
 
-  def session_data_for_csv
+  def email_csv_data
     options = params.permit(:activity_id, :start_date, :end_date, :filter_type, :responses_for_scoring).to_h.symbolize_keys
-    feedback_histories = FeedbackHistory.session_data_for_csv(**options)
-    results = []
-    feedback_histories.find_each(batch_size: 10_000) { |feedback_history| results << feedback_history.serialize_csv_data }
-    results.sort! { |a,b| b["datetime"] <=> a["datetime"] }
-    if results
-      render json: results
-    else
-      render plain: "The resource you were looking for does not exist", status: 404
-    end
+    email = current_user.email
+    logger.debug options
+    InternalTool::EmailFeedbackHistorySessionDataWorker.perform_async(options[:activity_id], options[:start_date], options[:end_date], options[:filter_type], options[:responses_for_scoring], email)
+    render plain: 'OK'
   end
 
   # GET /feedback_histories/1.json
--- a/services/QuillLMS/app/mailers/user_mailer.rb
+++ b/services/QuillLMS/app/mailers/user_mailer.rb
@@ -152,7 +152,17 @@ class UserMailer < ActionMailer::Base
     csv = CSV.generate(headers: true) do |csv_body|
       csv_body << FEEDBACK_HISTORY_CSV_HEADERS
       data.each do |row|
-        csv_body << [row["datetime"], row["session_uid"], row["conjunction"], row["attempt"], row["optimal"], (row['optimal'] || row['attempt'] == DEFAULT_MAX_ATTEMPTS).to_s, row["response"], row["feedback"], "#{row['feedback_type']}: #{row['name']}"]
+        csv_body << [
+          row["datetime"],
+          row["session_uid"],
+          row["conjunction"],
+          row["attempt"],
+          row["optimal"],
+          row['optimal'] || row['attempt'] == DEFAULT_MAX_ATTEMPTS,
+          row["response"],
+          row["feedback"],
+          "#{row['feedback_type']}: #{row['name']}"
+        ]
       end
     end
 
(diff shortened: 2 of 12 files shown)


services/QuillLMS/app/mailers/user_mailer.rb
```diff
@@ -146,6 +148,18 @@ def ell_starter_diagnostic_info_email(name, email)
     mail from: "The Quill Team <hello@quill.org>", to: email, subject: "ELL Starter Diagnostic Next Steps"
   end
 
+  def feedback_history_session_csv_download(email, data)
+    csv = CSV.generate(headers: true) do |csv_body|
+      csv_body << FEEDBACK_HISTORY_CSV_HEADERS
+      data.each do |row|
+        csv_body << [row["datetime"], row["session_uid"], row["conjunction"], row["attempt"], row["optimal"], (row['optimal'] || row['attempt'] == DEFAULT_MAX_ATTEMPTS).to_s, row["response"], row["feedback"], "#{row['feedback_type']}: #{row['name']}"]
```
brendanshean: For readability, I like to break method calls with 3+ parameters into multiple lines:

```
  csv_body << [
    row["datetime"], 
    row["session_uid"], 
    row["conjunction"], 
    row["attempt"], 
    row["optimal"], 
    row['optimal'] || row['attempt'] == DEFAULT_MAX_ATTEMPTS, 
    row["response"], 
    row["feedback"], 
    "#{row['feedback_type']}: #{row['name']}"
  ]
```

Also, I think the `to_s` on the boolean is unnecessary since the `CSV#generate` should call to_s internally.

services/QuillLMS/app/controllers/api/v1/session_feedback_histories_controller.rb
```diff
@@ -16,17 +16,11 @@ def index
...
-    results = []
-    feedback_histories.find_each(batch_size: 10_000) { |feedback_history| results << feedback_history.serialize_csv_data }
-    results.sort! { |a,b| b["datetime"] <=> a["datetime"] }
-    if results
-      render json: results
-    else
-      render plain: "The resource you were looking for does not exist", status: 404
-    end
+    email = current_user.email
+    InternalTool::EmailFeedbackHistorySessionDataWorker.perform_async(options[:activity_id], options[:start_date], options[:end_date], options[:filter_type], options[:responses_for_scoring], email)
```
brendanshean: For readability, I like to break method calls with 3+ parameters into multiple lines:
```
  InternalTool::EmailFeedbackHistorySessionDataWorker.perform_async(
    options[:activity_id], 
    options[:start_date], 
    options[:end_date], 
    options[:filter_type], 
    options[:responses_for_scoring], 
    current_user.email
  )
```

Also, no need for the temp variable `email` as it's only used as a parameter

services/QuillLMS/app/workers/internal_tool/email_feedback_history_session_data_worker.rb
```diff
@@ -0,0 +1,16 @@
+# frozen_string_literal: true
+
+class InternalTool::EmailFeedbackHistorySessionDataWorker
+  include Sidekiq::Worker
+  sidekiq_options queue: SidekiqQueue::LOW
+
+  def perform(activity_id, start_date, end_date, filter_type, responses_for_scoring, email)
+    feedback_histories = FeedbackHistory.session_data_for_csv({activity_id: activity_id, start_date: start_date, end_date: end_date, filter_type: filter_type, responses_for_scoring: responses_for_scoring})
```
brendanshean: For readability, I like to break method calls with 3+ parameters into multiple lines:

```
 feedback_histories = FeedbackHistory.session_data_for_csv(
    activity_id: activity_id, 
    start_date: start_date, 
    end_date: end_date, 
    filter_type: filter_type, 
    responses_for_scoring: responses_for_scoring
)
```

services/QuillLMS/spec/mailers/user_mailer_spec.rb
```diff
@@ -121,4 +121,35 @@
...
+          csv << [row["datetime"], row["session_uid"], row["conjunction"], row["attempt"], row["optimal"], (row['optimal'] || row['attempt'] == 5).to_s, row["response"], row["feedback"], "#{row['feedback_type']}: #{row['name']}"]
+        end
+      end
+
+      expect(mail.to).to eq(["team@quill.org"])
+      expect(mail.subject).to match("Feedback Sessions CSV Download")
+      expect(mail.attachments['feedback_sessions.csv'].mime_type).to match('text/csv')
+      expect(CSV.parse(mail.attachments['feedback_sessions.csv'].body.raw_source)).to eq(CSV.parse(csv_body))
+    end
+  end
```
brendanshean, line 177: Here's a refactoring of the spec with comments for clarifiation:

```
  describe 'feedback_history_session_csv_download' do
    # I like to structure specs starting with subject which matches the describe block
    subject { described_class.feedback_history_session_csv_download(email, data) }

    # factor out parameters provided to subject as let variables
    let(:email) { 'team@quill.org' }

    let(:data) do
      [
        {
          "datetime": "20220701",
          "session_uid": "sessionuid",
          "conjunction": "but",
          "attempt": "1",
          "optimal": "false",
          "response": "this is a test response",
          "feedback": "test feedback",
          "feedback_type": "spelling"
        }
      ]
    end

    # Add some constants to UserMailer to make explicit the coupling with the spec: 
    let(:csv_headers) { described_class::FEEDBACK_HISTORY_CSV_HEADERS }
    let(:csv_attachment) { subject.attachments[described_class::FEEDBACK_SESSIONS_CSV_FILENAME] }

    it 'should set the subject, receiver and the sender' do
      csv_body = CSV.generate(headers: true) do |csv|
        csv << csv_headers
        data.each do |row|
          #  break up multiple parameter method into multiple lines for readability
          csv << [
            row["datetime"],
            row["session_uid"],
            row["conjunction"],
            row["attempt"],
            row["optimal"],
            row['optimal'] || row['attempt'] == described_class::MAX_ATTEMPTS,
            row["response"],
            row["feedback"],
            "#{row['feedback_type']}: #{row['name']}"
          ]
        end
      end

      expect(subject.to).to eq [email]

      # Refer to constants to make coupling with string explicit
      expect(subject.subject).to match(described_class::FEEDBACK_SESSIONS_CSV_DOWNLOAD)
      expect(csv_attachment.mime_type).to match('text/csv')
      expect(CSV.parse(csv_attachment.body.raw_source)).to eq(CSV.parse(csv_body))
    end
  end
```

brendanshean (approved):
It makes a lot of sense moving this into a background job.  Thank you for making the change.

services/QuillLMS/client/app/bundles/Staff/utils/evidence/activityAPIs.ts
```diff
@@ -119,18 +119,15 @@ export const fetchActivitySessions = async ({ queryKey, }) => {
...
-
-  return {
-    csvResponseData,
-    error: handleApiError('Failed to fetch activity sessions, please refresh the page.', response),
-  };
+  const response = await mainApiFetch(url, {method: 'POST'});
+  if (!response.ok) {
+    handleApiError('Failed to initiate CSV download email, please try again', response)
+  } else {
+    alert("Data load has been initiated. The CSV Data will be sent to your email.")
```
eadams17: I think this would be a good opportunity to use a Snackbar to have a cleaner alert message design. You could return this success message here, and then add logic to set that message for the Snackbar in `handleLoadCSVDataClick()` (you can see an example [here](https://github.com/empirical-org/Empirical-Core/blob/develop/services/QuillLMS/client/app/bundles/Staff/components/newAdminOrDistrictUser.tsx) of how that would look)

services/QuillLMS/spec/mailers/user_mailer_spec.rb
brendanshean: That way specs kind of a have a structure of
```
describe ...
  subject ...
  let..
  let...

  it do
    expect(subject).to ....
  end
end
```
Moving the setup outside of the `it` block allows you to think of the test in three parts:  
1) setup 
2) test 
3) cleanup (this tends to be handled automatically)

Abstracting out setup allows for refactoring similar tests that only differ in setup.

eadams17 (approved):
This was already getting close to the timeout limit for some of the older activities, thanks for making this change!

emilia-friedberg (approved):

PR #10233, opened 2023-02-02
Lms/add admin role

anathomical:
## WHAT
Add a new `admin` value for `User.role` that will be used to flag Users as school or district admins.  Historically, we have left admins with the role of `teacher`, and just determined the user to be an admin if they have any records in the `schools_admins` table.  NOTE: while a user can have the role `admin` to get to the admin dashboard, if they don't have any entries in the `schools_admins` table, the dashboard is broken and loads empty.
## WHY
As we begin treating admins more like a real user class, we should differentiate them properly.  That said, we don't want existing user behavior to change, so admins should still have access to all the various teacher-specific dashboards and reports.
## HOW
- Add a new value for `User.role`
- Trace through a bunch of auth pathways to ensure that users with the `admin` role are authorized to view pages that have historically required `teacher` roles
- Add an `after_save` callback to the `SchoolsAdmins` model to ensure that any time a User is assigned as an admin for a School (the way our current code denotes admins) it updates the `User.role` value to its new expectation.

### Notion Card Links
https://www.notion.so/quill/Create-a-new-admin-user-role-and-migrate-existing-admins-19d6e4118a3d41a0a626f3c571c38793

PR Checklist | Your Answer
------------ | -------------

### DIFF
--- a/services/QuillLMS/app/models/concerns/teacher.rb
+++ b/services/QuillLMS/app/models/concerns/teacher.rb
@@ -24,7 +24,7 @@ module Teacher
     delegate :first, :find, :where, :all, :count, to: :scope
 
     def scope
-      User.where(role: User::TEACHER)
+      User.teacher
     end
   end
 
--- a/services/QuillLMS/app/models/concerns/teacher.rb
+++ b/services/QuillLMS/app/models/concerns/teacher.rb
@@ -24,7 +24,7 @@ module Teacher
     delegate :first, :find, :where, :all, :count, to: :scope
 
     def scope
-      User.where(role: 'teacher')
+      User.where(role: User::TEACHER)
     end
   end
 
@@ -359,7 +359,7 @@ module Teacher
   def update_teacher params
     return if !teacher?
 
-    params[:role] = 'teacher' if params[:role] != 'student'
+    params[:role] = User::TEACHER if params[:role] != User::STUDENT
     params.permit(
       :id,
       :name,
--- a/services/QuillLMS/app/workers/populate_annual_vitally_worker.rb
+++ b/services/QuillLMS/app/workers/populate_annual_vitally_worker.rb
@@ -21,7 +21,7 @@ class PopulateAnnualVitallyWorker
   end
 
   def schools_to_sync
-    School.select(:id).distinct.joins(:users).where('users.role = ?', 'teacher')
+    School.select(:id).distinct.joins(:users).merge(User.teacher)
   end
 
   def users_to_sync
--- a/services/QuillLMS/spec/controllers/teachers_controller_spec.rb
+++ b/services/QuillLMS/spec/controllers/teachers_controller_spec.rb
@@ -19,8 +19,7 @@ describe TeachersController, type: :controller do
       end
 
       it 'render admin dashboard' do
-        user = create(:user)
-        user.schools_admins.create
+        user = create(:user, role: User::ADMIN)
         allow(controller).to receive(:current_user) { user }
         get :admin_dashboard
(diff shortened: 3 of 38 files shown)


emilia-friedberg (commented):
This approach to the auth etc looks good to me, and I really appreciate you adding the logic around turning teachers into admins once they have an associated `school_admin` record. I think we might want to do the same thing for `district_admins`. I also think that, since we're trying to preserve all the teacher behavior for admins, there is a larger sweep we'll need to do for all the places we call `.teacher?` or check that a user's role is `teacher` in the codebase, to update them to call something more like `.teacher_or_admin?`.

services/QuillLMS/app/workers/populate_annual_vitally_worker.rb
```diff
@@ -21,7 +21,7 @@ def perform
   end
 
   def schools_to_sync
-    School.select(:id).distinct.joins(:users).where('users.role = ?', 'teacher')
+    School.select(:id).distinct.joins(:users).merge(User.teacher)
```
emilia-friedberg, line 24: Never seen this before, cool!

emilia-friedberg (approved):
LGTM! Thanks so much for doing all that.

services/QuillLMS/app/workers/populate_annual_vitally_worker.rb
anathomical: It was new for me, too!  I was trying to figure out how to use a `scope` with `joins`, and Google surfaced this technique.  It feels a lot cleaner to me.

dandrabik (approved):
Thanks for tracking all of these items down. Looks good to me as long as you tested this in the browser a bit.

One note on naming, you've flagged that `.teacher?` is weird. I think it's fine for now, but in the future, we could consider a refactor to rename around `authorization`, e.g. something like `teacher_authorized?`, `teacher_access?`, `teacher_permissions?`, etc. It might read better long-term for people who are non-teachers.

services/QuillLMS/app/models/concerns/teacher.rb
```diff
@@ -24,7 +24,7 @@ class << self
     delegate :first, :find, :where, :all, :count, to: :scope
 
     def scope
-      User.where(role: 'teacher')
+      User.where(role: User::TEACHER)
```
brendanshean: What about `User.teacher`?

services/QuillLMS/app/workers/populate_annual_vitally_worker.rb
brendanshean: Cool.  I haven't seen this either.

services/QuillLMS/spec/controllers/teachers_controller_spec.rb
```diff
@@ -19,8 +19,7 @@
       end
 
       it 'render admin dashboard' do
-        user = create(:user)
-        user.schools_admins.create
+        user = create(:user, role: User::ADMIN)
```
brendanshean: To keep with the pattern in users.rb, I would create a new factory

```
factory :admin do
  role { User::ADMIN }
end
```

The interesting question is do you embed this inside of the `factory :teacher` block definition.

Then you have `user = create(:admin)` which cuts down on boilerplate and standardizes the factory construction for easier refactoring later.

brendanshean: There are a few more of this admin construction in the specs of this in the PR.

brendanshean (approved):
Thanks for cleaning up this role logic!

PR #10236, opened 2023-02-03
Global footer updates

eadams17:
## WHAT
update the footers globally around the Quill site

## WHY
we want continuity in design for these footers across all pages

## HOW
combine footers into single footer (since the designs are very similar between authenticated and non-authenticated users), update links/ ordering, update logos and add logic for conditionally rendering differing elements

### Screenshots
Screen Shot 2023-02-03 at 11 43 32 AM
Screen Shot 2023-02-03 at 11 50 00 AM
Screen Shot 2023-02-03 at 11 52 14 AM

### Notion Card Links
https://www.notion.so/quill/New-Global-Footer-Updates-39bca72c9fe144d480f53b45dd4ee085

PR Checklist | Your Answer
------------ | -------------


services/QuillLMS/app/views/layouts/application.html.erb
```diff
@@ -39,7 +39,7 @@
      client side hydration -->
     <%= redux_store_hydration_data %>
     <% if current_user %>
-      <%= render partial: 'footer', media: 'all', 'data-turbolinks-track' => true %>
+      <%= render partial: 'pages/shared/footer', media: 'all', 'data-turbolinks-track' => true %>
     <% else %>
       <%= render partial: 'pages/shared/footer' %>
```
anathomical: Both sides of this conditional look like they do the same thing.  Should we get rid of this `if` split altogether?

services/QuillLMS/app/views/pages/shared/_footer.html.erb
```diff
@@ -1,126 +1,129 @@
...
-      <p>We’re helping every student succeed thanks to our partners:</p>
-      <a href="/about">
-        <div class="partner-links">
-          <img class="lazyload" src="https://assets.quill.org/images/homepage/gates_logo.svg" alt='Gates Foundation Logo'/>
-          <img class="lazyload" src="https://assets.quill.org/images/homepage/google_logo.svg" alt='Google.org Logo'/>
-          <img class="lazyload" src="https://assets.quill.org/images/homepage/footer-att.svg" alt='AT&T Logo'/>
-          <img class="lazyload" src="https://assets.quill.org/images/homepage/footer-fast-forward.svg" alt='Fast Forward Logo'/>
-          <img class="lazyload" src="https://assets.quill.org/images/homepage/footer-nbc.svg" alt='NBC Logo'/>
-          <img class="lazyload" src="https://assets.quill.org/images/homepage/footer-robin-hood.svg" alt='Robinhood Logo'/>
+  <% if current_user&.role == 'student' || current_user&.role == 'staff' %>
```
anathomical: I suspect that this logic is copied from another file, but I'm trying to get us to standardize on `current_user&.student?` and `current_user&.staff?` rather than doing direct comparisons to `role`, so would like to see this updated to `if current_user&.student? || current_user&.staff?`

services/QuillLMS/app/views/pages/shared/_footer.html.erb
```diff
@@ -1,126 +1,129 @@
...
+          <li>
+            <a href="/tools/proofreader">Quill Proofreader</a>
+          </li>
+          <li>
+            <a href="/tools/grammar">Quill Grammar</a>
+          </li>
+          <li>
+            <a href="/tools/evidence">Quill Reading for Evidence</a>
+          </li>
+        </ul>
```
anathomical: I don't know if it would be an improvement, but did you consider putting all of these blocks into partials the way we did with `explore_activities_footer` and so on?  Just to keep this file stripped down?  Could see that going either way.

anathomical (approved):
A couple of small notes, but looks generally solid to me.

services/QuillLMS/app/views/pages/shared/_footer.html.erb
emilia-friedberg: Agree with @anathomical. I think it would be even cleaner to declare a partial that's just something like `list_of_links` that can just accept an array of hashes with urls and names, and then iterate through to make all of those `<li><a>{...}</a></li>` tags, since we use them a lot throughout these footers.

emilia-friedberg (approved):

services/QuillLMS/app/views/pages/shared/_footer.html.erb
eadams17: good call, I'll just make a shared `list_of_links_with_header` partial and add a helper class for the link data

PR #10238, opened 2023-02-06
internal tool updates to support new admin self-onboarding flow

emilia-friedberg:
## WHAT
Internal tool updates to support new admin self-onboarding flow; namely: exposing admin verification status and sub-role in the user manager, and building a tool to allow staff members to grant or deny admin verification requests.

## WHY
These are requisite parts of the admin approval process.

## HOW
Just update the form and controller action for the user manager; for the new admin authenticity verification tool I built out the controller endpoints and then the frontend interface to spec.

### Screenshots

### Notion Card Links
https://www.notion.so/quill/Create-new-admin-related-user-properties-and-expose-them-in-the-user-manager-4a7800125a2b4742805fd4b48b7f968f
https://www.notion.so/quill/Implement-the-internal-tool-for-admin-authenticity-verification-e35979fced3d4b2abe1a8274549a2f18

PR Checklist | Your Answer
------------ | -------------
Have you deployed to Staging? | Not yet - going to have PS review everything all at once


services/QuillLMS/client/app/bundles/Staff/components/adminVerification.tsx/undoModal.tsx
```diff
@@ -0,0 +1,19 @@
...
+      </div>
+      <div className="button-section">
+        <button className="quill-button medium secondary outlined focus-on-light" onClick={closeModal} type="button">Cancel</button>
+        <button className="quill-button medium primary contained focus-on-light" onClick={undo} type="button">Confirm</button>
+      </div>
+    </div>
+  </div>
+)
+
+export default UndoModal
```
eadams17, line 14: I feel like these three modals could just be combined into one shared `AdminVerificationModal` and passed args like these: `({ headerText, bodyText, clickAction, closeModal )}`

eadams17 (approved):
Looks great! Just one comment about combining the modals into a shared component

cissyxyu (approved):

services/QuillLMS/app/controllers/cms/admin_verification_controller.rb
```diff
@@ -0,0 +1,63 @@
...
+      format.json {
+        pending = admin_records_by_approval_status(AdminInfo::PENDING)
+        completed = admin_records_by_approval_status([AdminInfo::APPROVED, AdminInfo::DENIED])
+        render json: { pending: pending, completed: completed }, status: 200
+      }
+    end
+  end
+
+  def set_approved
+    admin_info = AdminInfo.find_by(id: params[:admin_info_id])
```
anathomical: This line is repeated often enough that it might make sense to make a `set_admin_info` `before_action` for these methods similar to `set_user` or `set_user_by_token` in `app/controllers/accounts_controller`.  This would also let us centralize any logic we might want for cases where the `id` value being passed in doesn't correspond to an actual record (unlikely since we control the front-end).

On that note, this is purely a style thing, so probably not a big deal, but given that each `user` can only have a single `AdminInfo` record, it might make sense to use `user_id` instead of `admin_info_id` for this, and look up (or create) an `AdminInfo` object associated with the `user` when necessary.  Maybe not worth overhauling this approach which already works, but perhaps worth considering to keep things focused on the individual users.

services/QuillLMS/app/views/cms/users/show.html.erb
```diff
@@ -72,6 +72,16 @@
     <p><%= @user.role || 'N/A' %></p>
     <br />
 
+    <% if @user.admin? %>
+      <p style='font-weight: 700;'>Admin sub-role</p>
+      <p><%= @user.admin_info&.sub_role || 'N/A' %></p>
+      <br />
+
+      <p style='font-weight: 700;'>Admin approval status</p>
+      <p><%= @user.admin_info&.approval_status || 'N/A' %></p>
```
anathomical: You have already written some nice helper methods on the `User` model, I'd use those here, so `@user.admin_sub_role` over `@user.admin_info&.sub_role` since it will mean that we don't have to edit these views if we ever decide to change how we store sub-role information.  Ditto for approval status.

services/QuillLMS/app/controllers/cms/admin_verification_controller.rb
```diff
@@ -0,0 +1,63 @@
...
+    school = admin_info.user.school
+    DeniedAdminVerificationEmailWorker.perform_async(admin_info.user_id, school.id)
+    render json: {}
+  end
+
+  def set_pending
+    admin_info = AdminInfo.find_by(id: params[:admin_info_id])
+    admin_info.update(approval_status: AdminInfo::PENDING)
+    school = admin_info.user.school
+    SchoolsAdmins.find_by(user: admin_info.user, school: school)&.destroy
```
anathomical: Cool new feature in Rails 6!  You can now call `destroy_by` to get the same functionality as `find_by(...)&.destroy`.  So this can just be `SchoolsAdmins.destroy_by(user: admin_info.user, school: school)`.

services/QuillLMS/app/workers/approved_admin_verification_email_worker.rb
```diff
@@ -0,0 +1,14 @@
+# frozen_string_literal: true
+
+class ApprovedAdminVerificationEmailWorker
+  include Sidekiq::Worker
+
+  def perform(user_id, school_id)
+    user = User.find_by(id: user_id)
+    school_name = School.find_by(id: school_id)&.name
+
+    return if !(user && school_name)
```
anathomical: Minor style thing, but in Rails we generally use `unless` if the condition is going to be entirely negated.  So `return unless user && school_name` is the same as `return if !(user && school_name)` and doesn't require any mental gymnastics trying to figure out what the combinations are.

services/QuillLMS/app/workers/denied_admin_verification_email_worker.rb
```diff
@@ -0,0 +1,14 @@
+# frozen_string_literal: true
+
+class DeniedAdminVerificationEmailWorker
+  include Sidekiq::Worker
+
+  def perform(user_id, school_id)
+    user = User.find_by(id: user_id)
+    school_name = School.find_by(id: school_id)&.name
+
+    return if !(user && school_name)
```
anathomical: Same note as above about `unless` instead of `if` here

services/QuillLMS/spec/mailers/user_mailer_spec.rb
```diff
@@ -91,6 +91,30 @@
...
+    let(:user) { build(:user) }
+    let(:school) { build(:school) }
+    let(:mail) { described_class.denied_admin_email(user, school.name) }
+
+    it 'should set the subject, receiver and the sender' do
+      expect(mail.subject).to eq("We couldn’t verify you as an admin of #{school.name}")
+      expect(mail.to).to eq([user.email])
+      expect(mail.from).to eq(["hello@quill.org"])
+    end
+  end
```
anathomical: Another minor style thing that may not be worth changing anything here, but because you're doing `let(:user)` and `let(:school)` with the same values for both cases, you could wrap both of these `describe` blocks in another `describe` block named something like "admin approval emails" and put those two `let` calls in there.  Given that you only have two instances, it's not a ton of duplication, so probably doesn't really matter, but I figured I'd flag it just in case.

services/QuillLMS/spec/models/user_spec.rb
```diff
@@ -1511,4 +1511,63 @@
...
+      let!(:admin_info) { create(:admin_info, user: user)}
+
+      it 'sets the admin info to have the new sub role' do
+        user.admin_sub_role=AdminInfo::LIBRARIAN_SLASH_MEDIA_SPECIALIST
+        expect(user.admin_sub_role).to eq(AdminInfo::LIBRARIAN_SLASH_MEDIA_SPECIALIST)
+      end
+    end
+
+    context 'user has no admin info' do
+      let!(:admin_info) { create(:admin_info, user: user)}
```
anathomical: I think we need to remove this line in order for this test to be a case when "user has no admin info"?

anathomical (approved):
This looks solid to me.  I have a couple of notes that are mostly about style and organization, but I think that this could probably be merged as-is if you prefer.

PR #10249, opened 2023-02-08
Lms/user email verification

anathomical:
## WHAT
Add back-end infrastructure to handle email verification for accounts that we want to verify things for.
## WHY
When we allow admins to self-signup, we want have the ability to require them to verify that they have access to the email account they sign up with.
## HOW
- created a new model and table for `UserEmailVerification` that links to `User` records
- - Added a bunch of utility functions on the `User` model to simplify managing verification
- - The most useful one is probably `email_verification_pending?` which will return `true` if the user still needs to verify (because we require it but they haven't done it) and `false` in all other cases (we don't care if they verify, or they have verified already)
- Add a new controller that will allow for API calls to let a user verify their email address via a token we email them, or to allow Quill staff to manually mark an email address as verified
- Updated sign up and sign in code for Google and Clever to verify control of email addresses if it's required and a user authenticates with one of those methods

### Notion Card Links
https://www.notion.so/quill/Implement-the-email-verification-step-51c4bb9b2e6342bd9e52e039a1b1b893

PR Checklist | Your Answer
------------ | -------------


services/QuillLMS/app/controllers/verify_emails_controller.rb
```diff
@@ -0,0 +1,34 @@
...
+
+  def verify_by_token
+    token = token_verification_params[:token]
+
+    verification = UserEmailVerification.find_by(verification_token: token)
+
+    return render json: {'error': 'Invalid verification token'}, status: 400 unless verification
+
+    verification.verify(UserEmailVerification::EMAIL_VERIFICATION, token)
+    puts 'trying to render'
```
brendanshean: Probably want to remove the `puts`.

services/QuillLMS/app/controllers/auth/clever_controller.rb
```diff
@@ -50,6 +50,9 @@ def clever
     private def user_success(user, redirect=nil)
       CompleteAccountCreation.new(user, request.remote_ip).call if user.previous_changes["id"]
       user.update(ip_address: request.remote_ip)
+
+      user.verify_email(UserEmailVerification::CLEVER_VERIFICATION) if user.email_verification_pending?
```
brendanshean: For consistency, maybe add `verify_email_if_necessary` private method.

services/QuillLMS/app/models/user.rb
```diff
@@ -233,6 +235,29 @@ def self.valid_email?(email)
     ValidatesEmailFormatOf.validate_email_format(email).nil?
   end
 
+  def require_email_verification
+    create_user_email_verification unless user_email_verification
```
brendanshean, line 239: Where is `create_user_email_verification` defined?

services/QuillLMS/app/models/user.rb
```diff
@@ -233,6 +235,29 @@ def self.valid_email?(email)
...
+  def email_verified?
+    user_email_verification.present? && user_email_verification.verified?
+  end
+
+  def email_verification_pending?
+    requires_email_verification? && !email_verified?
+  end
+
+  def verify_email(verification_method, verification_token = nil)
+    # Set up email verification records if they don't exist yet
```
brendanshean, line 255: what about rewriting 255-258 something like:

```
  user_email_verification
    .find_or_create_by(...)
    .verify(verification_method, verification_token)
```

And then you can get rid of the `require_email_verification` method

services/QuillLMS/app/models/user_email_verification.rb
```diff
@@ -0,0 +1,64 @@
...
+  def verified?
+    verified_at.present?
+  end
+
+  def set_new_token
+    self.verification_token = SecureRandom.uuid
+  end
+
+  private def mark_as_verified(verification_method)
+    update(verification_method: verification_method, verification_token: nil, verified_at: DateTime.now)
```
brendanshean: I tend to [prefer](https://stackoverflow.com/a/12748804) `DateTime.current` to `DateTime.now`

brendanshean (approved):
Nice work on this tricky problem.  I really appreciate the thorough test coverage too.

services/QuillLMS/spec/controllers/verify_emails_controller_spec.rb
```diff
@@ -0,0 +1,29 @@
...
+    it 'should successfully mark a user as having a verified email when Quill staff says it should be' do
+      post :verify_by_staff, params: { user_id: user.id }, format: :json
+
+      expect(response.status).to be(200)
+      expect(user.reload.email_verified?).to be(true)
+      expect(user_email_verification.reload.verification_method).to eq(UserEmailVerification::STAFF_VERIFICATION)
+    end
+  end
+
+  describe '#verify_by_token' do
```
happythenewsad, line 20: How about adding a spec for the unhappy path (HTTP 4xx)?

services/QuillLMS/app/models/user_email_verification.rb
```diff
@@ -0,0 +1,64 @@
...
+    return verify_by_token(verification_token) if verification_method == EMAIL_VERIFICATION
+
+    mark_as_verified(verification_method)
+  end
+
+  def verified?
+    verified_at.present?
+  end
+
+  def set_new_token
```
happythenewsad, line 51: If you run this in the `before_create` hook instead of `after_create`, you should be able add a `not null` DB constraint on `verification_token`, possibly mitigating future bugs. 

Also, `user_id` can be `not null` as well.

happythenewsad (commented):
Looks good; couple comments. 

I understand that there will be some follow-on front end work for this feature, but just to clarify:  when an unverified user attempts to log in, will they be redirected to a "please verify email" treatment, in a following PR?.

services/QuillLMS/app/models/user.rb
anathomical: This is Rails magic.  https://guides.rubyonrails.org/association_basics.html#methods-added-by-belongs-to

services/QuillLMS/app/models/user.rb
anathomical: That refactor isn't unreasonable, but we actually want the `require_email_verification` for other reasons.  Specifically, we want, in future PRs, the ability to create a user and flag it as needing future email verification.  Having a convenience method that you can call on `User` to establish that state will be useful for those cases.  Given that we want to keep the function, do you feel like it makes sense to avoid using it here and refactor the query, or keep things as-is?

services/QuillLMS/app/models/user_email_verification.rb
anathomical: I think that using a `before_create` hook makes sense.  That said, we can't, at least in this implementation, `NOT NULL` the `verification_token` because we intentionally set the value to `NULL` when the user is verified.  No value in keeping the token around.

Making `user_id` a `NOT NULL` column seems reasonable, too.

anathomical:
> Looks good; couple comments.
> 
> I understand that there will be some follow-on front end work for this feature, but just to clarify: when an unverified user attempts to log in, will they be redirected to a "please verify email" treatment, in a following PR?.

Yes.  The plan is that if the user returns `true` for `email_verification_pending?` users will be redirected to a special landing page.  Screenshot is present in the Notion card linked in this PR.

services/QuillLMS/app/models/user.rb
brendanshean: That makes sense.  I didn't see `require_email_verification` called outside of user but if you're going to need it later, then I would ignore my suggestion.

happythenewsad (approved):

emilia-friedberg (approved):

PR #10251, opened 2023-02-09
Evidence: Remove services that we no longer use.

dandrabik:
## WHAT
The monorepo has a bunch of old Evidence services that we know longer use since they were either pulled into the LMS (go endpoint, spelling) or broken out into a separate repo (grammar). The removes the tests and linters from the code as well. 

You served us well 🫡

Tagging a lot of people as an FYI that this will be removed.
## WHY
If code code isn't going to be used, we should remove it from the codebase to make the repo easier to parse.

Also, this removes two tests runs that we don't need (go endpoint, and flake linter), so that'll save some compute time.
## HOW
`rm -r evidence`

PR Checklist | Your Answer
------------ | -------------
Have you added and/or updated tests? |   No, just deleted everything like a boss.


services/QuillLMS/lib/tasks/staging_config.rake
```diff
@@ -75,7 +75,6 @@ namespace :staging_config do
       quill-lms-eric
       quill-lms-pkong
       quill-lms-thomas
-      empirica-grammar-security
```
dandrabik, line 78: This is a tiny unrelated change that I had locally. We deleted this env.

happythenewsad (approved):

anathomical (approved):
At a previous job we tracked how many lines of code individual engineers removed from our codebase on a kind of leader board.  I think you're winning for this year.

brendanshean (approved):
Screenshot 2023-02-09 at 10 29 24 AM
!!!

Thanks for taking out the trash. 😆

PR #10258, opened 2023-02-10
CircleCI - Run focused tests based on diff.

dandrabik:
## WHAT
Edit 4/2/24 - I have this working in a way that will report to Github the correct status by passing a `should-run` parameter to each job from the job list.
CircleCI has a new-ish feature in which you can [run specific tests based on what is changed in the branch](https://circleci.com/docs/using-dynamic-configuration/#execute-specific-workflows-or-steps-based-on-which-files-are-modified), more specifically based on the file paths in the branch diff with base branch `develop`.
## WHY
This will make the test suite faster for some test runs, e.g. if you just edit javascript, it won't run the rails tests. It will also cut down on our CircleCI credit usage.
## HOW
Use CircleCI's `path-filtering` to define a few variables. Pass the path variable to the `should-run` variable of each job, which conditionally runs the job.

Note, most of this diff is just moving the current `jobs` code from `config.yml` to `continue_config.yml`. The only new code in the `continue_config.yml` is at the bottom `parameters` and `workflows`
### Screenshots
Some Sample runs in CircleCI confirming this works.
**Editing something in the CMS** (note all the jobs run, but they only do the one blank step):
Screenshot 2024-04-02 at 5 20 32 PM

### Notion Card Links

PR Checklist | Your Answer
------------ | -------------
Have you added and/or updated tests? |  Sort of.


.circleci/config.yml
```diff
@@ -1,351 +1,32 @@
...
-          <<: *default_filter
-      - rubocop_lint:
-          <<: *default_filter
+      # the path-filtering/filter job determines which pipeline
+      # parameters to update.
+      - path-filtering/filter:
+          name: check-updated-files
+          # 3-column, whitespace-delimited mapping. One mapping per
+          # line:
+          # <regex path-to-test> <parameter-to-set> <value-of-pipeline-parameter>
```
dandrabik, line 21: This is the regex path logic to determine what was edited.

.circleci/continue_config.yml
```diff
@@ -0,0 +1,388 @@
...
+    type: boolean
+    default: false
+  marking-logic-edits:
+    type: boolean
+    default: false
+
+# here we specify our workflows, most of which are conditionally
+# executed based upon pipeline parameter values. Each workflow calls a
+# specific job defined above, in the jobs section.
+workflows:
```
dandrabik, line 524: The `when` items here are the logic of when to run certain tests.

.circleci/config.yml
```diff
@@ -1,351 +1,32 @@
...
+          mapping: |
+            services/QuillLMS/(app/|bin/|config/|db/|lib/|engines/|Gemfile).* lms-edits true
```
brendanshean: What about `services/QuillLMS/spec` for lms-edits?

brendanshean: Also, are there assets in public/ whose change would trigger a failure?

brendanshean (approved):
This is really cool and a great idea.  Great job figuring out the configuration.

anathomical (approved):
Excited about faster running CI

happythenewsad (approved):
👍

dandrabik:
Closing until there is a way to only run focused tests AND properly have test-passing checks in github (since the tests sets that are run vary by PR, there's not a way to work to github checks to ensure mergers without tests being run).

.circleci/continue_config.yml
```diff
@@ -0,0 +1,576 @@
...
+          POSTGRES_USER: ubuntu
+          POSTGRES_PASSWORD: password
+      - image: pgvector/pgvector:pg15
+        name: test_replica
+        environment:
+          POSTGRES_DB: quill_test_db
+          POSTGRES_USER: ubuntu
+          POSTGRES_PASSWORD: password
+    steps:
+      - run:
```
dandrabik: Note, jobs must have one step. So in case the condition is false, include a blank step.

services/QuillCMS/Gemfile
```diff
@@ -6,7 +6,7 @@ git_source(:github) do |repo_name|
   repo_name = "#{repo_name}/#{repo_name}" unless repo_name.include?('/')
   "https://github.com/#{repo_name}.git"
 end
-
+# test comment
```
anathomical: Do we need to keep this?

anathomical (approved):
Super excited about CI going even faster.

.circleci/continue_config.yml
```diff
@@ -0,0 +1,576 @@
...
+# here we specify our workflows, most of which are conditionally
+# executed based upon pipeline parameter values. Each workflow calls a
+# specific job defined above, in the jobs section.
+define: &default_filter
+  filters:
+    branches:
+      ignore: /^deploy-.*/
+
+workflows:
+  build-test:
```
brendanshean, line 525: This is really cool!

brendanshean (approved):
Great work figuring this configuration out!  This is a big win for the whole team.

.circleci/continue_config.yml
brendanshean: This feels like something important that should be captured in the configuration itself.  Maybe replace each of the `echo 'hello'` sections with something like:

```
- run:
  name: Required first step in case proceeding condition is false
  command: echo 'about to check run conditions'
```

dandrabik:
I've done enough refactors from QA this morning that it probably needs a re-review. But this logic all works as I want it to now.

brendanshean (approved):

PR #10270, opened 2023-02-14
admin sign up

emilia-friedberg:
## WHAT
Allow admins to sign up as admins, with some changes to the onboarding flow (added steps for email verification, sub-role selection, and school verification) to support authenticating and verifying these new users. Also minor adjustments to the existing flows to make sure everything is linked together properly and they don't see things that would be confusing (like the option to select a non-US K12 school).

## WHY
This is one of our [Q1 OKRs](https://www.notion.so/quill/Q1-2023-Product-OKRs-344f5ad67c624bfe80c9c7a5783e459d?pvs=4#fdbe252337d146f9afc7cf12bd397b08).

## HOW
Build out new controllers for updating admin info and getting info about a user's associated school (necessary for the school verification page). Adapt some pieces of initial account creation to account for admin-specific properties, like email verification and approval status. Then really all just building out new React components and adjusting old ones to tie them together.

Additional items to note:
1. The button to enter the flow as a K-12 Administrator is currently hidden behind a query param (you have to access the page via `/account/new?show-admin=true` to see it. This is because the admin dashboard is not yet functional for teachers that haven't been approved as school admins (that will be in a follow-up PR)
2. There is some additional code necessary to get email verification working that exists in a follow-up PR.

### Screenshots
Screen Shot 2023-02-15 at 10 09 58 AM
Screen Shot 2023-02-15 at 10 10 34 AM
Screen Shot 2023-02-15 at 10 09 20 AM
Screen Shot 2023-02-15 at 10 11 18 AM

### Notion Card Links
https://www.notion.so/quill/Add-the-ability-to-sign-up-as-an-admin-1ec59d85e1b44ee9b61544e2fd937cc6?pvs=4
https://www.notion.so/quill/Implement-the-email-verification-step-frontend-ab610afbedfd4093a6b6acee0611f330?pvs=4
https://www.notion.so/quill/Implement-the-sub-role-selection-step-339c233b3ce04a6798b22dc09a89b2f9?pvs=4
https://www.notion.so/quill/Implement-the-admin-authenticity-verification-step-e24b3897f73146a98e3ef97bccc50372?pvs=4

PR Checklist | Your Answer
------------ | -------------
Have you added and/or updated tests? | YES
Have you deployed to Staging? | YES
Self-Review: Have you done an initial self-review of the code below on Github? | YES
Spec Review: Have you reviewed the spec and ensured this meets requirements and/or matches design mockups? | YES

PR #10272, opened 2023-02-15
get email verification working all the way through

emilia-friedberg:
## WHAT
Add actual email and email sending functionality; minor adjustments to get existing code to work with the frontend. 

## WHY
This just completes the necessary work to get email verification working, in addition to the frontend in #10270 and Thomas's work in #10249.

## HOW
Add the actual email template and a function to send it. Then call that method both when we instantiate the user email verification in the sign-up flow and again in a new controller action if the user wants to re-send the token.

### Screenshots
N/A

### Notion Card Links
https://www.notion.so/quill/Implement-the-email-verification-step-51c4bb9b2e6342bd9e52e039a1b1b893

PR Checklist | Your Answer
------------ | -------------
Have you deployed to Staging? | YES (in conjunction with parent PR)


services/QuillLMS/app/models/user_email_verification.rb
```diff
@@ -21,7 +21,7 @@
 #  fk_rails_...  (user_id => users.id)
 #
 class UserEmailVerification < ApplicationRecord
-  after_create :set_new_token, unless: :verification_token
+  after_create :set_new_token, unless: [:verification_token, :verified?]
```
anathomical, line 24: Good catch!

services/QuillLMS/spec/models/user_email_verification_spec.rb
```diff
@@ -61,23 +61,44 @@
     end
   end
 
+  describe '#send_email' do
+    let(:mailer) { double(:mailer, deliver_now!: true) }
+
+    before do
+      allow(UserMailer).to receive(:email_verification_email) { mailer }
+    end
```
anathomical, line 69: Do we need this `allow` when we have an `expect` for the same thing in the only spec in the block?

services/QuillLMS/spec/models/user_email_verification_spec.rb
```diff
@@ -61,23 +61,44 @@
...
 
       expect(verification.verification_token).to eq(new_token)
     end
+
+    it 'should not be called on create if the record is already verified' do
+      new_user = create(:user)
+      verification = UserEmailVerification.create(user: new_user, verified_at: Time.zone.today)
+
+      expect(verification.verification_token).to eq(nil)
+    end
```
anathomical: Since this spec says we want it not to be called, we should probably add an expectation for that.  I think `expect(SecureRandom).not_to receive(:uuid)` works here.

anathomical (approved):
A couple of minor notes, but this looks good to go for me.

services/QuillLMS/spec/models/user_email_verification_spec.rb
emilia-friedberg: Yes - we have to stub the `UserMailer` functionality for the tests (otherwise we get an error about `.deliver_now!` being called on `nil`), but we still want to ensure that it's actually getting called.

happythenewsad (approved):

eadams17 (approved):

PR #10273, opened 2023-02-15
Synthetic Seed Data - Allow Label Examples to be run on their own.

dandrabik:
## WHAT
Adding the ability to just create seed data for label examples.
## WHY
This is a request from the Curriculum Team. Occasionally they just need a little more data for a certain label, so this will give them more flexibility in creating that.
## HOW
Added a new "use_passage" option that gets passed from the form to the backend. Toggles whether we should run all of the generators or just the label examples
### Screenshots
Run Type Options
Screenshot 2023-02-15 at 4 27 42 PM

How it appears on the page:
Screenshot 2023-02-15 at 4 27 30 PM

### Notion Card Links
https://www.notion.so/quill/Add-ability-to-generate-responses-from-labeled-examples-separately-from-seed-data-24b9e54616aa4eb08384ae754b6a34fa

PR Checklist | Your Answer
------------ | -------------
Have you deployed to Staging? |  YES,

### DIFF
--- a/services/QuillLMS/client/app/bundles/Staff/components/evidence/syntheticData/seedDataForm.tsx
+++ b/services/QuillLMS/client/app/bundles/Staff/components/evidence/syntheticData/seedDataForm.tsx
@@ -5,7 +5,7 @@ import ReactHtmlParser from 'react-html-parser'
 import SubmissionModal from '../shared/submissionModal';
 import { fetchActivity, createSeedData } from '../../../utils/evidence/activityAPIs';
 import { renderHeader, quillCloseX } from "../../../helpers/evidence/renderHelpers";
-import { Input, Spinner } from '../../../../Shared/index';
+import { Input, Spinner, DropdownInput, } from '../../../../Shared/index';
 import { TITLE, BECAUSE, BUT, SO } from "../../../../../constants/evidence";
 
 const SeedDataForm = ({ history, match }) => {
@@ -14,7 +14,13 @@ const SeedDataForm = ({ history, match }) => {
   const [errorOrSuccessMessage, setErrorOrSuccessMessage] = React.useState<string>(null);
   const [errors, setErrors] = React.useState<string[]>([]);
   const [showSubmissionModal, setShowSubmissionModal] = React.useState<boolean>(false);
-  const queryClient = useQueryClient()
+  const [usePassage, setUsePassage] = React.useState<string>("true");
+  const queryClient = useQueryClient();
+
+  const runOptions = [
+    { value: "true", label: 'Full Run'},
+    { value: "false", label: 'Label Examples Only'},
+  ];
 
   const [activityNouns, setActivityNouns] = React.useState<string>('');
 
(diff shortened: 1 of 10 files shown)


services/QuillLMS/client/app/bundles/Staff/components/evidence/syntheticData/seedDataForm.tsx
```diff
@@ -14,7 +14,13 @@ const SeedDataForm = ({ history, match }) => {
   const [errorOrSuccessMessage, setErrorOrSuccessMessage] = React.useState<string>(null);
   const [errors, setErrors] = React.useState<string[]>([]);
   const [showSubmissionModal, setShowSubmissionModal] = React.useState<boolean>(false);
-  const queryClient = useQueryClient()
+  const [usePassage, setUsePassage] = React.useState<string>("true");
```
emilia-friedberg: Maybe it makes sense to make `'true'` a constant since you use it a couple of times throughout this file?

services/QuillLMS/client/app/bundles/Staff/components/evidence/syntheticData/seedDataForm.tsx
```diff
@@ -14,7 +14,13 @@ const SeedDataForm = ({ history, match }) => {
...
+  const queryClient = useQueryClient();
+
+  const runOptions = [
+    { value: "true", label: 'Full Run'},
+    { value: "false", label: 'Label Examples Only'},
+  ];
```
emilia-friedberg: Does something not work if you use booleans instead of string values here?

emilia-friedberg (approved):
LGTM, just curious about the `DropdownInput` implementation here. I don't think you need to change it, but for the future one option to consider that might match your use case more closely is using the `.quill-checkbox` styling (example [here](https://github.com/empirical-org/Empirical-Core/blob/develop/services/QuillLMS/client/app/bundles/Teacher/components/accounts/shared/subject_area_section.tsx#L14-L20)).

services/QuillLMS/client/app/bundles/Staff/components/evidence/syntheticData/seedDataForm.tsx
dandrabik: Yeah, `value` is typed as a `string` [here](https://www.notion.so/quill/Product-Retro-February-16-2023-18d01fcd081c4f37b5de00f41f0a0c53).

services/QuillLMS/client/app/bundles/Staff/components/evidence/syntheticData/seedDataForm.tsx
dandrabik: Good call!

dandrabik:
> LGTM, just curious about the `DropdownInput` implementation here. I don't think you need to change it, but for the future one option to consider that might match your use case more closely is using the `.quill-checkbox` styling (example [here](https://github.com/empirical-org/Empirical-Core/blob/develop/services/QuillLMS/client/app/bundles/Teacher/components/accounts/shared/subject_area_section.tsx#L14-L20)).

Thanks for the feedback. So I actually started with a checkbox (I agree that's the natural choice), but testing it in the browser, I found the UI a little confusing to use. I couldn't come up with language that made both states clear and felt intuitive, e.g. I might be over thinking this but it felt odd to either "check something to remove other unnamed things" `[x] Label Examples Only` or "not clear what the other state is"  `[x] Full Run` ). The dropdown UI, while a little weird since it's a boolean, felt more intuitive to me. I essentially wanted a toggle button that has both states in it.

PR #10277, opened 2023-02-16
email verification in user manager

emilia-friedberg:
## WHAT
Allow staff members to see and change email verification status from the user manager.

## WHY
In case a user has some issue completing email verification on their own and we need to set it for them.

## HOW
Add new getter and setter functions to `user` called `email_verification_status`, and use these in the erb forms to get and set the values.

### Screenshots

### Notion Card Links
https://www.notion.so/quill/Expose-email-verification-and-make-editable-in-the-user-manager-make-sure-admin-appears-as-selectab-3f59a08071564bb9a4e134fe7290e424?pvs=4

PR Checklist | Your Answer
------------ | -------------

### DIFF
--- a/services/QuillLMS/app/models/user.rb
+++ b/services/QuillLMS/app/models/user.rb
@@ -261,14 +261,17 @@ class User < ApplicationRecord
 
   def email_verification_status
     return UserEmailVerification::PENDING if email_verification_pending?
+
     return UserEmailVerification::VERIFIED if email_verified?
+
     return nil
   end
 
   def email_verification_status=(status)
-    if status == UserEmailVerification::VERIFIED
+    case status
+    when UserEmailVerification::VERIFIED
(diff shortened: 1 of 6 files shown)


services/QuillLMS/app/models/user.rb
```diff
@@ -259,6 +259,24 @@ def verify_email(verification_method, verification_token = nil)
...
+
+  def email_verification_status=(status)
+    case status
+    when UserEmailVerification::VERIFIED
+      verify_email(UserEmailVerification::STAFF_VERIFICATION)
+    when UserEmailVerification::PENDING
+      require_email_verification
+      user_email_verification.update(verification_method: nil, verified_at: nil)
+    end
+  end
```
anathomical, line 278: I really like the approach here with these helper methods on the `User` model.

anathomical (approved):
Looks great to me.

PR #10284, opened 2023-02-17
admin segment and vitally updates

emilia-friedberg:
## WHAT
Segment and Vitally updates to track new admin info.

## WHY
So we can see the new data that we're storing about these users in Segment and Vitally.

## HOW
Just adding new traits to each.

### Screenshots

### Notion Card Links
https://www.notion.so/quill/Implement-the-necessary-changes-to-Segment-and-Vitally-to-track-admin-creation-and-usage-9bc28fa5afb440f7b4a66e69bc6487b5?pvs=4

PR Checklist | Your Answer
------------ | -------------

### DIFF
--- a/services/QuillLMS/app/services/vitally_integration/serialize_vitally_sales_user.rb
+++ b/services/QuillLMS/app/services/vitally_integration/serialize_vitally_sales_user.rb
@@ -99,8 +99,8 @@ class SerializeVitallySalesUser
         admin_sub_role: @user.admin_sub_role,
         email_verification_status: @user.email_verification_status,
         admin_approval_status: @user.admin_approval_status,
-        admin_of_schools: admin_of_schools,
-        admin_of_districts: admin_of_districts
+        number_of_schools_administered: number_of_schools_administered,
+        number_of_districts_administered: number_of_districts_administered
       }.merge(account_data_params)
     }
   end
@@ -211,12 +211,12 @@ class SerializeVitallySalesUser
     @user.school.district&.name if @user.school.present?
   end
 
-  private def admin_of_schools
-    @user.schools_admins&.joins(:school)&.pluck(:name)&.join(', ')
+  private def number_of_schools_administered
+    @user.schools_admins.count || nil
   end
 
-  private def admin_of_districts
-    @user.district_admins&.joins(:district)&.pluck(:name)&.join(', ')
+  private def number_of_districts_administered
+    @user.district_admins.count || nil
   end
 
 end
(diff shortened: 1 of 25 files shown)


cissyxyu (approved):

services/QuillLMS/app/services/vitally_integration/serialize_vitally_sales_user.rb
```diff
@@ -94,7 +94,13 @@ def data
         completed_evidence_activities_per_student_last_year: get_from_cache("completed_evidence_activities_per_student"),
         date_of_last_completed_evidence_activity: date_of_last_completed_evidence_activity,
         premium_state: @user.premium_state,
-        premium_type: @user.subscription&.account_type
+        premium_type: @user.subscription&.account_type,
+        role: @user.role,
+        admin_sub_role: @user.admin_sub_role,
+        email_verification_status: @user.email_verification_status,
+        admin_approval_status: @user.admin_approval_status,
+        admin_of_schools: admin_of_schools,
```
anathomical: Should we be at all concerned about how long this could be?  If someone is the admin of, say, 100 schools, will this payload get big enough to break something?  And in cases where it doesn't break something, is a list of 100 schools going to be useful to anyone?

anathomical (approved):
Love all the new test specs.  Quick question about the actual values we're passing around with concern for whether Segment or Vitally might have issues with payloads that get too large.  Not sure where allowable payload sizes for either API might be documented, though.

services/QuillLMS/app/services/vitally_integration/serialize_vitally_sales_user.rb
emilia-friedberg: Yeah, this is a good question. I wasn't able to find anything in either Vitally or Segment's docs about a maximum length, but we do have some users with over 100 schools. I'll ask Peter S if it makes sense to add some kind of cap, or if we'd rather release as is and fix if necessary.

anathomical: That seems reasonable.  The stakes feel like they're probably not too high if we get this wrong since the Segment code is isolated javascript, so if it 500s, I don't think it'll impact user experience.  And the Vitally data is synced overnight in Sidekiq.  So if our payloads get too big even though there aren't documented limits, it should be invisible to users so we can clean it up later.

emilia-friedberg: Discussed with Peter S and we decided to turn these values into numbers: https://www.notion.so/Implement-the-necessary-changes-to-Segment-and-Vitally-to-track-admin-creation-and-usage-540e50dfe1694250a4e015c2e566312a?d=ce46559e466d4b78a78ad8287dc81d2e#50364f14078c4b48936f1d150e646dbe.

PR #10285, opened 2023-02-17
lms/procfile-sidekiq-tweaks

anathomical:
## WHAT
Update Procfile so that Sidekiq workers increase databse pool size based on the number of worker threads we intend to run on Sidekiq
## WHY
We want to make sure that there are enough database connections in the pool to serve however many worker threads we want to run
## HOW
Set the `MAX_THREADS` (which is used to set pool size in `config/database.yml`) to the value of `SIDEKIQ_PROCESS_MAX_THREADS` (which is used to set concurrency in `config/sidekiq.yml`) in the Procfile so that the number of database connections in the pool and the number of worker threads in Sidekiq are sycned up.

### Notion Card Links
https://www.notion.so/quill/Experiments-with-large-bg-workers-cc42dc8a144d4c3791f09afd39acd26c?pvs=4

PR Checklist | Your Answer
------------ | -------------
Have you added and/or updated tests? |  No.  Just a config change.

### DIFF
--- a/services/QuillLMS/Procfile
+++ b/services/QuillLMS/Procfile
@@ -3,14 +3,14 @@ web: bundle exec puma -C ./config/puma.rb
 # Note, using MALLOC_ARENA_MAX=2 based on this article:
 # https://github.com/mperham/sidekiq/wiki/Deployment#heroku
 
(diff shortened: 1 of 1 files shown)


dandrabik (approved):
:+1: Thanks for doing this testing.

brendanshean (approved):
Thanks for running this experiment!

PR #10288, opened 2023-02-21
Embed Salesmate sales forms inside Quill

cissyxyu:
## WHAT
Use iFrames to embed the new Salesmate-provided School Quote Request and District Quote Request forms directly in our website. 

## WHY
This way we don't have to redirect users to a third-party website to submit quote requests, and the quote requests can directly populate Salesmate.

## HOW
Create new routes and page templates for School Quote and District Quote requests. Add in the iFrame script provided by Salesmate.

### Screenshots
Screen Shot 2023-02-21 at 3 07 00 PM

### Notion Card Links
https://www.notion.so/quill/Update-Webforms-with-new-links-on-site-for-sales-e3e81c29b4434457bb9944472bd08132?pvs=4

PR Checklist | Your Answer
------------ | -------------

### DIFF
--- a/services/QuillLMS/app/assets/stylesheets/pages/request_quote.scss
+++ b/services/QuillLMS/app/assets/stylesheets/pages/request_quote.scss
@@ -1,0 +1,7 @@
+.quote-request-form {
+  width: 600px;
+  padding-top: 20px;
+  border-radius: 15px;
+  background-color: white;
+  margin: 0 auto !important;
+}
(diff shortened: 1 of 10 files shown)


services/QuillLMS/app/assets/stylesheets/pages/request_quote.scss
```diff
@@ -0,0 +1,7 @@
+.quote-request-form {
+  width: 600px;
```
emilia-friedberg: It probably doesn't matter much, but I think we should be in the habit of writing responsive CSS, especially on simple pages like this that shouldn't require an alternate design. One easy way to do that is usually just to do `max-width: 600px; width: 100%;` inside of specifying an exact width value.

emilia-friedberg (approved):

PR #10314, opened 2023-02-27
lms/add-sidekiq-latency-status

anathomical:
## WHAT
Add new action to statuses to get Sidekiq queue latency

NOTE: We should follow this up by updating our alarms in NewRelic
## WHY
We'd like to modify our NewRelic alarms to go off when the thing we actually care about (queue latency) are not in ideal ranges rather than relying a proxy for that value (assuming that high queue length means high latency)
## HOW
Add a new action to the `StatusController` that reports on Sidekiq queue latencies

PR Checklist | Your Answer
------------ | -------------

### DIFF
--- a/services/QuillLMS/app/controllers/statuses_controller.rb
+++ b/services/QuillLMS/app/controllers/statuses_controller.rb
@@ -36,7 +36,7 @@ class StatusesController < ApplicationController
 
   def sidekiq_queue_latency
     queues = Sidekiq::Queue.all
-    latency_hash = queues.map { |q| [q.name, q.latency] }.to_h
+    latency_hash = queues.to_h { |q| [q.name, q.latency] }
 
     render json: latency_hash
   end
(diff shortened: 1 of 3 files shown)


services/QuillLMS/app/controllers/statuses_controller.rb
```diff
@@ -34,6 +34,13 @@ def redis_queue
     render plain: 'OK'
   end
 
+  def sidekiq_queue_latency
+    queues = Sidekiq::Queue.all
+    latency_hash = queues.map { |q| [q.name, q.latency] }.to_h
```
brendanshean: Some code golf:
`latency_hash = queues.to_h { |q| [q.name, q.latency] }`

brendanshean (approved):
Great idea on making this the alert target.

services/QuillLMS/app/controllers/statuses_controller.rb
anathomical: Oh, hey, I didn't know you could do that with `to_h`.  Today I learned!

PR #10320, opened 2023-02-28
Send additional Segment Identify calls as part of the admin verification process

emilia-friedberg:
## WHAT
Add a worker to perform Segment identify calls and call it when a new admin a) fills out the verification form and b) has their verification approval status updated by a staff member.

## WHY
We need this data to update immediately in Ortto in order to create funnels correctly, as opposed to just whenever the user signs in.

## HOW
Add a new worker and call it in the relevant controller actions.

### Screenshots

### Notion Card Links
https://www.notion.so/quill/Send-additional-Segment-Identify-calls-as-part-of-the-admin-verification-process-04c19f2e8788463fa1b01b5c0cc2b895

PR Checklist | Your Answer
------------ | -------------

### DIFF
--- a/services/QuillLMS/app/controllers/admin_infos_controller.rb
+++ b/services/QuillLMS/app/controllers/admin_infos_controller.rb
@@ -10,6 +10,7 @@ class AdminInfosController < ApplicationController
     if admin_info.approval_status == AdminInfo::SKIPPED && admin_info.verification_url && admin_info.verification_reason
       admin_info.update(approval_status: AdminInfo::PENDING)
       UserRequestedAdminVerificationEmailWorker.perform_async(current_user.id)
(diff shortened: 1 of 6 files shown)


happythenewsad (approved):

anathomical (approved):
It feels kind of weird to me that we don't have a job for this already, but it seems that we don't.  This feels like a totally reasonable way to do it.

One note: you say that we don't want to wait, but we've put this in the `low` queue.  I think that sometimes that queue backs up 15-30 minutes.  I suspect the answer is "yes", but I did want to confirm that that's okay?

emilia-friedberg:
@anathomical thanks for flagging that. After further consideration I think it's probably best to keep this in the default queue, since we want this to fire close to the actual event for the purposes of the funnel flows working correctly.

PR #10322, opened 2023-02-28
lms/sidekiq-latency-400

anathomical:
## WHAT
Tweak the latency status controller to 400 in certain conditions
## WHY
As far as we can tell, the easiest way for us to construct a synthetic alarm in NewRelic that cares about multiple data points is to do so using NRQL, but the features that we can query in NRQL do not include the response body of a synthetic check.  We need to differentiate the response status code in "good" situations vs "bad" ones to work up an NRQL query that can figure out if we've had multiple failures in a row, which is what we want to be querying.
## HOW
Add a segment to determine whether to respond with a 200 status code or a 400 status code to the `sidekiq_queue_latency` action.  The determinant value is always a comparison of the `critical_external` queue's latency against an ENV variable called `SIDEKIQ_CRITICAL_EXTERNAL_LATENCY_LIMIT` (or a default of 5 if that value isn't present).

PR Checklist | Your Answer
------------ | -------------

### DIFF
--- a/services/QuillLMS/app/controllers/statuses_controller.rb
+++ b/services/QuillLMS/app/controllers/statuses_controller.rb
@@ -38,7 +38,7 @@ class StatusesController < ApplicationController
     queues = Sidekiq::Queue.all
     latency_hash = queues.to_h { |q| [q.name, q.latency] }
 
-    response_status = latency_hash.fetch(:critical_external, 0) > ENV.fetch('SIDEKIQ_CRITICAL_EXTERNAL_LATENCY_LIMIT', 5) ? 400 : 200
+    response_status = latency_hash.fetch(:critical_external, 0) > ENV.fetch('SIDEKIQ_CRITICAL_EXTERNAL_LATENCY_LIMIT', 60) ? 400 : 200
 
     render json: latency_hash, status: response_status
   end
--- a/services/QuillLMS/app/controllers/statuses_controller.rb
+++ b/services/QuillLMS/app/controllers/statuses_controller.rb
@@ -38,7 +38,9 @@ class StatusesController < ApplicationController
     queues = Sidekiq::Queue.all
     latency_hash = queues.to_h { |q| [q.name, q.latency] }
 
-    render json: latency_hash
+    response_status = latency_hash.fetch(:critical_external, 0) > ENV.fetch('SIDEKIQ_CRITICAL_EXTERNAL_LATENCY_LIMIT', 5) ? 400 : 200
+
+    render json: latency_hash, status: response_status
   end
 
   def sidekiq_queue_length
--- a/services/QuillLMS/spec/controllers/statuses_controller_spec.rb
+++ b/services/QuillLMS/spec/controllers/statuses_controller_spec.rb
@@ -53,50 +53,52 @@ describe StatusesController, type: :controller do
   end
 
   describe "#sidekiq_queue_latency" do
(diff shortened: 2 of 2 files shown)


brendanshean (approved):

services/QuillLMS/app/controllers/statuses_controller.rb
```diff
@@ -38,7 +38,9 @@ def sidekiq_queue_latency
     queues = Sidekiq::Queue.all
     latency_hash = queues.to_h { |q| [q.name, q.latency] }
 
-    render json: latency_hash
```
brendanshean, line 41: It might make sense to change the name of the method from sidekiq_queue_latency to `sidekiq_critical_external_latency`

anathomical: It could, but it does technically return the latency for every endpoint?  I suppose we could make the 400 part a parameter of some sort.  Like `/statuses/sidekiq_queue_latency/:queue/:latency` and if `queue` has a latency over `latency` it has status 400, otherwise 200, and it always returns the same body?

happythenewsad (approved):

services/QuillLMS/app/controllers/statuses_controller.rb
brendanshean: I was focusing on the `response_status` and forgot you were sending the `latency_hash` too.  Please disregard.

I think the parameter passing is a good idea if we end up using this on other queues.

services/QuillLMS/spec/controllers/statuses_controller_spec.rb
```diff
@@ -53,18 +53,35 @@
   end
 
   describe "#sidekiq_queue_latency" do
-    let(:queue_latency_pairs) do
-      {
+    it 'should build a hash of queue.name => queue.latency for each queue' do
+      queue_latency_pairs = {
```
dandrabik: I believe you can shorten this section like this:

```ruby
let(:queue_doubles) do 
  [ 
    double(name: 'critical', latency: 1.14),
    double(name: 'critical_external', latency: 0.34),
    ...
  ]
end

File: services/QuillLMS/spec/controllers/statuses_controller_spec.rb [hunk 3]
```diff
@@ -53,18 +53,35 @@
...
+      queue_doubles = queue_latency_pairs.map do |name, latency|
+        d = double
+        allow(d).to receive(:name).and_return(name)
+        allow(d).to receive(:latency).and_return(latency)
+        d
+      end
+
+      allow(Sidekiq::Queue).to receive(:all).and_return(queue_doubles)
+
+      expect {
```
dandrabik: I don't think you gain anything from this expects. If there is an error in the method, that would inherently fail the test.

services/QuillLMS/app/controllers/statuses_controller.rb
```diff
@@ -38,7 +38,9 @@ def sidekiq_queue_latency
...
+    response_status = latency_hash.fetch(:critical_external, 0) > ENV.fetch('SIDEKIQ_CRITICAL_EXTERNAL_LATENCY_LIMIT', 5) ? 400 : 200
```
dandrabik: This should be a higher limit than 5, maybe 60.

dandrabik (approved):
Thanks for working on this!

PR #10324, opened 2023-03-01
add migrations and model upgrades to support teachers submitting requests to become admins

emilia-friedberg:
## WHAT
Add a new column, `approver_role`, to the `admin_info` table, and a new table called `admin_approval_requests` to join `admin_info` records with the admins who've been requested to approve the user's upgrade to `admin`.

## WHY
In V1 of the admin project, we set up the infrastructure for users to sign up as admins and be verified by staff members. Now we're making it so that existing users who belong to a school that already has admins can make a request to become an admin, which will be verified by the existing admins. In order to do that, we need to a) be able to differentiate at the database level between users who need to be approved as admins by staff members vs existing admins, and b) in the latter case, store which existing admins the request will be sent to.

## HOW
Just add migrations and update model files and specs.

### Screenshots

### Notion Card Links
https://www.notion.so/quill/Migrations-model-changes-for-teacher-admin-upgrade-requests-40a94d42a8464eecacffdc2f791c8d68?pvs=4

PR Checklist | Your Answer
------------ | -------------

### DIFF
--- a/services/QuillLMS/app/models/admin_info.rb
+++ b/services/QuillLMS/app/models/admin_info.rb
@@ -6,6 +6,7 @@
 #
 #  id                  :bigint           not null, primary key
 #  approval_status     :string
+#  approver_role       :string
 #  sub_role            :string
 #  verification_reason :text
 #  verification_url    :string
@@ -23,6 +24,7 @@
 #
 class AdminInfo < ApplicationRecord
   belongs_to :user
+  has_many :admin_approval_requests
 
   validates :user_id, presence: true, uniqueness: true
 
@@ -43,9 +45,16 @@ class AdminInfo < ApplicationRecord
     BILLING_CONTACT = "Billing contact"
   ]
 
+  APPROVER_ROLES = [
+    User::ADMIN,
+    User::STAFF
+  ]
+
   validates :approval_status, :inclusion=> { :in => APPROVAL_STATUSES }, :allow_nil => true
 
   validates :sub_role, :inclusion=> { :in => SUB_ROLES }, :allow_nil => true
+
+  validates :approver_role, :inclusion=> { :in => APPROVER_ROLES }, :allow_nil => true
 
   def admin
     user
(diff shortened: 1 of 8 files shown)


services/QuillLMS/app/models/admin_info.rb
```diff
@@ -23,6 +24,7 @@
 #
 class AdminInfo < ApplicationRecord
   belongs_to :user
+  has_many :admin_approval_requests
```
brendanshean: Add the `dependent: :destroy`

services/QuillLMS/db/migrate/20230301160642_create_admin_approval_request.rb
```diff
@@ -0,0 +1,12 @@
+# frozen_string_literal: true
+
+class CreateAdminApprovalRequest < ActiveRecord::Migration[6.1]
+  def change
+    create_table :admin_approval_requests do |t|
+      t.references :admin_info, foreign_key: true, dependent: :destroy
```
brendanshean: I'm not familiar with `dependent: :destroy` in migration.  Is this a new convention?  Normally I put it in the model file.

happythenewsad (approved):

brendanshean (approved):
One minor comment, but otherwise looks good to me.  I like modeling this as a new table as opposed to adding more to admin_info.

services/QuillLMS/db/migrate/20230301160642_create_admin_approval_request.rb
emilia-friedberg: You know, I was looking at `https://github.com/empirical-org/Empirical-Core/blob/develop/services/quilllms/db/migrate/20230206203447_create_user_email_verification.rb#L6` as a reference and didn't really think about it, but further investigation indicates that it doesn't do anything in migrations so I think that was just a mistake. I've removed it and added it to the model file.

PR #10325, opened 2023-03-01
fix inline tag highlighting bug in evidence activities

happythenewsad:
## WHAT
Fix the bug in which:

- student selects an Evidence phrase to highlight
- the phrase is inside of inline markup, like `<em>` or `<strong>`
- the computed highlight only grabs the text within the inline markup tag, not the whole passage 

## WHY
The current behavior is inconsistent with normal highlighting. 

## HOW
Have the highlight onclick event use `currentTarget` rather than `target`, which is scoped to the parent tag, not the inline markup tag. More explanation here: https://medium.com/@etherealm/currenttarget-vs-target-in-js-2f3fd3a543e5

### Screenshots

### Notion Card Links
https://www.notion.so/quill/Highlighting-errors-in-Haitian-Rev-Evidence-activity-5a3c54610e0d4e40808f5b67262848f0

PR Checklist | Your Answer
------------ | -------------

### DIFF
--- a/services/QuillLMS/client/app/bundles/Evidence/components/studentView/container.tsx
+++ b/services/QuillLMS/client/app/bundles/Evidence/components/studentView/container.tsx
@@ -444,7 +444,6 @@ export const StudentViewContainer = ({ dispatch, session, isTurk, location, acti
   }
 
   function clickedTextIsWithinInlineMarkupTag(event) {
-    console.log("clickedText: event:", event)
     return INLINE_MARKUP_TAGS.includes(event?.target?.nodeName)
   }
 
--- a/services/QuillLMS/client/app/bundles/Evidence/components/studentView/container.tsx
+++ b/services/QuillLMS/client/app/bundles/Evidence/components/studentView/container.tsx
@@ -58,8 +58,6 @@ const STUDENT_HIGHLIGHT_ENDS_TEXT = "(highlighted text ends here)"
 const PASSAGE_HIGHLIGHT_STARTS_TEXT = "(yellow underlined text begins here)"
 const PASSAGE_HIGHLIGHT_ENDS_TEXT = "(yellow underlined text ends here)"
 
-const INLINE_MARKUP_TAGS = ['EM', 'B', 'STRONG', 'U']
-
 export const StudentViewContainer = ({ dispatch, session, isTurk, location, activities, handleFinishActivity, user }: StudentViewContainerProps) => {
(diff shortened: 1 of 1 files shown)


services/QuillLMS/client/app/bundles/Evidence/components/studentView/container.tsx
```diff
@@ -441,27 +443,36 @@ export const StudentViewContainer = ({ dispatch, session, isTurk, location, acti
...
 
     if (studentHighlights.includes(textWithHighlightContentRemoved)) {
       newHighlights = studentHighlights.filter(hl => hl !== textWithHighlightContentRemoved)
     } else {
-      newHighlights = studentHighlights.concat(text)
+      newHighlights = studentHighlights.concat(stringText)
     }
 
     setStudentHighlights(newHighlights)
-    callback && callback()
```
happythenewsad, line 464: cleanup of unused param and logic

anathomical (approved):
This seems like a solid solution to the problem to me.

services/QuillLMS/client/app/bundles/Evidence/components/studentView/container.tsx
```diff
@@ -441,27 +443,36 @@ export const StudentViewContainer = ({ dispatch, session, isTurk, location, acti
...
+    toggleStudentHighlight(e)
   }
 
-  function toggleStudentHighlight(text, callback=null) {
-    const textWithHighlightContentRemoved = text.replace(STUDENT_HIGHLIGHT_STARTS_TEXT, '').replace(STUDENT_HIGHLIGHT_ENDS_TEXT, '')
+  function toggleStudentHighlight(eventOrString) {
+    if (typeof(eventOrString) === 'string') {
+      let stringText = eventOrString
+    } else {
+      let stringText = clickedTextIsWithinInlineMarkupTag(eventOrString) ? eventOrString.currentTarget.textContent : eventOrString.target.textContent
```
emilia-friedberg: Out of curiosity, what happens if we just always use `currentTarget` instead of `target`?

emilia-friedberg (approved):

eadams17 (approved):

services/QuillLMS/client/app/bundles/Evidence/components/studentView/container.tsx
happythenewsad: I'm glad you asked that question - turns out we can use currentTarget in both code paths.

PR #10328, opened 2023-03-01
lms/rake-task-to-purge-classroom-activity-sessions

anathomical:
## WHAT
Create a rake task to purge classrooms and their activity_sessions and all models in between. 
## WHY
We don't intend to use this with any regularity, but in cases where teachers or admins come to us with a request to delete PII for some classes, but not for all activities of their students (so that we can't use our usual anonymization process which removes PII from user accounts)

As far as I know we've gotten one request to do this that has been approved, for a single teacher, so there's a chance that this task will be run exactly once, but I figured it was better to have it go through the PR process just to check my approach.
## HOW
Take a list of classroom ids, then find all the models that connect that `classroom` to `activity_sessions` and flag all of them for deletion from the database.

### Notion Card Links
https://www.notion.so/quill/Engineering-Question-1440929cef0b43bfbb25ec013e2b9541?pvs=4

PR Checklist | Your Answer
------------ | -------------
Have you added and/or updated tests? |  No tests on rake task

### DIFF
--- a/services/QuillLMS/lib/tasks/classrooms.rake
+++ b/services/QuillLMS/lib/tasks/classrooms.rake
@@ -1,0 +1,28 @@
+# frozen_string_literal: true
+
+require_relative './progress_bar'
+
+namespace :classrooms do
+  desc 'Purge activity history for a specified list of classroom ids'
+  task :purge_histories, [:classroom_ids] => :environment do |t, args|
+    classroom_ids = args[:classroom_ids].split
+
+    classrooms = Classroom.where(id: classroom_ids)
+    classrooms_teachers = ClassroomsTeacher.where(classroom_id: classroom_ids)
+    classroom_units = ClassroomUnit.where(classroom_id: classroom_ids)
+    activity_sessions = ActivitySession.where(classroom_unit: classroom_units)
+    concept_results = ConceptResult.where(activity_session: activity_sessions)
+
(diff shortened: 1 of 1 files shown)


happythenewsad (commented):
Looks fine, but I don't think we need/want to delete the concept results. We could anonymize them by setting the activity_session_id to an encrypted version of the original.

anathomical:
> Looks fine, but I don't think we need/want to delete the concept results. We could anonymize them by setting the activity_session_id to an encrypted version of the original.

Can you expand on this?  I think that avoiding deleting things we don't have to delete sounds great, but don't see an obvious way to do it here.  I suppose that `ConceptResult.activity_session_id` isn't a foreign key, so we could set it to anything we wanted.  It sounds like you have a specific idea of what we should set it to?

happythenewsad:
> > Looks fine, but I don't think we need/want to delete the concept results. We could anonymize them by setting the activity_session_id to an encrypted version of the original.
> 
> Can you expand on this? I think that avoiding deleting things we don't have to delete sounds great, but don't see an obvious way to do it here. I suppose that `ConceptResult.activity_session_id` isn't a foreign key, so we could set it to anything we wanted. It sounds like you have a specific idea of what we should set it to?

I didn't have anything super specific in mind - anything deterministic. That way, a student's work can still be logically queried/filtered/grouped in the future (as opposed to setting activity_session_id to `nil` for all runs of the rake task).

Slack me if you'd like to talk it out :-)

emilia-friedberg (commented):
Along the lines of what Peter Kong was thinking - can we just remove the `classroom_unit_id`  and the `user_id` from the activity sessions? That would let us preserve those sessions, and their concept results, without being traceable to the users who completed them.

anathomical:
I really like @emilia-friedberg 's suggestion.  I've refactored the task so that instead of deleting any records, it just orphans all related ActivitySession records.

emilia-friedberg (approved):
LGTM!

PR #10331, opened 2023-03-02
Lms/fix admin email verification

anathomical:
## WHAT
- Make sure that we flag users who sign up as admins with a Google or Clever account get marked as email-verified.
- Add a rake task to backfill the 14-ish admins who have signed up with Clever or Google and haven't been marked as verified
## WHY
While we don't want users signing up with Google or Clever to go through the email verification workflow, we do still want them to be marked as verified in our system for consistency.
## HOW
Add some logic the the account completion logic to just mark users as having verified their email when they're signing up with Clever or Google. 

### Notion Card Links
https://www.notion.so/quill/When-admins-sign-up-with-Google-or-Clever-set-their-Email-verification-status-to-Verified-ce876e661cb3406e906de936cd80d357?pvs=4

PR Checklist | Your Answer
------------ | -------------

### DIFF
--- a/services/QuillLMS/lib/tasks/users.rake
+++ b/services/QuillLMS/lib/tasks/users.rake
@@ -44,13 +44,14 @@ namespace :users do
     User.left_outer_joins(:user_email_verification)
       .joins(:admin_info) # only users who self-service sign up as admins have AdminInfo records
       .where.not(clever_id: nil)
-      .or(User.where.not(google_id: nil)) # note that the position of `or` matters a lot when you have multiple `where` clauses
+      .or(User.where.not(google_id: nil)) # the position of `or` matters a lot when you have multiple `where` clauses
       .where(role: User::ADMIN)
       .where(user_email_verification: { id: nil })
       .each do |user|
-        verification_method = UserEmailVerification::GOOGLE_VERIFICATION if user.google_id
-        verification_method = UserEmailVerification::CLEVER_VERIFICATION if user.clever_id
-        user.verify_email(verification_method)
+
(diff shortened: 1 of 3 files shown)


services/QuillLMS/lib/tasks/users.rake
```diff
@@ -39,4 +39,19 @@ namespace :users do
...
     end
   end
+
+  task mark_google_clever_admins_verified: :environment do
+    User.left_outer_joins(:user_email_verification)
+      .where.not(clever_id: nil)
+      .or(User.where.not(google_id: nil)) # note that the position of `or` matters a lot when you have multiple `where` clauses
+      .where(role: User::ADMIN)
+      # I was going to use the open-ended `..` here, but Rubocop apparently can't handle a `..` without a value at the end and chokes.  `..nil` is documented as being equivalent to an unterminated `..`
+      .where(created_at: DateTime.new(2023, 2, 27)..nil) # this is the date we launched self-service admin sign-up
```
emilia-friedberg: I think this probably works, but there could theoretically be some admins that were invited by an existing admin (bypassing the admin flow and therefore email verification) who signed up in that timeframe, so it might be safer to check for the existence of an `approval_status` value to be sure that they really did use the self-service flow.

emilia-friedberg (approved):
One note on the rake task backfill, but I think it's theoretical anyway. Thank you for the quick fix!

services/QuillLMS/lib/tasks/users.rake
anathomical: Good call.  Per our Zoom conversation, I'll update the logic here.

PR #10346, opened 2023-03-07
Update scrolling and alignment on Diagnostics Practice Recommendations page

cissyxyu:
## WHAT
We want the "Assign Test" and "Post Diagnostic Test Assigned" tables to be aligned together and statically positioned on the right side of this page, regardless of the scrolling behavior of the table on the left side. 

## WHY
It makes more visual sense for the "Post Diagnostic Test Assigned" table column to be aligned with the corresponding header "Assign Test", otherwise it's confusing because it's unclear what the "Post Diagnostic Test Assigned" table column is referring to.

## HOW
I had to change the React components a bit to make the right side table and the left side table into separate components. That way, I could put the left side table in one Div and the right side table in another Div to keep them both aligned with each other. 
I also had to change CSS style settings quite a bit to get the left side table to scroll independently (previously the entire container scrolled as one unit, but now the right side has to be static while the left side scrolls).

### Screenshots
Screen Shot 2023-03-07 at 4 43 29 PM

### Notion Card Links
https://www.notion.so/quill/UI-Enhancements-to-the-Diagnostic-Progress-Indicator-Designs-22a4b8c4a93c4c6a8f78a2a5e29db3c2?pvs=4

PR Checklist | Your Answer
------------ | -------------


services/QuillLMS/client/app/bundles/Teacher/components/progress_reports/diagnostic_reports/diagnostics/postTestTable.tsx
```diff
@@ -0,0 +1,194 @@
...
+  React.useEffect(() => {
+    onScroll()
+  }, [size])
+  React.useEffect(() => {
+    window.addEventListener("scroll", onScroll);
+
+    return () => {
+      window.removeEventListener("scroll", onScroll);
+    };
+  }, [handleScroll]);
```
emilia-friedberg: Nit: this section feels crowded - it would be helpful for readability to have whitespace between the function definitions and `React.useEffect` calls.

services/QuillLMS/client/app/bundles/Teacher/components/progress_reports/diagnostic_reports/diagnostics/postTestTable.tsx
```diff
@@ -0,0 +1,194 @@
...
+        <div className="activity-count-and-select-all">
+          <span className="activity-count">{activityCount} {activityCount === 1 ? 'activity' : 'activities'}</span>
+          <Tooltip
+            tooltipText="Select all column"
+            tooltipTriggerText={<button aria-label="Select all column" className="interactive-wrapper" onClick={handleSelectAllClick} type="button">{ellipsesIcon}</button>}
+          />
+        </div>
+      </th>
+    )
+  }
```
emilia-friedberg: I think this is identical to the `ActivityPackHeader` defined in the `recommendationsTable` file, right? If so, it should get split into a third file and imported into both rather than duplicated here.

services/QuillLMS/client/app/bundles/Teacher/components/progress_reports/diagnostic_reports/diagnostics/postTestTable.tsx
```diff
@@ -0,0 +1,194 @@
...
+    left: 0,
+    right: 0,
+    zIndex: 5
+  })
+  const [stickyTableRightOffset, setStickyTableRightOffset] = React.useState(0)
+  const recommendationsTableRef = React.useRef(null);
+  const handleScroll = React.useCallback(({ top, bottom, left, right, }) => {
+    if (top <= 0 && bottom > 92) {
+      setStickyTableStyle(oldStickyTableStyle => ({ ...oldStickyTableStyle, left: left }))
+      setStickyTableRightOffset(782)
```
emilia-friedberg: `782` here should be a constant so we both have a little more context for where it's coming from (the name) and also can change it more easily.

services/QuillLMS/client/app/bundles/Teacher/components/progress_reports/diagnostic_reports/diagnostics/postTestTable.tsx
```diff
@@ -0,0 +1,194 @@
...
+  position: string,
+  top: number,
+  left: number,
+  right: number,
+  zIndex: number,
+}
+
+const ellipsesIcon = <img alt="Open menu icon" src={`${baseDiagnosticImageSrc}/ellipses_icon.svg`} />
+
+const PostTestAssignmentTable = ({showPostTestAssignmentColumn, students, studentsWhoCompletedDiagnostic, postTestSelections, setPostTestSelections, previouslyAssignedPostTestStudentIds, studentsWhoCompletedAssignedRecommendations, postDiagnosticUnitTemplateId, }) => {
```
emilia-friedberg: I think this component duplicates too much code from the older one - I know there's a lot of tightly coupled logic in there, but I think it needs to get DRY-ed up, especially given the complexity of the logic at hand. Having to make any change in two places would get really complicated fast. I think your best bet would be creating a third, more generic component that can just accept props to differentiate itself. The good news is, I think a huge chunk of the code in both places can be deleted - I may be mistaken, but since Jack signed off on no longer using the sticky header, I'm pretty sure everything regarding it can just get deleted, which will really reduce the amount of code that needs deduplication.

emilia-friedberg (changes requested):
This is getting so close! I know this page is so frustrating to work with and I'm impressed with how far you've gotten it.

I left some comments about duplication, but I think happily a lot of the duplicated code can just get deleted. I also checked out your branch on staging and spotted a couple of visual bugs:

There is a random "s" in the bottom righthand corner of the recommendations table container.
Screen Shot 2023-03-07 at 8 24 49 AM

The "assign test" header still shows up when there is no post-test to be assigned.
Screen Shot 2023-03-07 at 8 25 02 AM

The whole class instruction section seems to just increase indefinitely with page width - we probably need a max width on that.
Screen Shot 2023-03-07 at 8 25 23 AM

services/QuillLMS/client/app/bundles/Teacher/components/progress_reports/diagnostic_reports/diagnostics/postTestTable.tsx
emilia-friedberg: Note that I think the comments above can all get ignored if I'm right about what can get deleted/de-duplicated.

cissyxyu: So I ended up deleting this file and simply combining this code into the `RecommendationsTable` component. That component now accepts a boolean indicating whether it's the post test table or not, and then makes some display differentiations based on that. So we longer have that duplicated code and duplicated parameters!

services/QuillLMS/client/app/bundles/Teacher/components/progress_reports/diagnostic_reports/diagnostics/recommendationsTable.tsx
```diff
@@ -352,63 +244,54 @@ const RecommendationsTable = ({ recommendations, responsesLink, students, select
...
-        {renderPostTestTableHeader(true)}
-      </table>
+      <thead>
+        <tr>
+          {isPostTest ? null : <th className="corner-header" style={sticky ? style : {}}>Name</th>}
+          {isPostTest ? renderPostTestTableHeader() : tableHeaders}
+        </tr>
+      </thead>
     )
   }
```
emilia-friedberg: I think the last couple of references to stickiness/changing style in here can get deleted (ie, we don't need the function parameter).

emilia-friedberg (approved):
This looks great - so much cleaner and it's so nice to have all that sticky code gone now that we aren't using it! One tiny cleanup suggestion, but I think once Jack has signed off on design and those final small adjustments are made we're good to go. Thanks for all your work on this!

eadams17 (approved):

PR #10349, opened 2023-03-08
Make ziptz a singleton

brendanshean:
## WHAT
The Ziptz gem is [used](https://github.com/empirical-org/Empirical-Core/blob/develop/services/QuillLMS/app/models/user.rb#L737) in a user callback for time_zone calculations.
This PR moves the `z = Ziptz.new` into an initializer that makes it a [singleton](https://ruby-doc.org/stdlib-2.5.1/libdoc/singleton/rdoc/Singleton.html).

## WHY
Each `z = Ziptz.new` call uses quite a bit of memory due to unzipping a [file](https://github.com/infused/ziptz/blob/master/lib/ziptz.rb#L64) containing all of the zip code info.    Here's two experiments for importing 30 students in Google Classroom.
30-students-no-ziptz
30-students-ziptz

By moving this to a singleton, we cut down significantly the memory and number of objects allocated.

Here's a sample import of 90 students, one with the singleton and one without
Screenshot 2023-03-08 at 8 09 41 AM

This should hopefully help with the memory errors we are seeing on our worker dynos.

## HOW
Add an initializer that makes Ziptz a singleton.

Then use `::Ziptz.instance` to call the object (rather than `:Ziptz.new`)

### Screenshots

### Notion Card Links

PR Checklist | Your Answer
------------ | -------------

### DIFF
--- a/services/QuillLMS/Gemfile
+++ b/services/QuillLMS/Gemfile
@@ -144,7 +144,6 @@ gem 'clever-ruby', '~> 2.0.1'
 
 # Memory profiling
 gem 'puma_worker_killer'
-gem 'memory_profiler', require: false
 
 # temp for migrations
 gem 'paperclip'
@@ -166,6 +165,7 @@ group :development do
   gem 'foreman'
   gem 'graphiql-rails'
   gem 'letter_opener'
+  gem 'memory_profiler'
   gem 'spring'
   gem 'spring-commands-rspec'
   gem 'annotate'
--- a/services/QuillLMS/Gemfile
+++ b/services/QuillLMS/Gemfile
@@ -144,6 +144,7 @@ gem 'clever-ruby', '~> 2.0.1'
 
 # Memory profiling
 gem 'puma_worker_killer'
+gem 'memory_profiler', require: false
 
 # temp for migrations
 gem 'paperclip'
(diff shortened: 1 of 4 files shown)


emilia-friedberg (approved):
Awesome reduction. Thanks for finding that.

services/QuillLMS/Gemfile
```diff
@@ -144,6 +144,7 @@ gem 'clever-ruby', '~> 2.0.1'
 
 # Memory profiling
 gem 'puma_worker_killer'
+gem 'memory_profiler', require: false
```
dandrabik: Do we need this gem installed on prod? Maybe we should move to the `development` section?

services/QuillLMS/config/initializers/ziptz.rb
```diff
@@ -0,0 +1,5 @@
+# frozen_string_literal: true
+
+class Ziptz
+  include Singleton
```
dandrabik, line 4: Nice! This Singleton module is new to me. Thanks for showing me that.

dandrabik (approved):
Thanks for tracking this down, I'm excited to view the impact on memory. Well done!

anathomical (approved):
Agree with @dandrabik that just making the class a singleton is a very simple and clever way of solving the issue here.

services/QuillLMS/Gemfile
brendanshean: Good call.

PR #10357, opened 2023-03-09
Admin dashboard/ freemium standards report

eadams17:
## WHAT
add a freemium view of the district standards reports on the admin dashboard (*note that this also fixes a UI issue with the Premium tab when subscriptions are expired and also fixes an issue where the standards reports was briefly showing for non-premium users)

## WHY
we want to provide a preview to non-premium admins of this data

## HOW
add a `freemium` param to the API to conditionally limit results, conditionally render report based on limited user admin status

### Screenshots
Screen Shot 2023-03-09 at 2 56 39 PM

### Notion Card Links
https://www.notion.so/Add-Standards-report-preview-to-the-freemium-state-of-the-admin-dashboard-1a8e57eb205c4afca380a9002646cbdc

PR Checklist | Your Answer
------------ | -------------


services/QuillLMS/app/models/schools_admins.rb
```diff
@@ -37,6 +38,7 @@ def wipe_cache
     $redis.del("#{DISTRICT_ACTIVITY_SCORES_CACHE_KEY_STEM}#{user_id}")
     $redis.del("#{DISTRICT_CONCEPT_REPORTS_CACHE_KEY_STEM}#{user_id}")
     $redis.del("#{DISTRICT_STANDARD_REPORTS_CACHE_KEY_STEM}#{user_id}")
+    $redis.del("#{FREEMIUM_DISTRICT_STANDARD_REPORTS_CACHE_KEY_STEM}#{user_id}")
```
brendanshean: We should avoid using global variables (i.e `$redis`).  Instead use [delete](https://api.rubyonrails.org/v7.0.4/classes/ActiveSupport/Cache/Store.html#method-i-delete) with a Rails wrapper:

 ```
Rails.cache.delete("#{FREEMIUM_DISTRICT_STANDARD_REPORTS_CACHE_KEY_STEM}#{user_id}")
```

The added benefit to this is that we abstract out the details of what the cache is.  If at some point we used a different cache store, we only have to change the config [here](https://github.com/empirical-org/Empirical-Core/blob/develop/services/QuillLMS/config/application.rb#L27) rather than replacing $redis everywhere.

File: services/QuillLMS/app/queries/progress_reports/district_standards_reports.rb [hunk 2]
```diff
@@ -3,8 +3,11 @@
 class ProgressReports::DistrictStandardsReports
   attr_reader :admin_id
```
brendanshean: Is `is_fremium_view` missppelled?  Kind of reminds me of fremen.   

I would rename variable to just `is_freemium` and add it to attr_reader here:

`attr_reader :admin_id, :is_freemium`

Then you can get rid of a lot of the parameter passing as follow:

```
def results
  ...
  ids = user_ids
  return [] if ids.empty?

  query(ids)

end

# is_freemium doesn't need to be passed if it's an attr_reader
private def query(user_ids)
  ...
end

# admin_id doesn't need to be passed since it's an attr_reader
private def user_ids
  ...
end
```

services/QuillLMS/client/app/actions/district_standards_reports.js
```diff
@@ -31,12 +31,13 @@ export const initializePusherForDistrictStandardsReports = (adminId) => {
...
-export const getDistrictStandardsReports = () => {
+export const getDistrictStandardsReports = (isFreemiumView) => {
+  const freemiumParam = isFreemiumView ? '?freemium=true' : ''
   return (dispatch) => {
     requestGet(
-      `${process.env.DEFAULT_URL}/api/v1/progress_reports/district_standards_reports`,
+      `${process.env.DEFAULT_URL}/api/v1/progress_reports/district_standards_reports${freemiumParam}`,
       (body) => {
-        if (body.id) {
+        if (body.id && process.env.RAILS_ENV !== 'development') {
```
brendanshean: What's the reasoning for excluding development here?  Should we also exclude test?

services/QuillLMS/spec/controllers/api/v1/progress_reports_controller_spec.rb
```diff
@@ -117,9 +117,18 @@
       allow_any_instance_of(ProgressReports::DistrictStandardsReports).to receive(:results) { "some data" }
     end
 
-    it 'should return the district activit scores progress reports' do
+    it 'should return the district activities scores progress reports for the standards reports page' do
+      expect($redis).to receive(:get).with("#{SchoolsAdmins::DISTRICT_STANDARD_REPORTS_CACHE_KEY_STEM}#{teacher.id}").and_return(nil)
```
brendanshean: `You want to replace $redis with the Rails.cache...`

services/QuillLMS/spec/controllers/api/v1/progress_reports_controller_spec.rb
```diff
@@ -117,9 +117,18 @@
...
+      expect(FindDistrictStandardsReportsWorker).to receive(:perform_async).with(teacher.id, false)
       get :district_standards_reports, as: :json
       expect(response.body).to eq({id: teacher.id}.to_json)
     end
+
+    it 'should return the district activities scores progress reports for the freemium section of the admin dashboard' do
+      expect($redis).to receive(:get).with("#{SchoolsAdmins::FREEMIUM_DISTRICT_STANDARD_REPORTS_CACHE_KEY_STEM}#{teacher.id}").and_return(nil)
```
brendanshean: `You want to replace $redis with the Rails.cache...`

services/QuillLMS/spec/models/schools_admin_spec.rb
```diff
@@ -41,6 +41,7 @@
       expect($redis).to receive(:del).with("#{SchoolsAdmins::DISTRICT_ACTIVITY_SCORES_CACHE_KEY_STEM}#{user.id}")
```
brendanshean: `You want to replace $redis with the Rails.cache...`

services/QuillLMS/app/workers/find_district_standards_reports_worker.rb
```diff
@@ -4,13 +4,18 @@ class FindDistrictStandardsReportsWorker
...
-    serialized_district_standards_reports = ProgressReports::DistrictStandardsReports.new(admin_id).results.to_json
-    $redis.set("#{SchoolsAdmins::DISTRICT_STANDARD_REPORTS_CACHE_KEY_STEM}#{admin_id}", serialized_district_standards_reports)
-    $redis.expire("#{SchoolsAdmins::DISTRICT_STANDARD_REPORTS_CACHE_KEY_STEM}#{admin_id}", serialized_district_standards_reports_cache_life)
-    PusherDistrictStandardsReportsCompleted.run(admin_id)
+    serialized_district_standards_reports = ProgressReports::DistrictStandardsReports.new(admin_id, is_fremium_view).results.to_json
+    if is_fremium_view
+      $redis.set("#{SchoolsAdmins::FREEMIUM_DISTRICT_STANDARD_REPORTS_CACHE_KEY_STEM}#{admin_id}", serialized_district_standards_reports)
+      $redis.expire("#{SchoolsAdmins::FREEMIUM_DISTRICT_STANDARD_REPORTS_CACHE_KEY_STEM}#{admin_id}", serialized_district_standards_reports_cache_life)
+    else
+      $redis.set("#{SchoolsAdmins::DISTRICT_STANDARD_REPORTS_CACHE_KEY_STEM}#{admin_id}", serialized_district_standards_reports)
```
brendanshean:
```
Rails.cache.write(
  "#{SchoolsAdmins::DISTRICT_STANDARD_REPORTS_CACHE_KEY_STEM}#{admin_id}", 
  serialized_district_standards_reports, 
  expires_in: serialized_district_standards_reports_cache_life
)
```

services/QuillLMS/app/workers/find_district_standards_reports_worker.rb
```diff
@@ -4,13 +4,18 @@ class FindDistrictStandardsReportsWorker
...
     return unless admin_id
 
     serialized_district_standards_reports_cache_life = 60*60*25
-    serialized_district_standards_reports = ProgressReports::DistrictStandardsReports.new(admin_id).results.to_json
-    $redis.set("#{SchoolsAdmins::DISTRICT_STANDARD_REPORTS_CACHE_KEY_STEM}#{admin_id}", serialized_district_standards_reports)
-    $redis.expire("#{SchoolsAdmins::DISTRICT_STANDARD_REPORTS_CACHE_KEY_STEM}#{admin_id}", serialized_district_standards_reports_cache_life)
-    PusherDistrictStandardsReportsCompleted.run(admin_id)
+    serialized_district_standards_reports = ProgressReports::DistrictStandardsReports.new(admin_id, is_fremium_view).results.to_json
+    if is_fremium_view
+      $redis.set("#{SchoolsAdmins::FREEMIUM_DISTRICT_STANDARD_REPORTS_CACHE_KEY_STEM}#{admin_id}", serialized_district_standards_reports)
```
brendanshean: Similar to `del` you'll want to rewrite this with Rails wrapper [write](https://api.rubyonrails.org/v7.0.4/classes/ActiveSupport/Cache/Store.html#method-i-write).  Note the `set` and `expire` are bundled into one call

```
Rails.cache.write(
  "#{SchoolsAdmins::FREEMIUM_DISTRICT_STANDARD_REPORTS_CACHE_KEY_STEM}#{admin_id}", 
  serialized_district_standards_reports, 
  expires_in: serialized_district_standards_reports_cache_life
)
```

brendanshean (approved):
I left some refactoring comments, but nothing major.  It's cool to see how the blur thing is implemented.  Nice job pulling these pieces together.

services/QuillLMS/app/controllers/api/v1/progress_reports_controller.rb
```diff
@@ -47,15 +47,15 @@ def district_concept_reports
   def district_standards_reports
```
emilia-friedberg, line 47: I probably would have just defined a different controller action and worker here, instead of having to introduce conditional logic into both of the existing ones. I don't think it's necessary to refactor it, but just something to consider since it feels like it's really doing two totally different things. You could still have the different action/worker call the `district_standards_reports` with the freemium parameter.

services/QuillLMS/client/app/bundles/admin_dashboard/containers/AdminDashboard.jsx
```diff
@@ -98,27 +97,6 @@ const AdminDashboard = ({ adminId, accessType, passedModel, }) => {
     );
   }
 
-  function resendLoginDetails(data) {
```
emilia-friedberg, line 101: Was this supposed to get deleted?

emilia-friedberg (approved):

services/QuillLMS/app/models/schools_admins.rb
eadams17: Thanks for the explanation! I'll make the changes for that here and the other places where it's used

services/QuillLMS/app/queries/progress_reports/district_standards_reports.rb
eadams17: yeah, that was a misspelling-- good catch

services/QuillLMS/client/app/actions/district_standards_reports.js
eadams17: There was just a lot of noise in the console while I was developing this so I turned it off, I should add test as well

services/QuillLMS/client/app/bundles/admin_dashboard/containers/AdminDashboard.jsx
eadams17: as far as I can tell, this isn't being used anywhere-- it was just dead code so I got rid of it

PR #10383, opened 2023-03-14
Turn back on pre-cache admin dashboard

brendanshean:
## WHAT
Turn back on pre-cache admin dashboard jobs.

## WHY
There were some slow queries that needed to be cleaned up before running this over all admins.

## HOW
I've added in a couple settings for these jobs
1)  queue: LOW so that these jobs are lowest priority 
2) retry: 0 so that they won't be re-run upon failure clogging up the retry queue

### Screenshots

### Notion Card Links

PR Checklist | Your Answer
------------ | -------------

### DIFF
--- a/services/QuillLMS/app/workers/pre_cache_admin_dashboards_worker.rb
+++ b/services/QuillLMS/app/workers/pre_cache_admin_dashboards_worker.rb
@@ -13,6 +13,9 @@ class PreCacheAdminDashboardsWorker
 
     active_admin_ids.each do |id|
       FindAdminUsersWorker.set(queue: SidekiqQueue::DEFAULT).perform_async(id)
+      FindDistrictActivityScoresWorker.set(queue: SidekiqQueue::DEFAULT).perform_async(id)
+      FindDistrictStandardsReportsWorker.set(queue: SidekiqQueue::DEFAULT).perform_async(id)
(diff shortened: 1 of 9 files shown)


services/QuillLMS/app/workers/pre_cache_admin_dashboards_worker.rb
```diff
@@ -13,6 +13,9 @@ def perform
 
     active_admin_ids.each do |id|
       FindAdminUsersWorker.set(queue: SidekiqQueue::DEFAULT).perform_async(id)
+      FindDistrictActivityScoresWorker.set(queue: SidekiqQueue::LOW, retry: 0).perform_async(id)
```
dandrabik, line 16: Nice! I always forget you can set settings inline like this. 

We could consider adding a standard `retry` limit to the job as well (maybe 0, maybe 1?), in case these queries grow and have similar issues.

dandrabik (approved):
Looks good to me, thanks for putting back in place with additional protections!

anathomical (approved):

services/QuillLMS/app/workers/pre_cache_admin_dashboards_worker.rb
brendanshean: Good idea.  I'll set to 2

PR #10402, opened 2023-03-17
add school info to segment

emilia-friedberg:
## WHAT
Add info about schools for admins to Segment.

## WHY
So that Peter Sharkey can use Ortto to send emails that include these figures to our admins.

## HOW
Add a service to cache and calculate this data, workers and a cron job to cache it every night, and then just read from the cache for `identify` calls for admin users.

### Screenshots

### Notion Card Links
https://www.notion.so/quill/Add-school-information-to-the-Segment-Identify-call-for-admins-0a1eeb6c8728470da619f16df99cfe89?pvs=4

PR Checklist | Your Answer
------------ | -------------


services/QuillLMS/spec/services/analytics/cache_segment_school_data_spec.rb
```diff
@@ -0,0 +1,144 @@
...
+      sc.student.update(last_sign_in: Time.zone.today - 366.days)
+    end
+    students_classrooms2.each do |sc|
+      sc.student.update(last_sign_in: Time.zone.today)
+    end
+  end
+
+  subject { described_class.new(school) }
+
+  # rubocop:disable RSpec/SubjectStub
```
emilia-friedberg, line 29: I think this linter rule is misfiring - we were getting the `Rspec/SubjectStub` error on every `expect(...)to receive` line in this test file, even though none of them are stubbing. Please let me know if I'm missing something here.

services/QuillLMS/app/services/analytics/cache_segment_school_data.rb
```diff
@@ -0,0 +1,97 @@
...
+  end
+
+  def classrooms_at_school
+    @classrooms_at_school ||= Classroom
+      .joins(:classrooms_teachers)
+      .where(:classrooms_teachers => { user_id: teachers_at_school.ids })
+  end
+
+  def teachers_at_school
+    @teachers_at_school ||= @school.users
```
happythenewsad, line 73: There may be an n+1 issue here, since you don't currently use `includes` on line 16. Would you mind verifying? Thanks

services/QuillLMS/app/services/analytics/cache_segment_school_data.rb
```diff
@@ -0,0 +1,97 @@
+# frozen_string_literal: true
+
+class CacheSegmentSchoolData
+  CACHE_KEYS = [
```
happythenewsad: This constant is not used anywhere.

services/QuillLMS/app/services/analytics/cache_segment_school_data.rb
```diff
@@ -0,0 +1,97 @@
...
+
+  def initialize(school)
+    @school = school
+  end
+
+  def read(namespace)
+    Rails.cache.read(@school.id, namespace: namespace)
+  end
+
+  def write(data, namespace)
```
happythenewsad: It's cleaner to use Rails.cache.fetch, as this collapses the read and write operations into a single method + block. `fetch` can also accept an expiry. 

Related: What about batching all this data into a single cache key and cache write? It would significantly reduce the rspec LoC and slightly reduce this file's LoC.

services/QuillLMS/app/services/analytics/cache_segment_school_data.rb
```diff
@@ -0,0 +1,97 @@
...
+    TOTAL_ACTIVITIES_COMPLETED_BY_STUDENTS_AT_SCHOOL_THIS_YEAR = 'TOTAL_ACTIVITIES_COMPLETED_BY_STUDENTS_AT_SCHOOL_THIS_YEAR'
+  ]
+
+  CACHE_LIFE = 60*60*25
+
+  def initialize(school)
+    @school = school
+  end
+
+  def read(namespace)
```
happythenewsad: +1 for adding a namespace, but IMO it should be a constant in this class, to avoid other values getting passed into the `read` and `write` methods. Thoughts?

happythenewsad (changes requested):

services/QuillLMS/app/services/analytics/cache_segment_school_data.rb
emilia-friedberg: I'm not sure how to verify, could you provide guidance? Line 16 isn't a lookup, it's just a definition.

services/QuillLMS/app/services/analytics/cache_segment_school_data.rb
emilia-friedberg: I used separate `read` and `write` methods instead of `fetch` because I don't want the cache to be written to on a read call if there's nothing present - this data is just for analytics, and since the queries are potentially somewhat expensive I think we'd rather stick with the model where they just get updated by the cron job and not on demand. That use case would only come up when new school admins are created for a school that doesn't already have any, but it still seemed worth avoiding to me. LMK if you feel differently.

I can update the data into one cache key - agreed that it doesn't seem like we'll need to pull it out separately.

eadams17 (approved):

services/QuillLMS/app/services/analytics/cache_segment_school_data.rb
happythenewsad: One of the easiest ways is just to run a sample execution in rails console, which, in development mode, prints the SQL statements: 
Screenshot 2023-03-21 at 1 28 19 PM

Notice that `d.teachers_at_school` triggers zero queries, whereas if you don't use `includes`, it will trigger a query.

In this case, there isn't an n+1 issue. However, I still think it's best to pre load the `users` relationships before passing the `school` to `CacheSegmentSchoolData`, since in the future we may implement a method like `admins_at_school` which could call `school.users` under the hood. The end result would be that the `school.users` query is called at least twice.

services/QuillLMS/app/services/analytics/cache_segment_school_data.rb
happythenewsad: Got it; makes sense

happythenewsad (approved):

PR #10411, opened 2023-03-20
lms/strip-stripe-ids-before-validation

anathomical:
## WHAT
Strip Stripe Invoice and Subscription ID whitespace before validation
## WHY
So that if someone copy/pastes an ID but misses leading or trailing whitespace, validation will still work, and the value we care about will be stored.  (This happened to Partnerships, and making their lives easier so that they don't have to super-carefully check the IDs they're pasting for whitespace seems like a reasonable thing to do.
## HOW
Add a `before_validation` callback in the `Subscription` model to handle cases with whitespace

### Notion Card Links
https://www.notion.so/quill/New-subs-not-adding-in-locker-0b51db87f5954b55b25379fca7d57afc?pvs=4

PR Checklist | Your Answer
------------ | -------------

### DIFF
--- a/services/QuillLMS/app/models/subscription.rb
+++ b/services/QuillLMS/app/models/subscription.rb
@@ -117,7 +117,7 @@ class Subscription < ApplicationRecord
   validates :stripe_invoice_id, allow_blank: true, stripe_uid: { prefix: :in }
   validates :stripe_subscription_id, allow_blank: true, stripe_uid: { prefix: :sub }
 
-  before_validation :strip_stripe_id_whitespace, only: [:stripe_invoice_id, :strip_subscription_id]
+  before_validation :strip_stripe_id_whitespace
 
   delegate :stripe_cancel_at_period_end, :stripe_subscription_url,
     to: :stripe_subscription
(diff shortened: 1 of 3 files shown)


services/QuillLMS/app/models/subscription.rb
```diff
@@ -117,6 +117,8 @@ class RenewalNilStripeCustomer < StandardError; end
   validates :stripe_invoice_id, allow_blank: true, stripe_uid: { prefix: :in }
   validates :stripe_subscription_id, allow_blank: true, stripe_uid: { prefix: :sub }
 
+  before_validation :strip_stripe_id_whitespace, only: [:stripe_invoice_id, :strip_subscription_id]
```
emilia-friedberg: I think you're missing an e in `stripe_subscription_id` here. I can't find `only` in the context of validations in the Rails docs - I think it maybe isn't doing anything at all and you want something more like `if: -> { will_save_change_to_stripe_invoice_id? || will_save_change_to_stripe_subscription_id? }`

emilia-friedberg (approved):
I think we want to fix that typo but I don't need to review again.

eadams17 (approved):

services/QuillLMS/app/models/subscription.rb
anathomical: Good catch.  I tried a couple of different options with `if` to restrict when this runs, but none of them were all that readable/functional.  I think the thing to do here is to just always run this hook since it's fast and safe to run whenever.

PR #10419, opened 2023-03-22
Adding teacher advisors section to Quill Team page

cissyxyu:
## WHAT
Add in a section to the Quill team page that shows our teacher advisors names, photos, and titles. Remove the previous "Advisors" section.

## WHY
We want to promote our teacher advisors.

## HOW
Add their images to quill-cdn and add in the extra section with new styling for displaying Teacher Advisors.

### Screenshots
Screen Shot 2023-03-22 at 4 47 48 PM

### Notion Card Links
https://www.notion.so/quill/P2-Add-Teacher-Advisors-to-the-Team-Page-c575254e59e24c3f9840987f4eb8b0d3?pvs=4

PR Checklist | Your Answer
------------ | -------------


eadams17 (approved):
Looks good, but seems like the count is off? We say "the following 13 teachers", but have 15 people in this section (I know that the first two don't have a title of teacher, but the mismatch seems odd to me).

emilia-friedberg (approved):
The structure and everything of this looks good to me, but I think as @eadams17 points out the inclusion of the first two advisors is a design mistake which is why the count is off - maybe worth checking in with Jack and Devin to confirm.

cissyxyu:
> The structure and everything of this looks good to me, but I think as @eadams17 points out the inclusion of the first two advisors is a design mistake which is why the count is off - maybe worth checking in with Jack and Devin to confirm.

Yep, looks like it was a design error! I've discussed with Jack and Devin and we're removing those two extra people from the teacher section.

PR #10424, opened 2023-03-23
support verification by existing admins in sign up

emilia-friedberg:
## WHAT
Add view to existing admin sign up flow for users who want to be admins for schools that already have admins in Quill, and build out the backend to support the interface and Segment events we need for those admins to actually verify them.

## WHY
This mimics the pattern we use for existing teachers who are requesting an upgrade, and will theoretically reduce load on support.

## HOW
Build out the frontend, and update all relevant routes to determine whether to use the Segment events for an existing teacher or a new user. Add a migration to add a `request_made_during_sign_up` column to the `admin_approval_requests` table so we can determine that value for the UI and Segment events.

### Screenshots
Screen Shot 2023-03-22 at 12 22 49 PM
Screen Shot 2023-03-22 at 11 13 07 AM

### Notion Card Links
https://www.notion.so/quill/Update-the-admin-sign-up-process-to-support-verification-by-existing-admins-c349feceeb0e4773993623246ca0a55d?pvs=4

PR Checklist | Your Answer
------------ | -------------


eadams17 (approved):
Everything looks good to me-- thanks for knocking this out!

services/QuillLMS/app/controllers/admin_access_controller.rb
```diff
@@ -19,13 +19,14 @@ def upgrade_to_admin
   def request_upgrade_to_admin_from_existing_admins
     admin_ids = request_upgrade_to_admin_from_existing_admins_params[:admin_ids]
     reason = request_upgrade_to_admin_from_existing_admins_params[:reason]
+    new_user = request_upgrade_to_admin_from_existing_admins_params[:new_user].to_s == 'true'
```
anathomical: I'm not sure that this is actually more readable, but you could use the built in Rails tool to figure out if a string should be `true` or `false` with `ActiveRecord::Type::Boolean.new.deserialize(request_upgrade_to_admin_from_existing_admins_params[:new_user])` instead of comparing to "true"

services/QuillLMS/client/app/bundles/Teacher/components/admin_access/adminTable.tsx
```diff
@@ -0,0 +1,71 @@
+import * as React from 'react'
```
anathomical, line 1: Just wanted to say that I like that you've broken this out into a stand-alone component.

services/QuillLMS/spec/controllers/admin_access_controller_spec.rb
```diff
@@ -70,15 +70,16 @@
   describe '#request_upgrade_to_admin_from_existing_admins' do
     it 'creates an admin info record with the requisite information, an admin approval request record, and fires the necessary segment workers' do
       reason = 'I really want to be an admin.'
+      new_user = [true, false].sample
```
anathomical: I like the impulse to not repeat all these expectations for both `true` and `false`, and from a practical perspective, this should be just fine.  However, it is possible (if statistically unlikely) that only either `true` or `false` paths are actually working and that `sample` just keeps returning the working value.  I think you can make sure that both branches run every test with a small tweak:
```ruby
[true, false].each do |new_user|
  # put all your test code from line 74 to 82 inside this block, and it'll run twice, once for each value
end
```

services/QuillLMS/spec/controllers/admins_controller_spec.rb
```diff
@@ -182,12 +182,10 @@
...
 
   describe '#approve_admin_request' do
-    before do
-      create(:admin_info, user: teacher, approval_status: AdminInfo::PENDING)
-    end
-
     it 'should update the teacher admin info to be approved, create a school admin record, call the analytics worker, and return a message' do
-      expect(TeacherApprovedToBecomeAdminAnalyticsWorker).to receive(:perform_async).with(teacher.id)
+      admin_info = create(:admin_info, user: teacher, approval_status: AdminInfo::PENDING)
+      admin_approval_request = create(:admin_approval_request, admin_info: admin_info, requestee: admin, request_made_during_sign_up: [true, false].sample )
```
anathomical: Given the use of the `[true, false].sample` pattern, I wonder if it might make sense to put that in the AdminInfo factory instead of setting it on each `create` call.

services/QuillLMS/spec/workers/admin_received_admin_upgrade_request_from_teacher_analytics_worker_spec.rb
```diff
@@ -10,12 +10,13 @@
     let!(:teacher) { create(:teacher) }
     let(:analyzer) { double(:analyzer) }
     let(:reason) { 'Please' }
+    let(:new_user) { [true, false].sample }
```
anathomical: This might be another good place to use a `[true, false].each do |new_user|` block.  Or, more precisely, in the `it` block on line 17.

anathomical (approved):
A couple of minor comments, but the approach here is solid.  You could merge as-is without any problems, I think.

services/QuillLMS/spec/controllers/admins_controller_spec.rb
emilia-friedberg: Good call!

PR #10437, opened 2023-03-28
add note for district admins

emilia-friedberg:
## WHAT
Add note that appears when a user selects the "District administrator" option, and also make sure the Teacher admin option doesn't show up on that page.

## WHY
To clarify that they can only select one school for now, but will receive an email with follow up.

## HOW
Just React/CSS.

### Screenshots
Screen Shot 2023-03-28 at 10 15 58 AM

### Notion Card Links
https://www.notion.so/quill/Remove-Teacher-admin-as-an-option-for-sub-role-on-sign-up-e0bb782ebb304c0f9ba89df9734bcdfe?pvs=4
https://www.notion.so/quill/Display-a-note-if-someone-selects-the-District-administrator-sub-role-when-signing-up-as-an-admin-bf5d156de4364d599661579f124ae9f8?pvs=4

PR Checklist | Your Answer
------------ | -------------


services/QuillLMS/app/models/admin_info.rb
```diff
@@ -45,9 +45,9 @@ class AdminInfo < ApplicationRecord
     BILLING_CONTACT = "Billing contact"
   ]
 
-  SUB_ROLES = SELECTABLE_SUB_ROLES.concat([
+  SUB_ROLES = SELECTABLE_SUB_ROLES + [
     TEACHER_ADMIN = 'Teacher admin'
-  ])
+  ]
```
eadams17, line 50: just curious about what the difference is between using `.concat([value])` and `+ [value]` ?

eadams17 (approved):

services/QuillLMS/app/models/admin_info.rb
emilia-friedberg: Great question! `.concat` was modifying the original array, so both of them were returning the same list.

PR #10443, opened 2023-03-29
evidence/add-positive-automl-concept-results

anathomical:
## WHAT
Modify Evidence conceptResult generation to include topic-optimals
## WHY
Basically, we want to include an optimal concept result for the AutoML concept when specific feedback types are received.  Specifically, if a feedback type is lower priority than AutoML, we want to flag that AutoML passed by generating an optimal ConceptResult.
## HOW
- Create a list of `feedback_type` values that count as "topic-optimal"
- Create a hard-coded list of the `concept_uid` for each of the conjunction types
- Modify the logic that assembles Evidence `ConceptResult` objects so that if it's building a non-optimal `ConceptResult` for a feedback_type that is in the topic-optimal list, then it also creates an optimal `ConceptResult` for the AutoML `Concept` for the stem of the prompt being processed.

NOTE: I would especially like feedback and thoughts on how brittle the coupling is between this code and what's actually in the database.  I think that the `concept_uid` values are likely to be pretty stable going forward, but am less sure about whether or not we'll be adding additional algorithms to the Feedback endpoint.

### Screenshots
image

### Notion Card Links
https://www.notion.so/quill/Edit-Logic-around-Evidence-Concept-Tags-e199a2a58719469ba3fe8712129eff77

PR Checklist | Your Answer
------------ | -------------


services/QuillLMS/client/app/bundles/Evidence/libs/conceptResults.ts
```diff
@@ -16,25 +16,59 @@ export const generateConceptResults = (currentActivity, submittedResponses) => {
     so: 3
   }
 
+  // This list is manually copied from the production database.
+  // It is a very brittle integration since there's nothing that keeps it
+  // synced with what's actually going on in the database
+  const conjunctionToSemanticConceptUid = {
+    because: "qkjnIjFfXdTuKO7FgPzsIg",
```
brendanshean: What about adding a `conjunctionToSemanticConceptUid` parameter to `generateConceptResults` and  then populating those uid higher upstream via some initializer in Rails?

anathomical: Am I understanding the suggestion here that we'd pull this data in Rails and then inject it into the front-end through the `ReactOnRails` initializer and then pass it through various components until it arrives here?

happythenewsad: I think that's what Brendan is saying. Also: could use the hydrate-from-rails strategy for `postAutomlFeedbackTypes` as well. 

Many valid ways to do this, with their respective pros and cons:
- use an AppSetting
- hardcode in js, like currently
- fetch on react component load (would want to cache)
- ... others

anathomical: I've refactored this per our Zoom discussion so that it now uses an API to fetch topic-optimal data for calculating concept results.

services/QuillLMS/client/app/bundles/Evidence/components/studentView/container.tsx
```diff
@@ -103,6 +103,7 @@ export const StudentViewContainer = ({ dispatch, session, isTurk, location, acti
   }, [hasStartedReadPassageStep]);
 
   React.useEffect(() => {
+    dispatch(getTopicOptimalInfo(activityUID))
```
happythenewsad, line 106: Curious why you opted for the redux-y `dispatch` pattern here, when `fetch + useState` or `useReducer` would have accomplished the same thing. 

I'm not a redux expert, but it seems like hooks would require less code in this case.

happythenewsad (approved):

services/QuillLMS/client/app/bundles/Evidence/components/studentView/container.tsx
anathomical: This is a totally reasonable point.  It may be a mistake, but my thought here was to follow the existing pattern already established in this file.  (Specifically, I'm copying the logic used in the `getActivity` action in this file already, since that's the other API call that gets made.)  I'm not sure this is the pattern we'd use if we were writing this component from scratch now, and would be willing to entertain the idea that we should build new components in the current state-of-the-art rather than re-using old approaches for new components.  Thoughts?

happythenewsad: IMO keeping the existing pattern here is fine, since you already have a PR hot and ready. 

In the future, I encourage us to prefer hooks over redux, in cases where we don't need complex reducer logic (which, for this app, is most cases).

anathomical: Sounds good.  Might also be worth getting input from @eadams17 or @emilia-friedberg while I'm gone to figure out what patterns they want to use for the front-end moving forward.

emilia-friedberg: Yes, agreed with @happythenewsad that we should be opting for hooks where possible in general - I think this app was written before we were really using them and we just haven't updated old code to move towards that pattern. I also agree that it isn't something that needs to change for this PR, but makes sense to keep in mind in the future.

emilia-friedberg (approved):

eadams17 (approved):

PR #10458, opened 2023-04-04
LearnWorlds SSO backend integration

brendanshean:
## WHAT
Add LearnWorlds SSO backend integration.

## WHY
We want teachers to be able to sign into LearnWorlds without having to create a new account.

## HOW
When user a clicks an auth_learn_worlds_course_url button/form, two actions occurs.

1) A POST request will be [sent](https://github.com/empirical-org/Empirical-Core/blob/a2f140bbba3f5c9df2cd0b89b62fc640264abd85/services/QuillLMS/app/services/auth/learn_worlds/sso_request.rb#L21) to the LearnWorlds SSO endpoint.
2) Upon successful response, the user is [redirected](https://github.com/empirical-org/Empirical-Core/blob/a2f140bbba3f5c9df2cd0b89b62fc640264abd85/services/QuillLMS/app/controllers/auth/learn_worlds_controller.rb#L9) to the LearnWorlds site, logged in.
  a) If the users has an existing learn_worlds_account, they are logged in via their learn_worlds_account.external_id
  b) Otherwise, for the initial sign on, the user is logged in via their email and a learn_worlds_account is attached to that user

### Screenshots

### Notion Card Links

PR Checklist | Your Answer
------------ | -------------

### DIFF
--- a/services/QuillLMS/spec/requests/auth/learn_worlds_controller_spec.rb
+++ b/services/QuillLMS/spec/requests/auth/learn_worlds_controller_spec.rb
@@ -5,7 +5,7 @@ require 'rails_helper'
 RSpec.describe Auth::LearnWorldsController do
   subject { post auth_learn_worlds_courses_path }
 
-  let(:courses_endpoint) { stub_const('Auth::LearnWorlds::COURSES_ENDPOINT', 'https://learnworlds.com/courses') }
+  let(:courses_endpoint) { 'https://learnworlds.com/courses' }
 
   before do
     allow(User).to receive(:find).with(user.id).and_return(user)
--- a/services/QuillLMS/spec/requests/auth/learn_worlds_controller_spec.rb
+++ b/services/QuillLMS/spec/requests/auth/learn_worlds_controller_spec.rb
@@ -1,0 +1,69 @@
+# frozen_string_literal: true
+
+require 'rails_helper'
+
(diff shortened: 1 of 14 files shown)


emilia-friedberg (approved):
LGTM!

services/QuillLMS/spec/requests/auth/learn_worlds_controller_spec.rb
```diff
@@ -0,0 +1,69 @@
+# frozen_string_literal: true
+
+require 'rails_helper'
+
+RSpec.describe Auth::LearnWorldsController do
+  subject { post auth_learn_worlds_courses_path }
+
+  let(:courses_endpoint) { stub_const('Auth::LearnWorlds::COURSES_ENDPOINT', 'https://learnworlds.com/courses') }
```
dandrabik: Do you need this `stub_const` here (vs. just the url string). I'm not sure the constant is used in these tests.

dandrabik (approved):
This looks good. I like the thorough tests, this feels very solid. Well done!

🚢

services/QuillLMS/spec/requests/auth/learn_worlds_controller_spec.rb
brendanshean: Good point.  Since I'm stubbing out the HTTParty call we don't need it.

PR #10461, opened 2023-04-05
Add LearnWorldsAccount model

brendanshean:
## WHAT
Create a new model `LearnWorldsAccount` and accompanying table.

## WHY
We need a place to persist individual LearnWorld's account external ids in our system.

## HOW
`rails generate model user:references external_id`

### Screenshots

### Notion Card Links

PR Checklist | Your Answer
------------ | -------------

### DIFF
--- a/services/QuillLMS/app/models/learn_worlds_account.rb
+++ b/services/QuillLMS/app/models/learn_worlds_account.rb
@@ -1,0 +1,23 @@
+# frozen_string_literal: true
+
+# == Schema Information
+#
+# Table name: learn_worlds_accounts
+#
+#  id          :bigint           not null, primary key
+#  created_at  :datetime         not null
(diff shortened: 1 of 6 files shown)


dandrabik (approved):
Overall looks good. My instinct with this kind of code is to abstract it a bit for future use, e.g. `ServiceAccounts` with a `type` field with the value `LearnWorld`, but since we don't have a planned second use, I think it's fine to be specific here. :+1:

brendanshean:
> Overall looks good. My instinct with this kind of code is to abstract it a bit for future use, e.g. `ServiceAccounts` with a `type` field with the value `LearnWorld`, but since we don't have a planned second use, I think it's fine to be specific here. 👍

I was wrestling with this idea too, but I think I will keep it simple here for now.

PR #10466, opened 2023-04-05
Teacher Center/ restrict drafts from recommended articles section

eadams17:
## WHAT
restrict drafts from recommended articles section

## WHY
we only want to show published posts in the recommended articles section

## HOW
update query to only select `draft: true`

### Screenshots

### Notion Card Links
https://www.notion.so/quill/Articles-that-are-no-longer-published-showing-up-in-TC-footer-43b44d8f6bf44b89ba889b1bbfd139ba?pvs=4

PR Checklist | Your Answer
------------ | -------------


services/QuillLMS/app/models/blog_post.rb
```diff
@@ -146,6 +146,7 @@ def related_posts
     BlogPost
       .where(topic: topic)
       .where.not(id: id)
+      .where(draft: false)
```
happythenewsad: There's already a scope on line 91 that does this. Let's use the scope, or delete the scope

happythenewsad (changes requested):

services/QuillLMS/app/models/blog_post.rb
happythenewsad: This would also be a classic place to use a default scope, but i'm personally not a fan of default scopes because they can make debugging difficult if you forget that they exist.

happythenewsad (approved):

PR #10489, opened 2023-04-13
Add Canvas{Account, Config, Instance, InstanceSchool} backend

brendanshean:
## WHAT
1. Add `CanvasInstance` model and table
1. Add `CanvasConfig` model and table
1. Add `CanvasAccount` model and table
1. Add `CanvasInstanceSchool` model and table
1. Add lockbox gem

## WHY
1. Creating a `CanvasInstance` will allow users to connect their canvas instance to Quill
1. A `CanvasConfig` contains credentials used when connecting a CanvasInstance to Quill
1. `CanvasAccount` is a join table that connect users to canvas instances
1. `CanvasInstanceSchool` is a join table connecting schools and canvas_instances (See linked notion card for more details on this addendum)
1. Each `CanvasConfig` contains client_id and client_secret attributes which are encrypted using Lockbox

## HOW
1.  Add the necessary model and migration
1. Add the necessary model and migration. (NB. The attribute names for the encrypted columns have the suffix. `_ciphertext`)
1. Add the necessary model and migration
1. Add the necessary model and migration
1. bundle install lockbox and set `ENV['LOCKBOX_MASTER_KEY'] = Lockbox.generate_key`

### Screenshots

### Notion Card Links
](https://www.notion.so/quill/Canvas-Configurations-33c5e500e9154a90b3bac1d963e2ef67?pvs=4)

PR Checklist | Your Answer
------------ | -------------


services/QuillLMS/db/migrate/20230411145241_create_canvas_configs.rb
```diff
@@ -0,0 +1,13 @@
+# frozen_string_literal: true
+
+class CreateCanvasConfigs < ActiveRecord::Migration[6.1]
+  def change
+    create_table :canvas_configs do |t|
```
dandrabik, line 5: How is this connected to a school or district?

services/QuillLMS/db/migrate/20230413140558_create_canvas_accounts.rb
```diff
@@ -0,0 +1,12 @@
+# frozen_string_literal: true
+
+class CreateCanvasAccounts < ActiveRecord::Migration[6.1]
+  def change
+    create_table :canvas_accounts do |t|
```
dandrabik, line 5: Should this be a habtm between users and configs rather than some relationship between schools/districts and configs? How do schools/districts fit into this current structure? I also wonder if we should lean on the SchoolAdmin/DistrictAdmin relationship for permission-ing access to users here.

It might be worth us talking through / mapping out the modeling of this feature and the rest of Canvas.

services/QuillLMS/Gemfile
```diff
@@ -153,6 +153,9 @@ gem 'paperclip'
 gem 'rb-readline'
 gem 'awesome_print'
 
+# TODO: to be removed with Rails 7
```
dandrabik: Just to clarify, is there a migration path from this to what Rails 7 uses?

Also, we might want to clarify in this comment that Rails has similar functionality to this gem, so we'd want to migrate our usage to the Rails 7 structure, then remove the gem.

services/QuillLMS/app/models/canvas_config.rb
```diff
@@ -0,0 +1,19 @@
...
+#
+#  id                       :bigint           not null, primary key
+#  client_id_ciphertext     :text             not null
+#  client_secret_ciphertext :text             not null
+#  url                      :string           not null
+#  created_at               :datetime         not null
+#  updated_at               :datetime         not null
+#
+class CanvasConfig < ApplicationRecord
+  has_encrypted :client_id, :client_secret
```
dandrabik, line 23: It's nice how easy this is.

dandrabik (changes requested):
Thanks for putting this together. Adding the gem looks good. I think we should talk through this modeling a bit, since I'm not seeing how schools/districts fit in.

services/QuillLMS/Gemfile
brendanshean: For a migration path, my plan was to create new fields on the  `client_id` and `client_secret` and copy the unencrypted values over to those new fields.  Then remove the attributes `client_id_ciphertext` and `client_secret_ciphertext` from the table.

brendanshean: I've updated the TODO comment.

services/QuillLMS/db/migrate/20230411145241_create_canvas_configs.rb
brendanshean: Good point.  Per our discussion I will need to add some reference to school[s] and/or district.

happythenewsad (approved):

services/QuillLMS/db/migrate/20230413140558_create_canvas_accounts.rb
```diff
@@ -0,0 +1,13 @@
...
+      t.string :external_id, null: false
+      t.references :user, foreign_key: true, null: false
+      t.references :canvas_config, foreign_key: true, null: false
```
dandrabik: I think we should talk through this table. I'm guessing that this came up after we chatted [about this schema](https://www.notion.so/quill/Canvas-Configurations-33c5e500e9154a90b3bac1d963e2ef67), but I'm wondering if we are concerned that we are creating multiple paths to `canvas_config` from a `user` that can drift from each other, 
i.e. `user->canvas_config`, `user->school->canvas_config`.

We might consider linking to `school` here or not linking to anything and always going through the `user->school->canvas_config` path. Or keep what you have here and code around possible duplication.

Either way, I think it's worth talking through the pros and cons.

brendanshean: While I was aware of the duplicate paths, my goal was to try to keep the schema structure similar to the `learn_worlds_account`.  After looking through this again, I would agree that the multiple paths should be removed.

services/QuillLMS/app/models/validators/url_validator.rb
```diff
@@ -0,0 +1,16 @@
+# frozen_string_literal: true
+
+class UrlValidator < ActiveModel::EachValidator
```
dandrabik, line 3: Nice, this is useful, thanks for adding this.

dandrabik (approved):
This looks good to me! Thanks for iterating through this with me. I think this landed in a strong place, and I appreciate your willingness to discuss different options until we shook out a good solution.
🚢

PR #10490, opened 2023-04-13
Bigquery integration

happythenewsad:
## WHAT
- rewrites an existing expensive postgres query in bigquery 
- creates integration and unit style test framework for bigquery going forward

## WHY
- so that we can leverage BigQuery for expensive reports 

## HOW
- Create LMS<>BigQuery ruby wrapper
- port SQL query to BigQuery SQL

### ALSO
The Notion card specifies implementing a second report. I plan to ship the second report once the overall approach in this PR is approved. 

In addition to the code tests, I tested on staging by:
1. select an admin
2. Download that admin's district concept report, CSV
3. Delete the user's report redis key `"SERIALIZED_DISTRICT_CONCEPT_REPORTS_FOR_3737095"`
4. Run the worker `FindDistrictConceptReportsWorker.new.perform(3737095)` 
5. Download the admin's report again, and spot check. Some values will be different, since staging data significantly lags production BigQuery.

### Notion Card Links
https://www.notion.so/quill/Product-Board-30f97e4eb01246dbb8264253fb385073?p=b06ee1bef5e34e17b4515137e73fedca&pm=s

PR Checklist | Your Answer
------------ | -------------


services/QuillLMS/app/queries/quill_big_query.rb
```diff
@@ -0,0 +1,70 @@
...
+
+    # Parsing - It returns row names and values separately in an odd setup
+    # This parses them down to [{"id" => 1, "name"=> "Dan", }, {"id" => 2, "name" => "Peter"}]
+    fields = json_body["schema"]["fields"].map {|h| h["name"]}
+    values = json_body["rows"].map {|r| r.values_at("f").flatten.map {|h| h.values.first}}
+    hash_results = values.map {|v| fields.zip(v).to_h}
+    floatify_fields(hash_results)
+  end
+
+  def self.get_response(query)
```
happythenewsad: Credit to Dan for writing the precursor to this method

services/QuillLMS/spec/queries/progress_reports/district_concept_reports_spec.rb
```diff
@@ -3,37 +3,32 @@
...
-    let!(:admin) { create(:teacher) }
-    let!(:classroom) { create(:classroom) }
-    let!(:student) { create(:student) }
-    let!(:schools_admins) { create(:schools_admins, school: school, user: admin) }
-    let!(:schools_users) { create(:schools_users, school: school, user: teacher) }
-    let!(:classrooms_teacher) { create(:classrooms_teacher, user: teacher, role: "owner", classroom: classroom) }
-    let!(:classroom_unit) { create(:classroom_unit, classroom: classroom, assigned_student_ids: [student.id]) }
-    let!(:activity_session) { create(:activity_session_without_concept_results, classroom_unit: classroom_unit, user_id: student.id) }
-    let!(:concept_result) { create(:concept_result, activity_session: activity_session) }
+  context 'integration', :integration do
```
happythenewsad: Notice the 'integration' filter here. Rspec will not run this unless passed `-t integration`

services/QuillLMS/spec/queries/progress_reports/district_concept_reports_spec.rb
```diff
@@ -3,37 +3,32 @@
...
+    let(:sample_prod_admin_id) { 3737095 }
+    let(:expected_row_keys) do
+      %w(school_name teacher_name classroom_name student_name correct incorrect percentage)
+    end
+
+    around do |a_spec|
```
happythenewsad, line 104: I wanted this to be a true integration test, so I'm bypassing VCR. I mock BigQuery responses in a following file, for unit style testing.

services/QuillLMS/app/queries/quill_big_query.rb
```diff
@@ -0,0 +1,70 @@
...
+  class UnsupportedSchemaError < StandardError; end
+
+  def self.valid_schema?(json_body)
+    return false unless json_body['schema']['fields'].respond_to?(:count)
+    return false unless json_body['rows'].respond_to?(:count)
+
+    true
+  end
+
+  def self.floatify_fields(array_of_hashes)
```
happythenewsad: This transform is specific to `ProgressReports::DistrictConceptReports`. As I add more queries to this framework, I plan to abstract idiosyncratic transforms into procs, which QuillBigQuery will apply as a post-processing step.

services/QuillLMS/spec/rails_helper.rb
```diff
@@ -81,6 +81,8 @@
   # focus tests
   config.filter_run focus: true
   config.filter_run_excluding benchmarking: true
+  config.filter_run_excluding integration: true
```
brendanshean: I think the 'integration' tag is confusing since other types of integration tests not labeled integration (e.g. system, request, controller) are included by default.

Maybe something descriptive about why it's being excluded like 'big_query', 'vcr', etc.

services/QuillLMS/app/queries/progress_reports/district_concept_reports.rb
```diff
@@ -11,41 +11,42 @@ def results
...
           classrooms.name AS classroom_name,
-          students.name AS student_name,
+          MAX(students.name) AS student_name,
           students.id AS student_id,
-          SUM(CAST(concept_results.correct as INT)) AS correct,
+          SUM(CAST(concept_results.correct as INT64)) AS correct,
           COUNT(concept_results) AS concept_results_count
-        FROM schools_admins
-        JOIN schools
+        FROM lms.schools_admins
```
brendanshean, line 29: Where does `lms` come from here?

services/QuillLMS/app/queries/quill_big_query.rb
```diff
@@ -0,0 +1,70 @@
...
+  def self.get_response(query)
+    user_agent = ["quill/0.0.0", "google-api-ruby-client/0.8.6", "Mac OS X",'(gzip)'].join(' ').delete("\n")
+    client = Google::APIClient.new(application_name: 'quill', user_agent: user_agent)
+
+    # API discovery https://github.com/googleapis/google-api-ruby-client/tree/v0.8.6#api-discovery
+    bigquery = client.discovered_api('bigquery', 'v2')
+    scope = 'https://www.googleapis.com/auth/bigquery'
+
+    # make_creds expects a file, so we wrap the JSON env var in an IO stream
+    json_credential_io = StringIO.new(ENV.fetch('BIGQUERY_CREDENTIALS'))
```
brendanshean: I would use an an initializer for BIGQUERY_CREDENTIALS in config/intiailzers and then reference that here.

services/QuillLMS/app/queries/quill_big_query.rb
```diff
@@ -0,0 +1,70 @@
...
+    return false unless json_body['rows'].respond_to?(:count)
+
+    true
+  end
+
+  def self.floatify_fields(array_of_hashes)
+    array_of_hashes.map do |hsh|
+      hsh.each {|k,v| FLOAT_FIELDS.include?(k) && hsh[k] = v.to_f }
+    end
+    array_of_hashes
```
brendanshean: I think this line is redundant since  `map` block should return the same object

services/QuillLMS/app/queries/quill_big_query.rb
```diff
@@ -0,0 +1,70 @@
...
+require 'stringio'
+
+module QuillBigQuery
+  PROJECT_ID = 'analytics-data-stores'
+  FLOAT_FIELDS = %w(correct incorrect percentage)
+
+  class UnsupportedSchemaError < StandardError; end
+
+  def self.valid_schema?(json_body)
+    return false unless json_body['schema']['fields'].respond_to?(:count)
```
brendanshean: I think this one-liner might be more readable as it avoids the false/unless:

`json_body['schema']['fields'].respond_to?(:count) && json_body['rows'].respond_to?(:count)`

services/QuillLMS/app/queries/progress_reports/district_concept_reports.rb
```diff
@@ -11,41 +11,42 @@ def results
...
   private def query
+    # For testing, see the Metabase version of this query: https://data.quill.org/question/1026-progressreports-districtconceptreports
     <<~SQL
       WITH results AS (
         SELECT
           schools.name AS school_name,
           teachers.name AS teacher_name,
           classrooms.name AS classroom_name,
-          students.name AS student_name,
+          MAX(students.name) AS student_name,
```
brendanshean, line 25: Why the change to `MAX`?

services/QuillLMS/app/queries/quill_big_query.rb
```diff
@@ -0,0 +1,70 @@
...
+    end
+    array_of_hashes
+  end
+
+  def self.transform_response(json_body)
+    raise UnsupportedSchemaError, json_body unless valid_schema?(json_body)
+
+    # Parsing - It returns row names and values separately in an odd setup
+    # This parses them down to [{"id" => 1, "name"=> "Dan", }, {"id" => 2, "name" => "Peter"}]
+    fields = json_body["schema"]["fields"].map {|h| h["name"]}
```
brendanshean: You can use pluck here:  ` json_body["schema"]["fields"].pluck("name")`

services/QuillLMS/app/queries/quill_big_query.rb
```diff
@@ -0,0 +1,70 @@
...
+    values = json_body["rows"].map {|r| r.values_at("f").flatten.map {|h| h.values.first}}
```
brendanshean: And then pluck again:
`values = json_body['rows'].pluck('f').map { |h| h.pluck('v') }`

services/QuillLMS/app/queries/quill_big_query.rb
```diff
@@ -0,0 +1,70 @@
...
+    hash_results = values.map {|v| fields.zip(v).to_h}
```
brendanshean: There's a `to_h` method on array that takes a block:

`hash_results = values.to_h { |v| fields.zip(v) }`

services/QuillLMS/app/queries/quill_big_query.rb
brendanshean: I would probably extract all of the client code (lines ~41-57) into its own file which allows for easier to stub out during testing while also providing a cleaner separation of concerns.

services/QuillLMS/spec/queries/quill_big_query_spec.rb
```diff
@@ -0,0 +1,66 @@
...
+      let(:invalid_response) { { foo: 1 } }
+      let(:another_invalid_response) do
+        {
+          "schema" => {
+            "fields" => 42
+          }
+        }
+      end
+
+      it 'should raise error' do
```
brendanshean: I would break this up into two tests since they are testing different exceptions:

```
it { expect { QuillBigQuery.transform_response(invalid_response)}.to raise_error(NoMethodError) }

it { expect { QuillBigQuery.transform_response(another_invalid_response)}.to raise_error(QuillBigQuery::UnsupportedSchemaError) }

```

brendanshean (approved):
Some minor refactoring comments and then one question about end-to-end-testing, but otherwise,  great job figuring out this integration.

services/QuillLMS/spec/queries/progress_reports/district_concept_reports_spec.rb
brendanshean: How is circleci picking up  'BIGQUERY_CREDENTIALS'

services/QuillLMS/app/queries/progress_reports/district_concept_reports.rb
happythenewsad: BigQuery is stricter with qualifying tables. Postgres does this implicitly, by assuming `mytable` qualifies to `public.mytable`

services/QuillLMS/spec/queries/progress_reports/district_concept_reports_spec.rb
happythenewsad: CircleCI doesn't see this file due to the `integration` filter. However, you can set up secret env vars in CircleCi in `Project > Settings`, same pattern as Heroku.

brendanshean: Good point

services/QuillLMS/app/queries/quill_big_query.rb
dandrabik: I agree that there's a `BigQueryRequest` type of class here that should be extracted. It could be an `ApplicationService` with a main `run` method and broken down into several constants and submethods.

services/QuillLMS/app/queries/quill_big_query.rb
```diff
@@ -0,0 +1,70 @@
+# frozen_string_literal: true
+
+# All messages to BigQuery should use this module
+
+require 'google/api_client'
+require 'googleauth'
+require 'stringio'
+
+module QuillBigQuery
```
dandrabik: Take or leave. Consider making this an `ApplicationService` with an `initialize` and `run` to follow a common pattern in our codebase.

services/QuillLMS/app/queries/quill_big_query.rb
```diff
@@ -0,0 +1,70 @@
...
+  PROJECT_ID = 'analytics-data-stores'
+  FLOAT_FIELDS = %w(correct incorrect percentage)
```
dandrabik: Hmm, this `FLOAT_FIELDS` constant only applies to this query. I don't think this is going to scale well, e.g. every query you'd have to remember to update this constant, plus what if there is an overlap in field names among different queries (correct is a float here, but could be a boolean in a different query). 

There is a field schema returned by API. I think we'd want to use that to determine any transforms. That's a more scalable approach.

services/QuillLMS/app/queries/quill_big_query.rb
dandrabik: I think we should reconsider this approach. Asking developers to define a post-query transform proc for every query has a lot of downsides (adds overhead to the query process, it will be hard to keep approaches consistent).

However, the API returns a schema definition for the result. I think it will be more simple and consistent to do basic transforms based on this schema rather than custom procs for each query. BigQuery only has [a handful of types](https://cloud.google.com/bigquery/docs/reference/standard-sql/data-types), we can just write transforms for each once, then every query would return results in the correct type. It's more work upfront, but will make this easier to use long-term.

dandrabik (changes requested):
Thanks for moving this along. Really great progress here.

I'm requesting changes since I think we need a different approach in two areas
1. How we transform the returned values. I think we want to do this universally based on the schema returned by the API, rather than ad-hoc per query.
2. I think some of the longer, more complex methods here should be refactored to service object classes. That will also make step 1 easier.

Happy to chat through this. Thanks for your work on this.
I think there is some opportunity for some refactoring of long methods here.

services/QuillLMS/app/queries/progress_reports/district_concept_reports.rb
happythenewsad: All SELECTed elements must either be aggregates, or a group by component. It's unexpected that postgres allowed the original.

happythenewsad:
Responded to feedback. Highlights:

- re-arrange code into new files (client, runner) within app/services/quill_big_query
- refactor transforms on the assumption that they will be universal, rather than proc-based (this may need iteration as we encounter any edge cases in upcoming queries)
- added a more comprehensive, stubbed spec for `ProgressReports::DistrictConceptReports`

happythenewsad:
Note: after speaking with Dan, we've decided to increase the scope of this PR to include all type coercions, not just the ones required for this PR. I will add this change to this PR today.

services/.rubocop.yml
```diff
@@ -554,7 +554,7 @@ Style/DocumentDynamicEvalDefinition:
   Enabled: true
 
 Style/EachWithObject:
-  Enabled: true
+  Enabled: false
```
happythenewsad, line 557: I find that `reduce` is much more intuitive than `each_with_object`. Both aliases point to a function that has reducing behavior.

services/QuillLMS/app/services/quill_big_query/transformer.rb
```diff
@@ -0,0 +1,57 @@
...
+      @field_to_lambda = self.class.build_field_to_lambda_lookup(fields)
+      @array_of_hashes = array_of_hashes
+    end
+
+    LAMBDA_IDENTITY     = ->(val) { val }
+    LAMBDA_TO_BOOLEAN   = ->(val) { val == 'true' }
+    LAMBDA_TO_DATETIME  = ->(val) { DateTime.parse(val) }
+    LAMBDA_TO_F         = ->(val) { val.to_f }
+
+    TYPE_TO_LAMBDA_LOOKUP = {
```
happythenewsad, line 20: You may notice that this list is a superset of https://cloud.google.com/bigquery/docs/reference/standard-sql/data-types 

This is because I noticed empirically, and in non-official BigQuery documentation, that `BOOLEAN` and `FLOAT64` are returned values, in addition to `BOOL` and `FLOAT`.

services/QuillLMS/app/services/quill_big_query/runner.rb
```diff
@@ -0,0 +1,51 @@
...
+      bigquery = client.discovered_api('bigquery', 'v2')
+
+      # API reference: https://cloud.google.com/bigquery/docs/reference/rest/v2/jobs/query?apix_params=%7B%22projectId%22%3A%22analytics-data-stores%22%2C%22resource%22%3A%7B%22query%22%3A%22SELECT%20email%20from%20lms.users%20LIMIT%201%22%7D%7D
+      begin
+        result = client.execute(
+          api_method: bigquery.jobs.query,
+          parameters: {'projectId' => PROJECT_ID},
+          body_object: {'query' => query, 'useLegacySql' => false}
+        )
+      rescue => e
```
happythenewsad, line 40: I added this rescue block because network calls from local dev to BigQuery fail occasionally, and the error is a parse error. The intent of this wrap and re-raise is to make it more obvious when an error arises from the network call.

services/QuillLMS/app/services/quill_big_query/transformer.rb
```diff
@@ -0,0 +1,59 @@
...
+      'INTEGER'     => LAMBDA_TO_F,
+      'INT64'       => LAMBDA_TO_F,
+      'NUMERIC'     => LAMBDA_TO_F,
+      'STRING'      => LAMBDA_IDENTITY,
+      'STRUCT'      => LAMBDA_IDENTITY,
+      'TIME'        => LAMBDA_TO_DATETIME,
+      'TIMESTAMP'   => LAMBDA_TO_DATETIME   # '2021-05-01 21:32:45'
+    }
+
+    def self.build_field_to_lambda_lookup(fields)
```
brendanshean, line 40: Is this method meant to be used outside of Transformer?

Otherwise, why not make a private instance method and remove the parameter passing.

services/QuillLMS/app/services/quill_big_query/runner.rb
```diff
@@ -0,0 +1,51 @@
...
+
+    def self.transform_response(json_body)
+      raise UnsupportedSchemaError, json_body unless valid_schema?(json_body)
+
+      # This parses to [{"id" => 1, "name"=> "Dan", }, {"id" => 2, "name" => "Peter"}]
+      raw_fields = json_body["schema"]["fields"]
+      fields = raw_fields.pluck("name")
+      values = json_body['rows'].pluck('f').map { |val_hash| val_hash.pluck('v') }
+      hash_results = values.map {|v| fields.zip(v).to_h}
+      QuillBigQuery::Transformer.new(raw_fields, hash_results).transform
```
brendanshean, line 25: Transformer seems like a good candidate for ApplicationService:

`QuillBigQuery::Transformer.run(raw_fields, hash_results)`

services/QuillLMS/app/services/quill_big_query/transformer.rb
brendanshean, line 1: I really like the `TYPE_TO_LAMBDA_LOOKUP` setup that you have here.

services/QuillLMS/app/services/quill_big_query/transformer.rb
```diff
@@ -0,0 +1,59 @@
...
+      fields.reduce({}) do |memo, field|
+        field_name = field['name']
+        field_type = field['type']
+        memo[field_name] = TYPE_TO_LAMBDA_LOOKUP[field_type]
```
dandrabik: What do we want to do if the type isn't present in our list? This will (eventually) raise an error here. 

Considering you found 2 undocumented types already, probably worth having an explicit:
- Default to STRING (no transform)
- Raise an explicit error.

Note that BigQuery's gem defaults to no transform. We might want to follow that.

services/QuillLMS/app/services/quill_big_query/transformer.rb
```diff
@@ -0,0 +1,59 @@
...
+        memo
+      end
+    end
+
+    def transform_pair(key, value)
```
dandrabik, line 49: How are null results handled?

services/QuillLMS/app/services/quill_big_query/transformer.rb
```diff
@@ -0,0 +1,59 @@
...
+
+    class UnsupportedBigQueryType < StandardError; end
+
+    def initialize(fields, array_of_hashes)
+      @field_to_lambda = self.class.build_field_to_lambda_lookup(fields)
+      @array_of_hashes = array_of_hashes
+    end
+
+    LAMBDA_IDENTITY     = ->(val) { val }
+    LAMBDA_TO_BOOLEAN   = ->(val) { val == 'true' }
```
dandrabik, line 14: This would convert a `nil` to `false` which is different.

services/QuillLMS/app/services/quill_big_query/transformer.rb
```diff
@@ -0,0 +1,59 @@
...
+      'BIGNUMERIC'  => LAMBDA_IDENTITY,     # '5.2e+37'
+      'BOOLEAN'     => LAMBDA_TO_BOOLEAN,   # 'true'
+      'BOOL'        => LAMBDA_TO_BOOLEAN,
+      'BYTES'       => LAMBDA_IDENTITY,
+      'DATE'        => LAMBDA_TO_DATETIME,  # '2022-01-01'
+      'DATETIME'    => LAMBDA_TO_DATETIME,  # '2022-05-06T17:08:44.915791'
+      'FLOAT'       => LAMBDA_TO_F,
+      'FLOAT64'     => LAMBDA_TO_F,
+      'GEOGRAPHY'   => LAMBDA_IDENTITY,
+      'INTEGER'     => LAMBDA_TO_F,
```
dandrabik: Is there a reason Integers are transformed `float`, rather than an `integer`?

services/QuillLMS/spec/services/quill_big_query/transformer_spec.rb
```diff
@@ -0,0 +1,82 @@
...
+
+      expected = {
+        'school_id' => QuillBigQuery::Transformer::LAMBDA_TO_F,
+        'color' => QuillBigQuery::Transformer::LAMBDA_IDENTITY
+      }
+      expect(lookup).to include(expected)
+    end
+  end
+
+  describe '#transform_pair' do
```
dandrabik, line 47: I think we need to expand the tests here, since I think there are a lot of things that would fail.

It would be good to add tests for transform_pair hitting a lot more things. I think there are some issues here.

These should be tested:
- Null results
- An example of every data type with an example of how that will be returned from BigQuery

Note, you might consider a small refactor to this method to make this testing easier, e.g. having a method in which you can send `type` and `value` (single line test) vs. `field` and `value` which requires a little more stubbing.

services/QuillLMS/app/services/quill_big_query/transformer.rb
```diff
@@ -0,0 +1,59 @@
...
+
+    def initialize(fields, array_of_hashes)
+      @field_to_lambda = self.class.build_field_to_lambda_lookup(fields)
+      @array_of_hashes = array_of_hashes
+    end
+
+    LAMBDA_IDENTITY     = ->(val) { val }
+    LAMBDA_TO_BOOLEAN   = ->(val) { val == 'true' }
+    LAMBDA_TO_DATETIME  = ->(val) { DateTime.parse(val) }
+    LAMBDA_TO_F         = ->(val) { val.to_f }
```
dandrabik, line 16: This would convert `nil` to `0.0`

services/QuillLMS/spec/services/quill_big_query/transformer_spec.rb
```diff
@@ -0,0 +1,82 @@
...
+      ]
+      input = [{ "a" => "1.0", "b" => "2", "c" => "3"}]
+      expected_output = [{ "a" => 1.0, "b" => 2, "c" => "3"}]
+      expect(QuillBigQuery::Transformer.new(fields, input).transform).to eq expected_output
+    end
+  end
+
+  describe 'datetime proc' do
+    # Many ISO 8601 date formats exist. Bigquery currently uses this one.
+    let(:example_iso_8601_date_string) { "2022-05-06T17:08:44.915791" }
```
dandrabik: Based on the [BigQuery gem](https://github.com/googleapis/google-cloud-ruby/blob/main/google-cloud-bigquery/lib/google/cloud/bigquery/convert.rb#L64-L110) I think we want to convert some of the date/time values differently

dandrabik (changes requested):
This is getting much tighter. I have a few more requested changes.

1. We should test the different types in the transform
2. Needs fallback for a TYPE not the list (in case there are new types) ✅ 
3. Needs testing for nulls
4. I believe integers should be converted to integers ✅ 
5. I believe that some of the date/time types need to be treated differently

Happy to chat through this if my comments don't make sense. I think most of these asks are small tweaks and tightening this up.

services/QuillLMS/app/services/quill_big_query/transformer.rb
happythenewsad: Good point; fixed.

services/QuillLMS/app/services/quill_big_query/transformer.rb
happythenewsad: I consider the complexity of the two options equal, and thus its a stylistic choice:

| current | proposed |
|--------|--------|
| pure function  | impure |
| public | private |
| needs param | does not|
| | needs initialization, for `fields` |

Happy to discuss further if you feel strongly about it, however.

services/QuillLMS/spec/services/quill_big_query/transformer_spec.rb
happythenewsad: FYI: I wrote a compact, metaprogrammy test runner as hinted above, but didn't like it's looks just wrote out the tests the standard way.

services/QuillLMS/app/services/quill_big_query/transformer.rb
happythenewsad: (discussed in zoom)

services/QuillLMS/app/services/quill_big_query/runner.rb
happythenewsad: Hmm, it would certainly work here. I tend to prefer PORO's over the ApplicationService pattern, but this seems like a good thing to discuss in the next engineering meeting, since it would behoove us all to align on the criteria for deciding on when to use this pattern.

happythenewsad: I went with your suggestion to use ApplicationService for the client in this PR, since that felt like more of a "service" to me. But I realize my definition of "service" is not tightly defined.

services/QuillLMS/app/services/quill_big_query/transformer.rb
```diff
@@ -0,0 +1,61 @@
...
+      fields.reduce({}) do |memo, field|
+        field_name = field['name']
+        field_type = field['type']
+        memo[field_name] = TYPE_TO_LAMBDA_LOOKUP.fetch(field_type, LAMBDA_IDENTITY)
+        memo
+      end
+    end
+
+    def transform_pair(key, value)
+      return nil unless value
```
happythenewsad, line 50: Adding the early return here, so that we don't have to test nil inputs in all of the lambdas.

dandrabik (approved):
This looks good to me, thanks for all of your work on this and for making the requested changes. This feels much tighter now.

🚢

PR #10504, opened 2023-04-21
lms/add-teacher-notifications

anathomical:
## WHAT
Add a model to store teacher notification records in
## WHY
We want to establish an abstraction to store messages in that can be used to enqueue messages to teachers and record their various delivery statuses.  This doesn't really do anything yet, but is foundational for further additions.
## HOW
Add a new table and model to store notification data in.  Add some strict validation on the `params` JSONB value so that it has a known and consistent shape.

### Notion Card Links

PR Checklist | Your Answer
------------ | -------------
Have you deployed to Staging? | No because this is foundational, and doesn't contain actual new functionality


services/QuillLMS/app/models/teacher_notification.rb
```diff
@@ -0,0 +1,68 @@
...
+    ]
+  }
+
+  belongs_to :user
+
+  validates :notification_type, presence: true, inclusion: {in: NOTIFICATION_TYPES}
+  validate :params_must_be_hash, :params_payload_has_required_keys
+
+  private def params_is_hash?
+    params.is_a?(Hash)
```
brendanshean: Since `params` has special meaning in rails in the context of the controller, to avoid confusion later, I would name this something more descriptive (e.g. custom_attrs).

services/QuillLMS/spec/factories/teacher_notification.rb
```diff
@@ -0,0 +1,14 @@
+# frozen_string_literal: true
+
+FactoryBot.define do
+  factory :teacher_notification, class: TeacherNotification do
```
brendanshean: I think the  `class: TeacherNotification` is unnecessary.

services/QuillLMS/spec/factories/teacher_notification.rb
```diff
@@ -0,0 +1,14 @@
...
+    association :user
```
brendanshean: I think you can just use `user` here.

brendanshean (approved):

services/QuillLMS/spec/models/teacher_notification_spec.rb
```diff
@@ -0,0 +1,69 @@
...
+      {
+        student_name: 'Student Name',
+        classroom_name: 'Classroom Name'
+      }
+    }
+
+    it 'should require params to be a hash' do
+      teacher_notification.params = ''
+
+      expect(teacher_notification.valid?).to be(false)
```
brendanshean: All four of the `.valid?` can be written using the `be_valid` matcher, for example:

expect(teacher_notification).to be_valid

services/QuillLMS/app/models/teacher_notification.rb
```diff
@@ -0,0 +1,68 @@
...
+  private def params_must_be_hash
+    return if params_is_hash?
+
+    errors.add(:params, "must be a hash")
+  end
+
+  private def params_payload_has_required_keys
+    return unless params_is_hash?
+
+    return if params.keys.map(&:to_s).to_set == NOTIFICATION_TYPE_VALIDATION[notification_type].map(&:to_s).to_set
```
happythenewsad: Nit: set inclusion would be a bit more flexible than set equality here. We don't care if the payload has more stuff than necessary, but we do care if the payload is lacking a key.

happythenewsad (approved):
Good stuff

services/QuillLMS/app/models/teacher_notification.rb
anathomical: @happythenewsad My thought on this was that we'll know exactly what keys we do want, and that additional keys will only be noise.  By doing equality instead of inclusion, my hope was to prevent bloat to these JSON blobs.  Having extra keys probably won't hurt anything other than potentially taking up space in the database, but what are your thoughts on preferring to allow extra keys instead of restricting them?

services/QuillLMS/app/models/teacher_notification.rb
```diff
@@ -0,0 +1,68 @@
...
+#  fk_rails_...  (user_id => users.id)
+#
+class TeacherNotification < ApplicationRecord
+  NOTIFICATION_TYPES = [
+    STUDENT_COMPLETED_DIAGNOSTIC = 'student-completed-diganostic',
+    STUDENT_COMPLETED_ALL_DIAGNOSTIC_RECOMMENDATIONS = 'student-completed-all-diagnostic-recommendations',
+    STUDENT_COMPLETED_ALL_ASSIGNED_ACTIVITIES = 'student-completed-all-assigned-activities'
+  ]
+
+  NOTIFICATION_TYPE_VALIDATION = {
```
dandrabik: So I'm not opposed to this setup, but some Rails-y things to consider using instead. 

Consider using [store](https://api.rubyonrails.org/classes/ActiveRecord/Store.html). It's a way to store the json, but treat them as proper attributes. You could then use standard Rails validations, e.g. `validates :student_name, presence: true` would work. You can make validations conditional with `if:` clauses if needed. Also code such as `notification.student_name = "Thomas"` would work as expected.

services/QuillLMS/app/models/teacher_notification.rb
happythenewsad: This reasoning makes sense to me - carry on.

I was thinking that it would be annoying to have specs fail when adding a new expected key to the payload, but that's more of a feature than a bug, given that we've suffered from heterogenous jsonb schemata in the past.

services/QuillLMS/db/migrate/20230420141952_create_teacher_notifications.rb
```diff
@@ -0,0 +1,14 @@
+# frozen_string_literal: true
+
+class CreateTeacherNotifications < ActiveRecord::Migration[6.1]
+  def change
+    create_table :teacher_notifications do |t|
+      t.references :user, index: true, foreign_key: true, null: false
```
dandrabik, line 6: FK on users, we should deploy this during low traffic.

services/QuillLMS/app/models/teacher_notification.rb
```diff
@@ -0,0 +1,68 @@
...
+# Indexes
+#
+#  index_teacher_notifications_on_user_id  (user_id)
+#
+# Foreign Keys
+#
+#  fk_rails_...  (user_id => users.id)
+#
+class TeacherNotification < ApplicationRecord
+  NOTIFICATION_TYPES = [
```
dandrabik: Note, if there is different logic for each of these type, we could consider [STI](https://medium.com/@GeneHFang/single-table-inheritance-in-rails-6-emulating-oop-principles-in-relational-databases-be60c84e0126) here. This is something we could also migrate to later if it's helpful.

dandrabik (approved):
Thanks for putting this together. I've left some comments for your consideration, but overall this is approved. I do think you should consider using ActiveRecord store to make validations/usage simpler, but I'll leave that up to you.

services/QuillLMS/app/models/teacher_notification.rb
```diff
@@ -0,0 +1,68 @@
...
+    STUDENT_COMPLETED_DIAGNOSTIC = 'student-completed-diganostic',
```
emilia-friedberg: Typo here - "diganostic".

emilia-friedberg (approved):
Typo but otherwise LGTM!

services/QuillLMS/app/models/teacher_notification.rb
anathomical: I ended up doing a refactor based on @dandrabik 's feedback, which has allowed us to avoid a custom validator of this type.  Does this address your concerns here?

services/QuillLMS/app/models/teacher_notification.rb
anathomical: I was originally struggling to figure out how to use `store` in conjunction with the fact that different notification types had different shapes, but it turns out that the solution was just to use the STI pattern.  So I've refactored the models in a way that I think takes advantage of both suggestions.  Let me know if this looks like a good solution.

happythenewsad (approved):

dandrabik (approved):
Looks good to me!

services/QuillLMS/app/models/teacher_notification/student_completed_all_diagnostic_recommendations.rb
```diff
@@ -0,0 +1,28 @@
...
+#
+# Indexes
+#
+#  index_teacher_notifications_on_user_id  (user_id)
+#
+# Foreign Keys
+#
+#  fk_rails_...  (user_id => users.id)
+#
+class TeacherNotification::StudentCompletedAllDiagnosticRecommendations < TeacherNotification
```
brendanshean: I just noticed there was this update with namespacing.  For some technical [reasons](https://rubystyle.guide/#namespace-definition), these should use the module definition over the shortened version

```
module TeacherNotification
  class StudentCompletedDiagnostic < ::TeacherNotification
    ..
   end
end
```

Make sure to do this for all subclasses.

PR #10505, opened 2023-04-21
update nces import to use nces id instead of name

emilia-friedberg:
## WHAT
Update the NCES import script to find existing districts based on NCES id and not name.

## WHY
We were inaccurately combining different district records into one in our system.

## HOW
Just update the `where` clause here.

### Screenshots

### Notion Card Links
https://www.notion.so/quill/NCES-District-Import-Fix-noting-for-future-imports-not-urgent-07d4a61c8cf64d81ad43352daa447216

PR Checklist | Your Answer
------------ | -------------

### DIFF
--- a/services/QuillLMS/lib/tasks/nces_sync.rake
+++ b/services/QuillLMS/lib/tasks/nces_sync.rake
@@ -36,7 +36,7 @@ namespace :nces_sync do
           grade_range: "#{row['Lowest Grade Offered']} - #{row['Highest Grade Offered']}"
         }
 
-        district = District.where("lower(name) = ?", attributes_hash[:name].downcase).first
+        district = District.where("nces_id = ?", attributes_hash[:nces_id]).first
         if district.present?
           district.update!(attributes_hash)
         else


services/QuillLMS/lib/tasks/nces_sync.rake
```diff
@@ -36,7 +36,7 @@ namespace :nces_sync do
           grade_range: "#{row['Lowest Grade Offered']} - #{row['Highest Grade Offered']}"
         }
 
-        district = District.where("lower(name) = ?", attributes_hash[:name].downcase).first
+        district = District.where("nces_id = ?", attributes_hash[:nces_id]).first
```
anathomical, line 39: Probably worth getting Cissy to confirm this, but my assumption is that the reason we even went for `name` in the first place was because some records in the NCES data file didn't have NCES ID values.  I wonder if what we really want here is "match on nces_id if present, otherwise match on name"?

emilia-friedberg: Yeah, good question. I'll ping Cissy since this isn't urgent.

cissyxyu: Hey y'all -- the reason we were matching on Name and not NCES ID originally is because our district database at the time was very simple. It only had a name field, not an NCES ID field. So we had no choice but to match on name. IIRC the NCES database is pretty complete for districts -- I don't recall any district not having an NCES ID. So I think it would be safe to match purely on NCES ID.

emilia-friedberg: Thank you!

anathomical (approved):
With @cissyxyu 's clarification, this looks good to me!

cissyxyu (approved):

PR #10510, opened 2023-04-25
start converting react routes

emilia-friedberg:
## WHAT
Convert three router files to the new v6 syntax, along with their components where necessary.

## WHY
We are going to slowly migrate each of our individual routers to use the new v6 syntax, in order to ultimately upgrade completely.

## HOW
Following the steps in this doc: https://github.com/remix-run/react-router/discussions/8753. Note that I started converting the teacher fix components to functional components, because they need to be functional to utilize the new routing hooks, but then I realized that they weren't actually using any functionality that required replacing at this time so I elected not to complete the conversion for all of them since it isn't actually necessary for this upgrade and adds a lot of additional work to an already large project (which is to say, I think we should only convert where necessary moving forward).

### Screenshots

### Notion Card Links
https://www.notion.so/quill/Upgrade-to-React-Router-v6-59c7bfc49d0b4a0484c9755a6960b7ea?pvs=4

PR Checklist | Your Answer
------------ | -------------


eadams17 (approved):

brendanshean (approved):
I appreciate the approach of doing this in stages.

PR #10519, opened 2023-04-26
Lms/add teacher notification settings

anathomical:
## WHAT
- Add the backend model to store settings for which Notifications a teacher wants to record for themselves
- Assign certain Notification Types defined by Product to new users by default moving forward
- Provide controller actions to allow a user to update their Notification settings

NOTE: This is targeted to merge into `lms/add-teacher-notifications` (#10504) since that PR is still review and this PR requires the changes from that one.  The plan would be to merge that PR when it's ready and then change the target of this PR to `develop` before merging.
## WHY
We want to provide teachers with granular control over which types of events they want to be notified about via email, but we also want to be able to set sensible defaults for new users rather than requiring them to discover this setting to begin receiving notification value.
## HOW
- Create a new `TeacherNotificationSetting` model that validates its `notification_type` from the list of values used for `TeacherNotification` records.
- Add an `after_create` hook for User records where `teacher?` is true to set some `notification_type` values to true for new users
- Create a controller with actions to list all valid `notification_type` values along with their current setting for the `current_user` and an action to update any number of `notification_type` settings for the `current_user`

### Notion Card Links

PR Checklist | Your Answer
------------ | -------------
Have you deployed to Staging? | No.  This is a building-block PR, and doesn't provide direct user functionality yet.


emilia-friedberg (approved):

services/QuillLMS/app/models/user.rb
```diff
@@ -860,5 +862,11 @@ def school_or_district_premium?
   private def update_invitee_email_address
     Invitation.where(invitee_email: email_before_last_save).update_all(invitee_email: email)
   end
+
+  private def generate_default_teacher_notification_settings
+    TeacherNotificationSetting::DEFAULT_FOR_NEW_USERS.each do |notification_type|
+      TeacherNotificationSetting.create!(user: self, notification_type: notification_type)
```
brendanshean: Since there's a has_many relation defined for teacher_notification_settings, you declare the creation via that, and omit the `user: self` param:
```
  TeacherNotificationSetting::DEFAULT_FOR_NEW_USERS.each do |notification_type|
    teacher_notification_settings.create(notification_type: notification_type)
  end
```

services/QuillLMS/app/models/teacher_notification_setting.rb
```diff
@@ -0,0 +1,35 @@
...
+    TeacherNotification::StudentCompletedAllDiagnosticRecommendations,
+    TeacherNotification::StudentCompletedAllAssignedActivities
+  ].map(&:name).freeze
+
+  def self.notification_types
+    TeacherNotification.subclasses.map(&:name)
+  end
+
+  belongs_to :user
+  validates :notification_type, presence: true, inclusion: {in: ->(model) { model.class.notification_types }, message: "%<value>s is not a valid TeacherNotification type"}
```
brendanshean: I tend to prefer multi-line if there's more than a couple args:

```
validates :notification_type, 
  presence: true, 
  inclusion: {in: ->(model) { model.class.notification_types }, 
  message: "%<value>s is not a valid TeacherNotification type"}
```

services/QuillLMS/app/models/teacher_notification_setting.rb
```diff
@@ -0,0 +1,35 @@
...
+#
+class TeacherNotificationSetting < ApplicationRecord
+  DEFAULT_FOR_NEW_USERS = [
+    TeacherNotification::StudentCompletedDiagnostic,
+    TeacherNotification::StudentCompletedAllDiagnosticRecommendations,
+    TeacherNotification::StudentCompletedAllAssignedActivities
+  ].map(&:name).freeze
+
+  def self.notification_types
+    TeacherNotification.subclasses.map(&:name)
```
brendanshean, line 30: Because of how Rails lazy loads classes, I don't think this will [always](https://guides.rubyonrails.org/autoloading_and_reloading_constants.html#single-table-inheritance) work.

As a stopgap until we update to zeitwerk, I've been [hard-coding](https://github.com/empirical-org/Empirical-Core/blob/develop/services/QuillLMS/app/models/provider_classroom_user.rb#L23) the subclass names into the base class

services/QuillLMS/app/controllers/api/v1/teacher_notification_settings_controller.rb
```diff
@@ -0,0 +1,43 @@
...
+      end
+    end
+
+    render json: types_and_values
+  rescue ActiveRecord::RecordNotDestroyed, ActiveRecord::RecordInvalid => e
+    render json: { errors: e }, status: 400
+  end
+
+  private def types_and_values
+    TeacherNotificationSetting.notification_types.map do |notification_type|
```
brendanshean: `to_h` can take a block:

```
TeacherNotificationSetting.notification_types.to_h do |notification_type|
  [notification_type, @current_settings.exists?(notification_type: notification_type)]
end
```

services/QuillLMS/app/controllers/api/v1/teacher_notification_settings_controller.rb
```diff
@@ -0,0 +1,43 @@
...
+  before_action :set_current_settings, only: [:index, :bulk_update]
+
+  def index
+    render json: types_and_values
+  end
+
+  def bulk_update
+    TeacherNotificationSetting.transaction do
+      teacher_notification_setting_params.each do |notification_type, active|
+        # Convert the string to a boolean so that 'false' is treated as false
```
brendanshean, line 13: Since the guard clauses have overlap, I feel like a traditional if/else block is more readable:

```
# Convert the string to a boolean so that 'false' is treated as false
if ActiveModel::Type::Boolean.new.cast(active)
  next if @current_settings.exists?(notification_type: notification_type)

  TeacherNotificationSetting.create!(user: current_user, notification_type: notification_type)
else
  # Destroy settings that are not in the payload params
  @current_settings.find_by(notification_type: notification_type)&.destroy!
end
```

services/QuillLMS/spec/factories/teacher_notification_setting.rb
```diff
@@ -0,0 +1,8 @@
+# frozen_string_literal: true
+
+FactoryBot.define do
+  factory :teacher_notification_setting, class: TeacherNotificationSetting do
```
brendanshean: `class: TeacherNotificationSetting` should not be needed as factory bot can infer from filename.

brendanshean (commented):
I have one question about `.subclasses`, otherwise I think this looks good to go.

services/QuillLMS/spec/rails_helper.rb
```diff
@@ -12,6 +12,9 @@
 require 'factory_bot_rails'
 require 'spec_helper'
 
+# This should eager load the TeacherNotification sub-classes which is necessary in our testing environment because we use the TeacherNotification.subclasses call for validation, and without eager loading that value starts as an empty array
+Dir[File.join(__dir__, '..', 'app', 'models', 'teacher_notifications', '*.rb')].sort.each { |file| require file }
```
brendanshean, line 16: I really like this approach to solving the lazy loading.

services/QuillLMS/spec/controllers/api/v1/teacher_notification_settings_controller_spec.rb
```diff
@@ -0,0 +1,87 @@
+# frozen_string_literal: true
+
+require 'rails_helper'
+
+describe Api::V1::TeacherNotificationSettingsController do
+  let!(:teacher) { create(:teacher) }
+  let(:notification_type) { TeacherNotifications::StudentCompletedDiagnostic }
```
brendanshean, line 7: If you don't want to couple to a specific type, you could [sample](https://github.com/empirical-org/Empirical-Core/blob/develop/services/QuillLMS/spec/models/provider_classroom_user_spec.rb#L22) the subclasses

services/QuillLMS/spec/controllers/api/v1/teacher_notification_settings_controller_spec.rb
```diff
@@ -0,0 +1,87 @@
...
+    end
+
+    it 'should set values to false for notification_types without saved settings' do
+      get :index
+
+      expect(JSON.parse(response.body)[notification_type_name]).to be(false)
+    end
+  end
+
+  describe '#bulk_update' do
```
brendanshean, line 39: I really appreciate the thorough testing here.

brendanshean (approved):
Great work on this!  I think you've setup a nice blueprint for STI in the future.

PR #10525, opened 2023-04-28
Lms/add teacher notification email frequency

anathomical:
## WHAT
- Add a column to `TeacherInfo` to store desired email notification frequency
- Add `after_create` hook on `User` to set a default value for this column on newly-created users
- Tweak the `TeacherInfosController` so that calls to the `create` action will use `find_or_create` so that when we create `TeacherInfo` records early in our `after_create` hook we don't break later calls to the `create` action by running into uniqueness constraints
## WHY
We want users to be able to specify how frequently the receive rollup emails to let them strike their own balance between number of emails and freshness of data.
## HOW
- Add a column to the TeacherInfo table
- Add a `User` `after_create` hook to set default values
- Update `TeacherInfosController` to accept the new value so that we can use it to update this value moving forward

### Notion Card Links

PR Checklist | Your Answer
------------ | -------------


services/QuillLMS/app/controllers/teacher_infos_controller.rb
```diff
@@ -2,42 +2,53 @@
...
-      maximum_grade_level: teacher_info.maximum_grade_level,
-      subject_area_ids: teacher_info.subject_area_ids
+      minimum_grade_level: @teacher_info.minimum_grade_level,
+      maximum_grade_level: @teacher_info.maximum_grade_level,
+      subject_area_ids: @teacher_info.subject_area_ids,
+      notification_email_frequency: @teacher_info.notification_email_frequency
     }, status: 200
   end
 
+  private def set_teacher_info
```
brendanshean, line 50: The logic here is a little surprising given the conventions for `create` and `update` in Rails controllers.

Unless there's a good reason not to, I think create should be reserved the case where `teacher_info` is created only.  Then use update when the object already exists;.

services/QuillLMS/app/controllers/teacher_infos_controller.rb
```diff
@@ -2,42 +2,53 @@
...
     if minimum_grade_level && maximum_grade_level
-      teacher_info.update!(minimum_grade_level: minimum_grade_level, maximum_grade_level: maximum_grade_level)
+      @teacher_info.update!(minimum_grade_level: minimum_grade_level, maximum_grade_level: maximum_grade_level)
     end
 
     if subject_area_ids
-      teacher_info.teacher_info_subject_areas.destroy_all
-      subject_area_ids.each { |subject_area_id| TeacherInfoSubjectArea.create!(teacher_info: teacher_info, subject_area_id: subject_area_id) }
+      @teacher_info.teacher_info_subject_areas.destroy_all
+      subject_area_ids.each { |subject_area_id| TeacherInfoSubjectArea.create!(teacher_info: @teacher_info, subject_area_id: subject_area_id) }
```
brendanshean: I like using the create directly from the base model:

`subject_area_ids.each { |subject_area_id| @teacher_info.teacher_info_subject_areas(subject_area_id: subject_area_id) }`

This pattern highlights the has_many relationship but also makes the context of the creation the @teacher_info object which is consistent with the controller.

services/QuillLMS/app/models/user.rb
```diff
@@ -860,5 +861,9 @@ def school_or_district_premium?
   private def update_invitee_email_address
     Invitation.where(invitee_email: email_before_last_save).update_all(invitee_email: email)
   end
+
+  private def generate_default_notification_email_frequency
+    TeacherInfo.create!(user: self, notification_email_frequency: TeacherInfo::DAILY_EMAIL)
```
brendanshean: I would use:
`teacher_info.create!(notification_email_frequency: TeacherInfo::DAILY_EMAIL)`

since it highlights the has_one relationship with teacher_info and has the implicit user context.

services/QuillLMS/spec/models/segment_integration/user_spec.rb
```diff
@@ -101,6 +101,8 @@
     let(:teacher_info) { create(:teacher_info, user: teacher) }
```
brendanshean: What about using the teacher_info from the teacher:

`let(:teacher_info) { teacher.teacher_info }`

then you could get rid of lines 104-105

brendanshean (approved):
There's a comment about create/update in the controller, but nothing that requires a re-review.  Nice work on this.

emilia-friedberg (approved):

PR #10530, opened 2023-05-02
Lms/teacher notification email rollups

anathomical:
## WHAT
Add workers and cron jobs to send rollup notifications, along with a mailer and views to actually assemble the rollup emails.

IMPORTANT NOTE: There is currently no code to generate `TeacherNotification` records.  That's my next (and final) step.  So deploying this code to production won't actually do anything at the moment.
## WHY
Email is a quick and dirty way of providing notification to teachers, so that's what we're starting with
## HOW
- Add a new worker to determine which teachers should get notification rollups at any given frequency
- Add calls to the worker in our cron model to trigger it at the three frequencies
- Add a worker that's called by the frequency worker that finds all new notifications for a teacher, sends them to a mailer, and then marks them with an `email_sent` value so they don't get sent again later
- Add the mailer to group `TeacherNotifications` into logical group by classroom and notification type, and then assemble the notifications into an HTML view to send to the teacher

### Notion Card Links

PR Checklist | Your Answer
------------ | -------------
Have you deployed to Staging? | Doing so now


services/QuillLMS/app/mailers/teacher_notifications/rollup_mailer.rb
```diff
@@ -0,0 +1,29 @@
...
+      TeacherInfo::WEEKLY_EMAIL => 'week'
+    }
+
+    default from: "The Quill Team <hello@quill.org>"
+
+    def rollup(user, teacher_notifications)
+      @user = user
+      @frequency = FREQUENCY_WORD_LOOKUP[user.teacher_info.notification_email_frequency]
+
+      @teacher_notifications = teacher_notifications.sort_by(&:student_name)
```
brendanshean, line 28: I'm curious why you went with Ruby methods instead of active record.  Potentially too much load on the db?

services/QuillLMS/app/views/teacher_notifications/rollup_mailer/_students_completed_all_assigned_activities.html.erb
```diff
@@ -0,0 +1,8 @@
+<h2>Completed all assigned activities</h2>
+<p><i>Assign your students more activities <a href="https://www.quill.org/assign">here</a>.</i></p>
+<ul>
+  <% notifications[..9].each do |notification| %>
```
brendanshean: `notifications.first(10).each` might be more readable.

services/QuillLMS/app/views/teacher_notifications/rollup_mailer/_truncate_student_names.html.erb
```diff
@@ -0,0 +1,3 @@
+<% if notifications.length > 10 %>
+  <li>...and <%= notifications.length - 10 %> other student<%= notifications.length == 11 ? '' : 's' %>.</li>
```
brendanshean: There's a [pluralize](https://api.rubyonrails.org/classes/ActionView/Helpers/TextHelper.html#method-i-pluralize) method that I think could simplify this:
```
<% num_other_students = notifications.length - 10 %>
<% if num_other_students > 0 %>
  <li>...and <%= pluralize(num_other_students, 'other student') %>.</li>
```

services/QuillLMS/app/views/teacher_notifications/rollup_mailer/rollup.html.erb
```diff
@@ -0,0 +1,17 @@
+<p>Hi <%= @user.first_name %>,</p>
+
+<p>Here’s what happened in the last <%= @frequency %> on Quill:</p>
+
+<% @teacher_notifications.each do |classroom_name, notifications_by_type| %>
+  <h1><%= classroom_name %></h1>
+  <% notifications_by_type.each do |notification_type, notifications| %>
+    <% if notification_type == "TeacherNotifications::StudentCompletedDiagnostic" %>
```
brendanshean: Since there's likely to be more of these in the future, I'd recommend using case statement for the partial:
```
partial = case notification_type
when "TeacherNotifications::StudentCompletedDiagnostic"
  :student_completed_diagnostic
when "TeacherNotifications::StudentCompletedAllDiagnosticRecommendations 
  :student_completed_all_diagnostic_recommendations
...
end

render partial, notifications: notifications
...

Comment (brendanshean) [reply services/QuillLMS/app/views/teacher_notifications/rollup_mailer/rollup.html.erb]: Instead of the case statement you could also do:
```
partial = notification_type.constantize.name.demodulize.underscore
render partial, notification: notification
```

File: services/QuillLMS/spec/factories/teacher_notification.rb [hunk 5]
```diff
@@ -0,0 +1,43 @@
...
+#  index_teacher_notifications_on_user_id  (user_id)
+#
+# Foreign Keys
+#
+#  fk_rails_...  (user_id => users.id)
+#
+FactoryBot.define do
+  factory :teacher_notification do
+    user
+
```
brendanshean, line 26: If you think `student_name` and `classroom_name` will be apart of all teacher notifications, I would just pull them up into the base factory.

brendanshean (approved):
Great work on pulling all this together.  I also really appreciate the thorough testing.

services/QuillLMS/app/views/teacher_notifications/rollup_mailer/_students_completed_all_assigned_activities.html.erb
happythenewsad: Let's make `9` a constant rather than a magic number

services/QuillLMS/app/views/teacher_notifications/rollup_mailer/rollup.html.erb
happythenewsad: I agree with Brendan's comment about minimizing logic in views. But to go a step further: you could abstract the if/case statement into a data structure, so you could then do 


```
notifications_by_type.each do |notification_type, notifications|
  <%= render LOOKUP_STRUCT[notification_type] %>
end
```

happythenewsad (changes requested):
Couple minor things. Looking good though

services/QuillLMS/app/mailers/teacher_notifications/rollup_mailer.rb
anathomical: I don't know if this has a huge performance impact, but `student_name` is a value inside a JSONB blob rather than a column.  It might still be better to sort in Postgres rather than Ruby, but that was my thought here.

services/QuillLMS/spec/factories/teacher_notification.rb
anathomical: At this point, I just don't know.  I don't have a clear sense of what the next five or six notification types might be.  I could imagine doing some that only have classroom_name (like "your whole class has done X").  I do imagine that any time we have student_name we'll probably want to use classroom_name for disambiguation.

services/QuillLMS/app/mailers/teacher_notifications/rollup_mailer.rb
brendanshean: I forgot they were JSONB.  That makes sense.

services/QuillLMS/app/views/teacher_notifications/rollup_mailer/_students_completed_all_assigned_activities.html.erb
anathomical: Both good suggestions.

services/QuillLMS/app/views/teacher_notifications/rollup_mailer/rollup.html.erb
anathomical: I like @happythenewsad 's suggestion of a lookup table, so I've gone with that approach.

happythenewsad (approved):

services/QuillLMS/app/views/teacher_notifications/rollup_mailer/rollup.html.erb
brendanshean: I think @happythenewsad 's suggestion was the best one.

I'm realizing that while `notification_type.constantize.name.demodulize.underscore` is shorter it makes the code harder to follow when things go wrong.

PR #10533, opened 2023-05-03
Add more analytics to Vitally District payload 

cissyxyu:
## WHAT
Add more analytics data -- total activities completed, total active students, activities completed per student, and last active time -- to the payload we send to Vitally for Districts. 

**Benchmark runtimes for queries:** 

**(for an RSpec test batch of 10,000 activities completed for a district with 80 schools):**
Average runtime for active students all time: 0.065 seconds
Average runtime for active students this year: 0.063 seconds
Average runtime for active students last year: 0.023 seconds
Average runtime for activities completed all time: 0.050 seconds
Average runtime for activities completed this year: 0.053 seconds
Average runtime for activities completed last year: 0.029 seconds
Average runtime for last sign in: 0.041 seconds
Average runtime to get entire data payload (all traits): 0.393 seconds

**(for Staging, 12,152 active districts):**
Average runtime to get entire data payload (all traits): 0.28 seconds

## WHY
To help Partnerships make more informed decisions.

## HOW
Firstly, I moved all the Vitally data calculations into a new class that handles Vitally data for districts. 
Then I added the `active_students`, `activities_completed`, `last_active_time`, and `activities_completed_all_time` methods to query and filter these stats and aggregate them for the vitally payload.

### Screenshots

### Notion Card Links
https://www.notion.so/quill/Product-Board-30f97e4eb01246dbb8264253fb385073?p=784b8ee0bd804c59a415c8ea11a8419f&pm=c

PR Checklist | Your Answer
------------ | -------------


services/QuillLMS/app/services/vitally_integration/serialize_vitally_sales_organization.rb
```diff
@@ -0,0 +1,194 @@
...
+        }
+      }
+    }).where(classification: {key: ActivityClassification::DIAGNOSTIC_KEY})
+      .where(activity_sessions: {completed_at: start..stop})
+      .distinct
+      .count
+  end
+
+  def active_students(start_date=nil, end_date=nil)
+    return activities_completed_all_time.group("students.id").length if start_date.blank? && end_date.blank?
```
cissyxyu, line 120: this is a newly added method.

services/QuillLMS/app/services/vitally_integration/serialize_vitally_sales_organization.rb
```diff
@@ -0,0 +1,194 @@
...
+
+    if end_date.present?
+      filtered_results = activities_completed_all_time.where("activity_sessions.completed_at <= ?", end_date).group("students.id")
+    end
+
+    filtered_results.length
+  end
+
+  def activities_completed(start_date=nil, end_date=nil)
+    return activities_completed_all_time.count if start_date.blank? && end_date.blank?
```
cissyxyu, line 134: this is a newly added method.

services/QuillLMS/app/services/vitally_integration/serialize_vitally_sales_organization.rb
```diff
@@ -0,0 +1,194 @@
...
+
+    if end_date.present?
+      filtered_results = activities_completed_all_time.where("activity_sessions.completed_at <= ?", end_date)
+    end
+
+    filtered_results.count
+  end
+
+  def last_active_time
+    User.select("students.last_sign_in")
```
cissyxyu, line 148: this is a new method.

services/QuillLMS/app/services/vitally_integration/serialize_vitally_sales_organization.rb
```diff
@@ -0,0 +1,194 @@
...
+      .where('districts.id = ?', @district.id).references(:district)
+      .includes(classrooms_teachers: [classroom: :students_classrooms])
+      .joins("JOIN users students ON students.id = students_classrooms.student_id")
+      .order("students.last_sign_in DESC")
+      .first
+      &.last_sign_in
+  end
+
+  def activities_completed_all_time
+    @activities_completed ||= ClassroomsTeacher.select("students.id")
```
cissyxyu, line 159: this is a new method.

services/QuillLMS/app/services/vitally_integration/serialize_vitally_sales_organization.rb
```diff
@@ -0,0 +1,194 @@
+# frozen_string_literal: true
```
happythenewsad, line 1: Refactoring this out of district.rb was a god idea.

services/QuillLMS/app/services/vitally_integration/serialize_vitally_sales_organization.rb
```diff
@@ -0,0 +1,194 @@
...
+
+class SerializeVitallySalesOrganization
```
happythenewsad, line 3: Note: This class's purpose is serialization, but it does not inherit from ApplicationSerializer, nor does it live in `serializers/`.  

I'm sort of OK leaving it as-is, but it's worth highlighting.

services/QuillLMS/app/services/vitally_integration/serialize_vitally_sales_organization.rb
happythenewsad: This one is less of an issue due to district-wise filtering

services/QuillLMS/app/services/vitally_integration/serialize_vitally_sales_organization.rb
happythenewsad: Would you mind benchmarking these new queries, to minimize the possibility of introducing performance issues to production?

You can use the pattern found in `plagiarism_check_spec.rb` to tag certain specs with `benchmarking`, so that CircleCI does not run them. 

Using print statements within the `benchmark` specs, you can get run times, which you can then attach to this PR's notes. 

Alternatively, you could benchmark the entire serializer, which would mean writing fewer specs. Up to you. 

I'm happy to pair on this if that's preferable - we can find an overlap time.

happythenewsad (commented):
Looks good, but I'd like to smoke test the query performance before approving.

services/QuillLMS/app/services/vitally_integration/serialize_vitally_sales_organization.rb
cissyxyu: I'm following the naming pattern we use for all the Vitally "serializers" (there's two others by that name in the folder, for schools and teachers). 

Good to note this -- I'll keep it for now, but maybe we can do a renaming next time there's additions to Vitally "serializers" here and rename all of these classes.

services/QuillLMS/app/services/vitally_integration/serialize_vitally_sales_organization.rb
anathomical: I think that `count` is preferable to `length` here.  ActiveRecord runs the `COUNT(*)` command in SQL, but I'm pretty sure that `length` pulls all items into Rails and the counts the length of the array which can be way less efficient.

services/QuillLMS/app/services/vitally_integration/serialize_vitally_sales_organization.rb
```diff
@@ -0,0 +1,194 @@
...
+
+    if start_date.present?
+      filtered_results = activities_completed_all_time.where("activity_sessions.completed_at >= ?", start_date).group("students.id")
+    end
+
+    if end_date.present?
+      filtered_results = activities_completed_all_time.where("activity_sessions.completed_at <= ?", end_date).group("students.id")
+    end
+
+    filtered_results.length
```
anathomical, line 130: Ditto the `count` vs `length` thing here.

services/QuillLMS/app/services/vitally_integration/serialize_vitally_sales_organization.rb
anathomical: I agree with the benchmarking suggestion.  This is probably something we should be doing with all of our Vitally stuff so that we can calculate how "expensive" any new values we add are to the overnight sync process.  So while this isn't something we've done historically, it's probably a good idea to do, and this seems like an okay time to start it.

anathomical (commented):
A minor flag about `length` vs `count`, but mostly I just agree with @happythenewsad

services/QuillLMS/app/services/vitally_integration/serialize_vitally_sales_organization.rb
cissyxyu: Benchmarking specs added! Runtime reports are in the PR. @happythenewsad

services/QuillLMS/app/services/vitally_integration/serialize_vitally_sales_organization.rb
cissyxyu: The reason I'm using `length` here is that this is a query using GROUP BY, so the result of the query if I use `.count` looks like { student1: # of activities, student2: # of activities, student3: # of activities ... ...}. 

The only way to get the actual number of students, and not just the count of activities each student completed, is by using the `.length` method.

happythenewsad (approved):
Thank you for adding the benchmarking, Cissy. 

Let's also log the run times in staging and prod, when you deploy. Thanks!

services/QuillLMS/spec/services/vitally_integration/serialize_vitally_sales_organization_spec.rb
```diff
@@ -0,0 +1,313 @@
...
+        diagnostic = create(:diagnostic_activity)
+        unit = create(:unit, activities: [diagnostic])
+        students = create_list(:student, 25, student_in_classroom: [classroom], last_sign_in: Time.zone.today)
+        classroom_unit = create(:classroom_unit, classroom: classroom, unit: unit, assigned_student_ids: students.map(&:id))
+        students.each do |student|
+          create_list(:activity_session, 5, state: 'finished', completed_at: Time.current, user: student, classroom_unit: classroom_unit, activity: diagnostic)
+        end
+      end
+
+      runtime = Benchmark.realtime { expect(SerializeVitallySalesOrganization.new(district).active_students).to eq(2000) }
```
happythenewsad, line 291: Note, for future PRs: You don't need to wrap the thing you are benchmarkinging in an expectation.

anathomical (approved):
Yay benchmarking!

PR #10536, opened 2023-05-03
Global mobile navigation updates

eadams17:
## WHAT
update our site headers and navigation functionality globally for mobile 

## WHY
we want to make our site more mobile-friendly and our header/navigation elements were in desperate need of an overhaul

## HOW
update the various rails views and React components that handle navigation rendering, add helper functions and partials for rendering 

just a note: the design slightly deviates from the Figma mockups, these differences were discussed with Jack and he has already given this approval

### Screenshots
Screen Shot 2023-05-03 at 4 37 32 PM
Screen Shot 2023-05-03 at 4 37 38 PM
Screen Shot 2023-05-03 at 4 38 00 PM
Screen Shot 2023-05-03 at 4 38 26 PM
Screen Shot 2023-05-03 at 4 38 46 PM
Screen Shot 2023-05-03 at 4 39 07 PM
Screen Shot 2023-05-03 at 4 39 29 PM
Screen Shot 2023-05-03 at 4 39 55 PM
Screen Shot 2023-05-03 at 4 41 45 PM

### Notion Card Links
https://www.notion.so/quill/New-Global-Header-Navigation-Premium-Tabs-for-Mobile-2a193c9e981c41918523872df450401b

PR Checklist | Your Answer
------------ | -------------


services/QuillLMS/app/helpers/navigation_helper.rb
```diff
@@ -115,4 +188,130 @@ def should_show_admin_access_tab?
   def admin_impersonating_user?(user)
     session[:admin_id].present? && session[:admin_id] != user.id
   end
+
+  # rubocop:disable Metrics/CyclomaticComplexity
```
eadams17: I added this disable block because I wasn't really sure if there was a better way to approach this portion of the code that didn't feel like over-engineering-- I'm open to suggestions though

anathomical: I think in some ideal world we probably want to rationalize our paths so that the sections are sort of named in the path and are thus easy and predictable to extract, but I think that's a project that's way bigger than the scope of this PR.

I don't know that this is necessarily better, but it at least avoids a bunch of `if/elsif` statements:
```ruby
ACTIVE_TAB_PATHS = {
  LEARNING_TOOLS => ['tools'],
  ABOUT_US => ['about', 'announcements', 'mission', 'impact', 'press', 'team', 'pathways', 'careers']
  ...
}

def determine_active_tab(current_path)
  ACTIVE_TAB_PATHS.each do |tab, paths|
    # return early once we find our match
    return tab if paths.any? { |path| current_path.include?(path) }
  end
  determine_dashboard_active_tab(current_path)
end
```

services/QuillLMS/app/views/application/_header.html.erb
```diff
@@ -1,16 +1,50 @@
...
+	function toggleMobileSecondaryDropdown(primaryDropdownOpen) {
+    const dropdown = document.querySelector('#mobile-subnav-dropdown');
+
+		if(!dropdown) { return }
+
+    const parent = dropdown.parentElement
+
+    if(primaryDropdownOpen) {
+      parent.classList.remove('open');
+      return;
```
anathomical, line 30: I think it should be safe to make this one line and just call `return parent.classList.remove('open');` since it doesn't look like it matters what value comes back from this function.  I could be wrong, though.

services/QuillLMS/spec/helpers/navigation_helper_spec.rb
```diff
@@ -196,4 +160,229 @@
       expect(helper.should_show_admin_access_tab?).to eq true
     end
   end
+
+  describe '#determine_mobile_navbar_tabs' do
```
anathomical, line 164: I love how extensive and thorough these tests are.

anathomical (approved):
One suggestion on possible tweaks to organization of code, but overall this code looks good.  And the update looks even better!

services/QuillLMS/app/helpers/navigation_helper.rb
```diff
@@ -1,11 +1,88 @@
...
 module NavigationHelper
+  routes = Rails.application.routes.url_helpers
+
+  # premium tiers
   PAID = 'paid'
   TRIAL = 'trial'
   LOCKED = 'locked'
   NONE = 'none'
 
+  # primary/secondary navigation
```
emilia-friedberg, line 12: Love all these constants!

services/QuillLMS/app/helpers/navigation_helper.rb
emilia-friedberg: Agreed with Thomas here - the easiest thing would be to turn the groupings into hashes and iterate. Totally up to you whether that feels worthwhile, though.

services/QuillLMS/spec/helpers/navigation_helper_spec.rb
emilia-friedberg: Agreed with Thomas. These tests do, however, make me feel like using a hash structure like the one Thomas was suggesting might be more maintainable - because then you could just iterate through that to test that things return as expected rather than manually writing out each path in both places.

emilia-friedberg (approved):
This looks so good! So excited for its launch. Some take-or-leave feedback on the controller constants, but overall I think it's ready to go. One thing that I think doesn't necessarily fit into this project, but might be worth considering in a future iteration, is whether we could generalize the subtab components so that we don't have to repeat classList logic in the various React files where it exists.

services/QuillLMS/app/helpers/navigation_helper.rb
eadams17: Thanks for the suggestion, @anathomical! I like this approach, I'll refactor to use this pattern

PR #10540, opened 2023-05-05
Lms/generate teacher notifications

anathomical:
## WHAT
Add code to actually generate `TeacherNotification` records so that they become available for email rollups.  Also add front-end access to the ability to change email rollup configuration so that teachers can control what they get and how often.

NOTE: I think there's a strong argument to be made to do the backend work and the front-end work in separate PRs, but launching one without the other doesn't make sense since we don't want to start sending emails without a way for teachers to control what they're getting, but we also don't want to imply that they can control message settings before they can actually get anything.  So if people want to review just the front-end or just the back-end of this PR, I think that could be helpful.
## WHY
We want teachers to be able to select specific student-related events to be notified of, and also to control how often they receive those notifications.  Once a teacher has their configuration picked (or if they leave the default settings in place), we also want them to actually receive roll-up emails of those notifications.
## HOW
- Add and update `Student` concern to have powerful utility methods: `assigned_activities` and `incomplete_assigned_activities` to help with later code to determine which notifications are relevant for a given student
- Hook up a Sidekiq worker to the `ActivitySession` completion code that determines which notifications a given activity completion might qualify for
- Have that worker write new `TeacherNotification` records for any relevant teachers
- Update the Teacher Account settings front-end to allow a teacher to change which student events they want to get notifications for, and how often they want to receive an email rollup of those notifications

### Screenshots
image

### Notion Card Links

PR Checklist | Your Answer
------------ | -------------


services/QuillLMS/app/controllers/api/v1/teacher_notification_settings_controller.rb
```diff
@@ -32,9 +32,11 @@ def bulk_update
   end
 
   private def types_and_values
-    TeacherNotificationSetting.notification_types.to_h do |notification_type|
-      [notification_type, @current_settings.exists?(notification_type: notification_type)]
-    end
+    {
```
brendanshean, line 35: Since current_settings is a TeacherNotificationSettings collection, you could extract it into a class method within `TeacherNotificationSetting` and store the `exists?` as a temp to limit the number of database calls.

app/models/teacher_notification_setting
```
def self.blah
  temp = pluck(:notification_type).uniq

  notification_types.to_h do |notification_type|
    [notification_type, notification_type.in?(temp)]
  end
end
```

Then this becomes

```
private def types_and_values
  { teacher_notification_settings: @current_settings.blah }
end
```

services/QuillLMS/app/models/user.rb
```diff
@@ -638,15 +638,13 @@ def generate_teacher_account_info
     user_attributes[:subscription]['subscriptionType'] = premium_state
     user_attributes[:minimum_grade_level] = teacher_info&.minimum_grade_level
     user_attributes[:maximum_grade_level] = teacher_info&.maximum_grade_level
+    user_attributes[:notification_email_frequency] = teacher_info&.notification_email_frequency
+    user_attributes[:teacher_notification_settings] = teacher_notification_settings_info
     user_attributes[:subject_area_ids] = subject_area_ids
 
-    if school && school.name
```
brendanshean, line 645: Thanks for cleaning this up!

services/QuillLMS/app/models/user.rb
```diff
@@ -873,5 +871,23 @@ def school_or_district_premium?
       teacher_notification_settings.create!(notification_type: notification_type)
     end
   end
+
+  private def teacher_notification_settings_info
+    TeacherNotificationSetting.notification_types.to_h do |notification_type|
```
brendanshean: You can reuse the method discussed in the comment on teacher_notification_settings_controller.rb

services/QuillLMS/app/models/concerns/student.rb
```diff
@@ -18,9 +18,28 @@ module Student
       class_name: "Classroom"
 
     has_many :activity_sessions, dependent: :destroy
-    has_many :assigned_activities, through: :classrooms, source: :activities
+    has_many :assigned_activities, -> { where("students_classrooms.student_id = ANY (classroom_units.assigned_student_ids)") }, through: :classrooms, source: :activities
```
brendanshean, line 21: Great job figuring out that the previous relation was incorrect.

services/QuillLMS/app/workers/teacher_notifications/send_notification_worker.rb
```diff
@@ -0,0 +1,82 @@
...
+        .includes(:user)
+        .includes(teachers: :teacher_notification_settings)
+        .includes(:classroom)
+        .includes(activity: :classification)
+        .includes(unit_template: :recommendations)
+        .first
+    end
+
+    private def send_complete_diagnostic
+      return unless @activity_session.activity.classification.key == ActivityClassification::DIAGNOSTIC_KEY
```
brendanshean: What about: `return unless @activity_session.is_diagnostic?`

services/QuillLMS/app/workers/teacher_notifications/send_notification_worker.rb
```diff
@@ -0,0 +1,82 @@
...
+      })
+    end
+
+    private def send_complete_all_diagnostic_recommendations
+      unit_template = @activity_session.unit_template
+      # Return early if the activity that was just completed wasn't a recommendation
+      return if unit_template.nil? || unit_template.recommendations.empty?
+
+      # Return early if the user has never completed a Diagnostic since any activities that they might be assigned can't be from recommendations in that case
+      return unless @activity_session.user.activity_sessions
```
brendanshean: What about moving lines 44-48 to user.rb

```
def completed_diagnostics?
...
end
```

This allows removal of the comment in line 44 and then just
`return unless @activity_session.user.completed_diagnostics?`

services/QuillLMS/app/workers/teacher_notifications/send_notification_worker.rb
```diff
@@ -0,0 +1,82 @@
...
+        .joins(activity: :classification)
+        .where(activity_classifications: {key: ActivityClassification::DIAGNOSTIC_KEY})
+        .count > 0
+
+      # Return early if the user has any incomplete assigned activities that are recommendations
+      # Note that if the student has recommendations from two different diagnostics, they have to finish all recommendations from BOTH or this will returrn early
+      return if @activity_session.user.incomplete_assigned_activities
```
brendanshean: Similarly, what about adding a new scope in user.rb:

def incomplete_recommended_activities?
...
end

services/QuillLMS/app/workers/teacher_notifications/send_notification_worker.rb
```diff
@@ -0,0 +1,82 @@
...
+        .joins(unit_templates: :recommendations).distinct.count > 0
+
+      send_notification(TeacherNotifications::StudentCompletedAllDiagnosticRecommendations, {
+        student_name: @activity_session.user.name,
+        classroom_name: @activity_session.classroom.name
+      })
+    end
+
+    private def send_complete_all_assigned_activities
+      return if @activity_session.user.incomplete_assigned_activities
```
brendanshean: Instead of .where.count, what about exists?

services/QuillLMS/app/workers/teacher_notifications/send_notification_worker.rb
brendanshean: Also, can the  .where..distinct.count > 0 could be rewritten as just `exists?`

brendanshean (approved):
I did a pass of the backend and it looked good to me.  Nice work on finishing up this last piece.

brendanshean (approved):
I did a pass of the backend and it looked good to me.  Nice work on finishing up this last piece.

services/QuillLMS/client/app/bundles/Teacher/components/accounts/edit/teacher_email_notifications.jsx
```diff
@@ -1,36 +1,172 @@
...
+
+    updateTeacherInfo({
+      notification_email_frequency: notificationEmailFrequency
+    })
   }
 
-  render() {
+  function renderNewsletterCheckbox() {
+    const selectedClass = (sendNewsletter) ? "selected" : "unselected"
+    const checkboxImg = (sendNewsletter) ? (<img alt="check" src={smallWhiteCheckSrc} />) : ""
```
eadams17: small nit, but I think we can do without the booleans or `img` element wrapped in parens

services/QuillLMS/client/app/bundles/Teacher/components/accounts/edit/teacher_email_notifications.jsx
```diff
@@ -1,36 +1,172 @@
...
+    if (active) { return }
+
+    reset()
+  }, [active])
+
+  // The following three useEffect calls should be unnecesary, but I was finding
+  // in testing that during render cycles where "passed" values updated, the useState
+  // calls weren't updating, so the UI would revert to the original state until refresh.
+  // I suspect that it might be because the save call makes three separate API calls,
+  // but at least adding these hooks makes things work.
```
eadams17: So I think what you're trying to achieve with using passed values for the initial state that will update on re-render without needing to use `useEffect` can be achieved with passing a key prop to this `TeacherEmailNotifications` component-- this post goes into a bit more detail of how that works: https://tkdodo.eu/blog/putting-props-to-use-state

eadams17 (approved):

happythenewsad (approved):

services/QuillLMS/app/models/user.rb
```diff
@@ -873,5 +871,21 @@ def school_or_district_premium?
...
   end
+
+  private def teacher_notification_settings_info
+    teacher_notification_settings.rollup_hash
+  end
+
+  private def school_account_info
+    return school if school&.name
+
+    School.find_by_name(School::NO_SCHOOL_SELECTED_SCHOOL_NAME)
```
dandrabik: Let's use `School.find_by(name: School::NO_SCHOOL_SELECTED_SCHOOL_NAME)`. (preferring `Model.find_by(attribute:)` is in a TODO Rubocop rule we haven't turned on yet).

services/QuillLMS/app/workers/teacher_notifications/send_notification_worker.rb
```diff
@@ -0,0 +1,86 @@
...
+        student_name: @activity_session.user.name,
+        classroom_name: @activity_session.classroom.name,
+        diagnostic_name: @activity_session.activity.name
+      })
+    end
+
+    private def send_complete_all_diagnostic_recommendations
+      unit_template = @activity_session.unit_template
+      # Return early if the activity that was just completed wasn't a recommendation
+      return if unit_template.nil? || unit_template.recommendations.empty?
```
dandrabik: Consider `return unless unit_template&.recommendations&.present?`

services/QuillLMS/app/workers/teacher_notifications/send_notification_worker.rb
```diff
@@ -0,0 +1,86 @@
...
+      })
+    end
+
+    private def send_notification(type, message_attrs)
+      @activity_session.teachers.each do |teacher|
+        type.create!(user: teacher, message_attrs: message_attrs) if teacher_receives_type(teacher, type)
+      end
+    end
+
+    private def teacher_receives_type(user, type)
```
dandrabik: This feels like a method that could be on user, e.g. `receives_notification_type?(type)`

services/QuillLMS/app/workers/teacher_notifications/send_notification_worker.rb
```diff
@@ -0,0 +1,86 @@
...
+
+module TeacherNotifications
+  class SendNotificationWorker
+    include Sidekiq::Worker
+
+    def perform(activity_session_id)
+      @activity_session = find_activity_session(activity_session_id)
+      return unless @activity_session
+
+      TeacherNotification.transaction do
```
dandrabik: Does this need to be wrapped in a transaction? We have a lot of DB contention on locks, so wondering if this could be removed. Maybe consider breaking into 3 jobs if the concern is keeping the job idempotent.

services/QuillLMS/app/workers/teacher_notifications/send_notification_worker.rb
```diff
@@ -0,0 +1,86 @@
...
+        .where.not(completed_at: nil)
+        .includes(:user)
+        .includes(teachers: :teacher_notification_settings)
+        .includes(:classroom)
+        .includes(activity: :classification)
+        .includes(unit_template: :recommendations)
+        .first
+    end
+
+    private def send_complete_diagnostic
```
dandrabik: Note, every method in this file has `activity_session` as its base object. Take or leave, you could consider moving some methods there, e.g. `activity_session.send_complete_diagnostic_notification`

services/QuillLMS/app/workers/teacher_notifications/send_notification_worker.rb
```diff
@@ -0,0 +1,86 @@
...
+    include Sidekiq::Worker
+
+    def perform(activity_session_id)
+      @activity_session = find_activity_session(activity_session_id)
+      return unless @activity_session
+
+      TeacherNotification.transaction do
+        send_complete_diagnostic
+        send_complete_all_diagnostic_recommendations
+        send_complete_all_assigned_activities
```
dandrabik: Another question. We'll be adding more of these notifications in the future. Is the plan to add them all in this file? That might make testing and maintenance more complicated than breaking them up into separate jobs.

dandrabik (approved):
Left some style notes and considerations, but nothing blocking approval. Thanks for pushing this along!

services/QuillLMS/app/workers/teacher_notifications/send_notification_worker.rb
anathomical: This is a good point.  Combined with your note below about expansion, I think it's probably a good idea to split it into multiple workers.  That said, each of the workers will be in a position where they write a variable number of records (based on how many teachers and co-teachers there are for the class an even takes place in), and I'm a little worried about one of those records saving, the job crashing, then re-running and doubling notifications for one of the teachers but not the other.  Or getting stuck in a retry loop where it keeps adding a record to one teacher, but the other always fails.  Maybe it's better to just take that risk than to add a transaction block, but that's what was on my mind when I added the transaction.

Thoughts?

PR #10551, opened 2023-05-10
rake task for activity session backfill

happythenewsad:
## WHAT
Some ActivitySessions have null values for updated_at. We want all rows to be non null, so that Airbyte can use this field as a cursor.

### Screenshots

### Notion Card Links
https://www.notion.so/quill/Product-Board-30f97e4eb01246dbb8264253fb385073?p=b9015bdad7c5444cbd7ad11f9f92beba&pm=s

PR Checklist | Your Answer
------------ | -------------

### DIFF
--- a/services/QuillLMS/lib/tasks/activity_sessions.rake
+++ b/services/QuillLMS/lib/tasks/activity_sessions.rake
@@ -4,7 +4,7 @@ namespace :activity_sessions do
   namespace :backfill do
     desc 'backfills updated_at so null values do not exist'
     task :updated_at => :environment do
-      sessions = ActivitySession.where(updated_at: nil)
+      sessions = ActivitySession.unscoped.where(updated_at: nil)
       sessions.each do |session|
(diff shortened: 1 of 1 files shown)


happythenewsad:
ActivitySession.unscoped.where(updated_at: nil).count
=> 23335 (staging)

rake activity_sessions:backfill:updated_at

ActivitySession.unscoped.where(updated_at: nil).count
=> 18514 (staging)


So, the script was only partially successful. This is either due to some records silently failing validations, or the fact that `joins` is an INNER JOIN, so some activity sessions are filtered out. Investigating.

I can't simply remove the `joins`, because of an `after_save` callback that assumes `activity` is available.

UPDATE: after reviewing the callbacks, I decided that they don't need to be run on super old data. So, I skipped validations via the `update_columns` API.

happythenewsad:
The latest commit works as intended - all nullish rows are backfilled on staging.

anathomical (approved):
I like the order of priority you've got here for figuring out the correct date to use.

dandrabik (approved):
:+1: LGTM

PR #10558, opened 2023-05-15
Lms/detach users from vitally schools

anathomical:
## WHAT
Add code to record when a Teacher changes which school they're in, and then modify nightly Vitally Sync code to `unlink` users from their old schools
## WHY
Vitally changed the way that they read our batch updates of users so that when a user has a different `accountId` than the last time we synced.  It used to be the case that the new value would override the old value, but the new handling attaches the user to both `accountId`s.  Having a user attached to two different accounts makes some uses Partnerships has harder to use.  Also, we just want Vitally to accurately reflect our actual data.
## HOW
- Begin recording `ChangeLog` records when a user changes what school they're assigned to
- During the nightly Vitally sync process, find ChangeLogs of User school values from the past 25 hours and unlink them in Vitally.

### Notion Card Links
https://www.notion.so/quill/Update-Vitally-Analytics-API-Rule-to-Unlink-Endpoint-only-show-users-at-their-current-school-asso-d875631f26fb456283c2c1eaaa2bfa5d?pvs=4

PR Checklist | Your Answer
------------ | -------------

### DIFF
--- a/services/QuillLMS/app/models/schools_users.rb
+++ b/services/QuillLMS/app/models/schools_users.rb
@@ -20,8 +20,25 @@ class SchoolsUsers < ApplicationRecord
 
   # When a teacher sets their school, we make sure they they have the appropriate subscription type.
   after_save :update_subscriptions
+  after_save :school_changed_change_log, if: :should_write_change_log?
 
   def update_subscriptions
     user&.updated_school(school_id)
+  end
+
+  private def should_write_change_log?
+    saved_change_to_attribute?(:school_id) &&
+      attribute_before_last_save(:school_id)
+  end
+
+  private def school_changed_change_log
+    ChangeLog.create({
+      action: ChangeLog::USER_ACTIONS[:update],
+      changed_record_type: 'User',
+      changed_record_id: user_id,
+      changed_attribute: 'school',
+      previous_value: attribute_before_last_save(:school_id),
+      new_value: attribute_in_database(:school_id)
+    })
   end
 end
--- a/services/QuillLMS/app/models/schools_users.rb
+++ b/services/QuillLMS/app/models/schools_users.rb
@@ -20,23 +20,18 @@ class SchoolsUsers < ApplicationRecord
 
   # When a teacher sets their school, we make sure they they have the appropriate subscription type.
   after_save :update_subscriptions
-  after_save :school_changed_change_log, if: :should_write_change_log?
+  after_update :school_changed_change_log, if: proc { saved_change_to_attribute?(:school_id) }
 
   def update_subscriptions
     user&.updated_school(school_id)
   end
 
-  private def should_write_change_log?
-    (saved_change_to_attribute?(:school_id) &&
-      attribute_before_last_save(:school_id))
-  end
-
   private def school_changed_change_log
     ChangeLog.create({
       action: ChangeLog::USER_ACTIONS[:update],
-      changed_record_type: 'User',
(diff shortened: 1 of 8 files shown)


cissyxyu (approved):

services/QuillLMS/app/models/schools_users.rb
```diff
@@ -20,8 +20,25 @@ class SchoolsUsers < ApplicationRecord
 
   # When a teacher sets their school, we make sure they they have the appropriate subscription type.
   after_save :update_subscriptions
+  after_save :school_changed_change_log, if: :should_write_change_log?
```
dandrabik: Should this be an `after_update` so we don't have a changelog on create?

services/QuillLMS/app/models/schools_users.rb
```diff
@@ -20,8 +20,25 @@ class SchoolsUsers < ApplicationRecord
...
 
   def update_subscriptions
     user&.updated_school(school_id)
   end
+
+  private def should_write_change_log?
+    (saved_change_to_attribute?(:school_id) &&
```
dandrabik: I think you only need this first one if you use `after_update`.

You could also consider removing this method and changing the `if` to `if: :saved_change_to_school_id?`

services/QuillLMS/app/workers/sync_vitally_unlinks_worker.rb
```diff
@@ -0,0 +1,13 @@
+# frozen_string_literal: true
+
+class SyncVitallyUnlinksWorker
```
dandrabik, line 3: Would a singular `Unlink` be more accurate?

services/QuillLMS/app/models/schools_users.rb
```diff
@@ -20,8 +20,25 @@ class SchoolsUsers < ApplicationRecord
...
+      attribute_before_last_save(:school_id))
+  end
+
+  private def school_changed_change_log
+    ChangeLog.create({
+      action: ChangeLog::USER_ACTIONS[:update],
+      changed_record_type: 'User',
+      changed_record_id: user_id,
+      changed_attribute: 'school',
```
dandrabik: Should the attribute be `school_id`?

services/QuillLMS/app/models/schools_users.rb
```diff
@@ -20,8 +20,25 @@ class SchoolsUsers < ApplicationRecord
...
+
+  private def should_write_change_log?
+    (saved_change_to_attribute?(:school_id) &&
+      attribute_before_last_save(:school_id))
+  end
+
+  private def school_changed_change_log
+    ChangeLog.create({
+      action: ChangeLog::USER_ACTIONS[:update],
+      changed_record_type: 'User',
```
dandrabik: You are using two strings to link logic between here and the run (`User` and `school`), it would be better if they referenced constants.

services/QuillLMS/app/workers/sync_vitally_worker.rb
```diff
@@ -39,6 +46,12 @@ def perform
     end
   end
 
+  private def unlinks_to_run
+    ChangeLog.where(changed_record_type: 'User')
```
dandrabik: Strings for logic should be constants.

dandrabik (approved):
:+1: Overall looks good. I left a few cleanup comments. Thanks for pushing this forward.

PR #10576, opened 2023-05-19
Upgrade ruby to 2.7.8

cissyxyu:
## WHAT
Upgrade our ruby version to 2.7.8.

## WHY
So our site is up to date with Ruby and latest security patches.

## HOW
Bump up the version and bump up related gems that need to be updated.

### Screenshots

### Notion Card Links
https://www.notion.so/quill/Upgrade-to-Ruby-2-7-8-d283a95af0cc46c49c01160155392625?pvs=4

PR Checklist | Your Answer
------------ | -------------


services/QuillLMS/spec/models/cron_spec.rb
```diff
@@ -121,6 +121,7 @@
     end
 
     it "enqueues TeacherNotifications::EnqueueUsersForRollupEmailWorker" do
+      expect(TeacherNotifications::EnqueueUsersForRollupEmailWorker).to receive(:perform_async).with(TeacherInfo::WEEKLY_EMAIL) if Time.zone.today.wday == 5
```
emilia-friedberg: Is this supposed to be in here?

emilia-friedberg (approved):

services/QuillLMS/Gemfile
```diff
@@ -72,6 +72,7 @@ gem 'configs'
 gem 'rack-test', '~> 0.6.3'
 gem 'secure_headers', '6.3.2'
 gem 'faker'
+gem 'cgi', '~> 0.3.6'
```
happythenewsad, line 75: Perhaps this diff goes away after rebasing?

services/QuillLMS/spec/models/cron_spec.rb
cissyxyu: Removed! Was fixing a broken spec I noticed on develop but it just got fixed in another PR, so this is unneeded.

services/QuillLMS/Gemfile
cissyxyu: So, I needed to manually put in this gem here because we were getting an RSpec error on one RSpec test after upgrading the ruby version. I looked it up and it turns out that on the new ruby version 2.7.8 we need to specify a certain version of this gem, cgi. Otherwise some RSpec unit tests will fail.

https://discuss.rubyonrails.org/t/invalid-domain-example-com-in-rspec-after-changing-session-store-to-domain-all/81922

happythenewsad (approved):

brendanshean (approved):

PR #10583, opened 2023-05-23
Session json bug

happythenewsad:
## WHAT / HOW
- Have our main server side auth flow handle JSON requests correctly, by return JSON responses instead of HTML responses
- There may be some instances when we want to hard-refresh a webpage due to a failing XHR request. This PR passes redirect information in JSON format, delegating this decision to individual client components.

'JSON requests' are here defined as requests with header `Accept: application/json`

Some subtleties to note:

1. In Rails, `as: :json` sets the `Content-Type` header, while `respond_to` interrogates the `Accept` header

2. `application_controller.rb:130` does a similar thing, but handles a different set of code paths (i.e. `auth_failed` is called in places where InvalidAuthenticityToken exceptions do not occur.

# WHY
So that we don't get unnecessary `ActionController::UnknownFormat` errors

### Screenshots

### Notion Card Links
https://www.notion.so/quill/Product-Board-30f97e4eb01246dbb8264253fb385073?p=61a1376540ed4657a8f27c2b2c6ec23c&pm=s

PR Checklist | Your Answer
------------ | -------------

### DIFF
--- a/services/QuillLMS/app/lib/quill_authentication.rb
+++ b/services/QuillLMS/app/lib/quill_authentication.rb
@@ -125,11 +125,10 @@ module QuillAuthentication
 
   def auth_failed
     if request.headers['Accept'] == 'application/json'
-      puts "\n\n HERE \n\n"
-      render(json: {redirect: new_session_path}, status: 401) and return
+      render(json: { redirect: new_session_path }, status: 401)
+    else
+      redirect_to(profile_path, notice: "404")
     end
-
-    redirect_to(profile_path, notice: "404")
   end
 
   def admin?
--- a/services/QuillLMS/app/lib/quill_authentication.rb
+++ b/services/QuillLMS/app/lib/quill_authentication.rb
@@ -123,15 +123,13 @@ module QuillAuthentication
     auth_failed
   end
 
-  def auth_failed(hard: true)
-    if hard
-      sign_out
-      session[:attempted_path] = request.fullpath
-      redirect_to(new_session_path, status: :see_other)
-    else
-      redirect_to(profile_path, notice: "404")
+  def auth_failed
+    if request.headers['Accept'] == 'application/json'
+      puts "\n\n HERE \n\n"
+      render(json: {redirect: new_session_path}, status: 401) and return
     end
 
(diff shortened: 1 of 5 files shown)


services/QuillLMS/app/lib/quill_authentication.rb
```diff
@@ -123,15 +123,13 @@ def signed_in!
...
   end
 
-  def auth_failed(hard: true)
-    if hard
-      sign_out
-      session[:attempted_path] = request.fullpath
-      redirect_to(new_session_path, status: :see_other)
-    else
-      redirect_to(profile_path, notice: "404")
+  def auth_failed
```
happythenewsad: Unrelatedly, the argument to auth_failed is never used, so I'm removing it

services/QuillLMS/app/lib/quill_authentication.rb
```diff
@@ -123,15 +123,13 @@ def signed_in!
...
+    if request.headers['Accept'] == 'application/json'
+      puts "\n\n HERE \n\n"
+      render(json: {redirect: new_session_path}, status: 401) and return
```
brendanshean: What about a regular if/else block and then get rid of the `and return`

services/QuillLMS/spec/controllers/lib/quill_authentication_spec.rb
```diff
@@ -4,10 +4,51 @@
...
+        get '/example_action', to: 'fake#example_action'
+      end
+    end
 
+    after do
+      Rails.application.reload_routes!
+    end
+
+    context 'Accept: text/html' do
+      it 'should return JSON, status 401' do
```
brendanshean: status 302?

brendanshean (commented):
I think this looks good

services/QuillLMS/spec/controllers/lib/quill_authentication_spec.rb
happythenewsad: Thanks, that's a copy-paste artifact

brendanshean (approved):
Great job tracking this one down.

anathomical (approved):

PR #10585, opened 2023-05-23
Lms/admin snapshot report backend milestone1

anathomical:
## WHAT
Add the initial skeleton for getting Admin Snapshot data for the dashboard.
## WHY
We want a solid, re-usable pattern for quickly adding new snapshot queries for easy expansion.  We also want the API calls to be performant.
## HOW
- Create a framework for snapshot queries that understands how to single-number rollup numbers for a given admin using BigQuery (note: this is based on my mini-RFC, but I would appreciate feedback on this pattern as I'm not totally confident in it)
- Implement a first batch of 11 queries that we know we'll want on that page
- Create a small lib utility that can take Snapshot query params and generate a consistent cache key from them
- Create a worker that can run the specified queries and stash them in the appropriate cache key, then send a Pusher notification to the front-end so that the data can be requested again
- Create a controller action that can be used to request single-number rollup stats that
- - checks the cache to see if the data is there and returns it if so
- - initiates the worker to generate the appropriate data if the cache is empty
- Create a controller action that lists the values that should be used in the report filter section based on current params (e.g. it populates the `schools` segment based on the current user)

### Notion Card Links
https://www.notion.so/quill/Implement-Admin-Usage-Snapshot-Report-Backend-0dd45ff4b92641328cec85660ea614fb

PR Checklist | Your Answer
------------ | -------------


services/QuillLMS/app/controllers/snapshots_controller.rb
```diff
@@ -0,0 +1,168 @@
...
+  end
+
+  private def find_timeframe(timeframe_value)
+    TIMEFRAMES.find { |timeframe| timeframe[:value] == timeframe_value }
+  end
+
+  private def calculate_timeframes
+    timeframe = find_timeframe(snapshot_params[:timeframe])
+
+    if snapshot_params[:timeframe] == 'custom'
```
emilia-friedberg: Should 'custom' here be `Snapshots::CacheKeys::CUSTOM_TIMEFRAME_NAME` from above?

emilia-friedberg (approved):
This looks great to me. One tiny question about a constant but I'm really impressed by the construction and readability here.

services/QuillLMS/app/controllers/snapshots_controller.rb
```diff
@@ -0,0 +1,168 @@
+# frozen_string_literal: true
+
+class SnapshotsController < ApplicationController
+  DEFAULT_TIMEFRAME = 'last-30-days'
+
+  TIMEFRAMES = [
+    {
+      value: 'last-30-days',
+      name: 'Last 30 days',
+      previous_start: proc { DateTime.current - 60.days },
```
dandrabik: I believe we want these start times to be pinned to the end of the day before. This is the SLA for these queries, and it will also help with caching since it'll be the same group of timeframes all day.

These procs likely need a param named `end_date` (or some other name) that is calculated once and passed in to each.

services/QuillLMS/app/controllers/snapshots_controller.rb
```diff
@@ -0,0 +1,168 @@
...
+    }, {
+      value: 'last-year',
+      name: 'Last year',
+      previous_start: proc { DateTime.current.beginning_of_year - 2.years },
+      current_start: proc { DateTime.current.beginning_of_year - 1.year },
+      current_end: proc { DateTime.current.beginning_of_year },
+    }, {
+      value: 'all-time',
+      name: 'All time',
+      previous_start: proc { DateTime.new(2010,1,1) }, # This is well before anydata exists in our system, so works for "all time"
```
dandrabik: So "All Time" data shouldn't have a previous period, I would expect something like `nil` here. How will the front end know not to display these queries and will we end up running these queries twice on the backend?

services/QuillLMS/app/controllers/snapshots_controller.rb
```diff
@@ -0,0 +1,168 @@
...
+    }, {
+      value: 'last-90-days',
+      name: 'Last 90 days',
+      previous_start: proc { DateTime.current - 180.days },
+      current_start: proc { DateTime.current - 90.days },
+      current_end: proc { DateTime.current },
+    }, {
+      value: 'this-month',
+      name: 'This month',
+      previous_start: proc { DateTime.current.beginning_of_month - 1.month },
```
dandrabik: So this `DateTime.current` (which should be EOD yesterday as noted above) needs to be run once and used by all the methods in the call, otherwise, these will be slightly off from each other, which could cause weird edge cases at the end of the day/month/year.

services/QuillLMS/app/lib/snapshots/cache_keys.rb
```diff
@@ -0,0 +1,24 @@
...
+        timeframe_part += "-#{timeframe_hash[:custom_start]}-#{timeframe_hash[:custom_end]}"
+      end
+
+      [
+        "admin-snapshot",
+        query,
+        timeframe_part,
+        "school-ids-#{school_ids.sort.join('-')}",
+        "grades-#{grades.map(&:to_s).sort.join('-')}",
+        user_id
```
dandrabik: Tiny optimization / nitpick, `user_id` should be moved up in this order to second, so the keys move from larger buckets to smaller.

Also `user_id` would work, but should this actually be `school_id` and/or `district_id` since that's what the report is specific to?

services/QuillLMS/app/lib/snapshots/cache_keys.rb
```diff
@@ -0,0 +1,24 @@
...
+    def self.generate_key(query, user_id, timeframe_hash, school_ids, grades)
+      timeframe_part = timeframe_hash[:name]
+
+      if timeframe_part == CUSTOM_TIMEFRAME_NAME
+        timeframe_part += "-#{timeframe_hash[:custom_start]}-#{timeframe_hash[:custom_end]}"
+      end
+
+      [
+        "admin-snapshot",
+        query,
```
dandrabik, line 8: So building on my pinning to yesterday's EOD. I think we need a date in the cache key, so that this is consistently wiped when a new day starts.

services/QuillLMS/app/queries/snapshots/options_query.rb
```diff
@@ -0,0 +1,68 @@
+# frozen_string_literal: true
+
+module Snapshots
+  class OptionsQuery
+    attr_accessor :admin_id
+
+    def self.run(*args)
```
dandrabik, line 7: Could consider making this a child class of `ApplicationService` to match other code in the app.

services/QuillLMS/app/controllers/snapshots_controller.rb
```diff
@@ -0,0 +1,168 @@
...
+
+  def options
+    render json: {
+      timeframes: TIMEFRAMES.map { |timeframe| format_timeframe_option(timeframe) },
+      schools: Snapshots::SchoolsOptionsQuery.run(current_user.id),
+      grades: GRADE_OPTIONS
+    }
+  end
+
+  private def cache_key
```
dandrabik: It worries me a bit that we are generating the cache key in two places, once in the controller, once in the worker. Those params could accidentally drift. Maybe we should pass the `cache_key` to the worker so that it doesn't have to generate it.

services/QuillLMS/app/workers/snapshots/cache_snapshot_count_worker.rb
```diff
@@ -0,0 +1,62 @@
...
+
+      previous_snapshot = QUERIES[query].run(user_id,
+        previous_timeframe_start,
+        current_timeframe_start,
+        school_ids,
+        grades)
+
+      payload = { current: current_snapshot, previous: previous_snapshot }
+
+      cache_key = Snapshots::CacheKeys.generate_key(query,
```
dandrabik: This feels like we should already know this cache key via the controller.

services/QuillLMS/app/controllers/snapshots_controller.rb
```diff
@@ -0,0 +1,168 @@
...
+    {value: "11", name: "11th"},
+    {value: "12", name: "12th"},
+    {value: "University", name: "University"},
+    {value: "Other", name: "Other"}
+  ]
+
+  def count
+    response = Rails.cache.read(cache_key)
+
+    return render json: response if response
```
dandrabik, line 32: So the controller sometimes returns a valid response and sometimes doesn't. Maybe we should always follow the same path and have a single source of the caching, say in the worker?

services/QuillLMS/app/workers/snapshots/cache_snapshot_count_worker.rb
```diff
@@ -0,0 +1,62 @@
...
+          name: timeframe_name,
+          custom_start: timeframe_name == Snapshots::CacheKeys::CUSTOM_TIMEFRAME_NAME ? current_timeframe_start : nil,
+          custom_end: timeframe_name == Snapshots::CacheKeys::CUSTOM_TIMEFRAME_NAME ? current_timeframe_end : nil
+        },
+        school_ids,
+        grades)
+
+      Rails.cache.write(cache_key, payload, expires_in: DEFAULT_CACHE_EXPIRATION)
+
+      PusherTrigger.run(user_id, PUSHER_EVENT,
```
dandrabik, line 24: Hmm, surprised to see this isn't pushing the data to the frontend. Is this just asking the frontend to trigger another web request? There are some benefits/drawbacks to each approach, probably worth discussing.

So this flow is:
- Web request
- Empty cache, trigger job
- Run job, populate cache, notify frontend
- Web request, hit cache

We could also do:
- Web request
- Always run job
- Fetch data (cached or not)
- Push data to frontend

services/QuillLMS/app/controllers/snapshots_controller.rb
```diff
@@ -0,0 +1,168 @@
...
+      grades: GRADE_OPTIONS
+    }
+  end
+
+  private def cache_key
+    Snapshots::CacheKeys.generate_key(snapshot_params[:query],
+      current_user.id,
+      {
+        name: snapshot_params[:timeframe_name],
+        custom_start: snapshot_params[:timeframe_custom_start],
```
dandrabik: Are we sure these custom timeframes will be `nil` when passed in from the frontend? E.g. if a user starts playing with the calendar select, then selects a set timeframe.

services/QuillLMS/app/controllers/snapshots_controller.rb
```diff
@@ -0,0 +1,168 @@
...
+    ]
+  end
+
+  private def custom_previous_start
+    timeframe_length = custom_current_end - custom_current_start
+    custom_current_start - timeframe_length
+  end
+
+  private def custom_current_start
+    DateTime.parse(snapshot_params[:timeframe_custom_start])
```
dandrabik: Could these be `nil`? It seems some of the code is expecting nils and some expecting dates in these fields.

services/QuillLMS/app/queries/snapshots/period_query.rb
```diff
@@ -0,0 +1,95 @@
...
+
+    def run
+      raise NotImplementedError
+    end
+
+    def run_query
+      QuillBigQuery::Runner.execute(query)
+    end
+
+    def query
```
dandrabik, line 26: So writing these generic query combiners is always tricky. I'm a bit on the fence about whether it would be better (but more redundant) to have individual queries share only the `where_clause` and explicitly write everything else out (there could be some shared constants). I worry a little bit that overwriting or modifying superclass vars might make the queries harder to parse (but it is less redundant). Maybe worth a discussion.

dandrabik (changes requested):
Nice work on this so far! Thanks for getting a PR up quickly. I like the general shape here. 

There's a few things that need to be adjusted slightly, and since this is a pattern we will build upon, I think it's worth discussing some of the alternate decisions possible here, just to kick the tires a bit.

services/QuillLMS/app/lib/snapshots/cache_keys.rb
dandrabik: Per conversation, even outside of caching, we likely need to know what school or district the user is viewing.

dandrabik: Per conversation, if we are using a school or district id from params, need to determine in the controller the user has access there.

services/QuillLMS/spec/queries/snapshots/active_students_query_spec.rb
```diff
@@ -0,0 +1,23 @@
+# frozen_string_literal: true
+
+require 'rails_helper'
+
+module Snapshots
+  describe 'ActiveStudentsQuery' do
+    context 'external_api', :external_api do
+      before do
```
brendanshean: I would add 
```
  config.around(:each, :external_api) do |example|
    VCR.configure { |c| c.allow_http_connections_when_no_cassette = true }
    example.run
    VCR.configure { |c| c.allow_http_connections_when_no_cassette = false }
  end
```
to spec_helper similar to the `caching` tag [configuration](https://github.com/empirical-org/Empirical-Core/blob/develop/services/QuillLMS/spec/rails_helper.rb#L92)

Then you can get rid of all the before after VCR in all of these tests in addition to the ones that @happythenewsad wrote

services/QuillLMS/app/controllers/snapshots_controller.rb
```diff
@@ -0,0 +1,168 @@
...
+      default: (DEFAULT_TIMEFRAME == timeframe[:value])
+    }
+  end
+
+  private def find_timeframe(timeframe_value)
+    TIMEFRAMES.find { |timeframe| timeframe[:value] == timeframe_value }
+  end
+
+  private def calculate_timeframes
+    timeframe = find_timeframe(snapshot_params[:timeframe])
```
happythenewsad: We should explicitly handle the case where find_timeframe returns `nil`.

services/QuillLMS/app/workers/snapshots/cache_snapshot_count_worker.rb
```diff
@@ -0,0 +1,62 @@
+# frozen_string_literal: true
+
+module Snapshots
+  class CacheSnapshotCountWorker
+    include Sidekiq::Worker
+
+    DEFAULT_CACHE_EXPIRATION = 24.hours
+    PUSHER_EVENT = 'admin-snapshot-count-cached'
+
+    QUERIES = {
```
happythenewsad, line 9: It would be great to use module reflection here, so that this hash doesn't become stale when new queries are added. 

One technique is:
`Snapshots.constants.filter {|q| Snapshots.const_get(q).is_a? Class }` 

Caveats:
1.  Ensure that Rails autoloads `Snapshots.constants` 
2. If you don't want the worker to perform every query in the module, consider earmarking runnable queries with a Snaphots::Autorun namespace, or a class method.

happythenewsad (commented):
Looking good. 

What order of magnitude is the cache key count? How large is the combined cache value size? It would be good to know before launch. If it's small enough, we could simplify the SnapshotsController::Count logic by just computing all values for everyone every night.

And if we don't use that strategy, we still want to know how big the cache can grow organically.

happythenewsad (commented):
Looking good. 

What order of magnitude is the cache key count? How large is the combined cache value size? It would be good to know before launch. If it's small enough, we could simplify the SnapshotsController::Count logic by just computing all values for everyone every night.

And if we don't use that strategy, we still want to know how big the cache can grow organically.

services/QuillLMS/app/workers/snapshots/cache_snapshot_count_worker.rb
anathomical: After touching base with Emilia she raised a valid point: Pusher has a payload limit of 10KB.  While the values we're expecting in these snapshot returns, so it shouldn't be an issue, a number of our other reports easily exceed this limit.  This suggests that maintaining the existing pattern is worth doing since there are some cases where the second alternative you mention isn't an option, but the former alternative should always work.

services/QuillLMS/app/workers/snapshots/cache_snapshot_count_worker.rb
anathomical: Do you have a good sense of how wee build a list of "human readable slugs" that refer to each report?  We need something for the front-end to be able to pass to say "I want this specific report" since we don't just automatically run all the queries on any single trigger at the moment.

anathomical:
> Looking good.
> 
> What order of magnitude is the cache key count? How large is the combined cache value size? It would be good to know before launch. If it's small enough, we could simplify the SnapshotsController::Count logic by just computing all values for everyone every night.
> 
> And if we don't use that strategy, we still want to know how big the cache can grow organically.

@happythenewsad This is a good question.  Cache key count is theoretically huge since people can create custom date ranges and we'd cache each timeframe for each query (so, like, 30-ish keys per timeframe).  In practice, most people probably only use a couple of standard timeframes, maybe 90-120-ish keys per user who uses this page?  Data stored in cache is very small (hundreds of bytes), since these snapshots are so limited.

I think that the ability for a user to do a bunch of custom filtering probably means pre-calculating overnight doesn't make a ton of sense, but I'd definitely bee open to discussion on that front.

As far as cache growth, I think that the small number of admins combined with the small payloads being cached means that even if we end up with quite a few keys per user will still leave us with a pretty negligible amount of data going into the cache even for very heavy users.

services/QuillLMS/app/workers/snapshots/cache_snapshot_count_worker.rb
happythenewsad: One option is to use `<nice_snake_case>.camelize.constantize` in Rails, so you can send `<nice_snake_case>` from the front end

services/QuillLMS/app/controllers/snapshots_controller.rb
```diff
@@ -0,0 +1,193 @@
...
+  end
+
+  private def school_ids_param_valid?
+    snapshot_params[:school_ids]&.any?
+  end
+
+  private def authorize_request
+    schools_user_admins = current_user.administered_schools.pluck(:id)
+
+    return render json: { error: 'user is not authorized for all specified schools' }, status: 403 unless snapshot_params[:school_ids]&.all? { |param_id| schools_user_admins.include?(param_id.to_i) }
```
brendanshean: This line is kind of long.  You could break it up as follows:

```
return if snapshot_params[:school_ids]&.all? { |param_id| schools_user_admins.include?(param_id.to_i) }

render json: { error: 'user is not authorized for all specified schools' }, status: 403 
```

services/QuillLMS/app/controllers/snapshots_controller.rb
brendanshean, line 1: Given all the logic surrounding timeframes, what about extracting the logic into its own class?

services/QuillLMS/spec/controllers/snapshots_controller_spec.rb
```diff
@@ -0,0 +1,235 @@
...
+    let(:query_class) { controller.class::QUERIES[query_name] }
+    let(:now) { DateTime.current }
+    let(:current_snapshot_stub) { 'CURRENT' }
+    let(:previous_snapshot_stub) { 'PREVIOUS' }
+    let(:school) { create(:school) }
+    let(:user) { create(:user, administered_schools: [school]) }
+    let(:school_ids) { [school.id.to_s] }
+
+    before do
+      allow(controller).to receive(:current_user).and_return(user)
```
brendanshean: I think this is redundant given line 9

services/QuillLMS/spec/controllers/snapshots_controller_spec.rb
```diff
@@ -0,0 +1,235 @@
...
+
+      expect(controller.send(:calculate_timeframes)).to eq([
+        end_of_yesterday - 60.days,
+        end_of_yesterday - 30.days,
+        end_of_yesterday
+      ])
+    end
+
+    it 'accurately calculates a custom timeframe previous_start value' do
+      now = DateTime.current.change(usec: 0) # strip fractional seconds to simplify conversion between string and DateTime
```
brendanshean: Cool.  I haven't seen `usec: 0` before

services/QuillLMS/app/queries/snapshots/period_query.rb
```diff
@@ -0,0 +1,88 @@
+# frozen_string_literal: true
+
+module Snapshots
+  class PeriodQuery
+    attr_accessor :timeframe_start, :timeframe_end, :school_ids, :grades
+
+    def self.run(*args)
```
brendanshean, line 7: You could inherit from `ApplicationService` here.

services/QuillLMS/app/controllers/snapshots_controller.rb
```diff
@@ -0,0 +1,193 @@
...
+    {value: "11", name: "11th"},
+    {value: "12", name: "12th"},
+    {value: "University", name: "University"},
+    {value: "Other", name: "Other"}
+  ]
+
+  before_action :validate_request, only: [:count]
+  before_action :authorize_request, only: [:count]
+
+  def count
```
dandrabik, line 25: In this method, you are calling `calculate_timeframes` 3 times (in `cache_key` 2 times, and in the controller). Those could all be different at the end of the day/month/year. You want to calculate them once and use them throughout, e.g.
```ruby
def count
  previous_start, current_start, current_end = calculate_timeframes
  cache_key = cache_key_for_time(previous_start, current_start, current_end, snapshots_params)
  ...

File: services/QuillLMS/app/controllers/snapshots_controller.rb [hunk 23]
```diff
@@ -0,0 +1,193 @@
...
+  private def format_timeframe_option(timeframe)
+    {
+      value: timeframe[:value],
+      name: timeframe[:name],
+      default: (DEFAULT_TIMEFRAME == timeframe[:value])
+    }
+  end
+
+  private def find_timeframe(timeframe_value)
+    TIMEFRAMES.find { |timeframe| timeframe[:value] == timeframe_value }
```
dandrabik: Do we want a default here? This could be `nil`. Maybe return the frontend default selection?

services/QuillLMS/app/controllers/snapshots_controller.rb
```diff
@@ -0,0 +1,193 @@
...
+    return render json: response if response
+
+    previous_timeframe_start, current_timeframe_start, timeframe_end = calculate_timeframes
+
+    Snapshots::CacheSnapshotCountWorker.perform_async(cache_key,
+      snapshot_params[:query],
+      current_user.id,
+      {
+        name: snapshot_params[:timeframe],
+        previous_start: previous_timeframe_start,
```
dandrabik: I would use the shorter key names here throughout, e.g. `previous_start` instead of `previous_timeframe_start`

services/QuillLMS/app/lib/snapshots/cache_keys.rb
```diff
@@ -0,0 +1,17 @@
+# frozen_string_literal: true
+
+module Snapshots
+  class CacheKeys
+    def self.generate_key(query, previous_timeframe_start, current_timeframe_start, timeframe_end, school_ids, grades)
```
dandrabik: Consider using the short `previous_start` names here as well.

services/QuillLMS/app/lib/snapshots/cache_keys.rb
```diff
@@ -0,0 +1,17 @@
...
+      [
+        "admin-snapshot",
+        query,
+        previous_timeframe_start,
```
dandrabik: I believe `previous_start` is redundant and isn't needed by this method.

services/QuillLMS/app/workers/snapshots/cache_snapshot_count_worker.rb
```diff
@@ -0,0 +1,54 @@
...
+      'activities-assigned' => Snapshots::ActivitiesAssignedQuery,
+      'activities-completed' => Snapshots::ActivitiesCompletedQuery,
+      'average-activities-completed-per-student' => Snapshots::AverageActivitiesCompletedPerStudentQuery,
+      'sentences-written' => Snapshots::SentencesWrittenQuery,
+      'student-learning-hours' => Snapshots::StudentLearningHoursQuery
+    }
+
+    def perform(cache_key, query, user_id, timeframe, school_ids, grades)
+      timeframe_name = timeframe[:name]
+      previous_timeframe_start = timeframe[:previous_start]
```
dandrabik, line 35: Consider carrying through shorter var names here as well, e.g. `previous_start` for readability.

services/QuillLMS/app/queries/snapshots/activities_assigned_query.rb
```diff
@@ -0,0 +1,37 @@
...
+
+    def select_clause
+      "SELECT DISTINCT classroom_units.id, classroom_units.unit_id, array_length(classroom_units.assigned_student_ids) AS students_assigned_count"
+    end
+
+    def relevant_date_column
+      "classroom_units.created_at"
+    end
+
+    def group_by_clause
```
brendanshean: This might be redundant given inheritance with PeriodQuery

services/QuillLMS/app/workers/snapshots/cache_snapshot_count_worker.rb
```diff
@@ -0,0 +1,54 @@
...
+          current_timeframe_start,
+          school_ids,
+          grades)
+      else
+        previous_snapshot = nil
+      end
+
+      payload = { current: current_snapshot, previous: previous_snapshot }
+
+      Rails.cache.write(cache_key, payload, expires_in: DEFAULT_CACHE_EXPIRATION)
```
dandrabik: Given we are pinning to yesterday now, we shouldn't have a constant 24 hours expiration. It should expire when the day changes. So this should be something like `expires_in: current_end + 1.day`

services/QuillLMS/spec/queries/snapshots/active_classrooms_query_spec.rb
```diff
@@ -0,0 +1,17 @@
+# frozen_string_literal: true
+
+require 'rails_helper'
+
+module Snapshots
+  describe 'ActiveClassroomsQuery' do
```
brendanshean: If you use a constant in the describe block (instead of a string) you can then use described_class in the block:

```
describe Snapshots::ActivitiesAssignedQuery do
  include_context 'Snapshot Query Params'

  context 'external_api', :external_api do
    it 'should successfully get data' do
      result = described_class.run(timeframe_start, timeframe_end, school_ids, grades)
    end
  end
end
```

services/QuillLMS/app/workers/snapshots/cache_snapshot_count_worker.rb
dandrabik: Note, I would consider this refactor a "nice to have, but not a requirement". This code is specific to this dashboard, so there will be 27 queries (not 100's), all set up in the next two weeks.

services/QuillLMS/spec/queries/snapshots/active_classrooms_query_spec.rb
brendanshean: Same recommendation applies to the other six snapshot query specs.

services/QuillLMS/app/controllers/snapshots_controller.rb
dandrabik: I noted this below, but you don't need `previous_start` for the cache_key calculation

services/QuillLMS/app/controllers/snapshots_controller.rb
```diff
@@ -0,0 +1,193 @@
+# frozen_string_literal: true
+
+class SnapshotsController < ApplicationController
+  DEFAULT_TIMEFRAME = 'last-30-days'
+
+  TIMEFRAMES = [
+    {
+      value: 'last-30-days',
+      name: 'Last 30 days',
+      previous_start: proc { |end_of_yesterday| end_of_yesterday - 60.days },
```
dandrabik: Take or leave. So `end_of_yesterday` is accurate to the value we are passing into this method right now, but you could pass any date in here (say we changed end date logic), so it might be too descriptive for a method param. I get that you are trying to document the logic, so there is value, but it couples this config to that one-line date calculation (and they could drift and get confusing).

services/QuillLMS/app/controllers/snapshots_controller.rb
```diff
@@ -0,0 +1,193 @@
...
+    if snapshot_params[:timeframe] == 'custom'
+      return [
+        send(timeframe[:previous_start]),
+        send(timeframe[:current_start]),
+        send(timeframe[:current_end])
+      ]
+    end
+
+    [
+      timeframe[:previous_start].call(end_of_yesterday),
```
dandrabik: Note, take or leave, but you could consider making the custom timeframe and general proc signature the same so you don't have to branch code here, e.g. 
```ruby
custom_params = snapshot_params.slice(:timeframe_custom_start, :timeframe_custom_end)
timeframe[:previous_start].call(end_of_yesterday, custom_params),

# config
{
      value: 'custom',
      name: 'Custom',
      current_start: proc {|_, params| Date.parse(params[:timeframe_custom_start]) }
      ...
}
```
The procs will be more complicated, so this is a take or leave, but I always like eliminating branches if it's not too complicated.

dandrabik (approved):
Nice work on this. I left a bunch of comments on things to tighten up, but approving with the understanding that the following two issues will be fixed before merge:

1. Only do the timeframe calculation once.
2. Fix the cache expiration to dynamically match the new EOD yesterday logic.

The rest of my comments are take or leave and/or code organization recommendations. Thanks for pushing this forward!

brendanshean (approved):
Great work on this.  It's really cool to see  how you've abstracted the snapshot queries.  I had one comment on Timeframes but the rest should be pretty straightforward.

brendanshean (approved):
Great work on this.  It's really cool to see  how you've abstracted the snapshot queries.  I had one comment on Timeframes but the rest should be pretty straightforward.

services/QuillLMS/app/controllers/snapshots_controller.rb
anathomical: This can't currently be `nil` because the `before_action` `validate_request` will return an error if a timeframe is not provided and is not in the list of valid timeframes.

PR #10590, opened 2023-05-24
Send LearnWorlds data to Vitally Sync

brendanshean:
## WHAT
1. Persist LearnWorlds data on Quill
2. Handle LearnWorlds webhooks for various course events
3. Sync data with Vitally

## WHY
Sales/Partnerships team is interested how LearnWorlds is being used

## HOW
1.  Create LearnWorldsCourse: stores the course title and LearnWorldsAccountCourseEvent: links LearnWorldsAccount with LearnWorlds course along with the event type
2.  Add Webhooks handler that for each of the three types: enrolled, completed and earned certificate
3.  Add user info about the enrolled, completed and earned certificate courses to the vitally sync

### Screenshots

### Notion Card Links
https://www.notion.so/quill/Send-data-from-Learnworlds-Quill-via-API-Vitally-5fd55ec9ed1b43a0a43108790ef6445c?pvs=4

PR Checklist | Your Answer
------------ | -------------

### DIFF
--- a/services/QuillLMS/app/controllers/learn_worlds_integration/webhooks_controller.rb
+++ b/services/QuillLMS/app/controllers/learn_worlds_integration/webhooks_controller.rb
@@ -6,7 +6,7 @@ module LearnWorldsIntegration
     class InvalidSignatureError < Error; end
     class UnknownEventTypeError < Error; end
 
-    EVENT_TYPE_HANDLER_MAPPING = {
+    HANDLER_MAPPING = {
       'enrolledFreeCourse' => LearnWorldsIntegration::Webhooks::EnrolledFreeCourseEventHandler,
       'courseCompleted' => LearnWorldsIntegration::Webhooks::CourseCompletedEventHandler,
       'awardedCertificate' => LearnWorldsIntegration::Webhooks::EarnedCertificateEventHandler
@@ -34,7 +34,7 @@ module LearnWorldsIntegration
     end
 
     private def payload
-      @payload ||= JSON.parse(request.body.read)
+      @payload ||= JSON.parse(request.body.string)
     end
 
     private def signature_header
--- a/services/QuillLMS/app/controllers/learn_worlds_integration/webhooks_controller.rb
+++ b/services/QuillLMS/app/controllers/learn_worlds_integration/webhooks_controller.rb
@@ -6,6 +6,12 @@ module LearnWorldsIntegration
     class InvalidSignatureError < Error; end
     class UnknownEventTypeError < Error; end
 
+    EVENT_TYPE_HANDLER_MAPPING = {
+      'enrolledFreeCourse' => LearnWorldsIntegration::Webhooks::EnrolledFreeCourseEventHandler,
+      'courseCompleted' => LearnWorldsIntegration::Webhooks::CourseCompletedEventHandler,
+      'awardedCertificate' => LearnWorldsIntegration::Webhooks::EarnedCertificateEventHandler
+    }.freeze
+
     SIGNATURE_HEADER_KEY = 'HTTP_LEARNWORLDS_WEBHOOK_SIGNATURE'
 
     protect_from_forgery except: :create
@@ -35,17 +41,16 @@ module LearnWorldsIntegration
       request.env[SIGNATURE_HEADER_KEY]
     end
 
+    private def valid_signature?
+      signature_header.split('=').last == Auth::LearnWorlds::WEBHOOK_SIGNATURE
+    end
+
     private def verify_signature
-      raise InvalidSignatureError unless signature_header.split('=').last == Auth::LearnWorlds::WEBHOOK_SIGNATURE
+      raise InvalidSignatureError unless valid_signature?
     end
 
     private def webhook_handler_class
-      case event_type
-      when 'enrolledFreeCourse' then Webhooks::EnrolledFreeCourseEventHandler
-      when 'courseCompleted' then Webhooks::CourseCompletedEventHandler
-      when 'awardedCertificate' then Webhooks::EarnedCertificateEventHandler
-      else raise UnknownEventTypeError, "Unknown event type: #{event_type}"
-      end
+      HANDLER_MAPPING.fetch(event_type) { raise UnknownEventTypeError, "Unknown event type: #{event_type}" }
     end
   end
 end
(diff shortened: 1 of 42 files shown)


services/QuillLMS/app/controllers/learn_worlds_integration/webhooks_controller.rb
```diff
@@ -0,0 +1,51 @@
...
+
+    private def signature_header
+      request.env[SIGNATURE_HEADER_KEY]
+    end
+
+    private def verify_signature
+      raise InvalidSignatureError unless signature_header.split('=').last == Auth::LearnWorlds::WEBHOOK_SIGNATURE
+    end
+
+    private def webhook_handler_class
```
dandrabik, line 52: Consider separating the config, the lookup, and the error by doing something like this.

```ruby
private def webhook_handler_class
  HANDLER_MAPPING.fetch(event_type)

rescue KeyError
  raise UnknownEventTypeError, "Unknown event type: #{event_type}"
end

# you can also do this:
HANDLER_MAPPING.fetch(event_type) { raise UnknownEventTypeError, "Unknown event type: #{event_type}" }

File: services/QuillLMS/app/controllers/learn_worlds_integration/webhooks_controller.rb [hunk 2]
```diff
@@ -0,0 +1,51 @@
...
+    private def payload
+      @payload ||= JSON.parse(request.body.read)
+    end
+
+    private def signature_header
+      request.env[SIGNATURE_HEADER_KEY]
+    end
+
+    private def verify_signature
+      raise InvalidSignatureError unless signature_header.split('=').last == Auth::LearnWorlds::WEBHOOK_SIGNATURE
```
dandrabik: Consider making a `signature_valid?` method for readability
```ruby
def signature_valid?
  signature_header.split('=').last == Auth::LearnWorlds::WEBHOOK_SIGNATURE
end

**Review by dandrabik — APPROVED** [review]
This looks great and a very robust feature-set. Thanks for putting this together quickly. I left a few take-or-leave code comments, but looks good to me! 🚢

File: services/QuillLMS/app/controllers/learn_worlds_integration/webhooks_controller.rb [hunk 3]
```diff
@@ -0,0 +1,56 @@
...
+    private def event_type
+      payload['type']
+    end
+
+    private def data
+      payload['data']
+    end
+
+    private def payload
+      @payload ||= JSON.parse(request.body.read)
```
dandrabik: One note here. `request.body.read` flushes the body of the request. If something queries the body after this, it will be empty.

If that's not ok or not the intention, use `req.body.string`

brendanshean: That's a good catch.  I was a little uneasy about the StringIO object but didn't realize there was something simpler "string"

services/QuillLMS/app/models/user.rb
```diff
@@ -109,6 +109,8 @@ class User < ApplicationRecord
 
   SCHOOL_CHANGELOG_ATTRIBUTE = 'school_id'
 
+  LEARN_WORLDS_ACCESS_OVERRIDE = 'learn_worlds_access_override'
```
anathomical: Should this constant be stored in the `AppSetting` model instead of the `User` model?  I don't have strong feelings either way, but was curious.

services/QuillLMS/spec/controllers/learn_worlds_integration/webhooks_controller_spec.rb
```diff
@@ -0,0 +1,70 @@
...
+
+    context 'invalid signature' do
+      let(:params) { {} }
+      let(:signature_header) { 'v2=invalid-signature' }
+      let(:error) { LearnWorldsIntegration::WebhooksController::InvalidSignatureError }
+
+      it 'handles the signature error and reports to new relic' do
+        expect(ErrorNotifier).to receive(:report).with(error)
+        subject
+        expect(response.status).to eq 400
```
anathomical, line 27: This is kind of nit-picky, but is an invalid signature a failure to authenticate?  Should this be a 403?

services/QuillLMS/app/services/vitally_integration/serialize_vitally_sales_user.rb
```diff
@@ -100,7 +105,14 @@ def data
         email_verification_status: @user.email_verification_status,
         admin_approval_status: @user.admin_approval_status,
         number_of_schools_administered: number_of_schools_administered,
-        number_of_districts_administered: number_of_districts_administered
+        number_of_districts_administered: number_of_districts_administered,
+        learn_worlds_num_enrolled_courses: learn_worlds_enrolled_courses&.count,
+        learn_worlds_enrolled_courses: learn_worlds_enrolled_courses&.titles_string,
```
anathomical, line 110: Do we have a sense of how big the list of enrolled courses might eventually get?  I imagine that there's some size at which passing a giant list of strings to Vitally is a bad idea.  Less than 10 should be fine, but how close to 100 do we want to allow this to comfortably get?

anathomical (approved):
Looks good to me.  A couple of small items for consideration, but the overall approach is solid.  I especially like the way you've set up the associations to simplify access to data in the User object.

services/QuillLMS/app/models/user.rb
brendanshean: Good call!

services/QuillLMS/app/services/vitally_integration/serialize_vitally_sales_user.rb
brendanshean: That's a good question.  I was under the impression this would be under 10.

services/QuillLMS/spec/controllers/learn_worlds_integration/webhooks_controller_spec.rb
brendanshean: I reused the stripe [structure](https://github.com/empirical-org/Empirical-Core/blob/develop/services/QuillLMS/app/controllers/stripe_integration/webhooks_controller.rb#L16) which was the code recommended in the stripe [docs](https://stripe.com/docs/payments/handling-payment-events).  Since the rescue clause is kind of bucket for any standard error I think it's more about sending a message back to the webhook that something failed.  While I agree with you point about it technically being the wrong code, I think for simplicity I'd rather just have the one clause for now.

services/QuillLMS/app/services/vitally_integration/serialize_vitally_sales_user.rb
brendanshean: I'll set a reminder to check every few months how many courses are in Quill Academy.

PR #10597, opened 2023-05-27
Make evidence factories visible in LMS for testing

cissyxyu:
## WHAT
Tweak configurations to make all evidence engine factories (Activity, Prompt, Rule, Feedback, etc.) usable in LMS specs. 

## WHY
This makes it much easier for us to write tests in the LMS because we no longer have to manually create objects (with a lot of manually configured required fields) when we want to test things related to the Evidence Engine.

## HOW
Tweak the configs in the engine to add the Evidence Engine factories to the FactoryBot directory whenever the Evidence Engine is loaded. Previously, this code was only running inside the spec setup code for Evidence Engine. 
Then change the code inside all previously written LMS specs to start using the evidence engine factories instead of manually creating records. 

### Screenshots

### Notion Card Links
https://www.notion.so/quill/Import-factories-from-engine-to-LMS-08495475acd54cfb979f1bc987ab2f0d?pvs=4

PR Checklist | Your Answer
------------ | -------------

### DIFF
--- a/services/QuillLMS/engines/evidence/lib/evidence/engine.rb
+++ b/services/QuillLMS/engines/evidence/lib/evidence/engine.rb
@@ -1,7 +1,7 @@
 # frozen_string_literal: true
 
 # require_relative '../quill_scaffold_controller'
-require 'factory_bot_rails'
+require 'factory_bot_rails' unless Rails.env.production?
 
 module Evidence
   class Engine < ::Rails::Engine
@@ -23,8 +23,10 @@ module Evidence
       g.fallbacks[:shoulda] = :test_unit
     end
 
-    initializer "evidence.factories", after: "factory_bot.set_factory_paths" do
-      FactoryBot.definition_file_paths << File.expand_path('../../../spec/factories/evidence', __FILE__) if defined?(FactoryBot)
+    unless Rails.env.production?
+      initializer "evidence.factories", after: "factory_bot.set_factory_paths" do
+        FactoryBot.definition_file_paths << File.expand_path('../../../spec/factories/evidence', __FILE__)
+      end
     end
   end
 end
--- a/services/QuillLMS/engines/evidence/lib/evidence/engine.rb
+++ b/services/QuillLMS/engines/evidence/lib/evidence/engine.rb
@@ -1,6 +1,8 @@
 # frozen_string_literal: true
 
 # require_relative '../quill_scaffold_controller'
+require 'factory_bot_rails'
+
 module Evidence
   class Engine < ::Rails::Engine
     config.eager_load_paths += %W{#{config.root}/lib/evidence}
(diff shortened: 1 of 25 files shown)


services/QuillLMS/engines/evidence/lib/evidence/engine.rb
```diff
@@ -20,5 +22,9 @@ class Engine < ::Rails::Engine
 
       g.fallbacks[:shoulda] = :test_unit
     end
+
+    initializer "evidence.factories", after: "factory_bot.set_factory_paths" do
```
brendanshean: Should this initializer block be wrapped in a Rails.env.test? conditional to prevent it loading in ~~development and~~ production.

brendanshean (approved):
Great idea.  Thanks for configuring this!

anathomical (approved):
Very nice bit of clean up.

services/QuillLMS/spec/factories/activities.rb
```diff
@@ -87,7 +87,7 @@
       end
     end
 
-    factory :evidence_activity do
+    factory :evidence_lms_activity do
```
happythenewsad, line 90: I believe this is now a redundant factory. In a future PR we should dedupe and clean this up, so that specs only use `evidence_activity`, not `evidence_lms_activity`

happythenewsad (approved):
Nice work!

services/QuillLMS/engines/evidence/lib/evidence/engine.rb
brendanshean: Actually, probably 

```
unless Rails.env.production?
  initializer "evidence.factories", after: "factory_bot.set_factory_paths" do
          FactoryBot.definition_file_paths << File.expand_path('../../../spec/factories/evidence', __FILE__) if defined?(FactoryBot)
  end
end
```
 
I feel like the defined?(FactoryBot) might be redundant given the require 'factory_bot_rails' at the top of the file.

services/QuillLMS/engines/evidence/lib/evidence/engine.rb
```diff
@@ -1,6 +1,8 @@
 # frozen_string_literal: true
 
 # require_relative '../quill_scaffold_controller'
+require 'factory_bot_rails'
```
brendanshean: Since this is development_dependency we probably want guard `unless Rails.env.production?`

services/QuillLMS/spec/factories/activities.rb
brendanshean: I also, did a double take on this.  I think the `evidence_lms_activity` is the Activity model in the LMS whereas `evidence_activity` is Evidence::Activity model.

While this seems correct, I'm wondering if there's better naming convention to avoid this confusion in the future.

PR #10603, opened 2023-05-31
store feedback session csv on disk instead of in memory

emilia-friedberg:
## WHAT
Update feedback session email worker to store the CSV on disk, instead of in memory, hopefully eliminating some weird errors we've been getting from `UserMailer` when trying to send these emails.

## WHY
We want these emails to go through successfully, but we're getting [EPIPE](https://quillorg-5s.sentry.io/issues/4218923083/?project=11238&query=is%3Aunresolved&referrer=issue-stream&stream_index=0) and [ECONNRESET](https://quillorg-5s.sentry.io/issues/4218931446/?project=11238&query=is%3Aunresolved&referrer=issue-stream&stream_index=2) errors. My hypothesis is that the CSV data we're attempting to pass is just too large, because the older version does work locally (where I have much smaller amounts of data), and that if the mailer doesn't have to keep all the csv data in memory it'll stop having these outages.

## HOW
Just update the worker to write to a file instead of passing the CSV directly, and the mailer attaches that file rather than what it gets passed.

### Screenshots

### Notion Card Links
https://www.notion.so/quill/Email-Me-CSV-Data-button-not-working-on-View-Sessions-page-86ad7b0514a64c32beb37cee28f4ca41

PR Checklist | Your Answer
------------ | -------------

### DIFF
--- a/services/QuillLMS/app/mailers/user_mailer.rb
+++ b/services/QuillLMS/app/mailers/user_mailer.rb
@@ -164,8 +164,8 @@ class UserMailer < ActionMailer::Base
     mail from: "The Quill Team <hello@quill.org>", to: email, subject: "ELL Starter Diagnostic Next Steps"
   end
 
-  def feedback_history_session_csv_download(email, csv)
-    attachments[FEEDBACK_SESSIONS_CSV_FILENAME] = {mime_type: 'text/csv', content: csv}
+  def feedback_history_session_csv_download(email, csv_file_path)
+    attachments[FEEDBACK_SESSIONS_CSV_FILENAME] = File.read(csv_file_path)
     mail from: "The Quill Team <hello@quill.org>", to: email, subject: FEEDBACK_SESSIONS_CSV_DOWNLOAD
   end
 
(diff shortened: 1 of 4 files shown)


happythenewsad (commented):
~I think you need to make the write/read synchronous (executed by a single dyno, at the same time), because of the [non-thread-safe and ephemeral nature](https://devcenter.heroku.com/articles/dynos#ephemeral-filesystem) of the Heroku filesystem.~

I see that the Worker calls the mailer, so it should work as-is.

anathomical (commented):
I mentioned this in a private message on Slack, but wanted to flag it publicly: I suspect that this might not work in production because the disc that the workers will write to and the disc that the web servers read from is not the same.  It'd work fine in local testing, but I'd like to see confirmation in a personal staging environment, which is already in progress.

emilia-friedberg:
> I mentioned this in a private message on Slack, but wanted to flag it publicly: I suspect that this might not work in production because the disc that the workers will write to and the disc that the web servers read from is not the same. It'd work fine in local testing, but I'd like to see confirmation in a personal staging environment, which is already in progress.

@anathomical checked and it does indeed work in my staging environment!

anathomical (approved):
One minor improvement that probably won't matter, but could in theory: after the mailer sends the file, we might want to delete it from the disk.  I could imagine a bunch of requests for this email landing on the same worker and taking up all the disk space.  Since we won't need the file after it's sent, we might as well just delete it.

anathomical (approved):
One minor improvement that probably won't matter, but could in theory: after the mailer sends the file, we might want to delete it from the disk.  I could imagine a bunch of requests for this email landing on the same worker and taking up all the disk space.  Since we won't need the file after it's sent, we might as well just delete it.

PR #10605, opened 2023-06-01
Add testing for snapshots queries on BigQuery 

brendanshean:
## WHAT
Add testing for snapshot queries on BigQuery using CTEs.

## WHY
Tests will give us transparency but also confidence that the queries are doing what we expect them to be doing.

## HOW

The QuillBigQuery::TestRunner [class](https://github.com/empirical-org/Empirical-Core/pull/10605/files#diff-96d2459e3ee946dd97d582a62b96aa844aa76cb61b3fba2e2afaf4df37fc722eR7) takes in some records, creates a CTE from those records, and then runs the query on that CTE.  It takes a couple of arguments: 
1) `query`:  populated with `described_class.new(*params).final_query` in this [file](https://github.com/empirical-org/Empirical-Core/blob/cea5d1a01c5dc7552eb0d84ccee3de8a37739649/services/QuillLMS/spec/support/shared_contexts/snapshots/snapshots_cte.rb#L7)
2) `cte_records`:  populated from the parent_class' shared_context cte which is one of:
  * snapshots_period_cte 
  * snapshots_count_cte 
  * snapshots_option_cte 

One can override this by specifying a different `cte_records`. (e.g. we override [here](https://github.com/empirical-org/Empirical-Core/blob/e25b1d95dffa531d1c9e65f641f00de97da3c3bb/services/QuillLMS/spec/queries/snapshots/sentences_written_query_spec.rb#L11) because the query has additional tables in the [from_and_join_clauses](https://github.com/empirical-org/Empirical-Core/blob/e25b1d95dffa531d1c9e65f641f00de97da3c3bb/services/QuillLMS/app/queries/snapshots/sentences_written_query.rb#L20))

### Screenshots

### Notion Card Links

PR Checklist | Your Answer
------------ | -------------
Have you added and/or updated tests? |  No.  Just tests

### DIFF
--- a/services/QuillLMS/app/queries/snapshots/activities_assigned_query.rb
+++ b/services/QuillLMS/app/queries/snapshots/activities_assigned_query.rb
@@ -2,7 +2,7 @@
 
 module Snapshots
   class ActivitiesAssignedQuery < CountQuery
-    def query_to_run
+    def final_query
       utilizing_subquery
     end
 
@@ -12,7 +12,7 @@ module Snapshots
           FROM (
             SELECT (sub_sub_query.students_assigned_count *
               COUNT(unit_activities.unit_id)) AS activities_assigned
-            FROM (#{query}) AS sub_sub_query
+            FROM (#{core_query}) AS sub_sub_query
             JOIN lms.unit_activities
               ON sub_sub_query.unit_id = unit_activities.unit_id
             GROUP BY sub_sub_query.id,
--- a/services/QuillLMS/app/queries/snapshots/activities_assigned_query.rb
+++ b/services/QuillLMS/app/queries/snapshots/activities_assigned_query.rb
@@ -2,8 +2,8 @@
 
 module Snapshots
   class ActivitiesAssignedQuery < CountQuery
-    def run_query
-      QuillBigQuery::Runner.execute(utilizing_subquery)
+    def query_to_run
+      utilizing_subquery
     end
 
     def utilizing_subquery
--- a/services/QuillLMS/app/queries/snapshots/activities_assigned_query.rb
+++ b/services/QuillLMS/app/queries/snapshots/activities_assigned_query.rb
@@ -2,17 +2,12 @@
 
 module Snapshots
   class ActivitiesAssignedQuery < CountQuery
-    def final_query
-      utilizing_subquery
-    end
-
-    def utilizing_subquery
+    def query
       <<-SQL
         SELECT SUM(sub_query.activities_assigned) AS count
           FROM (
-            SELECT (sub_sub_query.students_assigned_count *
-              COUNT(unit_activities.unit_id)) AS activities_assigned
-            FROM (#{core_query}) AS sub_sub_query
+            SELECT (sub_sub_query.students_assigned_count * COUNT(unit_activities.unit_id)) AS activities_assigned
+            FROM (#{super}) AS sub_sub_query
             JOIN lms.unit_activities
               ON sub_sub_query.unit_id = unit_activities.unit_id
             GROUP BY sub_sub_query.id,
--- a/services/QuillLMS/app/queries/snapshots/activities_assigned_query.rb
+++ b/services/QuillLMS/app/queries/snapshots/activities_assigned_query.rb
@@ -23,7 +23,12 @@ module Snapshots
     end
 
     def select_clause
-      "SELECT DISTINCT classroom_units.id, classroom_units.unit_id, array_length(classroom_units.assigned_student_ids) AS students_assigned_count"
+      <<-SQL
+        SELECT
+          DISTINCT classroom_units.id,
+          classroom_units.unit_id,
+          array_length(classroom_units.assigned_student_ids) AS students_assigned_count
+      SQL
     end
 
     def relevant_date_column
--- a/services/QuillLMS/spec/rails_helper.rb
+++ b/services/QuillLMS/spec/rails_helper.rb
@@ -96,11 +96,8 @@ RSpec.configure do |config|
     ActionController::Base.perform_caching = caching
   end
 
-  config.around(:each, :external_api) do |example|
-    VCR.configure { |c| c.allow_http_connections_when_no_cassette = true }
-    example.run
-    VCR.configure { |c| c.allow_http_connections_when_no_cassette = false }
-  end
+  config.around(:each, :external_api) { |example| with_vcr_disabled { example.run } }
+  config.around(:each, :big_query_snapshot) { |example| with_vcr_disabled { example.run } }
 
   if ENV.fetch('SUPPRESS_PUTS', false) == 'true'
     config.before do
@@ -110,6 +107,8 @@ RSpec.configure do |config|
   end
 end
 
-def vcr_ignores_localhost
-  VCR.configuration.ignore_localhost = true
+private def with_vcr_disabled
+  VCR.configure { |c| c.allow_http_connections_when_no_cassette = true }
+  yield
+  VCR.configure { |c| c.allow_http_connections_when_no_cassette = false }
 end
--- a/services/QuillLMS/spec/support/shared_contexts/snapshots/snapshots_cte.rb
+++ b/services/QuillLMS/spec/support/shared_contexts/snapshots/snapshots_cte.rb
@@ -40,7 +40,7 @@ RSpec.shared_context 'Snapshots CTE' do
     attr_type = record.class.column_for_attribute(attr).type
 
     if value.nil?
-      'NULL'
+      "''"
     elsif value.is_a?(Array)
       value.map { |v| attr_type_value(attr_type, v) }
     else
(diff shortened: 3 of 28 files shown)


services/QuillLMS/spec/support/shared_contexts/snapshots/snapshots_cte.rb
```diff
@@ -0,0 +1,60 @@
...
+    if value.nil?
+      "''"
+    elsif value.is_a?(Array)
+      value.map { |v| attr_type_value(attr_type, v) }
+    else
+      attr_type_value(attr_type, value)
+    end
+  end
+
+  def attr_type_value(attr_type, value)
```
brendanshean: There will probably be new types failures as new queries are created.  You'll just need to update them in here:

Screenshot 2023-06-01 at 7 40 17 AM

services/QuillLMS/spec/rails_helper.rb
```diff
@@ -96,11 +96,8 @@
     ActionController::Base.perform_caching = caching
   end
 
-  config.around(:each, :external_api) do |example|
-    VCR.configure { |c| c.allow_http_connections_when_no_cassette = true }
-    example.run
-    VCR.configure { |c| c.allow_http_connections_when_no_cassette = false }
-  end
+  config.around(:each, :external_api) { |example| with_vcr_disabled { example.run } }
+  config.around(:each, :big_query_snapshot) { |example| with_vcr_disabled { example.run } }
```
happythenewsad, line 100: Nice refactor

happythenewsad (commented):
I'm a little concerned that the snapshot classes now have the following accessible methods, which is bad for future intelligibility:

```
self.run
run
query
run_query
query_to_run
```

How about renaming to something like this?

```
query -> core query
query_to_run -> final/composed_query
run_query -> nil
```

happythenewsad (commented):
I'm a little concerned that the snapshot classes now have the following accessible methods, which is bad for future intelligibility:

```
self.run
run
query
run_query
query_to_run
```

How about renaming to something like this?

```
query -> core_query
query_to_run -> final/composed_query
run_query -> nil
```

brendanshean:
> I'm a little concerned that the snapshot classes now have the following accessible methods, which is bad for future intelligibility:
> 
> ```
> self.run
> run
> query
> run_query
> query_to_run
> ```
> 
> How about renaming to something like this?
> 
> ```
> query -> core_query
> query_to_run -> final/composed_query
> run_query -> nil
> ```

@happythenewsad To be honest, I wasn't really happy with the `query_to_run` name.  I think your suggestion is much clearer.  @anathomical Any objections with this rename going forward? (I would opt for final_query)

services/QuillLMS/spec/support/shared_contexts/snapshots/snapshots_cte.rb
```diff
@@ -0,0 +1,60 @@
...
+    <<-SQL
+      WITH
+        #{cte}
+      #{query}
+    SQL
+  end
+
+  let(:results) { QuillBigQuery::Runner.execute(bq_query) }
+
+  def cte
```
anathomical: Is there a utility file we can stash these functions in?  And then maybe replace the `let(:bq_query)` block with a function call?  Something like `big_query_with_context(query, array_of_relevant_models)`?  Then you can do `let(:query) { big_query_with_context(described_class.new(params).query_to_run, [classrooms, users, ...]) }`  This will make it easier to utilize this outside of the Snapshots shared context (such as progress reports).

services/QuillLMS/spec/support/shared_contexts/snapshots/snapshots_cte.rb
```diff
@@ -0,0 +1,60 @@
...
+    cte_table_collections
+      .reject { |collection| collection.empty? }
+      .map { |collection| "#{cte_alias(collection)} AS ( #{cte_query(collection)} )" }
+      .join(",\n\t")
```
anathomical: This is a pure style thing, but I think that this becomes safer, and maybe slightly more useful by extension, if we don't make strong assumptions that each collection is complete, or that there aren't multiple collections of the same model type.  Imagine that `cte_table_collections` is `[[classroom1, classroom2], [school], [edge_case_classroom]]`.  I think that you can handle that with Ruby's `group_by`:
```ruby
cte_table_collections
  .flatten
  .group_by { |record| record.class.table_name }
  .map do |table_name, collection|
    "#{table_name} AS ( #{collection.map { |record| [:SELECT, cte_select_clause(record)].join(' ') } } )"
  end.join(" UNION ALL \n")
```
Only, you know, cleaner than that.  I think that being able to be relatively imprecise with what goes into the `cte_table_collections` (or whatever we end up calling it) value will empower later developers to get simpler value out of this toolchain.

anathomical (approved):
I like this a lot.  I especially appreciate that it manages migrations (since it builds the structure it puts into context from the PostgreSQL table structure).  I've got one comment on code organization, and one comment on expected inputs and implementation, but overall this is awesome.

anathomical (approved):
I like this a lot.  I especially appreciate that it manages migrations (since it builds the structure it puts into context from the PostgreSQL table structure).  I've got one comment on code organization, and one comment on expected inputs and implementation, but overall this is awesome.

anathomical:
> @happythenewsad To be honest, I wasn't really happy with the `query_to_run` name. I think your suggestion is much clearer. @anathomical Any objections with this rename going forward? (I would opt for final_query)

No objections from me.  I like `final_query` or something like it.

services/QuillLMS/spec/support/shared_contexts/snapshots/snapshots_cte.rb
brendanshean: Good idea.  I will update.

services/QuillLMS/spec/support/shared_contexts/snapshots/snapshots_cte.rb
brendanshean: I think this is a better design and smart idea with group_by.  I think this framework much easier to add records in.

services/QuillLMS/app/queries/snapshots/activities_assigned_query.rb
```diff
@@ -2,8 +2,8 @@
 
 module Snapshots
   class ActivitiesAssignedQuery < CountQuery
-    def run_query
-      QuillBigQuery::Runner.execute(utilizing_subquery)
+    def final_query
```
dandrabik: Take or leave (and I'm on the fence on this), you could consider having a standard `query` method and then overwriting it for this class and use `super` where appropriate. Would eliminate the need for a separate `final_query`, `core_query`, etc. vars, and follows the pattern used other places in these files. e.g. 
```ruby
def query
  <<-SQL
        SELECT COUNT(*) AS count
          FROM (#{super})
      SQL
end

**Review by happythenewsad — APPROVED** [review]
LGTM

File: services/QuillLMS/spec/support/shared_contexts/snapshots/snapshots_cte.rb [hunk 6]
```diff
@@ -0,0 +1,8 @@
+# frozen_string_literal: true
+
+RSpec.shared_context 'Snapshots CTE' do
+  # Create before running the query, otherwise, some parameters (e.g. school_ids) will be nil
+  before { cte_records }
+
+  let(:results) { QuillBigQuery::TestRunner.run(described_class.new(*params).final_query, cte_records) }
```
dandrabik: I wonder if it would be better to restructure this a bit to still call `run`, but maybe making the runner a variable that we can stub in testing. This might make it easier to stub this out in other tests, and it adds a clean seam for testing. e.g.
```ruby
let(:runner) { QuillBigQuery::TestRunner.new(cte_records) }
let(:results) {described_class.run(*params, runner: runner) }

# Query class
def new(params, runner: QuillBigQuery::Runner)
  # initialization
end

def run
  runner.execute(query)
end

dandrabik (approved):
Overall this approach looks really good and this will be a valuable part of our testing structure.

I left one comment about a slightly different approach that I think is worth considering (adding a seam in the code to inject the `TestRunner`, so we can still call `SomeQuery.run`). Happy to chat about that if my comment isn't clear.

services/QuillLMS/spec/support/shared_contexts/snapshots/snapshots_cte.rb
dandrabik: My concern with the current setup is we are stubbing everything around a specific sub-method, and not using the path of the code used by the application.

services/QuillLMS/app/queries/snapshots/activities_assigned_query.rb
brendanshean: Much simpler.  Good idea

dandrabik (approved):
:+1: This looks great. I like this approach with the new Query base class, cuts down on boilerplate. Well done.

PR #10607, opened 2023-06-01
Lms/admin snapshot report backend milestone2

anathomical:
## WHAT
Add additional queries to the Admin snapshots backend
## WHY
In addition to simple counts, we also want "Top X" queries that find lists of things like the "Most Active Schools".
## HOW
- Add a new class of `Snapshot` query for calculating `TopX` data sets
- Write three specific `TopX` queries
- Add a new worker to cache `TopX` queries
- Add a new controller action to access `TopX` query data

### Notion Card Links
https://www.notion.so/quill/Technical-Review-Usage-Snapshot-Admin-Report-afa1f08d50524c81a758d32899f97ae4

PR Checklist | Your Answer
------------ | -------------


services/QuillLMS/app/controllers/snapshots_controller.rb
```diff
@@ -77,6 +60,32 @@ def options
     return render json: { error: 'user is not authorized for all specified schools' }, status: 403
   end
 
+  private def retrieve_cache_or_enqueue_worker(worker)
```
anathomical, line 75: Not totally sure about this abstraction as it sort of obfuscates what's going on in the actual actions, but it felt better than repeating all the code again.  Open to suggestions on approach tweaks.

services/QuillLMS/app/workers/snapshots/cache_snapshot_top_x_worker.rb
```diff
@@ -0,0 +1,40 @@
+# frozen_string_literal: true
+
+module Snapshots
+  class CacheSnapshotTopXWorker
```
anathomical, line 4: A ton of this code is just a copy of the `CacheSnapshotCountWorker`, and the duplication makes me a little uneasy.  But I figured doing it twice might be fine?  We also don't have a real history of inherited class hierarchies in worker code, so I was hesitant to refactor into a parent with two children.  Again, super open to suggestions.

happythenewsad: The `perform` method is nearly identical, so we could abstract that into a parent Worker class

services/QuillLMS/app/controllers/snapshots_controller.rb
```diff
@@ -77,6 +60,32 @@ def options
...
+    return { error: 'unrecognized query type for this endpoint' } unless worker::QUERIES.keys.include?(snapshot_params[:query])
```
happythenewsad: This is an HTTP 200, by default. I think we want this error to be less silent - perhaps use `ErrorNotifier`?

Alternatively you could fire off an error from the client, but pushing logic from server to client seems unnecessary.

happythenewsad (commented):
The new `top_x_query` classes are almost completely declarative, which indicates you've picked a good point of abstraction.

services/QuillLMS/app/controllers/snapshots_controller.rb
brendanshean: I agree with the point about using ErrorNotifier

services/QuillLMS/app/workers/snapshots/cache_snapshot_top_x_worker.rb
brendanshean: I usually wait to abstract classes until I have three examples.  Two can be premature.

services/QuillLMS/app/controllers/snapshots_controller.rb
brendanshean: I think the abstraction makes sense.  We kind of do something similar with google classrooms [here](https://github.com/empirical-org/Empirical-Core/blob/develop/services/QuillLMS/app/controllers/teachers/classroom_manager_controller.rb#L202)

The timeframe makes the data model a little more complicated, but I don't see any easy way of simplifying it.

services/QuillLMS/app/controllers/snapshots_controller.rb
```diff
@@ -77,6 +60,32 @@ def options
...
+      },
+      snapshot_params[:school_ids],
+      snapshot_params[:grades])
+
+    { message: 'Generating snapshot' }
+  end
+
   private def cache_key_for_timeframe(previous_start, current_start, current_end)
 
     Snapshots::CacheKeys.generate_key(snapshots_params[:query],
```
brendanshean: Is `snapshots_params[:query]` a typo?  If so, I would suggest maybe creating a new method:

```
private def query
  snapshot_params[:query]
end
```

since it's used in a few places in this controller.

services/QuillLMS/spec/controllers/snapshots_controller_spec.rb
```diff
@@ -9,72 +9,88 @@
...
       payload = {
         "current" => "CURRENT",
         "previous" => "PREVIOUS"
       }
 
-      expect(Rails.cache).to receive(:read).with(cache_key).and_return(payload)
+      expect(Rails.cache).to receive(:read).exactly(controller_actions.length).times.with(cache_key).and_return(payload)
 
-      get :count, params: { query: query_name, timeframe: timeframe_name, school_ids: school_ids }
+      controller_actions.each do |action, query_name|
```
brendanshean, line 41: Interesting I haven't seen this construction before.

brendanshean (approved):
Nice work on this.  I appreciate the thorough testing.

services/QuillLMS/spec/controllers/snapshots_controller_spec.rb
brendanshean: It does make me wonder if these should be the same endpoint but I think that would be premature.

services/QuillLMS/app/controllers/snapshots_controller.rb
anathomical: I like the idea of returning some non-200 status code here.  It'll make it easier for the front-end to know something went wrong.  Let me see if I can figure out how to make that work.

services/QuillLMS/spec/controllers/snapshots_controller_spec.rb
anathomical: I'm a little uncertain about all of this.  My thought was:
1. The payload shapes returned by these two snapshot types is different, so I figured it made sense for them to be two different actions.
2. But other than response shape, they basically are the same, so I didn't want to duplicate a bunch of code.

I could easily be convinced to do something different here, but it was the awkward compromise I settled on.

services/QuillLMS/app/queries/snapshots/top_x_query.rb
```diff
@@ -0,0 +1,41 @@
+# frozen_string_literal: true
+
+module Snapshots
+  class TopXQuery < PeriodQuery
+    NUMBER_OF_RECORDS = 3
```
emilia-friedberg: Just want to flag that we actually need 10 for the frontend queries, since all the three-line tables expand into ten-line modals (https://www.notion.so/quill/Technical-Review-Usage-Snapshot-Admin-Report-afa1f08d50524c81a758d32899f97ae4?pvs=4#778b83aa8cff488a936aa7ecf8edde3b). Alternately, we could pass the number of records and make two requests, but that seems like it is probably more expensive than just requesting ten once.

services/QuillLMS/app/queries/snapshots/top_concepts_assigned_query.rb
```diff
@@ -0,0 +1,56 @@
+# frozen_string_literal: true
+
+module Snapshots
+  class TopConceptsAssignedQuery < TopXQuery
+    def run_query
+      QuillBigQuery::Runner.execute(utilizing_subquery)
+    end
+
+    def utilizing_subquery
```
dandrabik: I think I wrote this on Brendan's testing PR, but you are overwriting the parent method for other methods, so you should consider doing that here as well to cut down on needing new methods, e.g.
```ruby
def query
 <<-SQL
      SELECT value, SUM(count) AS count
        FROM (#{super})
        GROUP BY value
        ORDER BY count DESC
        LIMIT #{NUMBER_OF_RECORDS}
      SQL
end

File: services/QuillLMS/app/workers/snapshots/cache_snapshot_top_x_worker.rb [hunk 8]
```diff
@@ -0,0 +1,40 @@
...
+          query: query,
+          timeframe: timeframe[:name],
+          school_ids: school_ids,
+          grades: grades
+        }
+      )
+    end
+
+    private def generate_payload(query, timeframe, school_ids, grades)
+      timeframe_start = timeframe[:current_start]
```
dandrabik: not sure you gain much from assigning these local vars over just passing `timeframe[:current_start]` directly into the method below.

dandrabik (approved):
Looks good to me. I left a few small style items, but all are take or leave. Nice continued work on this!

services/QuillLMS/spec/queries/snapshots/most_active_schools_query_spec.rb
```diff
@@ -0,0 +1,114 @@
...
+      }
+
+      it 'should include the top 10 results if they are present' do
+        runner = QuillBigQuery::TestRunner.new([
+          runner_context,
+          activity_sessions
+        ])
+        result = described_class.run(timeframe_start, timeframe_end, school_ids, grades, runner: runner)
+
+        expect(result).to eq([
```
dandrabik: I think it would be good to dry this up, something like this e.g.
```ruby
let(:expected_result) { (0..schools.length).map {|i| {"value"=>schools[i].name, "count"=>activity_sessions[i].length } }

expect(result).to eq(expected_result)

File: services/QuillLMS/spec/queries/snapshots/top_concepts_practiced_query_spec.rb [hunk 10]
```diff
@@ -0,0 +1,86 @@
...
+
+      # We have one activity connected to each concept.
+      # We want to create an activity_session for each activity, then a second activity_session for all but one activity, and so on.  The outcome we desire is a unique number of activity_sessions tied to each activity
+      let(:activity_session_bundles) do
+        activities.map.with_index do |activity, activity_index|
+          create_list(:activity_session, num_concepts - activity_index, activity: activity, classroom_unit: classroom_units.first)
+        end
+      end
+
+      let(:runner_context) {
```
dandrabik, line 26: So since you are including the [TestRunner shared_context](https://github.com/empirical-org/Empirical-Core/blob/develop/services/QuillLMS/spec/support/shared_contexts/quill_big_query_test_runner_setup.rb) (by proxy), you can standardize this all a bit by defining `cte_records` and `params` and the rest of the setup is done for you, e.g.

```ruby
let(:cte_records) { [runner_context, activity_session_bundles] }
# This is not needed since it's in the shared_context
# let(:params) { [timeframe_start, timeframe_end, school_ids, grades] }

# DRY up expected_results to a var.
expect(results).to eq(expected_results)

File: services/QuillLMS/spec/queries/snapshots/top_concepts_practiced_query_spec.rb [hunk 11]
```diff
@@ -0,0 +1,86 @@
...
+          {"count"=>activity_session_bundles[9].length, "value"=>concepts[9].name}
+        ])
+      end
+
+      it 'should not include data from the 11th most common concept' do
+        runner = QuillBigQuery::TestRunner.new([
+          runner_context,
+          activity_session_bundles
+        ])
+        result = described_class.run(timeframe_start, timeframe_end, school_ids, grades, runner: runner)
```
dandrabik: All of this can be removed if you define `cte_records` above and change `result` to `results`

services/QuillLMS/spec/queries/snapshots/top_concepts_practiced_query_spec.rb
```diff
@@ -0,0 +1,86 @@
...
+
+        expect(result.map { |r| r['value'] }).not_to include(concepts[10].name)
+      end
+
+      it 'should count concepts in unit_activities outside of the timeframe' do
+        too_old = create(:activity_session, activity: activities.first, classroom_unit: classroom_units.first, completed_at: timeframe_start - 1.day)
+        too_new = create(:activity_session, activity: activities.first, classroom_unit: classroom_units.first, completed_at: timeframe_end + 1.day)
```
dandrabik: these should be `let` declarations.

services/QuillLMS/spec/queries/snapshots/top_concepts_practiced_query_spec.rb
```diff
@@ -0,0 +1,86 @@
...
+
+        runner = QuillBigQuery::TestRunner.new([
```
dandrabik: Same, define `cte_records`

services/QuillLMS/spec/support/shared_contexts/snapshots/snapshots_top_x_cte.rb
```diff
@@ -0,0 +1,8 @@
+# frozen_string_literal: true
+
+RSpec.shared_context 'Snapshots TopX CTE' do
```
dandrabik: I think you should delete this context and just include the period context in the test.

services/QuillLMS/spec/workers/snapshots/cache_snapshot_top_x_worker_spec.rb
```diff
@@ -0,0 +1,68 @@
...
+  describe CacheSnapshotTopXWorker, type: :worker do
+    subject { described_class.new }
+
+    let(:cache_key) { 'CACHE_KEY' }
+    let(:query) { 'most-active-schools' }
+    let(:user_id) { 123 }
+    let(:timeframe_name) { 'last-30-days' }
+    let(:school_ids) { [1,2,3] }
+    let(:grades) { ['Kindergarten',1,2,3,4] }
+    let(:query_double) { double }
```
dandrabik: You could define this fully here and remove  the `allow` below.
```suggestion
    let(:query_double) { double(run: {}) }
```

services/QuillLMS/spec/workers/snapshots/cache_snapshot_top_x_worker_spec.rb
```diff
@@ -0,0 +1,68 @@
...
+
+      before do
+        stub_const("Snapshots::CacheSnapshotTopXWorker::QUERIES", {
+          query => query_double
+        })
+
+        # Intercept theses calls to isolate them in separate cases
+        allow(query_double).to receive(:run).and_return({})
+        allow(Rails.cache).to receive(:write)
+        allow(PusherTrigger).to receive(:run)
```
dandrabik: I think it's better to have specific `expect.to` and `expect.not_to` in each test, rather than a universal allow that some tests overwrite with `expect`. That setup will also catch circumstances when writes or runs are happening and they shouldn't.

services/QuillLMS/spec/queries/snapshots/top_concepts_assigned_query_spec.rb
```diff
@@ -0,0 +1,87 @@
...
+
+      it 'should successfully get data' do
+        runner = QuillBigQuery::TestRunner.new([
+          runner_context,
+          unit_activity_bundles
+        ])
+        result = described_class.run(timeframe_start, timeframe_end, school_ids, grades, runner: runner)
+
+        expect(result).to eq([
+          {"count"=>unit_activity_bundles[0].length, "value"=>concepts[0].name},
```
dandrabik: You should DRY this up, see my other comment.

services/QuillLMS/spec/queries/snapshots/most_active_schools_query_spec.rb
```diff
@@ -0,0 +1,114 @@
...
+          teachers,
+          classrooms_teachers,
+          schools,
+          schools_users,
+          classroom_units
+        ]
+      }
+
+      it 'should include the top 10 results if they are present' do
+        runner = QuillBigQuery::TestRunner.new([
```
dandrabik: This can be DRY'd up by leaning on the shared_context more, see my other comment.

dandrabik (changes requested):
Thanks for integrating these tests, I think it's worth doing another pass to DRY a few things up:
1. Lean on the [shared_context setup](https://github.com/empirical-org/Empirical-Core/blob/develop/services/QuillLMS/spec/support/shared_contexts/quill_big_query_test_runner_setup.rb) more (see comment)
2. DRY up some of the 10-line vars used that could be generated
3. Delete the new empty shared_context
4. Convert vars to `let` variables so they can be reused.

services/QuillLMS/spec/rails_helper.rb
```diff
@@ -85,6 +85,7 @@
   config.filter_run focus: true
   config.filter_run_excluding benchmarking: true
   config.filter_run_excluding external_api: true
+  config.filter_run_excluding big_query_snapshot: true
```
brendanshean, line 88: Whoops.  Good catch on this.  I guess we've been running these on lms_rails_build too.

services/QuillLMS/spec/queries/snapshots/most_active_schools_query_spec.rb
```diff
@@ -0,0 +1,72 @@
...
+          classrooms,
+          teachers,
+          classrooms_teachers,
+          schools,
+          schools_users,
+          classroom_units
+        ]
+      }
+
+      let(:runner) { QuillBigQuery::TestRunner.new(cte_records) }
```
brendanshean, line 32: `let(:runner)` and `let(:results)` can be removed since they are already defined in the shared_context [here](https://github.com/empirical-org/Empirical-Core/blob/develop/services/QuillLMS/spec/support/shared_contexts/quill_big_query_test_runner_setup.rb#L7)

services/QuillLMS/spec/queries/snapshots/top_concepts_assigned_query_spec.rb
```diff
@@ -0,0 +1,66 @@
...
+          classroom_units,
+          concepts,
+          activities,
+          classroom_units
+        ]
+      }
+
+      let(:cte_records) { [runner_context, unit_activity_bundles] }
+
+      let(:runner) { QuillBigQuery::TestRunner.new(cte_records) }
```
brendanshean, line 45: `let(:runner)` and `let(:results)` can be removed since they are already defined in the shared_context [here](https://github.com/empirical-org/Empirical-Core/blob/develop/services/QuillLMS/spec/support/shared_contexts/quill_big_query_test_runner_setup.rb#L7)

services/QuillLMS/spec/queries/snapshots/top_concepts_practiced_query_spec.rb
```diff
@@ -0,0 +1,61 @@
...
+          classrooms_teachers,
+          schools,
+          schools_users,
+          concepts,
+          activities,
+          classroom_units
+        ]
+      }
+
+      let(:runner) { QuillBigQuery::TestRunner.new(cte_records) }
```
brendanshean, line 39: `let(:runner)` and `let(:results)` can be removed since they are already defined in the shared_context [here](https://github.com/empirical-org/Empirical-Core/blob/develop/services/QuillLMS/spec/support/shared_contexts/quill_big_query_test_runner_setup.rb#L7)

services/QuillLMS/spec/support/shared/quill_big_query/test_runner.rb
```diff
@@ -64,7 +64,9 @@ def execute(query)
     private def attr_type_value(attr_type, value)
       case attr_type
       when :boolean, :decimal, :float, :integer then value
-      when :inet, :jsonb, :string then "'#{value}'"
+      when :inet then "'#{value}'"
+      when :jsonb then "'#{value.to_json}'"
+      when :string, :text then "\"#{value}\""
```
brendanshean, line 69: Thanks for cleaning this type conversions up

brendanshean (approved):
This looks great.  I really like the runner_context pattern you've established here.  Thank you for building out all of these testing examples.

services/QuillLMS/spec/queries/snapshots/most_active_schools_query_spec.rb
```diff
@@ -0,0 +1,72 @@
...
+
+      context 'all activity_sessions' do
+        let(:expected_result) do
+          (0..9).map { |i| {"value"=>schools[i].name, "count"=>activity_sessions[i].length} }
+        end
+        let(:cte_records) { [runner_context, activity_sessions] }
+
+        it { expect(results).to eq(expected_result) }
+        it { expect(results.length).to eq(10) }
+        it { expect(results.map { |r| r["value"] }).not_to include(schools[10].name) }
```
dandrabik, line 43: Since you have a test explicitly matching the entire `results`, these two additional tests are redundant.

services/QuillLMS/spec/queries/snapshots/top_concepts_assigned_query_spec.rb
```diff
@@ -0,0 +1,66 @@
...
+      let(:results) { described_class.run(timeframe_start, timeframe_end, school_ids, grades, runner: runner) }
+
+      context 'query LIMITs and shape' do
+        let(:expected_result) do
+          (0..9).map { |i| {"count"=>unit_activity_bundles[i].length, "value"=>concepts[i].name} }
+        end
+
+        it { expect(results).to eq(expected_result) }
+        it { expect(results.map { |r| r['value'] }).not_to include(concepts[10].name) }
```
dandrabik, line 54: You can remove this expects, since you do an explicit match above of the entire result.

services/QuillLMS/spec/queries/snapshots/top_concepts_practiced_query_spec.rb
```diff
@@ -0,0 +1,61 @@
...
+      let(:results) { described_class.run(timeframe_start, timeframe_end, school_ids, grades, runner: runner) }
+
+      context 'all activity_sessions' do
+        let(:expected_result) do
+          (0..9).map { |i| {"count"=>activity_session_bundles[i].length, "value"=>concepts[i].name} }
+        end
+        let(:cte_records) { [runner_context, activity_session_bundles] }
+
+        it { expect(results).to eq(expected_result) }
+        it { expect(results.map { |r| r['value'] }).not_to include(concepts[10].name) }
```
dandrabik, line 49: This expect is redundant with the exact match above.

dandrabik (approved):
Approved! I left a few tiny cleanups, but looks good.

Reading these written out, I personally think letting `results` and `runner` definitions just live in the `shared_context` is a little cleaner, but I understand your point of view. It might be worth having you and Brendan chat about this to agree on a pattern.

PR #10608, opened 2023-06-01
Add a model to track teacher logins when we send Segment events

anathomical:
## WHAT
Begin tracking each Teacher login
## WHY
We want access to similar time-series data for teacher logins as we have in Segment and Ortto.  Specifically, we want to be able report how many "active teachers" a given school has during arbitrary time periods.  While we acknowledge that teachers who haven't been force-logged-out can be active without signing in for significant periods of time, we want to start with a simple copy of what we're doing in Segment already.
## HOW
- Add a new read-only `UserLogin` model that with a `created_at` timestamp
- Write a new `UserLogin` record for any user who triggers a `Teacher signed in` event in Segment (note that we're not recording for students at this time to reduce volume of events being stored)

### Notion Card Links
https://www.notion.so/quill/Backend-for-Admin-Snapshot-Report-a90a9996669a4c60b8a242ce2306cb12?pvs=4#e9d3b1f40ff84e8abf0a26791a844f58

PR Checklist | Your Answer
------------ | -------------

### DIFF
--- a/services/QuillLMS/app/workers/user_login_worker.rb
+++ b/services/QuillLMS/app/workers/user_login_worker.rb
@@ -33,6 +33,6 @@ class UserLoginWorker
   end
 
   def record_user_login
-    UserLogin.create(user: @user)
+    @user.record_login
   end
 end
--- a/services/QuillLMS/app/workers/user_login_worker.rb
+++ b/services/QuillLMS/app/workers/user_login_worker.rb
@@ -16,6 +16,7 @@ class UserLoginWorker
         SegmentIo::BackgroundEvents::TEACHER_SIGNIN,
         properties: @user&.segment_user&.common_params
       )
+      record_user_login
     when User::STUDENT
       # keep these in the following order so the student is the last one identified
       teacher = @user.teacher_of_student
@@ -29,5 +30,9 @@ class UserLoginWorker
         )
       end
     end
+  end
+
+  def record_user_login
+    UserLogin.create(user: @user)
   end
 end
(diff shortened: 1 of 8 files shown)


services/QuillLMS/app/workers/user_login_worker.rb
```diff
@@ -16,6 +16,7 @@ def perform(id)
         SegmentIo::BackgroundEvents::TEACHER_SIGNIN,
         properties: @user&.segment_user&.common_params
       )
+      record_user_login
```
brendanshean, line 19: What about a method in user.rb:

```
def record_login
  user_logins.create
end
```

Then the we're left with `@user.record_login` which removes the dependency on `UserLogin`

brendanshean (approved):
Looks good to me.

brendanshean (approved):
Looks good to me.

brendanshean (approved):
Looks good to me.

services/QuillLMS/app/workers/user_login_worker.rb
anathomical: I did consider that.  There's a part of me that's become quite resistant to adding more code to `User` because that model is HUGE and complex and I don't want to add to it.  Am I just being silly with that since it really does make the most sense to have that method in the model?

brendanshean: `User` is definitely a god object, but I think that issue is unrelated to this addition.  In other words, I think this method belongs on user, whereas there are lots of other methods in user.rb that do not, or could be extracted via concerns and decorators.

happythenewsad: I kinda lean towards keeping it in `User`.  Another code smell is that UserLoginWorker is called from QuillAuthentication, not User. So eventually, it might be good to have QuillAuthentication call `User#record_login`, where:

```
class User
  def record_login
    UserLoginWorker.perform_async
    user_logins.create if teacher? || admin?
  end
...
end
```

happythenewsad (approved):

PR #10629, opened 2023-06-08
Delete Obsolete Sessions Optimization.

dandrabik:
## WHAT
Tiny PR. Was looking at stats and saw this very slow job. This job is really slow since the queries used by `in_batches` are a select with ordering (which seems to be very slow on this table).
## WHY
This job is slow and getting slower, it took 39 minutes last week.
## HOW
Remove batch selects from job which involve ordering (this is the part that is slow). Loop through delete statements and delete in chunks.
### Screenshots
Screenshot 2023-06-08 at 9 31 25 AM
Screenshot 2023-06-08 at 9 31 16 AM
Screenshot 2023-06-08 at 9 31 04 AM

PR Checklist | Your Answer
------------ | -------------
Have you added and/or updated tests? |  Yes, but mostly a pure refactor.

### DIFF
--- a/services/QuillLMS/app/workers/delete_obsolete_active_activity_sessions_worker.rb
+++ b/services/QuillLMS/app/workers/delete_obsolete_active_activity_sessions_worker.rb
@@ -4,13 +4,17 @@ class DeleteObsoleteActiveActivitySessionsWorker
   include Sidekiq::Worker
   sidekiq_options queue: SidekiqQueue::LOW
 
+  BATCH_SIZE = 100
+
   def perform
-    obsolete_query.in_batches.delete_all
+    while obsolete_query.delete_all > 0
+      # empty loop since condtional is also the action
+    end
   end
 
   private def obsolete_query
     ActiveActivitySession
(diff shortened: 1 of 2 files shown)


services/QuillLMS/app/workers/delete_obsolete_active_activity_sessions_worker.rb
```diff
@@ -4,13 +4,17 @@ class DeleteObsoleteActiveActivitySessionsWorker
   include Sidekiq::Worker
   sidekiq_options queue: SidekiqQueue::LOW
 
+  BATCH_SIZE = 100
+
   def perform
```
dandrabik, line 9: Note that `delete_all` ["Returns the number of rows affected."](https://api.rubyonrails.org/classes/ActiveRecord/Relation.html#method-i-delete_all) as an integer.

brendanshean (approved):
Thanks for speeding this up.  Clever solution

happythenewsad (approved):

PR #10638, opened 2023-06-09
Add Canvas LTI configuration

brendanshean:
## WHAT
Add a LTI Launch configuration.

## WHY
As a pre-requisite for Canvas SSO, we need a LTI app that users can install in their Canvas Instances.

## HOW
There are two endpoints:
1) `canvas_integration/lti/launch_config.xml`: This XML file contains an LTI configuration:
Screenshot 2023-06-09 at 8 30 40 AM

2) `canvas_integration/lti/launch`:  Using the launch_config.xml, the LTI app loads an iframe with the launch URL:
Screenshot 2023-06-09 at 8 36 03 AM

### Screenshots

### Notion Card Links

PR Checklist | Your Answer
------------ | -------------

### DIFF
--- a/services/QuillLMS/app/controllers/canvas_integration/lti_controller.rb
+++ b/services/QuillLMS/app/controllers/canvas_integration/lti_controller.rb
@@ -5,7 +5,7 @@ module CanvasIntegration
     # launch is loaded within Canvas in an iframe muddling the CSRF token
     skip_before_action :verify_authenticity_token, only: :launch
 
-    SSO_LINK_TEXT = 'This link will redirect you to Quill.org for SSO'
+    SSO_LINK_TEXT = 'Log in to Quill.org with your Canvas credentials'
 
     def launch
       @link_text = SSO_LINK_TEXT
--- a/services/QuillLMS/app/controllers/canvas_integration/lti_controller.rb
+++ b/services/QuillLMS/app/controllers/canvas_integration/lti_controller.rb
@@ -1,0 +1,21 @@
+# frozen_string_literal: true
+
+module CanvasIntegration
+  class LtiController < ApplicationController
+    # launch is loaded within Canvas in an iframe muddling the CSRF token
+    skip_before_action :verify_authenticity_token, only: :launch
+
+    SSO_LINK_TEXT = 'This link will redirect you to Quill.org for SSO'
+
+    def launch
+      @link_text = SSO_LINK_TEXT
+      @link_url = root_path # root_path will be replaced with SSO path in next PR
+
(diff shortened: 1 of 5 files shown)


services/QuillLMS/app/controllers/canvas_integration/lti_controller.rb
```diff
@@ -0,0 +1,21 @@
+# frozen_string_literal: true
+
+module CanvasIntegration
+  class LtiController < ApplicationController
+    # launch is loaded within Canvas in an iframe muddling the CSRF token
+    skip_before_action :verify_authenticity_token, only: :launch
+
+    SSO_LINK_TEXT = 'This link will redirect you to Quill.org for SSO'
```
anathomical: Is this the right text for this link?  I imagine that most normal users won't know what "SSO" means.  Maybe not a big deal, but I'm always curious about customer-facing copy.

services/QuillLMS/config/routes.rb
```diff
@@ -595,6 +595,11 @@
     post '/learn_worlds/courses', to: 'learn_worlds#courses'
   end
 
+  namespace :canvas_integration do
+    get 'lti/launch_config.xml' => 'lti#launch_config'
```
anathomical, line 599: My first thought was "there's got to be a cleaner way to restrict a route to a single format", but then I looked and there's not.  TIL!

anathomical (approved):
One comment about copy, but this looks solid to me.  I like the way you've organized your specs.

services/QuillLMS/config/routes.rb
dandrabik: I believe you can do this with [constraints](https://guides.rubyonrails.org/routing.html#request-based-constraints) if you want to, e.g.

```ruby
get 'lti/launch_config', constraints: lambda { |req| req.format == :xml }

File: services/QuillLMS/app/controllers/canvas_integration/lti_controller.rb (hunk 1 above)
Comment (dandrabik) [reply services/QuillLMS/app/controllers/canvas_integration/lti_controller.rb]: This is definitely placeholder copy, but let's make it closer to what it will be.
```suggestion
    SSO_LINK_TEXT = 'Log in to Quill.org with your Canvas credentials'
```

dandrabik (approved):
Looks good! Thanks for moving this along.

services/QuillLMS/config/routes.rb
brendanshean: Another way is to use: `get '/lti/launch_config', to: 'lti#launch_config', constraints: { format: 'xml' }` but I don't know if that's any better.

PR #10647, opened 2023-06-13
lms/allow-user-login-to-be-destroyed

anathomical:
## WHAT
Allow UserLogin records to be destroyed through dependency chains
## WHY
Originally making the record read-only meant that it couldn't be destroyed, and that prevented users from being destroyed due to the dependent: :destroy chain raising an error.
## HOW
Add a clause to the `readonly?` method in the model so that it is not read-only if it is being destroyed by an association.

### Notion Card Links
https://www.notion.so/quill/Staff-Demo-Account-Not-Resetting-Overnight-d34aba9e81204ec6a4829b8de79c961b

PR Checklist | Your Answer
------------ | -------------

### DIFF
--- a/services/QuillLMS/app/models/user_login.rb
+++ b/services/QuillLMS/app/models/user_login.rb
@@ -20,6 +20,8 @@ class UserLogin < ApplicationRecord
   belongs_to :user
 
   def readonly?
+    return false if destroyed_by_association
+
     !new_record?
   end
 end
--- a/services/QuillLMS/spec/models/user_login_spec.rb
+++ b/services/QuillLMS/spec/models/user_login_spec.rb
@@ -22,4 +22,15 @@ RSpec.describe UserLogin, type: :model do
   context 'should relations' do
     it { should belong_to(:user) }
   end
+
+  context 'dependent destroy' do
+    let(:user_login) { create(:user_login) }
+    let(:user) { user_login.user }
+
+    it 'should destroy all user_logins parent user is destroyed' do
+      user.destroy!
+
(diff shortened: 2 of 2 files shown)


services/QuillLMS/app/models/user_login.rb
```diff
@@ -20,6 +20,8 @@ class UserLogin < ApplicationRecord
   belongs_to :user
 
   def readonly?
+    return false if destroyed_by_association
```
brendanshean, line 23: TIL `destroyed_by_association`.  Cool!

brendanshean (approved):

services/QuillLMS/spec/models/user_login_spec.rb
```diff
@@ -22,4 +22,15 @@
   context 'should relations' do
     it { should belong_to(:user) }
   end
+
+  context 'dependent destroy' do
+    let(:user_login) { create(:user_login) }
+    let(:user) { user_login.user }
+
+    it 'should destroy all user_logins parent user is destroyed' do
+      user.destroy!
```
brendanshean, line 31: Another way to write this using change syntax

`it { expect { user.destroy! }.to change(UserLogin, :count).from(1).to(0) }`

anathomical: You know, that's what I tried first, but it requires `user_login` to be instantiated outside of the block, so I would have had to use `let!`.  I decided that the awkwardness of `find_by` was worth avoiding `let!` in this case, but could easily see going the other way.

PR #10652, opened 2023-06-15
Extract role information and import User before Canvas SSO flow

brendanshean:
## WHAT
Before launch, [extract](https://github.com/empirical-org/Empirical-Core/blob/7648af501a1ca808f776fe50e7a1a980e78bd2f9/services/QuillLMS/app/controllers/canvas_integration/lti_controller.rb#LL41C1-L42C1) role and [run](https://github.com/empirical-org/Empirical-Core/blob/7648af501a1ca808f776fe50e7a1a980e78bd2f9/services/QuillLMS/app/controllers/canvas_integration/lti_controller.rb#L12) `UserImporter`

## WHY
Unlike Clever which passes a [user_type](https://github.com/empirical-org/Empirical-Core/blob/develop/services/QuillLMS/app/services/clever_integration/sign_up/sub_main.rb#L8) value during SSO callback, there is no role information found in the `OmniAuth::AuthHash` object for Canvas SSO.  Fortunately, role information is available at launch, so we can import the user then.

## HOW
The steps are as follows:
1. User loads LTI launch app.  This page is passed from Canvas a `params[:ext_roles]` containing [role information](https://canvas.instructure.com/doc/api/file.tools_variable_substitutions.html#canvas-xuser-allroles-duplicates-ext_roles-which-is-sent-by-default).
1. The `role` is [pulled out](https://github.com/empirical-org/Empirical-Core/blob/f04e4e5763559f25e3cd7c0ffb1119e6279b589c/services/QuillLMS/app/controllers/canvas_integration/lti_controller.rb#L34) with `RoleExtractor` and  to [passed](https://github.com/empirical-org/Empirical-Core/blob/7648af501a1ca808f776fe50e7a1a980e78bd2f9/services/QuillLMS/app/controllers/canvas_integration/lti_controller.rb#L35) to `UserImporter` along with other some custom params via [LTI variable substitutions](https://canvas.instructure.com/doc/api/file.tools_variable_substitutions.html) (i.e. email, external_id, name, and url).
1.  `UserImporter` attempts to locate a User with an existing `CanvasAccount` with `canvas_instance_id` and `external_id` rather than [email](https://ed.link/community/why-you-shouldnt-match-users-by-email-address/).  Otherwise, either find or create a user create a user with that email and link it to a new `CanvasAccount`.  

NB:  I made a [judgment call](https://github.com/empirical-org/Empirical-Core/blob/f04e4e5763559f25e3cd7c0ffb1119e6279b589c/services/QuillLMS/app/services/canvas_integration/role_extractor.rb#L34) with an edge case where if a user has both student and teacher roles passed in from canvas, we resolve that user to be a teacher.  The logic here is very simple to change if needed.  Also, if for some reason unknown role information is passed in, we default that role to student.

### Screenshots

### Notion Card Links

PR Checklist | Your Answer
------------ | -------------


services/QuillLMS/app/services/canvas_integration/role_extractor.rb
```diff
@@ -0,0 +1,49 @@
...
+      LEARNER_ROLE => USER_STUDENT_ROLE,
+      STUDENT_INST_ROLE => USER_STUDENT_ROLE,
+    }
+
+    def initialize(canvas_roles)
+      @canvas_roles = canvas_roles
+    end
+
+    def run
+      teacher_roles? ? USER_TEACHER_ROLE : USER_STUDENT_ROLE
```
dandrabik, line 34: Seems like a reasonable judgement call to default to teacher. We should just remember this for potential bug reports if a student is converted to a teacher. Probably better to just monitor that edge case for now, since it would be a big architecture change to allow multiple roles in our system (although long-term, that's a good goal).

services/QuillLMS/app/controllers/canvas_integration/lti_controller.rb
```diff
@@ -10,7 +10,7 @@ class LtiController < ApplicationController
 
     def launch
       @link_text = LAUNCH_TEXT
-      @link_url = canvas_integration_lti_sso_path(canvas_instance_url: params[:custom_canvas_api_baseurl])
+      @link_url = canvas_integration_lti_sso_path(canvas_instance_url: custom_canvas_api_baseurl, role: role)
```
dandrabik: I have some concerns about this approach that are probably easier to chat through, i'll ping you.

dandrabik (approved):
Overall, this approach looks good. I have one small item for discussion that I'll just ping you about.

brendanshean:
@dandrabik Re-requesting due to large refactor.

services/QuillLMS/app/services/canvas_integration/role_extractor.rb
anathomical, line 34: Do we need to support admins?  Is that something that would supersede teacher?  So priority is "highest authority role of `admin` > `teacher` > `student`?  Or would we prefer to maintain feature parity with our other SSO flows?

services/QuillLMS/app/controllers/canvas_integration/lti_controller.rb
```diff
@@ -22,7 +23,22 @@ def launch_config
     def sso
       @canvas_instance_id = CanvasInstance.find_by(url: params[:canvas_instance_url])&.id
       @button_text = SSO_BUTTON_TEXT
-      @button_url = Auth::Canvas::ACCESS_PATH
+      @button_params = { canvas_instance_id: @canvas_instance_id }
+      @button_url = Auth::Canvas::OMNIAUTH_REQUEST_PATH
+    end
+
+    private def import_user
+      CanvasIntegration::UserImporter.run(
```
dandrabik, line 31: Not a huge issue, but we might consider moving this to an async job 1) for performance reasons and 2) to ensure the login page in Canvas loads (if there's an error raised in this flow, the xml page would 500).

We would lose an assurance that the user is created before login, so if that's an issue we could leave it synchronous.

services/QuillLMS/app/services/canvas_integration/user_importer.rb
```diff
@@ -0,0 +1,50 @@
...
+      CanvasAccount
+        .find_by(external_id: external_id, canvas_instance: canvas_instance)
+        &.user
+    end
+
+    private def new_canvas_account_user
+      user
+        .canvas_accounts
+        .build(canvas_instance: canvas_instance, external_id: external_id)
+        .tap(&:save!)
```
dandrabik, line 37: I think this can be shortened
```suggestion
        .create!(canvas_instance: canvas_instance, external_id: external_id)
```

dandrabik (approved):
Overall looks good! Small note about considering making the creation `async` and tiny refactor suggestion.

Thanks for pushing this along, I like this flow!

services/QuillLMS/app/services/canvas_integration/user_importer.rb
brendanshean: I'm using `build` here to [satisfy](https://github.com/empirical-org/Empirical-Core/blob/e7f553406423091b9c1c368c74098a9916dd5d3f/services/QuillLMS/app/models/user.rb#L872) the `requires_password` validation on User.
The `tap(&:save!)` returns saved object (i.e. the newly created canvas account) so that a subsequent call to `.user` can be made.

services/QuillLMS/app/controllers/canvas_integration/lti_controller.rb
brendanshean: My thought here is if there's an issue creating a User, we fail fast here and render some error for the user.  Putting it in a background job will push the error downstream.

I do still have to put a redirect for the error handling but that's pegged with a later PR.

services/QuillLMS/app/services/canvas_integration/user_importer.rb
brendanshean: But you're right to flag this.  I added a comment on why build (and not create) is being used here.

services/QuillLMS/spec/middleware/canvas_auth_spec.rb
```diff
@@ -0,0 +1,40 @@
...
+require 'rails_helper'
+
+RSpec.describe Auth::Canvas::Setup do
+  subject { described_class.new(env, strategy: strategy).run }
+
+  let(:canvas_instance) { create(:canvas_config).canvas_instance }
+  let(:rack_app) { ->(env) { [200, {}, ''] } }
+  let(:strategy) { OmniAuth::Strategies::Canvas.new(rack_app, {}) }
+
+  describe 'request phase' do
```
anathomical: Is it worth also testing setup and teardown logic here?  Or somewhere?

anathomical (approved):
A couple of notes about admins and test coverage, but could see this shipping as is as long as those issues have been considered and we've decided they're fine as-is.

services/QuillLMS/app/services/canvas_integration/role_extractor.rb
brendanshean: Good question.  I think the thought here is to keep admin verification done internally rather than through a third-party.

services/QuillLMS/spec/middleware/canvas_auth_spec.rb
brendanshean: This is a good call out.  I've been debating on this piece for a little bit.  It's kind of messy/clunky to test in the unit spec itself.  I'm probably going to add something in an system test in a subsequent PR.

PR #10653, opened 2023-06-15
Handle Canvas SSO callback phase

brendanshean:
## WHAT
Handle the callback phase of Canvas SSO by 
1. Saving the user's auth_credential
1. Signing the user in
1. Redirecting the user to their profile

## WHY
1. Saving an auth_credential allows for using the refresh_token with API calls to Canvas later
1. User should be authenticated on Quill after Canvas SSO
1. This is the current pattern for redirect after SSO

## HOW
1-3 are [handled](https://github.com/empirical-org/Empirical-Core/blob/edc2f7ef43afdc15f64cf43071dc9d8aced5fbb0/services/QuillLMS/app/controllers/auth/canvas_controller.rb) in `Auth::CanvasController#canvas`.

### Screenshots

### Notion Card Links
https://www.notion.so/quill/Canvas-MVP-a138ee5dc0a741df9bf7c7b1ea5b19ce?pvs=4#14cc6ab3bfd24d1493f39f5b9ef86af3 

PR Checklist | Your Answer
------------ | -------------


services/QuillLMS/app/services/canvas_integration/user_importer.rb
```diff
@@ -0,0 +1,42 @@
...
+    private def new_canvas_account_user
+      user
+        .canvas_accounts
+        .build(canvas_instance: canvas_instance, external_id: canvas_user_external_id)
+        .tap(&:save!)
+        .user
+    end
+
+    private def user
+      User.find_by(email: email) || User.new(email: email, name: name, role: role)
```
dandrabik: We'd want to restrict this role param to `student` or `teacher`

dandrabik (approved):
:+1: This looks good. One note about restricting roles to a subset of roles.

brendanshean:
@dandrabik Re-requesting due to large refactor.

services/QuillLMS/app/services/canvas_integration/auth_credential_saver.rb
```diff
@@ -0,0 +1,59 @@
...
+
+    def initialize(auth_hash)
+      @credentials = auth_hash[:credentials]
+      @external_id = auth_hash[:uid]
+      @info = auth_hash[:info]
+    end
+
+    def run
+      raise CanvasInstanceNotFoundError unless canvas_instance
+      raise CanvasAccountNotFoundError unless canvas_account
```
dandrabik, line 20: I think the only product question I have is if we should do something more nuanced when a canvas account doesn't exist rather than the generic 500 page, e.g.
- Give instructions 
- Create a student account for them
- Redirect back to Canvas

This is an edge case so raising is probably fine for now.

dandrabik (approved):
:+1: Looks good, I like how this is coming together. I left one comment about a change, but that it's a "this is good for now, but we might consider something different down the road". So it's not a blocker.

services/QuillLMS/app/services/canvas_integration/auth_credential_saver.rb
brendanshean: I have a running list of expanded error testing/handling that I'm planning on getting to at the end.  I feel like there will be some time for that when I'm having people do some internal testing.

brendanshean: I will add these notes to that list.

anathomical (approved):

PR #10661, opened 2023-06-20
Change the Bing Spell Check API endpoint

cissyxyu:
## WHAT
Bing is expiring their old spell check endpoint and they have moved their spell check services to a new resource. We need to update our code to point to a new resource.

## WHY
The old resource will expire in October of this year and we will not be able to get spelling feedback from it.

## HOW
I've already set up the new resource in Bing Azure, so we just need to point the endpoints we have hard-coded in to refer to this new resource.

### Screenshots

### Notion Card Links
https://www.notion.so/quill/Product-Board-30f97e4eb01246dbb8264253fb385073?p=8b26bb192e364fab8e9c633cc6f30a74&pm=c

PR Checklist | Your Answer
------------ | -------------

### DIFF
--- a/services/QuillLMS/engines/evidence/spec/controllers/evidence/feedback_controller_spec.rb
+++ b/services/QuillLMS/engines/evidence/spec/controllers/evidence/feedback_controller_spec.rb
@@ -102,7 +102,7 @@ module Evidence
         end
 
         it 'should return correct spelling feedback when endpoint returns 200' do
-          stub_request(:get, "https://api.cognitive.microsoft.com/bing/v7.0/SpellCheck?mode=proof&text=#{ERB::Util.url_encode(entry_with_error)}").to_return(:status => 200, :body => { :flaggedTokens => ([{ :token => "spelin" }]) }.to_json, :headers => ({}))
+          stub_request(:get, "https://api.bing.microsoft.com/v7.0/spellcheck?mode=proof&text=#{ERB::Util.url_encode(entry_with_error)}").to_return(:status => 200, :body => { :flaggedTokens => ([{ :token => "spelin" }]) }.to_json, :headers => ({}))
           post :create, params: { entry: entry_with_error, :prompt_id => prompt.id, :session_id => 1 }, as: :json
 
           parsed_response = JSON.parse(response.body)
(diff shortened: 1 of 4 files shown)


services/QuillLMS/engines/evidence/spec/dummy/db/schema.rb
```diff
@@ -248,15 +248,6 @@
...
   end
 
-  create_table "evidence_rule_hints", force: :cascade do |t|
-    t.bigint "rule_id", null: false
-    t.bigint "hint_id", null: false
-    t.datetime "created_at", precision: 6, null: false
-    t.datetime "updated_at", precision: 6, null: false
-    t.index ["hint_id"], name: "index_evidence_rule_hints_on_hint_id"
-    t.index ["rule_id"], name: "index_evidence_rule_hints_on_rule_id"
-  end
```
anathomical: Is this supposed to be in this PR?  It's probably fine, but I was a bit thrown off?

services/QuillLMS/engines/evidence/spec/controllers/evidence/feedback_controller_spec.rb
```diff
@@ -102,7 +102,7 @@ module Evidence
         end
 
         it 'should return correct spelling feedback when endpoint returns 200' do
-          stub_request(:get, "https://api.cognitive.microsoft.com/bing/v7.0/SpellCheck?mode=proof&text=#{ERB::Util.url_encode(entry_with_error)}").to_return(:status => 200, :body => { :flaggedTokens => ([{ :token => "spelin" }]) }.to_json, :headers => ({}))
+          stub_request(:get, "https://api.bing.microsoft.com/v7.0/spellcheck?mode=proof&text=#{ERB::Util.url_encode(entry_with_error)}").to_return(:status => 200, :body => { :flaggedTokens => ([{ :token => "spelin" }]) }.to_json, :headers => ({}))
```
anathomical: It feels like all of these references should point at the constant:
```
"#{::Evidence::SpellingCheck::BIG_API_URL}?mode=proof..."
```

anathomical (changes requested):
A question about recorded changes to the `schema.rb` file, and then a request to reference the constant everywhere possible instead of repeating it.

services/QuillLMS/engines/evidence/spec/controllers/evidence/feedback_controller_spec.rb
happythenewsad: All of the literal string references shoul use the BING_API_URL constant instead.

services/QuillLMS/engines/evidence/spec/dummy/db/schema.rb
```diff
@@ -248,15 +248,6 @@
     t.datetime "updated_at", precision: 6, null: false
   end
 
-  create_table "evidence_rule_hints", force: :cascade do |t|
```
happythenewsad: Just want to confirm: this is an artifact of `evidence_rule_hints` moving to `evidence_hints`, hosted in the umbrella app, not the engine?

happythenewsad (changes requested):
I'm requesting changes, but to speed up the turnaround time, I pre-approve this PR to deploy, once the string literal issue is addressed ✌️

services/QuillLMS/engines/evidence/spec/dummy/db/schema.rb
cissyxyu: I'm removing this file -- it should not be in this PR and was accidentally added in.

services/QuillLMS/engines/evidence/spec/dummy/db/schema.rb
cissyxyu: I'm removing this file. It was accidentally created when I was testing things on my local DB but should not be in this PR.

anathomical (approved):
Thanks for the quick turn-around.  Looks good to me!

happythenewsad (approved):

PR #10662, opened 2023-06-20
Lms/fix snapshot practice queries

anathomical:
## WHAT
Fix a couple of queries that were calculating slightly differently than we wanted
## WHY
We want the data we report in Snapshots to be intuitive and make sense
## HOW
Tweak some of the query models to adjust how they do their calculations

### Notion Card Links
https://www.notion.so/quill/Admin-Snapshot-Report-QA-e8acbd26f2a640dcb8343012a0edb2fd

PR Checklist | Your Answer
------------ | -------------

### DIFF
--- a/services/QuillLMS/app/queries/snapshots/average_activities_completed_per_student_query.rb
+++ b/services/QuillLMS/app/queries/snapshots/average_activities_completed_per_student_query.rb
@@ -1,10 +1,20 @@
 # frozen_string_literal: true
 
 module Snapshots
-  class AverageActivitiesCompletedPerStudentQuery < ActivitySessionCountQuery
-    def select_clause
-      # "greatest" avoids division by 0 error
-      "SELECT COUNT(DISTINCT activity_sessions.id) / greatest(COUNT(DISTINCT activity_sessions.user_id), 1) AS count"
+  class AverageActivitiesCompletedPerStudentQuery < PeriodQuery
+    def initialize(*params)
+      @params = params
+
+      super
+    end
+
+    def run
+      activities_completed = ActivitiesCompletedQuery.run(*@params)[:count]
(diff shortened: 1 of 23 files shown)


services/QuillLMS/app/queries/snapshots/average_activities_completed_per_student_query.rb
```diff
@@ -1,10 +1,20 @@
 # frozen_string_literal: true
 
 module Snapshots
-  class AverageActivitiesCompletedPerStudentQuery < CountQuery
-    def select_clause
-      # "greatest" avoids division by 0 error
-      "SELECT COUNT(DISTINCT activity_sessions.id) / greatest(COUNT(DISTINCT activity_sessions.user_id), 1) AS count"
+  class AverageActivitiesCompletedPerStudentQuery < PeriodQuery
+    def initialize(*params)
```
brendanshean, line 7: What about `attr_reader :params`?

services/QuillLMS/app/queries/snapshots/average_activities_completed_per_student_query.rb
```diff
@@ -1,10 +1,20 @@
...
+      @params = params
+
+      super
+    end
+
+    def run
+      activities_completed = ActivitiesCompletedQuery.run(*@params)[:count]
```
brendanshean: I find params assignment a little opaque.  I think it would be worth adding some tests to show what's being passed in, especially since this is establishing a new pattern of calling another snapshot query class within this one

brendanshean (approved):
I think the refactoring here makes sense.  I added a note about adding some tests but shouldn't require a re-review

eadams17 (approved):

PR #10665, opened 2023-06-20
lms/de-duplicate-vitally-user-accounts

anathomical:
## WHAT
Add a rake task to bulk unlink Vitally users from double accounts

This script is intended to process a ~70k line CSV file of users who are assigned to multiple accounts and we need to figure out which accounts need to be unlinked from the user
## WHY
Users who are assigned to multiple accounts in Vitally causes confusion and double-counting when producing reports in Vitally about school usage.  We have implemented code to prevent users from being double-assigned to an account, but we have this big batch of legacy users who still need to be updated.
## HOW
- Write a script to parse the provided CSV, and then batch users into groups for `VitallyApi.batch` calls
- For each user, figure out what school (account) they are currently associated with, then build an API batch payload to unlink the user from all other accounts that they're connected to

### Notion Card Links

PR Checklist | Your Answer
------------ | -------------
Have you added and/or updated tests? |  No tests on rake tests

### DIFF
--- a/services/QuillLMS/lib/tasks/vitally.rake
+++ b/services/QuillLMS/lib/tasks/vitally.rake
@@ -1,0 +1,48 @@
+# frozen_string_literal: true
+
+namespace :vitally do
+  task unlink_extra_account_assignments: :environment do
+    pipe_data = $stdin.read unless $stdin.tty?
+
+    unless pipe_data
+      puts 'No data detected on STDIN.  You must pass data to the task for it to run.  Example:'
+      puts '  rake vitally:unlink_extra_account_assignments < path/to/local/file.csv'
+      puts ''
+      puts 'If you are piping data into Heroku, you need to include the --no-tty flag:'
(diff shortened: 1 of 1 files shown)


brendanshean (approved):
Thanks for cleaning this records up!

services/QuillLMS/lib/tasks/vitally.rake
```diff
@@ -0,0 +1,48 @@
...
+          {
+            type: 'unlink',
+            userId: user.id,
+            accountId: id,
+            messageId: SecureRandom.uuid
+          }
+        end
+      end.flatten
+      batch_payload.each do |api_call_payload|
+        puts "Unlinking user #{api_call_payload[:userId]} from account #{api_call_payload[:accountId]}"
```
happythenewsad, line 43: +1 for printing useful stuff in rake tasks

happythenewsad (approved):

PR #10670, opened 2023-06-21
lms/fix-admin-snapshot-cache-expiry

anathomical:
## WHAT
Make sure that we express cache expires_in in seconds rather than passing a DateTime (which becomes seconds since epoch, which is way higher than our intended TTLs).
## WHY
Rails cache expects `expires_in` to be a TTL in seconds, but we're passing a DateTime which then gets cast via `to_i`, and that in turn becomes seconds-since-epoch which is way longer than we want it to be.
## HOW
Add a helper function to the workers that figures out how many seconds it is until our desired cache expiry time, and use that value for `expires_in`

### Notion Card Links
https://www.notion.so/quill/Implement-Admin-Usage-Snapshot-Report-Backend-0dd45ff4b92641328cec85660ea614fb?pvs=4

PR Checklist | Your Answer
------------ | -------------


brendanshean (approved):
Thanks you for fixing this.

PR #10677, opened 2023-06-22
Refactor auth_credential table to STI

brendanshean:
## WHAT
Reorganize the auth_credential table to use STI with subclasses based on the provider names.

## WHY
With the addition of another provider type (i.e. Canvas) it would simplify logic to move the respective providers into their own subclasses.

## HOW
1. Create subclasses `CanvasAuthCredential`, `CleverDistrictAuthCredential`, `CleverLibraryAuthCredential`, `GoogleAuthCredential`
1. Update the places where these AuthCredentials are saved to include these subtypes.
1. Add a backfill script to be run after migration.   Once the backfill is complete, a subsequent PR can add a `null: false` to the type column.

### Screenshots

### Notion Card Links

PR Checklist | Your Answer
------------ | -------------

### DIFF
--- a/services/QuillLMS/app/models/auth_credential.rb
+++ b/services/QuillLMS/app/models/auth_credential.rb
@@ -26,8 +26,6 @@
 #  fk_rails_...  (user_id => users.id)
 #
 class AuthCredential < ApplicationRecord
-  self.inheritance_column = :_type_disabled
-
   belongs_to :user
 
   def canvas_authorized?
--- a/services/QuillLMS/app/models/auth_credential.rb
+++ b/services/QuillLMS/app/models/auth_credential.rb
@@ -29,46 +29,20 @@ class AuthCredential < ApplicationRecord
   self.inheritance_column = :_type_disabled
 
   belongs_to :user
-  has_one :canvas_instance_auth_credential, dependent: :destroy
 
-  CANVAS_PROVIDER = 'canvas'
-
-  GOOGLE_PROVIDER = 'google'
-  GOOGLE_EXPIRATION_DURATION = 6.months
-
-  CLEVER_DISTRICT_PROVIDER = 'clever_district'
-  CLEVER_LIBRARY_PROVIDER = 'clever_library'
-  CLEVER_EXPIRATION_DURATION = 23.hours
-
-  def google_access_expired?
-    google_provider? && !refresh_token_valid?
-  end
-
-  def google_authorized?
-    google_provider? && refresh_token_valid?
-  end
-
-  def google_provider?
-    provider == GOOGLE_PROVIDER
+  def canvas_authorized?
+    false
   end
 
   def clever_authorized?
-    [CLEVER_DISTRICT_PROVIDER, CLEVER_LIBRARY_PROVIDER].include?(provider)
+    false
   end
 
-  def refresh_token_valid?
-    return false if !google_provider? || expires_at.nil? || refresh_token.nil?
-
-    Time.current < refresh_token_expires_at
+  def google_access_expired?
+    false
   end
 
-  def refresh_token_expires_at
-    return nil if !google_provider? || expires_at.nil?
-
-    expires_at + GOOGLE_EXPIRATION_DURATION
-  end
-
-  def token
-    access_token
+  def google_authorized?
+    false
   end
 end
--- a/services/QuillLMS/app/models/auth_credential.rb
+++ b/services/QuillLMS/app/models/auth_credential.rb
@@ -7,17 +7,15 @@
 #  id            :integer          not null, primary key
 #  access_token  :string           not null
 #  expires_at    :datetime
-#  provider      :string
 #  refresh_token :string
 #  timestamp     :datetime
-#  type          :string
+#  type          :string           not null
 #  created_at    :datetime
 #  updated_at    :datetime
 #  user_id       :integer          not null
 #
 # Indexes
 #
-#  index_auth_credentials_on_provider       (provider)
 #  index_auth_credentials_on_refresh_token  (refresh_token)
 #  index_auth_credentials_on_user_id        (user_id)
 #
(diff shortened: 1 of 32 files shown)


anathomical (approved):
I totally dig this refactor.  Should make keeping some of this stuff much easier to maintain and expand moving forward.

services/QuillLMS/app/models/auth_credential.rb
```diff
@@ -26,46 +27,27 @@
...
-    google_provider? && !refresh_token_valid?
-  end
-
-  def google_authorized?
-    google_provider? && refresh_token_valid?
-  end
-
-  def google_provider?
-    provider == GOOGLE_PROVIDER
+  def canvas_authorized?
```
dandrabik, line 38: If we are going the STI approach, we might think about converting these methods to more generic method like `authorized?` and changing some of the logic about how these are called. That might be better as a separate possible refactor, given the diff size.

services/QuillLMS/app/models/auth_credential.rb
```diff
@@ -26,46 +27,27 @@
 #
 class AuthCredential < ApplicationRecord
   belongs_to :user
```
dandrabik, line 27: Also, for the STI route, I think we'd want to change the relationship on `User` to be `has_one :google_auth_credential` (one for each). That could be better as a separate refactor, but then you could do `user.google_auth_credential&.authorized?` and/or move the authorization methods to `user` where they might read more naturally, e.g.
```ruby
# In User
def google_authorized?
  google_auth_credential&.authorized?
end

File: services/QuillLMS/app/services/canvas_integration/auth_credential_saver.rb [hunk 3]
```diff
@@ -29,10 +27,10 @@ def run
 
     private def auth_credential
       @auth_credential ||=
-        AuthCredential.create!(
+        CanvasAuthCredential.create!(
```
dandrabik, line 30: I don't have strong feelings either way, but one thing to consider (take or leave): Since you now have a subclass that you can put Canvas-specific logic in, do you still need this Saver service? Or would this be better moving this to the Rails model with validations?.

services/QuillLMS/db/migrate/20230622125712_add_type_to_auth_credential.rb
```diff
@@ -0,0 +1,7 @@
+# frozen_string_literal: true
+
+class AddTypeToAuthCredential < ActiveRecord::Migration[6.1]
+  def change
```
dandrabik: You might consider putting this migration and the data migration in a separate PR and deploying that first to cut down on potential errors and/or downtime (All of this logic depends on type being there and populated). Then you can deploy the code and re-run the data migration for the last few records.

dandrabik (approved):
Overall, looks good. I left a few comments "for your consideration" but nothing blocking. I do think it would be better to deploy the migration first, since this code needs that column right away.

Nice work on this!

services/QuillLMS/app/services/google_integration/user.rb
```diff
@@ -43,10 +43,11 @@ def update_or_initialize
 
   private def auth_credential_attributes(user)
     {
-      provider:      'google',
-      refresh_token: profile.refresh_token || user&.auth_credential&.refresh_token,
-      expires_at:    profile.expires_at,
       access_token:  profile.access_token,
+      expires_at:    profile.expires_at,
+      refresh_token: profile.refresh_token || user&.auth_credential&.refresh_token,
```
happythenewsad: This is orthogonal to this PR, but can we get rid of a code path and just do 

`refresh_token: profile.refresh_token` ?

happythenewsad (approved):

services/QuillLMS/app/services/google_integration/user.rb
happythenewsad: Also, this line will need to be updated if you implement Dan's suggestion of adding `has_one` associations in `User.rb` for each STI class

services/QuillLMS/db/migrate/20230622125712_add_type_to_auth_credential.rb
brendanshean: That's a good point.  I will pull it out in a separate PR.

services/QuillLMS/app/services/canvas_integration/auth_credential_saver.rb
brendanshean: I like the idea, but I don't want the scope of this to grow too much.  I have an overall larger refactor for later that I'm imagining with Clever/Canvas/Google so I will add this as a note to that.

services/QuillLMS/app/models/auth_credential.rb
brendanshean: I was thinking about the `has_one :canvas_auth_credential`, `has_one :google_auth_credential`, etc. but felt that would actually change how we use auth_credentials sitewide since it would allow for the possibility of multiple auth_credentials.  We should discuss offline

brendanshean: I was actually thinking about the `has_one :canvas_auth_credential`, `has_one :google_auth_credential`, etc. yesterday, but felt that would actually change how we use auth_credentials sitewide since it would allow for the possibility of multiple auth_credentials.  We should probably discuss offline

services/QuillLMS/app/services/google_integration/user.rb
brendanshean: Good call.  I will remove that.

services/QuillLMS/app/models/auth_credential.rb
brendanshean: I was thinking about a generic `authorized?` but hadn't gotten into specifics.  I like your solution.

services/QuillLMS/app/models/auth_credential.rb
brendanshean: After discussing offline, Dan and I agreed that this is a good potential solution long term, but out of scope for this project.

PR #10678, opened 2023-06-22
lms/handle-invalid-ids-in-vitally-unlink

anathomical:
## WHAT
Allow for cases where the Vitally data has invalid user ids
## WHY
Apparently the data file Vitally sent us includes some nonsense data in it which we need to account for
## HOW
Use `find_by` so that we don't error if the `id` value from Vitally is invalid, and use `next` to skip cases where there is no User.

### Notion Card Links

PR Checklist | Your Answer
------------ | -------------
Have you added and/or updated tests? |  No tests on rake tasks
Have you deployed to Staging? | Tiny change

### DIFF
--- a/services/QuillLMS/lib/tasks/vitally.rake
+++ b/services/QuillLMS/lib/tasks/vitally.rake
@@ -19,8 +19,7 @@ namespace :vitally do
     # has a batch limit of 100.
     CSV.parse(pipe_data, headers: true).each_slice(33) do |batch|
       batch_payload = batch.map do |row|
-        # TIL `includes` makes a separate query to the db, and `joins` doesn't store the joined value in ActiveRecord, so if you want a single query to pull multiple models you need to use both?
-        user = User.joins(:schools_users).includes(:schools_users).find_by(id: row['externalId'])
+        user = User.eager_load(:schools_users).find_by(id: row['externalId'])
 
         # Some of the data provided by Vitally is bad, so we need to account for that
         next unless user
--- a/services/QuillLMS/lib/tasks/vitally.rake
+++ b/services/QuillLMS/lib/tasks/vitally.rake
@@ -19,7 +19,12 @@ namespace :vitally do
     # has a batch limit of 100.
     CSV.parse(pipe_data, headers: true).each_slice(33) do |batch|
       batch_payload = batch.map do |row|
-        user = User.includes(:schools_users).find(row['externalId'])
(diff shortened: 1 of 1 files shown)


services/QuillLMS/lib/tasks/vitally.rake
```diff
@@ -19,7 +19,12 @@ namespace :vitally do
     # has a batch limit of 100.
     CSV.parse(pipe_data, headers: true).each_slice(33) do |batch|
       batch_payload = batch.map do |row|
-        user = User.includes(:schools_users).find(row['externalId'])
+        # TIL `includes` makes a separate query to the db, and `joins` doesn't store the joined value in ActiveRecord, so if you want a single query to pull multiple models you need to use both?
+        user = User.joins(:schools_users).includes(:schools_users).find_by(id: row['externalId'])
+
+        # Some of the data provided by Vitally is bad, so we need to account for that
+        next unless user
+
```
brendanshean, line 26: Very minor, but if we're adding a joins, is safe navigation necessary here:
 `quill_school_id = user.schools_users&.school_id`

services/QuillLMS/lib/tasks/vitally.rake
```diff
@@ -19,7 +19,12 @@ namespace :vitally do
     # has a batch limit of 100.
     CSV.parse(pipe_data, headers: true).each_slice(33) do |batch|
       batch_payload = batch.map do |row|
-        user = User.includes(:schools_users).find(row['externalId'])
+        # TIL `includes` makes a separate query to the db, and `joins` doesn't store the joined value in ActiveRecord, so if you want a single query to pull multiple models you need to use both?
```
brendanshean: I usually have to look this up.  Reminds me of the other [options](https://www.bigbinary.com/blog/preload-vs-eager-load-vs-joins-vs-includes) preload and eager_load too.  I feel like at one point Rails changed how includes worked which I think is why I find this confusing.

brendanshean (approved):

brendanshean (approved):

services/QuillLMS/lib/tasks/vitally.rake
anathomical: Good catch.  We actually want cases where there are no school assignments, so I think I actually need to make the `joins` into a `eager_load`.

happythenewsad (approved):

brendanshean (approved):

PR #10679, opened 2023-06-22
Lms/add admin snapshot filter options

anathomical:
## WHAT
- Refactor approach used in `options` generation to use PostgreSQL rather than BigQuery for improved stability and user experience
- Add additional options for filters: specifically a list of teachers and a list of classrooms for more specific filtering
- Modify the `options` endpoint to take params of already-selected filter options in order to narrow down higher-specificity filter values based on more general filter selections (e.g. showing only teachers from a selected school)
- Update the `query` endpoint to take additional filter params (`teacher_ids` and `classroom_ids`)
- Update Snapshot cache key generation to handle additional filter params
## WHY
Later versions of the Admin Snapshots project requires these additional filters, and letting the back-end update filter options based on other filter selections should simplify front-end logic.
## HOW
- Write new `OptionsQuery` sub-classes for Teachers and Classrooms
- Add the new queries to the `options` action
- Allow `options` action to accept filter params and pass them through to queries
- Adjust CacheKey generation logic to utilize new params

### Notion Card Links
https://www.notion.so/quill/Backend-for-Admin-Snapshot-Report-a90a9996669a4c60b8a242ce2306cb12?pvs=4

PR Checklist | Your Answer
------------ | -------------


services/QuillLMS/app/controllers/snapshots_controller.rb
```diff
@@ -37,10 +37,16 @@ def top_x
...
+    school_ids = option_params[:school_ids]
+    grades = option_params[:grades]
+    teacher_ids = option_params[:teacher_ids]
+
     render json: {
       timeframes: Snapshots::Timeframes.frontend_options,
       schools: Snapshots::SchoolsOptionsQuery.run(current_user.id),
-      grades: GRADE_OPTIONS
+      grades: GRADE_OPTIONS,
+      teachers: Snapshots::TeachersOptionsQuery.run(current_user.id, school_ids, grades),
```
dandrabik: What would happen if one of the 3 BigQuery queries gives the "job not complete" error. I believe it would cause this endpoint to throw a 500. Does the frontend fetch them again in that case? With 3 jobs for an endpoint, seems like we might be upping that likelihood (although I don't have a good feel about how often that happens).

A few options that might be worth a discussion:
- Should we consider fetching these from the Postgres DB (and maybe limit to data to the end of the previous day). These aren't intense queries, and they might be more reliable coming from Postgres. 
- Should we consider making multiple endpoints? Right now, if you select a teacher_id, we only have to re-run the `classrooms` options, but we re-run all 5. If that's much more complex to implement, probably not worth the gain, but worth considering.

services/QuillLMS/app/queries/snapshots/period_query.rb
```diff
@@ -4,7 +4,7 @@ module Snapshots
   class PeriodQuery < ::QuillBigQuery::Query
     attr_accessor :timeframe_start, :timeframe_end, :school_ids, :grades
 
-    def initialize(timeframe_start, timeframe_end, school_ids, grades = nil, options = {})
+    def initialize(timeframe_start, timeframe_end, school_ids, grades = nil, options: {})
```
dandrabik: I don't have strong feelings either way, but curious why this change? Changes to protocol for using the runner (the only use case) from `runner: runner` to `options: {runner: runner}`. NBD, just wondering if I'm missing something.

services/QuillLMS/spec/queries/snapshots/options_query_spec.rb
```diff
@@ -0,0 +1,68 @@
...
+      end
+    end
+
+    context 'external_api', :big_query_snapshot do
+      include_context 'Snapshots Option CTE'
+
+      let(:results) {test_options_query.run(*query_args, options: {runner: runner}) }
+      # We want to ensure that a single admin is connected to all schools
+      # `admin` is already the user associated with the first school, but not the others
+      let(:additional_schools_admins) { schools[1..-1].map { |s| SchoolsAdmins.create(school: s, user: admin) } }
```
happythenewsad: IMO `schools.filter { ... }.map ....`

would be more intelligible

services/QuillLMS/app/queries/snapshots/options_query.rb
```diff
@@ -2,10 +2,13 @@
 
 module Snapshots
   class OptionsQuery < ::QuillBigQuery::Query
-    attr_accessor :admin_id
+    attr_accessor :admin_id, :school_ids, :grades, :teacher_ids
 
-    def initialize(admin_id, options = {})
+    def initialize(admin_id, school_ids = nil, grades = nil, teacher_ids = nil, options: {})
```
happythenewsad: Now that the method signature is getting longer, why not make `school_ids`, `grades`, and `teacher_ids` keyword params as well?

services/QuillLMS/app/queries/snapshots/period_query.rb
happythenewsad: Same comment here, regarding keyword params

happythenewsad (approved):
Couple nits; looks good

dandrabik (commented):
Thanks for putting this together. I'm leaving this review as a "comment" since I think it overall looks good, and I'm not really requesting changes, but there are a few items that might be worth discussion. Those may incur large changes, so holding off on approving for that reason.

services/QuillLMS/app/controllers/snapshots_controller.rb
anathomical: Actually, moving these queries to PostgreSQL is probably a good thing to do from a purely product perspective.  Right now if an admin is assigned a new school, their filter options won't show the new school until BigQuery updates that data.  So I think moving to PostgreSQL makes sense to do as part of this PR.

anathomical:
Re-requesting review on this PR as I've made extensive changes in light of moving the Options data query from BigQuery to PostgreSQL.

services/QuillLMS/app/controllers/snapshots_controller.rb
```diff
@@ -4,7 +4,6 @@ class SnapshotsController < ApplicationController
   GRADE_OPTIONS = [
     {value: "Kindergarten", name: "Kindergarten"},
     {value: "1", name: "1st"},
-    {value: "2", name: "2nd"},
```
brendanshean: Just confirming no second grade?

services/QuillLMS/app/queries/snapshots/period_query.rb
```diff
@@ -42,10 +46,22 @@ def school_ids_where_clause
       "AND schools.id IN (#{school_ids.join(',')})"
     end
 
-    def grade_where_clause
+    def grades_where_clause
       return "" if grades.nil? || grades.empty?
```
brendanshean: Would `return "" if grade.blank?` apply here?

services/QuillLMS/app/queries/snapshots/period_query.rb
```diff
@@ -42,10 +46,22 @@ def school_ids_where_clause
...
 
       "AND classrooms.grade IN (#{grades.map { |g| "'#{g}'" }.join(',')})"
     end
+
+    def teacher_ids_where_clause
+      return "" if teacher_ids.nil? || teacher_ids.empty?
```
brendanshean: Would `return "" if teacher_ids.blank?` apply here?

services/QuillLMS/app/queries/snapshots/period_query.rb
```diff
@@ -42,10 +46,22 @@ def school_ids_where_clause
...
+
+      "AND schools_users.user_id IN (#{teacher_ids.join(',')})"
+    end
+
+    def classroom_ids_where_clause
+      return "" if classroom_ids.nil? || classroom_ids.empty?
```
brendanshean: Would `return "" if classroom_ids.blank?` apply here?

services/QuillLMS/app/lib/snapshots/cache_keys.rb
```diff
@@ -2,16 +2,41 @@
...
         current_start,
         current_end,
-        "school-ids-#{(school_ids || []).sort.join('-')}",
-        "grades-#{(grades || []).map(&:to_s).sort.join('-')}"
+        "school-ids-#{(school_ids || []).sort.join('-')}"
       ]
     end
+
+    def self.grades_segment(grades)
+      return if grades.nil? || grades.empty?
```
brendanshean: Would `return "" if grades.blank?` apply here?

services/QuillLMS/app/lib/snapshots/cache_keys.rb
```diff
@@ -2,16 +2,41 @@
...
+
+      "grades-#{(grades || []).map(&:to_s).sort.join('-')}"
+    end
+
+    def self.teacher_ids_segment(teacher_ids)
+      return if teacher_ids.nil? || teacher_ids.empty?
```
brendanshean: Would `return "" if teacher_id.blank?` apply here?

services/QuillLMS/app/lib/snapshots/cache_keys.rb
```diff
@@ -2,16 +2,41 @@
...
+
+      "teacher-ids-#{(teacher_ids || []).sort.join('-')}"
+    end
+
+    def self.classroom_ids_segment(classroom_ids)
+      return if classroom_ids.nil? || classroom_ids.empty?
```
brendanshean: Would `return "" if classroom_ids.blank?` apply here?

services/QuillLMS/spec/queries/snapshots/period_query_spec.rb
```diff
@@ -0,0 +1,85 @@
+# frozen_string_literal: true
+
+require 'rails_helper'
+
+module Snapshots
+  describe PeriodQuery do
+
+    let(:test_period_query) do
+      Class.new(described_class) do
```
brendanshean, line 9: Cool!  I like this pattern.

services/QuillLMS/spec/controllers/snapshots_controller_spec.rb
brendanshean, line 1: I like the thorough testing in this file.

brendanshean (approved):
Nice work on pulling all of these filters/options together.

services/QuillLMS/app/controllers/snapshots_controller.rb
```diff
@@ -39,11 +38,48 @@ def top_x
...
+    School.joins(:schools_admins)
+      .where(schools_admins: {user_id: current_user.id})
+  end
+
+  private def teacher_options
+    school_ids = option_params[:school_ids]
+    grades = option_params[:grades]
+
+    filtered_schools = school_options
+    filtered_schools = filtered_schools.where(id: school_ids) unless school_ids.nil?
```
dandrabik: You should consider making `filtered_schools` or `filtered_school_ids` its own method to make this method less complex.

services/QuillLMS/app/controllers/snapshots_controller.rb
```diff
@@ -39,11 +38,48 @@ def top_x
...
+    teachers = teachers.where(classrooms: {grade: grades}) unless grades.nil?
+
+    teachers
+  end
+
+  private def classroom_options
+    teacher_ids = option_params[:teacher_ids]
+
+    filtered_teachers = teacher_options
+    filtered_teachers = filtered_teachers.where(id: teacher_ids) unless teacher_ids.nil?
```
dandrabik: You should consider making `filtered_teachers` or `filtered_teacher_ids` its own method to make this method less complex.

services/QuillLMS/app/controllers/snapshots_controller.rb
```diff
@@ -39,11 +38,48 @@ def top_x
...
+    grades = option_params[:grades]
+
+    filtered_schools = school_options
+    filtered_schools = filtered_schools.where(id: school_ids) unless school_ids.nil?
+
+    teachers = User.joins(:schools_users)
+      .joins(:classrooms_i_teach)
+      .where(schools_users: {school_id: filtered_schools.pluck(:id)})
+
+    teachers = teachers.where(classrooms: {grade: grades}) unless grades.nil?
```
dandrabik: take or leave, consider the positive framing of this, e.g. `if grades.present?`

services/QuillLMS/app/controllers/snapshots_controller.rb
dandrabik: take or leave, consider the positive framing of this, e.g. `if teacher_ids.present?`

services/QuillLMS/app/lib/snapshots/cache_keys.rb
dandrabik: `.blank?` is a short version of this line. I don't think you need the `grades || []` and this check. I believe that's redundant.

Also, you could consider not doing the nil check and the keys will all have the same sections, e.g. grades-teachers-classrooms.

services/QuillLMS/app/lib/snapshots/cache_keys.rb
```diff
@@ -2,16 +2,41 @@
...
       ]
     end
+
+    def self.grades_segment(grades)
+      return if grades.nil? || grades.empty?
+
+      "grades-#{(grades || []).map(&:to_s).sort.join('-')}"
+    end
+
+    def self.teacher_ids_segment(teacher_ids)
```
dandrabik: Consider a helper method since you do this logic 4 times in the file, e.g.

```ruby
def id_segment(name, ids)
  #
end

File: services/QuillLMS/app/queries/snapshots/period_query.rb (hunk 7 above)
Comment (dandrabik) [inline services/QuillLMS/app/queries/snapshots/period_query.rb]: consider `.blank?`

File: services/QuillLMS/app/workers/snapshots/cache_snapshot_count_worker.rb [hunk 17]
```diff
@@ -17,18 +17,16 @@ class CacheSnapshotCountWorker
...
       Rails.cache.write(cache_key, payload, expires_in: cache_expiry)
 
       PusherTrigger.run(user_id, PUSHER_EVENT,
         {
           query: query,
           timeframe: timeframe['name'],
-          school_ids: school_ids,
-          grades: grades
-        }
+        }.merge(filters)
```
dandrabik, line 30: The cache key logic separates out school_ids from additional filters, and this puts school_ids in filters. Might be better to follow the same pattern in both.

dandrabik (approved):
Small code cleanup comments, but overall looks good to me. Thanks for moving this along!

services/QuillLMS/app/controllers/snapshots_controller.rb
anathomical: Whoops!  Good catch!

services/QuillLMS/spec/queries/snapshots/period_query_spec.rb
anathomical: I originally just declared a new `class` in the test file, but Rubocop has a rule against that.  So this pattern is one I stole from the Rubocop documentation for that rule.  Good job, linter!

PR #10693, opened 2023-06-23
lms/more-vitally-script-tweaks

anathomical:
## WHAT
Remove nil values from batch array
## WHY
The code to avoid missing users resulted in the data file resulted in some `nil` values in the `batch.map` result
## HOW
Just `reject` nil values from the array

### Notion Card Links

PR Checklist | Your Answer
------------ | -------------
Have you added and/or updated tests? |  No
Have you deployed to Staging? | The data difference to between staging and prod has made this hard to test, but I have finally worked out a way to at least make sure that the script can read the entire file without choking, so I did that

### DIFF
--- a/services/QuillLMS/lib/tasks/vitally.rake
+++ b/services/QuillLMS/lib/tasks/vitally.rake
@@ -42,7 +42,7 @@ namespace :vitally do
             messageId: SecureRandom.uuid
           }
         end
-      end.flatten
+      end.flatten.reject { |payload| payload.nil? }
       batch_payload.each do |api_call_payload|
         puts "Unlinking user #{api_call_payload[:userId]} from account #{api_call_payload[:accountId]}"
       end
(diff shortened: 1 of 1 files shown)


happythenewsad (approved):
If you feel like there may be additional issues, you could try moving the meat of this rake task into a plain ruby class, which you can test more easily with RSpec. I did this pattern with universal_rule_loader_spec.rb, among others.

services/QuillLMS/lib/tasks/vitally.rake
```diff
@@ -42,7 +42,7 @@ namespace :vitally do
             messageId: SecureRandom.uuid
           }
         end
-      end.flatten
+      end.flatten.reject(&:nil?)
```
brendanshean: If you want to play code golf: `end.flatten.compact`
But this is a rake task so not a big deal.

brendanshean (approved):

PR #10705, opened 2023-06-27
Bs intermittent spec failures

brendanshean:
## WHAT
Fix a couple of intermittent spec failures.

## WHY
Intermittent spec failures slow the team down

## HOW
1. Since classes are lazy loaded, it's possible that the GoogleIntegration::TeacherClassroomsCache is not loaded before the base class and when the TeacherClassroomsCache is called.
2.  `Date.current` can be called multiple times and return the same value.  We need to set `last_login` to something earlier to ensure that Current.now is different from last_login.
Screenshot 2023-06-27 at 10 55 47 AM

### Screenshots

### Notion Card Links

PR Checklist | Your Answer
------------ | -------------

### DIFF
--- a/services/QuillLMS/app/services/google_integration/teacher_classrooms_retriever.rb
+++ b/services/QuillLMS/app/services/google_integration/teacher_classrooms_retriever.rb
@@ -25,7 +25,7 @@ module GoogleIntegration
     end
 
     private def cache_classrooms_data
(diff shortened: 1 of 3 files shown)


emilia-friedberg (approved):
Yay! Thanks for getting to these so fast.

PR #10706, opened 2023-06-27
Bug Fix - Fix issue with string key params.

dandrabik:
## WHAT
BugFix for [this issue](https://quillorg-5s.sentry.io/issues/4276632769/?project=11238&query=is%3Aunresolved&referrer=issue-stream&stream_index=1). Seems to be an issue with converting string params to keyword arguments.
## WHY
We want these reports to work.
## HOW
Replicated the issue in the console with params from a production failed job. Fixed that issue (see screenshot). Wrote a failing test and fixed it (note, there's stubbing in the test before the error, but replicating it's not receiving the same params).
### Screenshots
Replicated and fixed with params from a production failed job.
Screenshot 2023-06-27 at 12 50 31 PM

### Notion Card Links

PR Checklist | Your Answer
------------ | -------------
Have you added and/or updated tests? |  'YES'.
Have you deployed to Staging? | YES,
Self-Review: Have you done an initial self-review of the code below on Github? | Yes
Spec Review: Have you reviewed the spec and ensured this meets requirements and/or matches design mockups? | N/A

PR #10708, opened 2023-06-27
Fix same snapshot issue for other worker.

dandrabik:
## WHAT
This issue [fixed here](https://github.com/empirical-org/Empirical-Core/pull/10706) was happening in another place. Fixing it here as well.
## WHY
Fix this similar bugs for TopX reports load.
## HOW
symbolize params. Write a failing test and fix it. Confirm in console since there is a lot of stubbing in tests.

Screenshot 2023-06-27 at 1 47 35 PM

PR Checklist | Your Answer
------------ | -------------
Have you deployed to Staging? | No, I don't have a good way to test on staging.


brendanshean (approved):
@anathomical Makes me wonder long term if it's worth moving the filter.symbolize higher up so that it applies to all queries.

PR #10713, opened 2023-06-28
Add Client for Canvas REST API

brendanshean:
## WHAT
1.  Add a RestClient for [access](https://canvas.instructure.com/doc/api/) to the Canvas LMS, in contrast to the [GraphQL](https://canvas.instructure.com/doc/api/file.graphql.html) api.
2. Add a teacher_classrooms endpoint for the RestClient

## WHY
1.  The client gives access to much of a particular canvas Instance's data.
2. This particular endpoint is needed for Classroom rostering

## HOW
1. Add the lms-api Gem which is a wrapper to the REST API.  
2. Pull data from courses data from courses endpoint and then iterate over each course to find each section and that sections students.

### Screenshots

### Notion Card Links

PR Checklist | Your Answer
------------ | -------------


services/QuillLMS/spec/factories/auth_credentials.rb
```diff
@@ -30,7 +30,20 @@
     expires_at 1.day.from_now
     user
 
-    factory :canvas_auth_credential, parent: :auth_credential, class: :CanvasAuthCredential
+    factory :canvas_auth_credential, parent: :auth_credential, class: :CanvasAuthCredential do
```
brendanshean, line 33: This factory's complexity is related to the associated `canvas_instance_auth_credential` record, which in practice should always be created after a `canvas_auth_credential` creation.  However, this introduces [circularity](https://github.com/empirical-org/Empirical-Core/pull/10713/files#diff-db6f061c3f4219b39d595eccb79944da776260222746a15fd99940eb9d5b29e5R24) when unit testing `canvas_instance_auth_credential` hence the `canvas_auth_credential_without_canvas_instance_auth_credential` factory.  I'm not sure of a cleaner way to do this as factory bot doesn't seem to allow for conditional callbacks and I don't want to introduce two new traits for default behavior.  I'm definitely open to suggestions here.

anathomical: This is an awkward compromise, but I can't think of a cleaner way to do it offhand.

services/QuillLMS/spec/services/canvas_integration/classroom_data_adapter_spec.rb
```diff
@@ -0,0 +1,54 @@
...
+    it { expect(subject).to eq classroom_data }
+  end
+
+  def classroom_data
+    {
+      name: classroom_name,
+      external_classroom_id: section_data[:id],
+      students: students
+    }
+  end
```
anathomical: Thoughts on this just being a `let` call at the top?  The rest of the spec will look the same, but no one has to go look for this function definition.

anathomical (approved):
Question about spec organization, but this looks solid to me.

services/QuillLMS/spec/services/canvas_integration/classroom_data_adapter_spec.rb
brendanshean: Good idea!  Originally I had this as a function that passed parameters into but now I would agree that it should just be a let.

PR #10718, opened 2023-06-28
Lms/add no school assigned filter

anathomical:
## WHAT
Add a new "No grade set" option to the grades fillters in the admin Snapshot report
## WHY
Google classrooms never set grades, so we want to make it clear to admins that some of their classrooms may not have grades assigned, and allow them to filter for them explicitly.
## HOW
- Add an additional option to the grades option hash
- Update `Snapshots::PeriodQuery` to do special handling if the value `"null"` is in the list of grades.  (I considered just using actual `nil` here, but was worried there was probably too many assumptions in our code in multiple places that values would be truth-y.  Open to discussion on this.)
- Update the BigQuery TestRunner to treat `NULL` values as `NULL`
- Refactor a spec to work cleanly with the new `NULL` handling in the test runner

### Screenshots
image

### Notion Card Links
https://www.notion.so/quill/Add-None-selected-to-the-grade-filter-on-the-Usage-Snapshot-Report-e801ad2206874d08bf9eca1771337496

PR Checklist | Your Answer
------------ | -------------

### DIFF
--- a/services/QuillLMS/spec/support/shared/quill_big_query/test_runner.rb
+++ b/services/QuillLMS/spec/support/shared/quill_big_query/test_runner.rb
@@ -54,7 +54,7 @@ module QuillBigQuery
       attr_type = record.class.column_for_attribute(attr).type
 
       if value.nil?
-        "''"
+        "NULL"
       elsif value.is_a?(Array)
         value.map { |v| attr_type_value(attr_type, v) }
       else
--- a/services/QuillLMS/app/controllers/snapshots_controller.rb
+++ b/services/QuillLMS/app/controllers/snapshots_controller.rb
@@ -2,6 +2,7 @@
 
 class SnapshotsController < ApplicationController
   GRADE_OPTIONS = [
+    {value: "null", name: "No grade set"},
     {value: "Kindergarten", name: "Kindergarten"},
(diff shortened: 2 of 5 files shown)


services/QuillLMS/spec/support/shared/quill_big_query/test_runner.rb
```diff
@@ -54,7 +54,7 @@ def execute(query)
       attr_type = record.class.column_for_attribute(attr).type
 
       if value.nil?
-        "''"
+        "NULL"
```
brendanshean, line 57: At one [point](https://github.com/empirical-org/Empirical-Core/pull/10605/commits/53e6e1208538696274449cf0f01db2111c23c84c#diff-33db411271006d2a00f8922a18eee3429ddf7a8e545c472b0f36c835faf2a4e0R43) in building out the the test runner, I had `'NULL'` in here like this but couldn't figure out why it wasn't working properly (probably the disjunction with IS NULL).  Nice work on figuring it out.

brendanshean (approved):

brendanshean:
I liked the the HOW section of this PR description.

emilia-friedberg (approved):

PR #10729, opened 2023-07-02
Canvas Rostering Backend Part 1:  Google refactoring

brendanshean:
## WHAT
In preparation for Canvas rostering, this PR will standardize the three rostering endpoints within the context of Google Classrooms:

## 1. Retrieve Classrooms
check-google-retrieve-classrooms

## 2.  Import Classes
check-google-classroom-import

## 3.  Import Students
check-google-classroom-students-import
 

## WHY
Standardizing the endpoints for Google Classroom and Clever will provide a pattern to use for Canvas Rostering which in turn will lead to a sensible abstraction.

## HOW
Create a GoogleIntegration::TeachersController and do the following:

1.  Move classroom_managers/teachers#retrieve_google_classrooms to google_integration/teachers#retrieve_classrooms
1.  Move classroom_managers/teachers#update_google_classrooms to google_integration/teachers#import_classrooms
1.  Move classroom_managers/teachers#import_google_students to google_integration/teachers#import_students

The rest invovles some refactoring to get the controller looking as close as possible to the [CleverIntegration::TeachersController](https://github.com/empirical-org/Empirical-Core/blob/bs-google-rostering-backend-refactoring/services/QuillLMS/app/controllers/clever_integration/teachers_controller.rb).

### Screenshots

### Notion Card Links

PR Checklist | Your Answer
------------ | -------------


services/QuillLMS/app/workers/google_integration/import_classroom_students_worker.rb
```diff
@@ -0,0 +1,30 @@
...
+
+module GoogleIntegration
+  class ImportClassroomStudentsWorker
+    include Sidekiq::Worker
+    sidekiq_options queue: SidekiqQueue::CRITICAL_EXTERNAL
+
+    PUSHER_EVENT = 'google-classroom-students-imported'
+    PUSHER_FAILED_EVENT = 'google-account-reauthorization-required'
+
+    attr_reader :teacher_id
```
emilia-friedberg: What's the point of using the `attr_reader` and private method here when the teacher_id just gets passed into `perform`? Couldn't we just do `teacher ||= ::User.find(teacher_id)` in line 14?

emilia-friedberg (approved):
LGTM -- it'll be so nice to have all this unified! Just one question about an approach in a worker.

services/QuillLMS/app/workers/google_integration/hydrate_teacher_classrooms_cache_worker.rb
```diff
@@ -0,0 +1,17 @@
...
+module GoogleIntegration
+  class HydrateTeacherClassroomsCacheWorker
+    include Sidekiq::Worker
+    sidekiq_options queue: SidekiqQueue::CRITICAL_EXTERNAL
+
+    def perform(user_id)
+      user = ::User.find_by(id: user_id)
+
+      return if user.nil?
+      return unless user.teacher? && user.google_authorized?
```
dandrabik: Consider avoiding unless with multiple conditions

(side note, I think we should consider [tightening this cop](https://www.rubydoc.info/gems/rubocop/RuboCop/Cop/Style/UnlessLogicalOperators) to `forbid_logical_operators` but that's a team discussion.)

services/QuillLMS/app/controllers/google_integration/teachers_controller.rb
```diff
@@ -0,0 +1,72 @@
...
+module GoogleIntegration
+  class TeachersController < ApplicationController
+    before_action :authorize_owner!
+
+    def import_classrooms
+      run_student_importer(imported_classroom_ids)
+      delete_teacher_classrooms_cache
+      hydrate_teacher_classrooms_cache
+
+      render json: { classrooms: current_user.google_classrooms }.to_json
```
dandrabik: I think `to_json` is implicit, consider removing.

services/QuillLMS/app/controllers/google_integration/teachers_controller.rb
```diff
@@ -0,0 +1,72 @@
...
+      run_student_importer(selected_classroom_ids)
+
+      render json: { user_id: current_user.id }
+    end
+
+    def retrieve_classrooms
+      if !current_user.google_authorized?
+        render json: { user_id: current_user.id, reauthorization_required: true }
+      elsif serialized_classrooms_data
+        render json: JSON.parse(serialized_classrooms_data)
```
dandrabik: Do you need the `JSON.parse` here? Small item, but wondering if there is a way to avoid this (e.g. store as a hash maybe), since this line:  takes `json`, transforms it to a `hash`, then transforms that `hash` back to `json`.

services/QuillLMS/app/workers/google_integration/import_classroom_students_worker.rb
brendanshean: I usually default to finding of objects in their own memoized method.  That way if different methods call on that object, its found in one place.  However, in this case, you're right to point out that the memoization isn't really buying us much since there's only the `perform` method.

services/QuillLMS/app/controllers/google_integration/teachers_controller.rb
```diff
@@ -0,0 +1,72 @@
...
+      else
+        hydrate_teacher_classrooms_cache
+        render json: { user_id: current_user.id, quill_retrieval_processing: true }
+      end
```
anathomical: Purely stylistically: thoughts on structuring this around a pair of early returns instead of an if-else block?
```ruby
return render json: { user_id: current_user.id, reauthorization_required: true } unless current_user.google_authorized?
return render json: JSON.parse(serialized_classrooms_data) if serialized_classrooms_data

hydrate_teacher_classrooms_cache
render json: { user_id: current_user.id, quill_retrieval_processing: true }
```

services/QuillLMS/client/app/bundles/Teacher/components/classrooms/import_google_classroom_students_modal.tsx
```diff
@@ -30,12 +30,11 @@ export default class ImportGoogleClassroomStudentsModal extends React.Component<
     this.handleImportStudents = this.handleImportStudents.bind(this)
   }
 
-  initializePusherForGoogleStudentImport(id) {
-    if (process.env.RAILS_ENV === 'development') {
-      Pusher.logToConsole = true;
-    }
+  initializePusherForGoogleStudentImport(userId) {
+    if (process.env.RAILS_ENV === 'development') { Pusher.logToConsole = true; }
```
anathomical, line 34: Not worth changing in this PR since you're just changing formatting here, but I wonder if this is something that we can configure at a higher level.

anathomical (approved):
This feels cleaner and better organized to me.  Nice job.

services/QuillLMS/app/workers/google_integration/import_classroom_students_worker.rb
```diff
@@ -0,0 +1,24 @@
...
+    def perform(teacher_id, selected_classroom_ids = nil)
+      teacher = ::User.find(teacher_id)
+
+      if teacher.google_authorized?
+        TeacherClassroomsStudentsImporter.run(teacher, selected_classroom_ids)
+        PusherTrigger.run(teacher_id, PUSHER_EVENT, "Google classroom students imported for #{teacher_id}.")
+      else
+        PusherTrigger.run(teacher_id, PUSHER_FAILED_EVENT, "Reauthorization needed for user #{teacher_id}.")
+      end
+    rescue => e
```
dandrabik: Does this rescue prevent this retryable errors from being retried, e.g. http timeouts to google? Should we consider not rescuing here?

dandrabik (approved):
I left a few small take or leave comments, but overall looks good!

services/QuillLMS/app/controllers/google_integration/teachers_controller.rb
brendanshean: Good call.  I will update in Part 3 [where](https://github.com/empirical-org/Empirical-Core/pull/10731/files#diff-ece0c97f96ae128da5307cfc1a7961a8d0d9cf327b0cd3fb03be58ada1ea845bR6) this all gets abstracted away.

services/QuillLMS/client/app/bundles/Teacher/components/classrooms/import_google_classroom_students_modal.tsx
brendanshean: I'll make a note to take a look when I tackle the frontend work.

services/QuillLMS/app/controllers/google_integration/teachers_controller.rb
brendanshean: I will address in a later PR so that it applies to all three providers.

PR #10731, opened 2023-07-02
Canvas Rostering Backend Part 6: Canvas Rostering

brendanshean:
## WHAT
1.  Add CanvasController endpoints `import_classrooms', 'import_students' and 'retrieve_classrooms' via the same pattern in [Part 3](https://github.com/empirical-org/Empirical-Core/pull/10751) and [Part 4](https://github.com/empirical-org/Empirical-Core/pull/10753)
2.  Add processing for importing teacher_classrooms
3.  Add processing for importing classroom_students

## WHY
1.  Using the same endpoints make this code more flexible for future modifications
2.  Teacher classrooms data is used in the `import_classrooms` and `retrieve_classrooms` endpoints
3.   Classroom students data is used in the `import_students` endpoint

## HOW
1.  Use the same concerns from Google and Clever Rostering
2. Use the same processing for TeacherClassroomsData in Google and Clever
3. Use the same processing for ClassroomStudentsData in Google and Clever.

### Screenshots

### Notion Card Links

PR Checklist | Your Answer
------------ | -------------

### DIFF
--- a/services/QuillLMS/app/services/canvas_integration/classroom_updater.rb
+++ b/services/QuillLMS/app/services/canvas_integration/classroom_updater.rb
@@ -1,0 +1,60 @@
+# frozen_string_literal: true
+
+module CanvasIntegration
+  class ClassroomUpdater < ApplicationService
+    OWNER = ClassroomsTeacher::ROLE_TYPES[:owner].freeze
+    COTEACHER = ClassroomsTeacher::ROLE_TYPES[:coteacher].freeze
+
+    attr_reader :classroom, :data, :teacher_id
+
+    def initialize(classroom, data)
+      @classroom = classroom
+      @data = data
+      @teacher_id = data[:teacher_id]
+    end
+
+    def run
+      update
+      associate_teacher_and_classroom
+      classroom.reload
+    end
+
+    private def associate_teacher_and_classroom
+      return if ClassroomsTeacher.exists?(classroom: classroom, user_id: teacher_id)
+
+      ClassroomsTeacher.create!(
+        classroom: classroom,
+        user_id: teacher_id,
+        role: classroom.owner ? COTEACHER : OWNER
+      )
+    end
+
+    private def custom_name?
+      classroom.name != classroom.synced_name
+    end
+
+    private def name
+      custom_name? ? classroom.name : valid_name
+    end
+
+    private def other_owned_classroom_names
+      teacher.classrooms_i_own.reject { |c| c.id == classroom.id }.pluck(:name)
+    end
+
+    private def synced_name
+      data[:name]
+    end
+
+    private def teacher
+      ::User.find(teacher_id)
+    end
+
+    private def update
+      classroom.update!(name: name, synced_name: synced_name, visible: true)
+    end
+
+    private def valid_name
+      ::DuplicateNameResolver.run(data[:name], other_owned_classroom_names)
+    end
+  end
+end
--- a/services/QuillLMS/app/workers/canvas_integration/import_teacher_classrooms_students_worker.rb
+++ b/services/QuillLMS/app/workers/canvas_integration/import_teacher_classrooms_students_worker.rb
@@ -1,0 +1,24 @@
+# frozen_string_literal: true
+
+module CanvasIntegration
+  class ImportTeacherClassroomsStudentsWorker
+    include Sidekiq::Worker
+    sidekiq_options queue: SidekiqQueue::CRITICAL_EXTERNAL
+
+    PUSHER_EVENT = 'canvas-classroom-students-imported'
+    PUSHER_FAILED_EVENT = 'canvas-account-reauthorization-required'
+
+    def perform(teacher_id, selected_classroom_ids = nil)
+      teacher = ::User.find(teacher_id)
+
+      if teacher.canvas_authorized?
+        TeacherClassroomsStudentsImporter.run(teacher, selected_classroom_ids)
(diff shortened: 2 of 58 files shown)


services/QuillLMS/app/services/canvas_integration/classroom_updater.rb
```diff
@@ -0,0 +1,60 @@
...
+      classroom.reload
+    end
+
+    private def associate_teacher_and_classroom
+      return if ClassroomsTeacher.exists?(classroom: classroom, user_id: teacher_id)
+
+      ClassroomsTeacher.create!(
+        classroom: classroom,
+        user_id: teacher_id,
+        role: classroom.owner ? COTEACHER : OWNER
```
anathomical, line 28: I'm guessing that Canvas doesn't really pass us the information we need to actually assign roles in a consistent way, so making the first teacher to import a class the "owner" seems reasonable to me.

services/QuillLMS/app/services/canvas_integration/classroom_updater.rb
```diff
@@ -0,0 +1,60 @@
...
+      )
+    end
+
+    private def custom_name?
+      classroom.name != classroom.synced_name
+    end
+
+    private def name
+      custom_name? ? classroom.name : valid_name
+    end
```
anathomical, line 38: Is there a weird edge case where User A imports a classroom which does not have a name conflict, and thus doesn't need a custom name (so `custom_name?` is `false`, but that User B, a co-teacher, imports the class, triggering the updater, and DOES have a name conflict which causes the name to change out from under User A?  

This happening probably wouldn't be a big deal since the `DuplicateNameResolver` shouldn't change the name too much.  However, when User A syncs the classroom again won't it re-run through the `DuplicateNameResolver` again, but with different `other_owned_classroom_names` which will reset the name to the original?  So you could end up in a situation where the classroom name switches back and forth based on who last synced it?

Again, the edge case may be so rare that we don't need to bother fixing it, but felt worth checking on.

services/QuillLMS/app/workers/canvas_integration/import_teacher_classrooms_students_worker.rb
```diff
@@ -0,0 +1,24 @@
+# frozen_string_literal: true
+
+module CanvasIntegration
+  class ImportTeacherClassroomsStudentsWorker
+    include Sidekiq::Worker
+    sidekiq_options queue: SidekiqQueue::CRITICAL_EXTERNAL
+
+    PUSHER_EVENT = 'canvas-classroom-students-imported'
+    PUSHER_FAILED_EVENT = 'canvas-account-reauthorization-required'
```
anathomical, line 9: As with my note on another PR, is there front-end code set up to listen for these Pusher notifications?  (That may be a future thing, and not necessary yet, I just want to make sure that it's on the radar.)

services/QuillLMS/app/workers/canvas_integration/import_teacher_classrooms_students_worker.rb
```diff
@@ -0,0 +1,24 @@
...
+
+    def perform(teacher_id, selected_classroom_ids = nil)
+      teacher = ::User.find(teacher_id)
+
```
anathomical, line 13: No early return if `teacher.nil?`?  Maybe not necessary, but it's a pattern we've used in a lot of other places.

services/QuillLMS/spec/models/provider_classroom_delegator_spec.rb
```diff
@@ -70,4 +70,40 @@
...
+        user_external_id: student1.user_external_id(canvas_instance: canvas_instance)
+      )
+
+      create(
+        :canvas_classroom_user,
+        :unsynced,
+        classroom_external_id: classroom.classroom_external_id,
+        user_external_id: student2.user_external_id(canvas_instance: canvas_instance)
+      )
+    end
```
anathomical, line 97: Interesting to see this pattern as opposed to `let!`.  I don't know that I have a preference for either approach, but it's neat to see this.

anathomical (approved):
Looks good.  Excited to see the fruit of your very extensive labor.

happythenewsad (approved):
👍

services/QuillLMS/spec/models/provider_classroom_delegator_spec.rb
brendanshean: I tend to use before blocks over let! if I don't use the let! variable anywhere.

services/QuillLMS/app/workers/canvas_integration/import_teacher_classrooms_students_worker.rb
brendanshean: That's a good question. My intent here is to actually have a raised exception vs silent failure.  Maybe something to discuss in engineering.

services/QuillLMS/app/workers/canvas_integration/import_teacher_classrooms_students_worker.rb
brendanshean: Thank you for raising this.  In this case, there isn't canvas frontend code yet.

services/QuillLMS/app/services/canvas_integration/classroom_updater.rb
brendanshean: This is a good call out.  I will check again with some test data.

PR #10736, opened 2023-07-06
new version of the update assigned students page

emilia-friedberg:
## WHAT
New version of the update assigned students page that makes it easier for teachers to see what is changing.

## WHY
We get a large number of support requests due to teachers accidentally unassigning work from their students. Hopefully these design changes will help with that.

## HOW
Largely just rebuilding the frontend (which was so old and so satisfying to get rid of!), plus some controller endpoints for the new milestone (choosing not to see the "unassign warning modal") and to revert an inadvertent unassignment. Note: I chose to just make a special endpoint that handled that behavior for that function, since it is pretty straightforward and would have been a much larger change if I'd had to update the existing `Units::Updater` to make the session visible again. However, I am open to changing that if people feel strongly that it should follow the same flow.

### Screenshots
Screen Shot 2023-07-06 at 4 12 49 PM
Screen Shot 2023-07-06 at 4 13 04 PM
Screen Shot 2023-07-06 at 4 16 02 PM

### Notion Card Links
https://www.notion.so/quill/My-Activities-Page-Page-to-Add-Remove-Students-from-Pack-0d5c9696446d4f0ab3529b539ab05411?pvs=4
https://www.notion.so/quill/Q2-2023-OKR-My-Activities-Page-Page-to-Add-Remove-Students-from-Pack-55f84c50ecd842f49bda3312520ed1a3?pvs=4

PR Checklist | Your Answer
------------ | -------------


services/QuillLMS/app/models/classroom.rb
```diff
@@ -168,7 +168,7 @@ def hide_all_classroom_units
   end
 
   def with_students
-    attributes.merge({students: students})
+    attributes.merge({students: students.sort_by { |s| s.last_name }})
```
brendanshean: The outermost hash can be removed, but also you can use pretzel colon shorthand for last_name:

`attributes.merge(students: students.sort_by(&:last_name))`

services/QuillLMS/app/controllers/teachers/units_controller.rb
```diff
@@ -54,14 +54,29 @@ def update_classroom_unit_assigned_students
...
+    classroom_unit = ClassroomUnit.find_by_id(params[:classroom_unit_id])
+
+    new_assigned_student_ids = classroom_unit
+      .assigned_student_ids
+      .union([params[:student_id].to_i])
+      .sort
+
+    classroom_unit.update(assigned_student_ids: new_assigned_student_ids)
+
+    ActivitySession.unscoped.where(user_id: params[:student_id], classroom_unit_id: params[:classroom_unit_id]).update(visible: true)
```
brendanshean: Consider rewriting the query over multiple lines as it makes it easier to scan:

```
ActivitySession
  .unscoped
  .where(user_id: params[:student_id], classroom_unit_id: params[:classroom_unit_id])
  .update(visible: true)
```

brendanshean (approved):
Thanks for this UI/UX improvement!

services/QuillLMS/app/assets/stylesheets/pages/lesson_planner_and_unit_template_editor.scss
```diff
@@ -353,15 +353,7 @@
...
-  .unselecting-students-note {
-    background-color: #ffefef;
-    border: solid 1px rgba(255, 115, 112, 0.72);
-    i, svg {
-      color: #ff4542;
-    }
-  }
-
-  section:not(.activity-row):not(.selected-activity-row):not(.filter-and-sort) {
+  section:not(.activity-row):not(.selected-activity-row):not(.filter-and-sort):not(.classroom-section) {
```
anathomical: I don't think that we should update our syntax for this right now, but I'm wondering if the newer comma-separated syntax (introduced in CSS Selectors Level 4) would be more readable in cases where we use `not`:
```
section:not(.activity-row, .selected-activity-row, .filter-and-sort, .classroom-section)
```

services/QuillLMS/app/controllers/teachers/classroom_manager_controller.rb
```diff
@@ -21,6 +21,8 @@ class Teachers::ClassroomManagerController < ApplicationController
 
   def lesson_planner
     set_classroom_variables
+    unassign_warning_milestone = Milestone.find_by_name(Milestone::TYPES[:dismiss_unassign_warning_modal])
+    @unassign_warning_hidden = UserMilestone.find_by(milestone_id: unassign_warning_milestone&.id, user_id: current_user&.id)
```
anathomical: If I'm reading this right, we really just need a boolean value here.  If that's the case we can use `exists?` in place of `find_by`.  It's trivially more performant, but importantly just returns a boolean.

services/QuillLMS/app/controllers/teachers/classroom_manager_controller.rb
```diff
@@ -21,6 +21,8 @@ class Teachers::ClassroomManagerController < ApplicationController
 
   def lesson_planner
     set_classroom_variables
+    unassign_warning_milestone = Milestone.find_by_name(Milestone::TYPES[:dismiss_unassign_warning_modal])
```
anathomical: Given that we expect this to effectively be a constant (i.e. it may be different in different environments, but it should be always be the same in a specific environment), we should move this into a class constant within the controller.  So, like, on line 21 just do `UNASSIGN_WARNING_MILESTONE = Milestone.find_by_name(Milestone::TYPES[:dismiss_unassign_warning_modal])`.

The big advantage of doing this is that the database lookup will happen the first time the controller loads, and then the result will be held in memory for the lifetime of the controller.  This lets us avoid doing a database lookup for this value every time the action runs.

services/QuillLMS/app/controllers/teachers/units_controller.rb
```diff
@@ -54,14 +54,29 @@ def update_classroom_unit_assigned_students
     activities_data = UnitActivity.where(unit_id: params[:id]).order(:order_number).pluck(:activity_id).map { |id| { id: id } }
 
     if activities_data.any?
-      classroom_data = JSON.parse(params[:unit][:classrooms], symbolize_names: true)
+      classroom_data = params[:unit][:classrooms].as_json.map(&:symbolize_keys)
```
anathomical: I think that `deep_symbolize_keys` will get you the same result as `map`, though maybe we only want specific values symbolized and not others, in which case just keep what you've got here.  You could keep it anyway since I don't think there's a significant difference, but for whatever reason `deep_symbolize_keys` reads as "easier to understand" to me.

services/QuillLMS/app/controllers/teachers/units_controller.rb
```diff
@@ -54,14 +54,29 @@ def update_classroom_unit_assigned_students
...
       render json: {errors: 'Unit can not be found'}, status: 422
     end
   end
 
+  def restore_classroom_unit_assignment_for_one_student
+    classroom_unit = ClassroomUnit.find_by_id(params[:classroom_unit_id])
+
+    new_assigned_student_ids = classroom_unit
+      .assigned_student_ids
+      .union([params[:student_id].to_i])
```
anathomical: I think that the syntax here is a little cleaner if we use `.push` instead of `.union`, since `push` doesn't require you to wrap the data in an array.
```ruby
.assigned_student_ids
.push(params[:student_id].to_i)
```

services/QuillLMS/spec/controllers/milestones_controller_spec.rb
```diff
@@ -79,6 +79,16 @@
     end
   end
 
+  describe '#complete_dismiss_unassign_warning_modal' do
+    let!(:milestone) { create(:dismiss_unassign_warning_modal) }
+
+    it 'should push the milestone into users milestones' do
+      expect(user.milestones).to_not include milestone
+      post :complete_dismiss_unassign_warning_modal
+      expect(user.milestones).to include milestone
```
anathomical: I think that this may be a cleaner spec if we use `expect { ... }.to change`.
```
expect {
  post :complete_dismiss_unassign_warning_modal
}.to change(user.milestones).from([]).to([milestone])
```
Note: I'm not sure that's exactly the proper `from` and `to` value, but given that these are clean tests I'm guessing new users are expected to have no milestones.

services/QuillLMS/spec/controllers/teachers/classroom_manager_controller_spec.rb
```diff
@@ -20,6 +21,7 @@
       expect(assigns(:tab)).to eq "test tab"
       expect(assigns(:grade)).to eq "test grade"
       expect(assigns(:students)).to eq user.students.any?
+      expect(assigns(:unassign_warning_hidden)).to eq UserMilestone.find_by(milestone_id: milestone.id, user_id: user.id)
```
anathomical: If you take my `exists?` suggestion in the controller, you'll need to update this test, too.

anathomical (approved):
Some syntax notes on the Ruby code, but the front-end stuff looks good to me.  So glad to see this section of our website get some modern code.

services/QuillLMS/app/assets/stylesheets/pages/lesson_planner_and_unit_template_editor.scss
emilia-friedberg: This is a good find. I was actually thinking as I added it that I didn't like the syntax, but it didn't occur to me to look into a better way of doing it. Thank you.

services/QuillLMS/app/controllers/teachers/units_controller.rb
emilia-friedberg: `.deep_symbolize_keys` is a hash method, not an array method, so I updated it to `.map(&:deep_symbolize_keys)` here.

eadams17 (approved):

PR #10739, opened 2023-07-06
Lms/tweak snapshot timeframe logic

anathomical:
## WHAT
Tweak the way that we calculate and pass timeframes around 
## WHY
In the original design, things were set up such that the "previous period" and "current period" being compared in `Count` queries were a fixed number of days on either side of a fixed date.  That is, when looking at "this month" on the 7th day of the month, the timeframe would compare the first seven days of the current month against the seven days immediately preceding the start of that time (so the last seven days of the previous month).

Further discussion with product has us shifting that definition so that for "this month" or "this year" we want to compare the first X days of the current period to the first X days of the previous period.  So If it's July 7, and you select "this month" as your period, you want to compare July 1-7 to June 1-7 (instead of to June 24-30).
## HOW
- Update the `Snapshots::Timeframes` lib so that i calculates a `previous_end` value for each timeframe
- Update the way we calculate Snapshot cache keys to accommodate the new timeframe logic
- Update the `Count` worker to take the new timeframe (note that `TopX` queries don't care about previous timeframes, so that code doesn't need updating)
- Update the Snapshots controller to pass the new timeframe data to the worker

### Notion Card Links
https://www.notion.so/quill/On-the-Usage-Snapshot-Report-change-the-way-the-increase-or-decrease-is-calculated-for-This-mont-ef92a6aff8dd4e20b6d3eb5cb8d267b8#896228ada74c474c9e3f146ef03e01f6

PR Checklist | Your Answer
------------ | -------------

### DIFF
--- a/services/QuillLMS/app/controllers/snapshots_controller.rb
+++ b/services/QuillLMS/app/controllers/snapshots_controller.rb
@@ -118,10 +118,10 @@ class SnapshotsController < ApplicationController
 
   private def retrieve_cache_or_enqueue_worker(worker)
 
-    previous_start, current_start, current_end = Snapshots::Timeframes.calculate_timeframes(snapshot_params[:timeframe],
+    previous_start, previous_end, current_start, current_end = Snapshots::Timeframes.calculate_timeframes(snapshot_params[:timeframe],
       snapshot_params[:timeframe_custom_start],
       snapshot_params[:timeframe_custom_end])
-    cache_key = cache_key_for_timeframe(previous_start, current_start, current_end)
+    cache_key = cache_key_for_timeframe(snapshot_params[:timeframe], current_start, current_end)
     response = Rails.cache.read(cache_key)
 
     return { results: response } if response
@@ -145,7 +145,7 @@ class SnapshotsController < ApplicationController
     { message: 'Generating snapshot' }
   end
 
-  private def cache_key_for_timeframe(previous_start, current_start, current_end)
+  private def cache_key_for_timeframe(previous_start, previous_end, current_start, current_end)
 
     Snapshots::CacheKeys.generate_key(@query,
       previous_start,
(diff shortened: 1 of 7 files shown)


brendanshean (approved):
Nice work on adjusting this to accommodate the updated requirement.

services/QuillLMS/app/controllers/snapshots_controller.rb
```diff
@@ -127,10 +127,10 @@ def options
 
   private def retrieve_cache_or_enqueue_worker(worker)
 
-    previous_start, current_start, current_end = Snapshots::Timeframes.calculate_timeframes(snapshot_params[:timeframe],
-      snapshot_params[:timeframe_custom_start],
-      snapshot_params[:timeframe_custom_end])
-    cache_key = cache_key_for_timeframe(previous_start, current_start, current_end)
+    previous_start, previous_end, current_start, current_end = Snapshots::Timeframes.calculate_timeframes(snapshot_params[:timeframe],
+      custom_start: snapshot_params[:timeframe_custom_start],
```
happythenewsad, line 131: +1 nicer method signature

services/QuillLMS/app/lib/snapshots/cache_keys.rb
```diff
@@ -2,19 +2,19 @@
...
+    def self.generate_key(query, timeframe_name, current_start, current_end, school_ids, additional_filters: {})
+      root_key(query, timeframe_name, current_start, current_end, school_ids)
         .append(grades_segment(additional_filters[:grades]))
         .append(teacher_ids_segment(additional_filters[:teacher_ids]))
         .append(classroom_ids_segment(additional_filters[:classroom_ids]))
         .compact
     end
 
-    def self.root_key(query, previous_start, current_start, current_end, school_ids)
+    def self.root_key(query, timeframe_name, current_start, current_end, school_ids)
```
happythenewsad, line 13: Not an issue now, but we might have to care about cache-warming these sorts of key expirations in the future, as they become highly trafficked.

happythenewsad (approved):

PR #10742, opened 2023-07-07
Lms/admin snapshot backend milestone5

anathomical:
## WHAT
Add final six queries for the Admin Snapshots project and hook them up to the front-end
## WHY
So that we can show all of our snapshot widgets with no "Coming Soon" stuff
## HOW
- Some refactoring around "Average" queries so that the pattern is more re-usable
- Added some new TopX rollups along with specs for them
- Removed the `comingSoon` flags on the front-end so that the new queries begin working

### Notion Card Links
https://www.notion.so/quill/Technical-Review-Usage-Snapshot-Admin-Report-afa1f08d50524c81a758d32899f97ae4

PR Checklist | Your Answer
------------ | -------------


services/QuillLMS/app/queries/snapshots/average_query.rb
```diff
@@ -0,0 +1,41 @@
...
+    def query
+      ''
+    end
+
+    def run
+      # This query, and both of the queries below, inherit their `initialize` method from
+      # PeriodQuery.  This means that they should all have identical method signatures.
+      numerator = numerator_query.run(*params)[:count]
+      denominator = denominator_query.run(*params)[:count]
+      # max prevents us from dividing by 0 if there are no active students
```
brendanshean: Maybe update to `# max prevents us from dividing by 0 (e.g. if there are no active students)` since the denominator is generic.

services/QuillLMS/app/queries/snapshots/average_query.rb
```diff
@@ -0,0 +1,41 @@
...
+      # We pull params here because we're going to do a naive pass-through to sibling queries below
+      @params = params
+
+      super
+    end
+
+    # Adding a no-op overwrite of query to avoid confusion in debugging where this class
+    # might assemble a query even though it never intends to run one because it pulls data
+    # from other queries instead
+    def query
```
happythenewsad, line 14: You're right to call out this difference. This class's behavior is different enough from the others that I'd like to consider inheriting from a new parent class, `ComposedQuery`, which has no `query` method.

services/QuillLMS/app/queries/snapshots/average_query.rb
```diff
@@ -0,0 +1,41 @@
...
+    end
+
+    def run
+      # This query, and both of the queries below, inherit their `initialize` method from
+      # PeriodQuery.  This means that they should all have identical method signatures.
+      numerator = numerator_query.run(*params)[:count]
+      denominator = denominator_query.run(*params)[:count]
+      # max prevents us from dividing by 0 if there are no active students
+      {
+        'count': numerator / [denominator, 1].max.to_f
```
happythenewsad: Kind of a nit, but we have this "avoid zero division" helper in `AverageActivitiesCompletedPerStudentQuery` as well. Good opportunity to abstract into a shared helper method.

happythenewsad (commented):
Looking good. I'll review after a small discussion, below

happythenewsad (approved):

services/QuillLMS/app/queries/snapshots/average_query.rb
brendanshean: I concur with Peter's suggestion of extracting out a new `ComposeQuery` class.

brendanshean (approved):
This looks great.   Nice work on figuring out some new query shapes.

PR #10743, opened 2023-07-07
Lms/remove snapshot coming soon

anathomical:
## WHAT
Remove the "Coming Soon" logic from the front-end of the Admin Snapshots code
## WHY
With #10742 all of the queries will be running in the back-end, and thus there will be no more "coming soon" widgets.  While this code may be useful again at some point in the future, it feels better to me to remove it for the moment given how lightweight it is and that we don't KNOW that we need it.

That said, I'm doing so in a separate PR so that it's super easy to revert if we need to.
## HOW
- Stop passing the `comingSoon` value between components since there are no longer any components where that value is set
- Remove logic that checks `comingSoon` values since they don't exist anymore

PR Checklist | Your Answer
------------ | -------------
Have you deployed to Staging? | Small change

### DIFF
--- a/services/QuillLMS/app/assets/stylesheets/pages/premium_hub/usage-snapshot-report.scss
+++ b/services/QuillLMS/app/assets/stylesheets/pages/premium_hub/usage-snapshot-report.scss
@@ -179,12 +179,6 @@
           font-size: 18px;
           line-height: 22px;
         }
-        .coming-soon {
-          font-weight: 700;
-          font-size: 18px;
-          line-height: 22px;
-          color: #757575;
-        }
         .count-and-label {
           display: flex;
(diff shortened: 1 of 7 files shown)


emilia-friedberg (approved):
LGTM! This isn't strictly necessary but for the purposes of really cleaning up this code, it probably also makes sense to delete the `.coming-soon` CSS rules.

anathomical:
> LGTM! This isn't strictly necessary but for the purposes of really cleaning up this code, it probably also makes sense to delete the `.coming-soon` CSS rules.

Good call.  I'll go ahead and do that now since it's top of mind.

eadams17 (approved):

cissyxyu (approved):

PR #10744, opened 2023-07-10
Move SetImpactsMetrics Queries to BigQuery

cissyxyu:
## WHAT
Move the queries in SetImpactMetricsWorker to BigQuery.
Relatedly, refactor BigQuery Runner to optionally accept parameters in the form of an array of strings or integers.

## WHY
We want to speed up the SetImpactMetricsWorker by moving it to BigQuery.
Because one of the queries in SetImpactMetricsWorker uses a very long array of IDs as a parameter in the SQL query, we need to modify our BigQuery Runner logic to format this query such that BigQuery will accept it. 

## HOW
Write query classes for all three of the new queries we need to run in BigQuery.
In the BigQuery Runner class, pre-process the query through a new class, PreTransformer, that will format the query correctly accounting for optional array parameters.

### Screenshots

### Notion Card Links
https://www.notion.so/quill/Move-SetImpactMetricsWorker-to-BigQuery-75b35d8d2f8741b5a25d4e0c45db682a?pvs=4

PR Checklist | Your Answer
------------ | -------------


services/QuillLMS/app/services/quill_big_query/pre_transformer.rb
```diff
@@ -0,0 +1,57 @@
...
+
+    def self.type_lookup(object)
+      raise UnsupportedArrayParameterTypeError unless object.is_a?(String) || object.is_a?(Integer)
+
+      return STRING_TYPE if object.is_a? String
+
+      return INT_TYPE
+    end
+
+    def transformed_query
```
anathomical: It feels like we should maybe just call this `transform` to maintain the naming pattern established in the post transformer.

services/QuillLMS/app/services/quill_big_query/pre_transformer.rb
```diff
@@ -0,0 +1,57 @@
...
+    end
+
+    # according to Google Cloud BigQuery API as documented here:
+    # https://cloud.google.com/bigquery/docs/reference/rest/v2/QueryParameter
+    def self.build_array_query_params(*array_params)
+      array_params.map do |array_param|
+        raise EmptyArrayParameterError if array_param.values.first.first.blank?
+
+        name = array_param.keys.first
+        array = array_param.values.first
```
anathomical: Given that we're passing hashes around for `array_params`, I think we can simplify this code if we use double splat (`**`) instead of single splat (`*`) when defining all of these function signatures.  If we use `**array_params` (with maybe a different name such as `**query_params`?) everywhere, then it will be a hash instead of an array and we can do:
```ruby
array_params.map do |name, values|
  type = type_lookup(values.first)
  {
    'name' => name.to_s,
   ...
  'parameterValue' => {
    'arrayValues' => values.map { |v| {'value' => v.to_s} }
  }
}
```

services/QuillLMS/app/services/quill_big_query/pre_transformer.rb
```diff
@@ -0,0 +1,57 @@
...
+      end
+    end
+
+    def self.type_lookup(object)
+      raise UnsupportedArrayParameterTypeError unless object.is_a?(String) || object.is_a?(Integer)
+
+      return STRING_TYPE if object.is_a? String
+
+      return INT_TYPE
+    end
```
anathomical, line 45: Not sure of an easy way to do this, but I wonder if we can link the `ARRAY_TYPE`, `STRING_TYPE`, and `INT_TYPE` constants to the values in the `PostTransformer::TYPE_TO_LAMBDA_LOOKUP` constant.  The keys in that hash should line up with the various types BigQuery supports.  May not be worth struggling with in this PR, but it'd be nice to establish a link between the two sides of this logic if we can figure out an easy way to do it.

services/QuillLMS/app/services/quill_big_query/pre_transformer.rb
```diff
@@ -0,0 +1,57 @@
...
+
+    def transformed_query
+      body_object = {
```
anathomical: No need to name our returned object here since we're just raw returning it.  So we can dump `body_object =`

services/QuillLMS/app/services/quill_big_query/runner.rb
```diff
@@ -27,20 +27,22 @@ def self.transform_response(json_body)
...
 
+      body_object = QuillBigQuery::PreTransformer.new(query, *array_params).transformed_query
+
       # API reference: https://cloud.google.com/bigquery/docs/reference/rest/v2/jobs/query?apix_params=%7B%22projectId%22%3A%22analytics-data-stores%22%2C%22resource%22%3A%7B%22query%22%3A%22SELECT%20email%20from%20lms.users%20LIMIT%201%22%7D%7D
       begin
         result = client.execute(
           api_method: bigquery.jobs.query,
           parameters: {'projectId' => PROJECT_ID},
-          body_object: {'query' => query, 'useLegacySql' => false}
+          body_object: body_object
```
anathomical: Since we don't reference `body_object` anywhere in this method except for this one line, we can skip saving the value to a variable and just call the transformer in this spot:
```ruby
result = client.execute(
  ...
  body_object: QuillBigQuery::PreTransformer.new(query, *array_params).transformed_query
)
```

services/QuillLMS/app/services/quill_big_query/runner.rb
```diff
@@ -49,8 +51,13 @@ def self.get_response(query)
...
     end
 
-    def self.execute(query)
-      response = get_response(query)
+    # NOTE: array_params must be an array of integers or strings, e.g.
+    # {teacher_ids: [1,2,3,4,5]}
+    # because limitations on query length in BigQuery require us to specially format array params.
+
+    # other types of params should be written directly inside the query string, not passed in as array_params
+    def self.execute(query, *array_params)
```
anathomical: See my other note on single vs double splats.

services/QuillLMS/app/workers/set_impact_metrics_worker.rb
```diff
@@ -3,33 +3,30 @@
...
+    low_income_schools_query = schools_query.select { |school| school["free_lunches"].present? && school["free_lunches"] > FREE_LUNCH_MINIMUM}
 
     number_of_sentences = finished_activity_sessions_query.count.floor(-5) * 10
     $redis.set(PagesController::NUMBER_OF_SENTENCES, number_of_sentences)
     number_of_students = finished_activity_sessions_query.count("DISTINCT(user_id)").floor(-5)
     $redis.set(PagesController::NUMBER_OF_STUDENTS, number_of_students)
     number_of_teachers = teacher_ids.size.floor(-2)
     $redis.set(PagesController::NUMBER_OF_TEACHERS, number_of_teachers)
-    number_of_schools = schools_query.count
+    number_of_schools = schools_query.size
```
anathomical: Not a huge deal, but since `schools_query` should be an array, let's standardize on `length` for getting the size of an array.  (In Ruby, `Array.size` is just an alias `Array.length` anyway, so we might as well call `length` so that we do that everywhere.)

services/QuillLMS/app/workers/set_impact_metrics_worker.rb
```diff
@@ -3,33 +3,30 @@
...
-    teacher_ids = teachers_query.to_a.map(&:id)
-    schools_query = School.joins(:schools_users).where(schools_users: {user_id: teacher_ids}).distinct
-    low_income_schools_query = schools_query.where("free_lunches > ?", FREE_LUNCH_MINIMUM)
+    finished_activity_sessions_query = QuillBigQuery::ActivitiesAllTimeQuery.run
+
+    teachers_query = QuillBigQuery::ActiveTeachersAllTimeQuery.run
+    teacher_ids = teachers_query.to_a.map {|teacher| teacher["id"]}
+
+    schools_query = QuillBigQuery::SchoolsContainingCertainTeachersQuery.run(teacher_ids)
+    low_income_schools_query = schools_query.select { |school| school["free_lunches"].present? && school["free_lunches"] > FREE_LUNCH_MINIMUM}
```
anathomical: Since these variables store the result of the queries, we should probably name them after the data they have.  So `finished_activity_sessions` instead of `finished_activity_sessions_query`, and so on.

services/QuillLMS/app/queries/quill_big_query/active_teachers_all_time_query.rb
```diff
@@ -0,0 +1,37 @@
+# frozen_string_literal: true
+
+module QuillBigQuery
+  class ActiveTeachersAllTimeQuery < ::QuillBigQuery::Query
```
anathomical: This is an organizational thing more than anything else, but I sort of feel like the three new queries we're adding here belong in their own module rather than in the generic `QuillBigQuery` module.  Sort of like we collect all of the Admin Snapshots queries in the `Snapshots` module.  Not sure what the appropriate name for said module would be.  Maybe `ImpactMetrics` since that's what the worker that uses them is called?  (And if we're going to make a module for that, maybe we want to put the worker into that module, too?  Don't feel super-strong about that.)

services/QuillLMS/app/workers/set_impact_metrics_worker.rb
```diff
@@ -3,33 +3,30 @@
...
 
     number_of_sentences = finished_activity_sessions_query.count.floor(-5) * 10
```
anathomical: This works because `finished_activity_sessions_query` is an array, and in Rails `count` with no args on an array works like `length`.  But this could be way faster if you just ran the `COUNT(*)` in the query rather than pulling all the `activity_sessions.id` values and then counting them in Ruby.  That's super-inefficient.  So I think you need to modify the BigQuery query to just return the `count` number.

But if you do that, see my note below...

services/QuillLMS/app/workers/set_impact_metrics_worker.rb
```diff
@@ -3,33 +3,30 @@
...
     $redis.set(PagesController::NUMBER_OF_SENTENCES, number_of_sentences)
     number_of_students = finished_activity_sessions_query.count("DISTINCT(user_id)").floor(-5)
```
anathomical: So because `finished_activity_sessions_query` is now an array, not an `ActiveRecord` query object, `count("DISTINCT(user_id)")` doesn't actually do what you want anymore.  In Rails, `count('foo')` on an array means: `select { |v| v == 'foo' }.length`.  So this will attempt to find the number of elements in an array of `activity_sessions.id` values that are equal to the literal `DISTINCT(user_id)`.  Which, while syntactically valid (and thus not throwing an error), will always be none of them, meaning that we aren't getting the data we want.

Because you're basically counting a different thing, I think you'll need to do a separate `QuillBigQuery::Query` sub-class explicitly for this value rather than being able to use one query for total count of ActivitySessions and a separate one for distinct Users.  However, if you do that you can make both of them just return the `COUNT` results, so you don't have to do any additional stuff with `count` here.

services/QuillLMS/spec/queries/quill_big_query/active_teachers_all_time_query_spec.rb
```diff
@@ -0,0 +1,15 @@
+# frozen_string_literal: true
+
+require 'rails_helper'
+
+module QuillBigQuery
+  describe ActiveTeachersAllTimeQuery do
+    context 'for active teachers all time', :big_query_snapshot do
+      include_context 'Snapshots Activity Session Count CTE'
```
anathomical: Since you're not actually using the `Snapshots` module stack, it's probably most appropriate to include either a new shared context for these queries, or the base `QuillBigQuery TestRunner Setup` shared context.  If you do use the base context, you'll need to `let(:cte_records) { [] }` (or some array of factory-created models) since that context expects that value to be defined in addition to `query_args`.

services/QuillLMS/spec/queries/quill_big_query/active_teachers_all_time_query_spec.rb
```diff
@@ -0,0 +1,15 @@
...
+
+      let(:query_args) { [] }
+
+      it { expect(results).to match_array([]) }
```
anathomical: It would be nice to have some conditions which actually expect results to be returned so that we can confirm that the queries really are doing what we want.  So in this case we probably want something like
```ruby
  let(:num_teachers) { 10 } # I'm just picking an arbitrary number
  let(:users) { create_list(:teacher, num_teachers) }
  let(:units) { users.map { |user| create(:unit, user: user) } } # create one unit for each user
  let(:classroom_units) { units.map { |unit| create(:classroom_unit, unit: unit) } } # one classroom_unit for each unit
  let(:activity_sessions) { classroom_units.map { |classroom_unit| create(:activity_session, classroom_unit: classroom_unit) } } # one activity_session per classroom_unit

  let(:cte_records) { [users, units, classroom_units, activity_sessions] } # This fakes the models you've built into BigQuery for the length of the test

  it { expect(results).to match_array(activity_sessions.pluck(:id)) }

  context 'only one teacher with activity_sessions' do
    let(:activity_sessions) { [create(:activity_session, classroom_unit: classroom_units.first)] }

    it { expect(results.length).to eq(activity_sessions.length) } # or if you take my suggestion to have the queries return the counts instead of actual arrays of values, you can check that `expect(results).to eq(activity_sessions.length)`
  end

  context 'users with multiple activity_sessions' do
    let(:activity_sessions) { create_list(:activity_session, 2, classroom_unit: classroom_units.first) }

    it { expect(results.length).to eq(1) }
  end
```

services/QuillLMS/spec/queries/quill_big_query/activities_all_time_query_spec.rb
```diff
@@ -0,0 +1,24 @@
...
+
+require 'rails_helper'
+
+module QuillBigQuery
+  describe ActivitiesAllTimeQuery do
+    context 'for activities finished all time', :big_query_snapshot do
+      include_context 'QuillBigQuery TestRunner Setup'
+
+      let(:activity_sessions) { create_list(:activity_session, 20, state: 'finished') }
+      let(:unfinished_activity_sessions) { create_list(:activity_session, 10, state: 'started') }
```
anathomical: We have a "trait" in the `activity_session` factory for finished and started sessions, so this can be slightly cleaner as: `create_list(:activity_session, 20, :finished)` and `create_list(:activity_session, 10, :started)`.  (These traits also take care of setting `percentage`, `completed_at`, and `is_final_score` on the objects to make them more "valid".)

services/QuillLMS/app/workers/set_impact_metrics_worker.rb
anathomical: (Also, I realize that this isn't actually your code, but I believe you can remove all references to `floor` in these calculations since we expect all of our counts to be integers, and `floor` is used to is used to figure out how many decimal places to show for a floating point number: https://apidock.com/ruby/Numeric/floor.)

services/QuillLMS/spec/queries/quill_big_query/schools_containing_certain_teachers_query_spec.rb
```diff
@@ -0,0 +1,34 @@
...
+require 'rails_helper'
+
+module QuillBigQuery
+  describe SchoolsContainingCertainTeachersQuery do
+    context 'for schools all time', :big_query_snapshot do
+      include_context 'QuillBigQuery TestRunner Setup'
+
+      let(:schools) { create_list(:school, 10) }
+      let(:chosen_school_one) { create(:school) }
+      let(:chosen_school_two) { create(:school) }
```
anathomical: Would love to see the addition of a `let(:excluded_school) { create(:school) }` that gets included in `cte_records` just to make sure that our filtering is actually doing what we want and not just returning all schools.

services/QuillLMS/spec/services/quill_big_query/pre_transformer_spec.rb
```diff
@@ -0,0 +1,116 @@
+# frozen_string_literal: true
+
+require 'rails_helper'
+
+describe QuillBigQuery::PreTransformer do
```
anathomical, line 5: Love how extensive the tests are for this class.  Very nicely done.

services/QuillLMS/spec/workers/set_impact_metrics_worker_spec.rb
```diff
@@ -5,54 +5,56 @@
...
+        {"id" => 333, "free_lunches" => 30},
+        {"id" => 334, "free_lunches" => 0},
+        {"id" => 545, "free_lunches" => 54}
+      ]
+    }
+
+    before do
+      allow(QuillBigQuery::ActivitiesAllTimeQuery).to receive(:run).and_return(example_json1)
+      allow(QuillBigQuery::ActiveTeachersAllTimeQuery).to receive(:run).and_return(example_json2)
+      allow(QuillBigQuery::SchoolsContainingCertainTeachersQuery).to receive(:run).with(teachers.map(&:id)).and_return(example_json3)
```
anathomical: Super minor, but you can replace `map(&:id)` with `pluck(:id)` which I THINK is a pattern that we're preferring when we can get away with it?

services/QuillLMS/spec/workers/set_impact_metrics_worker_spec.rb
```diff
@@ -5,54 +5,56 @@
...
     it 'should set the NUMBER_OF_SENTENCES redis value' do
       subject.perform
-      expect($redis.get(PagesController::NUMBER_OF_SENTENCES)).to eq(@number_of_sentences.to_s)
+      expect($redis.get(PagesController::NUMBER_OF_SENTENCES)).to eq((example_json1.size.floor(-5) * 10).to_s)
     end
 
     it 'should set the NUMBER_OF_STUDENTS redis value' do
       subject.perform
-      expect($redis.get(PagesController::NUMBER_OF_STUDENTS)).to eq(@number_of_students.to_s)
+      expect($redis.get(PagesController::NUMBER_OF_STUDENTS)).to eq((example_json1.count('DISTINCT(user_id)').floor(-5)).to_s)
```
anathomical: I flagged this in the worker itself: `count('DISTINCT(user_id)')` is syntactically valid, but doesn't actually do what you want it to do.  The risk of duplicating this code here is that you've masked the failure.  You want to confirm, at least in this case, that `expect($redis.get(PagesController::NUMBER_OF_STUDENTS)).to eq(example_json.length.to_s)` or something along those lines.

Similar issue on line 37 above.  The tests below where you are expecting explicit numbers are better, but I've got a note on those, too.

services/QuillLMS/spec/workers/set_impact_metrics_worker_spec.rb
```diff
@@ -5,54 +5,56 @@
...
     end
 
     it 'should set the NUMBER_OF_TEACHERS redis value' do
       subject.perform
-      expect($redis.get(PagesController::NUMBER_OF_TEACHERS)).to eq(@number_of_teachers.to_s)
+      expect($redis.get(PagesController::NUMBER_OF_TEACHERS)).to eq("100")
```
anathomical: Much better to expect values to equal constants, but these current numbers are "magic".  I can't read this expectation and understand WHY this should be `"100"`.  Especially if I look above and see that `teachers` is a list of 150 teachers.  Why isn't this value 150?  I think you want this to be explicit in some way.  Either expecting `to eq(teachers.length.to_s)` or doing something like
```ruby
let(:num_teachers) { 150 }
let(:teachers) { create_list(:teacher, num_teachers) }

expect(...).to eq(num_teachers.to_s)
```
Same note on the two constant strings below on 52 and 57.

anathomical (changes requested):
The core of this code, the use of a pre-transformer to handle query params, is an excellent one, and I like the way you've laid things out.  I've got a lot of syntactic and nit-pick-y style things that I'd like you to take a look at.  A bunch of them I think you can use your judgment on.

There's also some core functionality that I think straight-up doesn't work as written which is caused by the switch from ActiveRecord queries (which treat `count` differently) to BigQuery queries.  That, I think, needs some cleaner automated tests to ensure work as expected.

services/QuillLMS/app/workers/set_impact_metrics_worker.rb
happythenewsad: Nit, since this was here before: abstract the magical arithmetic `floor(-5) * 10` into a helper method

services/QuillLMS/app/services/quill_big_query/pre_transformer.rb
happythenewsad: A quick and dirty way to effect this is just to do `PostTransformer::TYPE_TO_LAMBDA_LOOKUP.fetch(your thing)`

This will validate that everything passed into `type_lookup` is a BigQuery API-approved type.

happythenewsad (commented):
I think this will be pretty tight after implementing Thomas' comments. I'm 'commenting' rather than approving, since I'd like to do another pass afterward.

services/QuillLMS/app/services/quill_big_query/pre_transformer.rb
cissyxyu: Skipping this for now since I couldn't find a very straightforward way to backwards-link the types to constants in the other file.

services/QuillLMS/app/workers/set_impact_metrics_worker.rb
cissyxyu: The "floor" is needed not to round to an integer, but because we are rounding these numbers down to the nearest 10,000 (because these are very big numbers so it makes for better display to show them nicely rounded down to five zeros).

services/QuillLMS/app/workers/set_impact_metrics_worker.rb
```diff
@@ -4,30 +4,39 @@ class SetImpactMetricsWorker
...
 
-    teachers_query = QuillBigQuery::ActiveTeachersAllTimeQuery.run
-    teacher_ids = teachers_query.to_a.map {|teacher| teacher["id"]}
+    teachers = ImpactMetrics::ActiveTeachersAllTimeQuery.run
+    teacher_ids = teachers.to_a.map {|teacher| teacher["id"]}
 
-    schools_query = QuillBigQuery::SchoolsContainingCertainTeachersQuery.run(teacher_ids)
-    low_income_schools_query = schools_query.select { |school| school["free_lunches"].present? && school["free_lunches"] > FREE_LUNCH_MINIMUM}
+    schools = ImpactMetrics::SchoolsContainingCertainTeachersQuery.run(teacher_ids)
+    low_income_schools = schools.select { |school| school["free_lunches"].present? && school["free_lunches"] > FREE_LUNCH_MINIMUM}
```
anathomical: Super minor style note.  You could use `school.fetch('free_lunches', 0) > FREE_LUNCH_MINIMUM` instead of `.present? && > FREE_LUNCH_MINIMUM`.  You can take or leave that, I just like to avoid in-line checks for `present?` if possible.

anathomical (approved):
Thanks for the clean-up.  This looks good to me now.  (Love all the tests, as usual.)

happythenewsad (approved):

PR #10746, opened 2023-07-10
stop making unnecessary calls for archived classrooms

emilia-friedberg:
## WHAT
Stop pulling the number of completed activity sessions for archived classrooms, because we aren't showing it on the frontend.

## WHY
I noticed while testing a user with a very large number of classrooms and students (I think it was an admin that imported all of their school's classes) that the classrooms page was very slow to load (at least locally, change was less significant in prod though still a little slow) because we're now pulling a count of each student's completed activity sessions for each classroom. This is totally unnecessary for archived classes since we don't show this info on the frontend.

## HOW
Only make this call when the classroom is visible.

### Screenshots

### Notion Card Links

PR Checklist | Your Answer
------------ | -------------

### DIFF
--- a/services/QuillLMS/app/controllers/teachers/classrooms_controller.rb
+++ b/services/QuillLMS/app/controllers/teachers/classrooms_controller.rb
@@ -205,11 +205,14 @@ class Teachers::ClassroomsController < ApplicationController
   end
 
   private def format_students_for_classroom(classroom)
-    sorted_students = classroom.students.sort_by(&:last_name)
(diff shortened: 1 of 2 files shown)


anathomical (approved):
Looks good.  Worth surfacing here that we had a discussion via Slack where it was pointed out that we can't just removed the `unscoped` clause on line 186 because we do want classrooms where `visible = false`, we just don't need this particular expanded data for those classrooms because we show limited information about archived classrooms.

PR #10754, opened 2023-07-12
Canvas Rostering Backend Part 5: Remove canvas_instance from ProviderClassroomUser and add external_id helper methods

brendanshean:
## WHAT
1. Remove the `canvas_instance` reference from the `ProviderClassroomUser`.
2. Add code to manage `classroom_external_id` for `CanvasClassroom` 
3. Add code to manage `user_external_id` of `CanvasAccount`

## WHY
1.  Unlike Google and Clever, each `CanvasClassroom` and `CanvasAccount` effectively has a composite key for identification since each object will need an external_id and canvas_instance_id.  Initially, I chose add a `canvas_instance_id` to `ProviderClassroomUser` keep referential integrity with the `CanvasInstance`.  However, In building out the rostering, treating the composite key unifies the data importing pipeline
2.  Standardize the handling of `classroom_external_id` with `google_classroom_id` and `clever_id` in `ProviderClassroomUser`
3.  Standardize the handling of `user_external_id` with `google_id` and `clever_id` in `ProviderClassroomUser`

## HOW
1.  Add a migration with   `remove_reference :provider_classroom_users, :canvas_instance, foreign_key: true`
2. Add build, find_by, unpack and validate methods for `classroom_external_id`
3. Add create_by, find_by, unpack and validate methods for `user_external_id`

### Screenshots

### Notion Card Links

PR Checklist | Your Answer
------------ | -------------


services/QuillLMS/app/models/canvas_account.rb
```diff
@@ -22,10 +22,42 @@
...
+  def self.custom_create_by_user_external_id!(user_external_id, user_id)
+    canvas_instance_id, external_id = unpack_user_external_id!(user_external_id)
+
+    create!(canvas_instance_id: canvas_instance_id, external_id: external_id, user_id: user_id)
+  end
+
+  def self.custom_find_by_user_external_id(user_external_id)
+    return nil unless valid_user_external_id_format?(user_external_id)
+
+    canvas_instance_id, external_id = user_external_id.split(':')
```
anathomical, line 45: Should this be a call to `unpack_user_external_id` instead of `split`?

services/QuillLMS/app/models/classroom.rb
```diff
@@ -206,12 +217,15 @@ def google_classroom?
...
+    return 'Canvas' if canvas_classroom?
   end
 
   def classroom_external_id
     return google_classroom_id if google_classroom?
 
     return clever_id if clever_classroom?
+
+    return canvas_classroom&.classroom_external_id
   end
```
anathomical, line 229: With only three options, this probably isn't a big deal, but I wonder if in the long term we should look into some sort of lookup-Hash for types rather than a bunch of `if` statements.  Maybe not worth worrying about if this is the only function that does this, but I often prefer hashes like that over a whole bunch of `if` clauses.  I could imagine something like:
```ruby
def classroom_external_id
  {
    'Google' => -> (classroom) { classroom.google_classroom_id },
    'Clever' => -> (classroom) { classroom.clever_id },
    'Canvas' => -> (classroom) { classroom.canvas_classroom&.classroom_external_id }
  }.fetch(classroom_provider).call(self)
end
```

anathomical (approved):
Looks good to me.  I totally dig this pattern that you've established for providers here.

services/QuillLMS/app/models/canvas_account.rb
happythenewsad: If I'm reading this right, external_id contains a colon, so this line will yield

[canvas_instance_id, first_part_of_external_id, last_part_of_external_id] 

Might be worth making a supporting class does the packing and unpacking, and validations.

services/QuillLMS/app/models/classroom.rb
```diff
@@ -178,6 +183,8 @@ def classroom_type_for_segment
       'Google Classroom'
     elsif clever_id
       'Clever'
+    elsif canvas_classroom
```
happythenewsad, line 186: Using a switch statement here, but not enforcing the existence of one-and-only-one property, could lead to hard-to-diagnose bugs or unexpected behavior downstream. 

Let's throw an exception here if, say, a Classroom has a google ID AND a canvas classroom. 

Or you could enforce this integrity in a different way, I don't have a strong preference about the precise implementation.

services/QuillLMS/app/models/provider_classroom_user.rb
```diff
@@ -9,24 +9,18 @@
...
-#
-#  fk_rails_...  (canvas_instance_id => canvas_instances.id)
+#  index_provider_type_and_classroom_id_and_user_id  (type,classroom_external_id,user_external_id) UNIQUE
 #
 class ProviderClassroomUser < ApplicationRecord
   ACTIVE = :active
   DELETED = :deleted
 
-  TYPES = %w[CleverClassroomUser GoogleClassroomUser].freeze
+  TYPES = %w[CanvasClassroomUser CleverClassroomUser GoogleClassroomUser].freeze
```
happythenewsad, line 23: Interesting, I didn't know hard brackets with `%w` syntax was a thing.

services/QuillLMS/app/models/user.rb
```diff
@@ -674,6 +675,10 @@ def generate_username(classroom_id=nil)
     self.username = GenerateUsername.run(first_name, last_name, get_class_code(classroom_id))
   end
 
+  def canvas_authorized?
```
happythenewsad, line 682: The path where this function returns nil should be coerced to a boolean, to match the function-name-with-question-mark convention.

happythenewsad (changes requested):
Some minor suggestions, but the approach looks very solid, same as part 4

services/QuillLMS/app/models/provider_classroom_user.rb
brendanshean: The delimiters are stylistic, you can use different ones:
  
Square brackets: %w[...]
Curly braces: %w{...}
Parentheses: %w(...)
Less than/greater than: %w<...>

I tend to prefer square brackets since the return type is an array.

services/QuillLMS/app/models/canvas_account.rb
brendanshean: I'm not using `unpack_user_external_id!` since I wanted a split method that did not raise an error on exception.

services/QuillLMS/app/models/classroom.rb
brendanshean: I think this is an interesting pattern.  I  will make a note to consider it when I refactor out clever_id and google_classroom_id from the classroom table.

services/QuillLMS/app/models/classroom.rb
brendanshean: The problem here is that we already have some records with both.  Following earlier conventions, the precedence for now is to give google classroom priority.

When the classroom schema has google_classroom_id and clever_id removed, this will be enforced at the database level.

happythenewsad (approved):

PR #10755, opened 2023-07-12
get mvp version of key target skill concepts working

emilia-friedberg:
## WHAT
Update the apps (except Evidence, which doesn't tie concepts to prompts in the same way) to all pass back `question_uid` and `question_concept_uid`, then use the `question_concept_uid` to generate a `key_target_skill_concept` to display in the Activity Analysis report.

## WHY
This is the first step of a project to emphasize one key concept per question over all the concept results we store per activity session.

## HOW
Update the various `conceptResults` functions in the apps to pass back the data we need, then update the `public_progress_reports` controller and frontend to handle the new attribute.

### Screenshots
Screen Shot 2023-07-12 at 11 09 47 AM
Screen Shot 2023-07-12 at 11 09 35 AM

### Notion Card Links
https://www.notion.so/quill/Update-Activity-Analysis-student-report-to-use-key-target-skill-concept-82699ddcbad843249dcf2eec662d1153?pvs=4

PR Checklist | Your Answer
------------ | -------------


eadams17 (approved):

services/QuillLMS/client/app/bundles/Lessons/libs/conceptResults/classroomLessons.js
```diff
@@ -18,6 +18,7 @@ export function generateConceptResult(questionData, studentSubmission) {
       prompt: questionData.prompt,
       answer: studentSubmission.data,
       attemptNumber: 1,
+      question_concept_uid: 'X37oyfiNxSphA34npOb-Ig'
```
anathomical: Might be worth a quick comment explaining why this particular value is hard-coded here.  (I assume that all Lessons have the same concept, so they don't need to pass custom values around, but I could be wrong and other developers might not find it obvious.)

anathomical (commented):
A sort of high-level question here: should we use a migration to add `question_uid` and `question_concept_uid` as columns to the `concept_results` table rather than inserting them into `metadata`?  I think that, in general, we should try to avoid putting "regular" data into `metadata`.  This would require either a refactor of the payload shapes in all of the apps to submit these values outside of `metadata`, or a tweak to the `KNOWN_METADATA_KEYS` and `init_from_json` code in the `ConceptResult` model so I'd be open to discussion on this rather than committing to extra work.

Other than that, this looks really good to me.

anathomical (approved):
After a discussion via Slack, I've come to the conclusion that doing things this way is the simplest immediate implementation, and that there are enough questions about a more robust implementation that we really should have a longer discussion on what our long-term solution would be.  However, given the deadlines for getting this out the door, keeping things simple is a good call.

So approving as-is, and hoping that we get a chance to revisit the way we think about this stuff relatively soon.

PR #10763, opened 2023-07-13
lms/handle-rare-teacher-notification-errors

anathomical:
## WHAT
Handle notifications for activity sessions with no classroom_unit
## WHY
ClassroomUnit is the only way an ActivitySession knows which classroom it should be associated with, and we require that information for the ways that we roll up notifications.  So in the rare case where an ActivitySession is completed without a ClassroomUnit, we return early.
## HOW
Just add an early return in the parent worker that all of the various notification creating workers inherit from

### Notion Card Links
https://www.notion.so/quill/Investigate-occasional-errors-with-TeacherNotification-workers-43a29e6a49514350b1b6a20e90dda81d

PR Checklist | Your Answer
------------ | -------------

### DIFF
--- a/services/QuillLMS/app/workers/teacher_notifications/send_notification_worker.rb
+++ b/services/QuillLMS/app/workers/teacher_notifications/send_notification_worker.rb
@@ -9,6 +9,7 @@ module TeacherNotifications
     def perform(activity_session_id)
       @activity_session = find_activity_session(activity_session_id)
       return unless @activity_session
+      return unless @activity_session.classroom
       return unless should_send_notification?
 
       send_notification(notification_type, message_attrs)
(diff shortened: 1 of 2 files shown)


services/QuillLMS/app/workers/teacher_notifications/send_notification_worker.rb
```diff
@@ -9,6 +9,7 @@ class SendNotificationWorker
     def perform(activity_session_id)
       @activity_session = find_activity_session(activity_session_id)
       return unless @activity_session
+      return unless @activity_session.classroom
```
brendanshean, line 12: can 11-12 be rewritten as `return unless @activity_session&.classroom`?

brendanshean (approved):

PR #10775, opened 2023-07-20
activity analysis redesign

emilia-friedberg:
## WHAT
Update the Activity Analysis student report to emphasize key target skills over concept results, and update the tooltip in the Activity Summary to do the same.

## WHY
This is part of a larger project to change how we score activity sessions and display data about them to teachers.

## HOW
Update the backend (minorly for the activity summary page, just to get the name of the classification, and more significantly for the tooltip query to pass the array of session data that we need). Then update the frontend with a bunch of styling changes as spec-ed.

### Screenshots
Screen Shot 2023-07-21 at 10 40 22 AM
Screen Shot 2023-07-21 at 10 41 26 AM
Screen Shot 2023-07-21 at 10 41 34 AM

### Notion Card Links
https://www.notion.so/quill/Update-Activity-Analysis-to-Support-Target-Skill-3fbb99bda9ec4538ac753e65da804bac?pvs=4

PR Checklist | Your Answer
------------ | -------------


services/QuillLMS/client/app/bundles/Teacher/components/progress_reports/diagnostic_reports/student_report.tsx
```diff
@@ -26,119 +28,183 @@ interface StudentReportProps extends RouteComponentProps {
...
+  React.useEffect(() => {
+    if (isInitialMount.current) {
+      isInitialMount.current = false
+      getStudentData()
+    } else {
+      setLoading(true)
+      getStudentData()
     }
-  }
+  }, [params])
```
eadams17, line 51: just curious why you opted to go with this `isInitialMount` variable reference rather than using `[]` as an arg in `useEffect`?

eadams17 (approved):
LGTM! Thanks for getting all of this into place 🎉

services/QuillLMS/app/controllers/grades_controller.rb
```diff
@@ -9,64 +10,59 @@ def index
...
+    activity_sessions = ActivitySession
+      .includes(:concept_results, :activity, :unit)
+      .where(
+        classroom_unit_id: tooltip_params[:classroom_unit_id].to_i,
+        user_id: tooltip_params[:user_id].to_i,
+        activity_id: tooltip_params[:activity_id].to_i,
+        visible: true,
+        state: ActivitySession::STATE_FINISHED
+      )
+      .order(:completed_at)
```
anathomical, line 30: Love seeing the old raw SQL get replaced by ActiveRecord calls like this.

anathomical (approved):

services/QuillLMS/client/app/bundles/Teacher/components/progress_reports/diagnostic_reports/student_report.tsx
emilia-friedberg: Using `[]` would have it fire only on initial load, but we want to make sure this gets called both on initial load and also when the params change. However, we only want it to `setLoading(true)` on later loads, so we need that bit of logic in there to see if it's the first time it's loaded or not.

PR #10780, opened 2023-07-21
update demo account data to use correct scoring

emilia-friedberg:
## WHAT
Add a rake task to update the data we use to generate the demo account so that it reflects our new scoring system, where we give full credit for questions where a student reached an optimal response regardless of attempt. Also add a new session based on one I completed on prod so PG can see how the new scoring looks on demo. Then regenerate our `.yml` files so those database changes are persisted whenever we create new demo accounts. 

## WHY
We want our demo account data to be up to date. Note: this update does not add the data we need to generate target skill values, because there isn't really a way to backdate that, so reports will all use the default "Conventions of Language" as their target skill. I'm checking in with PG about this, but we may need to have staff members actually play a bunch of sessions at some point so we can populate these.

## HOW
See "What".

Note: I did have to update two tests that concretized the number of concept results -- the new generated files seem to have  six fewer total than the previous ones did. I think this is due to some update to the production data between now and when they were originally generated, and not the rake task, which shouldn't be destructive. Since the number is small and we may be regenerating these sessions entirely soon, I didn't think it was worth delving into deeply at this stage, but I can do so if anyone feels strongly about it.

### Screenshots

### Notion Card Links
https://www.notion.so/quill/Update-demo-data-to-reflect-new-scoring-41e8090ec98d490d9cadfd4f129e89d5?pvs=4

PR Checklist | Your Answer
------------ | -------------
Have you added and/or updated tests? |  N/A


services/QuillLMS/lib/tasks/update_scoring_for_demo_account_activity_sessions.rake
brendanshean, line 1: I think breaking the task into smaller methods would help with readability:

```
  task update: :environment do
    include PublicProgressReports

    REPLAYED_PAIR = [Demo::ReportDemoCreator::REPLAYED_ACTIVITY_ID, Demo::ReportDemoCreator::REPLAYED_SAMPLE_USER_ID].freeze
    
    ACTIVITY_USER_PAIRS = 
      Demo::ReportDemoCreator::ACTIVITY_PACKS_TEMPLATES
        .map {|hash| hash[:activity_sessions].map(&:to_a)}
        .flatten(2)
        .push(REPLAYED_PAIR)

    fetch_activity_sessions(ACTIVITY_USER_PAIRS)
      .reject { |activity_session| ActivityClassification::UNSCORED_KEYS.include?(activity_session.classification.key) }
      .each { |activity_session|  update_activity_session_percentage(activity_session) }

  end

  def fetch_activity_sessions(id_pairs)
    id_pairs.map do |activity_id, user_id|
      ActivitySession.unscoped.find_by(activity_id: activity_id, user_id: user_id, is_final_score: true)
    end.compact
  end

  def update_activity_session_percentage(activity_session)
    correct_questions_count = 0
    grouped_concept_results = activity_session.concept_results.group_by(&:question_number)

    grouped_concept_results.each_value do |concept_results|
      if get_score_for_question(concept_results) > 0
        concept_results.each { |cr| cr.update(question_score: 100) }
        correct_questions_count += 1
      end
    end

    activity_session.update(percentage: (correct_questions_count.to_f / grouped_concept_results.size).round(2))
  end
end

brendanshean (approved):
I think it's fine to ignore the decrease of 6 records for now and just monitor.

happythenewsad (approved):

anathomical (approved):

PR #10798, opened 2023-07-25
fix completed_at date for tooltip

emilia-friedberg:
## WHAT
Send back `completed_at` as datetime rather than integer so frontend can handle it correctly.

## WHY
We want users to see accurate information.

## HOW
Just wrap the time definition in a `Time.at` call.

### Screenshots
Screen Shot 2023-07-25 at 1 05 51 PM

### Notion Card Links
https://www.notion.so/quill/Completed-date-and-target-skills-displaying-incorrect-info-cf955d7f14734e3a9095558a3320f626?d=7d620c9da0d24ccca659c1069c0e85c8

PR Checklist | Your Answer
------------ | -------------

### DIFF
--- a/services/QuillLMS/app/controllers/grades_controller.rb
+++ b/services/QuillLMS/app/controllers/grades_controller.rb
@@ -46,7 +46,7 @@ class GradesController < ApplicationController
       percentage: activity_session.percentage,
       description: activity_session.activity.description,
       due_date: unit_activity.due_date,
-      completed_at: Time.at(activity_session.completed_at.to_i + current_user.utc_offset.seconds),
+      completed_at: activity_session.completed_at + current_user.utc_offset.seconds,
       grouped_key_target_skill_concepts: format_grouped_key_target_skill_concepts(key_target_skill_concepts),
       number_of_questions: questions.length,
       number_of_correct_questions: correct_key_target_skill_concepts.length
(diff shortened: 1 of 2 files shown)


anathomical (approved):

eadams17 (approved):

services/QuillLMS/app/controllers/grades_controller.rb
```diff
@@ -46,7 +46,7 @@ def tooltip
       percentage: activity_session.percentage,
       description: activity_session.activity.description,
       due_date: unit_activity.due_date,
-      completed_at: activity_session.completed_at.to_i + current_user.utc_offset.seconds,
+      completed_at: Time.at(activity_session.completed_at.to_i + current_user.utc_offset.seconds),
```
anathomical: Quick note, I think that because of the nature of Datetime math in Ruby, you can also just remove `to_i` so that you add seconds directly to the `completed_at` value.  But this works, too.

happythenewsad: nils are probably not going to happen, but you could add the safe navigation operator

`activity_session&.completed_at ...`

happythenewsad (approved):

services/QuillLMS/app/controllers/grades_controller.rb
emilia-friedberg: this is cleaner, thank you.

emilia-friedberg: I don't think that's necessary here since this method is only called on a `.map` of `ActivitySession` records from the database, and I don't want to add too much unnecessary branching complexity, but thank you for flagging.

PR #10825, opened 2023-07-31
add migration and model files for diagnostic question skills

emilia-friedberg:
## WHAT
Add migration and model file updates for the new `diagnostic_question_skills` table.

## WHY
This is the table we're using to replace `skills` for the new diagnostic scoring logic.

## HOW
Just generate the migration, build a very simple model file, and add tests.

### Screenshots

### Notion Card Links
https://www.notion.so/quill/Spec-Base-the-Diagnostic-Scoring-on-the-Question-ID-not-the-Concept-Result-82251d621dcb4411996830ec37d154d6?pvs=4

PR Checklist | Your Answer
------------ | -------------

### DIFF
--- a/services/QuillLMS/db/migrate/20230731184420_create_diagnostic_question_skills.rb
+++ b/services/QuillLMS/db/migrate/20230731184420_create_diagnostic_question_skills.rb
@@ -1,0 +1,13 @@
+# frozen_string_literal: true
+
+class CreateDiagnosticQuestionSkills < ActiveRecord::Migration[6.1]
+  def change
+    create_table :diagnostic_question_skills do |t|
+      t.string :name, null: false
+      t.references :question, index: true, foreign_key: true, null: false
+      t.references :skill_group, index: true, foreign_key: true, null: false
+
+      t.timestamps
+    end
+  end
+end
--- a/services/QuillLMS/app/models/diagnostic_question_skill.rb
+++ b/services/QuillLMS/app/models/diagnostic_question_skill.rb
@@ -1,0 +1,31 @@
+# frozen_string_literal: true
+
+# == Schema Information
+#
+# Table name: diagnostic_question_skills
+#
+#  id             :bigint           not null, primary key
+#  name           :string           not null
+#  created_at     :datetime         not null
+#  updated_at     :datetime         not null
+#  question_id    :bigint           not null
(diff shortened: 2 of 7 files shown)


eadams17 (approved):

services/QuillLMS/db/migrate/20230731184420_create_diagnostic_question_skills.rb
```diff
@@ -0,0 +1,13 @@
...
+
+class CreateDiagnosticQuestionSkills < ActiveRecord::Migration[6.1]
+  def change
+    create_table :diagnostic_question_skills do |t|
+      t.string :name, null: false
+      t.references :question, index: true, foreign_key: true, null: false
+      t.references :skill_group, index: true, foreign_key: true, null: false
+
+      t.timestamps
+    end
```
anathomical, line 11: Given that, from what I can tell of the use-case, there will only ever be a single skill associated with any given question, what do you think about the idea of just storing a `skill_group_id` on the `questions` table itself?  It reduces flexibility in the case that we ever want to do something different, but it does let us avoid adding an additional model that we may not really need?  (That also lets us avoid an additional index and foreign key required to tie this record to `questions`, but given how small the `questions` table is, I doubt that's a huge deal.

Not a direct objection to this approach, but it felt like I should check.

anathomical (commented):
Quick question about whether it might not make sense to just add a column to `questions` rather than creating a new table.

services/QuillLMS/db/migrate/20230731184420_create_diagnostic_question_skills.rb
emilia-friedberg: I think that's a really good idea, and will more directly reflect the frontend (where what used to be referred to as "skills" is now going to be referred to as "questions"). I'll go ahead and make that change now.

anathomical (approved):

services/QuillLMS/db/migrate/20230731184420_create_diagnostic_question_skills.rb
emilia-friedberg: Update after discussing with Hannah: since questions can be used in multiple diagnostics, we do need this join table so that the same question can be linked to multiple skill groups.

PR #10835, opened 2023-08-01
add skill to question mapping csv and rake task

emilia-friedberg:
## WHAT
Add a CSV with the data for the `diagnostic_question_skills` table, and add a rake task to store that data.

## WHY
We need these database links for the new scoring system.

## HOW
See "What".

### Screenshots

### Notion Card Links
https://www.notion.so/quill/Add-script-to-upload-new-diagnostic-question-skills-ccd7ff27351443e893633d828670cb97?pvs=4

PR Checklist | Your Answer
------------ | -------------
Have you added and/or updated tests? |  NO - no tests for migrations
Have you deployed to Staging? | NO - ran locally, will run on staging once I have the full data

### DIFF
--- a/services/QuillLMS/lib/tasks/populate_diagnostic_question_skills.rake
+++ b/services/QuillLMS/lib/tasks/populate_diagnostic_question_skills.rake
@@ -1,0 +1,20 @@
+# frozen_string_literal: true
+
+namespace :diagnostic_question_skills do
+  desc 'Populate diagnostic question skills table'
+  task :populate => :environment do
+    skill_to_question_mapping = CSV.parse(File.read("lib/data/skill_to_question_mapping.csv"), headers: true)
+
+    ActiveRecord::Base.transaction do
+      skill_to_question_mapping.each do |row|
+        begin
+          question = Question.find_by_uid(row['Question ID'])
+          DiagnosticQuestionSkill.create(question: question, skill_group_id: row['Skill Group ID'], name: row['Skill Name'])
+        rescue => e
+          puts "ID:#{row['Question ID']}: #{e}"
+        end
+      end
+    end
+
+  end
+end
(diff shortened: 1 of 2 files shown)


brendanshean (approved):
Looks good!

services/QuillLMS/lib/tasks/populate_diagnostic_question_skills.rake
```diff
@@ -0,0 +1,20 @@
+# frozen_string_literal: true
+
+namespace :diagnostic_question_skills do
+  desc 'Populate diagnostic question skills table'
+  task :populate => :environment do
+    skill_to_question_mapping = CSV.parse(File.read("lib/data/skill_to_question_mapping.csv"), headers: true)
```
anathomical, line 6: No reason that you need to update this since the CSV is small and doesn't contain any data that's not suitable for publication (via being in our public repo), and given that we want all developers to be able to put this in their local databases, it may make sense for this to be committed.  That said, we have a pattern for using CSVs that lets us avoid actually committing the data CSV to the repo.  (You can see an example in `lib/tasks/concept_results_migration.rake`.)

services/QuillLMS/lib/tasks/populate_diagnostic_question_skills.rake
```diff
@@ -0,0 +1,20 @@
...
+
+    ActiveRecord::Base.transaction do
+      skill_to_question_mapping.each do |row|
+        begin
+          question = Question.find_by_uid(row['Question ID'])
+          DiagnosticQuestionSkill.create(question: question, skill_group_id: row['Skill Group ID'], name: row['Skill Name'])
```
anathomical, line 12: Again, not something to bother with here, but there's a nice alternative magical syntax in Rails that you could use in situations like this:
```ruby
# You can create_<name of association from has_one or whatever>
question.create_diagnostic_skill_question(skill_group_id: row['Skill Group ID'], name: row['Skill Name'])
```

anathomical (approved):
A couple of comments on alternative approaches you could take, but I think this is fine as-is.

PR #10841, opened 2023-08-01
use diagnostic question skills in scoring and update frontend to handle correctly

emilia-friedberg:
## WHAT
Update the backend to use `diagnostic_question_skills`, rather than `skills`, in the diagnostic scoring, and update the frontend to handle empty states for diagnostics completed before the requisite data was available (namely, the `question_uid` in `concept_results.extra_metadata`) and remove components that are totally nonsensical in the new system.

Note: there is a lot of renaming and cleanup we'll want to do here to be consistent with the frontend (ie, all references to `skill` should be to `question`. I updated some I found really confusing, but I think we should do the rest as a follow-up task due to the time-sensitive nature of this project.

## WHY
This update is going to give us greater parity with the other reports, and makes the teacher reports consistent with the scoring we intend to use for the admin reports.

## HOW
See "What"

### Screenshots
Screen Shot 2023-08-02 at 11 47 35 AM
Screen Shot 2023-08-02 at 11 47 46 AM
Screen Shot 2023-08-02 at 11 47 58 AM
Screen Shot 2023-08-02 at 11 48 27 AM
Screen Shot 2023-08-02 at 11 51 08 AM

### Notion Card Links
https://www.notion.so/quill/Teacher-Reports-UI-changes-for-Switching-from-Skills-to-Questions-1-Add-Empty-States-2-Remov-537f32480b4a469199d95ff47c33d7eb?pvs=4
https://www.notion.so/quill/Base-the-Diagnostic-Scoring-on-the-Question-ID-not-the-Concept-Result-82251d621dcb4411996830ec37d154d6?pvs=4

PR Checklist | Your Answer
------------ | -------------


brendanshean (approved):
I think the two concerns:
app/controllers/concerns/growth_results_summary.rb
app/controllers/concerns/results_summary.rb

are in need of some major refactoring, but I think that will have to wait for a follow-up task like you mention.

Great work on sorting out all these details.

eadams17 (approved):
Thank you for knocking all of this out so quickly! Much appreciated given the time-sensitive nature

services/QuillLMS/app/services/diagnostics_organized_by_classroom_fetcher.rb
```diff
@@ -28,46 +30,27 @@ def run
       index_of_existing_classroom = classrooms.find_index { |c| c['id'] == record['classroom_id'] }
       name = grouped_name(record)
 
-      next if record['post_test_id'] &&
```
emilia-friedberg, line 30: I can't figure out why this was necessary to do previously, and in fact it seemed to be causing a bug where activities without post-tests (like the various College Board-related diagnostics) were showing up with duplicates, so I'm removing it. I'll monitor in case we get related bug reports.

services/QuillLMS/app/services/diagnostics_organized_by_classroom_fetcher.rb
```diff
@@ -28,46 +30,27 @@ def run
...
           activity_name: Activity.find_by_id(record['post_test_id'])&.name,
           unit_template_id: ActivitiesUnitTemplate.find_by_activity_id(record['post_test_id'])&.unit_template_id
         }
-
-        grouped_record[:pre] = record_with_aggregated_activity_sessions(
-          record['activity_id'],
-          record['classroom_id'],
-          nil
-        )
-      else
```
emilia-friedberg, line 56: Similarly, here, I think this special handling of pre-tests that don't have post-tests is no longer necessary, so I cleaned up the code.

emilia-friedberg:
@brendanshean @eadams17 I'm re-requesting a review here because I discovered an edge case in testing where teachers who had assigned a diagnostic to a given classroom prior to the question-scoring switchover, and then again afterwards, were always getting the empty state even though we do sometimes have good data to show them. This is unavoidable without significant UX changes in instances where not every student who completed the given diagnostic in the given classroom before the switchover has completed a new session after the switchover, but I think the changes I've made will minimize the number of users with the suboptimal experience here. I think you only need to review what's changed since last time.

services/QuillLMS/spec/services/diagnostics_organized_by_classroom_fetcher_spec.rb
brendanshean, line 1: More of a stylistic thing, but you could shorten all six of these `it` blocks to one liners since the on the test doesn't seem to add much information:
```
 it 'should return the record with eligible_for_question_scoring = false' do
    expect(subject.record_with_aggregated_activity_sessions(activity.id, classroom.id['eligible_for_question_scoring']).to eq(false)
end
```
becomes
```
it { expect(subject.record_with_aggregated_activity_sessions(activity.id, classroom.id['eligible_for_question_scoring']).to eq(false) }
```

brendanshean (approved):
A step in the right direction of reducing complexity here!

eadams17 (approved):

PR #10843, opened 2023-08-01
Premium Hub Data Export

eadams17:
## WHAT
add a Data Export tab to the Premium Hub (also fix regression where diamond icons weren't displaying in Premium Hub navbar)

## WHY
we want to add this view (and in the near future, downloadable reports) for admins

## HOW
- add a new view, using existing `Filters` and `DataTable` components to display and filter results
- add a new action in the `SnapshotsController` and a new `DataExportQuery`
- add `CachePremiumReportsWorker` to cache results nightly

### Screenshots
Screen Shot 2023-08-01 at 6 29 03 PM
Screen Shot 2023-08-01 at 6 12 09 PM

### Notion Card Links
https://www.notion.so/quill/Technical-Review-Data-Export-Admin-Report-3bbefc045bdd4745b484aea86cd8daab

PR Checklist | Your Answer
------------ | -------------


services/QuillLMS/client/app/bundles/PremiumHub/components/dataExportTableAndFields.tsx
```diff
@@ -0,0 +1,234 @@
...
+      const classroomIdsAreEqual = unorderedArraysAreEqual(message.classroom_ids, selectedClassroomIds.map(id => String(id)))
+      const gradesAreEqual = unorderedArraysAreEqual(message.grades, selectedGrades.map(grade => String(grade))) || (!message.grades && !selectedGrades.length)
+
+      if (queryKeysAreEqual && timeframesAreEqual && schoolIdsAreEqual && gradesAreEqual && teacherIdsAreEqual && classroomIdsAreEqual) {
+        getData()
+      }
+    });
+  };
+
+  function toggleCheckbox(e) {
```
emilia-friedberg, line 173: I like this generic function.

services/QuillLMS/client/app/bundles/PremiumHub/components/dataExportTableAndFields.tsx
```diff
@@ -0,0 +1,234 @@
...
+  function renderCheckboxes() {
+    return Object.keys(fields).map((fieldLabel, i) => {
+      // skip Student Name because we want this field to always be present
+      if (i === 0) { return }
+      const selected = fields[fieldLabel].checked;
+      const selectedClass = selected ? "selected" : "unselected";
+      const checkboxImg = selected ? <img alt="check" src={smallWhiteCheckIcon.src} /> : ""
+      return (
+        <div className="checkbox-container">
+          <div className={`quill-checkbox ${selectedClass}`} id={fieldLabel} onClick={toggleCheckbox}>{checkboxImg}</div>
```
emilia-friedberg: I don't think we've been totally consistent about this, but in the interests of accessibility I think it's best to use buttons with `aria-labels` for the checkboxes -- example here: https://github.com/empirical-org/Empirical-Core/blob/develop/services/QuillLMS/client/app/bundles/Teacher/components/update_assigned_students/unassign_warning_modal.tsx#L8.

services/QuillLMS/client/app/bundles/PremiumHub/containers/DataExportContainer.tsx
```diff
@@ -0,0 +1,284 @@
...
+import Filters from '../components/usage_snapshots/filters'
+import { CUSTOM } from '../components/usage_snapshots/shared'
+import { Spinner } from '../../Shared/index'
+import { requestGet, } from '../../../modules/request'
+import { unorderedArraysAreEqual, } from '../../../modules/unorderedArraysAreEqual'
+import DataExportTableAndFields from '../components/dataExportTableAndFields';
+
+const filterIconSrc = `${process.env.CDN_URL}/images/icons/icons-filter.svg`
+
+const DataExportContainer = ({ adminInfo, accessType, }) => {
```
emilia-friedberg, line 16: Given just how much code is duplicated between this file and the UsageSnapshotsContainer, I wonder if it's worth extracting a lot of that shared code into a parent component for both of them. Maybe that's something we could do as a follow-up card.

services/QuillLMS/config/routes.rb
```diff
@@ -738,7 +738,8 @@
     collection do
       post :count
       get :options
-      post :top_x
```
emilia-friedberg, line 733: I would double check with @anathomical about this, but I think we do need this to be a `post` (and that the `data_export` call may need to be one too due to the length of the params).

emilia-friedberg (commented):
LTGM overall! An accessibility concern and a question about request types that I think need attention, and I do think we probably want to do a larger refactor here since `DataExportContainer` duplicates so much code from `UsageSnapshotsContainer`, but that can probably be a follow-up.

services/QuillLMS/client/app/bundles/PremiumHub/components/dataExportTableAndFields.tsx
eadams17: that makes sense-- thanks for the code reference! I'll update

services/QuillLMS/client/app/bundles/PremiumHub/containers/DataExportContainer.tsx
eadams17: I agree, I think we should do this as a follow-up card

services/QuillLMS/config/routes.rb
eadams17: I checked with him and you're right, I'll change this to a post request

emilia-friedberg (approved):
LGTM! Thanks for those changes.

anathomical (changes requested):
All of this code looks good to me.  I would like to see a set of tests for `CachePremiumReportWorker`, though.

anathomical (approved):

PR #10840, opened 2023-08-01
Upgrade to Ruby 3

brendanshean:
## WHAT
Upgrade LMS to Ruby 3

## WHY
Ruby 2.7.8 is end of life.
Ruby 3 also has many performance benefits

## HOW
Most of the work involved updating the use of [keyword parameters](https://juanitofatas.com/ruby-3-keyword-arguments)

VCR and webmock needed to be [upgraded](https://github.com/empirical-org/Empirical-Core/pull/10840/files#diff-a4f485e86e4309c49eb4ff9dbae5144afed471cb84720599f23a3c6df2e239a0R111) in order to disable them for bigquery testing.

NB:  I'm temporarily turning off the lms integration_test suite until I figure out the selenium configuration.  It's not a huge worry if that is off for a few days.

### Screenshots
(If applicable. Also, please censor any sensitive data)

### Notion Card Links
(Please provide links to any relevant Notion card(s) relevant to this PR.)

PR Checklist | Your Answer
------------ | -------------
Have you added and/or updated tests? |  YES
Have you deployed to Staging? | Not yet - deploying now!
Self-Review: Have you done an initial self-review of the code below on Github? | YES
Spec Review: Have you reviewed the spec and ensured this meets requirements and/or matches design mockups? | N/A

PR #10845, opened 2023-08-02
Remove doorkeeper

brendanshean:
## WHAT
Remove the doorkeeper gem and its references.

## WHY
The upcoming [Ruby 3](https://github.com/empirical-org/Empirical-Core/pull/10840) upgrade requires a [bumping](https://github.com/doorkeeper-gem/doorkeeper/blob/main/CHANGELOG.md#550rc2) doorkeeper to 5.5.0 which has breaking [changes](https://github.com/doorkeeper-gem/doorkeeper/blob/main/CHANGELOG.md#540rc2) from 5.4.0.  
These breaking changes require setting up a client for proper OAuth which would require a `client_id` and `client_secret` whose values are not known, which is most likely because there is no current client.

Doorkeeper was [installed](https://github.com/empirical-org/Empirical-Core/commit/0e8ba9b6bfa3dc7e8f03bdb190c72d5c6d0d818f) in the LMS' infancy and where companion apps communicated to the main app via doorkeeper and OAuth and the tables oauth_access_token, oauth_access_grants, and oauth_applications.

After investigation, I believe this functionality is no longer used.  The latest records for those associated tables is from 2019 and most are from 2018.

## HOW
Remove the gem and calls to doorkeeper.

Also, change before_action `doorkeeper_authorize!` calls to `staff!`.

### Screenshots

### Notion Card Links

PR Checklist | Your Answer
------------ | -------------


services/QuillLMS/app/controllers/api/v1/activities_controller.rb
```diff
@@ -44,12 +53,11 @@ def create
   end
 
   def destroy
-    if @activity.destroy!
```
anathomical, line 47: Haha.  I hadn't noticed that this `destroy!` call had an `if/else` wrapped around it that, obviously, could never run the `else` branch because failure to `destroy!` raises an exception.  Nice catch.

services/QuillLMS/app/lib/quill_authentication.rb
```diff
@@ -22,6 +22,37 @@ module QuillAuthentication
     )
   end
 
+  def admin!
```
anathomical, line 25: So this is kind of nit-picky, but technically, this is not `authentication` (even though we're putting it in a module called `QuillAuthentication`), it is `authorization`.  The fact that we have `current_user` means that the user has already authenticated, and now we're seeing if they're authorized for a resource.  This almost certainly doesn't matter, but it is sometimes useful to maintain a separation of the two concerns.

No suggested alternative, so this is fine really, I just wanted to flag it for possible future consideration.

services/QuillLMS/config/routes.rb
```diff
@@ -456,16 +455,16 @@
       get 'activity_health' => 'rule_feedback_histories#activity_health'
       post 'email_csv_data' => 'session_feedback_histories#email_csv_data'
 
-      resources :activities,              except: [:index, :new, :edit]
-      resources :activity_flags,          only: [:index]
-      resources :activity_sessions,       except: [:index, :new, :edit]
-      resources :feedback_histories,      only: [:index, :show, :create]
-      resources :lessons_tokens,          only: [:create]
+      resources :activities, only: [:create, :show, :update, :destroy]
```
anathomical, line 458: Very much appreciate replacement of `except` with `only`.  `except` is an easy way to accidentally add stuff you didn't want.

services/QuillLMS/spec/controllers/api/v1/firebase_tokens_controller_spec.rb
```diff
@@ -21,29 +21,6 @@
       end
     end
 
-    context 'when authenticated via OAuth' do
```
anathomical, line 24: It's hard for me to imagine that this entire controller is even in use anymore, but I appreciate cleaning up the tests anyway.

anathomical (approved):

services/QuillLMS/app/lib/quill_authentication.rb
brendanshean: I was debating  a `QuillAuthorization` concern but opted not to.  I think your comment means I should.

brendanshean: Although looking at it again I realized that the `auth_failed` method is kind of doing double duty.  I'll make a note to refactor this later.  It's definitely a good call out

services/QuillLMS/config/routes.rb
brendanshean: My feeling exactly

happythenewsad (approved):

PR #10863, opened 2023-08-07
User Rails.cache wrapper over $redis

brendanshean:
## WHAT
Make the following replacements:
 `$redis.set` -> `Rails.cache.write`
 `$redis.get` -> `Rails.cache.read`
 `$redis.del` -> `Rails.cache.delete`
 `$redis.flushdb` -> `Rails.cache.clear`
 `$redis.exists` -> `Rails.cache.exists?`

Also, fold in `$redis.expire` with `Rails.cache.write(key, value, expires_in: [expiry])` 

## WHY
[Calls](https://github.com/empirical-org/Empirical-Core/blob/c920c57566ee6cf5eeae0e266178058ef429228b/services/QuillLMS/app/models/concerns/teacher.rb#L257) to `$redis.set` with positional and keyword arguments no longer work in Ruby 3

It's also better if we have an agnostic wrapper `Rails.cache` vs the global variable `$redis`

## HOW
Find and replace

### Screenshots

### Notion Card Links

PR Checklist | Your Answer
------------ | -------------

### DIFF
--- a/services/QuillLMS/app/models/concerns/user_flagset.rb
+++ b/services/QuillLMS/app/models/concerns/user_flagset.rb
@@ -72,8 +72,6 @@ module UserFlagset
       },
     }
 
-    DEFAULT_ACTIVITY_SEARCH_CACHE_KEY = 'default_activity_search'
-
     validates :flagset, inclusion: { in: FLAGSETS.keys.map(&:to_s) }
   end
 
--- a/services/QuillLMS/app/models/concerns/user_flagset.rb
+++ b/services/QuillLMS/app/models/concerns/user_flagset.rb
@@ -6,7 +6,6 @@ module UserFlagset
 
   included do
     FLAGSETS = {
-
       production: {
         display_name: 'production',
         flags: {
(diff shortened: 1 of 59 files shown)


brendanshean:
@happythenewsad @cissyxyu @anathomical The original [PR](https://github.com/empirical-org/Empirical-Core/pull/10862) got mixed up in some rebasing.  Please ignore that PR as this is the correct one.

services/QuillLMS/app/models/concerns/user_flagset.rb
```diff
@@ -71,9 +70,10 @@ module UserFlagset
           Flags::PRIVATE =>           { display_name: 'Private' }
         }
       },
-
     }
 
+    DEFAULT_ACTIVITY_SEARCH_CACHE_KEY = 'default_activity_search'
```
anathomical: No objection to this being a constant, but is this the right file for it to be defined in?  Just checking.

anathomical (approved):
Thanks for taking the time to clean these up.

services/QuillLMS/app/models/concerns/user_flagset.rb
brendanshean: I put it in that concern because the other cache keys were populated via iterating over UserFlagset:
Screenshot 2023-08-07 at 4 35 08 PM

However, on second thought I think it belongs in activity.rb.

PR #10876, opened 2023-08-08
update scoring names around the site

emilia-friedberg:
## WHAT
Update language we use around the site to say "Frequently/sometimes/rarely demonstrated skill" rather than "Proficient/Nearly proficient/Not yet proficient".

## WHY
This is a follow-up to the scoring changes we've made recently.

## HOW
Find the places identified in the spec and update them. Note: the code itself still largely uses the language of proficiency; I made a follow-up card to clean that up but wanted to get this out sooner so teachers are less likely to be confused by a change.

### Screenshots

### Notion Card Links
https://www.notion.so/quill/Update-Names-of-Scoring-Explanation-of-Frequently-Sometimes-Rarely-Demonstrated-Skill-a439297c7de4473fbb2a4481e202b78c?pvs=4

PR Checklist | Your Answer
------------ | -------------

### DIFF
--- a/services/QuillLMS/app/models/activity_session.rb
+++ b/services/QuillLMS/app/models/activity_session.rb
@@ -59,6 +59,10 @@ class ActivitySession < ApplicationRecord
   NOT_YET_PROFICIENT = 'Not yet proficient'
   PROFICIENT = 'Proficient'
 
+  SOMETIMES_DEMONSTRATED_SKILL = 'Sometimes demonstrated skill'
+  RARELY_DEMONSTRATED_SKILL = 'Rarely demonstrated skill'
+  FREQUENTLY_DEMONSTRATED_SKILL = 'Frequently demonstrated skill'
+
   RESULTS_PER_PAGE = 25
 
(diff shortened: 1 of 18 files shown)


services/QuillLMS/app/models/activity_session.rb
```diff
@@ -59,6 +59,10 @@ class StudentNotAssignedActivityError < StandardError; end
   NOT_YET_PROFICIENT = 'Not yet proficient'
   PROFICIENT = 'Proficient'
```
anathomical, line 60: Do we need to remove the old constants, or are they still in use anywhere?

anathomical (approved):
Quick question about keeping the old constants around, but this looks good to me.  Thanks for taking care of it so quickly.

eadams17 (approved):

services/QuillLMS/app/models/activity_session.rb
emilia-friedberg: They're still in use, that'll be part of my follow-up.

PR #10891, opened 2023-08-09
Lms/admin snapshot qa 2

anathomical:
## WHAT
Address additional edge cases and issues identified during the QA process
## WHY
We want the Admin Usage Snapshots page to work as specced in all details
## HOW
- Tweak the back-end to parse the string`"null"` into the value `nil` when it's passed in as part of a request payload
- Update the display of "average" values in cases where there are 0 denominators to say "N/A"
- Tweak `options` action so that it returns `unscoped` Classrooms so that even archived classrooms end up in filter options
- Tweak front-end filter logic so that if all possible options are selected, instead of sending an array of all possible values, just send `null` and let the back-end recognize that as a lack of additional filters
- Tweak Pusher logic to recognize that `null` values for a given filter type means "all possible values"
- Adjust the base `PeriodQuery` logic to `LEFT OUTER JOIN` classroom-related tables so that it doesn't automatically exclude users without classrooms (but will exclude users if we're filtering for classrooms)

### Notion Card Links
https://www.notion.so/quill/Admin-Usage-Snapshot-Report-QA-Week-of-July-17-767225ae4fd74382bfed44861d36f589?pvs=4

PR Checklist | Your Answer
------------ | -------------


services/QuillLMS/app/lib/utils/string.rb
```diff
@@ -7,4 +7,10 @@ def self.last_name_or_name(first_name_last_name_string)
     first, last = first_name_last_name_string.split
     last || first
   end
+
+  def self.parse_null_to_nil(input)
+    return nil if input == 'null'
```
brendanshean: What about a constant for 'null'?

services/QuillLMS/app/lib/utils/numeric.rb
```diff
@@ -26,7 +26,7 @@ def self.human_readable_time_to_seconds(time)
     (time.split(":")[0].to_i * 60) + (time.split(":")[1].to_i)
   end
 
-  def self.safe_division(numerator, denominator, fallback = 0)
+  def self.safe_division(numerator, denominator, fallback = nil)
```
brendanshean, line 29: This actually seems like a better default than 0.  Not sure why I didn't think of nil.

brendanshean (approved):
Looks good to me.  I like utils spec.

services/QuillLMS/client/app/bundles/PremiumHub/components/usage_snapshots/snapshotRanking.tsx
```diff
@@ -112,17 +112,24 @@ const SnapshotRanking = ({ label, queryKey, headers, searchCount, selectedGrades
     })
   }
 
+  function selectionsEqual(firstSelection, secondSelection) {
```
emilia-friedberg: We should maybe move this function out to the `shared` file in this folder since the code is duplicated.

emilia-friedberg (approved):

happythenewsad (approved):

PR #10938, opened 2023-08-21
Evidence/parallelize invalid highlights

anathomical:
## WHAT
Refactor activity listing so that it looks up individual `invalid_highlights` per activity in parallel instead of the old method which included all of this data in the primary API call
## WHY
We had to disable pulling this data in the primary API call because it was too computationally expensive with all of the activities and was resulting in the API call getting killed by Heroku's 30 second request limit.  By running these in parallel on demand we: A) don't run them if the activity in question isn't going to be displayed, and B) keep each individual call short so that the page loads quickly
## HOW
- Add a new `invalid_highlights` controller action that provides an API call to find the invalid highlights for an individual Evidence Activity
- Refactor the front-end so that there's a stand-alone component responsible for handling the display of the checkbox/red-x for Activity Highlight invalidity
- Configure that new component to handle the "still loading data" case by displaying nothing, allowing all the activities to load into the page while for normal operation while we fetch their highlight status

### Notion Card Links
https://www.notion.so/quill/Very-long-internal-tool-loading-time-a98b11b3fa4b4c998bdee11caad45be5?pvs=4

PR Checklist | Your Answer
------------ | -------------


services/QuillLMS/client/app/bundles/Staff/utils/evidence/activityAPIs.ts
```diff
@@ -162,3 +162,14 @@ export const fetchActivityVersions = async ({ queryKey, }) => {
...
 }
+
+export const fetchInvalidHighlights = async ({ queryKey, }) => {
+  const [key, activityId]: [string, string] = queryKey
+  const response = await apiFetch(`activities/${activityId}/invalid_highlights`);
+  const invalid_highlights = await response.json();
+
+  return {
+    invalid_highlights: invalid_highlights.invalid_highlights,
+    error: handleApiError('Failed to fetch highlights_valid value, please refresh the page.', response)
```
eadams17: I think you meant to have `invalid_highlights` instead of `highlights_valid` here?

eadams17 (approved):
LGTM!

PR #10943, opened 2023-08-22
Backfill capitalize names

brendanshean:
## WHAT
A previous [PR](https://github.com/empirical-org/Empirical-Core/pull/10941) will be restricting the `before_save :capitalize_name` callback so that it only runs when the name is changed.  This backfill will ensure that we run CapitalizeNames.capitalize on all users at least once.

## WHY
user.capitalize_names was recently updated to utilize a gem that handles capitalizing of names properly.  We'd like to ensure that this is run on all user names so that their names are properly capitalized.

## HOW
`rake users:back_fill_capitalize_names`

FYI, it took 24 minutes on staging.
bundle exec rake users:backfill_capitalize_names  629.65s user 24.22s system 45% cpu 23:55.67 total

### Screenshots

### Notion Card Links

PR Checklist | Your Answer
------------ | -------------


services/QuillLMS/lib/tasks/temporary/users.rake
```diff
@@ -90,4 +90,13 @@ namespace :users do
     User.where(id: user_ids).destroy_all
     puts "Deleted #{user_ids.count} users"
   end
+
+  task backfill_capitalize_names: :environment do
+    progress_bar = ProgressBar.new(User.count)
+
+    User.find_each do |user|
+      user.capitalize_name
```
anathomical, line 98: Do we need a `save` call after the `capitalize_name` call?

anathomical (commented):
Quick question about whether this actually works as intended.

services/QuillLMS/lib/tasks/temporary/users.rake
brendanshean: Great point.  This actually doesn't do anything at the moment.

brendanshean:
After adding the `user.save` the rake task will now take in excess of a 14 days on staging.  Due to this cost, I'm inclined to not do the backfill.

PR #10958, opened 2023-08-24
Fix bug with Clever district_admin flow

brendanshean:
## WHAT
1. Fix a [bug](https://quillorg-5s.sentry.io/issues/3665849322/?environment=production&project=11238&query=clever&referrer=issue-stream&statsPeriod=24h&stream_index=1) involving Clever district admin login/signup flow
1. Clean up handling of auth_hash

## WHY
1. Clever district admins are sent to 500 page when the attempt to login/signup
1. The parser is importing the wrong keys

## HOW
1. Add a new template district_success, indicating that basic district info was imported but also indicate District Admins cannot login.
1. Use the [:info][:district] key and then use district client to fetch the district name.  The name key in the omniauth hash is the district admin's name.

### Screenshots
Screenshot 2023-08-24 at 10 08 03 AM

### Notion Card Links

PR Checklist | Your Answer
------------ | -------------


services/QuillLMS/app/controllers/auth/clever_controller.rb
```diff
@@ -16,8 +16,8 @@ def clever
       @auth_hash ||= request.env['omniauth.auth']
     end
 
-    private def district_success(data, redirect=nil)
-      render status: 200, nothing: true # Don't bother rendering anything.
+    private def district_success(data, redirect)
```
happythenewsad: can we remove both params here, so that the method signature matches the one in controllers/clever_controller?

services/QuillLMS/app/views/clever/district_success.html.erb
```diff
@@ -0,0 +1,28 @@
+<!DOCTYPE html>
+<html>
+  <head>
+    <style>
+      body {
+        background-color: #f7f5f5
```
happythenewsad, line 6: No variable for hex color

happythenewsad (changes requested):

services/QuillLMS/app/views/clever/district_success.html.erb
brendanshean: That particular color is not in variables.scss and there are 45 other occurrences.  I think that's out of scope for this PR.

Screenshot 2023-08-24 at 10 38 56 AM

happythenewsad (approved):

PR #10988, opened 2023-08-30
Lms/admin snapshots filters teachers without classrooms

anathomical:
## WHAT
- Update admin filter generation logic to include Teachers that do not have classrooms (as long as they're associated with a relevant school)
- Update the logic for applying filters to API calls so that it better understands when "all" items from a filter are selected because we want to submit `null` as the value for that category filter in those cases instead of having a huge array of redundant IDs
- Update `custom` timeframe calculation so that it has the same start and end of day as other time-frames
## WHY
- We want to count and include teachers in our stats even if they don't have classrooms
- Simplifying the "all selected" logic makes our API calls and caching less complex, and also ensures that we don't mistakenly exclude items in cases where the logic that populates the filters is off but the user doesn't care to apply filters of that class
- We want someone who puts in a custom timeframe to get the same results as a matching pre-defined timeframe
## HOW
- Use `LEFT OUTER JOINS` for `classroom` relationships when generating teacher filters so that if the teacher has no classroom they still end up in the results
- Add an extra condition when deciding whether or not to send `null` as a filter value so that if `selected` items is a superset of `all` items (containing at least every member of `all`), we count that as all-selected even if `selected` has additional items in it.
- Just add `beginning_of_day` and `end_of_day` to custom timeframe calculation lambdas

### Notion Card Links
https://www.notion.so/quill/Admin-Usage-Snapshot-Report-QA-Week-of-July-17-767225ae4fd74382bfed44861d36f589?d=1c41c3ee821141fcacc2c8054d38b703#2fa451e21bca4905bac0a92959ac9097

PR Checklist | Your Answer
------------ | -------------

### DIFF
--- a/services/QuillLMS/client/app/bundles/PremiumHub/containers/UsageSnapshotsContainer.tsx
+++ b/services/QuillLMS/client/app/bundles/PremiumHub/containers/UsageSnapshotsContainer.tsx
@@ -4,6 +4,7 @@ import * as _ from 'lodash'
 import * as Pusher from 'pusher-js';
 
 import { FULL, restrictedPage, } from '../shared';
+import { selectionsEqual, } from '../components/usage_snapshots/shared'
 import CustomDateModal from '../components/usage_snapshots/customDateModal'
 import SnapshotSection from '../components/usage_snapshots/snapshotSection'
 import Filters from '../components/usage_snapshots/filters'
@@ -239,7 +240,12 @@ const UsageSnapshotsContainer = ({ adminInfo, accessType, }) => {
   function handleClickDownloadReport() { window.print() }
 
   function mapItemsIfNotAll(selectedItems, allItems, mapKey = 'id') {
-    if (unorderedArraysAreEqual(selectedItems, allItems)) return null
+    // selectedItems may, by design, be a superset of allItems, but if everything in allItems is in selectedItems, we want to treat it as "everything" being selected
+    const allItemsSelected = allItems.every((i) => {
+      return _.some(selectedItems, i)
+    })
+
+    if (allItemsSelected || selectionsEqual(selectedItems, allItems)) return null
 
     return selectedItems.map(i => i[mapKey])
   }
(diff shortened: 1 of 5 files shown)


services/QuillLMS/client/app/bundles/PremiumHub/containers/UsageSnapshotsContainer.tsx
```diff
@@ -239,7 +240,12 @@ const UsageSnapshotsContainer = ({ adminInfo, accessType, }) => {
   function handleClickDownloadReport() { window.print() }
 
   function mapItemsIfNotAll(selectedItems, allItems, mapKey = 'id') {
-    if (unorderedArraysAreEqual(selectedItems, allItems)) return null
+    // selectedItems may, by design, be a superset of allItems, but if everything in allItems is in selectedItems, we want to treat it as "everything" being selected
+    const allItemsSelected = allItems.every((i) => {
```
emilia-friedberg: nit, but I always read `i` as `index`.

emilia-friedberg (approved):

services/QuillLMS/client/app/bundles/PremiumHub/containers/UsageSnapshotsContainer.tsx
anathomical: Good call-out.  Worth making that clearer.

services/QuillLMS/client/app/bundles/PremiumHub/containers/UsageSnapshotsContainer.tsx
```diff
@@ -239,7 +240,12 @@ const UsageSnapshotsContainer = ({ adminInfo, accessType, }) => {
   function handleClickDownloadReport() { window.print() }
 
   function mapItemsIfNotAll(selectedItems, allItems, mapKey = 'id') {
-    if (unorderedArraysAreEqual(selectedItems, allItems)) return null
+    // selectedItems may, by design, be a superset of allItems, but if everything in allItems is in selectedItems, we want to treat it as "everything" being selected
+    const allItemsSelected = allItems.every((item) => {
```
happythenewsad, line 244: Personally I think
`_.intersection(...).length` is more readable and probably a lower runtime complexity, although I didn't check

happythenewsad (approved):

services/QuillLMS/client/app/bundles/PremiumHub/containers/UsageSnapshotsContainer.tsx
anathomical: The idea here would be to do something like:
```javascript
const allItemsSelected = _.intersection(selectedItems, allItems).length === allItems.length
```

PR #11007, opened 2023-09-05
Loosen sentence length requirement on Connect/Diagnostic

cissyxyu:
## WHAT
Shorten the minimum sentence length from 2 words shorter than the optimal response, to 4 words shorter than the optimal response.

## WHY
Right now our requirement for sentence minimum length is a bit too strict. It requires sentences to be at most two words shorter than the optimal response, which is too strict.

## HOW
Change the logic in the marking logic to allow for up to 3 words shorter than the optimal response.

### Screenshots

### Notion Card Links

PR Checklist | Your Answer
------------ | -------------

### DIFF
--- a/lambdas/rematching/package.json
+++ b/lambdas/rematching/package.json
@@ -11,7 +11,7 @@
   "dependencies": {
     "lodash": "^4.17.21",
     "pg": "^7.7.1",
-    "quill-marking-logic": "0.15.10",
+    "quill-marking-logic": "0.15.11",
     "request": "^2.88.0",
     "request-promise": "^4.2.2",
     "sequelize": "^4.44.3",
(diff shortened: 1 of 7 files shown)


emilia-friedberg (commented):
I see the test updates here but I don't actually see a change to the algorithm itself (which makes me wonder if we need negative tests as well as positive ones -- for instance confirming that it returns false if the response string is only three words shorter).

cissyxyu:
> I see the test updates here but I don't actually see a change to the algorithm itself (which makes me wonder if we need negative tests as well as positive ones -- for instance confirming that it returns false if the response string is only three words shorter).

Added the change. And added new tests to confirm the change!

emilia-friedberg (approved):

brendanshean (approved):

PR #11025, opened 2023-09-12
handle stripe subscription with no card

emilia-friedberg:
## WHAT
Handle stripe subscriptions where the user paid with a bank account directly instead of a card.

## WHY
We don't want this page to error when the user paid with a bank account.

## HOW
Update the `stripe_payment_method` function to just return `nil` when `&.card` is not a function.

Note: the UI for this page will show "Credit card ending in null" etc, instead of properly addressing it as a bank account. Since that would require a much more substantial rewrite, and this is currently only affecting one teacher, I think that's an acceptable tradeoff for now, but will flag it to support in case the issue becomes more widespread and we want to take it on as a product task.

### Screenshots

### Notion Card Links
https://www.notion.so/quill/500-error-on-teacher-s-My-Subscriptions-page-95016a01881e4f8a82faef4e758a29dd

PR Checklist | Your Answer
------------ | -------------

### DIFF
--- a/services/QuillLMS/app/models/stripe_integration/subscription.rb
+++ b/services/QuillLMS/app/models/stripe_integration/subscription.rb
@@ -53,7 +53,7 @@ module StripeIntegration
 
     private def stripe_payment_method
       Stripe::PaymentMethod.retrieve(stripe_payment_method_id)&.card
-    rescue Stripe::InvalidRequestError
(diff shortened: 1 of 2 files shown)


brendanshean (approved):
Nice sleuthing on this!  I also agree with the tradeoff decision here.

PR #11042, opened 2023-09-14
surface better errors for staff school creation

happythenewsad:
## WHAT / HOW
Rescues exceptions during ActiveRecord validation that are not handled by ActiveRecord. Typically, these will be Postgres errors. Example: when an index constraint is violated. 

## WHY
So instead of this:
Screenshot 2023-09-14 at 2 09 30 PM

Users get an actionable error message, and are redirected back the the create form:

Screenshot 2023-09-14 at 7 16 49 PM

### Screenshots

### Notion Card Links
https://www.notion.so/quill/Unable-to-create-New-Schools-in-CMS-006dc7f847cf4787baf3221e972e12d8

PR Checklist | Your Answer
------------ | -------------

### DIFF
--- a/services/QuillLMS/app/models/application_record.rb
+++ b/services/QuillLMS/app/models/application_record.rb
@@ -6,15 +6,26 @@ class ApplicationRecord < ActiveRecord::Base
   connects_to database: { writing: :primary, reading: :replica }
 
   # Rescues previously unhandled exceptions, like ActiveRecord::RecordNotUnique
-  def safer_save
+  def save_with_error_handling
     save
   rescue ActiveRecord::RecordNotUnique => e
-    /Key \((\w+)\)/ =~ e.message
+    handle_not_unique_error(e)
+  rescue StandardError => e
+    handle_standard_error(e)
+  end
+
+  def handle_not_unique_error(exception)
+    /Key \((\w+)\)/ =~ exception.message
     attribute_sym = Regexp.last_match(1) ? Regexp.last_match(1).to_sym : :base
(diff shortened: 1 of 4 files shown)


services/QuillLMS/app/models/application_record.rb
```diff
@@ -4,4 +4,15 @@ class ApplicationRecord < ActiveRecord::Base
   self.abstract_class = true
 
   connects_to database: { writing: :primary, reading: :replica }
+
+  # Rescues previously unhandled exceptions, like ActiveRecord::RecordNotUnique
+  def safer_save
+    save
+  rescue ActiveRecord::RecordNotUnique => e
+    errors.add(:nces_id, 'not unique')
+    false
```
happythenewsad, line 22: My intention is that this method can be extended with additional rescue blocks if necessary. If you don't want to craft pretty messaging like I did here, you can let the last rescue print a more general message.

emilia-friedberg (approved):

services/QuillLMS/app/models/application_record.rb
anathomical: Do we want to use `ErrorNotifier` to send these errors to Sentry and NewRelic, or do we think that swallowing these edge cases until/unless someone investigates payload responses for errors?  I could see either way making sense, but just wanted to confirm where your head's at.

happythenewsad: Yeah, I think adding ErrorNotifier would be nice here.

PR #10812, opened 2023-07-27
Upgrade RailsAdmin to 3.1.2

brendanshean:
## WHAT
Upgrade RailsAdmin gem to 3.1.2

## WHY
There's a [regression](https://quillorg-5s.sentry.io/issues/4343112355/?project=11238&query=is%3Aunresolved&referrer=issue-stream&statsPeriod=14d&stream_index=1) 
Also, it's required for Rails 7

## HOW
Bump gemfile.
Clean up `config.authorize_with` block to calculate current_user directly.

### Screenshots
(If applicable. Also, please censor any sensitive data)

### Notion Card Links
(Please provide links to any relevant Notion card(s) relevant to this PR.)

PR Checklist | Your Answer
------------ | -------------
Have you added and/or updated tests? |  No. Configuration change.
Have you deployed to Staging? | Not yet - deploying now!
Self-Review: Have you done an initial self-review of the code below on Github? | YES
Spec Review: Have you reviewed the spec and ensured this meets requirements and/or matches design mockups? | N/A

PR #11045, opened 2023-09-15
Upgrade LMS Ruby to 3.1.4

cissyxyu:
## WHAT
Upgrade LMS Ruby version to 3.1.4.

## WHY
For security reasons we should always have an up-to-date ruby version. Also Heroku requires this version to upgrade to the latest Heroku stack.

## HOW
Set the ruby version to 3.1.4.

Other Gemfile remediations:
1. The gem `timecop` needed an update to be compatible with Ruby 3.1
2. The gem `mini_racer` was no longer compatible, but we're no longer using it so I removed the gem entirely.
3. The gem `matrix` needs to be explicitly installed now for compatibility with Heroku deploy. So I moved it to the Gemfile.
4. The gem `psych` has a known issue where only versions below 4.0 are compatible with Ruby >= 3.1. [This issue is documented here.](https://stackoverflow.com/questions/71191685/visit-psych-nodes-alias-unknown-alias-default-psychbadalias) Currently the most efficient way to solve it is to lock our psych gem at version <= 4.0. 

Non-gem remediations:
1. Ruby 3.1 has some differences when used with RailsAdmin 3.1. We need to explicitly set a setting called `config.asset_source` and add some stylesheets. All these changes were automatically generated by running `rails g rails_admin:install` as recommended by the command line after I upgraded to Ruby 3.1. 
2. One spec was failing because error messaging generated in Ruby 3.1 is more detailed and has more information. So I fixed the failing spec by loosening the expect statement to find a substring within the error string rather than an exact match.

### Screenshots

### Notion Card Links
https://www.notion.so/quill/Upgrade-LMS-Ruby-version-to-3-1-4-3079da026ff448e9979b4c98e603469f?pvs=4

PR Checklist | Your Answer
------------ | -------------


services/QuillLMS/Gemfile
```diff
@@ -144,6 +144,7 @@ gem 'sentry-ruby'
 gem 'sentry-rails'
 gem 'sentry-sidekiq'
 gem 'rack-heartbeat'
+gem "matrix", '~> 0.4.2'
```
brendanshean: If this is only for heroku deploy, let's move this to the :production, :staging group

services/QuillLMS/config/initializers/rails_admin.rb
```diff
@@ -2,6 +2,7 @@
 
 Rails.application.config.eager_load do
   RailsAdmin.config do |config|
+    config.asset_source = :webpacker
```
brendanshean: I [upgraded](https://github.com/empirical-org/Empirical-Core/pull/10812) RailsAdmin to 3.1.2 in preparation for the Rails 7 upgrade a little over a month ago.  AFAICT, the warning about asset source is incorrect and it still exists on this branch even after adding the asset_source = :webpacker.  (e.g. run rails console in your local environment):

Screenshot 2023-09-18 at 8 16 00 AM

Unless there is something in the Rails Admin that you noticed not working properly, I would recommend removing the following changes in this PR:
```
services/QuillLMS/app/javascript/packs/rails_admin.js
services/QuillLMS/app/javascript/stylesheets/rails_admin.scss
services/QuillLMS/config/initializers/rails_admin.rb
services/QuillLMS/package.json
```
since I don't think they do anything.

brendanshean (changes requested):
Nice work on figuring out the psych complication.  There's a comment about removing some code but once that's done, this should be good to go.

anathomical (approved):

brendanshean (approved):
Thank you for figuring out the dependencies on this upgrade

PR #11051, opened 2023-09-19
Deduplicate logic shared in Premium Admin pages

eadams17:
## WHAT
deduplicate shared logic and refactor Premium hub pages

## WHY
we want to separate out this shared logic to keep things cleaner

## HOW
create a new `PremiumFilterableReportsContainer` parent container that houses the shared filtering logic; refactor `DataExportContainer` and `UsageSnapshotsContainer`

### Screenshots

### Notion Card Links
https://www.notion.so/quill/Admin-Reports-deduplication-Data-Export-Usage-Snapshot-Report-Growth-Reports-48bcc3b119df409bb72631d72e05cf60?pvs=4

PR Checklist | Your Answer
------------ | -------------


services/QuillLMS/client/package-lock.json
eadams17, line 1: I'm not really sure why this is getting added-- I pulled from develop but it's still showing as changed. Please disregard-- I will investigate tomorrow

services/QuillLMS/app/helpers/footer_helper.rb
```diff
@@ -64,4 +64,11 @@ def about_us_links
       { href: '/contact', label: 'Contact Us' }
     ]
   end
+
+  def show_footer(path)
+    excluded_paths = ['/teachers/premium_hub/usage_snapshot_report', '/teachers/premium_hub/data_export']
+    return false if excluded_paths.include?(path)
+
+    true
```
emilia-friedberg: I think you could just do `excluded_paths.exclude?(path)` here as the last line.

services/QuillLMS/client/app/bundles/PremiumHub/containers/PremiumFilterableReportsContainer.tsx
```diff
@@ -0,0 +1,295 @@
...
+          passedEndDate={customEndDate}
+          passedStartDate={customStartDate}
+          setCustomDates={setCustomDates}
+        />
+      )}
+      <Filters
+        {...filterProps}
+      />
+      {shouldRenderDataExportContainer && <DataExportContainer {...sharedProps} />}
+      {shouldRenderUsageSnapshotsContainer && <UsageSnapshotsContainer {...sharedProps} />}
```
emilia-friedberg, line 290: I think you should be able to use a `<Routes><Route>...` structure here (or a `<Switch><Route>...` if the v6 syntax doesn't work for some reason), right?

emilia-friedberg (approved):
A couple of small comments, but nothing blocking -- I do think we should try to get the component rendering working with React Router if possible, but not the end of the world if not. Thanks for this cleanup, it's so nice to have allllllll that code deduplicated.

services/QuillLMS/app/helpers/footer_helper.rb
brendanshean: Since this method is return a boolean lets rename it `show_footer?`.  You could also pull out a constant here:

```
 EXCLUDED_FOOTER_PATHS = [
  '/teachers/premium_hub/usage_snapshot_report', 
  '/teachers/premium_hub/data_export'
].freeze

def show_footer?(path)
  !path.in?(EXCLUDED_FOOTER_PATHS)
end
```

brendanshean (approved):
Thank you for cleaning up the duplication here!  It's nice to see some more RTL

services/QuillLMS/client/app/bundles/PremiumHub/containers/PremiumFilterableReportsContainer.tsx
eadams17: I originally tried to go that route, but what was happening is that with the Router structure, these components were getting re-rendered every single time one of the filter-related state variables was changed (and then passed via `sharedProps`). This was causing `getData` to be called prematurely instead of just when `searchCount` updates

eadams17:
> Thank you for cleaning up the duplication here! It's nice to see some more RTL

Thanks for the suggestion! I've made that update

PR #11061, opened 2023-09-25
DataTable aggregate rows feature

eadams17:
## WHAT
add functionality to the `DataTable` component to handle aggregate row data

## WHY
we need this in preparation for the admin Growth Diagnostic reports

## HOW
update logic in the `DataTable` component to check for and handle aggregate row data and add an example to Backpack

just a note: I went with the name `aggregate_rows` for the identifier. @anathomical, I know we had said `break_outs`, but I felt like this made more sense. I'm totally open to changing it back, or to any other naming convention that @emilia-friedberg might feel like works better. I also plan on adding more thorough testing once the reports are actually implemented since this functionality may end up changing

### Screenshots
Screen Shot 2023-09-25 at 4 49 00 PM
Screen Shot 2023-09-25 at 5 11 49 PM

### Notion Card Links
https://www.notion.so/quill/New-Project-Card-Admin-Diagnostic-Reports-Spec-For-Three-Reports-15ace9cefb7c4bf7b5475b041fff56d9?pvs=4

PR Checklist | Your Answer
------------ | -------------


services/QuillLMS/client/app/bundles/Shared/components/shared/dataTable.tsx
```diff
@@ -340,10 +367,24 @@ export class DataTable extends React.Component<DataTableProps, DataTableState> {
...
+    const aggregateRowIdentifier = `${sectionText}-${row.id}`
+    const shouldDisplayAggregateIcon = isAggregateRow && i === 0
+    const shouldDisplayToggleButton = row.aggregate_rows && row.aggregate_rows.length && i === 0
+    const aggregateRowsDisplayed = expandedParentRowIdentifier && expandedParentRowIdentifier === aggregateRowIdentifier
+    let aggregateRowButtonOrIcon = <span />
+
+    if (shouldDisplayAggregateIcon) {
+      aggregateRowButtonOrIcon = <img alt="" className="aggregate-row-icon" src={aggregateRowArrowSrc} />
+    } else if (shouldDisplayToggleButton && aggregateRowsDisplayed) {
+      aggregateRowButtonOrIcon = <button className="aggregate-row-toggle interactive-wrapper focus-on-light" onClick={this.handleHideAggregateRows}><img alt="" className="aggregate-row-toggle-icon" src={toggleArrowExpandedSrc} /></button>
```
emilia-friedberg: This button, and the next one, need an aria-label so that screenreader users know what this button does.

services/QuillLMS/client/app/bundles/Shared/components/shared/dataTable.tsx
```diff
@@ -376,17 +418,31 @@ export class DataTable extends React.Component<DataTableProps, DataTableState> {
...
+    const { expandedParentRowIdentifier } = this.state
     const { headers, } = this.props
     const rowClassName = `data-table-row ${row.checked ? 'checked' : ''} ${row.className || ''}`
-    const rowSections = headers.map(header => this.renderRowSection(row, header))
+    const rowSections = headers.map((header, i) => this.renderRowSection({ row, header, isAggregateRow, i }))
     const rowContent = <React.Fragment>{this.renderRowCheckbox(row)}{this.renderRowDragHandle(row)}{rowSections}{this.renderRowRemoveIcon(row)}</React.Fragment>
-
+    let rowElement = <tr className={rowClassName} key={String(row.id)}>{rowContent}</tr>
+    const aggregateRowIdentifier = `${row[headers[0].attribute]}-${row.id}`
+    const showAggregateRows = row.aggregate_rows && row.aggregate_rows.length && expandedParentRowIdentifier && expandedParentRowIdentifier === aggregateRowIdentifier
```
emilia-friedberg: I think this could just be shortened to `const showAggregateRows = row.aggregate_rows?.length && expandedParentRowIdentifier === aggregateRowIdentifier`, because neither of those calls should error so the initial null checks aren't necessary.

services/QuillLMS/client/app/bundles/Shared/components/shared/dataTable.tsx
```diff
@@ -340,10 +367,24 @@ export class DataTable extends React.Component<DataTableProps, DataTableState> {
     }
 
     const rowDisplayText = linkDisplayText || sectionText
+    const aggregateRowIdentifier = `${sectionText}-${row.id}`
+    const shouldDisplayAggregateIcon = isAggregateRow && i === 0
+    const shouldDisplayToggleButton = row.aggregate_rows && row.aggregate_rows.length && i === 0
+    const aggregateRowsDisplayed = expandedParentRowIdentifier && expandedParentRowIdentifier === aggregateRowIdentifier
+    let aggregateRowButtonOrIcon = <span />
```
emilia-friedberg: could this line just be `let aggregateRowButtonOrIcon` without setting it equal to `<span />`? It doesn't really matter visually but it does create pretty busy HTML (as evidence by all those snapshot updates). I know sometimes the null element thing creates weird behavior though so if it doesn't work to remove it it's all good.

emilia-friedberg (approved):
I like this implementation a lot. I have some minor tweaks requested around concision and accessibility, but nothing I need to review again. Thanks for taking care of this!

emilia-friedberg:
@eadams17 one more question -- I think we'd been planning on making some of the generic styling changes to `DataTable` as part of this project (with the new header background color, etc). Is that still the case?

anathomical (approved):
This looks reasonable to me.  I wonder if it's worth adding some additional tests to the `dataTable.test.tsx` file.  We've got React testing in use, so maybe checking to make sure that things toggle as expected via tests is worth doing if that's straightforward is worth spending a bit of time on?

eadams17:
> @eadams17 one more question -- I think we'd been planning on making some of the generic styling changes to `DataTable` as part of this project (with the new header background color, etc). Is that still the case?

that will be out of scope for this project, but I think the plan is still to eventually do a universal overhaul with the changes (I'll check in with Jack about it)

eadams17:
> This looks reasonable to me. I wonder if it's worth adding some additional tests to the `dataTable.test.tsx` file. We've got React testing in use, so maybe checking to make sure that things toggle as expected via tests is worth doing if that's straightforward is worth spending a bit of time on?

for sure-- I made a note though in the How section about my plan to add more tests once we actually implement this toggle functionality with the new reports as we will probably end up having to make some tweaks to it

services/QuillLMS/client/app/bundles/Shared/components/shared/dataTable.tsx
eadams17: yeah, that's a good point-- I'll make this change, and if there is any weird behavior down the line, I'll just revert to using a span

services/QuillLMS/client/app/bundles/Shared/components/shared/dataTable.tsx
eadams17: will do! thanks for calling out

PR #11064, opened 2023-09-26
Add Vertex AI backfill rake task

brendanshean:
## WHAT
Add a rake task to populate pre-existing `Evidence::AutomlModel` records

## WHY
There is a UI to add new Evidence::AutomlModel but this rake task will add existing ones to our database

## HOW
There are two rake tasks.
`populate_prompt_endpoint_and_model_ids`:  This will populate a csv with the correct values for `endpoint_external_id`, `model_external_id`,`prompt_id`. and `labels`.  Curriculum requested this updated file which is why there are two rake tasks.

With the values from the previous rake task, run`backfill_evidence_automl_models` to create the new records.
Raw sql is used since the Evidence::AutomlModel does not point to the newer table yet (that change happens downstream in `bs-vertex-backfill`)

Logistics:
The `populate_prompt_endpoint_and_model_ids` task has to be run first on a downstream branch `bs-vertex-ai` so that it has access to the vertex ai platform and the params builder.  Once the values are populated into vertex_backfill.csv, `backfill_evidence_automl_models` can be run on this branch `bs-vertex-backfill`. 

I've also added a deploy script [here](https://www.notion.so/quill/Implement-Vertex-AI-in-Evidence-with-parity-for-AutoML-features-cc6f6f4562a544b489bd2eff7ec0f686?pvs=4#4823dc30da824bafa4c137f8e2e1f669)

### Screenshots

### Notion Card Links

PR Checklist | Your Answer
------------ | -------------

### DIFF
--- a/services/QuillLMS/lib/tasks/temporary/vertex_ai_migration.rake
+++ b/services/QuillLMS/lib/tasks/temporary/vertex_ai_migration.rake
@@ -9,12 +9,8 @@ namespace :vertex_ai  do
 
     endpoint_client = ::Google::Cloud::AIPlatform::V1::EndpointService::Client.new
     parent = "projects/#{ENV.fetch('VERTEX_AI_PROJECT_ID')}/locations/#{ENV.fetch('VERTEX_AI_LOCATION')}"
-    i = 0
 
     CSV.parse(pipe_data, headers: true) do |row|
-      puts i if i % 10 == 0
-      i += 1
-
       row_data = row.to_h
       conjunction = row_data['Conjunction'].downcase
       evidence_activity_id = row_data['Activity ID']
@@ -33,7 +29,7 @@ namespace :vertex_ai  do
       rows_data << row_data
     end
 
-    CSV.open("vertex_output.csv", "wb") do |csv|
+    CSV.open("vertex_backfill.csv", "wb") do |csv|
       csv << rows_data.first.keys
 
       rows_data.each do |row|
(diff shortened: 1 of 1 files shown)


services/QuillLMS/lib/tasks/temporary/vertex_ai_migration.rake
```diff
@@ -0,0 +1,80 @@
...
+        csv << row.values
+      end
+    end
+  end
+
+  task backfill_evidence_automl_models: :environment do
+    pipe_data = $stdin.read unless $stdin.tty?
+
+    CSV.parse(pipe_data, headers: true) do |row|
+      next if row['Model ID'].blank? || row['Endpoint ID'].blank? || row['Prompt ID'].blank? || row['Model ID'].blank? || row['Name'].blank? || row['Labels'].blank?
```
anathomical: Is `row['Model ID']` in this list of checks twice?

services/QuillLMS/lib/tasks/temporary/vertex_ai_migration.rake
```diff
@@ -0,0 +1,80 @@
...
+      endpoint = endpoint_client.list_endpoints(parent: parent).find { |e| e.display_name.start_with?(name) }
+      row_data['Endpoint ID'] = endpoint&.name&.split('/')&.last
+      model = endpoint&.deployed_models&.find { |m| endpoint.display_name == m.display_name && m.display_name.start_with?(name) }
+      row_data['Model ID'] = model&.model&.split('/')&.last
+      row_data['Name'] = endpoint&.display_name
+      row_data['Labels'] = Evidence::VertexAI::ParamsBuilder.run(row_data['Name'])[:labels]&.join(',')
+      rows_data << row_data
+    end
+
+    CSV.open("vertex_backfill.csv", "wb") do |csv|
```
anathomical, line 32: What happens if we run this using `heroku run ...`?  I presume we can't actually access a local file on whatever container does the job.  Given that this is planned to be a "run once" thing, it may not matter, but maybe a note that this job only works when run on a local machine might be worth throwing in?

anathomical (approved):
Seems like a reasonable approach to me.  I assume that there won't be rows in the source CSV for the first task that will be clobbered by the job.  I'm assuming that the plan, which I don't remember being made explicit, is to run these rake tasks from a local machine connected to the production database?

services/QuillLMS/lib/tasks/temporary/vertex_ai_migration.rake
brendanshean: Good point!

services/QuillLMS/lib/tasks/temporary/vertex_ai_migration.rake
brendanshean: That makes sense.  I kind of view these lib/tasks/temporary as one-off scripts but I should make my assumptions transparent.

PR #11072, opened 2023-09-28
lms/classroom-roster-lists-bugfix

anathomical:
## WHAT
Use a safe accessor to handle weird edge cases
## WHY
If a user imports a roster through Google or Clever and then later disconnects their account from third-party auth, the current code will error because it tries to access value out of null
## HOW
Use a safe accessor (`?.`) so that in cases where `user.provider` is `null`, and thus we have an undefined value from `providerConfigLookup`, we don't try to access `title` from it for comparison.  Instead we'll get `undefined` which will set the value of `isDisabled` to `false` (since `undefined` won't equal whatever `classroomProvider` is set to).

### Notion Card Links
https://www.notion.so/quill/My-Classes-page-goes-blank-when-attempting-to-open-rosters-ee0707bb13764c62804cd01ed20c555e

PR Checklist | Your Answer
------------ | -------------
Have you added and/or updated tests? |  No.  I considered trying to refactor the test for `react-testing` library and adding a case for this, but it seemed like overkill for a safe accessor.

### DIFF
--- a/services/QuillLMS/client/app/bundles/Teacher/components/classrooms/classroom_student_section.tsx
+++ b/services/QuillLMS/client/app/bundles/Teacher/components/classrooms/classroom_student_section.tsx
@@ -585,7 +585,7 @@ const ClassroomStudentSection = ({
     const { classroomProvider } = classroom
 
     if (classroomProvider) {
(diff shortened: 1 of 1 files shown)


brendanshean (approved):
Thank you for fixing this bug.

PR #11075, opened 2023-09-29
fix diagnostic report display issue by stricter filtering by class roster

happythenewsad:
fix diagnostic report display issue by stricter filtering by class roster

## WHAT / HOW
For diagnostic reports, check assigned_student_ids against the students rostered in the given classroom 

## WHY
If Classroom Units exist from recycling classes, or other teacher fixes, students who have transferred out of class A may still appear in class A's diagnostic report, causing confusion. 

### Screenshots

### Notion Card Links
https://www.notion.so/quill/Students-appearing-in-Practice-recommendations-report-for-a-class-in-which-they-are-no-longer-roster-66b201d11b6240babbf971585212bd4d

PR Checklist | Your Answer
------------ | -------------

### DIFF
--- a/services/QuillLMS/app/controllers/concerns/diagnostic_reports.rb
+++ b/services/QuillLMS/app/controllers/concerns/diagnostic_reports.rb
@@ -83,6 +83,7 @@ module DiagnosticReports
   private def assigned_student_ids_filtered_by_classroom_roster(classroom_units)
     classroom_id = classroom_units.first&.classroom_id
     return [] unless classroom_id
+
     assigned_student_ids = classroom_units.map { |cu| cu.assigned_student_ids }.flatten.uniq
     rostered_student_ids = Classroom.find(classroom_id).students.pluck(:id)
     assigned_student_ids.intersection(rostered_student_ids)
--- a/services/QuillLMS/app/controllers/concerns/diagnostic_reports.rb
+++ b/services/QuillLMS/app/controllers/concerns/diagnostic_reports.rb
@@ -57,7 +57,6 @@ module DiagnosticReports
     end
   end
 
-  # rubocop:disable Metrics/CyclomaticComplexity
   def set_activity_sessions_and_assigned_students_for_activity_classroom_and_unit(activity_id, classroom_id, unit_id=nil, hashify_activity_sessions: false)
     if unit_id
       classroom_unit = ClassroomUnit.find_by(unit_id: unit_id, classroom_id: classroom_id)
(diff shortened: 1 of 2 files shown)


services/QuillLMS/app/controllers/concerns/diagnostic_reports.rb
```diff
@@ -57,7 +57,6 @@ def summarize_correct_skills(number_correct, number_incorrect)
     end
   end
 
-  # rubocop:disable Metrics/CyclomaticComplexity
```
happythenewsad, line 60: less cyclomatic complexity - bonus!

services/QuillLMS/app/controllers/concerns/diagnostic_reports.rb
```diff
@@ -80,7 +79,12 @@ def set_activity_sessions_and_assigned_students_for_activity_classroom_and_unit(
 
     @activity_sessions = @activity_sessions.to_h { |session| [session.user_id, session] }
   end
-  # rubocop:enable Metrics/CyclomaticComplexity
+
+  private def assigned_student_ids_filtered_by_classroom_roster(classroom_id, classroom_units)
+    assigned_student_ids = classroom_units.map { |cu| cu.assigned_student_ids }.flatten.uniq
+    rostered_student_ids = Classroom.find(classroom_id).students.pluck(:id)
+    assigned_student_ids.intersection(rostered_student_ids)
+  end
```
anathomical: Since `classroom_units` include a `classroom_id` value, I wonder if we can simplify this function further by not even passing a `classroom_id` and just checking the appropriate `Classroom` based on the `ClassroomUnit`.  Given that `classroom_units` should all belong to the same `classroom_id`, you could just do `rostered_student_ids = classroom_units.first.classroom.students.pluck(:id)`.  I suppose that does risk some future use of this function where the array of `classroom_units` don't all belong to the same `Classroom`, but I think that might be a risk worth taking given that avoids redundancy in the method signature.

anathomical (approved):
Love getting to remove cyclomatic complexity disables, so nice job on that.  The general approach seems fine to me.  I've got a take-it-or-leave it suggestion on the way that you're constructing the filtering function, but it seems fundamentally sound.

services/QuillLMS/app/controllers/concerns/diagnostic_reports.rb
happythenewsad: I got rid of the param, via slightly different route

emilia-friedberg (approved):

PR #11081, opened 2023-09-29
lms/remove-source-classroom-unit-id

anathomical:
## WHAT
Remove a column that we never ended up using
## WHY
We added this column when we thought it would be necessary for Admin Diagnostic reporting, but realized that it wouldn't we should clean it up.  (A query to the production database confirms that this column is `null` for all rows.)
## HOW
Add a migration to remove the column since it's not in use.

PR Checklist | Your Answer
------------ | -------------
Have you added and/or updated tests? |  No, we never implemented code to use this

### DIFF
--- a/services/QuillLMS/app/models/classroom_unit.rb
+++ b/services/QuillLMS/app/models/classroom_unit.rb
@@ -4,15 +4,14 @@
 #
 # Table name: classroom_units
 #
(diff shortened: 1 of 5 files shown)


dandrabik (approved):
:+1: 🔪

happythenewsad (approved):

PR #11089, opened 2023-10-02
lms/fix-team-image-size

anathomical:
## WHAT
Add a max-width to team-member divs
## WHY
We were relying on image size of 219x219px to set the size of these elements, but we're considering using 2x images for better display on hi-res displays where normal images can look fuzzy.  In order to support that, we'll need the CSS of the page itself to control the size of these divs.
## HOW
Just add a `max-width` rule to the `team-member` to explicitly control the width of team member divs instead of just relying on the defined size of the images.

NOTE: there's a more robust version of this approach using [Responsive Image approaches](https://developer.mozilla.org/en-US/docs/Learn/HTML/Multimedia_and_embedding/Responsive_images), but I think we should try to resolve the immediate issue quickly and leave ourselves time to have a longer discussion about what our broader goals are and which approach would serve them best.

### Screenshots
image
image

### Notion Card Links
https://www.notion.so/quill/New-photos-on-Team-Page-are-too-large-on-mobile-view-e48071888a2949579eb94cbb6d0a1b82

PR Checklist | Your Answer
------------ | -------------
Have you added and/or updated tests? |  No tests on pure style changes

### DIFF
--- a/services/QuillLMS/app/assets/stylesheets/pages/team.scss
+++ b/services/QuillLMS/app/assets/stylesheets/pages/team.scss
@@ -18,6 +18,7 @@
     flex-wrap: wrap;
   }
   .team-member {
+    max-width: 219px;
     margin-right: 24px;
     margin-bottom: 40px;
   }


emilia-friedberg (approved):

anathomical:
Closing this in favor of a different approach in #11093

PR #11103, opened 2023-10-06
Lms/admin snapshot front end retry

anathomical:
## WHAT
Add a bit of logic to retry Admin Snapshot data calls on the front-end if it takes more than 20 seconds for Pusher to trigger a data reload
## WHY
It looks like some amount of the time Pusher isn't properly triggering data reloads.  Even in these cases, it looks like the backend jobs are completing and the data is making it into cache, it's just not being reloaded so it appears as if the data never loads.  By using `setTimeout` we provide a fallback mechanism to load that data out of cache in these cases.
## HOW
- Add a new value to component to state to store a timeout ID
- Use a React effect to set a timeout on component initialization to `getData` 20 seconds after component initialization
- When processing data from API calls, if there is data, clear the `timeout` interval so that we don't trigger unnecessary API calls

### Notion Card Links
https://www.notion.so/quill/Add-a-front-end-retry-after-20s-for-Admin-Snapshot-queries-605032ace8d84a5fbc6fa6d592603277?pvs=4

PR Checklist | Your Answer
------------ | -------------
Have you added and/or updated tests? |  No test coverage on this behavior

### DIFF
--- a/services/QuillLMS/client/app/bundles/PremiumHub/components/usage_snapshots/snapshotCount.tsx
+++ b/services/QuillLMS/client/app/bundles/PremiumHub/components/usage_snapshots/snapshotCount.tsx
@@ -52,6 +52,10 @@ const SnapshotCount = ({ label, size, queryKey, searchCount, selectedGrades, sel
     getData()
   }, [searchCount])
 
+  React.useEffect(() => {
+    console.log(`retryTimeout set to ${retryTimeout}`)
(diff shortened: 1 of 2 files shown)


emilia-friedberg (approved):

eadams17 (approved):

dandrabik (approved):
:+1: Thanks for adding this.

PR #11106, opened 2023-10-06
Lms/add timing logging

anathomical:
## WHAT
- Add a utility to send ErrorNotifications to NewRelic and Sentry when a block of code takes too long to execute
- Add reporting of Admin Snapshot queries that take longer than 20 seconds
## WHY
We don't have a ton of good visibility into BigQuery performance.  This should provide us with a decent view of how often BigQuery is returning slowly and what parameters make it more likely for this to happen.
## HOW
- Create a new `ErrorNotifier.report_long_running` method to generalize slow-running code reporting
- Use this new method to wrap both `Count` and `TopX` workers' query logic to isolate BigQuery performance

### Notion Card Links
https://www.notion.so/quill/Add-back-end-logging-for-Admin-Snapshot-workers-when-queries-take-longer-than-20s-4994ef4de3e94606905589a5a889b255?pvs=4

PR Checklist | Your Answer
------------ | -------------


services/QuillLMS/app/workers/snapshots/cache_snapshot_count_worker.rb
```diff
@@ -4,9 +4,12 @@ module Snapshots
   class CacheSnapshotCountWorker
     include Sidekiq::Worker
 
+    class SnapshotCountTooSlow < StandardError; end
```
dandrabik: Let's rename this to include `Error` and remove redundant language from the parent class, e.g. the current name is `CacheSnapshotCountWorker::SnapshotCountTooSlow`. Something like `CacheSnapshotCountWorker::SlowQueryError` is more standard.

services/QuillLMS/app/workers/snapshots/cache_snapshot_count_worker.rb
```diff
@@ -59,23 +62,29 @@ def perform(cache_key, query, user_id, timeframe, school_ids, filters)
       timeframe_end = parse_datetime_string(timeframe['current_end'])
       filters_symbolized = filters.symbolize_keys
 
-      current_snapshot = QUERIES[query].run(**{
+      ErrorNotifier.report_long_running(SnapshotCountTooSlow.new("Snapshot Count query took more than #{TOO_SLOW_THRESHOLD}"), TOO_SLOW_THRESHOLD, {
```
dandrabik: This error with message could be a constant (and/or you could define`message` in the error class).

services/QuillLMS/app/lib/error_notifier.rb
```diff
@@ -6,4 +6,19 @@ def self.report(error, options = {})
     Sentry.capture_exception(error, extra: options)
     NewRelic::Agent.notice_error(error, options)
   end
+
+  def self.report_long_running(error, threshold, options = {})
+    start = current_time
+    result = yield
```
dandrabik: Do you need to return `result` from this method?
e.g. Could this be:
```ruby
start = current_time
yield
runtime = current_time - start
return if runtime < threshold.to_i

report(error, options.merge({time_to_execute: runtime}))

File: services/QuillLMS/app/lib/error_notifier.rb [hunk 4]
```diff
@@ -6,4 +6,19 @@ def self.report(error, options = {})
     Sentry.capture_exception(error, extra: options)
     NewRelic::Agent.notice_error(error, options)
   end
+
+  def self.report_long_running(error, threshold, options = {})
```
dandrabik: Take or leave since this is a tiny piece of code, but this might be cleaner as a separate class/service so you can setup the config and usage separately, e.g.

```ruby
long_process_notifier = LongProcessNotifier.new(...configs)

long_process_notifier.run do
  # code here
end

**Review by dandrabik — APPROVED** [review]
So the code looks reasonable to me. I left a few notes for your consideration to make this a little cleaner to use. Note, it's not a deal breaker, but this current setup does make some of the methods a little harder to parse since there's a bunch of config for the block, then a long block. Might be worth looking for ways to make this a little cleaner.

**Review by happythenewsad — APPROVED** [review]

File: services/QuillLMS/app/lib/error_notifier.rb (hunk 3 above)
Comment (anathomical) [reply services/QuillLMS/app/lib/error_notifier.rb]: The current code layout does require the return.  The caching takes place outside of the code that's being benchmarked, which means that the result of the benchmarking block needs to bubble up in `return` chains so that the originally calling function can put the results in the cache.  We could refactor this to do the cache writing inside the benchmarking block instead if you think not returning values from benchmarking is worth it.

File: services/QuillLMS/app/lib/long_process_notifier.rb [hunk 5]
```diff
@@ -0,0 +1,26 @@
+# frozen_string_literal: true
+
+class LongProcessNotifier < ApplicationService
```
dandrabik: I believe this should be in `app/services/` not `app/lib`

services/QuillLMS/app/workers/snapshots/cache_snapshot_top_x_worker.rb
```diff
@@ -41,13 +48,28 @@ def perform(cache_key, query, user_id, timeframe, school_ids, filters)
...
-      }.merge(filters_symbolized))
+      long_process_notifier = LongProcessNotifier.new(
+        SlowQueryError.new,
+        TOO_SLOW_THRESHOLD,
+        {
+          query: query,
+          timeframe_start: timeframe_start,
+          timeframe_end: timeframe_end,
+          school_ids: school_ids
+        }.merge(filters_symbolized)
```
dandrabik: Note that in Ruby 3.1, there's a fun new shortening syntax, you can do this here (this is equivalent code for the first 4 items, since it):

```ruby
{query:, timeframe_start:, timeframe_end:, school_ids:}.merge(filters_symbolized)
# or since it's only for logging
{query:, timeframe_start:, timeframe_end:, school_ids:, filters:}

File: services/QuillLMS/app/lib/long_process_notifier.rb [hunk 7]
```diff
@@ -0,0 +1,26 @@
...
+  attr_reader :error, :threshold, :options
+
+  def initialize(error, threshold, options = {})
+    @error = error
+    @threshold = threshold
+    @options = options
+  end
+
+  def run
```
dandrabik: Note, since you are using `attr_reader`, you should use the non-instance variables in the class, e.g. `threshold` not `@threshold`

services/QuillLMS/app/lib/long_process_notifier.rb
```diff
@@ -0,0 +1,26 @@
...
+    start = self.class.current_time
+    result = yield
+    runtime = self.class.current_time - start
+    return result if runtime < @threshold.to_i
```
dandrabik: Take or leave, but since result is always returned, this might be more readable
```ruby
if runtime < @threshold.to_i
  ErrorNotifier.report(@error, @options.merge({time_to_execute: runtime}))
end

result

dandrabik (approved):
Small comments, but :+1:

PR #11121, opened 2023-10-11
Lms/hash filters for pusher comparison

anathomical:
## WHAT
Use a hash function for the Pusher payload sent to the front-end for comparison to figure out which Admin Snapshot to reload from cache
## WHY
In cases with a lot of filters selected, sending the full payload of all filters was causing Pusher to throw errors because the payloads were too big.
## HOW
Implement identical logic for hashing the values that were being compared between the front and back ends so that the backend hashes the payload to keep Pusher payloads very small and then the front-end can confirm if that hash matches the payload it expects as a signal to reload data from cache.

### Notion Card Links
https://www.notion.so/quill/Hash-Admin-Snapshot-filter-payloads-on-the-front-end-and-back-end-for-Pusher-comparison-instead-of-t-6b1ae512decd4d13bfd5425e24a33e4a?pvs=4
https://www.notion.so/quill/Pusher-Payload-Too-Large-Errors-4f82617b954f46999ca623acd6cb0984?pvs=4

PR Checklist | Your Answer
------------ | -------------


services/QuillLMS/app/workers/snapshots/cache_snapshot_count_worker.rb
```diff
@@ -35,13 +35,16 @@ def perform(cache_key, query, user_id, timeframe, school_ids, filters)
...
-        }.merge(filters)
-      )
+      filter_hash = Digest::MD5.hexdigest([
+        query,
+        timeframe['name'],
+        school_ids.join('-'),
+        filters['grades']&.join('-'),
+        filters['teacher_ids']&.join('-'),
+        filters['classroom_ids']&.join('-')
+      ].join('-'))
```
dandrabik: Small refactor, make an array, flatten in, then you can do the join once
```ruby
params = [query, timeframe['name'], school_ids, filters['grades'], filters['teacher_ids'], filters['classroom_ids']]

filter_hash = params
  .flatten
  .join('-')

**Review by emilia-friedberg — APPROVED** [review]

File: services/QuillLMS/app/workers/snapshots/cache_snapshot_top_x_worker.rb [hunk 2]
```diff
@@ -23,13 +23,16 @@ def perform(cache_key, query, user_id, timeframe, school_ids, filters)
...
 
-      SendPusherMessageWorker.perform_async(user_id, PUSHER_EVENT,
-        {
-          query: query,
-          timeframe: timeframe['name'],
-          school_ids: school_ids
-        }.merge(filters)
-      )
+      filter_hash = Digest::MD5.hexdigest([
+        query,
```
dandrabik, line 34: Same comment, here, we can do this `join('-')` once.

services/QuillLMS/client/app/bundles/PremiumHub/components/usage_snapshots/snapshotCount.tsx
```diff
@@ -1,5 +1,7 @@
 import * as React from 'react'
 
+import md5 from 'md5'
```
dandrabik: I believe that JS has a [native digest function](https://developer.mozilla.org/en-US/docs/Web/API/SubtleCrypto/digest). Let's use that if possible.

services/QuillLMS/client/app/bundles/PremiumHub/components/usage_snapshots/snapshotCount.tsx
```diff
@@ -111,16 +113,18 @@ const SnapshotCount = ({ label, size, queryKey, searchCount, selectedGrades, sel
...
+        selectedSchoolIds.join('-'),
+        selectedGrades?.join('-'),
+        selectedTeacherIds?.join('-'),
+        selectedClassroomIds?.join('-')
+      ].join('-')
 
-      if (queryKeysAreEqual && timeframesAreEqual && schoolIdsAreEqual && gradesAreEqual && teacherIdsAreEqual && classroomIdsAreEqual) {
-        getData()
-      }
+      const filterHash = md5(filterTarget)
```
dandrabik: Can we extract this out into a function to make it testable. I'd like you to add a test for this.

services/QuillLMS/client/app/bundles/PremiumHub/components/usage_snapshots/snapshotCount.tsx
```diff
@@ -111,16 +113,18 @@ const SnapshotCount = ({ label, size, queryKey, searchCount, selectedGrades, sel
...
-      const queryKeysAreEqual = message.query === queryKey
-      const timeframesAreEqual = message.timeframe === selectedTimeframe
-      const schoolIdsAreEqual = selectionsEqual(message.school_ids, selectedSchoolIds)
-      const teacherIdsAreEqual = selectionsEqual(message.teacher_ids, selectedTeacherIds)
-      const classroomIdsAreEqual = selectionsEqual(message.classroom_ids, selectedClassroomIds)
-      const gradesAreEqual =  selectionsEqual(message.grades, selectedGrades?.map(grade => String(grade))) || (!message.grades && !selectedGrades.length)
+      const filterTarget = [
+        queryKey,
+        selectedTimeframe,
+        selectedSchoolIds.join('-'),
```
dandrabik: Would have to do equivalent change on the frontend.

services/QuillLMS/app/workers/snapshots/cache_snapshot_count_worker.rb
```diff
@@ -35,13 +35,16 @@ def perform(cache_key, query, user_id, timeframe, school_ids, filters)
...
       Rails.cache.write(cache_key, payload, expires_in: cache_expiry)
 
-      SendPusherMessageWorker.perform_async(user_id, PUSHER_EVENT,
-        {
-          query: query,
-          timeframe: timeframe['name'],
-          school_ids: school_ids
-        }.merge(filters)
-      )
+      filter_hash = Digest::MD5.hexdigest([
```
dandrabik: Let's make this a separate service and test it explicitly with generic data, e.g.
```ruby
digest = HashDigest.run(array)

**Review by dandrabik — CHANGES_REQUESTED** [review]
Thanks for your work on this, I think this is getting close. There are a few small changes in my comments, but the larger item is adding better tests (and making the code testable):

1. Use the native JS digest library if possible.
2. Break out these digest calculations into services / functions and write explicit unit tests for them (both in Ruby and js ), e.g. confirm that HashService.run(['hello','world]) == "adsfasdasdfadsf"
3. Include the same sample tests in Ruby and Javascript, e.g. pass this array, get this explicit hash back. We need these outputs to match, and there's not a great way to link these tests between js and ruby, so let's just copy/paste the same example input/outputs in the tests.

File: services/QuillLMS/client/app/bundles/PremiumHub/components/usage_snapshots/snapshotCount.tsx (hunk 3 above)
Comment (anathomical) [reply services/QuillLMS/client/app/bundles/PremiumHub/components/usage_snapshots/snapshotCount.tsx]: Per our discussion on Slack, the difficulty of getting native stuff to work both in-browser and in test environment, I'm sticking with a separate md5 library for now.

File: services/QuillLMS/app/services/payload_hasher.rb [hunk 7]
```diff
@@ -0,0 +1,14 @@
+# frozen_string_literal: true
+
+class PayloadHasher < ApplicationService
+  attr_reader :payload
+
+  def initialize(payload)
```
dandrabik: One small naming nit, `payload` isn't a generic param, since you are doing `.join('-')` it needs to be an `array`, you might want to name it as such.

services/QuillLMS/client/app/bundles/PremiumHub/shared.tsx
```diff
@@ -44,3 +46,9 @@ export function mapItemsIfNotAll(selectedItems, allItems, mapKey = 'id') {
 
   return selectedItems.map(i => i[mapKey])
 }
+
+/* Should produce the same output as backend `app/services/payload_hasher.rb:run` */
+export function hashPayload(payload) {
+  const joinedPayload = payload.join('-')
```
dandrabik: same here, you might want a more specific name for `payload` or type it, since this is typescipt.

dandrabik (approved):
Small naming nit, but overall, looks much better. Thanks for doing this refactor and adding the tests. This feels much more solid, well done.

PR #11127, opened 2023-10-12
Add hint to vertex-ai-form

brendanshean:
## WHAT
Add some language to dropdown selector for Vertex AI model creation

## WHY
When creating endpoints in the Vertex AI UI, if a leading/trailing whitespace is added to the name, the Google Cloud UI does not display it so when comparing Endpoint and Model names it's difficult to reconcile why they are different.  For example, 
Screenshot 2023-10-12 at 2 57 53 PM)

But in the UI, the whitespace is invisible:
Screenshot 2023-10-12 at 2 58 25 PM

## HOW
Initially, I thought about using `.strip` on the input coming in from Google AI, but since our dropdown only lists model-endpoint pairs that are exactly the same, we can validate the names upstream immediately.  In the old automl platform there was only `model` to worry about but now we have model and endpoint to contend with.  This copy prevents gotchas downstream where we have to remember `.strip` anytime we're using `.display_name` on models or endpoints.

Curriculum has updated their endpoint creation documentation to include a note about leading/trailing whitespace with names.  

This hint is just another place where users will be reminded to update the gold standard name on Vertex AI.

### Screenshots
Screenshot 2023-10-12 at 2 50 48 PM

### Notion Card Links

PR Checklist | Your Answer
------------ | -------------
Have you added and/or updated tests? |  No.  Minor copy change.

### DIFF
--- a/services/QuillLMS/client/app/bundles/Staff/components/evidence/semanticRules/modelForm.tsx
+++ b/services/QuillLMS/client/app/bundles/Staff/components/evidence/semanticRules/modelForm.tsx
@@ -89,7 +89,7 @@ const ModelForm = ({ location, history, match }) => {
       <DropdownInput
         handleChange={handleSetName}
         isSearchable={true}
-        label="Name"
+        label="Name (note: if you're not finding a particular name, ensure that endpoint and model name are identical including leading/trailing whitespace)"
         options={nameOptions}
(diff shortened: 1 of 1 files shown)


dandrabik (approved):
Seems like a good stop gap that should help the team. :+1:

anathomical (approved):

PR #11132, opened 2023-10-13
Fix set impact metrics worker

happythenewsad:
## WHAT
The SetImpactMetricsWorker has been failing for three types of reasons:

1. Two queries used an outdated `initialize` method, leading to unhandled exceptions
2. Due to a change in the structure of Ruby BigQuery responses, two queries were retrieving hash values by string key instead of symbol key, resulting in nil values
3. One query's size grew to exceed BigQuery's query size limit 

This PR addresses all three issue groups, so that the worker completes without errors. 

## WHY
We want this worker to be healthy. It drives the topline usage metrics displayed on the Quill homepage.

## HOW
1. Update query `initialize` methods
2. update hash key types
3. refactor a query to avoid the generation of very large query strings 

### Screenshots

### Notion Card Links
https://www.notion.so/quill/Investigate-topline-homepage-metrics-187f08569ad74bb798bde21f35574d3f

PR Checklist | Your Answer
------------ | -------------


services/QuillLMS/app/queries/impact_metrics/schools_containing_certain_teachers_query.rb
```diff
@@ -29,7 +22,15 @@ def from_and_join_clauses
 
     def where_clause
       <<-SQL
-        WHERE schools_users.user_id IN (#{teacher_ids.join(', ')})
```
happythenewsad, line 32: This was the line that resulted in a query that was too large for BigQuery to accept

services/QuillLMS/app/workers/set_impact_metrics_worker.rb
```diff
@@ -7,20 +7,19 @@ class SetImpactMetricsWorker
   SENTENCES_PER_ACTIVITY_SESSION = 10
 
   def perform
-    finished_activity_sessions_count = ImpactMetrics::ActivitiesAllTimeQuery.run[0]["count"]
-    active_students_count = ImpactMetrics::ActiveStudentsAllTimeQuery.run[0]["count"]
+    finished_activity_sessions_count = ImpactMetrics::ActivitiesAllTimeQuery.run[0][:count]
+    active_students_count = ImpactMetrics::ActiveStudentsAllTimeQuery.run[0][:count]
 
     teachers = ImpactMetrics::ActiveTeachersAllTimeQuery.run
```
dandrabik: Is this query used anywhere else? If not, can we change it to a `count` query instead of returning all of the records and then counting them?

brendanshean (approved):
Nice work on restructuring the query and fixing the other items.

services/QuillLMS/app/workers/set_impact_metrics_worker.rb
```diff
@@ -7,20 +7,19 @@ class SetImpactMetricsWorker
...
-    teacher_ids = teachers.to_a.map {|teacher| teacher["id"]}
 
-    schools = ImpactMetrics::SchoolsContainingCertainTeachersQuery.run(teacher_ids: teacher_ids)
-    low_income_schools = schools.select { |school| (school["free_lunches"] || 0) > FREE_LUNCH_MINIMUM}
+    schools = ImpactMetrics::SchoolsContainingCertainTeachersQuery.run
```
dandrabik: Probably out of scope for this fix, but we could probably do the same thing here, i.e. make this a count query with an option for free lunch minimum.

dandrabik (approved):
Thanks for fixing this up. A few comments about potential performance improvements that are take or leave.

services/QuillLMS/app/queries/impact_metrics/schools_containing_certain_teachers_query.rb
```diff
@@ -2,13 +2,6 @@
 
 module ImpactMetrics
   class SchoolsContainingCertainTeachersQuery < ::QuillBigQuery::Query
```
anathomical: Not a high priority, but should we consider renaming this now that the logic for "certain teachers" is baked into the query itself instead of being passed in?  `SchoolsWithMinimumActivitySessions` or something?

anathomical (approved):
Nice clean-up, especially considering how many different axes you had to fix this on.

PR #11137, opened 2023-10-16
add shogun to spellcheck, relax bad words list to allow kirishitan

happythenewsad:
## WHAT / HOW
- Add permutations of the word 'shogun' to exceptions list, as bing was tripping on them
- relax the "*shit*" bad words entry, to allow Kirishitan. Note: we could add infrastructure to handle prompt-wise bad word exceptions, but this would add complexity to our system and we should first discuss if the juice is worth the squeeze.

### Screenshots

### Notion Card Links
https://www.notion.so/quill/Add-words-to-spelling-exceptions-profanity-exceptions-and-opinion-algorithm-91c436efe2614cc6a2416a8b7fe4cc3b

PR Checklist | Your Answer
------------ | -------------
Have you added and/or updated tests? |  manual - config only

### DIFF
--- a/services/QuillLMS/engines/evidence/lib/evidence/spelling_check.rb
+++ b/services/QuillLMS/engines/evidence/lib/evidence/spelling_check.rb
@@ -26,6 +26,8 @@ module Evidence
       'then',
       'sanchez',
       'sánchez',
+      "shoguns'",
+      "shogun's",
       'kanaka',
       'kānaka',
       'worldwatch',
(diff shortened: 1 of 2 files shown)


services/QuillLMS/engines/evidence/lib/evidence/spelling_check.rb
```diff
@@ -89,7 +91,7 @@ def error
     private def bing_response
       @response ||= HTTParty.get(BING_API_URL.to_s,
         headers: {
-          "Ocp-Apim-Subscription-Key": ENV['OCP-APIM-SUBSCRIPTION-KEY']
+          "Ocp-Apim-Subscription-Key": ENV['BING_SPELL_KEY']
```
happythenewsad, line 94: I changed the env var since env vars with hyphens are silently ignored in our local environments. I've updated staging and prod with this new env var.

brendanshean: That's good to know.  Thanks for figuring that out.

brendanshean (approved):

PR #11138, opened 2023-10-16
Lms/split snapshot count timeframes

anathomical:
## WHAT
When fetching the current and previous counts for a given metric in Admin Snapshots, fetch the two values in parallel on the front-end instead of using a single endpoint/worker combo that we were using before which made two BigQuery queries in sequence.
## WHY
Parallelizing this data lookup should result in a reduction in the average time it takes before a user sees data on this page due to the cases where current data comes back alone in the new system faster than current+previous data could come back in the old system.
## HOW
- Add a new controller action for `previous_count` so that the front-end can fetch that data separately
- Create a new worker for processing `previous_count`, but it does exactly the same thing as `count` except it sends a different Pusher event name when it completes
- Modify the controller helpers so that when processing data for `previous_count` the previous period of the timeframe is passed to the worker
- Update count worker to only take a timeframe_start and timeframe_end for timeframe data rather than the old method of getting previous and current timeframe bounds (we now handle that split in the controller)
- Short-circuit the controller action so that if there is no valid timeframe data we don't bother trying to calculate anything (currently intended for use with `previous_count` when the timeframe is "all-time")
- Update the front-end to make two different API calls on component load (along with two different retry timers for those two loads)
- Register two different Pusher listeners so that we can handle callbacks for both current and previous count
- Move a bunch of the calculation logic that is intended to handle API data into an effect so that we can consolidate that code into one place despite having two different API calls that could come back in either order

### Notion Card Links
https://www.notion.so/quill/Split-Admin-Snapshot-Count-queries-into-two-different-API-Worker-Cache-flows-one-for-Current-Timefr-3796837d69c84170a4e110e3cfc48be9?pvs=4

PR Checklist | Your Answer
------------ | -------------

### DIFF
--- a/services/QuillLMS/app/controllers/snapshots_controller.rb
+++ b/services/QuillLMS/app/controllers/snapshots_controller.rb
@@ -22,16 +22,21 @@ class SnapshotsController < ApplicationController
 
   WORKERS_FOR_ACTIONS = {
     "count" => Snapshots::CacheSnapshotCountWorker,
+    "previous_count" => Snapshots::CacheSnapshotPreviousCountWorker,
     "top_x" => Snapshots::CacheSnapshotTopXWorker,
     "data_export" => Snapshots::CachePremiumReportsWorker
   }
 
-  before_action :set_query, only: [:count, :top_x, :data_export]
-  before_action :validate_request, only: [:count, :top_x, :data_export]
-  before_action :authorize_request, only: [:count, :top_x, :data_export]
+  before_action :set_query, only: [:count, :previous_count, :top_x, :data_export]
+  before_action :validate_request, only: [:count, :previous_count, :top_x, :data_export]
+  before_action :authorize_request, only: [:count, :previous_count, :top_x, :data_export]
 
   def count
     render json: retrieve_cache_or_enqueue_worker(WORKERS_FOR_ACTIONS[action_name])
+  end
+
+  def previous_count
+    render json: retrieve_cache_or_enqueue_worker(WORKERS_FOR_ACTIONS[action_name], true)
   end
 
   def top_x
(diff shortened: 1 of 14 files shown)


emilia-friedberg (approved):
LGTM. Thanks for your work on this.

eadams17 (approved):
LGTM!

dandrabik (changes requested):
Thanks for putting this together, glad we were able to get this in place quickly. The previous and current controller endpoints, workers, and requests are identical, so I think there's a way to DRY this up by having one endpoint, worker, and request and passing one additional param, e.g. `previous: true/false` or `current: true/false` or whatever param makes you feel is clearest. That would make long-term maintenance easier and prevent these paths from drifting. I think it'll be a pretty small refactor.

services/QuillLMS/app/controllers/snapshots_controller.rb
```diff
@@ -134,10 +134,18 @@ def options
 
   private def retrieve_cache_or_enqueue_worker(worker)
 
-    previous_start, previous_end, current_start, current_end = Snapshots::Timeframes.calculate_timeframes(snapshot_params[:timeframe],
+    previous_start, previous_end, timeframe_start, timeframe_end = Snapshots::Timeframes.calculate_timeframes(snapshot_params[:timeframe],
```
dandrabik: Would this be cleaner if you passed `previous_timeframe` to the `calculate_timeframes` method and returned the correct start/end from there, versus swapping below. This current setup does 4 calculations for timeframes, then always throws away 2.

Looking at the calculate_timeframe code, you could use the `previous_timeframe` boolean to determine whether to run/return `[:previous_start, :previous_end]` or `[:current_start, :current_end] `

dandrabik (approved):
One small refactor suggestion, but otherwise approved. Thanks for pushing this along. Also, Let's make sure to do some testing on staging and prod to make sure this is working properly, in case the refactor caused any bugs.

PR #11139, opened 2023-10-16
AGDR/ skill section frontend

eadams17:
## WHAT
add the frontend UI for the Skill tab of the Admin Growth Diagnostic Report

## WHY
this is part of the larger Admin Growth Diagnostic Report project

## HOW
add the new component `SkillSection` and do some minor refactoring

### Screenshots
Screen Shot 2023-10-16 at 5 16 23 PM

### Notion Card Links
https://www.notion.so/quill/New-Project-Card-Admin-Diagnostic-Reports-Spec-For-Three-Reports-15ace9cefb7c4bf7b5475b041fff56d9?d=d6bcd80e0df94fe5902f127b4f789bae#f1801ca73f2d4b98825d7f42e30797ef

PR Checklist | Your Answer
------------ | -------------


services/QuillLMS/client/app/bundles/PremiumHub/shared.tsx
```diff
@@ -13,6 +13,31 @@ export const PENDING = 'Pending'
...
+  { label: 'ELL Starter Baseline Diagnostic (Pre)', value: 'ELL Starter Baseline Diagnostic (Pre)' },
+  { label: 'ELL Starter Growth Diagnostic (Post)', value: 'ELL Starter Growth Diagnostic (Post)' },
+  { label: 'ELL Intermediate Baseline Diagnostic (Pre)', value: 'ELL Intermediate Baseline Diagnostic (Pre)' },
+  { label: 'ELL Intermediate Growth Diagnostic (Post)', value: 'ELL Intermediate Growth Diagnostic (Post)' },
+  { label: 'ELL Advanced Baseline Diagnostic (Pre)', value: 'ELL Advanced Baseline Diagnostic (Pre)' },
+  { label: 'ELL Advanced Growth Diagnostic (Post)', value: 'ELL Advanced Growth Diagnostic (Post)' },
+  { label: 'AP Writing Skills Survey', value: 'AP Writing Skills Survey' },
+  { label: 'Pre-AP Writing Skills Survey 1', value: 'Pre-AP Writing Skills Survey 1' },
+  { label: 'Pre-AP Writing Skills Survey 2', value: 'Pre-AP Writing Skills Survey 2' },
+  { label: 'SpringBoard Writing Skills Survey', value: 'SpringBoard Writing Skills Survey' }
```
anathomical, line 40: I wonder if these should have the value of the `activity_id` for these Diagnostics.  And that makes me wonder if we already have that mapped as a constant somewhere in the code.  It seems like something that comes up enough that we probably do?  (It may be worth doing this as-written for now and then doing an update when we have a better idea of what the back-end API is going to need.)

services/QuillLMS/client/app/bundles/PremiumHub/containers/__tests__/skillSection.test.tsx
```diff
@@ -0,0 +1,56 @@
...
+  openMobileFilterMenu: jest.fn()
+}
+
+describe('SkillSection', () => {
+  describe('loading state', () => {
+    test('it should render loading spinner', () => {
+      const { asFragment } = render(<SkillSection {...props} />);
+      expect(asFragment()).toMatchSnapshot();
+      const loadingSpinner = screen.getByRole('img')
+      expect(loadingSpinner.getAttribute('class')).toEqual('spinner')
```
emilia-friedberg: This was just added, but as of [this change](https://github.com/empirical-org/Empirical-Core/blob/cea38a0bbad6826da3ddfcadb490ebb2434c381b/services/QuillLMS/client/app/bundles/Shared/components/shared/spinner.tsx), it's better to `queryByAltText` rather than look for attributes like class because those aren't visible to the user. So in this instance you would do something like: `expect(screen.getByAltText(/loading spinner/i)).toBeInTheDocument().`

services/QuillLMS/client/app/bundles/PremiumHub/shared.tsx
```diff
@@ -13,6 +13,31 @@ export const PENDING = 'Pending'
...
+]
```
emilia-friedberg, line 41: Since the label and value are identical in every case, it might be cleanest to do something like this:

```export const diagnosticTypeDropdownOptions = ['Starter Baseline Diagnostic (Pre)', 'Starter Growth Diagnostic (Post)', ...].map(name => ({ label: name, value: name }))```

emilia-friedberg: You could also split the array into its own const of `diagnosticTypes` and then do the mapping in the `diagnosticTypeDropdownOptions` definition.

emilia-friedberg (approved):
Small test tweak and cleanup idea but overall LGTM!

anathomical (approved):

services/QuillLMS/client/app/bundles/PremiumHub/containers/__tests__/skillSection.test.tsx
eadams17: oh great, that's def cleaner-- thanks for flagging

services/QuillLMS/client/app/bundles/PremiumHub/shared.tsx
eadams17: I like this approach, but I think I'm going to leave this as is until we have the backend portion finished (I think Thomas may be right that the `value` values will probably need to be the diagnostic IDs)

services/QuillLMS/client/app/bundles/PremiumHub/shared.tsx
eadams17: Yeah, I think you're probably right. I'm going to leave this as-is though since it's more of a placeholder for right now, and I can tweak once the backend portion is ready. I do like the idea of standardizing this in the codebase because I think we do have some similar consts for the Diagnostic cards on the Assign page

PR #11143, opened 2023-10-17
Allow google users to reset their non-existent password

brendanshean:
## WHAT
Allow google users the ability to 'reset' their password.

## WHY
This will provide a mechanism for google users to switch to username/password authentication.
The process of 'creating-a-password' will also have the side effect of unlinking _teacher_ accounts from Google.

There's two roles to consider in relation to the login problem and whether or not to drop the google_id from the user:

1) Student
The frontend UI and background job sync rely on google_id less for the student role.  Also, depending on the teacher's permissions, the student's google_id might change from nil back to the original so I would argue just keep it for consistency.

2) Teacher
Initially, the scope was targeting students only, but apparently some teachers might accidentally be set as under 18 (e.g. see this [thread](https://support.google.com/a/thread/168978251?hl=en&msgid=169381975)), so we have to plan for the case where they will also be restricted.  In this case we _do_ want to unlink their google account when `creating-a-password`.   Various background jobs run for a google teacher that rely on the google_id as a proxy for access.  Keeping the google_id will result in these background job running unnecessarily causing errors that will pollute the sidekiq queues.  Also, the frontend UI for various pages will present buttons to the teacher that are incorrect (e.g. Import Google Classrooms students) when they actually don't have access to this.

While this does present an asymmetry with respect to when google_id is kept, the other three cases have downsides (i.e. incorrect frontend UI buttons and background job failures) that warrant this approach.

## HOW
Remove check for google_id in the PasswordResetController
Unlink google account before sending the password_reset email.

### ****ADDENDUM
Added code to handle expired google access that was triggering a confirm session loop.  Instead check for users who have `google_access_expired` and have not gone through the 'create-a-password' route.

### Screenshots
Screenshot 2023-10-19 at 8 57 07 AM

### Notion Card Links
https://www.notion.so/quill/Allow-users-connected-to-Google-to-sign-in-with-an-email-or-username-Part-1-1df0445245e045298a9abd986b8947b8?pvs=4

PR Checklist | Your Answer
------------ | -------------
Spec Review: Have you reviewed the spec and ensured this meets requirements and/or matches design mockups? |  YES

### DIFF
--- a/services/QuillLMS/app/models/user.rb
+++ b/services/QuillLMS/app/models/user.rb
@@ -868,6 +868,10 @@ class User < ApplicationRecord
     update!(clever_id: nil, google_id: nil, signed_up_with_google: false)
   end
 
+  def unlink_google_account!
+    update!(google_id: nil, signed_up_with_google: false)
+  end
+
   private def validate_flags
     invalid_flags = flags - VALID_FLAGS
 
--- a/services/QuillLMS/app/models/user.rb
+++ b/services/QuillLMS/app/models/user.rb
@@ -710,6 +710,10 @@ class User < ApplicationRecord
     google_id.present? && auth_credential.present? && auth_credential.google_access_expired?
   end
 
+  def google_access_expired_and_no_password?
+    google_access_expired? && password_digest.nil?
+  end
+
   # Note this is an incremented count, so could be off.
   def completed_activity_count
     user_activity_classifications.sum(:count)
--- a/services/QuillLMS/app/assets/stylesheets/shared/session.scss
+++ b/services/QuillLMS/app/assets/stylesheets/shared/session.scss
@@ -87,7 +87,7 @@
         margin: auto;
         justify-content: center;
         border-radius: 4px;
-        box-shadow: 0 0 0 1px #bdbdbd;
+        box-shadow: 0 0 0 1px $quill-grey-20;
         background-color: $quill-white;
         font-size: 15px;
         font-weight: 600 !important;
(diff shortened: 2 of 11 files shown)


emilia-friedberg (approved):

anathomical (approved):

services/QuillLMS/app/models/user.rb
```diff
@@ -868,6 +868,10 @@ def unlink_clever_and_google_accounts!
     update!(clever_id: nil, google_id: nil, signed_up_with_google: false)
   end
 
+  def unlink_google_account!
+    update!(google_id: nil, signed_up_with_google: false)
```
dandrabik, line 876: Out of scope for this, but I think at some point, we should clean up / remove the "signed_up_with_google" attribute and just rely on `google_id`. Seeing code like this really shows that it's redundant (I agree with setting it to `false` here because there's a lot of logic tied to it, but if we are not keeping it as `true` here, I'm not sure there's a scenario in which it stores different info than `google_id`.

dandrabik (approved):
Looks good to me. Thanks for the explanation on the PR description about the different states was very helpful, since I would have had questions about it when looking at the code. If Peter Sharkey approves unlinking teachers and not students, that seems reasonable to me.

services/QuillLMS/app/models/user.rb
brendanshean: That's a good point.  AFAICT it's a redundant attribute that was used for something in the past. 

I will add it to my growing list of things to clean up at some point.

emilia-friedberg (approved):

anathomical (approved):

dandrabik (approved):
:+1: LGTM! Thanks for moving this along

PR #11162, opened 2023-10-23
lms/fix-argument-count-error

anathomical:
## WHAT
Update data-export worker signature to match new call signature
## WHY
We updated the controller to always call jobs with 7 params even though only one of the worker classes being used here needs all 7, which is awkward.  I think we should almost certainly refactor this in a cleaner way rather than have intentionally unused params in method signatures, but I wanted to get this out quickly rather than let the bug linger for a week.

What we did was add an additional param to help the `Count` worker handle both "current" and "previous" timeframes (which currently need different Pusher events to be sent).  Because we've generalized the calling of these workers, I decided to update the `TopX` worker to also take the additional param even though it never used it.  This was awkward, but felt like an okay compromise.  I missed that `DataExport` worker is hooked into the same call stack, so it also needs this value.  Having unused params in two different classes makes that compromise a lot worse, and it definitely needs some kind of refactor.  However, given that I'm off this week, and that I can get a quick-and-dirty fix out the door before fully checking out, I figured it was worth doing this and following up next week.

NOTE: It may be worth holding off on merging this version of the fix and hold out for something cleaner and closer to what we want the end state to be given that the page that uses this worker is not public (it's another one of those that we've put into production, but without a public link to the page).
## HOW
Add an unused param to the method signature for the worker to match the new expected call

### Notion Card Links
https://www.notion.so/quill/Sentry-Error-Sidekiq-Snapshots-CachePremiumReports-816ca6db3c8e46129aab1d677b1f08b6

PR Checklist | Your Answer
------------ | -------------

### DIFF
--- a/services/QuillLMS/app/workers/snapshots/cache_premium_reports_worker.rb
+++ b/services/QuillLMS/app/workers/snapshots/cache_premium_reports_worker.rb
@@ -10,7 +10,7 @@ module Snapshots
       'data-export' => Snapshots::DataExportQuery
     }
 
-    def perform(cache_key, query, user_id, timeframe, school_ids, filters)
+    def perform(cache_key, query, user_id, timeframe, school_ids, filters, previous_timeframe)
       payload = generate_payload(query, timeframe, school_ids, filters)
       Rails.cache.write(cache_key, payload.to_a, expires_in: cache_expiry)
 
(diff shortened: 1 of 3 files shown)


dandrabik (approved):
:+1: Better to get this out and then refactor if needed.

PR #11171, opened 2023-10-24
save and apply saved admin report filter selections 

emilia-friedberg:
## WHAT
Save admin filter selections when they're applied and then fetch them to use on page load.

## WHY
So that we can persist users' last selections per report without them having to do it manually each time.

## HOW
Add a controller to get and create/update `admin_report_filter_selections`, then update the frontend to use the new endpoints both before the filters are loaded and when new ones are applied. Also made a more minor tweak to the `options` endpoint to get all classrooms, teachers, and schools on first load, since we can no longer rely on all options being selected in the initial request.

### Screenshots

### Notion Card Links
https://www.notion.so/quill/Save-a-user-s-filter-selections-on-the-new-admin-reports-6012eccc6d1a47ed9f42d632cce33299?pvs=4

PR Checklist | Your Answer
------------ | -------------


eadams17 (approved):
Looks great to me! Thanks for putting this together

services/QuillLMS/app/controllers/snapshots_controller.rb
```diff
@@ -83,6 +86,24 @@ def options
     teachers
   end
 
+  private def all_sorted_teacher_options
```
brendanshean, line 125: Since this method his doing some heavy querying could pull this into a scope in user.rb:
```
# in user.rb
scope :teachers_in_schools, ->(school_ids) {
  distinct
    .joins(:schools_users)
    .left_outer_joins(:classrooms_teachers)
    .joins("LEFT OUTER JOIN classrooms ON classrooms_teachers.classroom_id = classrooms.id")
    .where(schools_users: { school_id: school_ids })
}
```

the method becomes:

```
private def all_sorted_teacher_options
  User
    .teachers_in_schools(school_options.pluck(:id))
    .sort_by(&:last_name)
end
```

services/QuillLMS/app/controllers/snapshots_controller.rb
```diff
@@ -192,8 +213,17 @@ def options
...
+      :report,
+      school_ids: [],
       grades: [],
-      teacher_ids: [])
+      teacher_ids: []
+    )
+  end
+
+  private def initial_load?
+    option_params[:is_initial_load] == "true" || option_params[:is_initial_load] == true
```
brendanshean: `option_params[:is_initial_load].in?([true, "true"])`

services/QuillLMS/spec/controllers/admin_report_filter_selections_controller_spec.rb
```diff
@@ -0,0 +1,75 @@
+# frozen_string_literal: true
+
+require 'rails_helper'
+
+describe AdminReportFilterSelectionsController, type: :controller do
+  let(:user) { create(:user) }
+  let(:valid_attributes) { { report: AdminReportFilterSelection::REPORTS.sample, user_id: user.id } }
```
brendanshean: `.sample` in specs can lead to some tricky intermittent spec failures.  If it's important to check test all values, I would just iterate over all of them.  If it's really just an issue of testing integration, then I would hardcode one value.

services/QuillLMS/spec/controllers/admin_report_filter_selections_controller_spec.rb
```diff
@@ -0,0 +1,75 @@
...
+  describe "POST #create_or_update" do
+    context "with valid params" do
+      it "creates a new AdminReportFilterSelection" do
+        expect {
+          post :create_or_update, params: { admin_report_filter_selection: valid_attributes }
+        }.to change(AdminReportFilterSelection, :count).by(1)
+      end
+
+      it "updates the filter selections" do
+        report = AdminReportFilterSelection::REPORTS.sample
```
brendanshean: Ditto earlier comment about `.sample`.

services/QuillLMS/app/controllers/snapshots_controller.rb
```diff
@@ -48,7 +48,10 @@ def options
       schools: format_option_list(school_options),
       grades: GRADE_OPTIONS,
       teachers: format_option_list(sorted_teacher_options),
-      classrooms: format_option_list(classroom_options)
+      classrooms: format_option_list(classroom_options),
+      all_classrooms: initial_load? ? format_option_list(all_classroom_options) : nil,
```
brendanshean: You could consider pulling initial_load? options into its own method:

```

def build_options_hash
  {
    timeframes: Snapshots::Timeframes.frontend_options,
    schools: format_option_list(school_options),
    grades: GRADE_OPTIONS,
    teachers: format_option_list(sorted_teacher_options),
    classrooms: format_option_list(classroom_options)
  }.merge(initial_load_options)
end

def initial_load_options
  return {} unless initial_load?

  {
    all_classrooms: format_option_list(all_classroom_options),
    all_teachers: format_option_list(all_sorted_teacher_options),
    all_schools: format_option_list(school_options)
  }
end
```

brendanshean (approved):
I left some refactoring comments, but overall looks good.  Nice job pulling the backend-frontend together here.

services/QuillLMS/client/app/bundles/PremiumHub/containers/PremiumFilterableReportsContainer.tsx
```diff
@@ -136,23 +151,75 @@ export const PremiumFilterableReportsContainer = ({ accessType, adminInfo, }) =>
     return timeframes?.find(timeframe => timeframe.default) || null
   }
 
+  function reportPath() {
+    return location.pathname.slice(location.pathname.lastIndexOf("/") + 1, location.pathname.length)
```
anathomical, line 155: I generally feel like reading paths directly is a weird pattern, but in this particular case I think it's quite a clever way of letting this component store its state any time it's used without having to pass extra attributes into it.  👍

services/QuillLMS/client/app/bundles/PremiumHub/containers/PremiumFilterableReportsContainer.tsx
```diff
@@ -136,23 +151,75 @@ export const PremiumFilterableReportsContainer = ({ accessType, adminInfo, }) =>
...
+  }
+
+  function setSelectedAndLastSubmitted(grades, schools, teachers, classrooms, timeframe) {
```
anathomical, line 158: Good call moving this into its own function.

services/QuillLMS/config/routes.rb
```diff
@@ -736,6 +736,13 @@
     end
   end
 
+  resources :admin_report_filter_selections, only: [] do
+    collection do
+      post '/show', to: 'admin_report_filter_selections#show'
```
anathomical, line 742: This almost certainly doesn't matter, but it feels like maybe this should be a `GET` call instead of `POST` since it doesn't change data and doesn't expect a large set of param data.  But keeping it POST for consistency would be fine.  If you do decide to change this, I think we can also remove this line and just add `:show` to the currently-empty `only: []` array since `show` is a "standard" action in Rails.

services/QuillLMS/spec/controllers/admin_report_filter_selections_controller_spec.rb
```diff
@@ -0,0 +1,75 @@
...
+          'timeframe' => timeframe,
+          'schools' => [school]
+        }
+        updated_attributes = { report: report, filter_selections: updated_filter_selections }
+
+        post :create_or_update, params: { admin_report_filter_selection: updated_attributes }
+        admin_report_filter_selection.reload
+
+        expect(admin_report_filter_selection.filter_selections['timeframe']['value']).to eq(timeframe['value'])
+        expect(admin_report_filter_selection.filter_selections['schools'][0]['name']).to eq(school['name'])
```
anathomical: Whenever I run a `.reload` in the middle of a test, I consider wrapping the test in an `expect ... to change` block.  In this case, I think that might look like:
```ruby
expect do
  post :create_or_update, params: { admin_report_filter_selection: updated_attributes }
end.to change(admin_report_filter_selection.filter_selections['timeframe']['value']).from('last_week').to(timeframe['value'])
  .and change(admin_report_filter_selection.filter_selections['schools'][0]['name']).from(nil).to(school['name'])
```
If you decide to do this, it's probably also be worth putting `'last_week'` into a variable so you don't have the string literal in two places.

services/QuillLMS/spec/controllers/admin_report_filter_selections_controller_spec.rb
```diff
@@ -0,0 +1,75 @@
+# frozen_string_literal: true
+
+require 'rails_helper'
+
+describe AdminReportFilterSelectionsController, type: :controller do
+  let(:user) { create(:user) }
+  let(:valid_attributes) { { report: AdminReportFilterSelection::DATA_EXPORT, user_id: user.id } }
+  let(:invalid_attributes) { { report: 'invalid_report', user_id: user.id } }
```
anathomical: It feels a little odd to include `user_id` in these attribute variables given that you use them as part of request payloads (and `user_id` is not a valid param for the controllers.  I'm pretty sure you're doing this so that you can use the same value for both the request and for the `create` calls for model creation, so this isn't a huge deal, but I think I'd be inclined to break "params for HTTP requests" and "data for creating backend models" into two separate values if only to avoid confusion if someone reads this file and thinks they can pass arbitrary `user_id` values in.  Maybe something like:
```ruby
let(:valid_attributes) { { report: AdminReportFilterSelection::DATA_EXPORT, user_id: user.id } }
let(:invalid_attributes) { { report: 'invalid_report' } }
let(:valid_model_attributes { valid_attributes.merge({ user_id: user.id }) }
```

services/QuillLMS/app/controllers/snapshots_controller.rb
```diff
@@ -43,12 +43,26 @@ def data_export
...
+  end
+
+  private def build_options_hash
+    {
       timeframes: Snapshots::Timeframes.frontend_options,
       schools: format_option_list(school_options),
       grades: GRADE_OPTIONS,
       teachers: format_option_list(sorted_teacher_options),
       classrooms: format_option_list(classroom_options)
+    }.merge(initial_load_options)
```
anathomical, line 85: I quite like the way that you've implemented this `initial_load?` override to keep the code simple and readable.

anathomical (approved):
A couple of notes regarding test structure, but this code looks good to me.

PR #11173, opened 2023-10-24
Add segment event to track google users that set a password

brendanshean:
## WHAT
Add a segment event that tracks when a google user sets a password.

## WHY
This will give us an indirect metric for seeing how many users have been hit by the Google Workspace for Education login change in authorization.

## HOW
Add a ~~before_save~~ before_update callback to on the User model that checks to see if a google user has set a password from nil to something non-nil.

Addendum: I changed the callback to `before_update` since we don't want this running on creation, only on update.  A user would only have a google_id present after creation if the oauth2 flow had no authorization errors.

### Screenshots

### Notion Card Links
https://www.notion.so/quill/Create-new-Segment-event-for-Google-Users-creating-a-password-f1a6a2a9771d4e699a0c6e8bbc23cb97?pvs=4

PR Checklist | Your Answer
------------ | -------------


dandrabik (approved):
:+1: Thanks for putting this together quickly!

PR #11183, opened 2023-10-26
Add Checklist for new District Admin attaching to schools

cissyxyu:
## WHAT
Add a checklist feature when adding a new district admin. From this checklist, Quill staff can select which schools the admin gets privileges for. 
Also, automatically add the admin to the first school selected with Premium access. This gives them a Premium subscription so they can access full Quill features.

## WHY
This allows Quill staff to fine tune admin access and turn it on and off for different schools throughout a district. It saves them a lot of time because now they have to manually remove certain schools access after creating an admin.

## HOW
Create a front end checklist that passes checked schools to the back end. In the back end, attach schools from the passed parameters rather than attaching all schools automatically.

### Screenshots
Screenshot 2023-10-26 at 5 21 32 PM

### Notion Card Links
[(Please provide links to any relevant Notion card(s) relevant to this PR.)](https://www.notion.so/quill/Product-Board-30f97e4eb01246dbb8264253fb385073?p=9e8319cfdc764f0cb81ebca73964497a&pm=c)

PR Checklist | Your Answer
------------ | -------------

### DIFF
--- a/services/QuillLMS/app/controllers/cms/district_admins_controller.rb
+++ b/services/QuillLMS/app/controllers/cms/district_admins_controller.rb
@@ -8,7 +8,7 @@ class Cms::DistrictAdminsController < Cms::CmsController
     user = User.find_by(email: params[:email])
     school_ids = params[:school_ids]
     if user && @district.district_admins.exists?(user_id: user.id)
-      render json: { message: t('district_admin.already_assigned', district_name: @district.name) }
+      render json: { message: t('district_admin.already_assigned', district_name: @district.name) }, status: 200
     elsif user
       create_district_admin_user_for_existing_user(user, school_ids)
       InternalTool::MadeDistrictAdminEmailWorker.perform_async(user.id, @district.id)
@@ -18,6 +18,8 @@ class Cms::DistrictAdminsController < Cms::CmsController
       InternalTool::DistrictAdminAccountCreatedEmailWorker.perform_async(user.id, @district.id)
       render json: { message: t('district_admin.new_account') }, status: 200
     end
+  rescue ActiveRecord::RecordInvalid => e
+    render json: { error: e.record.errors.messages }
   end
 
   def destroy
@@ -44,26 +46,20 @@ class Cms::DistrictAdminsController < Cms::CmsController
 
   private def create_district_admin_user_for_existing_user(user, school_ids)
     district_admin = @district.district_admins.build(user_id: user.id)
-    if district_admin.save!
-      district_admin.attach_schools(school_ids)
-      attach_as_teacher_to_first_premium_school(user)
-      admin_info = AdminInfo.find_or_create_by!(user: user)
-      admin_info.update(approver_role: User::STAFF, approval_status: AdminInfo::APPROVED)
-    else
-      render json: { error: district_admin.errors.messages }
-    end
+    district_admin.save!
+    district_admin.attach_schools(school_ids)
+    attach_as_teacher_to_first_premium_school(user)
+    admin_info = AdminInfo.find_or_create_by!(user: user)
+    admin_info.update(approver_role: User::STAFF, approval_status: AdminInfo::APPROVED)
   end
 
   private def create_new_account_for_district_admin_user
     user = User.new(user_params)
-    if user.save!
-      user.refresh_token!
-      ExpirePasswordTokenWorker.perform_in(30.days, user.id)
-      create_district_admin_user_for_existing_user(user, params[:school_ids])
-      user
-    else
-      render json: { error: user.errors.messages }
-    end
+    user.save!
+    user.refresh_token!
+    ExpirePasswordTokenWorker.perform_in(30.days, user.id)
+    create_district_admin_user_for_existing_user(user, params[:school_ids])
+    user
   end
 
   private def attach_as_teacher_to_first_premium_school(admin)
--- a/services/QuillLMS/app/controllers/cms/district_admins_controller.rb
+++ b/services/QuillLMS/app/controllers/cms/district_admins_controller.rb
@@ -54,6 +54,7 @@ class Cms::DistrictAdminsController < Cms::CmsController
 
     if district_admin.save!
       district_admin.attach_schools(school_ids)
+      attach_as_teacher_to_first_premium_school(user)
       admin_info = AdminInfo.find_or_create_by!(user: user)
       admin_info.update(approver_role: User::STAFF, approval_status: AdminInfo::APPROVED)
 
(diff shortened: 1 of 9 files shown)


services/QuillLMS/client/app/bundles/Staff/components/newAdminOrDistrictUser.tsx
```diff
@@ -53,7 +68,8 @@ const NewAdminOrDistrictUser = ({ type, returnUrl, schoolId, districtId }: NewAd
       const params = {
         first_name: firstName.trim(),
         last_name: lastName.trim(),
-        email: email.trim()
+        email: email.trim(),
+        school_ids: checklistSchools.filter((s) => s.checked).map((s) => s.id),
```
emilia-friedberg: NBD but for styling consistency, we usually don't wrap single parameters in parentheses (as you don't elsewhere on this page), so I would write these as `(s => s.checked)` and `(s => s.id)`.

emilia-friedberg (approved):

services/QuillLMS/client/app/bundles/Staff/components/newAdminOrDistrictUser.tsx
brendanshean: It looks like we could use an eslint [rule](https://eslint.org/docs/latest/rules/arrow-parens) for that, specifically the `as-needed` config

services/QuillLMS/app/controllers/cms/districts_controller.rb
```diff
@@ -84,6 +85,17 @@ def update
     @district = District.find(params[:id])
   end
 
+  private def schools_formatted_for_new_admin
+    @district.schools.map { |s|
```
brendanshean: The ruby convention is to use do..end for multiline blocks as well as full variable names instead of abbreviations:

```
@district.schools.map do |school|
  {
    id: school.id,
    name: school.name
    checked: school.subscription.present?,
    has_subscription: school.subscription.present?
  }
end
```

services/QuillLMS/spec/models/district_admin_spec.rb
```diff
@@ -21,12 +21,13 @@
   it { should belong_to(:district) }
   it { should belong_to(:user) }
 
-  it { is_expected.to callback(:attach_schools).after(:create) }
   it { is_expected.to callback(:detach_schools).after(:destroy) }
 
   let!(:user) { create(:user, email: 'test@quill.org') }
   let!(:district) { create(:district) }
   let!(:district_admin) { create(:district_admin, user: user, district: district) }
+  let!(:school1) { create(:school) }
```
brendanshean: Since the variables school1, school2 are only used in '#attach_schools' lets pet the declarations inside of the block for clarity.

services/QuillLMS/spec/controllers/cms/district_admins_controller_spec.rb
```diff
@@ -34,6 +35,23 @@
           expect(ActionMailer::Base.deliveries.last.to).to eq(['test@email.com'])
         end
       end
+
```
brendanshean, line 39: Let's extract a `let(:email) { 'test@email.com' }` at the beginning of `describe 'for a new user' do` since it occurs many places:

Screenshot 2023-10-26 at 9 50 57 AM

brendanshean (commented):
I'm leaving some simple comments to complete offline.

Let's plan to discuss refactoring `create_district_admin_user_for_existing_user(user, school_ids, new_user: false)` since it uses a boolean parameter that is hiding some complexity.

In any event this looks to be close to ready.  Nice work on working through some of the tricky logic with these attachments.

services/QuillLMS/app/controllers/cms/district_admins_controller.rb
brendanshean, line 1: You're getting closer here but there's still an issue with the handling of `save!` methods specifically,

`create_district_admin_user_for_existing_user` and `create_new_account_for_district_admin_user methods`.  When using save!, it will raise an exception if it fails, so the else branch with render will never actually be reached in those private methods. So it's not quite a double render problem, because those render calls are in places they will never be hit due to the use of `save!`. (If non-bang `save` were used than the render calls would get hit and you would have a double render issue)

I think it's instructive to try and figure out how to do that, but I'll leave a possible answer below if you get stuck.  In any event, I think this is a great example to try and reason control flow with `save/save!` and potential double render issues.  I remember the first time I hit one of these it was definitely confusing.


One solution is to remove the conditionals altogether in those methods, and than capture the error higher up in create, so something like:

```

 def create
    user = User.find_by(email: params[:email])
    school_ids = params[:school_ids]
    
    if user && @district.district_admins.exists?(user_id: user.id)
      render json: { message: t('district_admin.already_assigned', district_name: @district.name) }
    elsif user
      create_district_admin_user_for_existing_user(user, school_ids)
      InternalTool::MadeDistrictAdminEmailWorker.perform_async(user.id, @district.id)
      render json: { message: t('district_admin.existing_account') }, status: 200
    else
      user = create_new_account_for_district_admin_user
      InternalTool::DistrictAdminAccountCreatedEmailWorker.perform_async(user.id, @district.id)
      render json: { message: t('district_admin.new_account') }, status: 200
    end
  rescue ActiveRecord::RecordInvalid => e
    render json: { error: e.record.errors.messages }, status: 422
  end

private def create_district_admin_user_for_existing_user(user, school_ids)
    district_admin = @district.district_admins.build(user_id: user.id)
    district_admin.save!
    district_admin.attach_schools(school_ids)
    attach_as_teacher_to_first_premium_school(user)
    admin_info = AdminInfo.find_or_create_by!(user: user)
    admin_info.update(approver_role: User::STAFF, approval_status: AdminInfo::APPROVED)
  end

  private def create_new_account_for_district_admin_user
    user = User.new(user_params)
    user.save!
    user.refresh_token!
    ExpirePasswordTokenWorker.perform_in(30.days, user.id)
    create_district_admin_user_for_existing_user(user, params[:school_ids])
    user
  end
  ```
  
Overall, this create method is probably doing too much.  If we added anymore logic in the future I would probably break it up into different endpoints.

File: services/QuillLMS/app/controllers/cms/district_admins_controller.rb [hunk 5]
```diff
@@ -6,10 +6,17 @@ class Cms::DistrictAdminsController < Cms::CmsController
 
   def create
     user = User.find_by(email: params[:email])
-    if user
-      create_district_admin_user_for_existing_user(user)
+    school_ids = params[:school_ids]
+    if user && @district.district_admins.exists?(user_id: user.id)
+      render json: { message: t('district_admin.already_assigned', district_name: @district.name) }
```
brendanshean: For consistency does this need a status 200?

brendanshean (changes requested):
You're very close here.  One comment on cleaning up some exception handling

brendanshean (approved):
Nice work on finishing this up!

happythenewsad (approved):

PR #11189, opened 2023-10-30
Admin csv export: add spinner, snackbar, new copy

happythenewsad:
## WHAT
- add spinner when download button is clicked
- add snackbar when download button is clicked
- add final snackbar and email copy texts
- remove query truncation, which was put in place for testing

## WHY
To effect a smooth and informative UX for users. 

## HOW
- For both effects, hook into POST callback 

### Screenshots

### Notion Card Links
https://www.notion.so/quill/Product-Board-30f97e4eb01246dbb8264253fb385073?p=1039497a407e4e94b1a988530861d7c1&pm=s

PR Checklist | Your Answer
------------ | -------------

### DIFF
--- a/services/QuillLMS/client/app/bundles/PremiumHub/components/dataExportTableAndFields.tsx
+++ b/services/QuillLMS/client/app/bundles/PremiumHub/components/dataExportTableAndFields.tsx
@@ -161,6 +161,7 @@ export const DataExportTableAndFields = ({ queryKey, searchCount, selectedGrades
   }
 
   function createCsvReportDownload() {
+    const buttonDisableTime = 2000
     const requestParams = {
       query: 'create_csv_report_download',
       timeframe: selectedTimeframe,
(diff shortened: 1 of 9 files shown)


services/QuillLMS/client/app/bundles/PremiumHub/components/dataExportTableAndFields.tsx
```diff
@@ -166,7 +173,11 @@ export const DataExportTableAndFields = ({ queryKey, searchCount, selectedGrades
       grades: selectedGrades,
       headers_to_display: getHeaders().map(header => header.attribute)
     }
+    setDownloadButtonBusy(true)
+
     requestPost('/snapshots/create_csv_report_download', requestParams, (body) => {
+      setShowSnackbar(true)
+      setTimeout(() => {setDownloadButtonBusy(false)}, buttonDisableTime);
```
happythenewsad, line 180: I recall that previous React versions needed to explicitly clear timers. I notice in our code we don't do that. Perhaps the latest version of React handles timer cleanup, or it's just not relevant for short-lived components?

emilia-friedberg: Good question. React does not handle clearing timeouts on its own. My understanding is that clearing timeouts is important so that you don't have weird side effects that can occur if the timeout gets called after the component has unmounted. I generally only find myself using them in pretty high-level components (such that we are less likely to "unmount" rather than just navigate away, like this page), so for practical purposes it doesn't matter, but I think there's an argument that we should still be explicitly clearing the timeout just in case.

emilia-friedberg (approved):
LGTM!

PR #11196, opened 2023-10-31
lms/admin-snapshot-cache-with-filters

anathomical:
## WHAT
Update overnight cache worker to use stored filters

NOTE: This code is intended to supplement work done in #11171, and should be tested again in staging after that PR is merged.
## WHY
We want to pre-cache the first report that the user loads, and if they have stored filters those are the ones that will load when they navigate to the snapshot page.  So we should cache the data for that version of the report.
## HOW
In the code that constructs the payload to send to the cache workers, check to see if the admin in question has stored filters.  If they do, apply the data from those filters to the payloads sent to the workers so that the report matching those filters is the one cached.

### Notion Card Links
https://www.notion.so/quill/Cache-Admin-Snapshot-queries-overnight-for-every-admin-using-the-default-filters-da22a602494146628c8d00cefb81a29e?pvs=4

PR Checklist | Your Answer
------------ | -------------

### DIFF
--- a/services/QuillLMS/app/workers/cache_admin_snapshots_worker.rb
+++ b/services/QuillLMS/app/workers/cache_admin_snapshots_worker.rb
@@ -4,17 +4,17 @@ class CacheAdminSnapshotsWorker
   include Sidekiq::Worker
   sidekiq_options queue: SidekiqQueue::CRITICAL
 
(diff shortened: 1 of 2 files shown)


anathomical:
This project is on hold so I'm going to close this PR.

PR #11201, opened 2023-11-01
persist custom dates for saved filter selections

emilia-friedberg:
## WHAT
Make sure we're saving the custom start and end dates along with the timeframe when they've been selected.

## WHY
Otherwise the frontend is in a weird state because the timeframe itself is saved but it doesn't know what dates to use.

## HOW
Just add these values to the ones that get sent to the backend and then handle them if there are returned values.

### Screenshots

### Notion Card Links

PR Checklist | Your Answer
------------ | -------------
Have you deployed to Staging? | NO - testing locally with staging data


anathomical (approved):
Thanks for the quick fix.

PR #11204, opened 2023-11-02
Truncate long classroom names

brendanshean:
## WHAT
Fix an issue with teachers being unable to import classrooms with very long names (700+ characters)

## WHY
The classroom name field is a `string` type which in older versions of postgres is capped at length 255.  When a teacher attempts to import a classroom with a longer name, the import process fails due to this constraint.

## HOW
Truncate names any longer than 233 characters(255 characters minus a random string length that might be added for duplicate name resolving)

Rename `DuplicateNameResolver` to `ValidNameBuilder` since it incorporates truncation and duplicates resolving.

### Screenshots

### Notion Card Links
https://www.notion.so/quill/Teachers-at-District-Cannot-Import-Clever-Classes-Class-Name-Issue-ecce211fe7fe42a7a229c893354a116b?pvs=4

PR Checklist | Your Answer
------------ | -------------

### DIFF
--- a/services/QuillLMS/app/services/duplicate_name_resolver.rb
+++ b/services/QuillLMS/app/services/duplicate_name_resolver.rb
@@ -4,9 +4,12 @@ class DuplicateNameResolver < ApplicationService
   attr_reader :name, :existing_names, :max_before_randomized
 
   MAX_BEFORE_RANDOMIZED = 10
+  MAX_STRING_LENGTH = 255
+  RANDOM_STRING_LENGTH = SecureRandom.urlsafe_base64.length
+  TRUNCATE_LENGTH = MAX_STRING_LENGTH - RANDOM_STRING_LENGTH
 
-  def initialize(name, existing_names, max_before_randomized=MAX_BEFORE_RANDOMIZED)
-    @name = name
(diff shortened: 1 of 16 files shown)


services/QuillLMS/app/services/duplicate_name_resolver.rb
```diff
@@ -4,9 +4,12 @@ class DuplicateNameResolver < ApplicationService
   attr_reader :name, :existing_names, :max_before_randomized
 
   MAX_BEFORE_RANDOMIZED = 10
+  MAX_STRING_LENGTH = 255
+  RANDOM_STRING_LENGTH = SecureRandom.urlsafe_base64.length
+  TRUNCATE_LENGTH = MAX_STRING_LENGTH - RANDOM_STRING_LENGTH
```
anathomical: We might also want to account for the `-` that's inserted by `name_with_random_string`.  So this might need another `- 1`?

anathomical (approved):
I do like centering the max length logic into the duplicate name resolver itself.

anathomical (approved):
Looks good to me.  I like updating the name to communicate the slightly broader responsibilities.

PR #11209, opened 2023-11-02
Lms/fix admin diagnostic percentage bug

anathomical:
## WHAT
- Tweak the way we initialize Pusher listeners in the Count so that there is only one listener per widget
- Pass the `custom_timeframe` related keys that are sent to the controller through to the workers so that they can include the contents of those values in the hashed key to be sent back to the front-end via Pusher
- Add `queryKey` to the Pusher event names rather than using the same name for all queries
- Handle comparison of Pusher hashes to front-end hashes in React effect
- Tweak the retry logic for Snapshot Admin frontends so that they're triggered by requests for data rather than during component initialization
- Make updates to the DataExport front-end and back-end to bring it in line with Admin Snapshot logic
- - Special focus on using hashes in Pusher payloads
## WHY
- Having multiple listeners was combining with another issue to cause Pusher messages to cause multiple different filter configurations to request data from cache, resulting in the data flickering behind the scenes as different values rapidly loaded from cache
- This was the other half of the issue causing flickering.  We added additional information to the Pusher payload to include "custom" timeframe start and end data from the initial request.  The Pusher payload included information to try to avoid requesting data for "old" filter values that the front-end checked before requesting data from cache.  However, the only thing that was being done to differentiate timeframes was to check their names.  When using a "custom" timeframe, the timeframe name was always `custom`.  This meant that if you requested data for a custom timeframe, and then did so for another timeframe, you could initiate a race condition where either the first or second filter config could come back from Pusher, and the front-end couldn't tell the difference.
- This is a small performance tweak.  What we were doing is having every individual widget subscribe to all Pusher messages for each other widget of the same type.  This meant that widgets were processing each message from Pusher for widgets of their type and then throwing away the messages that were for the wrong report.  This update means that most Pusher messages that are processed by the front-end are relevant (rather than the old system where most were discarded).
- Comparing Pusher payload hashes to a hash of the current local filter state (to ensure that if you changed filters before all the data came back for your earlier filters, the front-end didn't load data for the previous filters when the Sidekiq job eventually completed) was sometimes failing to recognize good matches.  This appears to have been related to an encapsulation issue.  Moving the logic for the comparison into a React effect got us around encapsulation.
- Moving retry logic into the request code does two things: 1) simplifies the code used for retry setup (by not using an effect), and 2) it allows us to do multiple retries as if the retry call still doesn't get data it can set up another retry timer
- This code had drifted from the updates made for Admin Snapshot code.  Most of this wasn't a big deal, but failing to carry over the hashing of the Pusher payload would likely result in the same "Pusher payload too large" errors we were getting in Admin Snapshots that led to hashing in the first place.
## HOW
- Remove the `initializePusher` call from the `searchCount` effect in `snapsthotCount.ts`
- Modify `SnapshotController` so that it passes the custom timeframe params through to the worker, include those params in the payload hashed and sent through Pusher, update the front-end hash comparison code with the new params that are included in the hash
- Update both front-end and back-end Pusher logic to expect event names to include the query name
- Update `initializePusher` logic so that when Pusher messages are received they get pushed into state and hook the evaluation logic into a React effect that monitors that state
- Instead of calling `setTimeout` during the `searchCount` effect, do it when processing results from API calls such that when the backend doesn't have data in cache we establish a re-load timeout, and when there is data we clear the timeout
- Just generally update the code in the front-end and worker for Data Export to be in line with what we're doing in Count and Ranking

### Screenshots
(We expect an "optimal" run of the page to involve 90 requests: 19 current counts + 19 previous counts + 7 rankings, each of which gets called once to initialize workers and then a second time to load data from cache.  In production we're seeing a number of cases where it's much higher.)
image
(This is what we're consistently seeing in the new branch.  The extra 1 request is the new call to save the filter settings to the back-end.)
image

### Notion Card Links
https://www.notion.so/quill/Percentages-Jumping-on-Usage-Snapshot-dashboard-f2e3a9af6062473a8277f0dac98ffcf1?pvs=4
https://www.notion.so/quill/Usage-Snaphot-too-many-requests-for-count-c23482dc11aa4baabaad1f8619738209?pvs=4

PR Checklist | Your Answer
------------ | -------------

### DIFF
--- a/services/QuillLMS/client/app/bundles/PremiumHub/components/dataExportTableAndFields.tsx
+++ b/services/QuillLMS/client/app/bundles/PremiumHub/components/dataExportTableAndFields.tsx
@@ -57,6 +57,8 @@ export const DataExportTableAndFields = ({ queryKey, searchCount, selectedGrades
   const [showSnackbar, setShowSnackbar] = React.useState<boolean>(false);
   const [data, setData] = React.useState<any>(null);
   const [pusherMessage, setPusherMessage] = React.useState<any>(null)
+  const [customTimeframeStartString, setCustomTimeframeStartString] = React.useState(null)
+  const [customTimeframeEndString, setCustomTimeframeEndString] = React.useState(null)
 
   useSnackbarMonitor(showSnackbar, setShowSnackbar, defaultSnackbarTimeout)
 
@@ -136,6 +138,18 @@ export const DataExportTableAndFields = ({ queryKey, searchCount, selectedGrades
   }, [searchCount])
 
   React.useEffect(() => {
+    if (!customTimeframeStart) return
+
+    setCustomTimeframeStartString(customTimeframeStart.toISOString())
+  }, [customTimeframeStart])
+
+  React.useEffect(() => {
+    if (!customTimeframeEnd) return
+
+    setCustomTimeframeEndString(customTimeframeEnd.toISOString())
+  }, [customTimeframeEnd])
+
+  React.useEffect(() => {
     if (!pusherMessage) return
 
     if (filtersMatchHash(pusherMessage)) getData()
@@ -192,8 +206,8 @@ export const DataExportTableAndFields = ({ queryKey, searchCount, selectedGrades
     const filterTarget = [].concat(
       queryKey,
       selectedTimeframe,
-      customTimeframeStart?.toISOString().split('T',1)[0],
-      customTimeframeEnd?.toISOString().split('T',1)[0],
+      customTimeframeStartString,
+      customTimeframeEndString,
       selectedSchoolIds,
       selectedGrades,
       selectedTeacherIds,
--- a/services/QuillLMS/client/app/bundles/PremiumHub/components/dataExportTableAndFields.tsx
+++ b/services/QuillLMS/client/app/bundles/PremiumHub/components/dataExportTableAndFields.tsx
@@ -138,9 +138,7 @@ export const DataExportTableAndFields = ({ queryKey, searchCount, selectedGrades
   React.useEffect(() => {
     if (!pusherMessage) return
 
-    const { hash, timeframe, } = pusherMessage
-
-    if (filtersMatchHash(hash) && customTimeframeMatches(timeframe)) getData()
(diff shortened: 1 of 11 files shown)


services/QuillLMS/client/app/bundles/PremiumHub/components/dataExportTableAndFields.tsx
```diff
@@ -181,20 +202,39 @@ export const DataExportTableAndFields = ({ queryKey, searchCount, selectedGrades
...
+
+    const filterHash = hashPayload(filterTarget)
+
+    return hashMessage == filterHash
+  }
+
+  function customTimeframeMatches(timeframe) {
+    if (!customTimeframeStart || !customTimeframeEnd) return true
+
+    const remoteStart = timeframe?.custom_start?.split('T', 1)[0]
```
dandrabik: You might consider a helper method here for repeated logic, e.g. `format_iso_string(string)`

dandrabik (approved):
Thanks for cleaning this up. If your testing shows that the appropriate number of requests are being made (one per metric needed, and percentages aren't jumping around) then this let's ship this. 

Notes for future PRs:
- This PR is slightly difficult to review since there are many functionality changes lumped together. For bug fixes, it can be tricky to find stopping points if there are multiple things wrong (like here), but you'll get stronger PR reviews on the logic of the code if the PR is more focused, e.g. keep an eye out for complete shippable pieces that you can break off into a PR
- If you have a PR that does many things to achieve a single goal, consider a before/after screenshot/video to show that the goal has been achieved, e.g. show the console with a lot of requests, then only the appropriate amount. It's a good check for yourself and confirms the goal for the reviewer.

services/QuillLMS/client/app/bundles/PremiumHub/components/dataExportTableAndFields.tsx
anathomical: Good suggestion.  Working on such a helper earlier in development led to me realizing I didn't need this code at all, and this function in now unused.  (Another point in favor of what you mentioned about too many moving parts in the fix, meaning this didn't get cleaned up.)

PR #11216, opened 2023-11-06
admin reports: only save filterSelections to backend if selectedX is different from allX

emilia-friedberg:
## WHAT
Update the `PremiumFilterableReportsContainer` to only save a given filterSelection to the backend if `selectedX` is different from `allX`, and reciprocally, correctly interpret `null` values coming out of backend to mean `allSelected`.

## WHY
Thomas noticed an edge case where the fact that `allX` was being persisted to the backend as an array meant that the addition of new schools, teachers, or classrooms would result in those new values not being selected when we loaded the filters, which is potentially confusing UX. This fixes that.

## HOW
Just update behavior around the two request calls.

### Screenshots

### Notion Card Links
https://www.notion.so/quill/Tweak-Admin-report-filter-saving-logic-to-understand-all-selected-logic-76c2ed187e9a40dc835ba74640e450c9?pvs=4

PR Checklist | Your Answer
------------ | -------------

### DIFF
--- a/services/QuillLMS/client/app/bundles/PremiumHub/containers/PremiumFilterableReportsContainer.tsx
+++ b/services/QuillLMS/client/app/bundles/PremiumHub/containers/PremiumFilterableReportsContainer.tsx
@@ -1,5 +1,4 @@
 import * as React from 'react'
-import queryString from 'query-string';
 import moment from 'moment';
 import * as _ from 'lodash'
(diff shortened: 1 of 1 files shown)


anathomical (approved):
Nice, clean solution.  I was sort of expecting this to be more complicated, so nice job keeping it so straight-forward.

eadams17 (approved):

PR #11223, opened 2023-11-08
admin csv QA

happythenewsad:
## WHAT
QA items for Admin CSV download:

1. Set MIMEtype when generating a presigned S3 URL. 
2. Format specific CSV column values 

## WHY
1. Otherwise, Chrome will change the downloaded file's extension from `csv -> txt`
2. So that they look better to the user. 

## HOW
1. Fog gem config 
2. Add a switch statement in the CSV generator class, which applies various transforms. 

### Screenshots

### Notion Card Links
https://www.notion.so/quill/Admin-Data-Export-QA-November-6-f1bfffdb6c7d45a9b2705c6883456e6d

PR Checklist | Your Answer
------------ | -------------

### DIFF
--- a/services/QuillLMS/app/queries/adapters/csv/admin_premium_data_export.rb
+++ b/services/QuillLMS/app/queries/adapters/csv/admin_premium_data_export.rb
@@ -58,7 +58,7 @@ module Adapters
       end
 
       def self.validate_input!(bigquery_result, sym_columns)
-        if !(ORDERED_COLUMNS.keys.to_set >= sym_columns.to_set)
+        if (ORDERED_COLUMNS.keys & sym_columns).length != sym_columns.length
           raise UnhandledColumnError, "Requested column(s) not supported: #{sym_columns - ORDERED_COLUMNS.keys}"
         end
 
--- a/services/QuillLMS/app/queries/adapters/csv/admin_premium_data_export.rb
+++ b/services/QuillLMS/app/queries/adapters/csv/admin_premium_data_export.rb
@@ -47,7 +47,7 @@ module Adapters
         when :timespent
           (value / 60).to_i
         when :score
-          value&.round(2)
+          value.respond_to?(:*) ? "#{(value*100).round(0)}%" : ''
         else
           value
         end
@@ -58,7 +58,7 @@ module Adapters
       end
 
       def self.validate_input!(bigquery_result, sym_columns)
-        if sym_columns.to_set > ORDERED_COLUMNS.keys.to_set
+        if !(ORDERED_COLUMNS.keys.to_set >= sym_columns.to_set)
           raise UnhandledColumnError, "Requested column(s) not supported: #{sym_columns - ORDERED_COLUMNS.keys}"
         end
 
--- a/services/QuillLMS/app/queries/adapters/csv/admin_premium_data_export.rb
+++ b/services/QuillLMS/app/queries/adapters/csv/admin_premium_data_export.rb
@@ -58,7 +58,7 @@ module Adapters
       end
 
       def self.validate_input!(bigquery_result, sym_columns)
-        if !(sym_columns.to_set <= ORDERED_COLUMNS.keys.to_set)
+        if sym_columns.to_set > ORDERED_COLUMNS.keys.to_set
           raise UnhandledColumnError, "Requested column(s) not supported: #{sym_columns - ORDERED_COLUMNS.keys}"
         end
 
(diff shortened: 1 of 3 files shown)


services/QuillLMS/app/queries/adapters/csv/admin_premium_data_export.rb
```diff
@@ -6,43 +6,66 @@ module Adapters
...
           end
         end
       end
 
-      def self.validate_input!(bigquery_result)
+      def self.human_displayable_csv_headers(sym_columns)
+        sym_columns.map{|col| ORDERED_COLUMNS[col] }
+      end
+
+      def self.format_cell(sym_column, value)
```
brendanshean, line 43: I tend to prefer `when` ... `then` one-liners if the instruction is short:

```
def format_cell(sym_column, value)
  case sym_column
    when :completed_at then value&.strftime('%F')
    when :timespent then (value / 60).to_i
    when :score then value.respond_to?(:*) ? "#{(value * 100).round}%": ''
    else value.to_s
  end
end
```

services/QuillLMS/app/queries/adapters/csv/admin_premium_data_export.rb
```diff
@@ -6,43 +6,66 @@ module Adapters
...
+        case sym_column
+        when :completed_at
+          value&.strftime("%F")
+        when :timespent
+          (value / 60).to_i
+        when :score
+          value.respond_to?(:*) ? "#{(value*100).round(0)}%" : ''
```
brendanshean: Unless we're expecting some other type, what about `value.is_a?(Numeric)`

brendanshean (approved):
Some minor stylistic comments but looks good.

services/QuillLMS/app/queries/adapters/csv/admin_premium_data_export.rb
```diff
@@ -6,43 +6,66 @@ module Adapters
...
+      def self.human_displayable_csv_headers(sym_columns)
+        sym_columns.map{|col| ORDERED_COLUMNS[col] }
+      end
+
+      def self.format_cell(sym_column, value)
+        case sym_column
+        when :completed_at
+          value&.strftime("%F")
+        when :timespent
+          (value / 60).to_i
```
anathomical: I'm pretty sure that `timespent` is an Integer, so this will do integer division, meaning we shouldn't need to cast the `to_i` at the end.  Unless Ruby is interpreting the BigQuery data differently than it does Postgres.

services/QuillLMS/app/queries/adapters/csv/admin_premium_data_export.rb
```diff
@@ -6,43 +6,66 @@ module Adapters
...
+        when :score
+          value.respond_to?(:*) ? "#{(value*100).round(0)}%" : ''
+        else
+          value
+        end
+      end
+
+      def self.row_contains_requested_columns?(row, requested_columns)
+        (row.keys & requested_columns).length == requested_columns.length
+      end
```
anathomical, line 58: If, somehow, `requested_columns` ended up having duplicate key names, this comparison will fail as `&` will collapse duplicates.  My bet is that we have no reason to believe that this will ever happen, so no need to make a change here, but I wanted to flag it to make sure we'd at least considered the possibility.

services/QuillLMS/spec/queries/adapters/csv/admin_premium_data_export_spec.rb
```diff
@@ -5,6 +5,14 @@
 describe Adapters::Csv::AdminPremiumDataExport do
   subject { described_class }
 
+  describe '#format_cell' do
+    it { expect(described_class.format_cell(:completed_at, DateTime.new(2020,1,1))).to eq('2020-01-01')}
+    it { expect(described_class.format_cell(:activity_pack, 'foo')).to eq('foo')}
+    it { expect(described_class.format_cell(:timespent, 61)).to eq(1)}
+    it { expect(described_class.format_cell(:score, 0.66777)).to eq('67%')}
+    it { expect(described_class.format_cell(:not_a_special_case, 1.5)).to eq(1.5)}
```
anathomical, line 13: I always appreciate when arbitrary values like this help me understand what the test case is.  Nice!

anathomical (approved):
I flagged a possible redundant `to_i` call, and an edge case that I don't think we need to worry about, but felt was worth considering.  But overall this looks good to me.

PR #11227, opened 2023-11-08
Lms/admin diagnostic reports overview

anathomical:
## WHAT
- Create a background worker to populate AdminDiagnosticOverview reports into cache
- Create a controller that can either pull a report from cache or kick off a background worker to populate said cache
- Add a route to the new controller
- Add a parent query for AdminDiagnosticOverview queries that understands how to aggregate data via different parameters
- - Create a child Query to find Pre-diagnostics Assigned count
- - Create a child query to find Pre-diagnostics Completed count and average score per playthrough
- - Create a child query to find Recommendations completed with average activities per student and average time spent per student
- - Create a child query to find Post-diagnostics Assigned count
- - Create a child query to find Post-diagnostics Completed count and average score per playthrough
## WHY
The five queries that are configured to be run all require different aggregation and join pathing to retrieve data efficiently so splitting them apart like this lets us run them in parallel for best performance.  We're basically following the pattern established in Admin Snapshots.
## HOW
- This is basically just a slightly modified copy of `Snapshots::CacheSnapshotTopXWorker`
- This is basically just a slightly modified copy of `SnapshotsController`
- Add an entry to `config/routes`
- This parent query consolidates shared query parts and logic around rolling up grade/teacher/classroom-level data into diagnostic-level data
- Each of the child queries adds enough details to actually retrieve the rolled-up data we care about and then describe how its data should be rolled up to the diagnostic level

### Notion Card Links
https://www.notion.so/quill/Admin-Diagnostic-Growth-Report-Performance-Overview-Backend-0f4c5bae40a74e68b590093fdc036e48?pvs=4

PR Checklist | Your Answer
------------ | -------------


services/QuillLMS/app/controllers/admin_diagnostic_reports_controller.rb
```diff
@@ -0,0 +1,96 @@
+# frozen_string_literal: true
+
+class AdminDiagnosticReportsController < ApplicationController
```
anathomical, line 3: This is 95% just a copy of `SnapshotsController` with tweaks to actions and `WORKERS_FOR_ACTIONS`.  I plan to do PRs for the other two reports in the Admin Diagnostic project, and then follow up with one that refactors all of the shared logic for these controllers along with things like Timeframes, CacheKeys, and Workers that consolidates all of the shared logic and DRYs this stuff up.

services/QuillLMS/app/queries/admin_diagnostic_reports/diagnostic_aggregate_query.rb
```diff
@@ -0,0 +1,150 @@
...
+      'classroom'
+    ]
+    DIAGNOSTIC_ORDER_BY_ID = [
+      1663, # Starter Baseline Diagnostic (Pre)
+      1668, # Intermediate Baseline Diagnostic (Pre)
+      1678, # Advanced Baseline Diagnostic (Pre)
+      1161, # ELL Starter Baseline Diagnostic (Pre)
+      1568, # ELL Intermediate Baseline Diagnostic (Pre)
+      1590  # ELL Advanced Baseline Diagnostic (Pre)
+    ]
```
anathomical: I'm not super-happy with hard-coding the IDs in here, and there are methods for identifying active Pre Diagnostics without doing this, but because we have a specific order that we want to maintain for these queries, this was the cleanest thing I could come up with.  I'd be totally open to alternate approaches that involve less risk around forgetting to update these and not being able to figure out why a new Diagnostic isn't loading in this report.

services/QuillLMS/app/queries/admin_diagnostic_reports/diagnostic_aggregate_query.rb
```diff
@@ -0,0 +1,150 @@
...
+
+    def where_clause
+      super + <<-SQL
+          #{relevant_diagnostic_where_clause}
+      SQL
+    end
+
+    def relevant_diagnostic_where_clause
+      "AND activities.id IN (#{DIAGNOSTIC_ORDER_BY_ID.join(',')})"
+    end
```
anathomical, line 66: Another weird potential gotcha: the `activities` table is not actually included as part of this abstract query, and is instead joined (via different paths) in each inheritor.  This introduces a bit of undocumented magic that I'd be glad to eliminate if anyone has a good suggestion.  (We join via different paths for different queries because sometimes this join comes from `activity_sessions`, but sometimes it comes from a join on `follow_up_activity_id` on separate `activities` join when trying to find post diagnostic information or through `recommendations` when finding follow up activities.)

services/QuillLMS/app/queries/admin_diagnostic_reports/diagnostic_aggregate_query.rb
```diff
@@ -0,0 +1,150 @@
...
+
+    def group_by_clause
+      "GROUP BY activities.id, activities.name, #{aggregation_clause}"
+    end
```
anathomical: Another weird potential gotcha: the `activities` table is not actually included as part of this abstract query, and is instead joined (via different paths) in each inheritor.  This introduces a bit of undocumented magic that I'd be glad to eliminate if anyone has a good suggestion.  (We join via different paths for different queries because sometimes this join comes from `activity_sessions`, but sometimes it comes from a join on `follow_up_activity_id` on separate `activities` join when trying to find post diagnostic information or through `recommendations` when finding follow up activities.)

services/QuillLMS/app/workers/admin_diagnostic_reports/diagnostic_overview_worker.rb
```diff
@@ -0,0 +1,79 @@
+# frozen_string_literal: true
+
+module AdminDiagnosticReports
```
anathomical, line 3: This is basically a copy of `Snapshots::CacheSnapshotTopXWorker` with tweaks to `PUSHER_EVENT` and `QUERIES`.  I mentioned a planned refactor to unify Controller logic in a parent, and I intend to do the same thing with these workers in that refactor.

services/QuillLMS/spec/controllers/admin_diagnostic_reports_controller_spec.rb
```diff
@@ -0,0 +1,235 @@
...
+
+        it 'should 400 if the school_ids param is empty' do
+          controller_actions.each do |action, query_name|
+            post action, params: { query: query_name, timeframe: timeframe_name, school_ids: [] }
+
+            expect(response.status).to eq(400)
+          end
+        end
+
+        it 'should 400 if the query name provided is not valid for the endpoint' do
```
dandrabik, line 142: It might be worth testing the message of the body. I don't think this returns the error you are expecting, since the missing `school_ids` error will go first.

services/QuillLMS/app/queries/admin_diagnostic_reports/diagnostic_aggregate_query.rb
```diff
@@ -0,0 +1,180 @@
...
+          next row.merge({name: "No grade selected"}) if row[:name].nil?
+          # to_i returns 0 for non-numeric strings, so this will only apply to numbered grades
+          next row.merge({name: "Grade #{row[:name]}"}) if row[:name].to_i > 0
+
+          row
+        end
+    end
+
+    private def aggregate_sort(diagnostic_rows)
+      diagnostic_rows.sort do |a, b|
```
dandrabik: Purely a preference thing, but it might be easier to run these sorts using [`sort_by`](https://ruby-doc.org/core-3.0.2/Enumerable.html#method-i-sort_by), since the syntax is a little easier to parse, e.g. 

```ruby
diagnostic_rows.sort_by do |row|
  ## TODO grade logic
  ## TODO teacher logic
  row[:name]
end

File: services/QuillLMS/app/queries/admin_diagnostic_reports/diagnostic_aggregate_query.rb [hunk 8]
```diff
@@ -0,0 +1,180 @@
...
+        'grade' => "classrooms.grade",
+        'classroom' => "classrooms.name",
+        'teacher' => "users.name"
+      }.fetch(additional_aggregation)
+    end
+
+    private def post_process(result)
+      return [] if result.empty?
+
+      result.group_by { |row| row[:diagnostic_name] }
```
dandrabik, line 119: Looking at the amount of post-processing, I'm curious if there is a better query to run to get the data in the right shape, e.g. using a few WITH queries.

dandrabik (approved):
Thanks for putting this together. I'm a little wary of the amount of post-processing needed to shape the data, and I wonder if 1) there is a way to clean that up a little bit or 2) there's a better SQL query that would return the data closer to the correct format.

I'm approving since it would be good to move these along (and we can refactor), but worth a pass to consider those options (happy to look at this together on Thursday).

Thanks for your work on this.

services/QuillLMS/spec/controllers/admin_diagnostic_reports_controller_spec.rb
anathomical: Good catch.  I'll add expectations around the specific message coming back.

services/QuillLMS/app/queries/admin_diagnostic_reports/diagnostic_aggregate_query.rb
```diff
@@ -0,0 +1,197 @@
...
+    class InvalidAggregationError < StandardError; end
+
+    attr_reader :additional_aggregation
+
+    AGGREGATION_OPTIONS = [
+      'grade',
+      'teacher',
+      'classroom'
+    ]
+    DIAGNOSTIC_ORDER_BY_ID = [
```
happythenewsad, line 14: This list is different from the one generated by 

`Activity.diagnostic_activity_ids` 

... which is different from the list in assignmentFlowConstants. We have many lists of diagnostic activities in the codebase :-/

services/QuillLMS/app/lib/snapshots/cache_keys.rb
```diff
@@ -2,17 +2,17 @@
 
 module Snapshots
   class CacheKeys
-    def self.generate_key(query, timeframe_name, current_start, current_end, school_ids, additional_filters: {})
-      root_key(query, timeframe_name, current_start, current_end, school_ids)
+    def self.generate_key(report, query, timeframe_name, current_start, current_end, school_ids, additional_filters: {})
```
happythenewsad, line 5: I like your decision to separate the concern of key generation from the concern of cache fetching.

happythenewsad (approved):

services/QuillLMS/app/queries/admin_diagnostic_reports/diagnostic_aggregate_query.rb
anathomical: Agreed that the proliferation isn't great.  In this case, though, this is a legitimately different list.  It's specifically only the pre-diagnostics plus the College Board diagnostics.  Post diagnostics are not included in this list.  In theory we could source this from `assignmentFlowConstants` as `[starterPreTest, intermediatePreTest, advancedPreTest, ellStarterPreTest, ellIntermediatePreTest, ellAdvancedPreTest, apWritingSkills, preAPWritingSkills1, preAPWritingSkills2, springboardWritingSkills]`.  If that data was available in the front-end, anyway.

PR #11228, opened 2023-11-09
Bugfix and Speed Improvements on Regex Rules Admin Panel

cissyxyu:
## WHAT
This page has very slow load times, mostly due to unnecessary API calls fetching all rules for a rule type or activity rather than the specific needed rules. This PR fixes slow load times and adds some requested features like loading spinners.

## WHY
Admin use this page a lot and they currently have a lot of trouble creating and editing regex rules efficiently because of slow load times.

## HOW
**Back-end:** Add a parameter to the backend endpoint activity/rules that specifies a rule_type. Only fetch rules for that rule type if the parameter is specified.
**Front-end:** 
-Only fetch rules of the specified activity_id and rule_type when calling the Rules API, instead of fetching all rules.
-Loading spinners: Add separate loading spinners for each regex rule section that appear after an update_rule_order call, so the admin know when a call is being made and they don't have to continuously reorder the rules attempting to see changes.
-Update order bugfix: add a .then() clause so the refresh is only called after the rule order call has returned. Only fetch rules for the updated rule_type, not all rules.
-Delete rule bugfix: After delete, add a .then() clause to refetch query so we can be sure the API is called after the rule is deleted.

### Screenshots
Screenshot 2023-11-09 at 3 46 48 PM

### Notion Card Links
https://www.notion.so/quill/Improve-Performance-of-Creating-and-Deleting-Regex-Rules-616f66e402c1416f9ab5df57dfa15c46?pvs=4
https://www.notion.so/quill/Update-the-Regex-Rules-Page-to-more-easily-access-the-View-link-d20145415c544eed9bc6c8581b417e9e?pvs=4
https://www.notion.so/quill/Investigate-Regex-Rules-Priority-Sequencing-78e763c8b92c4844be226a2a3862b6c0?pvs=4

PR Checklist | Your Answer
------------ | -------------

### DIFF
--- a/services/QuillLMS/client/app/bundles/Staff/components/evidence/configureRules/ruleViewForm.tsx
+++ b/services/QuillLMS/client/app/bundles/Staff/components/evidence/configureRules/ruleViewForm.tsx
@@ -92,11 +92,9 @@ const RuleViewForm = ({
 
   const queryClient = useQueryClient()
 
-  const promptIds = activityData.prompts.map(p => p.id)
-
   // cache ruleSets data for handling rule suborder
   const { data: rulesData } = useQuery({
-    queryKey: [`rules-${activityId}-${rule_type}`, activityId, promptIds.join(','), rule_type],
+    queryKey: [`rules-${activityId}-${rule_type}`, activityId, null, rule_type],
     queryFn: fetchRules,
   });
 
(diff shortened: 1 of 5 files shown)


emilia-friedberg (approved):
LGTM!

happythenewsad (approved):

PR #11237, opened 2023-11-13
fix react dom render console error

emilia-friedberg:
## WHAT
Update `react_on_rails` gem and `react-on-rails` package to use React 18 syntax.

## WHY
To get rid of the console error we're encountering. Note: while the card was written about the console.errors from the jest tests, it is not possible to fix that error with a package upgrade because it is actually coming from enzyme, which does not support React 18 and has no future versions coming. This fixes the same issue on production but we will still see the failures in the tests. The only ways to get rid of it would be to either migrate all `enzyme` tests to `react-testing-library`, which is theoretically the plan over time but a large undertaking to do all at once, or choose to silence jest console errors.

## HOW
Just upgrade the packages (and the one spot in the app where we were explicitly using the wrong syntax ourselves) and test to confirm everything works and the console error no longer appears.

### Screenshots

### Notion Card Links
https://www.notion.so/quill/Console-log-cleanup-ReactDOM-render-3652-668312eb777b4bd8a0a921386995c907?pvs=4

PR Checklist | Your Answer
------------ | -------------
Have you deployed to Staging? | Not yet - deploying now


happythenewsad (approved):

services/QuillLMS/bin/bundle
eadams17, line 1: I think this file wasn't meant to be added?

eadams17 (approved):

PR #11242, opened 2023-11-15
add wipe_question_cache method and add spec

emilia-friedberg:
## WHAT
Update the CMS so that we wipe the questions cache when responses are created, updated, or deleted.

## WHY
We were seeing a bug with newly-added questions where grading changes were not being reflected in what was coming out of the cache, which persists for 24 hours and never gets deleted otherwise, so turk users were unable to progress past certain questions.

## HOW
Add a `wipe_question_cache` function to the `Response` model so that changes to non-count variables will wipe the cache, allowing users to get the updated grading. 

### Screenshots

### Notion Card Links
https://www.notion.so/quill/Query-to-pull-turk-responses-broken-271d4032c6c64761ba2c3defd59a1dba?d=d40b61cc98e44cff83973508969f58cc

PR Checklist | Your Answer
------------ | -------------

### DIFF
--- a/services/QuillCMS/spec/models/response_spec.rb
+++ b/services/QuillCMS/spec/models/response_spec.rb
@@ -1,84 +1,84 @@
 require "rails_helper"
 
 RSpec.describe Response do
-  context "ActiveRecord callbacks" do
-    it "after_create_commit calls #create_index_in_elastic_search and #wipe_question_cache" do
-      response = Response.new()
-      expect(response).to receive(:create_index_in_elastic_search)
-      expect(response).to receive(:conditional_wipe_question_cache)
-      response.save
-    end
-
-    it "after_update_commit calls #update_index_in_elastic_search and #conditional_wipe_question_cache" do
-      response = Response.create()
-      expect(response).to receive(:update_index_in_elastic_search)
-      expect(response).to receive(:conditional_wipe_question_cache)
-      response.update(text: 'covfefe')
-    end
-
-    it "before_destroy calls #destroy_index_in_elastic_search and #wipe_question_cache" do
-      response = Response.create()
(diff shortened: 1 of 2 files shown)


services/QuillCMS/spec/models/response_spec.rb
```diff
@@ -41,4 +44,58 @@
...
+      expect(response).not_to receive(:wipe_question_cache)
+      response.update(child_count: 3)
+    end
+
+    it 'does not call wipe_question_cache when first_attempt_count is updated' do
+      expect(response).not_to receive(:wipe_question_cache)
+      response.update(first_attempt_count: 2)
+    end
+  end
+end
```
anathomical: Looks like all of this needs one level of indentation.

happythenewsad (approved):

services/QuillCMS/spec/models/response_spec.rb
```diff
@@ -41,4 +44,58 @@
...
+      Rails.cache.write(cache_key, 'cached content')
+    end
+
+    it 'clears the cache for the given question UID' do
+      expect(Rails.cache.read(cache_key)).to eq('cached content')
+
+      response = Response.create(question_uid: question_uid)
+      response.wipe_question_cache
+
+      expect(Rails.cache.read(cache_key)).to be_nil
```
anathomical: This is pure preference, so feel free to skip it, but I find this kind of logic more readable in an `expect to change` wrapper:
```ruby
expect do
  response = Response.create(question_uid: question_uid)
  response.wipe_question_cache
end.to change(Rails.cache.read(cache_key)).from('cached content').to(nil)
```

anathomical (approved):
Two small style things in the spec file, but this seems like a good solution to me.  I also appreciate how thorough you got with the specs.

PR #11245, opened 2023-11-15
add extra attributes to results summary payload

emilia-friedberg:
## WHAT
Add extra attributes to results summary payload so we can hopefully track down a weird bug.

## WHY
We're occasionally (looks like three reports in the last two sprints) seeing a bug where teachers wind up with a totally unrelated classroom's data in their diagnostic results summary cache. This problem has been tricky to track down, both because the cache either gets overwritten or wiped within 24 hours and also because there's just very little to go on in terms of how this might be happening. Even when I have caught it in the replicable window, it's clear that the wrong data is in the cache but all the methods that should be populating the cache appear to work as intended. I'm hoping that adding a little more data will give us more insight into how this might be happening. If anyone else has any ideas, please let me know.

## HOW
Just add more data about what's actually getting passed into this method so we can see where things might be going wrong. This seems redundant, but I'm hoping we'll see that what's getting passed in is not what we expect, and then we can work from there. If the correct classroom, unit, and activity ids are getting passed in but the student results are still unrelated, we'll know it's the code in this file that's not working as expected rather than the caching or controller logic.

### Screenshots

### Notion Card Links
https://www.notion.so/quill/Students-who-are-not-rostered-in-a-class-appearing-in-Student-results-report-2264796d865144a98178856820d17983?pvs=4
https://www.notion.so/quill/Students-that-are-not-rostered-in-class-appearing-in-Student-results-report-c2245d18791744a58a1950418b4dfd15

PR Checklist | Your Answer
------------ | -------------

### DIFF
--- a/services/QuillLMS/app/controllers/concerns/results_summary.rb
+++ b/services/QuillLMS/app/controllers/concerns/results_summary.rb
@@ -21,7 +21,10 @@ module ResultsSummary
 
     {
       student_results: student_results,
-      skill_group_summaries: @skill_group_summaries
+      skill_group_summaries: @skill_group_summaries,
+      activity_id: activity_id,
+      classroom_id: classroom_id,
+      unit_id: unit_id
     }
   end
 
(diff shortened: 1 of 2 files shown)


brendanshean (approved):

happythenewsad (approved):
If the queries that are hydrating the problematic caches are kicked off by sidekiq jobs, we could also append diagnostic information there. Particularly, the user who triggered the job.

anathomical (approved):
I think this is a really good approach to chasing down these kinds of ephemeral bugs.

PR #11247, opened 2023-11-15
fix intermittently failing query_spec specs

happythenewsad:
## WHAT
Fix failing specs 

## WHY
Failing spec bad. 

## HOW
It's an order issue, not a time issue. It's not obvious, but 
`unit_activities.publish_date >= NOW() AS scheduled` can yield a nil value 

So when the order of the query results is not the intended order, an error occurs. 

I fixed this by retrieving the desired row from the result set, rather than taking the first row. 

### Screenshots

### Notion Card Links
n/a

PR Checklist | Your Answer
------------ | -------------
Have you added and/or updated tests? |  n/a spec change only


brendanshean (approved):
Thanks for fixing this!

emilia-friedberg (approved):
Thank you for fixing this!

anathomical (approved):

PR #11255, opened 2023-11-16
lms/restrict-period-query-to-owner-teachers

anathomical:
## WHAT
- Restrict PeriodQueries to only include owner teachers
- Modify two queries (`ActiveTeachers` and `TeacherAccountsCreated`) so that they ignore the new restriction since they are explicit exceptions to the new rule
## WHY
This lets us avoid double-counting cases where a classroom has a coteacher and we aren't using DISTINCT to avoid said double-count and protects us from the need to guarantee DISTINCT clauses in future queries to avoid double counting in cases where the presence of a co-teacher will result in doubling (or tripling, etc.) rows because multiple teachers are associated with, say, a `ClassroomUnit`.
## HOW
Just add a new WHERE clause to the base `PeriodQuery` that excludes users who are not classroom owners from our queries.  And in cases where a teacher is a teacher for one class and a coteacher for another class, only include data for which the teacher is an owner.

### Notion Card Links
https://www.notion.so/quill/Update-PeriodQuery-so-that-it-only-includes-data-for-owner-teachers-4d404ad519734594a732d01020c439ef?pvs=4

PR Checklist | Your Answer
------------ | -------------

### DIFF
--- a/services/QuillLMS/app/queries/snapshots/active_teachers_query.rb
+++ b/services/QuillLMS/app/queries/snapshots/active_teachers_query.rb
@@ -18,5 +18,10 @@ module Snapshots
     def relevant_date_column
       "user_logins.created_at"
     end
+
+    # For this query we want to count all teachers, not just classroom owners
+    def owner_teachers_only_where_clause
+      ""
+    end
   end
 end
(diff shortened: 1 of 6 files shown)


happythenewsad (approved):

dandrabik (approved):
:+1: Thanks for putting this together!

PR #11274, opened 2023-11-21
fix children with the same key errors + unique "key" errors

emilia-friedberg:
## WHAT
Fix all instances of "Each child in a list should have a unique "key" prop." and "Encountered two children with the same key" console errors that appear in our frontend test suite.

## WHY
We are trying to clean up console errors.

## HOW
Manually go through and fix each instance where the test suite logged an error (mostly in the components themselves, but sometimes it was an issue with the test setup).

### Screenshots

### Notion Card Links
https://www.notion.so/quill/Console-log-cleanup-child-unique-key-prop-43-ce34e6b73a724b37a5f36e907d8638e4?pvs=4

https://www.notion.so/quill/Console-log-cleanup-children-with-the-same-key-148-2f9685628b2f4b99bc7aca8b9387e36c?pvs=4

PR Checklist | Your Answer
------------ | -------------


eadams17 (approved):
Thanks for going through all of these and sorting them out!

services/QuillLMS/client/app/bundles/Teacher/components/unitTemplate/unitTemplateSelectedActivitiesTable.tsx
```diff
@@ -26,10 +26,10 @@ export const UnitTemplateSelectedActivitiesTable = ({ activities, selectedActivi
   function renderHeaders() {
     return(
       <tr className="data-table-headers">
-        {unitTemplateDataTableFields.map(header => {
+        {unitTemplateDataTableFields.map((header, i) => {
           const { width, name } = header;
           const style = { width: `${width}`, minWidth: `${width}` }
-          return <th className="data-table-header" scope="col" style={style}>{name}</th>;
+          return <th className="data-table-header" key={i} scope="col" style={style}>{name}</th>;
```
anathomical: I don't think it's likely to matter in this use-case, but what I've read suggests that using index values for `key` can lead to unexpected behavior (since the order of an array can change, but the keys will remain the same, which will signal to React that it does not need to re-render).

anathomical (approved):
One note on using `i` as a value for `key`, but overall this is amazing.  So exciting to see this getting cleaned up, especially given how much work it was probably chasing all of these down.  Thanks!

cissyxyu (approved):
thank you for doing this!

PR #11276, opened 2023-11-21
BiqQuery - 3 bug fixes

happythenewsad:
## WHAT
1. For DataExportQuery, converts INNER JOIN on standards to LEFT OUTER JOIN
2. Updates QuillBigQuery::Runner to use `to_a`
3. Wraps DataExportQuery in a GROUP BY 

## WHY
1. So that we don't filter out activities (like some diagnostics) that don't have a standard
2. Because our previous invocation of the BigQuery gem was silently truncating result sets. This forces the return of all pages of a paginated result set. This potentially impacts all BigQuery queries.
3. So that DataQueryExport aligns with ActivitiesCompletedQuery in returning 1 row per activity session

### Screenshots

### Notion Card Links
https://www.notion.so/quill/DataExportQuery-and-ActivityCompletedCount-disrepancy-4e68dc50a3814cd7948275aac36e16d4

PR Checklist | Your Answer
------------ | -------------

### DIFF
--- a/services/QuillLMS/app/services/quill_big_query/runner.rb
+++ b/services/QuillLMS/app/services/quill_big_query/runner.rb
@@ -5,7 +5,7 @@
 module QuillBigQuery
   class Runner
     def self.execute(query)
-      Google::Cloud::Bigquery.new.query(query)
+      Google::Cloud::Bigquery.new.query(query).all.to_a
     rescue => e
       raise e, "Query: #{query}"
     end
--- a/services/QuillLMS/spec/queries/snapshots/data_export_query_spec.rb
+++ b/services/QuillLMS/spec/queries/snapshots/data_export_query_spec.rb
@@ -64,7 +64,6 @@ module Snapshots
       let(:cte_records) { [runner_context] }
 
       context 'basic shape tests' do
-        # TODO: this test should pass when the production query bug is fixed
         it 'should have one row per activity session' do
           expect(results.map{|r| r[:activity_session_id] }.uniq).to eq(
             results.map{|r| r[:activity_session_id] }
--- a/services/QuillLMS/spec/queries/snapshots/data_export_query_spec.rb
+++ b/services/QuillLMS/spec/queries/snapshots/data_export_query_spec.rb
@@ -65,11 +65,11 @@ module Snapshots
 
       context 'basic shape tests' do
         # TODO: this test should pass when the production query bug is fixed
-        # it 'should have one row per activity session' do
-        #   expect(results.map{|r| r[:activity_session_id] }.uniq).to eq(
-        #     results.map{|r| r[:activity_session_id] }
-        #   )
-        # end
+        it 'should have one row per activity session' do
+          expect(results.map{|r| r[:activity_session_id] }.uniq).to eq(
+            results.map{|r| r[:activity_session_id] }
+          )
+        end
 
         it { expect(results.count).to eq 10 }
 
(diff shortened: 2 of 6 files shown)


services/QuillLMS/spec/queries/snapshots/untruncated_data_export_query_spec.rb
```diff
@@ -0,0 +1,100 @@
...
+      let(:runner_context) {
+        [
+          classrooms,
+          teachers,
+          classrooms_teachers,
+          schools,
+          schools_users,
+          classroom_units,
+          activities,
+          classroom_units,
```
brendanshean: There is a duplicate `classroom_units` in this runner context.

services/QuillLMS/spec/queries/snapshots/untruncated_data_export_query_spec.rb
```diff
@@ -0,0 +1,100 @@
...
+          units,
+          students
+        ]
+      }
+
+      let(:cte_records) { [runner_context] }
+
+      context 'basic shape tests' do
+        it 'should have one row per activity session' do
+          expect(results.map{|r| r[:activity_session_id] }.uniq).to eq(
```
brendanshean: What about `pluck` here:  
```expect(results.pluck(:activity_session_id).uniq).to eq(results.pluck(:activity_session_id))```

**Review by brendanshean — APPROVED** [review]
Nice work figuring these bugs out, especially the truncated result sets.

File: services/QuillLMS/spec/queries/snapshots/data_export_query_spec.rb [hunk 3]
```diff
@@ -64,12 +64,11 @@ module Snapshots
...
 
       context 'basic shape tests' do
-        # TODO: this test should pass when the production query bug is fixed
-        # it 'should have one row per activity session' do
-        #   expect(results.map{|r| r[:activity_session_id] }.uniq).to eq(
-        #     results.map{|r| r[:activity_session_id] }
-        #   )
-        # end
+        it 'should have one row per activity session' do
+          expect(results.map{|r| r[:activity_session_id] }.uniq).to eq(
```
brendanshean: What about pluck here:
```expect(results.pluck(:activity_session_id).uniq).to eq(results.pluck(:activity_session_id))```

File: services/QuillLMS/app/services/quill_big_query/runner.rb [hunk 4]
```diff
@@ -5,7 +5,7 @@
 module QuillBigQuery
   class Runner
     def self.execute(query)
-      Google::Cloud::Bigquery.new.query(query)
+      Google::Cloud::Bigquery.new.query(query).all.to_a
```
dandrabik, line 8: This seems like a reasonable change after reading up on the API.

https://cloud.google.com/ruby/docs/reference/google-cloud-bigquery/latest/Google-Cloud-Bigquery-Data#Google__Cloud__Bigquery__Data_all_instance_

Given the smaller numbers for other reports, it's likely this wasn't an issue for the Usage Snapshot (since there wouldn't only every be a first page), but this is worth running by @anathomical since this touches his query work.

dandrabik (approved):
Thanks for fixing this up, LGTM! See my comment, might be worth running this by @anathomical.

services/QuillLMS/app/services/quill_big_query/runner.rb
happythenewsad: Agreed - I should have written

```
This potentially impacts all BigQuery queries where rows > ~30k`, which, as of today, only applies to the DataExportQuery.
```

PR #11287, opened 2023-11-28
Filter out broken caching spec

brendanshean:
## WHAT
Exclude broken caching spec from test suite.

## WHY
The `retry: 3` tag is not preventing failure.

## HOW
Add a new tag `broken` and exclude it from test runs.

### Screenshots

### Notion Card Links

PR Checklist | Your Answer
------------ | -------------

### DIFF
--- a/services/QuillLMS/spec/models/concerns/user_cacheable_spec.rb
+++ b/services/QuillLMS/spec/models/concerns/user_cacheable_spec.rb
@@ -2,7 +2,7 @@
 
 require 'rails_helper'
 
-RSpec.describe UserCacheable, type: :model, retry: 3 do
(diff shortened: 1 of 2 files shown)


anathomical (approved):
This seems like a nicely reusable solution here.  I wonder if we also want a `TODO` comment to that we can find this more easily when we need to, but maybe we can just use a search for `broken: true` to find anything we flag with this.

dandrabik (approved):
:+1: I like this pattern.

brendanshean:
> This seems like a nicely reusable solution here. I wonder if we also want a `TODO` comment to that we can find this more easily when we need to, but maybe we can just use a search for `broken: true` to find anything we flag with this.

@anathomical Good idea.

PR #11292, opened 2023-11-30
Lms/admin diagnostic performance by skill

anathomical:
## WHAT
- Add a new query to aggregate Diagnostic data for admins by skill (instead of by diagnostic)
- Add a worker and controller to cache and fetch the data from this query

NOTE: This query is currently quite time-consuming (~15 seconds regardless of the size of the dataset being analyzed).  We're intent is to address performance concerns with BigQuery in general next.
## WHY
- We need this data for the "By Skill" part of the Admin Diagnostic reports
- We have an established pattern for these workers and controllers, so we're just doing a bunch of copy/paste work here.  We should follow this project up with a refactor to consolidate the shared logic into some sort of shared code rather than doing all of this duplication
## HOW
- The nature of the data calculation we need to do involves multiple steps of aggregation.  In order to facilitate this, I developed a query with a series of `WITH` statements that feed into each other to successively aggregate the necessary data.  This involved a mild refactor of the parent query in order to support an arbitrary number of `WITH` statements
- - This also involved adding support for a new "type" of aggregation in the parent query to handle the calculation of percentages
- Copy and paste from the reference worker and controller (in this case, the one for the Admin Diagnostic Overview), and modify as appropriate

### Notion Card Links
https://www.notion.so/quill/Admin-Diagnostic-Growth-Report-Growth-Per-Skill-1cdc878fb976477ea40089ddef6711ca?pvs=4

PR Checklist | Your Answer
------------ | -------------

### DIFF
--- a/services/QuillLMS/app/controllers/admin_diagnostic_skills_controller.rb
+++ b/services/QuillLMS/app/controllers/admin_diagnostic_skills_controller.rb
@@ -1,0 +1,116 @@
+# frozen_string_literal: true
+
+class AdminDiagnosticSkillsController < ApplicationController
+  CACHE_REPORT_NAME = 'admin-diagnostic-skills'
+  WORKERS_FOR_ACTIONS = {
+    "report" => AdminDiagnosticReports::DiagnosticSkillsWorker
+  }
+
+  before_action :set_query
+  before_action :validate_request
+  before_action :authorize_request
+
+  def report
+    render json: retrieve_cache_or_enqueue_worker(WORKERS_FOR_ACTIONS[action_name])
+  end
+
+  private def set_query
+    @query = permitted_params[:query]
+    @group_by = permitted_params[:group_by]
+    @diagnostic_id = permitted_params[:diagnostic_id]
+  end
+
+  private def validate_request
+    return render json: { error: 'timeframe must be present and valid' }, status: 400 unless timeframe_param_valid?
+
+    return render json: { error: 'school_ids are required' }, status: 400 unless school_ids_param_valid?
+
+    return render json: { error: 'group_by param is required' }, status: 400 unless school_ids_param_valid?
+
+    return render json: { error: 'diagnostic_id param is required' }, status: 400 unless school_ids_param_valid?
+
+    return render json: { error: 'unrecognized query type for this endpoint' }, status: 400 unless WORKERS_FOR_ACTIONS[action_name]::QUERIES.keys.include?(@query)
+  end
+
+  private def timeframe_param_valid?
+    Snapshots::Timeframes.find_timeframe(permitted_params[:timeframe]).present?
+  end
+
+  private def school_ids_param_valid?
+    permitted_params[:school_ids]&.any?
+  end
+
+  private def group_by_param_valid?
+    permitted_params[:group_by].present?
+  end
+
+  private def diagnostic_id_param_valid?
+    permitted_params[:diagnostic_id].present?
+  end
+
+  private def authorize_request
+    schools_user_admins = current_user.administered_schools.pluck(:id)
+
+    return if permitted_params[:school_ids]&.all? { |param_id| schools_user_admins.include?(param_id.to_i) }
+
+    return render json: { error: 'user is not authorized for all specified schools' }, status: 403
+  end
+
+  private def retrieve_cache_or_enqueue_worker(worker)
+    timeframe_start, timeframe_end = Snapshots::Timeframes.calculate_timeframes(
+      permitted_params[:timeframe],
+      custom_start: permitted_params[:timeframe_custom_start],
+      custom_end: permitted_params[:timeframe_custom_end])
+    cache_key = cache_key_for_timeframe(permitted_params[:timeframe], timeframe_start, timeframe_end)
+    response = Rails.cache.read(cache_key)
+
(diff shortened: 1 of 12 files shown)


services/QuillLMS/app/controllers/admin_diagnostic_skills_controller.rb
```diff
@@ -0,0 +1,116 @@
...
+  private def validate_request
+    return render json: { error: 'timeframe must be present and valid' }, status: 400 unless timeframe_param_valid?
+
+    return render json: { error: 'school_ids are required' }, status: 400 unless school_ids_param_valid?
+
+    return render json: { error: 'group_by param is required' }, status: 400 unless school_ids_param_valid?
+
+    return render json: { error: 'diagnostic_id param is required' }, status: 400 unless school_ids_param_valid?
+
+    return render json: { error: 'unrecognized query type for this endpoint' }, status: 400 unless WORKERS_FOR_ACTIONS[action_name]::QUERIES.keys.include?(@query)
```
brendanshean: For consistency, what about a boolean method for this one too:

```return render json: { error: 'unrecognized query type for this endpoint' }, status: 400 unless valid_query_type?```

File: services/QuillLMS/spec/queries/admin_diagnostic_reports/diagnostic_performance_by_skill_query_spec.rb [hunk 2]
```diff
@@ -0,0 +1,149 @@
...
+      # Ideally we'd break this up into multiple cases, and also account for more varied inputs, but for some reason this query takes 15-ish seconds to run even with this dummy data, so writing a bunch of separate cases would balloon our run time.  Writing one super-case for now.
+      # TODO: Come back and expand the test cases in this spec once they can be run in reasonable time
+      it 'maximal results test' do
+        result = results.first
+        all_pre_concept_results = pre_activity_sessions.map(&:concept_results).flatten
+        all_post_concept_results = post_activity_sessions.map(&:concept_results).flatten
+
+        expect(result[:aggregate_rows].length).to eq(classrooms.map(&:grade).uniq.length)
+        expect(result[:skill_name]).to eq(skill_group.name)
+        expect(result[:pre_score]).to eq(all_pre_concept_results.select { |cr| cr.correct }.length.to_f / all_pre_concept_results.length * 100)
```
brendanshean: I think the following four `select { |cr| cr.correct }` can be rewritten `select(&:correct)`

services/QuillLMS/spec/queries/admin_diagnostic_reports/diagnostic_performance_by_skill_query_spec.rb
```diff
@@ -0,0 +1,149 @@
...
+        expect(result[:post_score]).to eq(all_post_concept_results.select { |cr| cr.correct }.length.to_f / all_post_concept_results.length * 100)
+        expect(result[:pre_correct_total]).to eq(all_pre_concept_results.select { |cr| cr.correct }.length)
+        expect(result[:pre_total_questions]).to eq(all_pre_concept_results.length)
+        expect(result[:post_correct_total]).to eq(all_post_concept_results.select { |cr| cr.correct }.length)
+        expect(result[:post_total_questions]).to eq(all_post_concept_results.length)
+
+        # The setup for these tests gives users 50% correct rates on their pre-diagnostic, and 100% correct rates on their post-diagnostic, which correspond to "imrpoved proficiency"
```
brendanshean: *improved

brendanshean (commented):
Great job figuring out these complex aggregations.  I also think the underlying refactor make this more readable.

services/QuillLMS/app/queries/admin_diagnostic_reports/diagnostic_aggregate_query.rb
```diff
@@ -42,16 +43,33 @@ def run_query
...
+      SQL
+    end
+
+    def with_sql
+      with_queries.map do |name, query|
+        "#{name} AS (#{query})"
+      end.join(",\n")
+    end
+
+    def with_queries
```
dandrabik, line 53: One thing that would be nice is a way to generate the SQL consistently across these classes, for debugging/optimizing purposes. For the Snapshot queries, that method was `query` and for these it's `union_query` which took me a second to find. We should consider making this `query` for all, or adding an additional method like `to_sql` that can be used consistently.

services/QuillLMS/app/queries/admin_diagnostic_reports/diagnostic_performance_by_skill_query.rb
```diff
@@ -0,0 +1,233 @@
...
+      <<-SQL
+        SELECT
+          #{aggregate_by_clause} AS aggregate_id,
+          #{aggregate_sort_clause} AS name,
+          '#{additional_aggregation}' AS group_by,
+          #{specific_select_clause}
+      SQL
+    end
+
+    def with_queries
```
dandrabik, line 56: Yeah, this is a huge query, the 15 seconds makes more sense now. We'll have to investigate ways of doing this differently (multiple queries, pre-processing data to get it in a better shape, etc.)

dandrabik (approved):
Overall the setup looks good. Now we just have to dig into the query performance. Thanks for putting this together.

PR #11294, opened 2023-11-30
fix edge cases with filtering logic

emilia-friedberg:
## WHAT
Fix two bugs in filtering logic: one where custom timeframes had been selected, but other filters had not, and one where a teacher makes filter selections and then clears them. 

## WHY
These both result in unexpected behavior with the filters, and we want them to work correctly.

## HOW
In the first case, make sure we're passing the set `customStartDate` and `customEndDate` to the `setSelectedAndLastSubmitted` function so that, if they were set in `getFilterSelections`, they aren't overwritten in `getFilters`. In the second, make sure we're comparing `selectedX` to `originalAllX`, rather than just `allX`, because `allX` can actually be a subset depending on the other selected filters (ex: if one teacher is selected, `allClassrooms` will only consist of their classrooms, while `originalAllClassrooms` will still contain every classroom under that admin's purview). Question for reviewers: does that mean we should rename `allX` in this file to something like `allXForDisplay`? What about files that receive it as a prop?

### Screenshots

### Notion Card Links
https://www.notion.so/quill/Resolve-edge-case-in-the-admin-report-filter-saving-logic-for-all-selected-81bf07e0ae3849939a06356c10485562?pvs=4

https://www.notion.so/quill/Custom-time-frames-are-not-being-saved-properly-on-the-admin-reports-21efd8057b5f477b8a0aea1d0f2a3f14?pvs=4

PR Checklist | Your Answer
------------ | -------------


anathomical (approved):
I favor the idea of renaming these variables.  The `allXForDisplay` nomenclature seems fine.  Possible alternatives could be `filteredX` or `availableX`.  I don't have strong feelings about which we pick.  That said, I usually find it useful if a passed-through property like this has the same name in as many places as possible, so if we do go for a rename, picking a name that "makes sense" in as many of the components that use it as possible may be a good way to evaluate name options.

eadams17 (approved):

emilia-friedberg:
Thanks, @anathomical. Rerequesting a review from you and @eadams17 because I did go ahead and do that rename (I liked your `availableX` suggestion).

anathomical (approved):
Looks great.  Thanks!

PR #11296, opened 2023-12-01
fix window.scrollTo test error

emilia-friedberg:
## WHAT
Fix window.scrollTo error in Jest tests

## WHY
This is one of the console errors we were trying to address, but is actually just a Jest thing, and intersects with work PK is doing with Vite (where it looks like this error causes actual failures, not just warnings).

## HOW
Just mock this behavior.

### Screenshots

### Notion Card Links
https://www.notion.so/quill/Console-log-cleanup-Not-implemented-window-scrollTo-15-7c17b5ad74b6465e897e6e458fe32669?pvs=4

PR Checklist | Your Answer
------------ | -------------
Have you added and/or updated tests? | Updated setupTests

### DIFF
--- a/services/QuillLMS/client/setupTests.ts
+++ b/services/QuillLMS/client/setupTests.ts
@@ -40,3 +40,5 @@ global.IntersectionObserver = class IntersectionObserver {
 };
 
 window.process.env = processEnvMock.env
+
+window.scrollTo = jest.fn();


anathomical (approved):
Nice, simple fix that works for all of our tests and doesn't require us to update any of them?  I love it!

happythenewsad (approved):

PR #11303, opened 2023-12-04
don't generate student usernames with spaces in them (replace with hyphens)

emilia-friedberg:
## WHAT
Update our `generate_usernames` file to replace any spaces in the student's name with hyphens.

## WHY
Shannon reported an issue where password reset was resulting in students having spaces in their passwords. I realized that this was actually due to a bug in initial username creation, where the username was being created with spaces and then any further updates to the user model failed to save because the username itself was invalid. This change should make it so that users imported from Google or Clever have valid usernames, so reset password should always work as expected. 

## HOW
Just `gsub` the name values.

### Screenshots

### Notion Card Links
https://www.notion.so/quill/Password-Reset-Message-for-Two-Word-Last-Names-Only-Showing-One-Name-ab47dddc31ae498eac32a376934efd82?d=9ac20a832c8747729ce10dce7a88319f

PR Checklist | Your Answer
------------ | -------------
Have you deployed to Staging? | NO

### DIFF
--- a/services/QuillLMS/app/services/generate_username.rb
+++ b/services/QuillLMS/app/services/generate_username.rb
@@ -4,8 +4,8 @@ class GenerateUsername < ApplicationService
   MAX_LOOPS = 1_000
 
   def initialize(first_name, last_name, classcode = nil)
-    @first_name = first_name.gsub(' ', '-')
-    @last_name = last_name.gsub(' ', '-')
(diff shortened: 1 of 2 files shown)


eadams17 (approved):
his name is my name too!

brendanshean (approved):
Whenever we go out

anathomical (approved):
The people always shout

PR #11309, opened 2023-12-06
Add PdfSubscription model

brendanshean:
## WHAT
Create a model `PdfSubscription`
title: name of pdf
frequency: 'weekly', 'monthly'
filter_selections: same as AdminReportFilterSelection model

## WHY
This will store user preferences related to periodic email of a particular PDF.

## HOW
rails generate model PdfSubscription

NB: Initially, I thought about adding a foreign key to `AdminReportFilterSelection` records but that particular record gets overwritten every time a user updates filter selections in the UI.  The specs for this require that we persist the filters with the PDF file itself.

### Screenshots

### Notion Card Links
https://www.notion.so/quill/UI-Report-Subscription-Email-subscription-for-Admin-Usage-Snapshot-Report-b0b949a1f69744ed9ba2a6fc3a1faad6?pvs=4

PR Checklist | Your Answer
------------ | -------------

### DIFF
--- a/services/QuillLMS/app/models/pdf_subscription.rb
+++ b/services/QuillLMS/app/models/pdf_subscription.rb
@@ -4,22 +4,22 @@
 #
 # Table name: pdf_subscriptions
 #
-#  id                :bigint           not null, primary key
-#  filter_selections :jsonb
-#  frequency         :string           not null
-#  title             :string           not null
-#  created_at        :datetime         not null
-#  updated_at        :datetime         not null
-#  user_id           :bigint           not null
+#  id                               :bigint           not null, primary key
+#  frequency                        :string           not null
+#  created_at                       :datetime         not null
+#  updated_at                       :datetime         not null
+#  admin_report_filter_selection_id :bigint           not null
+#  user_id                          :bigint           not null
 #
 # Indexes
 #
-#  index_pdf_subscriptions_on_frequency  (frequency)
-#  index_pdf_subscriptions_on_title      (title)
-#  index_pdf_subscriptions_on_user_id    (user_id)
+#  index_pdf_subscriptions_on_admin_report_filter_selection_id  (admin_report_filter_selection_id)
+#  index_pdf_subscriptions_on_frequency                         (frequency)
+#  index_pdf_subscriptions_on_user_id                           (user_id)
 #
 # Foreign Keys
 #
+#  fk_rails_...  (admin_report_filter_selection_id => admin_report_filter_selections.id)
 #  fk_rails_...  (user_id => users.id)
 #
 class PdfSubscription < ApplicationRecord
@@ -28,13 +28,8 @@ class PdfSubscription < ApplicationRecord
     MONTHLY = 'monthly'
   ]
 
-  TITLES = [
-    ADMIN_USAGE_SNAPSHOT_REPORT = 'admin_usage_snapshot_report'
-  ]
-
   belongs_to :user
+  belongs_to :admin_report_filter_selection
 
   validates :frequency, presence: true, inclusion: { in: FREQUENCIES }
-  validates :title, presence: true, inclusion: { in: TITLES }
-  validates :user_id, presence: true
 end
--- a/services/QuillLMS/app/models/pdf_subscription.rb
+++ b/services/QuillLMS/app/models/pdf_subscription.rb
@@ -9,18 +9,11 @@
 #  created_at                       :datetime         not null
 #  updated_at                       :datetime         not null
 #  admin_report_filter_selection_id :bigint           not null
-#  user_id                          :bigint           not null
 #
 # Indexes
 #
 #  index_pdf_subscriptions_on_admin_report_filter_selection_id  (admin_report_filter_selection_id)
 #  index_pdf_subscriptions_on_frequency                         (frequency)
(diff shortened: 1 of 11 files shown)


emilia-friedberg (approved):
LGTM! Too bad about the lack of reusability for the `AdminReportFilterSelections`. I guess one option would be to add a new report type for that table (`usage_snapshot_report_pdf` or something) and go ahead with your foreign key idea in reference to that record, which wouldn't keep getting overwritten, but I'm not sure that's really better than what you have here.

services/QuillLMS/app/models/pdf_subscription.rb
```diff
@@ -0,0 +1,40 @@
...
+
+  TITLES = [
+    ADMIN_USAGE_SNAPSHOT_REPORT = 'admin_usage_snapshot_report'
+  ]
+
+  belongs_to :user
+
+  validates :frequency, presence: true, inclusion: { in: FREQUENCIES }
+  validates :title, presence: true, inclusion: { in: TITLES }
+  validates :user_id, presence: true
```
anathomical: It may actually be worth adding a validation for the presence of `filter_selections` because, somewhat deceptively, even a user who hasn't selected any specific filters (thus using the default) will have at least an array of `school_ids` along with `timeframe_start` and `timeframe_end` set for them.  So we should always have some filter value for these.

anathomical (approved):
Minor suggestion about additional validation given that we should always have some filter values for these reports.

anathomical:
For what it's worth, I like @emilia-friedberg's suggestion of creating a new "report type" and maybe being able to re-use the existing filters table assuming that it's not too complicated to do that.

brendanshean:
> LGTM! Too bad about the lack of reusability for the `AdminReportFilterSelections`. I guess one option would be to add a new report type for that table (`usage_snapshot_report_pdf` or something) and go ahead with your foreign key idea in reference to that record, which wouldn't keep getting overwritten, but I'm not sure that's really better than what you have here.



> For what it's worth, I like @emilia-friedberg's suggestion of creating a new "report type" and maybe being able to re-use the existing filters table assuming that it's not too complicated to do that.

Unless I'm missing something, that approach doesn't solve the overwrite issue which highlights that this a different use case.

services/QuillLMS/app/models/pdf_subscription.rb
brendanshean: I'll defer to you and @emilia-friedberg on this question.  Also, does it make sense to add a similar validation for `AdminReportFilterSelections`

emilia-friedberg:
> > LGTM! Too bad about the lack of reusability for the `AdminReportFilterSelections`. I guess one option would be to add a new report type for that table (`usage_snapshot_report_pdf` or something) and go ahead with your foreign key idea in reference to that record, which wouldn't keep getting overwritten, but I'm not sure that's really better than what you have here.
> 
> > For what it's worth, I like @emilia-friedberg's suggestion of creating a new "report type" and maybe being able to re-use the existing filters table assuming that it's not too complicated to do that.
> 
> Unless I'm missing something, that approach doesn't solve the overwrite issue which highlights that this a different use case.

The idea is that you'd add a new report type that wouldn't be subject to getting overwritten (only the existing ones get overwritten, something like `usage_snapshot_report_pdf` wouldn't be).

brendanshean:
> > > LGTM! Too bad about the lack of reusability for the `AdminReportFilterSelections`. I guess one option would be to add a new report type for that table (`usage_snapshot_report_pdf` or something) and go ahead with your foreign key idea in reference to that record, which wouldn't keep getting overwritten, but I'm not sure that's really better than what you have here.
> > 
> > 
> > > For what it's worth, I like @emilia-friedberg's suggestion of creating a new "report type" and maybe being able to re-use the existing filters table assuming that it's not too complicated to do that.
> > 
> > 
> > Unless I'm missing something, that approach doesn't solve the overwrite issue which highlights that this a different use case.
> 
> The idea is that you'd add a new report type that wouldn't be subject to getting overwritten (only the existing ones get overwritten, something like `usage_snapshot_report_pdf` wouldn't be).

On second thought, that does sound better.  Good thinking!

services/QuillLMS/db/migrate/20231206143410_create_pdf_subscriptions.rb
```diff
@@ -3,9 +3,8 @@
 class CreatePdfSubscriptions < ActiveRecord::Migration[7.0]
   def change
     create_table :pdf_subscriptions do |t|
-      t.string :title, index: true, null: false
       t.string :frequency, index: true, null: false
-      t.jsonb :filter_selections
+      t.references :admin_report_filter_selection, index: true, foreign_key: true, null: false
```
anathomical: We've been shying away from foreign keys due to concerns about migrations and performance, but that's mostly been around larger tables like User. I imagine that this one will not pie significant risks since it will only ever service admins.

anathomical (approved):

services/QuillLMS/db/migrate/20231206143410_create_pdf_subscriptions.rb
brendanshean: Foreign keys have been removed.

PR #11310, opened 2023-12-06
admin usage snapshot report pdf styling and some config

emilia-friedberg:
## WHAT
Add HTML, CSS, and some config to get the PDF version of the Usage Snapshot Report rendering (statically) as expected.

## WHY
We want these PDFs to be pretty.

## HOW
Largely copied over HTML structure from live version, with additions and changes where necessary to reflect Figma and PDF constraints. Then write CSS, again based largely on the extant CSS file, but removing extraneous code. I initially wrote this as plain CSS when I had it just in a `<style>` tag in the `<head>`, but once I managed to get the stylesheet import working I realized it still worked as SCSS, which I prefer, so I used ChatGPT to convert it back. If this becomes a problem down the line, the regular CSS file is in commit history so we can go back to that. I also got the typekit import working.

Note: the command needs to be run like this:

```
Grover.configure do |config|
  config.options = {
    print_background: true,
    format: 'Letter',
    margin: {
      top: '50px',
      bottom: '50px'
    },
  }  
end

reload!; File.open('file.pdf', 'wb') { |file| file << Grover.new(ApplicationController.new.render_to_string(template: 'pages/pdfz', layout: 'pdf', locals: {})).to_pdf }
```

### Screenshots
[file.pdf](https://github.com/empirical-org/Empirical-Core/files/13588712/file.pdf)

### Notion Card Links
https://www.notion.so/quill/Placeholder-CSS-Design-of-PDF-report-abbcbd46804b4d9096d4b289f74c9f2e?pvs=4

PR Checklist | Your Answer
------------ | -------------
Have you added and/or updated tests? |  Manually testsed
Have you deployed to Staging? | NO - no great way to test on staging because the PDF generation isn't hooked up to anything yet


services/QuillLMS/app/views/layouts/pdf.html.erb
```diff
@@ -3,8 +3,10 @@
   <head>
     <title>Quill</title>
     <meta content='text/html; charset=UTF-8' http-equiv='Content-Type'/>
+    <link rel="stylesheet" href=<%= "#{ENV['DEFAULT_URL']}/assets/admin_usage_snapshot_report_pdf.scss" %>>
```
emilia-friedberg, line 6: I couldn't get this working as a relative path -- I don't think this should be an issue, but if it is I can mess around more or go back to using a `<style>` tag.

brendanshean: I remember reading something about relative paths here:  https://github.com/Studiosity/grover#relative-paths

I wonder if that is the issue.

PR #11320, opened 2023-12-07
Lms/admin diagnostic show only assigned

anathomical:
## WHAT
Update Admin Diagnostic queries that calculate completed activities to only count ActivitySession records that correspond to students that are currently in the associated `ClassroomUnit.assigned_student_ids`

I also made a small tweak to the `ActivitySession` factory so that when it adds a user to a provided `ClassroomUnit`'s `assigned_student_ids`, it appends that user instead of wiping out any existing users already assigned.
## WHY
Students may be assigned an activity, complete it, and then later leave the classroom (resulting in them being unassigned).  On Teacher reports, this excludes their completion from reporting, so we want to do that for admins as well.  (This also prevents us from having more people complete an activity than are assigned to it.)
## HOW
Add an additional conditional when JOINing relevant completed `ActivitySession` records for rollup in admin reports so that `ActivitySession` records for students who are not currently assigned to the `ClassroomUnit` will be excluded

### Notion Card Links
https://www.notion.so/quill/Admin-Diagnostic-Growth-Report-QA-Overview-Tab-December-7-0524fc96c41f424a9532228535b11553#3c9299885a674ad390701f30a01039e8

PR Checklist | Your Answer
------------ | -------------


services/QuillLMS/spec/queries/admin_diagnostic_reports/diagnostic_recommendations_query_spec.rb
```diff
@@ -64,6 +64,17 @@ module AdminDiagnosticReports
         it { expect(results).to eq([]) }
       end
 
+      context 'finished activity session from a user who is no longer assigned to the classroom unit' do
+        # The ActivitySession factory aggressively attaches students to ClassroomUnit.assigned_student_ids, so it's easier to manually remove them than to try to override a `let` call for this condition
+        before do
+          # Make sure all factories run so that the change we make below doens't get overwritten
```
brendanshean: *doesn't

services/QuillLMS/spec/queries/admin_diagnostic_reports/post_diagnostic_completed_query_spec.rb
```diff
@@ -42,6 +42,17 @@ module AdminDiagnosticReports
         it { expect(results).to eq([]) }
       end
 
+      context 'finished activity session from a user who is no longer assigned to the classroom unit' do
+        # The ActivitySession factory aggressively attaches students to ClassroomUnit.assigned_student_ids, so it's easier to manually remove them than to try to override a `let` call for this condition
+        before do
+          # Make sure all factories run so that the change we make below doens't get overwritten
```
brendanshean: *doesn't

services/QuillLMS/spec/queries/admin_diagnostic_reports/pre_diagnostic_completed_query_spec.rb
```diff
@@ -43,6 +43,17 @@ module AdminDiagnosticReports
         it { expect(results).to eq([]) }
       end
 
+      context 'finished activity session from a user who is no longer assigned to the classroom unit' do
+        # The ActivitySession factory aggressively attaches students to ClassroomUnit.assigned_student_ids, so it's easier to manually remove them than to try to override a `let` call for this condition
+        before do
+          # Make sure all factories run so that the change we make below doens't get overwritten
```
brendanshean: *doesn't

services/QuillLMS/spec/queries/admin_diagnostic_reports/diagnostic_recommendations_query_spec.rb
brendanshean: You could introduce a transient attribute in the ActivitySession factory to skip the before and after callbacks.  I'm not sure if the other items in the callbacks are necessary though.

```
...
    transient { skip_classroom_unit_assignment { true } }

    before(:create) do |activity_session, evaluator|
      unless evaluator.skip_classroom_unit_assignment
        ...rest of block
      end
    end

    after(:create) do |activity_session, evaluator|
      unless evaluator.skip_classroom_unit_assignment
        ..rest of block
      end
    end
```

Then you instantiate: `create(:activity_session, skip_classroom_unit_assignment: true)`

This would allow for a reusable piece later.  Whereas future uses of the other pattern have the long comment.

brendanshean (approved):
Nice job on figuring out this tricky case.

services/QuillLMS/spec/queries/admin_diagnostic_reports/diagnostic_recommendations_query_spec.rb
brendanshean: Related note:  I feel like this factory in general is trying to do much which is resulting in this complex behavior.

dandrabik (approved):
:+1: Thanks for fixing this up.

services/QuillLMS/spec/queries/admin_diagnostic_reports/diagnostic_recommendations_query_spec.rb
anathomical: This was a great suggestion.  @brendanshean and I paired up to try to implement it, but ran into an issue where `ActivitySession` has a `on: :create` validator that checks to make sure that the `ActivitySession.user`'s id is in the `ActivitySession.classroom_unit.assigned_student_ids` array, and fails if it's not.  This must be why this `before(:create)` even exists in the `ActivitySession` factory, but also means that without significant refactoring we can't just skip this step in `ActivitySession` creation.  We'll revisit at a later date.

happythenewsad:
@anathomical The 'approve' button is not working for me at the moment, so consider this comment an approval

anathomical:
Closing this as we realized that there were additional discrepancies that have been solved through a combination of data clean-up and other code changes.

PR #11322, opened 2023-12-08
Generate Admin Usage Snapshot Report PDF

brendanshean:
## WHAT
Generate the Admin Usage Snapshot Report PDF with fake data.

## WHY
Eventually, we'll be sending this PDF to Admins with real data.

## HOW
1.  Generate the HTML
2.  Convert relative links to absolute ones
3.  Send to preprocessed HTML to Grover 
4.  Have Grover convert to PDF.

NB:  In previous versions of this test that are on develop, I had the Application.new.render_to_string actually render the template.  However, with the addition of scss files, and a 

```  <%= stylesheet_link_tag "admin_usage_snapshot_report_pdf", media: "all" %>```

the asset pipeline, requires all assets to be compiled ahead of time which resulted in a new error:
Screenshot 2023-12-11 at 2 15 00 PM

I think this type of error hasn't come up before in other specs because 
a) we don't have stylesheet_include_tag in the template
b) we don't explicitly test the the html.erb file.

Long term we could add npm builds into the circleci configuration for rails tests (or even extract the PDF testing into its own configuration along with some snapshot testing), but at this point it seems kind of heavy for one use case.

### Screenshots

### Notion Card Links
https://www.notion.so/quill/UI-Report-Subscription-Email-subscription-for-Admin-Usage-Snapshot-Report-b0b949a1f69744ed9ba2a6fc3a1faad6?pvs=4

PR Checklist | Your Answer
------------ | -------------

### CODE REVIEW

File: services/QuillLMS/app/assets/stylesheets/admin_usage_snapshot_report_pdf.scss [hunk 1]
```diff
@@ -0,0 +1,354 @@
...
+      font-size: 24px;
+      margin-bottom: 32px;
+    }
+  }
+
+  /* Applied Filters Styles */
+  .applied-filters {
+    padding: 20px;
+    border-radius: 8px;
+    border: 1px solid #E8AB88;
```
emilia-friedberg, line 26: I just realized that, now that this file exists on its own and is scss, we could probably import the actual color variables here. I'm going to make a follow-up card so I can make those changes when I have time.

emilia-friedberg (approved):

services/QuillLMS/app/assets/stylesheets/admin_usage_snapshot_report_pdf.scss
brendanshean: Good idea.  I have a follow-up PR to this one, so I can try adding it in there.

services/QuillLMS/app/workers/blah_worker.rb
```diff
@@ -0,0 +1,32 @@
+# frozen_string_literal: true
+
+class BlahWorker
```
dandrabik, line 3: Name? `AdminUsageSnapshotReportPdfUploadWorker` if you follow the naming below? Or something more general if it's reusable.

dandrabik (approved):
Overall, this looks good for getting the general structure in place. I'm curious what will change when we are generating the page from data and finalizing, e.g. is it worth having Grover replace links or should we just manual links, what goes in the layout vs. the view, do these global margin stay, is there more [config](https://github.com/Studiosity/grover?tab=readme-ov-file#configuration) we will use, etc.

It's great to see this progress! Thanks for moving this along!

services/QuillLMS/app/workers/blah_worker.rb
brendanshean: I have this renamed in a follow-up PR.  The BlahWorker is called anywhere as of yet.

PR #11337, opened 2023-12-12
add post query transform hook to QuillBigQuery::Query

happythenewsad:
## WHAT / HOW
1. Adds insertion point for post query transforms to the base QuillBigQuery::Query class
2. Adds a particular DateTime transform to DateExportQuery 

## WHY
1. So that there's a standard way to add post query transforms in the future
2. To satisfy the specification of this user story 

## DISCUSSION

An alternative is to keep transforms completely separate from queries, like this:
```
result = DataExportQuery.run(...)
transformed_result = DataExportQueryTransformer.run(result)
``` 

I opted to add an insertion point into ::Query because:

1. It keeps transforms and queries together, conceptually and file-wise 
2. It allows unit testing of transforms, avoiding BigQuery data setup and network calls
3. It's extensible for future transforms, and 
4. You don't have to remember to call a transform; it's there by default. (It follows that transforms following this pattern should apply to every instance of the given query. One-off transforms should live elsewhere).

### Screenshots

### Notion Card Links
https://www.notion.so/quill/Take-an-admin-s-time-zone-setting-into-account-when-populating-the-Completed-date-value-in-the-adm-1d3785111c4b4e748b47bf77ff8d5d43

PR Checklist | Your Answer
------------ | -------------

### DIFF
--- a/services/QuillLMS/app/queries/quill_big_query/query.rb
+++ b/services/QuillLMS/app/queries/quill_big_query/query.rb
@@ -12,8 +12,12 @@ module QuillBigQuery
       raise NotImplementedError
     end
 
+    def post_query_transform(query_result)
+      query_result
+    end
+
     def run_query
-      runner.execute(query)
+      post_query_transform(runner.execute(query))
     end
 
     def query
--- a/services/QuillLMS/app/queries/snapshots/data_export_query.rb
+++ b/services/QuillLMS/app/queries/snapshots/data_export_query.rb
@@ -29,6 +29,15 @@ module Snapshots
       SQL
     end
 
+    def post_query_transform(query_result)
+      return query_result unless user&.time_zone
+      query_result.map do |row|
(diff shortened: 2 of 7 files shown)


services/QuillLMS/app/queries/quill_big_query/query.rb
```diff
@@ -12,8 +12,12 @@ def run
       raise NotImplementedError
     end
 
+    def post_query_transform(query_result)
+      query_result
+    end
+
     def run_query
-      runner.execute(query)
+      post_query_transform(runner.execute(query))
```
brendanshean, line 20: I like standardizing the post_query_transform in one place.  For this reason, I think it would be a good idea to update Thomas'  `post_process` method [here](https://github.com/empirical-org/Empirical-Core/blob/develop/services/QuillLMS/app/queries/admin_diagnostic_reports/diagnostic_aggregate_query.rb#L37) with your pattern, that way the approach has some momentum.

brendanshean (approved):
Nice work on this.

PR #11344, opened 2023-12-13
Generate SQL for Snapshot Queries

dandrabik:
## WHAT
This is a helper task to generate the SQL for the Snapshot Queries
## WHY
We need to analyze these queries for performance, and since they are compiled from code, the underlying SQL is a bit obscured. This allows us to run queries in the BigQuery console and iterate quickly on potential changes.
## HOW
I used `Thor` instead of `Rake` (thor is included with Rails these days), since I prefer its syntax (and to demo it), but basically just loop through the queries and called `.query` and write the output to a file in a folder.

I committed one folder for a test user, since I think it will be helpful to have some shared queries for debugging purposes. I'm not sure if these will stay in the repo long term.

PR Checklist | Your Answer
------------ | -------------
Have you added and/or updated tests?| No, it's an offline developer task


happythenewsad (approved):
Cool!

services/QuillLMS/lib/query_examples/snapshots/9874030-2023-08-01-2023-12-01/active-classrooms.sql
```diff
@@ -0,0 +1,26 @@
+      /* Data Processed By Query: 0.53 GB */
```
happythenewsad, line 1: +1 to gitignoring these files, eventually. They take up space and will eventually drift from what's in production.

brendanshean (approved):
This seems very useful.  Good call on using Thor which I also prefer to rake.

PR #11348, opened 2023-12-13
AGDR/ configure Skills frontend

eadams17:
## WHAT
configure Skills section of the Admin Diagnostic Growth Reports to receive backend data; fix various issues with the Overview section

## WHY
this is part 2 of 3 of the AGDR project

## HOW
- hook into skills API, format data and tweak logic to handle additional filters
- this also fixes some issues with Pusher for the Overview section as well as adds some additional testing

### Screenshots
Screen Shot 2023-12-13 at 5 22 04 PM

### Notion Card Links
https://www.notion.so/quill/Admin-Diagnostic-Growth-Report-Skills-Frontend-948eecede33144ce90e6222025f35bad?pvs=4
https://www.notion.so/quill/UI-Issues-Admin-Reports-Diagnostic-Growth-Report-ea6fdd1ec7714fea9e8548e9de8732c0?pvs=4

PR Checklist | Your Answer
------------ | -------------


services/QuillLMS/client/app/bundles/PremiumHub/containers/__tests__/diagnosticGrowthReportsHelpers.test.tsx
```diff
@@ -167,7 +167,7 @@ const mockCombinedData = [{
       averageActivitiesAndTimeSpent: "22 Activities (26:56)",
       id: "5",
       name: "Grade 5",
-      overallSkillGrowth: <p className="emphasized-content">+29%</p>,
+      overallSkillGrowth: <button className="interactive-wrapper emphasized-content" value={1663}>+29%</button>,
```
emilia-friedberg: Why are these buttons with no `onClick` handler?

services/QuillLMS/client/app/bundles/PremiumHub/containers/__tests__/overviewSection.test.tsx
```diff
@@ -37,23 +100,71 @@ describe('OverviewSection', () => {
...
+      expect(screen.getByRole('cell', { name: /17 of 35 students/i })).toBeInTheDocument()
+      expect(screen.getByRole('button', { name: /\+10%/i })).toBeInTheDocument()
+
+      expect(screen.getByRole('cell', { name: /grade 9/i })).toBeInTheDocument()
+      expect(screen.getByRole('cell', { name: /30 of 30 Students/i })).toBeInTheDocument()
+      expect(screen.getByRole('cell', { name: /9 Students/i })).toBeInTheDocument()
+      expect(screen.getByRole('cell', { name: /23 activities \(26:35\)/i })).toBeInTheDocument()
+      expect(screen.getByRole('cell', { name: /10 of 30 students/i })).toBeInTheDocument()
+      expect(screen.getByRole('button', { name: /\+30%/i })).toBeInTheDocument()
+    })
```
emilia-friedberg: I appreciate the thoroughness of the tests both here and in the following file, but it kind of feels to me like a snapshot might actually be more useful here --if something were to break here, I think it would be hard to figure out what had happened (as opposed to a snapshot when you can actually see the whole HTML structure). I don't think we need to remove what you have here, but maybe add snapshots too? Also, spacing between the tests would help with readability.

services/QuillLMS/client/app/bundles/PremiumHub/containers/diagnosticGrowthReports/overviewSection.tsx
```diff
@@ -90,33 +90,51 @@ export const OverviewSection = ({
...
 
   React.useEffect(() => {
     if (!pusherMessage) return
 
-    if (filtersMatchHash(pusherMessage)) getData()
+    if (filtersMatchHash(pusherMessage) && !updatedFromPusher) {
+      // this prevents getData from being called multiple times since we receive Pusher messages for each of the 5 different API calls
+      setUpdatedFromPusher(true)
+      getData()
+    }
```
emilia-friedberg: I'm confused about how this works -- what if one of the queries is ready way before the other ones? Won't this always just fire when the first query comes back, potentially leaving the user in a state where some of the page has no data?

emilia-friedberg (commented):
So excited to see this so close to the finish line! Just a couple of things I'm unclear about here.

services/QuillLMS/client/app/bundles/PremiumHub/containers/__tests__/diagnosticGrowthReportsHelpers.test.tsx
eadams17: good catch, I'll add those now

services/QuillLMS/client/app/bundles/PremiumHub/containers/__tests__/overviewSection.test.tsx
eadams17: that's a good point-- I'll add snapshots and add some spacing to break up the tests

services/QuillLMS/client/app/bundles/PremiumHub/containers/diagnosticGrowthReports/overviewSection.tsx
eadams17: I tested this with multiple different filter changes and it seems to be working as expected. We only really need to worry about matching on one of the five hashes returned to know that we need to refetch data for the newly applied filters. The queries run in parallel and need to be run all together regardless of any change, so we only need to worry about running `getData` once and setting this value to prevent unnecessary subsequent calls.

Further down, we do a check for change in value for `hasAdjustedFiltersSinceLastSubmission` to set `updatedFromPusher` back to false if the filters have been adjusted since the last submission, which will then enable calling `getData` again. 

Does that make sense? We can chat more on Slack if you'd like, I know this is a bit complicated to follow. This seemed like the most elegant solution, but I'd be open to other ideas if you think we should try another route.

emilia-friedberg: I'm curious about @anathomical's take here, but I'm still concerned that this implementation is going to lead, maybe very infrequently, to a case where `getData` is called for the second time before all of queries are actually finished running. I think it would be better to wait until the last, not the first of them, has its `pusherMessage` fired (something I think you'd have to do by tracking each of them individually in the state, like `preDiagnosticAssignedQueryFinished`, `preDiagnosticCompletedQueryFinished`, etc, and don't call `getData` again until all five are `true`). I agree that the change would be more complicated code, but I think that's probably worth the trade-off for preventing that edge case. Maybe they all run so fast, or at such a similar speed, that we can be reasonably confident that case will never occur in production, but then we should at least leave a comment to here to that effect because the code itself does not ensure that won't happen.

eadams17: @emilia-friedberg I have an idea. In each of the 5 functions that call a query, I can initially reset their respective pieces of state back to null, like this:

```
  function getPreDiagnosticAssignedData () {
    **setPreDiagnosticAssignedData(null)**
    const searchParams = getSearchParams(PRE_DIAGNOSTIC_ASSIGNED_QUERYSTRING)

    requestPost('/admin_diagnostic_reports/report', searchParams, (body) => {
      if (!body.hasOwnProperty('results')) {
        return
      } else {
        const { results, } = body
        setPreDiagnosticAssignedData(results)
      }
    })
  }
  ```
  
which will ensure that this useEffect accounts for this after initial load (for each query, the result value will always be either an array of values or just an empty array):

```
  React.useEffect(() => {
    if (preDiagnosticAssignedData && postDiagnosticAssignedData && preDiagnosticCompletedData && postDiagnosticCompletedData && recommendationsData) {
      aggregateOverviewData({
        preDiagnosticAssignedData,
        postDiagnosticAssignedData,
        preDiagnosticCompletedData,
        postDiagnosticCompletedData,
        recommendationsData,
        setAggregatedData,
        hasAdjustedFiltersFromDefault,
        handleSetNoDiagnosticDataAvailable,
        setLoading
      })
    }
  }, [preDiagnosticAssignedData, postDiagnosticAssignedData, preDiagnosticCompletedData, postDiagnosticCompletedData, recommendationsData])
  ```

anathomical: This is a good flag from @emilia-friedberg .  I think that the real risk is that there's some hidden complexity on the back-end.  When you make a request for data to an endpoint on the back-end, one of two things happen: 1) the data is in the cache, and you get it back or 2) the data is not in the cache which starts a Sidekiq job to put the data into the cache and then send a Pusher message to let the front-end know that the data is ready.  Consider this scenario:

1. Iniital page load: `getData` triggers 5 separate API calls.  Assume that the cache is empty, so all 5 queries kick off Sidekiq jobs.  Assume that the jobs take 1, 2, 3, 4, and 5 seconds to complete respectively.
2. The first job completes at 1 second, and sends a Pusher messages.  `getData` gets called, making 5 new API calls.  The first query, the one that sent the Pusher notification, has data in the cache, so that API call returns data.  Let's assume that somewhere in the time between Pusher sending a message and everything getting processed, the second Sidekiq job has completed, so that API call also returns data from cache.  The next three Sidekiq jobs have not completed, so we actually enqueue 3 Sidekiq jobs to repeat the work for those 3 bits of data.
3. Another Sidekiq jobs completes and sends a Pusher notification, so we call `getData` again.  The first and second job have data in cache, and now a third job (the one that just sent a Pusher message) has completed, so it's in cache, too.  But again some new jobs can get enqueued since they're still running.

Ultimately, I think this points to, at the very least, value in tracking Pusher notifications for each `queryString` separately and, as Emilia suggests, not re-calling `getData` until all five have returned.  There are a couple approaches to this, but I think that the easiest way to do so is to have a set of five new state values.  `preCompletedQueryReadyForRefresh` or something similar.  Crib the code from `filterMatchesHash` in an `Effect` that triggers on `pusherMessage` update to set whichever state value is appropriate.  Add a new effect that listens to all five of those states, and calls `getData` if all 5 are set.  And, per your suggestion above, just re-set those states to `false` or `null` whenever you call `getData`.  I think that covers most of the issues?

emilia-friedberg: Another option that just occurred to me, which I think is similar to what both you and @anathomical just suggested but gets around having to add five new state values, is to reset the state pieces like you have in your example and add that new effect and then do something like this to your `pusherMessage` hook:

```
  React.useEffect(() => {
    if (!pusherMessage) return

    switch(pusherMessage) {
      case getFilterHash(PRE_DIAGNOSTIC_ASSIGNED_QUERYSTRING):
        getPreDiagnosticAssignedData()
        break;
      case getFilterHash(PRE_DIAGNOSTIC_COMPLETED_QUERYSTRING):
        getPreDiagnosticCompletedData()
        break;
      [so on and so forth for the other possible filter hashes]
    }
  }, [pusherMessage])
```

Basically, if we just isolate the individual calls instead of calling `getData` all together, I think we'll minimize unnecessary extra calls while also ensuring that the data we have on the page is current.

eadams17: Thanks for the thorough input, y'all! @emilia-friedberg I think that's an excellent solution. I just made this change and everything seems to be working as expected

emilia-friedberg (approved):
Yay! So glad that worked. LGTM!

anathomical (approved):
Yeah, I like the solution you have here.

PR #11354, opened 2023-12-14
Add Question Counts to Activity Table For Queries

dandrabik:
## WHAT
Add Question Counts to Activity Table
## WHY
We want to make queries for “sentences written” to be less intensive in BigQuery. Right now that query requires hitting the `concept_result` table, which is very inefficient (slow and expensive).
## HOW
Add a QuestionCounter service object, since there is some custom logic for each type of activity. Add a column on activites to store this calculation. There's no functionality around this yet, since it's mainly for data queries.

### Notion Card Links
https://www.notion.so/quill/Store-Activity-Question-Count-in-DB-3958b13fd5ab40558688c7b93f57d326

Added a query for the [data on staging](https://data.quill.org/question/1109-activity-question-counts)

Screenshot 2023-12-14 at 4 02 03 PM

PR Checklist | Your Answer
------------ | -------------


services/QuillLMS/spec/services/question_counter_spec.rb
```diff
@@ -0,0 +1,88 @@
...
+
+  context 'grammar' do
+    let(:activity) { build(:grammar_activity, data: {questions: question_list}) }
+
+    it { expect(subject).to eq question_list.size }
+
+    context 'blank question list, use concept list' do
+      let(:activity) { build(:grammar_activity, data: {questions: [], concepts: concept_list}) }
+
+      it { expect(subject).to eq 8 }
```
emilia-friedberg: Do you want `concept_list.size` here?

emilia-friedberg (approved):
LGTM! Appreciate the thorough tests.

services/QuillLMS/spec/services/question_counter_spec.rb
dandrabik: It's not the size, but the sum of "quantities" in the sub objects. But taking this feedback, I've attempted to make the test more explicit.

services/QuillLMS/app/services/question_counter.rb
```diff
@@ -0,0 +1,53 @@
...
+    evidence: 3
+  }
+  DEFAULT = 10
+
+  def initialize(activity)
+    @activity = activity
+    @tool_key = activity&.classification&.key&.to_sym
+  end
+
+  def run
```
brendanshean, line 32: Are the `TOOL_CALCULATOR` lambdas meant to be called outside of the service object? If not , I would probably get rid of the structure and make them all private instance methods.

```
  def run
    tool_calculators || TOOL_DEFAULTS[tool_key] || DEFAULT
  end

  private def tool_calculators
    case tool_key
    when :passage then passage_calculator
    when :sentence then grammar_calculator
    when [:connect, :diagnostic] then question_array_size
    when :lessons then lessons_calculator
    when :evidence then evidence_calculator
    end
  end

  # move TOOL_CALCALATOR lambdas into private methods.
```

This eliminates the parameter passing of the activity and tightens up the access to the object at large.

brendanshean (approved):
This seems like a smart column to precompute.

services/QuillLMS/app/services/question_counter.rb
dandrabik: Hmm, that's interesting feedback. I personally find the original code easier to read/maintain, so I think I'm going to leave that in place on a personal preference basis. I see the upside of what you are proposing, but, to be honest, if I was reviewing the proposed code, I'd actually suggest you refactor it to how I have it (haha, which I why I think this is purely a 'personal preference' here, your code is perfectly fine). 

If you are curious why I prefer the original approach:
1. I really enjoy code that allows you to add functionality by just updating a config and not touching any of the logic that calls it (a version of following the [open-closed principle](https://www.freecodecamp.org/news/open-closed-principle-solid-architecture-concept-explained/) in SOLID OOO). This is often achieved by extracting mappings from the code, essentially separating the mapping and the lookup logic (a case statement combines the two). Sandi Metz briefly touches on this in [this refactoring talk](https://youtu.be/8bZh5LMaSmE?feature=shared&t=1902).
2. I actively dislike `case/switch` statements and almost always try to replace them with a mapping config. Some reasons: they do two things at once (define a config, define lookup logic), order matters (when you usually don't want it to), the syntax is subtle and can lead to non obvious bugs (^ the code above has a bug which would have been found in testing, it should be  `when :connect, :diagnostic` not `when [:connect, :diagnostic]` ), and I find them harder to parse than maps (although the when/then on a single line is a better than the multiline version).

brendanshean: I also dislike case/switch statements and to use them as a bridge to more objected oriented pattern.  I kind of made a half-assed suggestion here instead of actually suggesting what I really was thinking.  There seems to be some code here concerning subtypes that:
Screenshot 2023-12-15 at 1 47 00 PM
could lead to some architectural refactoring of the activity model.

Introduce activity subtypes for  passage:, sentence:, connect:, diagnostic:, lessons:, evidence: and then have the tool calculator logic and default logic within each subtype.

In any event, that's probably overkill for this story and I would agree with your points on open-closed being ideal with the lamdas.

dandrabik: Yeah, there's definitely subtypes within Activity that could be coded differently. It would be a bit of a refactor, but it might be something we benefit from one day.

PR #11363, opened 2023-12-18
Snapshot Sentences Written Query - use activity.question_count

dandrabik:
## WHAT
Use the new activity.question_count for sentences written.
## WHY
It's much more efficient to calculate
## HOW
Update query and tests. Still needs to be QA'd to make sure this metric works well (my initial QA shows the numbers are very close to the previous calculation.

I also added thor task to print out an individual query, since that was helpful for testing.
### Screenshots
Screenshot 2023-12-18 at 1 18 55 PM

### Notion Card Links
https://www.notion.so/quill/Use-Activity-question_count-for-sentences-writen-f5ae726c8875404ab3e1908de11e97f7

PR Checklist | Your Answer
------------ | -------------
Have you added and/or updated tests? |   'YES'
Have you deployed to Staging? |  YES, this is on my staging env.

### DIFF
--- a/services/QuillLMS/lib/tasks/queries.thor
+++ b/services/QuillLMS/lib/tasks/queries.thor
@@ -45,7 +45,7 @@ class Queries < Thor
 
   # user_id 9874030 is a test user
   # e.g. bundle exec thor queries:print_snapshot_sql 9874030 'sentences-written'
-  desc 'print_snapshot_sql user_id query_key', 'Output .sql fils for all snapshots for a user'
+  desc 'print_snapshot_sql user_id query_key', 'Print the SQL and data usage for a snaphot query to the console'
   def print_snapshot_sql(user_id, query_key, start_time = DEFAULT_START, end_time = DEFAULT_END)
 
     timeframe_start = DateTime.parse(start_time)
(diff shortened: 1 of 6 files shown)


brendanshean (approved):
Nice query optimization! :racing_car:

services/QuillLMS/lib/tasks/queries.thor
```diff
@@ -43,6 +43,26 @@ class Queries < Thor
     end
   end
 
+  # user_id 9874030 is a test user
+  # e.g. bundle exec thor queries:print_snapshot_sql 9874030 'sentences-written'
+  desc 'print_snapshot_sql user_id query_key', 'Print the SQL and data usage for a snaphot query to the console'
```
anathomical, line 48: Really like the idea of having this tool and building it into our regular processes to track performance and improvements.

anathomical (approved):
I would expect any discrepancies between the old query and the new query to be from cases where an Activity changed the number of questions it had at some point in history, and the `concept_results` approach counting whatever the number was at the time of completion, while the new approach just assumes that all completions were of the current count.  That seems like a reasonable approximation to me.

PR #11378, opened 2023-12-21
add mailer previews

happythenewsad:
## BACKGROUND / WHY
[Mailer Preview is a Rails subsystem](https://guides.rubyonrails.org/action_mailer_basics.html#previewing-emails) that renders emails in-browser, as an email client would. This allows fast email render debugging, since
- You have access to the full dev tool suite of the browser
- You can simply refresh the email display, rather than sending an email to yourself 

We have a stub directory of mailer previews, but it hasn't been touched since 2015, and it wasn't working, likely due to Rails upgrades since 2015. 

## WHAT 
- fix mailer preview configuration, so that it works
- move the mailer previews from `app/lib` (not standard) to `spec/` (standard)
- add email previews for `rollup` and  `ell_starter_diagnostic_info_email` email templates 

### Screenshots
Screenshot 2023-12-21 at 12 23 39 PM
Screenshot 2023-12-21 at 12 23 33 PM

### Notion Card Links

PR Checklist | Your Answer
------------ | -------------
Have you added and/or updated tests? |  no - spec code only
Have you deployed to Staging? | no

### DIFF
--- a/services/QuillLMS/spec/mailers/previews/rollup_mailer_preview.rb
+++ b/services/QuillLMS/spec/mailers/previews/rollup_mailer_preview.rb
@@ -1,0 +1,15 @@
+# frozen_string_literal: true
+
+  class RollupMailerPreview < ActionMailer::Preview
+    def rollup
+      teacher = User.find_or_initialize_by(
+        name: 'Teacher MailerPreview',
+        email: 'teacher@rollup_mailer_preview.com',
+        role: 'teacher'
+      )
+
+      teacher_info = TeacherInfo.find_or_initialize_by(user: teacher)
+      teacher_notification = TeacherNotifications::StudentCompletedDiagnostic.find_or_initialize_by(user: teacher).tap{|row| row.student_name = 'John' }
+      TeacherNotifications::RollupMailer.rollup(teacher, [teacher_notification])
+    end
+  end
--- a/services/QuillLMS/spec/mailers/previews/user_mailer_preview.rb
+++ b/services/QuillLMS/spec/mailers/previews/user_mailer_preview.rb
@@ -1,11 +1,10 @@
 # frozen_string_literal: true
 
-module MailerPreviews
   class UserMailerPreview < ActionMailer::Preview
     def welcome_email
+      puts "hello"
       UserMailer.welcome_email(User.find_by(email: 'marcello.sachs@gmail.com')).tap do |email|
         Premailer::Rails::Hook.delivering_email(email)
       end
     end
   end
-end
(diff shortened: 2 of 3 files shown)


services/QuillLMS/spec/mailers/previews/rollup_mailer_preview.rb
```diff
@@ -0,0 +1,15 @@
...
+
+  class RollupMailerPreview < ActionMailer::Preview
+    def rollup
+      teacher = User.find_or_initialize_by(
+        name: 'Teacher MailerPreview',
+        email: 'teacher@rollup_mailer_preview.com',
+        role: 'teacher'
+      )
+
+      teacher_info = TeacherInfo.find_or_initialize_by(user: teacher)
```
happythenewsad, line 11: Note that we don't currently load FactoryBot into the `development` Rails env. Thus, I opted to initialize (not persist) records using normal AR syntax.

brendanshean: Is `teacher_info` needed here?

services/QuillLMS/spec/mailers/previews/user_mailer_preview.rb
```diff
@@ -0,0 +1,7 @@
+# frozen_string_literal: true
+
+  class UserMailerPreview < ActionMailer::Preview
+    def ell_starter_diagnostic_info_email
+      UserMailer.ell_starter_diagnostic_info_email('pk', 'pkong@quill.org')
```
brendanshean: Maybe use an ENV variable instead of real email.

brendanshean (commented):
Thanks for fixing this functionality.

services/QuillLMS/spec/mailers/previews/user_mailer_preview.rb
happythenewsad: It doesn't actually send an email

services/QuillLMS/spec/mailers/previews/rollup_mailer_preview.rb
happythenewsad: Yes - `see rollup_mailer.rb:32`

services/QuillLMS/spec/mailers/previews/user_mailer_preview.rb
brendanshean: My comment was more about your real email being hard coded in the github repo than it being sent.  Since it's not being sent, no need for an ENV variable.  I would put in something generic like:`user@example.com`

brendanshean (approved):
I would change the email and this should be good to go.

anathomical (approved):

PR #11379, opened 2023-12-21
Lms/audit update all

anathomical:
## WHAT
Audit all places in the code where we use `update_all` and make sure to set `updated_at` when the model supports it
## WHY
We want our `updated_at` to be a reliable indicator of when the row was last modified in the database, and the Rails magic that handles `updated_at` doesn't work for `update_all` calls.  So we need to make sure we explicitly update those values when appropriate.  Specifically, we rely on `activity_sessions.updated_at` to handle syncing updates to that data to BigQuery, and no one thought about the idea that `updated_at` might not actually be up to date, so we should clean this up on any tables we can so that we don't get surprised like this in the future.
## HOW
Just add `updated_at: DateTime.current` to the params of every `update_all` call for models that support `updated_at` (not all do).

PR Checklist | Your Answer
------------ | -------------

### DIFF
--- a/services/QuillLMS/app/models/concerns/student.rb
+++ b/services/QuillLMS/app/models/concerns/student.rb
@@ -137,9 +137,9 @@ module Student
     .where("users.id = ?", id)
     .where("classroom_units.id = ?", classroom_unit_id)
     .where("activity_sessions.visible = true")
-    .order("activity_sessions.is_final_score DESC, activity_sessions.percentage ASC, activity_sessions.started_at")
+    .order("activity_sessions.is_final_score DESC, activity_sessions.percentage DESC, activity_sessions.started_at")
     .offset(1)
-    .update_all(visible: false)
+    .update_all(visible: false, updated_at: DateTime.current)
   end
 
   def merge_student_account(secondary_account, teacher_id=nil)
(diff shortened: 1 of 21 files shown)


services/QuillLMS/app/models/concerns/student.rb
```diff
@@ -137,9 +137,9 @@ def hide_extra_activity_sessions(classroom_unit_id)
     .where("users.id = ?", id)
     .where("classroom_units.id = ?", classroom_unit_id)
     .where("activity_sessions.visible = true")
-    .order("activity_sessions.is_final_score DESC, activity_sessions.percentage ASC, activity_sessions.started_at")
+    .order("activity_sessions.is_final_score DESC, activity_sessions.percentage DESC, activity_sessions.started_at")
```
anathomical, line 140: IMPORTANT NOTE: This is explicitly a change to the code, but from what I can understand from the descriptions in the specs, this is the intended behavior (and we just didn't have a case to prove that this functionality wasn't working before).  It is possible that we shouldn't try to make this change during this PR, but since I was adding the test case anyway, it felt like it should work as intended...

brendanshean (approved):
Good call to update all of these instances of `update_all`

emilia-friedberg (approved):

PR #11400, opened 2024-01-10
Add punctuation feedback to quill-marking-logic

cissyxyu:
## WHAT
Add catch in marking logic to provide feedback when students accidentally type two single quotes instead of a double quote. 
(This code is actually currently implemented and live, [because it went out in a reverted PR ](https://github.com/empirical-org/Empirical-Core/pull/11375) and has already been deployed to NPM. However, because the PR was reverted, the code changes were removed from the repo).

## WHY
We want to start catching this quotation mark mistake and providing appropriate feedback.

## HOW
Catch the mistake using regex and provide the corresponding feedback.

### Screenshots
Screenshot 2023-12-14 at 6 46 12 PM

### Notion Card Links
https://www.notion.so/quill/P3-Fix-quotation-mark-grading-in-Connect-and-Grammar-751cb7e21e6a44feb5e2a99aa92c2d08?pvs=4

PR Checklist | Your Answer
------------ | -------------

### DIFF
--- a/packages/quill-marking-logic/package.json
+++ b/packages/quill-marking-logic/package.json
@@ -1,6 +1,6 @@
 {
   "name": "quill-marking-logic",
-  "version": "0.15.11",
+  "version": "0.15.13",
   "description": "Shared Quill logic for grading and contextualizing student writing.",
   "main": "./dist/lib",
   "module": "./dist/lib",
--- a/packages/quill-marking-logic/src/libs/constants/error_authors.ts
+++ b/packages/quill-marking-logic/src/libs/constants/error_authors.ts
@@ -14,6 +14,7 @@ export interface ErrorAuthors {
   "Punctuation End Hint": string
   "Punctuation Hint": string
   "Punctuation and Case Hint": string
+  "Quotation Mark Hint": string
   "Required Words Hint": string
   "Spacing After Comma Hint": string
   "Starting Capitalization Hint": string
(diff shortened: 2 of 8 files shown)


emilia-friedberg (commented):
Do you know what went wrong with the deploy last time? Also, just for clarity's sake: I don't think this code is currently live, even though the updated package was published to NPM, because part of what got reverted was the `package.json` update for the LMS that directs it to use the most recent version of the package. In testing, it doesn't look like this change is live on our site.

cissyxyu:
> Do you know what went wrong with the deploy last time? Also, just for clarity's sake: I don't think this code is currently live, even though the updated package was published to NPM, because part of what got reverted was the `package.json` update for the LMS that directs it to use the most recent version of the package. In testing, it doesn't look like this change is live on our site.

Yes, sorry, you're totally right -- this code isn't currently "live" on the site because the LMS hasn't been bumped up to this package version yet. It is "live" on my staging environment in the sense that it is being used and tested on there in my staging LMS. 

I still don't know exactly what caused the deploy error last time, I still haven't been able to replicate that error. However I'm sure that pushing this code won't cause any deploy issues since it's only code for the npm package. (This code won't be deployed to LMS or any production application since it's already "deployed" in npm).

I'm hoping that as I keep testing and tracking down that deploy bug from last time I can at least split up my deploys into separate PRs so that this code can at least go out first, and then the package version bumps in the LMS can go out in a second deploy. Although I'm honestly not sure what the best approach is given I still can't replicate the deploy bug...but I've tested this npm package code extensively in staging and haven't run into any issues.

emilia-friedberg (approved):

PR #11411, opened 2024-01-11
lms/fix-detailed-student-activity-score-reports

anathomical:
## WHAT
Remove `if (row.percentage)` guard statement

Also rename `notLessonsOrDiagnostic` to match the name of the function where it is exported from (since it also makes sure an activity is not Evidence in addition to Lessons or Diagnostic).

Note: I'd discussed an alternative solution with Emilia, but further investigation suggested that this even simpler option was better.
## WHY
This guard statement was presumably introduced to prevent trying to build averages on `NULL` values, which would error, but this also excludes cases where students get a `0` score on an activity which means that this report excludes those activities in the rollup at the top.  We also already have a guard statement for `shouldCountForScoring` that should prevent that.

NOTE: I checked our database for legacy data that might still need this guard clause, but the last case of a non Lessons/Diagnostic/Evidence activity that was completed with a `NULL` score was at the end of the 2019-2020 school year, so I think we should be safe to skip it.
## HOW
Remove the guard clause.  Everything else operates exactly the way it always has before.

### Screenshots
Overview row for a student with some `0` score activities
image
Current PROD detailed summary
image
Staging detailed summary:
image

### Notion Card Links
https://www.notion.so/quill/Average-score-discrepancy-on-Activity-Scores-report-b4fc5d31127443d6834e82297f8b90d5

PR Checklist | Your Answer
------------ | -------------
Have you added and/or updated tests? |  No tests on this file

### DIFF
--- a/services/QuillLMS/client/app/bundles/Teacher/components/progress_reports/student_overview.jsx
+++ b/services/QuillLMS/client/app/bundles/Teacher/components/progress_reports/student_overview.jsx
@@ -6,7 +6,7 @@ import _ from 'underscore';
 import CSVDownloadForProgressReport from './csv_download_for_progress_report.jsx';
 import StudentOveriewTable from './student_overview_table.jsx';
 
-import notLessonsOrDiagnostic from '../../../../modules/activity_classifications.js';
+import shouldCountForScoring from '../../../../modules/activity_classifications.js';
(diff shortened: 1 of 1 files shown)


emilia-friedberg (approved):
Love the cleaner solution!

happythenewsad (approved):

PR #11412, opened 2024-01-11
Admin Usage Snapshot PDF

brendanshean:
## WHAT
1.  Manage `PdfSubscription` records pertaining to AdminUsageSnapshot
2. Add cron job that sends AdminUsageSnapshot PDFs for existing PdfSubscription

## WHY
1.  Admins need the ability to subscribe/unsubscribe from AdminUsageSnapshot PDF
2. If an admin is subscribed to a report, the PDF should be sent at the requested cadence (i.e. weekly or monthly)

## HOW
1.  Add a subscribe button to the `UsageSnapshotsContainer`.  Upon clicking the subscribe button reveals the `ReportSubscriptionsModal`.  If user turns subscription on create or update a `PdfSubscription` record.  If the user turns off the report, destroy the `PdfSubscription` record (if it exists).  Users can also unsubscribe via a link in the sent email.
2.  Within the cron job find existing `PdfSubscription` records and send their id to `Pdfs::AdminUsageSnapshotEmailWorker`. This has main parts: building the data and then populating `pdfs/admin_usage_snapshot_report.html.erb` with that data from `DataAggregator` which is then used to construct a PDF. 
Specifically, `DataAggregator` pulls together:
  a) Filtered section: utilizes newly added instance methods to `AdminReportFilterSelection`
  b) Snapshot sections: Uses `SnapshotSectionBuilder` and iterates over hashes in the data frame and, depending on the type key, injects either CountData or RankingData.  The `CountDataInjector` and `RankingDataInjector` both rely on `ResultsFetcher` to retrieve query results either cached or from BigQuery calls.

### Screenshots
[Example PDF Report](https://github.com/empirical-org/Empirical-Core/files/14024683/ADMIN_SNAPSHOT_REPORT__1_12-08-23_4e44b0.pdf)

[Subscribe to this report modal flow](https://github.com/empirical-org/Empirical-Core/assets/2057805/620ed2c1-061c-495c-beb2-2780456d4cf0)

### Notion Card Links
https://www.notion.so/quill/Email-subscription-for-Admin-Usage-Snapshot-Report-1d4f8e91d30549d89f89ec5f8ff3ec8a?pvs=4

PR Checklist | Your Answer
------------ | -------------


services/QuillLMS/.gitignore
```diff
@@ -74,3 +74,4 @@ node_modules
 # https://vitejs.dev/guide/env-and-mode.html#env-files
 *.local
 
+.cache
```
brendanshean, line 77: Generated from Grover

services/QuillLMS/app/controllers/admin_report_filter_selections_controller.rb
```diff
@@ -40,4 +32,7 @@ def create_or_update
         ]
       )
   end
+
+  private def report = admin_report_filter_selection_params[:report]
+  private def user_id = current_user.id
```
anathomical, line 37: Love the syntax.

services/QuillLMS/app/assets/stylesheets/pages/premium_hub/filterable-reports.scss
```diff
@@ -367,6 +373,78 @@
     }
   }
 
+  .manage-report-subscription-modal {
+    height: 500px;
+    width: 600px;
```
emilia-friedberg: You might want to make the width a `max-width`, and the height a `min-height`, so this doesn't get cut off on mobile.

services/QuillLMS/app/assets/stylesheets/pages/premium_hub/filterable-reports.scss
```diff
@@ -367,6 +373,78 @@
...
+      align-items: flex-start;
+      align-self: stretch;
+      flex-direction: column;
+      display: flex;
+      gap: 8px;
+      justify-content: flex-start;
+
+      .title {
+        color: $quill-black;
+        font-family: 'Open Sans';
```
emilia-friedberg: We shouldn't mess with the font-family -- if you remove it, is it not showing up correctly?

services/QuillLMS/app/assets/stylesheets/pages/premium_hub/filterable-reports.scss
```diff
@@ -367,6 +373,78 @@
...
+    width: 600px;
+    background: $quill-white;
+    border-radius: 6px;
+    border: 1px solid $quill-grey-5;
+    display: inline-flex;
+    padding: 40px 32px;
+
+    .title-and-description {
+      align-items: flex-start;
+      align-self: stretch;
```
emilia-friedberg: I generally don't need to use this property, or `word-break` below -- are they necessary here?

services/QuillLMS/app/assets/stylesheets/pages/premium_hub/filterable-reports.scss
```diff
@@ -367,6 +373,78 @@
...
+        word-break: break-word;
+      }
+    }
+
+    .is-subscribed-and-frequency {
+      .is-subscribed-options {
+        padding-bottom: 32px;
+
+        .is-subscribed-label {
+          color: #7F7F7F;
```
emilia-friedberg: Should this be a variable? Also, looks like this CSS is identical to `.frequency-label` below, so those could be combined.

services/QuillLMS/app/helpers/application_helper.rb
```diff
@@ -65,7 +65,7 @@ def on_sign_up?
   end
 
   def on_sign_up_or_log_in?
-    %w[account/new /session password_reset sign-up finish_set_up /canvas_integration/lti/sso]
+    %w[account/new /session password_reset sign-up finish_set_up /canvas_integration/lti/sso /pdf_subscriptions/unsubscribe]
       .any? { |str| request.path.include?(str) }
   end
```
anathomical, line 70: Outside the scope of this PR, but I wonder if we should rename this helper function at some point so that it's clear what it does.  (This would necessitate updating calls in all the templates, so something to do later.)

services/QuillLMS/app/assets/stylesheets/shared/modals.scss
```diff
@@ -496,13 +496,15 @@
...
-.already-has-premium-modal {
+.already-has-premium-modal,
+.manage-report-subscription-modal {
   width: 560px;
   overflow: visible;
   top: 50%;
   transform: translate(-50%, -50%);
   position: fixed !important;
   text-align: left;
+  z-index: 1000;
```
emilia-friedberg: Do we need to change the `z-index` for all of these modals? It might be better to just set this in the other file for `.manage-report-subscription-modal` -- I doubt it, but tracking down CSS bugs is really annoying, so better to not mess with a lot of places at once.

services/QuillLMS/app/controllers/admin_report_filter_selections_controller.rb
emilia-friedberg, line 37: One liners!!

services/QuillLMS/app/services/pdfs/admin_usage_snapshot_reports/data_aggregator.rb
```diff
@@ -0,0 +1,21 @@
...
+      attr_reader :admin_report_filter_selection
+
+      delegate :classrooms, :grades, :schools, :teachers, :timeframe_name,
+        to: :admin_report_filter_selection
+
+      def initialize(admin_report_filter_selection)
+        @admin_report_filter_selection = admin_report_filter_selection
+      end
+
+      def run = { filter_sections:, snapshot_sections:}
```
anathomical: Minor nitpick: should there be a space between the final `:` and `}`?

services/QuillLMS/app/services/pdfs/admin_usage_snapshot_reports/results_fetcher.rb
```diff
@@ -0,0 +1,70 @@
...
+            timeframe_value,
+            timeframe_start,
+            timeframe_end,
+            school_ids,
+            additional_filters: { grades: grade_values, teacher_ids:, classroom_ids: }
+          )
+      end
+
+      private def cached_results = Rails.cache.read(cache_key)
+      private def run_query = worker.new.perform(*worker_args)
```
anathomical, line 49: I like using a synchronous worker here to ensure that the cache is populated, but I'm a little worried about how we ensure that people who call `run` understand that this could take a non-trivial amount of time to run.  Don't want people to be "surprised" that this goes to BigQuery.

services/QuillLMS/app/controllers/pdf_subscriptions_controller.rb
```diff
@@ -0,0 +1,55 @@
...
+
+  def unsubscribe
+    @pdf_subscription = PdfSubscription.find_by(token:)
+
+    if @pdf_subscription&.destroy
+      render json: {}, status: :ok
+    else
+      redirect_to root_path, flash: { error: 'Subscription not found' }
+    end
+  end
```
emilia-friedberg: NBD but it might help to add a comment to these methods explaining the difference because I had to do some hunting to understand why we need both.

services/QuillLMS/app/services/pdfs/admin_usage_snapshot_reports/snapshot_sections_builder.rb
```diff
@@ -0,0 +1,299 @@
+# frozen_string_literal: true
+
+module Pdfs
+  module AdminUsageSnapshotReports
+    class SnapshotSectionsBuilder < ApplicationService
+      COUNT = 'count'
```
emilia-friedberg, line 9: Probably unnecessary, but it might be nice to have a comment in here linking to the relevant frontend file and vice versa in case we change something in one place and need to remember to change it in the other?

services/QuillLMS/client/app/bundles/PremiumHub/components/usage_snapshots/__tests__/reportSubscriptionModal.test.tsx
```diff
@@ -0,0 +1,65 @@
+import * as React from 'react';
```
emilia-friedberg, line 1: Great tests!

services/QuillLMS/client/app/bundles/PremiumHub/components/usage_snapshots/reportSubscriptionModal.tsx
```diff
@@ -0,0 +1,138 @@
...
+  function renderFrequencyOptions() {
+    const selectedFrequency = frequencyOptions.find(option => frequency === option.value)
+
+    return (
+      <div className="frequency-options">
+        <label className="frequency-label" htmlFor="frequency">
+          <span>Frequency</span>
+        </label>
+        <DropdownInput
+          className={`${isSubscribed ? '' : 'disabled'}`}
```
emilia-friedberg: Tiny, but you could do `className={isSubscribed ? '' : 'disabled'}` instead of having to wrap the whole thing in ````${}````.

services/QuillLMS/app/controllers/pdf_subscriptions_controller.rb
```diff
@@ -0,0 +1,55 @@
+# frozen_string_literal: true
+
+class PdfSubscriptionsController < ApplicationController
+  def existing
```
anathomical: I don't care here, but should we consider calling this `def index` to keep with Rails conventions of `GET` actions on "naked" resources?

emilia-friedberg (approved):
Looks awesome. Really nice work on the frontend and great tests. I made some very small comments, mostly on CSS, but nothing that requires rereview. I'm really impressed by your work here, especially in an area of the code you don't touch much!

services/QuillLMS/app/services/pdfs/admin_usage_snapshot_reports/count_data_injector.rb
```diff
@@ -0,0 +1,57 @@
...
+      end
+
+      def run
+        return item unless valid_query_key?
+
+        item.merge(count:, change:)
+      end
+
+      private def change
+        CountDataChangeCalculator.run(current_count: count, previous_count: previous_results[:count])
```
anathomical: I don't hate this pattern, but is the calculation here really complicated enough to be worth the mental overhead of creating a local class to handle it?  Is this primarily intended to simplify testing?

services/QuillLMS/app/services/pdfs/file_builder.rb
```diff
@@ -0,0 +1,52 @@
...
+    def initialize(data:, template:)
+      @data = data
+      @template = template
+    end
+
+    def run
+      tempfile.write(pdf)
+      tempfile.rewind
+      tempfile
+    end
```
anathomical: Documentation for `Tempfile` suggests using a `begin... ensure` block or something similar to guarantee that `close` and `unlink` get called on your temp file.  https://ruby-doc.org/stdlib-2.4.0/libdoc/tempfile/rdoc/Tempfile.html#class-Tempfile-label-Good+practices

services/QuillLMS/config/database.yml
```diff
@@ -34,6 +34,13 @@ development:
     database: emp_gr_development
     replica: true
 
+  #  For development group use this configuration temporarily for staging database locally
+  # primary:
+  #   url: <%= ENV['DATABASE_URL'] %>
+  # replica:
+  #   url: <%= ENV['DATABASE_URL'] %>
+  #   replica: true
+
```
anathomical: Do we actually want to leave this here for future developers to use, or is it an accidental commit?

services/QuillLMS/client/app/bundles/PremiumHub/components/usage_snapshots/__tests__/reportSubscriptionModal.test.tsx
```diff
@@ -0,0 +1,65 @@
...
+      <ReportSubscriptionModal
+        cancel={mockCancel}
+        existingPdfSubscription={subscription}
+        isOpen={isOpen}
+        save={mockSave}
+      />
+    );
+
+  beforeEach(() => {
+    jest.clearAllMocks();
```
anathomical, line 21: Oh, I wasn't familiar with `clearAllMocks()`.  What a useful function.

services/QuillLMS/client/app/bundles/PremiumHub/components/usage_snapshots/reportSubscriptionModal.tsx
```diff
@@ -0,0 +1,138 @@
...
+  }
+
+  function renderIsSubscribedAndFrequency() {
+    return (
+      <div className="is-subscribed-and-frequency">
+        {renderIsSubscribedOptions()}
+        {renderFrequencyOptions()}
+      </div>
+    )
+  }
```
anathomical: This is a super-minor ordering thing, but I generally like wrapper functions like this to come AFTER all the function definitions that it wraps in the code.  Sometimes I get confused if I have to look "down" to find the things called in a function.  So I'd put this definitino after `renderIsusbscribedOptions`.

services/QuillLMS/client/app/bundles/PremiumHub/containers/PremiumFilterableReportsContainer.tsx
```diff
@@ -189,7 +189,7 @@ export const PremiumFilterableReportsContainer = ({ accessType, adminInfo, }) =>
     })
   }
 
-  function saveFilterSelections() {
+  function saveFilterSelections(report = reportPath(), successCallback = () => { }) {
```
anathomical, line 192: I really like moving these values up here into the function signature.  So much easier to test.

services/QuillLMS/client/app/bundles/PremiumHub/containers/UsageSnapshotsContainer.tsx
```diff
@@ -35,15 +38,63 @@ export const UsageSnapshotsContainer = ({
...
+      saveFilterSelections(PDF_REPORT, (adminReportFilterSelection) => {
+        createOrUpdatePdfSubscription(adminReportFilterSelection, frequency)
+      })
+    } else if (existingPdfSubscription) {
+      deletePdfSubscription(existingPdfSubscription.id)
+    }
+  }
+
+  function createOrUpdatePdfSubscription(adminReportFilterSelection, frequency) {
+    if (adminReportFilterSelection && adminReportFilterSelection.id) {
```
anathomical: I think we can use a safe-accessor here `if(adminReportFilterSelection?.id) {`

Also, we often prefer early returns over fully-wrapped `if` statements, so maybe this should just be `if(!adminReportFilterSelection?.id) return`

services/QuillLMS/client/app/bundles/PremiumHub/containers/UsageSnapshotsContainer.tsx
```diff
@@ -35,15 +38,63 @@ export const UsageSnapshotsContainer = ({
...
+      requestPost('/pdf_subscriptions/create_or_update', pdfSubscriptionParams, (pdfSubscription) => {
+        setExistingPdfSubscription(pdfSubscription)
+      })
+    }
+  }
+
+  function deletePdfSubscription(pdfSubscriptionId) {
+    requestDelete(`/pdf_subscriptions/${pdfSubscriptionId}`, {}, null, error => { throw (error) })
+    setExistingPdfSubscription(null)
+  }
```
anathomical: Not a huge deal or anything, but I note that we use a callback for `setExistingPdfSubscription` on `createOrUpdate`, but don't bother when using `delete`.  From a user experience perspective, I doubt this matters, but I wonder if there's some value in syncing these up for consistency sake.  (I recognize that we MUST use a callback in the former case because we want to set the value to the response, and it doesn't really matter for the deletion, but I figured I'd flag this anyway.)

services/QuillLMS/config/routes.rb
```diff
@@ -758,6 +758,14 @@
     end
   end
 
+  resources :pdf_subscriptions, only: [:destroy] do
+    collection do
+      post :create_or_update
+      get :existing
```
anathomical: If you do rename `existing` to `index`, you'll want to remove this and add `:index` to the `only` array above.

services/QuillLMS/spec/controllers/pdf_subscriptions_controller_spec.rb
```diff
@@ -0,0 +1,120 @@
...
+      end
+    end
+
+    context 'when updating an existing subscription' do
+      let(:old_freq) { PdfSubscription::WEEKLY }
+      let(:new_freq) { frequency }
+
+      context 'when the pdf subscription belongs to the current user' do
+        let(:admin_report_filter_selection_id) { create(:admin_report_filter_selection, user:).id }
+        let(:pdf_subscription) { create(:pdf_subscription, admin_report_filter_selection_id:, frequency: old_freq) }
```
anathomical: This `let` call is repeated in this context and the next, suggesting that it could be moved up one level into the immediately containing context.

services/QuillLMS/spec/lib/snapshots/cache_keys_spec.rb
```diff
@@ -20,20 +20,24 @@ module Snapshots
       }
     end
 
+    it { expect(timeframe_value.class).to eq String }
+    it { expect(current_start.class).to eq DateTime }
+    it { expect(current_end.class).to eq DateTime }
```
anathomical, line 25: I like the addition of this type checking in the spec.  Can't enforce usage, but is good for communicating intent.

services/QuillLMS/spec/mailers/previews/premium_hub_user_mailer_preview.rb
```diff
@@ -0,0 +1,7 @@
+# frozen_string_literal: true
+
+class PremiumHubUserMailerPreview < ActionMailer::Preview
+  def admin_usage_snapshot_report_pdf_email
+    PremiumHubUserMailer.admin_usage_snapshot_report_pdf_email(download_url: 'localhost:3000/blah', pdf_subscription: PdfSubscription.last)
```
anathomical, line 9: Since this is for local testing, this may not matter, but if a user hits this without any valid `PdfSubscription` records in their local database, will they get a useful error so they can figure out what to fix?

services/QuillLMS/spec/models/admin_report_filter_selection_spec.rb
```diff
@@ -22,8 +22,154 @@
...
+
+  describe '.segment_admin_report_subscriptions' do
+    subject { described_class.segment_admin_report_subscriptions }
+
+    it { expect(subject).to match_array [] }
+
+    context 'when there are pdf subscriptions' do
+      let(:usage_snapshot) { described_class::SEGMENT_MAPPING[described_class::USAGE_SNAPSHOT_REPORT_PDF] }
+
+      before { create(:pdf_subscription) }
```
anathomical, line 39: Are we preferring `before { ... }` over `let!(:foo) { ... }`?  I'm okay with the `subject` pattern, and just want to make sure I know what to do with the next spec i write.  Or is there some other reason to use `subject` here that I'm missing?

services/QuillLMS/spec/models/segment_integration/user_spec.rb
```diff
@@ -52,7 +53,17 @@
...
-        expect(admin.segment_user.common_params).to eq params
+      end
+
+      it { is_expected.to eq common_params }
+
+      context 'when the admin is a premium admin' do
+        before { allow(admin).to receive(:premium_admin?).and_return(true) }
+
+        it {
+          binding.pry
```
anathomical: Is this a leftover `binding.pry` from testing?

services/QuillLMS/spec/services/pdfs/file_builder_spec.rb
```diff
@@ -0,0 +1,36 @@
+# frozen_string_literal: true
+
+require 'rails_helper'
+
+module Pdfs
+  RSpec.describe FileBuilder do
+    subject { described_class.run(data:, template:) }
+
+    let(:application_controller) { double(render_to_string: pdf_string) }
```
anathomical: Since we have an explicit `allow(application_controller).to receive(:render_to_string)` in the `before` block below, do we actually get value out of the `render_to_string: pdf_string` param in the `double` declaration?

services/QuillLMS/spec/workers/pdfs/admin_usage_snapshot_email_worker_spec.rb
```diff
@@ -0,0 +1,79 @@
...
+    subject { described_class.new.perform(pdf_subscription_id) }
+
+    let(:admin_report_filter_selection) { create(:admin_report_filter_selection, :with_default_filters) }
+    let(:pdf_subscription) { create(:pdf_subscription, admin_report_filter_selection:) }
+    let(:pdf_subscription_id) { pdf_subscription.id }
+    let(:user) { pdf_subscription.user }
+    let(:user_id) { user.id }
+
+    before do
+      allow(PdfSubscription).to receive(:find).with(pdf_subscription_id).and_return(pdf_subscription)
```
anathomical, line 16: Is this `allow` necessary?  Creating the records should make `find` work as expected without this, shouldn't it?

anathomical (approved):
This is a huge amount of new functionality.  Nice job on putting it all together, and especially nice job having such extensive spec coverage.  I like the services pattern that you're pointing at here with very clean, simple, almost functional services feeding into each other.  I've got a bunch of little notes throughout, but nothing that I think you really have to change.  More just food for thought.

services/QuillLMS/app/controllers/admin_report_filter_selections_controller.rb
```diff
@@ -1,21 +1,13 @@
...
 
 class AdminReportFilterSelectionsController < ApplicationController
-
   def show
-    admin_report_filter_selection = AdminReportFilterSelection.find_by(
-      report: admin_report_filter_selection_params[:report],
-      user_id: current_user.id
-    )
-
+    admin_report_filter_selection = AdminReportFilterSelection.find_by(report:, user_id:)
```
dandrabik, line 5: Ah, interesting, this is the new syntax you were referencing. Very concise!

services/QuillLMS/app/controllers/pdf_subscriptions_controller.rb
dandrabik: I agree with this comment. I'd recommend adding a comment or making the endpoint name more verbose, or both.

services/QuillLMS/app/mailers/premium_hub_user_mailer.rb
```diff
@@ -22,6 +22,17 @@ def admin_premium_download_report_email(user_first_name, file_url, email)
...
 
+  def admin_usage_snapshot_report_pdf_email(download_url:, pdf_subscription:)
+    @download_url = download_url
+    @user = pdf_subscription.user
+    @frequency = pdf_subscription.frequency
+    @user_first_name = @user.first_name
+    @premium_hub_url = teachers_premium_hub_url
+    @unsubscribe_url = unsubscribe_pdf_subscriptions_url(token: pdf_subscription.token)
+
+    mail to: @user.email, subject: "🚀 Your #{@frequency} Quill Admin Usage Snapshot Report is Ready!"
```
dandrabik, line 33: 🚀

services/QuillLMS/app/models/cron.rb
```diff
@@ -68,6 +71,7 @@ def self.run_weekday
 
   def self.run_friday
     TeacherNotifications::EnqueueUsersForRollupEmailWorker.perform_async(TeacherInfo::WEEKLY_EMAIL)
+    Pdfs::SendWeeklySubscriptionsWorker.perform_async
```
dandrabik: Not a code question, but something we might consider in a small v2: we might have better open rates/engagement if we send these on Monday for the week before.

services/QuillLMS/client/app/bundles/PremiumHub/components/usage_snapshots/reportSubscriptionModal.tsx
brendanshean: I like that better

services/QuillLMS/app/services/pdfs/admin_usage_snapshot_reports/snapshot_sections_builder.rb
brendanshean: Good idea

services/QuillLMS/client/app/bundles/PremiumHub/components/usage_snapshots/__tests__/reportSubscriptionModal.test.tsx
brendanshean: There's also `resetAllMocks`
https://dev.to/edwinwong90/jestclearallmocks-vs-jestresetallmocks-vs-jestrestoreallmocks-explained-5aal

services/QuillLMS/spec/models/segment_integration/user_spec.rb
brendanshean: Whoops.  Good catch.

services/QuillLMS/app/helpers/application_helper.rb
brendanshean: I agree.  It took me awhile to remember why this particular helper gave me the minimal header I wanted.  I've made a note to update this.

services/QuillLMS/app/services/pdfs/admin_usage_snapshot_reports/data_aggregator.rb
brendanshean: I agree.  I need to figure out why my editor didn't automatically format that.

services/QuillLMS/app/services/pdfs/admin_usage_snapshot_reports/count_data_injector.rb
brendanshean: I originally wrote the tests without this local class but then opted for one to simplify testing.  I'm open other ideas here.

services/QuillLMS/app/controllers/pdf_subscriptions_controller.rb
brendanshean: Good call.  I will probably forget why I wrote the different endpoints in a week.

services/QuillLMS/config/database.yml
brendanshean: It was intentional.  I did add an additional comment here: https://github.com/empirical-org/Empirical-Core/pull/11412/files#diff-21b1db8d989ec4a2050f898f6c86b08e86868bde2518302eab8e26f41336235cR38
to discourage people from checking in the configuration uncommented.

Maybe something for engineering meeting discussion?

services/QuillLMS/client/app/bundles/PremiumHub/components/usage_snapshots/reportSubscriptionModal.tsx
brendanshean: I think this is an interesting preference as I tend to just order alphabetically so I don't need to think about where things go.  However, I can understand your preference for ordering.  I'll move it per your request here but I think this would be an interesting item for discussion in engineering.

services/QuillLMS/client/app/bundles/PremiumHub/containers/UsageSnapshotsContainer.tsx
brendanshean: Much cleaner

services/QuillLMS/client/app/bundles/PremiumHub/containers/UsageSnapshotsContainer.tsx
brendanshean: Are you thinking something like this:

```
function deletePdfSubscription(pdfSubscriptionId) {
    requestDelete(`/pdf_subscriptions/${pdfSubscriptionId}`, {}, () => setExistingPdfSubscription(null), error => { throw (error) })
  }
```

services/QuillLMS/spec/services/pdfs/file_builder_spec.rb
brendanshean: Good catch!

services/QuillLMS/spec/workers/pdfs/admin_usage_snapshot_email_worker_spec.rb
brendanshean: This one is kind of tricky.  The find will work as expected, however, I don't think the subsequent stubbing of       `allow(user).to receive(:school_or_district_premium?).and_return(school_or_district_premium)` will work properly  since the `user` is something stored in memory from the let call rather than the `user` that from `PdfSubscription.find(pdf_subscription_id).user`.  There might be a better way to do this but this.

services/QuillLMS/app/services/pdfs/file_builder.rb
brendanshean: I forgot about this.  Thank you for the reminder.

dandrabik (approved):
:+1: This looks great. Thanks for all of your work on this. I'm loving the use of the new compact syntax. Well done on this.

services/QuillLMS/app/services/pdfs/file_builder.rb
brendanshean: I've added the cleanup to FileBuilder along with a required block parameter.  Here's an example of its use: https://github.com/empirical-org/Empirical-Core/pull/11412/files#diff-ab8318030d12bb7e3bfd2cc803ef661fd315cba1da9c07beff41f832f9f43118R29

services/QuillLMS/client/app/bundles/PremiumHub/containers/UsageSnapshotsContainer.tsx
anathomical: Yeah, that was what I was contemplating.  Again, I don't know if it's actually better, so feel free to retain the current approach if you prefer it.

services/QuillLMS/spec/workers/pdfs/admin_usage_snapshot_email_worker_spec.rb
anathomical: Good point.  The only alternative I can think of to doing it this way is `allow_any_instance_of`, and we definitely don't want to do that.  This seems like a reasonable, if slightly awkward, solution to the problem and I don't have a better suggestion in mind.

services/QuillLMS/app/models/cron.rb
brendanshean: After talking with Peter, we will go with Monday now.

services/QuillLMS/app/controllers/pdf_subscriptions_controller.rb
brendanshean: After discussing with Thomas, changing to `GET existing` to `GET current`.  The index semantics is for collections which doesn't apply here and show requires an id which we are not passing here.

services/QuillLMS/app/services/pdfs/admin_usage_snapshot_reports/count_data_injector.rb
brendanshean: Removed the inner class.  The tests had duplication but I think it's better.

services/QuillLMS/spec/mailers/previews/premium_hub_user_mailer_preview.rb
brendanshean: Good call.  I've added an error

PR #11422, opened 2024-01-17
validate dom nesting error - part one

emilia-friedberg:
## WHAT
Fix some, but not all, of the validate DOM nesting errors we're getting in the test suite.

## WHY
We want to clean up these errors (it's one of our OKRs for the quarter, and good practice besides). However, this PR is already large because of the change to `DataTable`, and since I have other higher priorities for this sprint I thought I would submit this changeset now and the rest later.

## HOW
Fix one instance by just adding a `tbody` element, and the larger `DataTable` issue with links by removing the `a` that wrapped all the `td` in favor of a new component that adds an `a` inside the `td` when necessary, in addition to some styling that makes the link clickable across the entire width and height of the row.

### Screenshots
No visual change.

### Notion Card Links
https://www.notion.so/quill/Console-log-cleanup-validateDOMNesting-26-1f9dc32dcd034942b40f08a05543f339?pvs=4

PR Checklist | Your Answer
------------ | -------------


services/QuillLMS/client/app/bundles/Teacher/components/shared/__tests__/__snapshots__/view_as_student_modal.test.jsx.snap
```diff
@@ -5138,7 +5138,9 @@ exports[`ViewAsStudentModal component without a defaultClassroomId should render
                     }
                   }
                 >
-                  Bleep Bloop
+                  <TableCellContentWrapper>
+                    Bleep Bloop
```
brendanshean, line 5142: 🤖

brendanshean (approved):
Thank you for fixing this error.  I also appreciate keeping the changeset atomic for a quick review.

services/QuillLMS/client/app/bundles/Shared/components/shared/dataTable.tsx
```diff
@@ -69,6 +69,14 @@ interface DataTableState {
   hoveredColIndex: null|number;
 }
 
+const TableCellContentWrapper = ({ link, children, }) => {
```
anathomical, line 72: This approach not only solves an issue, but also feels cleaner to me.  I like this kind of modularity.

services/QuillLMS/client/app/bundles/Staff/components/styleGuide/__tests__/dataTable.test.jsx
```diff
@@ -1,6 +1,6 @@
 import * as React from 'react';
-
-import { shallow } from 'enzyme';
+import { render, screen } from '@testing-library/react';
```
anathomical, line 2: I'm still appreciating the approach in `@testing-library`.  I find it a lot easier to understand what the test is trying to check for.

anathomical (approved):

services/QuillLMS/client/app/bundles/Shared/components/shared/dataTable.tsx
eadams17: agreed, I like this approach

eadams17 (approved):

PR #11423, opened 2024-01-17
Lms/tweak admin diagnostic query

anathomical:
## WHAT
- Fix a bug that was over-counting diagnostic assignments and completions, causing the numbers in the admin report to not match the teacher-level report numbers by one or two in a couple of cases
- Fix a bug in the logic that converts "time spent" values from pure seconds to "minutes:seconds" format.  The existing code was throwing away any values over one hour and displaying only what was left.  So 1 hour, 1 minute, 1 second was displaying as "01:01" instead of "61:01"
## WHY
- We want the numbers we're showing teachers and the numbers we're showing admins to match so that we don't create confusion in cases where they compare
- We want to see the full values for time spent as a number of teachers are doing enough practice on Quill that their students are legitimately spending multiple hours on practice activities
## HOW
- Make sure to exclude `ActivitySession` records that have been deleted (had `visible` set to `false`) when doing admin roll-ups
- Skip the process of modulo-ing off the units larger than an hour in the process that calculates how to display time in our Admin report tables

### Screenshots
Average time spent numbers can go over `59:59` now
image

### Notion Card Links
https://www.notion.so/quill/Admin-Diagnostic-Growth-Report-QA-Overview-Tab-December-7-0524fc96c41f424a9532228535b11553#e67342bc03694ef0ba3c94b4f044addd

PR Checklist | Your Answer
------------ | -------------
Have you added and/or updated tests? |  Yes, the query specs have been updated to reflect the new `visible` expectations, and front-end jest expectations have been updated.

### DIFF
--- a/services/QuillLMS/app/queries/admin_diagnostic_reports/diagnostic_recommendations_query.rb
+++ b/services/QuillLMS/app/queries/admin_diagnostic_reports/diagnostic_recommendations_query.rb
@@ -14,6 +14,7 @@ module AdminDiagnosticReports
       super + <<-SQL
         JOIN lms.activity_sessions
           ON classroom_units.id = activity_sessions.classroom_unit_id
+            AND activity_sessions.visible = true
         JOIN lms.units
           ON classroom_units.unit_id = units.id
         JOIN lms.recommendations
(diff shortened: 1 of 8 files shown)


happythenewsad (approved):

brendanshean (approved):
Thank you for tracking down and fixing this bug.

eadams17 (approved):

PR #11425, opened 2024-01-18
various premium hub updates

emilia-friedberg:
## WHAT
Various Premium Hub fixes and improvements, including:

- fixing a bug where "Clear filters" always shows on Usage Snapshot Report
- make School Subscriptions page accessible to users who have expired premium (as opposed to having never had premium)
- make a more informative page for non-premium visitors to the Usage Snapshot Report

## WHY
We want the Premium Hub to be bug-free, and to be informative for users who don't necessarily have access to every feature.

## HOW
For the first bug, it was happening as an edge case when one of the teachers an admin had oversight over had a classroom with no grade associated. That teacher would automatically get filtered out, because the saved filter selections explicitly included every grade if no selection was made, rather than defaulting to null the way we do for the other filters. I fixed this by having it save null if no grade filters were actually selected. The other two were just React/CSS updates.

### Screenshots
Screenshot 2024-01-18 at 9 50 29 AM

### Notion Card Links
https://www.notion.so/quill/The-Clear-filters-button-is-enabled-on-the-Usage-Snapshot-Report-even-if-not-filters-are-applied-94d77102fb8744e2b1f6c5054563a310?pvs=4

https://www.notion.so/quill/Allow-admins-that-have-an-expired-subscription-to-access-the-Subscriptions-page-on-the-Premium-Hub-11beb4455d594719bbb2a1b87e3638ae?pvs=4

https://www.notion.so/quill/Update-the-Usage-Snapshot-Report-locked-page-for-non-premium-admins-de3669363a324a18801760bc6dfca639?pvs=4

PR Checklist | Your Answer
------------ | -------------


happythenewsad (approved):

anathomical (approved):

services/QuillLMS/app/helpers/footer_helper.rb
```diff
@@ -71,8 +63,4 @@ def about_us_links
       { href: '/contact', label: 'Contact Us' }
     ]
   end
-
-  def show_footer?(path)
-    !path.in?(EXCLUDED_FOOTER_PATHS)
-  end
```
eadams17, line 77: if we're removing this method, we should also remove the tests

eadams17 (approved):
LGTM!

PR #11438, opened 2024-01-23
add bq docs, sql file structure

happythenewsad:
## WHAT
- create and standardize a way to manage BigQuery mat views via the `bq` command line tool

## WHY
- we want to version our mat views in sql files, in order to track changes
- CLI is a bit tighter and faster than copy-pasting into the BigQuery studio 

### Screenshots

### Notion Card Links
https://www.notion.so/quill/Product-Board-30f97e4eb01246dbb8264253fb385073?p=962bf7c18cad4612a0c64613bde5ac9d&pm=s

PR Checklist | Your Answer
------------ | -------------

### DIFF
--- a/services/QuillLMS/db/external/bigquery/mat_views/recent_reporting_sessions.sql
+++ b/services/QuillLMS/db/external/bigquery/mat_views/recent_reporting_sessions.sql
@@ -1,0 +1,47 @@
+CREATE MATERIALIZED VIEW analytics-data-stores.lms.recent_reporting_sessions
+OPTIONS (
+  enable_refresh = true, refresh_interval_minutes = 1440,
+  max_staleness = INTERVAL "24" HOUR,
+    allow_non_incremental_definition = true
+)
+AS (
+  SELECT
+   activity_sessions.user_id as student_id,
+   activity_sessions.activity_id as activity_id,
+   activities.name as activity_name,
+   schools.id as school_id,
+   schools.name as school_name,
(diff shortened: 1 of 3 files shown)


services/QuillLMS/db/external/bigquery/mat_views/recent_reporting_sessions.sql
```diff
@@ -0,0 +1,47 @@
+CREATE MATERIALIZED VIEW analytics-data-stores.lms.recent_reporting_sessions
+OPTIONS (
+  enable_refresh = true, refresh_interval_minutes = 1440,
```
happythenewsad: We want this to auto refresh off peak hours, to minimize downtime.

dandrabik: Maybe I'm not understanding, but how is this refreshing off peak hours? We need the stats to be up to date for the day before, but I'm not sure how this setting ensures that. I'm reading this as refresh every 24-hours with a max staleness of 24 hours. Could this refresh at noon one day and someone looking at in the next morning wouldn't have yesterday's full stats?

dandrabik (changes requested):
Overall the structure seems fine. I don't think the refresh rate fits our SLA for data up to midnight of the day before, so I think we need to fix that.

dandrabik (approved):
Thanks for making the changes. 🚢

PR #11439, opened 2024-01-24
Update values in “Log-in method” column on the “My Classes” page

brendanshean:
## WHAT
Update the values in the “Log-in method” column on the “My Classes” page

## WHY
We'd like to provide users with the most accurate information.

## HOW
Add a conditional that uses provider and/or email value to calculate the value for Log-in Method.

### Screenshots
Screenshot 2024-01-24 at 7 33 42 AM

### Notion Card Links
https://www.notion.so/quill/Update-the-values-in-the-Log-in-method-column-on-the-My-Classes-page-e0ee787389c1403eaf9149eece1979ac?pvs=4

PR Checklist | Your Answer
------------ | -------------

### DIFF
--- a/services/QuillLMS/client/app/bundles/Teacher/components/classrooms/classroom_student_section.tsx
+++ b/services/QuillLMS/client/app/bundles/Teacher/components/classrooms/classroom_student_section.tsx
@@ -21,7 +21,7 @@ const googleSetupInstructionsPdf = `${process.env.CDN_URL}/documents/setup_instr
 
 function activeHeaders(hasClassroomProvider: boolean) {
   const name = {
-    width: '220px',
+    width: '205px',
     name: 'Name',
     attribute: 'name'
   }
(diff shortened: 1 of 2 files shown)


services/QuillLMS/client/app/bundles/Teacher/components/classrooms/classroom_student_section.tsx
```diff
@@ -489,6 +489,15 @@ const ClassroomStudentSection = ({
     )
   }
 
+  const logInMethod = ({ email, provider }) => {
+    if (provider === 'Google') { return 'Google or email' }
+    if (provider === 'Clever') { return 'Clever' }
+    if (provider === 'Canvas') { return 'Canvas' }
```
emilia-friedberg: Nbd but we might want to constantize the provider names since we're using them for logic now.

emilia-friedberg (approved):

services/QuillLMS/client/app/bundles/Teacher/components/classrooms/classroom_student_section.tsx
brendanshean: Good point

PR #11450, opened 2024-01-26
Upgrade LMS Ruby Version to 3.2

cissyxyu:
## WHAT
Upgrade Ruby in LMS to version 3.2.

## WHY
To have the most up-to-date and secure version of Ruby.

## HOW
Update all references to Ruby. 
Changed some syntax to be compatible with Ruby 3.2 conventions.
Pinned grpc, graphql and sprockets-rails to previous versions to prevent errors on precompiling assets.

### Screenshots

### Notion Card Links
[(Please provide links to any relevant Notion card(s) relevant to this PR.)](https://www.notion.so/quill/Upgrade-LMS-Ruby-3-2-3e6b3363b0b048ab999d9357f3c060d1?pvs=4)

PR Checklist | Your Answer
------------ | -------------


services/QuillLMS/config/vite.json
```diff
@@ -2,7 +2,8 @@
   "all": {
     "sourceCodeDir": "client/app",
     "entryPointsDir": "client/app/entrypoints",
-    "watchAdditionalPaths": ["app/assets/stylesheets"]
+    "watchAdditionalPaths": ["app/assets/stylesheets"],
+    "skipCompatibilityCheck": true
```
cissyxyu, line 6: I added this because vite kept stopping compile with the Error "ruby 3.2 may not be compatible with vite_ruby 3.5.0". The gem is in fact compatible. But the warning is causing the code not to compile, so I'm removing the warning for now since during future updates of this gem we will always manually test to make sure everything is compatible.

services/QuillLMS/Gemfile
```diff
@@ -132,7 +132,7 @@ gem 'vite_rails', '~> 3.0.14'
 # ASSET/UI
 gem 'uglifier', require: false
 gem 'kaminari', '~> 1.2.1'
-gem 'sprockets-rails', '~> 3.2.2'
+gem 'sprockets-rails', '~> 2.3.3'
```
brendanshean, line 135: Can you explain the downgrade here as 2.3.3 is from 2015 whereas 3.2.2 is from 2020.

happythenewsad: I assume this is a typo; if it wasn't, we need to find an alternative to rolling back so far, as Brendan suggests.

Relatedly: I believe the only thing we use sprockets for now (post-vite) is this directive:

```
  config.assets.precompile += %w(
                                 cms.js
                                 scorebook/*
                                 scorebook/**/*
                                 sign_up_email.css,
                                 admin_usage_snapshot_report_pdf.scss
                                 ) 
```
                                 
So versioning these files directly in public/ could allow us to drop sprockets completely, which would be nice.

services/QuillLMS/config/vite.json
happythenewsad: The[ vite ruby docs ](https://vite-ruby.netlify.app/config/index.html#skipcompatibilitycheck)specify that doing this `could lead to hard-to-debug errors such as missing entrypoints in the build.`

Thus, it's worth the extra research into resolving this issue, rather than using the escape hatch.

happythenewsad (changes requested):
Hi Cissy - through no fault of your own, there are two dependency issues that IMO need more research to be solved correctly. 

I'm here to pair over zoom, if that's helpful 👋

happythenewsad:
@cissyxyu This page indicates that upgrading from Node 16 to 18 may fix the issue. If that's the case, let's roll out the Node 18 upgrade independently.

https://www.cnblogs.com/facingscreen/p/17871005.html

anathomical (approved):
I don't see anything worth flagging that hasn't already been flagged, so I'm approving this with the expectation that issues other folks have flagged get addressed.

dandrabik (commented):
@cissyxyu Looking at the two issues here that have been flagged, I think we'll want to hold off on this Ruby upgrade until we can address them.
The issues: 
1. We don't want to downgrade sprockets (we may want to remove it instead) 
2. We don't want to turn off compatibility check for vite. 

Given these could be larger projects, let's timebox at most 2 days to resolving these via upgrades or other means. Otherwise, let's put this Ruby upgrade on hold for now, with the idea that we will revisit it in the future when these libraries have been updated. (The work of this upgrade PR isn't wasted, it's very helpful to flag these issues).

brendanshean:
Let's make sure to upgrade `TargetRubyVersion` in rubocop configuration
for-3 2-upgrade

PR #11464, opened 2024-01-30
Snapshot Timeframes - Standardize Times

dandrabik:
## WHAT
We want the timeframes for the snapshots to be consistent, in particular
1. The period should start at the beginning of a day, and end at the end of a day
2. In most cases, current and previous period should be the same length
3. The periods should be the expected length.
## WHY
For (1) above, this setup makes it easy to re-use the time frames for data that is grouped by date, e.g. upcoming Materialized Views. The rest is just general consistency and accuracy checks (which should make us feel these are solid).
## HOW
Mostly wrote a lot of specs (and spec helpers), then adjusted the code to work.

Note, I am a little worried that these tests may end up being flaky since they are based on current time. But that also might reveal some edge cases, so I'm slightly inclined to put them out and adjust if they are flaky

### Notion Card Links
https://www.notion.so/quill/Make-Snapshot-Timeframes-consistent-824b7335593142359e96c21fbf261c13

PR Checklist | Your Answer
------------ | -------------


anathomical (approved):
I appreciate the layout of the shared examples that make it clear which timeframe types expect current and previous to be equal and which ones allow differences.

brendanshean (approved):
This standardizing is a real improvement.  Thank you for getting deep into the minutiae of time.

services/QuillLMS/spec/support/shared_examples/snapshots/snapshots_timeframe_times_of_day_rspec.rb
brendanshean, line 1: Since these matchers are used elsewhere, you could move matchers into their own directory: `spec/support/matchers` which establish a pattern for custom matchers.  (e.g. I debated using a custom [matcher](https://github.com/empirical-org/Empirical-Core/blob/eaeda4c6b098892b1135621d1088d96f3d686060/services/QuillLMS/spec/support/matchers/change_after_waiting.rb) but ultimately it didn't solve the problem) for a recent intermittent failure )

happythenewsad (approved):

services/QuillLMS/spec/support/shared_examples/snapshots/snapshots_timeframe_times_of_day_rspec.rb
dandrabik: Good call, moved it there.

PR #11472, opened 2024-02-01
fix assignment to nil error and adjust font size

emilia-friedberg:
## WHAT
Fix this bug on Sentry: https://quillorg-5s.sentry.io/issues/4935272728/?project=11238&referrer=github-pr-bot and also some font sizes I noticed.

## WHY
We don't want any errors occurring for our users.

## HOW
This is a weird one, because it's only happening on the `individual_student_diagnostic_responses` path, for which `@skill_group_summaries` should never be defined, and it's inconsistent -- when I wipe the cache I haven't been able to replicate it for any of these users. However, this safeguard should ensure that the error Sentry is reporting cannot happen. 

### Screenshots

### Notion Card Links

PR Checklist | Your Answer
------------ | -------------
Have you added and/or updated tests? |  Manually tested that the page loads correctly


brendanshean (approved):
Thanks for tracking this down

PR #11475, opened 2024-02-01
lms/fix-empty-sentences-written-count

anathomical:
## WHAT
Update new SentencesWritten query logic to return 0 instead of null by using `IFNULL`
## WHY
The old method used COUNT which returns 0 if there are no relevant rows to count, while the new method uses SUM which returns NULL if there are no relevant rows to sum.  
## HOW
Just wrap the rollup in an IFNULL to deal with this case

### Screenshots
OLD
image
NEW
image

### Notion Card Links
https://www.notion.so/quill/On-the-Usage-Snapshot-Report-display-0-for-Sentences-written-if-none-have-been-written-07d8e207913845939448a42b402e103e

PR Checklist | Your Answer
------------ | -------------

### DIFF
--- a/services/QuillLMS/app/queries/snapshots/sentences_written_query.rb
+++ b/services/QuillLMS/app/queries/snapshots/sentences_written_query.rb
@@ -3,7 +3,7 @@
 module Snapshots
   class SentencesWrittenQuery < ActivitySessionCountQuery
     def select_clause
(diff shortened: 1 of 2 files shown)


dandrabik (approved):
:+1: Good catch, thanks for fixing up the bug I surfaced.

PR #11480, opened 2024-02-01
Consolidation db local development configuration

brendanshean:
## WHAT
~~Add command line to script to switch between local and staging db configuration~~
NB:  This approach has been changed using Peter Kong's insight.

Consolidate local development db configuration
 
## WHY
I got tired of commenting and uncommenting these files which is also error prone.  

## HOW
In `.env` set `DATABASE_URL=` whatever the staging db value is and comment out the line.

If you want to use the staging db, simply uncomment the DATABASE_URL value.

If the DATABASE_URL is not set in `.env`, the fallback is `postgres://127.0.0.1:5432/emp_gr_development`.
You can also pass `DATABASE_URL` via a command line

### Screenshots

Uploading toggle-staging-db.mov…

### Notion Card Links

PR Checklist | Your Answer
------------ | -------------
Have you added and/or updated tests? |  No.  Just script.


anathomical (approved):
Neat script.  One note that shouldn't block merging this: we don't really standardize `.env`, which means that the assumptions made here (that there's one commented out and one not-commented out `DATABASE_URL` entry, one of which is `postgresql://postgres@localhost`) that are not universally true.  For instance, my localhost config line is `DATABASE_URL=postgresql://localhost:5432/emp_gr_development`.

Which isn't to say I object to merging this.  In fact, I think we should, and then probably want to do a project at some point to standardize `.env` so that we can make reasonable assumptions about what people have in theirs.

brendanshean:
> Neat script. One note that shouldn't block merging this: we don't really standardize `.env`, which means that the assumptions made here (that there's one commented out and one not-commented out `DATABASE_URL` entry, one of which is `postgresql://postgres@localhost`) that are not universally true. For instance, my localhost config line is `DATABASE_URL=postgresql://localhost:5432/emp_gr_development`.
> 
> Which isn't to say I object to merging this. In fact, I think we should, and then probably want to do a project at some point to standardize `.env` so that we can make reasonable assumptions about what people have in theirs.

My localhost config line is also `DATABASE_URL=postgresql://localhost:5432/emp_gr_development`, I just used the shorter string for differentiating the pattern with the staging db, but I would agree with your point about confirming assumptions on what these keys are.

happythenewsad (commented):
This technique adds extra maintenance for developers. When `database.yml` configs change, this script must also be updated. It will be easy to forget this step. 

We can get the same benefit, without the complexity, by adding this directive to the `development` (or `staging`) clause of `database.yml`:

`url: <%= ENV['DATABASE_URL']  || 'postgres://127.0.0.1:5432/emp_gr_development' %>`

And then invoke it like:

``DATABASE_URL=`heroku config:get DATABASE_URL -a empirical-grammar-staging` ; foreman start Procfile.dev``

brendanshean:
> This technique adds extra maintenance for developers. When `database.yml` configs change, this script must also be updated. It will be easy to forget this step.
> 
> We can get the same benefit, without the complexity, by adding this directive to the `development` (or `staging`) clause of `database.yml`:
> 
> `url: <%= ENV['DATABASE_URL'] || 'postgres://127.0.0.1:5432/emp_gr_development' %>`
> 
> And then invoke it like:
> 
> `` DATABASE_URL=`heroku config:get DATABASE_URL -a empirical-grammar-staging` ; foreman start Procfile.dev ``

I think the worry about extra maintenance is negligible since the development section of the config/database.yml file has been meaningfully modified a couple times in 10 years.

The approach of passing DATABASE_URL key with command line calls is problematic if you open another process (e.g. running rails console or a debugger) since you now have to remember to include the DATABASE_URL keys in two places.  This seems more like a burden on the future developer.

brendanshean:
@happythenewsad Actually, I think your solution points to just consolidating the local staging db and local db configurations.  Thank you for the suggestion.

emilia-friedberg (approved):

emilia-friedberg:
Love this very clean solution!

happythenewsad (approved):

anathomical (approved):

PR #11483, opened 2024-02-02
fix issue with responses falsely showing as human graded on diagnostic and connect cms

emilia-friedberg:
## WHAT
Fix issue with responses falsely showing as human graded on diagnostic and connect cms by using the logic we had in Grammar for all of the apps.

## WHY
We don't want responses that are actually algorithm-matched to show up as human-matched just because they have an empty `<br />` feedback string or something. Elsewhere, the standard we use is around having a boolean value for `optimal`, not `feedback`, so it makes sense to be consistent here.

## HOW
Just pull the logic we were using for Grammar into the "/Shared" bundle, use it everywhere, and get rid of the old files.

Note: I *think* this is the root of the problem from my testing, but I'm not 100% sure because the bug is inconsistent. I thought it might be coming from a difference in the `GradedResponses` `MaterializedView` and the actual `Responses` table in the CMS, but testing didn't really bear that out. If we're still seeing the issue after this fix goes out, I'll take another look at that.

### Screenshots

### Notion Card Links
https://www.notion.so/quill/Rematch-not-working-properly-in-Diagnostic-CMS-5d2c646290c44cb083b31d4796b0ce1e?pvs=4

PR Checklist | Your Answer
------------ | -------------


anathomical (approved):
Love the consolidation.

cissyxyu (approved):

eadams17 (approved):

PR #11484, opened 2024-02-02
send pusher notification to frontend when rematching finishes

emilia-friedberg:
## WHAT
Update the CMS and various app frontends to use Pusher to show an alert when rematching has completed for a given question.

## WHY
Currently there is no indication on the frontend that rematching has completed, leaving our team members unsure about how long the process takes.

## HOW
Update the rematching workers to send a pusher notification to the frontend when the last worker completes, and update the frontend to show an alert when it receives that pusher notification.

### Screenshots
Screenshot 2024-02-02 at 1 32 43 PM

### Notion Card Links
https://www.notion.so/quill/Rematch-Responses-Button-is-not-working-in-the-Diagnostic-Connect-or-Grammar-CMSs-af7035355e2d424bba263cd66fd69ec0?pvs=4

PR Checklist | Your Answer
------------ | -------------
Have you added and/or updated tests? |  (YES


services/QuillCMS/app/workers/rematch_response_worker.rb
```diff
@@ -21,13 +21,17 @@ class RematchResponseWorker
...
-  def perform(response_id, question_type, question_hash, reference_response_ids)
+  def perform(response_id, question_type, question_hash, reference_response_ids, options)
     response = Response.find_by(id: response_id)
     return unless response
 
     reference_responses = Response.where(id: reference_response_ids).to_a
 
     rematch_response(response, question_type, question_hash, reference_responses)
+
+    return unless options['fire_pusher_alert']
```
happythenewsad: changing this line to 

`return unless options['fire_pusher_alert'] && options['question_key']`

removes an edge case where we inadvertently send Pusher events with a null question ID

happythenewsad (approved):

services/QuillCMS/spec/workers/rematch_response_worker_spec.rb
```diff
@@ -128,16 +129,19 @@
         create(:response, params)
       end
     end
+    let(:options) { { 'fire_pusher_alert' => true, 'question_key' => 'some_question_key'} }
```
anathomical: Minor suggestion, but I'd be tempted to have an earlier `let(:question_key) { 'some_question_key}` and then have `'question_key' => question_key` here.  That way you can refer to `question_key` as a variable in your expectation below rather than typing the string literal again and risk that getting messed up somehow.

services/QuillCMS/app/workers/rematch_response_worker.rb
```diff
@@ -21,13 +21,17 @@ class RematchResponseWorker
...
     response = Response.find_by(id: response_id)
     return unless response
 
     reference_responses = Response.where(id: reference_response_ids).to_a
 
     rematch_response(response, question_type, question_hash, reference_responses)
+
+    return unless options['fire_pusher_alert'] && options['question_key']
+
+    RematchingFinished.run(options['question_key'])
```
anathomical: Minor note: the way that this is structured, if there's an error in `RematchingFinished.run` (like, say, Pusher API is down or something), that'll cause this entire job to be put into the retry queue.  So the `rematch_response` function will get called on this specific response again.

The way around this would be to put `RematchingFinished` logic into its own independent worker, and just have this job enqueue that job when it needs to.  That said, I think that the rematch for a single response should be fast enough that if we end up re-running one of them because of a Pusher error, it wouldn't be the end of the world, so it may not be worth refactoring this from a practical perspective even if it would be "better".

anathomical (approved):
A comment about how possible extra work that could be done if a worker errors, but I don't know that it necessarily needs to be fixed.

services/QuillCMS/app/workers/rematch_response_worker.rb
emilia-friedberg: Thanks for flagging, @anathomical. I agree that I think repeating one rematch is pretty low-stakes, but I will keep that approach in mind for the future.

emilia-friedberg: Thanks, @anathomical. I agree that having one rematch job rerun is probably pretty low-stakes, but since it's an easy change I'll go ahead and make it.

PR #11493, opened 2024-02-06
fix rematching individual response bug

emilia-friedberg:
## WHAT
Fix bug where rematching individual responses that were previously human-graded does not correctly update concept results.

## WHY
We want the newly matched response to be accurate.

## HOW
The issue was coming from the fact that, while the response itself gets updated when you unmatch it in the CMS database, we were usually pulling the `GradedResponses` materialized view, which wouldn't be up to date. Therefore, when we rematched, we were rematching against a set of responses that included a graded version of the one we were trying to match, and it was always just finding itself. This new endpoint allows us to get an actually-up-to-date version of the responses, so it'll rematch correctly.

### Screenshots

### Notion Card Links
https://www.notion.so/quill/Rematch-not-working-properly-in-Diagnostic-CMS-5d2c646290c44cb083b31d4796b0ce1e?pvs=4

PR Checklist | Your Answer
------------ | -------------


anathomical (approved):
Good catch.  Also, good solution.  I think it's probably a good idea to have cache-bypass stuff for most things in the CMS so that the admin tools can make sure to be getting fresh data.

PR #11502, opened 2024-02-15
update admin demo data

emilia-friedberg:
## WHAT
Make various changes to admin demo account generation, including:

- adding a cron job to run a task to generate it annually at the start of the school year
- increasing the number of students per classroom to 25
- giving every teacher two classrooms
- adding a `reset` function that destroys the existing admin account and associated schools so running the script multiple times doesn't result in duplicate data
- adding a delay between classroom generation with a default of one minute, because each instance of demo classroom account generation (of which we have 72 here -- two classrooms per teacher for 36 teachers) kicks off a ton of workers, since we enqueue a `SaveActivitySessionConceptResultsWorker` for each activity session completed for that classroom (~200). Running these without a delay on staging actually caused us to run out of redis memory. Staging has many fewer workers allocated and much less redis memory, so I don't think we'd run into that issue on production, but I still didn't want to overload the sidekiq queue unnecessarily.

## WHY
These changes were requested by the partnerships team in order to make the admin data, and specifically the Usage Snapshot Report, more compelling when they demo it to schools.

## HOW
Update the feeder `.yml` file and the services themselves to create the correct data, and the worker and cron to be runnable in that format.

Note on testing: this has been run and rerun repeatedly on staging.quill.org, and I am confident that the data is correct, but it's not possible to view the usage snapshot report for this account on staging because we only use a production instance of BigQuery. I discussed this with Dan and my plan is to launch this to production, but run the generation with an email that is not actually linked to quill.org/admin_demo. Then I can confirm that the usage snapshot report looks the way it should and have the partnerships team confirm before using that data to replace the actual admin demo account. 

### Screenshots

### Notion Card Links
https://www.notion.so/quill/Update-Admin-Demo-Account-For-Partnerships-Team-Sales-Demos-f07a9fe1074e4901adc55c53b960f1a7?pvs=4

PR Checklist | Your Answer
------------ | -------------


services/QuillLMS/app/services/demo/report_demo_creator.rb
```diff
@@ -294,7 +294,7 @@ def self.create_units(teacher, is_teacher_demo)
     ACTIVITY_PACKS_TEMPLATES.map do |activity_pack_template|
       # the following line sets the unit template id to nil for the quill_staff_demo account by request of the partnerships team, because they want to be able to assign the starter baseline recommendations
       # and it ensures the unit template actually exists in our database
-      unit_template_id = is_teacher_demo ? UnitTemplate.find_by_id(activity_pack_template[:unit_template_id])&.id : nil
+      unit_template_id = teacher.email == STAFF_DEMO_EMAIL ? nil : UnitTemplate.find_by_id(activity_pack_template[:unit_template_id])&.id
```
emilia-friedberg, line 297: @brendanshean this effectively reverts a change you made in a recent PR -- I don't think we want to do this for all non-teacher-demo demo accounts (like all the ones generated by the admin_demo script), but specifically only the staff_demo one, so I put it back. Let me know if you'd like to discuss further.

services/QuillLMS/app/workers/demo/reset_admin_demo_account_worker.rb
```diff
@@ -0,0 +1,13 @@
+# frozen_string_literal: true
+
+# destroy admin demo account and recreates it from scratch
+class Demo::ResetAdminDemoAccountWorker
+  include Sidekiq::Worker
+  sidekiq_options retry: false
+
+  ADMIN_DEMO_EMAIL = "hello+demoadmin-admindemoschool@quill.org"
```
anathomical, line 8: I know you didn't pick it, but this email address amuses me.

services/QuillLMS/app/services/demo/create_admin_report.rb
```diff
@@ -4,24 +4,36 @@ class Demo::CreateAdminReport
 
   NUMBER_OF_CLASSROOMS_TO_DESTROY_SESSIONS_FOR = 20
   RANGE_OF_NUMBER_OF_SESSIONS_TO_DESTROY = 14..28 # 10-20% of 140
+  BATCH_DELAY = 1.minute
 
   attr_reader :teacher_email
```
brendanshean: I realize this isn't new code, but what about initializing data and getting rid of the @passed_data instance variable.

```
attr_reader :data, :delay, :teacher_email

def initialize(teacher_email, passed_data=nil, delay=BATCH_DELAY)
    @teacher_email = teacher_email
    @data = passed_data || Demo::SessionData.new.admin_demo_data
    @delay = delay
  end
```

This gets rid of the def data method that has memoization and disjunction which can be difficult to parse.

services/QuillLMS/app/services/demo/report_demo_creator.rb
brendanshean: Would you mind removing the `is_teacher_demo` parameter since it's not needed anymore?

services/QuillLMS/spec/workers/demo/reset_admin_demo_account_worker_spec.rb
```diff
@@ -0,0 +1,20 @@
...
+require 'rails_helper'
+
+describe Demo::ResetAdminDemoAccountWorker, type: :worker do
+  let(:worker) { described_class.new }
+
+  let(:admin_report_instance) { instance_double("Demo::CreateAdminReport") }
+
+  before do
+    allow(Demo::CreateAdminReport).to receive(:new).with(described_class::ADMIN_DEMO_EMAIL).and_return(admin_report_instance)
+    allow(admin_report_instance).to receive(:reset)
```
anathomical: Given that you `expect` this same call in the test below, this line is technically unnecessary.  You could choose to leave it in place to make it easier for theoretical future tests that might need this `allow`, but my inclination is to just remove this line for now and let someone add it back later if they need it.

services/QuillLMS/app/services/demo/create_admin_report.rb
```diff
@@ -64,8 +76,9 @@ def data
       classroom = Classroom.create(name: row['Classroom'])
       all_classrooms.push(classroom)
       ClassroomsTeacher.create(classroom: classroom, user: teacher, role: ClassroomsTeacher::ROLE_TYPES[:owner])
-      student_names = (1..5).to_a.map { |i| "#{Faker::Name.first_name} #{Faker::Name.last_name}" }
+      student_names = (1..25).to_a.map { |i| "#{Faker::Name.first_name} #{Faker::Name.last_name}" }
```
brendanshean: I would extract a constant `NUM_STUDENTS_PER_CLASSROOM` = 25

services/QuillLMS/spec/services/demo/create_admin_report_spec.rb
```diff
@@ -59,17 +59,17 @@
...
 
-    it "should create every classroom from the data hash, associated with the correct teacher, each with five students" do
+    it "should create every classroom from the data hash, associated with the correct teacher, each with 25 students" do
       subject.data.each do |row|
         classroom = Classroom.find_by_name(row['Classroom'])
         teacher = User.find_by_name(row['Teacher'])
         expect(classroom).to be
         expect(classroom.owner).to eq(teacher)
-        expect(classroom.students.count).to eq(5)
+        expect(classroom.students.count).to eq(25)
```
brendanshean: I would reference the `NUM_STUDENTS_PER_CLASSROOM` defined earlier.

services/QuillLMS/app/services/demo/report_demo_creator.rb
```diff
@@ -387,10 +387,15 @@ def self.create_activity_sessions(students, classroom, session_data, is_teacher_
     students.each_with_index do |student, num|
       ACTIVITY_PACKS_TEMPLATES.each do |activity_pack|
         name = unit_name(activity_pack[:name], is_teacher_demo)
-        unit = Unit.where(name:).last
+        unit = Unit.where(name:, user: classroom.owner).last
         classroom_unit = ClassroomUnit.find_by(classroom: classroom, unit: unit)
+
+        # Calculate the index for activity_sessions using modulo
+        # This will cycle through the activity_sessions for more students
+        activity_sessions_index = num % activity_pack[:activity_sessions].length
```
anathomical, line 395: This is a clever solution to the problem of there now being more students to fake data for than students we're referencing for fake data.

services/QuillLMS/app/services/demo/create_admin_report.rb
```diff
@@ -4,24 +4,36 @@ class Demo::CreateAdminReport
 
   NUMBER_OF_CLASSROOMS_TO_DESTROY_SESSIONS_FOR = 20
   RANGE_OF_NUMBER_OF_SESSIONS_TO_DESTROY = 14..28 # 10-20% of 140
+  BATCH_DELAY = 1.minute
```
brendanshean, line 7: Can you add a note in the PR description about the purpose of the BATCH_DELAY, in other words, why it is necessary?

brendanshean (approved):
Some minor refactoring comments, but overall looks good.  Thank you for setting this up!  It will be nice to have some data in here for snapshots.

services/QuillLMS/app/services/demo/create_admin_report.rb
```diff
@@ -4,24 +4,36 @@ class Demo::CreateAdminReport
...
+    School.where(mail_street: school_street).destroy_all
+    create_demo
+  end
+
   def data
     @data ||= @passed_data || Demo::SessionData.new.admin_demo_data
   end
 
+  private def school_street
+    @school_street ||= "#{Time.zone.now.year} Demo School Lane"
```
anathomical, line 29: Do we need to be worried about the `reset` logic breaking when the year changes?  When the yearly job runs, it won't be able to find the "old" demo school anymore to blow up.  (Maybe that's fine, but seemed worth checking.)

services/QuillLMS/app/services/demo/report_demo_creator.rb
emilia-friedberg: @brendanshean it's still getting used in line 298!

services/QuillLMS/app/services/demo/create_admin_report.rb
```diff
@@ -64,8 +76,9 @@ def data
       classroom = Classroom.create(name: row['Classroom'])
       all_classrooms.push(classroom)
       ClassroomsTeacher.create(classroom: classroom, user: teacher, role: ClassroomsTeacher::ROLE_TYPES[:owner])
-      student_names = (1..5).to_a.map { |i| "#{Faker::Name.first_name} #{Faker::Name.last_name}" }
+      student_names = (1..25).to_a.map { |i| "#{Faker::Name.first_name} #{Faker::Name.last_name}" }
       Demo::ReportDemoCreator.create_demo_classroom_data(teacher, is_teacher_demo: false, classroom: classroom, student_names: student_names)
+      sleep @delay
```
anathomical, line 78: Using a `sleep` call to make sure that callbacks and everything get completed before doing another round of inserts is a fine way to do this, but the longer the task runs the higher the chance that something will go wrong and the job will crash out and go into the retry queue.  Given that we only run this once a year, I think that if that does happen and the whole job runs a second time from scratch would be just fine.

A more robust way of handling this is to actually make `Demo::ReportDemoCreator.create_demo_classroom_data` another worker and just enqueuing those workers individually into Sidekiq.  If we want to ensure that the associated workers all have time to finish before picking up one of these, we might even move this job into a lower priority queue so that all the spin-off jobs get worked on before another demo classroom gets made.

Anyway, I think that since this is a once-a-year thing, it's probably fine as-is, but if we were going to call it more often I'd rather handle the timing control by enqueuing jobs in Sidekiq directly.

services/QuillLMS/app/services/demo/create_admin_report.rb
emilia-friedberg: Explanation added! Thanks for flagging.

brendanshean: Thank you.  That is super helpful.

services/QuillLMS/app/services/demo/report_demo_creator.rb
brendanshean: Sorry.  My mistake. :grimacing:

anathomical (approved):

services/QuillLMS/app/services/demo/create_admin_report.rb
emilia-friedberg: That's a good question. I just confirmed that `.destroy_all` doesn't error if called on an empty array, so it should be fine, but thank you for checking.

services/QuillLMS/app/services/demo/create_admin_report.rb
emilia-friedberg: Hey @anathomical -- so this is a good point, and I tried to rewrite it your way, but some other issues came up that made testing it difficult and since there is some time sensitivity here I think it's better to just get it up the way it is and rewrite in the future if necessary. Specifically, I started getting Redis:OOO memories again on staging (the issue the batch delay was added to correct), but because each instance of the demo creation was a worker, this resulted in a bajillion concept result workers getting created again on every retry. I think there are various ways around this, including adding a new queue like you suggested, but for now I think we should leave as is.

anathomical: @emilia-friedberg That seems reasonable to me.

PR #11510, opened 2024-02-16
Materialized View Refresher

dandrabik:
## WHAT
This adds an endpoint, worker, service, and config structure to refresh a Materialized View in BigQuery by a post request.

## WHY
We need a way to refresh materialized views in BigQuery since certain Airbyte operations (dropping tables) make them unusable. We are going to use a webhook in Airbyte to trigger this. Breaking this off from other Mat View code since it can be released independently.
## HOW
I aimed to make this extensible for other table refreshes.
1. Added a generic config/folder structure (so the next refresh job can just add a config and put SQL in the desginated folder)
2. Made a BigQuery user with write permissions that is only used by this endpoint.
3. Added an api_key that will be stored in an ENV to secure the endpoint.
4. Structure the code to make it reasonably testable without hitting bigquery

### Notion Card Links
https://www.notion.so/quill/Mat-View-Refresh-Logic-395476a3e5194afc956af3618f0ec9b4

PR Checklist | Your Answer
------------ | -------------


services/QuillLMS/app/controllers/big_query/materialized_views_controller.rb
brendanshean, line 1: Can the module `BigQuery` be folded into `QuillBigQuery` or is this a reason for the separation?

services/QuillLMS/spec/services/quill_big_query/materlialized_view_refresher_spec.rb
```diff
@@ -0,0 +1,43 @@
...
+  describe 'config' do
+    subject {described_class.new(query_key)}
+
+    it 'should find SQL file' do
+      expect { subject.send(:create_sql)}.not_to raise_error
+    end
+  end
+
+  describe 'run' do
+    let(:report_sessions_sql) {'SELECT 2'}
```
brendanshean: Is this line needed?

brendanshean (approved):
I like the pattern you have setup here especially using the Configs for managing sql since that will allow for quick updates with minimal code.  Nice work on this.

services/QuillLMS/app/controllers/big_query/materialized_views_controller.rb
dandrabik: The thought I had here was this is a controller, so the default namespace will appear in the url, so it would be, `www.quill.org/quill_big_query/refresh` instead of `www.quill.org/big_query/refresh`. To be honest, since this is technically external, I debated doing something more generic like `data/refresh` or `data_views/refresh` to remove implementation details, but thought that might be more confusing internally, so landed on `big_query`. If others find it confusing, I could change it to `QuillBigQuery` and change the default namespace path.

services/QuillLMS/spec/services/quill_big_query/materlialized_view_refresher_spec.rb
dandrabik: Ah good catch, my intention was to use this for documentation that that is coming from the SQL file, which might be less obvious when this is not in PR, but I left it out. Let me put it back in.

services/QuillLMS/app/services/quill_big_query/materialized_view_refresher.rb
```diff
@@ -0,0 +1,45 @@
...
+    end
+
+    # Adding a 'SELECT 1;' query since the BigQuery library errors when returning the contents
+    # of a Materialized View, i.e. it runs the 'CREATE' operation successfully,
+    # but the API errors when parsing the Materialized View details to return:
+    # "Google::Apis::ClientError: invalid: Cannot list a table of type MATERIALIZED_VIEW."
+    private def create_view
+      <<-SQL.squish
+        CREATE MATERIALIZED VIEW #{name} #{create_options} AS (#{create_sql.squish});
+        SELECT 1;
```
anathomical: Potential alternative: `SELECT COUNT(*) FROM #{name}`.  One thing I've noticed in testing is that defining a materialized view doesn't actually populate it.  It's only on the first request that uses it that the view actually seems to get built (at least for non-incremental views).  So doing a query against the view should ensure that it gets "warmed up" on refresh.

It's probably not a big deal if we do this the way you already have it, but I figured I'd throw out the idea.

services/QuillLMS/app/services/quill_big_query/materialized_view_refresher.rb
```diff
@@ -0,0 +1,45 @@
...
+      SQL
+    end
+
+    private def drop_view
+      "DROP MATERIALIZED VIEW IF EXISTS #{name}"
+    end
```
anathomical: Thoughts on this being a one-liner?  I am generally digging that syntax.
```
private def drop_view "DROP MATERIALIZED VIEW IF EXISTS #{name}"
```

services/QuillLMS/db/big_query/views/reporting_sessions_view.sql
```diff
@@ -0,0 +1,38 @@
+SELECT
+ activity_sessions.user_id as student_id,
+ activity_sessions.activity_id as activity_id,
+ activities.name as activity_name,
+ schools.id as school_id,
+ schools.name as school_name,
+ classrooms_teachers.user_id as teacher_id,
+ users.name as teacher_name,
+ classrooms.id as classroom_id,
```
anathomical, line 9: Maybe worth including `classrooms.name AS classroom_name` in this table to finish the denormalization?

services/QuillLMS/db/big_query/views/reporting_sessions_view.sql
```diff
@@ -0,0 +1,38 @@
...
+ classrooms.grade as grade,
+ DATETIME_TRUNC(activity_sessions.completed_at, DAY) as completed_date,
```
anathomical, line 12: Are we getting an advantage out of truncating this value?  If so, should we try to use this truncation in other admin reports, too?

services/QuillLMS/db/big_query/views/reporting_sessions_view.sql
```diff
@@ -0,0 +1,38 @@
...
+ INNER JOIN lms.classroom_units ON classroom_units.id = activity_sessions.classroom_unit_id
+ INNER JOIN lms.classrooms ON classrooms.id = classroom_units.classroom_id
+ INNER JOIN lms.classrooms_teachers ON classrooms_teachers.classroom_id = classrooms.id
+ INNER JOIN lms.schools_users ON schools_users.user_id = classrooms_teachers.user_id
+ INNER JOIN lms.schools ON schools.id = schools_users.school_id
+ INNER JOIN lms.users on users.id = classrooms_teachers.user_id
+WHERE
+  classrooms_teachers.role = 'owner'
+  AND activity_sessions.completed_at IS NOT NULL
+  AND activity_sessions.completed_at > '2021-07-31'
```
anathomical, line 27: No objection to setting a reasonable date cut-off like this, but should we be worried about this "invisibly" reducing the timeframe reporting available?  For instance, "all-time" isn't all time when queried against this view.

services/QuillLMS/spec/services/quill_big_query/materlialized_view_refresher_spec.rb
```diff
@@ -0,0 +1,45 @@
...
+describe QuillBigQuery::MaterializedViewRefresher do
+  let(:query_key) { 'reporting_sessions_view' }
+
+  describe 'config' do
+    subject {described_class.new(query_key)}
+
+    it 'should find SQL file' do
+      expect { subject.send(:create_sql)}.not_to raise_error
+    end
+  end
```
anathomical: Not having a `before` block to `stub_const` the `QUERY_FOLDER` value before this `describe` means that we're testing for the actual file, not the spec fixture.  That might be intentional here, but if so it means that the value in `query_key` is coupled to actual configuration.  Which may be exactly what you're intending, so I just wanted to confirm that that's the case.

anathomical (approved):
This looks good to me.  I appreciate the use of the yaml config for extensibility.  I also like that each view requires a separate HTTP call to refresh.  The added flexibility that gives us over some sort of "refresh them all" seems like a good thing.

I have a couple of comments, but I think they're all either requests for clarification or take-it-or-leave it items.

services/QuillLMS/app/controllers/big_query/materialized_views_controller.rb
brendanshean: While I do find the `www.quill.org/quill_big_query/refresh` less aesthetically pleasing than `www.quill.org/big_query/refresh`.  I think having some files using the BigQuery an namespace and not others is surprising and/or confusing.  For example, in this worker:

```
module BigQuery
  class MaterializedViewRefreshWorker
    include Sidekiq::Worker

    def perform(query_key)
      QuillBigQuery::MaterializedViewRefresher.run(query_key)
    end
  end
end
```
it's not clear why it's in the module `BigQuery` whereas the service object is in `QuillBigQuery`.  Down the road, there's now a choice for developers to decide which namespace to use which in my mind doesn't really offer much benefit..  Why not have all QuillBiqQuery code in the same namespace.  Also, the name is different enough that it's not to be confused with third-party code `Google::Cloud::Bigquery`.

services/QuillLMS/app/services/quill_big_query/materialized_view_refresher.rb
happythenewsad: ^^^ Good idea

services/QuillLMS/app/services/quill_big_query/materialized_view_refresher.rb
```diff
@@ -0,0 +1,45 @@
+# frozen_string_literal: true
+
+module QuillBigQuery
+  class MaterializedViewRefresher < ApplicationService
+    attr_reader :query_key
+
+    QUERY_FOLDER = Rails.root.join('db/big_query/views/')
+    CONFIG = Configs[:big_query_views]
```
happythenewsad, line 8: Is there a reason to use the `Evidence::Configs` initializer here? Seems like an unnecessary indirection. How about `BigQuery::Configs`?

happythenewsad (approved):
Looks good, apart from a naming suggestion

services/QuillLMS/app/services/quill_big_query/materialized_view_refresher.rb
dandrabik: I think there might be some confusion here. `configs` is a [small gem](https://github.com/kickstarter/configs) that is helpful for querying `yaml` files. This isn't a reference to Evidence::Configs but the Configs gem.

services/QuillLMS/db/big_query/views/reporting_sessions_view.sql
dandrabik: I had asked PS about this a while ago and he was on board, but worth re-visiting. I could also edit this in a later PR when we start to use this code, so not considering this a blocker.

services/QuillLMS/db/big_query/views/reporting_sessions_view.sql
dandrabik: I believe this makes it slightly more efficient to query by the simpler date, but I've added so much grouping to the query since I originally added this so it might not have that much value (so it could be a take or leave). There is something nice about making the column a simple date since our time filters will always be a day granularity. So this is a reasonable shape.

dandrabik: Will leave this for now, but can edit/change this in the PR to use this code.

services/QuillLMS/app/services/quill_big_query/materialized_view_refresher.rb
happythenewsad: Got it, carry on

services/QuillLMS/app/services/quill_big_query/materialized_view_refresher.rb
dandrabik: Will do!

services/QuillLMS/app/services/quill_big_query/materialized_view_refresher.rb
dandrabik: Will do

services/QuillLMS/app/controllers/big_query/materialized_views_controller.rb
dandrabik: Ok, you've convinced me. I've update all of the uses the consistent QuillBigQuery.

services/QuillLMS/spec/services/quill_big_query/materlialized_view_refresher_spec.rb
dandrabik: This is intentional. I've updated the test description, but basically this is testing the production file path that is being stubbed in all other tests. That stub could cover a production bug (it was hiding one in my initial code), and this helps give some coverage to that.

PR #11511, opened 2024-02-16
Fix bug with quill_staff_demo assigning recommendations

brendanshean:
## WHAT
Fix a bug where the staff account is unable to assign recommendations

## WHY
When a staff demo account currently attempts to [assign](https://github.com/empirical-org/Empirical-Core/blob/develop/services/QuillLMS/app/services/independent_practice_packs_assigner.rb#L71) recommendations, a Unit name validation error occurs because the a unit with that name was already created in the `DemoReportCreator.create_demo(STAFF_DEMO_EMAIL)` daily cron job.

## HOW
Update the `DemoReportCreator.create_demo(STAFF_DEMO_EMAIL)` to create units with names that append "(STAFF DEMO)" so that there is no conflict during the assign recommendations flow.

### Screenshots

### Notion Card Links
https://www.notion.so/quill/Cannot-assign-Practice-recommendation-activities-from-quill-org-quill_staff_demo-account-fb35eb78ba734155a6f2753c5f0e638b?pvs=4

PR Checklist | Your Answer
------------ | -------------
Have you added and/or updated tests? |  No.  Manual check

### DIFF
--- a/services/QuillLMS/app/services/demo/report_demo_creator.rb
+++ b/services/QuillLMS/app/services/demo/report_demo_creator.rb
@@ -305,7 +305,7 @@ module Demo::ReportDemoCreator
   end
 
   def self.unit_name(name, is_teacher_demo)
-    is_teacher_demo ? name : "#{name} (Staff Demo)"
+    is_teacher_demo ? name : "#{name} (Demo)"
   end
 
   def self.activity_ids_for_config(template_hash)
--- a/services/QuillLMS/app/services/demo/report_demo_creator.rb
+++ b/services/QuillLMS/app/services/demo/report_demo_creator.rb
@@ -170,42 +170,40 @@ module Demo::ReportDemoCreator
   STUDENT_COUNT = STUDENT_TEMPLATES.count
   UNITS_COUNT = ACTIVITY_PACKS_TEMPLATES.count
 
-  def self.create_demo(email = nil, teacher_demo: false)
+  def self.create_demo(email = nil, is_teacher_demo: false)
     ActiveRecord::Base.transaction do
       teacher = create_teacher(email)
       create_teacher_info(teacher)
       create_subscription(teacher)
-      create_demo_classroom_data(teacher, teacher_demo: teacher_demo)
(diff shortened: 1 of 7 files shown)


services/QuillLMS/app/services/demo/report_demo_creator.rb
```diff
@@ -292,19 +290,24 @@ def self.non_demo_classrooms(teacher)
...
+      unit = Unit.find_or_create_by(name:, user: teacher, unit_template_id:)
+      activity_ids = activity_ids_for_config(activity_pack_template)
+      activity_ids.each { |activity_id| UnitActivity.find_or_create_by(activity_id:, unit:) }
 
       unit
     end
   end
 
+  def self.unit_name(name, is_teacher_demo)
+    is_teacher_demo ? name : "#{name} (Staff Demo)"
```
emilia-friedberg: My only concern here is that the staff demo isn't the only time we pass `is_teacher_demo: false` to this file. Right now, we do it for the admin demo reports too. I might actually be changing this in an upcoming PR, but I think it might be better to just do `(Demo)` here to avoid confusion in the future?

emilia-friedberg (approved):
Approving but with a note about the unique identifier for the pack.

emilia-friedberg (commented):
One other thing actually -- if we change the argument name from `teacher_demo` to `is_teacher_demo`, do we also need to change its invocation? I think we call both `create_demo` and `create_demo_classroom_data` outside of this file.

brendanshean:
> One other thing actually -- if we change the argument name from `teacher_demo` to `is_teacher_demo`, do we also need to change its invocation? I think we call both `create_demo` and `create_demo_classroom_data` outside of this file.

Good point.  You beat [circleci](https://app.circleci.com/pipelines/github/empirical-org/Empirical-Core/27044/workflows/e442cfa2-266f-420f-9718-33899c83dd61/jobs/302057?utm_campaign=vcs-integration-link&utm_medium=referral&utm_source=github-checks-link&utm_content=summary) in catching this.

PR #11514, opened 2024-02-16
Add QA section to PR template.

dandrabik:
## WHAT
I'm adding a main section to the PR for the code writer to give QA details, i.e. how did you verify that this feature works properly outside of automated tests. 

Some examples of a good description:
- (Adding a new worker): I triggered the worker in the console on staging with a variety of inputs (nil, one, many) and verified that the data in the database was changed the correct way each time by the job.
- (Adding a new frontend page): I verified the page loaded properly on staging and re-read through the spec and confirmed the requirements were met, visually and logically. I had Jack review the visual elements as well and he approved.
- (Changing a universal component): I verified the new form element renders properly in all of its uses across the site (student, teacher, admin, and staff views). I tested each form in chrome and safari on desktop, and the chrome mobile view (phone size). I tested submitting using the new form element on 4 forms (one student, one teacher, one admin, one staff internal tool). 

Some examples of a lacking description:
- I clicked around a bunch
- Tests pass
- I did a basic QA on staging of the feature.
## WHY
We've had a few bugs slip through the QA process. QA can also be really tricky to determine whether one has done enough. I like the idea of adding it to the PR template so that others can "review" the process. So this achieves two goals:
1. It's a stopping point for the code writer to think about whether they have done enough QA.
2. It's a chance for reviewers to point out potential areas for more QA. 
## HOW
Updated the PR template

I did a basic proofread (and pasted into a doc for a spellcheck). Then I used the new template for this PR.

PR Checklist | Your Answer
------------ | -------------
Have you added and/or updated tests? |  No, non-app code
Spec Review: Have you reviewed the spec and ensured this meets requirements and/or matches design mockups? | Debating whether we should remove this section with the new section in place.


anathomical (approved):
I really like this as a way to allow reviewers to suggest angles of QA that may have been missed by the original feature writer.  I know that I, at least, will appreciate having folks help catch use-cases that I might miss.

pull_request_template.md
```diff
@@ -10,6 +10,9 @@
 ### Notion Card Links
 (Please provide links to any relevant Notion card(s) relevant to this PR.)
 
+### What have you done to QA this feature?
```
brendanshean, line 13: Maybe add a link to this PR or some document with a list of examples for QA in case people need a refresher? 
This keeps the size of the PR template down but provides breadcrumbs for self-analysis.

brendanshean (approved):
I think this is a good idea.  Also, spec review at the bottom feels kind of redundant with this so probably not needed.

emilia-friedberg (approved):
I agree this probably makes "spec review" redundant, but I also don't mind editing a one word response if we leave that row in the checklist.

pull_request_template.md
eadams17: I like the idea of linking back to this PR to keep things more concise.

eadams17 (approved):

happythenewsad (approved):

PR #11521, opened 2024-02-21
fix inline editing for incorrect sequences and focus points

emilia-friedberg:
## WHAT
Change the values we're using for keys in incorrect sequence and focus point components so that curriculum team members can edit the texts inline while also being able to reorder them.

Also fix bug where entering a space would result in weird behavior, due to dnd-kit using space commands for accessibility purposes.

## WHY
We want the curriculum team members to be able to edit these things inline.

## HOW
This problem was caused by the fact that we were using the incorrect sequence/focus point's text as the element's key, resulting in the whole thing rerendering when the text changed (by editing it) and focus being lost. The fix here is fairly straightforward for focus points, because they already have keys that are actual unique identifiers which we could use instead. It doesn't work well for incorrect sequences, however, because their "keys" in the database are just indices, which don't play well with the drag and drop library. Ultimately, I think it would be best to migrate to using a structure more like the one we have for focus points, but that is a major rewrite, so in order to solve this bug while preserving reorderability I've opted to just use temporary uids in these components. These uids do not actually save to the database due to the code [here](https://github.com/empirical-org/Empirical-Core/blob/develop/services/QuillLMS/app/controllers/api/v1/incorrect_sequences_controller.rb#L49-L50). I'm open to other suggestions if we don't like this approach.

See photo of a question's data structure with both focus points and incorrect sequences here:
Screenshot 2024-02-21 at 2 58 13 PM

### Screenshots

### Notion Card Links
https://www.notion.so/quill/Inline-editing-not-working-in-Connect-CMS-b6c43cf969d440cc81f85610e3728023?pvs=4

PR Checklist | Your Answer
------------ | -------------
Have you deployed to Staging? | Tested locally with staging db

### DIFF
--- a/services/QuillLMS/client/app/bundles/Connect/components/focusPoints/focusPointsContainer.jsx
+++ b/services/QuillLMS/client/app/bundles/Connect/components/focusPoints/focusPointsContainer.jsx
@@ -64,7 +64,7 @@ export class FocusPointsContainer extends Component {
     const { fpOrderedIds, focusPoints } = this.state
     if (fpOrderedIds) {
       const focusPointsCollection = hashToCollection(focusPoints)
-      return fpOrderedIds.map(id => focusPointsCollection.find(fp => fp.text === id))
+      return fpOrderedIds.map(id => focusPointsCollection.find(fp => fp.key === id))
     } else {
       return hashToCollection(focusPoints).sort((a, b) => a.order - b.order);
     }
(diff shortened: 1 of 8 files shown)


brendanshean (approved):
Nice sleuthing on this one.  I didn't notice it was using the text as key which would explain why my approach of adding more fine-grained ids with sequenceString array didn't work. 

Regarding the 
> Note: it looks like there is still some weird behavior when you try to add spaces in either the name or the sequences, but I think it's unrelated. Working on figuring that one out now.

That might be from the regex split [here](https://github.com/empirical-org/Empirical-Core/blob/develop/services/QuillLMS/client/app/bundles/Connect/components/incorrectSequence/incorrectSequenceContainer.jsx#L227)

anathomical (approved):
I agree that long-term we should have Incorrect Sequences work the same way on the back-end as Focus Points, but I also agree that in the short term this is a clean solution that accomplishes what we want without having to do some sort of big migration.

PR #11526, opened 2024-02-22
fix usage snapshot report bugs

emilia-friedberg:
## WHAT
Fix two usage snapshot report bugs: one where going from a report with previous values to "All time" keeps showing arrows indicating change, and one where switching quickly between time frames results in the switched-away from report loading after the switched-to report is already in place. 

## WHY
We want these numbers to accurately reflect the filter.

## HOW
Make sure to clear the `previousCount` when `searchCount` is changed, and also clear any existing timeouts so that 

### Screenshots

### Notion Card Links
https://www.notion.so/quill/Don-t-show-the-percentage-change-values-on-the-Usage-Snapshot-Report-on-the-All-time-timeframe-7203bfcf82ae4e68bc3276e97740a07e?pvs=4

PR Checklist | Your Answer
------------ | -------------


anathomical (approved):
Good catch.

PR #11528, opened 2024-02-22
lms/retry-missing-bigquery-views

anathomical:
## WHAT
This is intended more as an RFC with working sample code than a PR for merging, so discussion highly encouraged.

Allow our `QuillBigQuery::Query` classes to specify a list of which BigQuery materialized views that they reference.  Then, when making the query, detect if an error is returned from BigQuery indicating that any of the described views are broken or missing, and if they are retry the query replacing all the materialized views with their raw SQL.
## WHY
We want our BigQuery queries that use materialized views to have fallbacks that allow them to keep working (rather than just crashing) if the underlying materialized views are missing or being rebuilt.  Our original plan of using `WITH` queries worked in some cases, but not all, so this approach does the fallback in Ruby rather than BigQuery.
## HOW
- Add a new `materialized_views_used` value to our `QuillBigQuery::Query` class
- Add a `rescue` block to `run_query` that rescues the two types of errors that BigQuery will return when something goes wrong pulling data from a view
- - In the rescue, check the error message to see if it references once of the registered materialized views
- - If so, retry the query, but substitute out the materialized view for the raw SQL used to generate the view

PR Checklist | Your Answer
------------ | -------------
Have you deployed to Staging? | No.  This code is intended for discussion, and is not ready for merging


anathomical:
Closing in favor of a different approach: #11547

PR #11535, opened 2024-02-26
Only check diagnostic responses for focus points and incorrect sequences

cissyxyu:
## WHAT
We no longer want to label incorrect diagnostic responses with various matchers like Minimum Length Hint, Whitespace Hint, etc. Because diagnostic responses don't show students any feedback, we're not concerned with what the type of hint they received was. Instead, we only want to match these responses with incorrect sequences and focus points.

## WHY
To simplify the grading process for diagnostic questions.

## HOW
Remove all checkers from the marking logic code except focus points and incorrect sequences.

### Screenshots

### Notion Card Links
https://www.notion.so/quill/Product-Board-30f97e4eb01246dbb8264253fb385073?p=7efb2c2930024679a943d999fd243de0&pm=c

I've already implemented the npm unit tests. Once the code changes are approved, I will publish the new npm package, then deploy to staging and manually test in the Diagnostic admin panel to make sure responses are matched correctly. 
I will manually test:
-when a student enters a correct response.
-when a student enters a response triggering a focus point
-when a student enters a response triggering an incorrect sequence
-when a student's response is incorrect and doesn't trigger any sequences or focus points, so it should simply be marked as incorrect

PR Checklist | Your Answer
------------ | -------------
Have you deployed to Staging? | Not yet - will deploy once code is approved because I don't want to publish un-approved code on npm.


emilia-friedberg (approved):
This looks good to me overall, but I think maybe we should talk as a team about the process with npm modules -- I'm not sure it makes sense to have the code reviewed before every publication, especially since nobody else is using it and our various apps are always set to use a pinned version of the library. I appreciate your caution here, but I think if you or anyone else needed to do several iterations for things you'd catch yourself in QA, it just starts to feel unnecessarily onerous.

Another thing I wanted to ask is that you QA to make sure the rematching lambda works -- staging.quill.org is currently connected to the staging version of the lambda, so it should be pretty straightforward to test there and then deploy to the production lambda when you're deploying the rest of the code to production.

Finally, just to confirm, we are not making changes to fill in the blank and sentence fragment questions as part of this project?

anathomical (approved):
Looks good.  One thing we should probably do (and probably should have been doing for a while) is to make sure that we also update `lambdas/rematching/package.json` and plan to redeploy our lambdas, too, since they use `quill-marking-logic`.

cissyxyu:
> This looks good to me overall, but I think maybe we should talk as a team about the process with npm modules -- I'm not sure it makes sense to have the code reviewed before every publication, especially since nobody else is using it and our various apps are always set to use a pinned version of the library. I appreciate your caution here, but I think if you or anyone else needed to do several iterations for things you'd catch yourself in QA, it just starts to feel unnecessarily onerous.
> 
> Another thing I wanted to ask is that you QA to make sure the rematching lambda works -- staging.quill.org is currently connected to the staging version of the lambda, so it should be pretty straightforward to test there and then deploy to the production lambda when you're deploying the rest of the code to production.
> 
> Finally, just to confirm, we are not making changes to fill in the blank and sentence fragment questions as part of this project?

Sounds good, I'll make sure to test the rematching lambda as well. 

Yes -- we're not changing the fill in blank and sentence fragment questions grading for now, just diagnostic questions.

PR #11536, opened 2024-02-26
Refactor diagnostic responses components for admin CMS

cissyxyu:
## WHAT
Refactor the two components `Response` and `ResponseList` into typescript and the latest React conventions. The impetus is various bugs in these components -- right now, when responses are rematched or edited, it triggers a page refresh and newly searches responses to update the entire page. This search also shuffles all the responses because search is not predictably ordered.

We don't have to do a new search every time one single response is edited, we can simply update the data for that one response. In order to do this, I've updated the parent component `responseComponent` to store responses in state and handle the updating of a single response. (we should eventually also refactor this component into typescript and latest React).

I've also fixed various small bugs I found in this page while refactoring.

## WHY
Admin don't want the entire page to refresh every time they simply update one response.

## HOW
See above.

### Screenshots

### Notion Card Links
https://www.notion.so/quill/Responses-shuffle-in-diagnostic-CMS-upon-every-autosave-f98d8d08910545c2956553ddc9142b1f?pvs=4

This refactor is a larger code change that touches all parts of the response code, so it needs thorough testing. I've asked Rachel to join me in doing QA testing and trying out all parts of the Response and Responses panel that admin typically use while editing a question. Rachel and I are both doing this sanity check through staging QA tests.

PR Checklist | Your Answer
------------ | -------------


services/QuillLMS/client/app/bundles/Diagnostic/libs/grading/rematching.ts
```diff
@@ -75,14 +86,14 @@ export function rematchAll(mode: string, question: Question, questionID: string,
   });
 }
 
-export function rematchOne(response: string, mode: string, question: Question, questionID: string, callback:Function) {
+export function rematchOne(response: Response, mode: string, question: Question, questionID: string, callback:Function) {
```
emilia-friedberg: This was actually just incorrect previously and `response` was never a string here, right?

services/QuillLMS/client/app/bundles/Diagnostic/libs/grading/rematching.ts
```diff
@@ -75,14 +86,14 @@ export function rematchAll(mode: string, question: Question, questionID: string,
...
   const matcher = getMatcher(mode);
   getGradedResponses(questionID).then((data) => {
     question.key = questionID
     const matcherFields = getMatcherFields(mode, question, formatGradedResponses(data));
     const promise = rematchResponse(matcher, matcherFields, response);
     if (promise) {
-      promise.then(() => { callback(); });
+      promise.then((result) => { callback(result); });
```
emilia-friedberg, line 96: It doesn't look like we're doing something with the `result` in `searchResponses`, am I missing something?

services/QuillLMS/client/app/bundles/Diagnostic/components/questions/response.tsx
```diff
@@ -0,0 +1,446 @@
...
+
+  let conceptResultsTemp = {}
+  if (passedResponse.concept_results) {
+    if (typeof passedResponse.concept_results === 'string') {
+      conceptResultsTemp = JSON.parse(passedResponse.concept_results)
+    } else {
+      conceptResultsTemp = passedResponse.concept_results
+    }
+  }
+  const [conceptResults, setConceptResults] = React.useState(conceptResultsTemp)
```
emilia-friedberg: I would refactor this section so you have a function outside of the component that looks like this:

```
function extractConceptResultsFromResponse(response) {
  const { concept_results, } = response

  if (!concept_results) { return {} }

  if (typeof concept_results === 'string') {
    return JSON.parse(concept_results)
  } else {
    return concept_results
  }
}
```

And then line 64 can just be 

```
const [conceptResults, setConceptResults] = React.useState(extractConceptResultsFromResponse(passedResponse))
```

services/QuillLMS/client/app/bundles/Diagnostic/components/questions/response.tsx
```diff
@@ -0,0 +1,446 @@
...
+    const { concept_results, } = response;
+    let conceptResults = {}
+    if (concept_results) {
+      if (typeof concept_results === 'string') {
+        conceptResults = JSON.parse(concept_results)
+      } else {
+        conceptResults = concept_results
+      }
+    }
+    setConceptResults(conceptResults)
```
emilia-friedberg: Once you have the function above, you can replace this code with an invocation of that as well.

services/QuillLMS/client/app/bundles/Diagnostic/components/questions/responseComponent.jsx
```diff
@@ -72,6 +72,8 @@ class ResponseComponent extends React.Component {
     } else if (this.props.states[this.props.questionID] === C.SHOULD_RELOAD_RESPONSES && prevProps.states[prevProps.questionID] !== C.SHOULD_RELOAD_RESPONSES) {
       this.props.dispatch(questionActions.clearQuestionState(this.props.questionID));
       this.searchResponses();
+    } else if (!_.isEqual(this.props.filters.responses, prevProps.filters.responses)) {
+      this.setState({responses: this.props.filters.responses})
```
emilia-friedberg: Even though this file isn't up to this standard in general, I still like to go with the linter standards on new code  (like extracting `{ filters, }` from `this.props` and using that in these lines) because it does make it easier to read.

services/QuillLMS/client/app/bundles/Diagnostic/components/questions/responseList.tsx
```diff
@@ -0,0 +1,130 @@
...
+  states: Object,
+  updateResponse: Function,
+}
+
+
+const ResponseList = ({admin, ascending, concepts, dispatch, expand, expanded, getChildResponses, getResponse, massEditResponses, mode, question, questionID, responses, selectedFocusPoints, selectedIncorrectSequences, states, updateResponse }: ResponseListProps) => {
+
+  function allResponsesChecked() {
+    return !responses.some((r) => {
+      return !(
```
emilia-friedberg: This double negative is a little confusing to me. I think we could just do `.every` instead of `.some` here without any bang operators, right?

services/QuillLMS/client/app/bundles/Diagnostic/components/questions/responseList.tsx
```diff
@@ -0,0 +1,130 @@
...
+        state={states[questionID]}
+        states={states}
+        updateParentResponse={(id, response) => updateResponse(id, response)}
+      />
+    )
+  }
+
+  function isValidAndNotEmptyRegex(string) {
+    return string.length && isValidRegex(string)
+  }
```
emilia-friedberg: This can get moved outside of the component entirely.

services/QuillLMS/client/app/bundles/Diagnostic/components/questions/response.tsx
```diff
@@ -0,0 +1,446 @@
...
+        (<a className="card-footer-item" key="update" onClick={() => updateResponse(response.key)} >Update</a>)
+      ];
+    } else {
+      buttons = [
+        (<a className="card-footer-item" key="edit" onClick={() => editResponse(response.key)} >Edit</a>),
+        (<a className="card-footer-item" key="delete" onClick={() => handleDeleteResponse(response.key)} >Delete</a>)
+      ];
+    }
+    if (statusCode > 1) {
+      buttons = buttons.concat([(<a className="card-footer-item" key="rematch" onClick={() => rematchResponse()} >Rematch</a>)]);
```
emilia-friedberg: since `rematchResponse` doesn't take an argument here, it can just be `onClick={rematchResponse}`.

services/QuillLMS/client/app/bundles/Diagnostic/components/questions/response.tsx
```diff
@@ -0,0 +1,446 @@
...
+      buttons = [
+        (<a className="card-footer-item" key="cancel" onClick={() => cancelResponseEdit(response.key)} >Cancel</a>),
+        (<a className="card-footer-item" key="unmatch" onClick={() => unmatchResponse(response.key)} >Unmatch</a>),
+        (<a className="card-footer-item" key="update" onClick={() => updateResponse(response.key)} >Update</a>)
+      ];
+    } else {
+      buttons = [
+        (<a className="card-footer-item" key="edit" onClick={() => editResponse(response.key)} >Edit</a>),
+        (<a className="card-footer-item" key="delete" onClick={() => handleDeleteResponse(response.key)} >Delete</a>)
+      ];
```
emilia-friedberg: You can take or leave this, but I think in all these onClick functions, because the response is in the state, we could extract the `response.key` inside of the function that's getting called and then we don't have to pass an argument to the onClick handler and can rewrite them as `onClick={cancelResponseEdit}`, which obeys our linter rules and is easier to read.

services/QuillLMS/client/app/bundles/Diagnostic/components/questions/response.tsx
```diff
@@ -0,0 +1,446 @@
...
+    if (response.weak) {
+      icon = '⚠️';
+    }
+    const authorStyle = { marginLeft: '10px', };
+    const showTag = response.author && (statusCode === 2 || statusCode === 3)
+    const author = showTag ? <span className="tag is-dark" style={authorStyle}>{response.author}</span> : undefined;
+    const checked = massEditResponses.selectedResponses.includes(response.id) ? 'checked' : '';
+    return (
+      <div className={bgColor} style={{ display: 'flex', alignItems: 'center', }}>
+        <input checked={checked} onChange={() => onMassSelectCheckboxToggle(response.id)} style={{ marginLeft: '15px', }} type="checkbox" />
```
emilia-friedberg: Same comment about argument to `onClick` handler.

services/QuillLMS/client/app/bundles/Diagnostic/components/questions/response.tsx
```diff
@@ -0,0 +1,446 @@
...
+      author: null,
+      parent_id: null,
+      concept_results: { [defaultConceptUID]: false, },
+    }
+    dispatch(submitResponseEdit(rid, newResp, questionID, rerenderResponse));
+  };
+
+  function rematchResponse() {
+    rematchOne(response, mode, question, questionID, rerenderResponse);
+  };
```
emilia-friedberg: Does this replace the `rematchOne` invocation in the parent component? If so, we should delete that function there.

emilia-friedberg (commented):
This is good work overall -- I just have some comments in various places about code cleanup. As you note, there is more refactoring we could do in both the `responseComponent` file and `response.tsx`, but they're looking much better than they were previously!

services/QuillCMS/app/models/response.rb
```diff
@@ -58,6 +58,30 @@ def as_indexed_json(options={})
     }
   end
 
+  def serialized_for_admin_cms(options={})
```
anathomical, line 61: I presume that this method signature was copied from elsewhere, but I note that `options` doesn't actually seem to get used in this method.

services/QuillLMS/client/app/bundles/Diagnostic/components/questions/responseComponent.jsx
```diff
@@ -80,6 +82,12 @@ class ResponseComponent extends React.Component {
     this.clearResponses();
   }
 
+  updateResponse = (id, response) => {
+    const newResponses = this.state.responses
```
anathomical: We extract `{ responses, } = this.state` on 236, so we might as well do so here and below on line 165.

anathomical (approved):

services/QuillLMS/client/app/bundles/Diagnostic/libs/grading/rematching.ts
cissyxyu: yes, that's correct!

services/QuillLMS/client/app/bundles/Diagnostic/components/questions/response.tsx
```diff
@@ -0,0 +1,439 @@
...
+    const { concept_results, } = response
+
+    if (!concept_results) { return {} }
+
+    if (typeof concept_results === 'string') {
+      return JSON.parse(concept_results)
+    } else {
+      return concept_results
+    }
+  }
```
emilia-friedberg: It would be best to move this outside of the component itself, because otherwise it's ambiguous whether `response` refers to the argument or the state variable. You could also just rename the argument, but since the function isn't using anything inside of the component itself there's no reason to have it inside anyway.

services/QuillLMS/client/app/bundles/Diagnostic/components/questions/responseComponent.jsx
```diff
@@ -154,7 +164,8 @@ class ResponseComponent extends React.Component {
   };
 
   rematchResponse = rid => {
-    const response = this.props.filters.responses[rid];
+    const { responses } = this.state
+    const response = responses[rid];
     const callback = this.searchResponses;
     rematchOne(response, this.props.mode, this.props.question, this.props.questionID, callback);
   };
```
emilia-friedberg: Isn't this no longer getting used at all?

emilia-friedberg (approved):
Two small comments, but nothing I need to review again. Thanks for all your work on this! These files look much better.

PR #11553, opened 2024-02-29
Materialized View Query Structure

dandrabik:
## WHAT
Generalized Structure to perform MatView Queries by add
## WHY
This is a pattern we are using in a few places, so want to use the same structure. These queries have a MatView `query` and a `fallback query` in case the MatView is "broken" due to a tables being synced. The fallback queries all follow the same pattern, i.e. put the MatView SQL into `WITH` clauses before the query, and edit the query to use the local WITH clause, not the MatView.
## HOW
Create a `QuillBigQuery::MaterializedViewQuery` that requires the subclass to define a list of materialized_views, then that class handles all of the fallback logic.

This is the Base class that will have specific implementation in upcoming PRs. I've tested this code in the console, verifying the query fallback logic works for both a single mat view and multiple mat views (there were several code revisions that Thomas and I went back and forth on to make this work for both). I've also done some manual testing with the querying and fallback logic by raising errors in the code.

PR Checklist | Your Answer
------------ | -------------
Have you added and/or updated tests? | YES'


services/QuillLMS/db/big_query/README.md
```diff
@@ -0,0 +1,68 @@
...
+- Pre-filter data. Run queries that only read a particular subset of the table.
+- Pre-join data. Query joins, especially between large and small tables.
+- Recluster data. Run queries that would benefit from a clustering scheme that differs from the base tables.
+```
+
+Important to know:
+- BigQuery automatically updates materialized views as new data comes in from the underlying tables
+- There are some [restrictions](https://cloud.google.com/bigquery/docs/materialized-views-create#query_limitations) on what the materialized view can contain, e.g. you can't use `OUTER JOINS`.
+- There are [tips](https://cloud.google.com/bigquery/docs/materialized-views-create#which_materialized_views_to_create) on creating efficient materialized views, e.g. put the largest, most frequently changing table first in the query
+- To minimize devops work, views should be configured to self-refresh whenever possible.
```
anathomical: Note that this is accurate regarding incremental materialized views, but we do have the intent to use at least one non-incremental materialized view where the first point is not true (you have to explicitly set OPTIONS to configure timed refreshed), and there are different restrictions than the example pointed out in the second point (OUTER JOINS are allowed).

services/QuillLMS/db/big_query/README.md
```diff
@@ -0,0 +1,68 @@
...
+2. Add .sql file referenced in config to `db/big_query/views/`. Note, this should note include the "CREATE" command, just the inner SQL.
+2. Run a command to `create/drop/refresh` the view (you need the write permission credentials to do this locally, else do this on staging or production)
+#### Available Commands
+```
+thor quill_big_query:materialized_views:create view_key   # Creates a specified Materialized View
+thor quill_big_query:materialized_views:drop view_key     # Drops a specified Materialized View
+thor quill_big_query:materialized_views:refresh view_key  # Refreshes (drops and creates) a specified Materialized View
+```
+### Migrating to new View Version
+This is something we haven't done yet, so this policy may change, but the easiest way to change/edit a view is to create a new view with a different key, put that in place, point the code to it, and drop the old view, remove the config.
```
anathomical: No strong feelings on this one, and given that we haven't had to do this yet we don't have to have a strong opinion right now, but is there a reason not to do these migrations the same way we do them for Postgres?  That is: deploy your migration first, run migration, then deploy updated code in a secondary deploy?

anathomical (approved):
Couple of items for potential discussion in the README, but this looks good to me.

services/QuillLMS/db/big_query/README.md
dandrabik: I think we might saying the same thing (I wrote this quickly, so the wording may not be as clear). What I'm saying is that instead of editing an existing view (which would break existing code during the middle of `deploy your migration first, run migration (existing code breaks), then deploy updated code (new code works)`), you should create a new view, run the migration to put it in place, then deploy the code to point to the new view. That method has no downtime.

services/QuillLMS/db/big_query/README.md
dandrabik: I've attempted to update this Readme with a little more context. I also made the migration section a little less opinionated (just listed the ways we could migrate).

PR #11559, opened 2024-03-01
allow teachers to change the setting for allowing students to see their scores

emilia-friedberg:
## WHAT
Add frontend and backend support for allowing teachers to change the setting for allowing students to see their scores.

## WHY
This attribute defaults to `true`, but we want teachers to have the option to change it from their account settings page.

## HOW
Add validations to the model file since the last PR was just migrations, then controller updates and frontend updates to enable toggling this feature on and off.

### Screenshots

### Notion Card Links
https://www.notion.so/quill/Teachers-settings-page-6fb6c75b53d04f619ea5c4d9390070bf?pvs=4

Went through a sign-up flow as a new teacher to make sure the record was created as expected, then tried toggling it on and off from the account settings page, on its own and in combination with other `teacher_info` attributes to make sure updating something else didn't have weird repercussions for this value. I am also having Peter Sharkey and Jack review it on staging.

PR Checklist | Your Answer
------------ | -------------

### DIFF
--- a/services/QuillLMS/app/assets/stylesheets/pages/my_account.scss
+++ b/services/QuillLMS/app/assets/stylesheets/pages/my_account.scss
@@ -47,9 +47,7 @@
   .dropdown-toggle.btn.btn-default {
     width: 300px;
   }
-  .error {
 
-  }
   .gray-text {
     color: #969696;
     .row {
@@ -130,28 +128,6 @@
     color: $quillforestgreen;
     margin-left: 3px;
   }
-  // .modal-dialog {
-  //   .modal-header {
-  //     min-height: 59px;
-  //   }
-  //   .account-form {
-  //     margin-bottom: 50px;
-  //   }
-  //   .close {
-  //     background-color: black;
-  //   }
-  //   .form-label, #zip{
-  //     float: none;
-  //     height: inherit;
-  //   line-height: inherit;
-  //   }
-  //   .row {
-  //     padding: 0px;
-  //   }
-  //   #select_school {
-  //     width: 50%;
-  //   }
-  // }
 }
 
 @media all and (-ms-high-contrast: none), (-ms-high-contrast: active) {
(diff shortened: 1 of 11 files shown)


services/QuillLMS/app/controllers/teacher_infos_controller.rb
```diff
@@ -39,11 +20,16 @@ def update
       @teacher_info.update!(notification_email_frequency: notification_email_frequency)
     end
 
+    if [true, false].include?(show_students_exact_score)
```
brendanshean: You could shorten these two conditionals to one-liner:

```
@teacher_info.update!(notification_email_frequency:) if notification_email_frequency
@teacher_info.update!(show_students_exact_score:) if show_students_exact_score.in?([true, false])
```

brendanshean (approved):
Looks great!  Nice work on this.

services/QuillLMS/app/assets/stylesheets/pages/my_account.scss
```diff
@@ -130,28 +128,6 @@
     color: $quillforestgreen;
     margin-left: 3px;
   }
-  // .modal-dialog {
```
anathomical, line 133: Should we just remove this code?

anathomical (approved):

services/QuillLMS/app/controllers/teacher_infos_controller.rb
```diff
@@ -4,26 +4,7 @@ class TeacherInfosController < ApplicationController
...
-    )
-
-    subject_areas = SubjectArea.where(id: subject_area_ids)
-    @teacher_info.subject_areas.push(subject_areas)
-
-    render json: {}, status: 200
-  rescue ActiveRecord::RecordInvalid => e
-    render json: {errors: e}, status: 400
-  end
-
```
eadams17, line 24: maybe I'm missing something, but I'm not sure if the `role_selected_at_signup` portion of this is accounted for in `create_teacher_info`?

services/QuillLMS/app/assets/stylesheets/pages/my_account.scss
emilia-friedberg: @anathomical this code was removed in this PR! It had been previously commented out.

eadams17 (approved):
Excited to see this go out!

services/QuillLMS/app/controllers/teacher_infos_controller.rb
emilia-friedberg: @eadams17 you're right, but I deleted this controller action because it wasn't actually getting used anywhere. Some quick investigation suggests that `role_selected_at_signup` is never being set, but I don't know when that became the case. Maybe worth filing a bug report about?

PR #11560, opened 2024-03-01
Timeframse: loosen specs for year. Add threshold spec.

dandrabik:
## WHAT
Leap year makes the 'this year' timeframe off by one day. Loosening the spec on this to be within one day.
## WHY
Making the day counts match might be stranger for edge cases, e.g. On the last day of the year, the previous year would include the first day of this year. I think the `Start of year to this day of the year` logic is a little easier to understand, so just sticking with that for now. Will adjust if there is a product requirement to change.
## HOW
Add a new threshold spec.

No QA, since this is just a spec fix. The tests pass.

PR Checklist | Your Answer
------------ | -------------


emilia-friedberg (approved):
LGTM! Thanks for picking this up.

PR #11562, opened 2024-03-04
add "your progress" section to student dashboard

emilia-friedberg:
## WHAT
Add a key metrics section to the student dashboard.

## WHY
To give all students an easy way to see how many activities and how much time they've spent on Quill, for a given classroom, in a given duration.

## HOW
Create a new query class, sharing some logic with the teacher key metric functions, which provides the relevant data, then pass that data up as part of the `student_profile` request to be displayed on the frontend by a file with nearly the same HTML, though different logic, to the teacher key metrics component.

### Screenshots
Screenshot 2024-03-04 at 1 05 43 PM

### Notion Card Links
https://www.notion.so/quill/Product-Board-30f97e4eb01246dbb8264253fb385073?p=db7c274e00e947edb8445aab04f88ab7&pm=c

Logged in as various students on staging and confirmed that the data shows up as expected, and also that it looks good on multiple screen sizes. Jack and Peter Sharkey are also reviewing.

PR Checklist | Your Answer
------------ | -------------


services/QuillLMS/app/queries/dashboard_metrics.rb
```diff
@@ -0,0 +1,25 @@
+# frozen_string_literal: true
+
+module DashboardMetrics
+
+  def today
+    @today ||= Date.current
+  end
+
+  def days_since_last_sunday
+    @days_since_last_sunday ||= today.wday == 0 ? 0 : ((today.wday + 6) % 7) + 1
```
anathomical: Is this not the same as simply doing `today.wday % 7`?  Also, if today is Sunday (`today.wday == 0`), doesn't that mean that this will return this Sunday, not last Sunday?  Is that the intention?

services/QuillLMS/app/queries/dashboard_metrics.rb
```diff
@@ -0,0 +1,25 @@
...
+  end
+
+  def last_sunday
+    @last_sunday ||= today - days_since_last_sunday
+  end
+
+  def july_first_of_this_year
+    @july_first_of_this_year ||= Date.parse("01-07-#{today.year}")
+  end
```
anathomical: Is it worth building this off of `School.school_year_start`?

services/QuillLMS/app/queries/dashboard_metrics.rb
```diff
@@ -0,0 +1,25 @@
...
+
+  def last_july_first
+    @last_july_first ||= today.month > 7 ? july_first_of_this_year : july_first_of_this_year - 1.year
+  end
```
anathomical: Again, should we use `School.school_year_start(today.year - 1-year)`?

services/QuillLMS/app/queries/student_dashboard_metrics.rb
```diff
@@ -0,0 +1,39 @@
...
+  end
+
+  def completed_sessions
+    @completed_sessions ||= begin
+      classroom_unit_ids = ClassroomUnit.where(classroom_id: @classroom_id).pluck(:id)
+
+      ActivitySession
+        .unscoped
+        .where(classroom_unit_id: classroom_unit_ids, user_id: @user.id)
+        .where.not(completed_at: nil)
```
anathomical: I think you could do this as one query:
```ruby
ActivitySession
  .unscoped
  .joins(:classroom_units)
  .where(user: @user, classroom_units: {classroom_id: @classroom_id})
  .where.not(completed_at.nil)
```

services/QuillLMS/app/queries/dashboard_metrics.rb
```diff
@@ -0,0 +1,25 @@
+# frozen_string_literal: true
+
+module DashboardMetrics
```
anathomical: I really appreciate the consolidation into this file, and realize that you're basically just copying that code, so feel free to ignore any suggest changes here to maintain continuity with the old code.

anathomical (approved):

services/QuillLMS/app/queries/student_dashboard_metrics.rb
brendanshean: To piggyback from Thomas, if you use `attr_reader :classroom_id, :user` above def initialize, then you have:

```
ActivitySession
  .unscoped
  .joins(:classroom_units)
  .where(user:, classroom_units: {classroom_id:})
  .where.not(completed_at.nil)
```

services/QuillLMS/app/queries/student_dashboard_metrics.rb
```diff
@@ -0,0 +1,39 @@
...
+       month: metrics_from_start_date(today.beginning_of_month),
+       year: metrics_from_start_date(last_july_first)
+     }
+  end
+
+  def metrics_from_start_date(start_date)
+    completed_sessions_for_timeframe = completed_sessions.where("completed_at >= ?", start_date)
+    {
+      activities_completed: completed_sessions_for_timeframe.count,
+      timespent: completed_sessions_for_timeframe.pluck(:timespent).sum { |num| num || 0 }
```
brendanshean: I think this is more efficient since it avoids explicit nil checks:
`timespent: completed_sessions_for_timeframe.pluck(:timespent).compact.sum`

services/QuillLMS/spec/queries/student_dashboard_metrics_spec.rb
```diff
@@ -0,0 +1,72 @@
...
+  let(:today_timespent) { 30 }
+  let(:yesterday_timespent) { 45 }
+  let(:first_of_month_timespent) { 60 }
+  let(:july_second_timespent) { 90}
+
+  let!(:student) { create(:student) }
+  let!(:classroom) { create(:classroom) }
+  let!(:classroom_unit) { create(:classroom_unit, classroom: classroom) }
+  let!(:activity_sessions) do
+    [
```
brendanshean, line 28: The four `classroom_unit: classroom_unit,` clauses can be shortened to `classroom_unit:,`

services/QuillLMS/app/queries/dashboard_metrics.rb
brendanshean, line 1: This feels more like a concern.  I would probably move this to `app/queries/concerns` and add a `extend ActiveSupport::Concern` right below `module DashboardMetrics`, a pattern which is consistent with other directories where module/mixins segregated from classes.

brendanshean (approved):
Thanks for building this out.  This seems like a really nice feature.

Some more minor refactoring comments, but overall looks good.

eadams17 (approved):
LGTM!

services/QuillLMS/app/queries/dashboard_metrics.rb
emilia-friedberg: Haha, yes, I just tested and all this is the same as `today.wday % 7`. Good catch. I think that's correct -- last Sunday is inclusive of today if today is Sunday.

services/QuillLMS/spec/queries/student_dashboard_metrics_spec.rb
emilia-friedberg: Thank you! I am going to internalize this soon!

services/QuillLMS/app/queries/dashboard_metrics.rb
emilia-friedberg: Good call! I think we can actually replace the method entirely with `school_year_start`.

PR #11565, opened 2024-03-05
Keep evidence migrations automatically in sync with LMS

brendanshean:
## WHAT
Add functionality that automates the syncing of migrations between the main app and engine.
Also, change back the schema format for evidence to schema.rb since the `vector` type actually does work with this format. 

## WHY
When migrations are created in an engine, we also want them to be kept in sync with the main app.  
This can be done via the [command](https://guides.rubyonrails.org/engines.html#engine-setup)  `bin/rails evidence:install:migrations`

However, based on looking through the migration history, this command is not always run, instead it looks like manual copying of migrations is performed (there are about ten migrations that lack `.evidence` or `.comprehension` suffixes:

Instead, let's have one standard command that `rails db:migrate` that keeps these migrations in sync automatically and also gets rid of the copying over of migration files, which in some cases, the diff isn't preserved due to later modifications.

## HOW
I added a config initializer to the evidence engine based on this [post](https://medium.com/@AlexKalininHedin/migrations-inside-rails-engine-b1efbe184f10).

I then went through and copied all relevant evidence migration files from db/migrate over to engines/evidence/db/migrate and did some things.  I'll describe what I did through an example:

1. Consider two migration files M (main app) and E (evidence engine)
    M: db/migrate/20220456_create_evidence_blah.evidence.rb
    E: engines/evidence/db/migrate/20220123_create_evidence_blah.rb

2. Move file M two same directory as file E:
    M: engines/evidence/db/migrate/20220456_create_evidence_blah.evidence.rb
    E: engines/evidence/db/migrate/20220123_create_evidence_blah.rb

3. Rename with file M as same timestamp as same timestamp as file E
    M: engines/evidence/db/migrate/20220456_create_evidence_blah.evidence.rb
    E: engines/evidence/db/migrate/20220456_create_evidence_blah.rb

4.  Compare the contents of the two files.  The gold standard should be file M (since this is what is on production), so if there are any differences, copy them over to file E. 

5. Finally remove file M.  (We don't need it anymore since the initializer 
    E: engines/evidence/db/migrate/20220456_create_evidence_blah.rb

Notes
1.  There were several files in the db/migrate that did not include `{.evidence,.comprehension}` suffix so I temporarily added that in for differentiation with the evidence migration file.
2.  The file `db/migrate/20210505150457_remove_null_restriction_from_feedback_history_ratings.comprehension.rb` was listed as comprehension yet it's not actually in the engine so I renamed it.

### Screenshots

### Notion Card Links

To check the structure.sql for the main app, I ran `r db:drop; r db:create; r db:schema:load` than did the local data rake task.  There was no change in the structure.sql and all migrations were up and named except for a few '*** NO FILE ***' in some of our early migrations:
Screenshot 2024-03-05 at 10 18 49 AM

For easier comparison, I changed the evidence  schema format back to schema.rb.  I then ran a diff between this branch and an evidence schema.rb generated from the `develop` branch and it returned empty.

PR Checklist | Your Answer
------------ | -------------


brendanshean:
I'm running into some issues now with the a new migration on evidence so I'm going to temporarily close this for now as I'd like to get moving with the real issues.

PR #11567, opened 2024-03-05
fix some more incorrect sequence bugs

emilia-friedberg:
## WHAT
There is a super fun issue, which existed prior to the recent round of incorrect sequence updates but has been uncovered by them, where the way we save incorrect sequences when they are first created for new questions (as a hash) is not how they are saved when they are reordered (as an array). This is fixable on its own, but I'm afraid that changing the structure here is going to mess with the actual grading logic, which doesn't actually take an "order" attribute into account for any of the question types the way it does for focus points. At some point soon, I think we really do need to do that revisitation of how we store and update these things, but in the meanwhile I think this series of changes will get us around that problem.

## WHY
We want the curriculum team to be able to create and edit incorrect sequences.

## HOW
Make incorrect sequences retrievable regardless of whether they actually have an index as a key or a real key.

### Screenshots

### Notion Card Links
https://www.notion.so/quill/Can-t-edit-Incorrect-Sequences-in-Connect-d75bc38569c546bda091c7611b3ed265?d=a7bdb1ffb07146ec80f31c63efc6bb32

I tested creating new questions, adding incorrect sequences to them, editing those incorrect sequences from both the main page and the individual edit page, and reordering them for both Connect and Diagnostic. Grammar did not appear to have this problem for some reason. I am also asking Rachel to QA so hopefully we can stop getting related bug tickets.

PR Checklist | Your Answer
------------ | -------------

### DIFF
--- a/services/QuillLMS/client/app/bundles/Connect/components/incorrectSequence/editIncorrectSequenceContainer.jsx
+++ b/services/QuillLMS/client/app/bundles/Connect/components/incorrectSequence/editIncorrectSequenceContainer.jsx
@@ -3,6 +3,7 @@ import { connect } from 'react-redux';
 import questionActions from '../../actions/questions';
 import sentenceFragmentActions from '../../actions/sentenceFragments';
 import IncorrectSequencesInputAndConceptSelectorForm from '../shared/incorrectSequencesInputAndConceptSelectorForm.jsx';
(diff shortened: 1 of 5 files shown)


eadams17 (approved):
Thank you for figuring out this work around!

PR #11568, opened 2024-03-06
fix rails admin search

happythenewsad:
## WHAT
- rescues Json parse failures within RailsAdmin

## WHY
- so that RailsAdmin search is not broken 

## HOW
monkey patch 

### Screenshots

### Notion Card Links
https://www.notion.so/quill/500-error-when-searching-in-Master-CMS-38cdffcf5a374879b627763d141423b4

1. Browse to http://localhost:5000/staff/activity 
2. Search activities for a integer value, confirm successful result
3. Search activities for a string value, confirm successful result

PR Checklist | Your Answer
------------ | -------------
Have you added and/or updated tests? |  No; config change only

### DIFF
--- a/services/QuillLMS/config/initializers/rails_admin.rb
+++ b/services/QuillLMS/config/initializers/rails_admin.rb
@@ -146,6 +146,14 @@ end
 # Monkey patch for a known issue: RailsAdmin tries to parse search strings as JSON
 # https://github.com/railsadminteam/rails_admin/issues/2502
 class RailsAdmin::Config::Fields::Types::Json
+  register_instance_option :formatted_value do
+    if value.is_a?(Hash) || value.is_a?(Array)
+      JSON.pretty_generate(value)
+    else
+      value
+    end
+  end
+
   def parse_value(value)
     value.present? ? JSON.parse(value) : nil
   rescue JSON::ParserError
--- a/services/QuillLMS/config/initializers/rails_admin.rb
+++ b/services/QuillLMS/config/initializers/rails_admin.rb
@@ -142,3 +142,13 @@ Rails.application.config.eager_load do
     config.excluded_models.push(*MODELS_TO_EXCLUDE)
   end
 end
+
+# Monkey patch for a known issue: RailsAdmin tries to parse search strings as JSON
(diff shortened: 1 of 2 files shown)


services/QuillLMS/config/initializers/rails_admin.rb
```diff
@@ -142,3 +142,13 @@
     config.excluded_models.push(*MODELS_TO_EXCLUDE)
   end
 end
+
+# Monkey patch for a known issue: RailsAdmin tries to parse search strings as JSON
+# https://github.com/railsadminteam/rails_admin/issues/2502
```
dandrabik, line 147: This seems to use half of the solution listed [here](https://github.com/railsadminteam/rails_admin/issues/2502#issuecomment-504612818). Is there a reason we aren't only using this piece?

brendanshean (approved):
Thanks for figuring this out.

services/QuillLMS/config/initializers/rails_admin.rb
```diff
@@ -142,3 +142,13 @@
...
+class RailsAdmin::Config::Fields::Types::Json
+  def parse_value(value)
```
dandrabik, line 157: Also, looks like this maps to this [Sentry issue](https://quillorg-5s.sentry.io/issues/4475454913/?environment=production&project=11238&query=is%3Aunresolved&referrer=issue-stream&statsPeriod=24h&stream_index=20). Can we write a failing test for this and then fix it?

dandrabik (changes requested):
Let's add a test to this to confirm this is a working fix before merging in.

services/QuillLMS/config/initializers/rails_admin.rb
happythenewsad: That part was not necessary during my QA testing, but I added it in, in case there's some nuance that the author caught.

services/QuillLMS/config/initializers/rails_admin.rb
happythenewsad: added

dandrabik (approved):
:+1: Thanks for fixing this and adding a test. 🚢

PR #11570, opened 2024-03-06
Display questions and responses as HTML in admin panels

cissyxyu:
## WHAT
Display the questions (i.e. prompts) and responses as HTML instead of as a pure string on the admin CMS for connect, grammar, diagnostic.

## WHY
We save question and response strings inconsistently, sometimes as readable strings and sometimes as html (e.g. with HTML formatting tags like bold/italic tags and HTML entities for punctuation). But we only display questions and responses as regular strings, which leads to the HTML markup being displayed to admin. 

This causes frustration on the admin side because they want to see readable question and responses. Ultimately, we should solve the problem by getting rid of the inconsistencies in the data; however, for now we can just display both questions and responses to admin as HTML, because both regular non-HTML strings and HTML strings can be displayed this way in a readable manner.

## HOW
Use dangerouslySetHTML to display html directly wherever we display questions or responses in code to admin.

### Screenshots
Screenshot 2024-03-06 at 8 52 47 PM
Screenshot 2024-03-06 at 8 52 34 PM
Screenshot 2024-03-06 at 8 51 37 PM

### Notion Card Links
[(Please provide links to any relevant Notion card(s) relevant to this PR.)](https://www.notion.so/quill/Apostrophes-and-formatting-aren-t-rendering-consistently-in-Diagnostic-Connect-and-Grammar-CMSs-37f9b862406f4962812bc829bb172bd3?pvs=4)

Admin provided me a list of problematic questions and responses in the Notion spec, and I tested them before and after on Staging to make sure they are visually appearing in the desired way. Additionally, since I changed all the questions/responses display for all the apps, I manually went in and checked that all question types are displaying normally and the links are working in Connect, Grammar, and Diagnostic CMS.

PR Checklist | Your Answer
------------ | -------------
Have you added and/or updated tests? |  NO, no front-end unit tests for this part of the code base.

### DIFF
--- a/services/QuillLMS/client/app/bundles/Connect/components/concepts/concept.tsx
+++ b/services/QuillLMS/client/app/bundles/Connect/components/concepts/concept.tsx
@@ -2,6 +2,7 @@ import * as React from 'react'
 import { connect } from 'react-redux'
 import { Link } from 'react-router-dom'
 import _ from 'underscore'
+import { decode } from 'html-entities'
 
 import { hashToCollection, Spinner } from '../../../Shared/index'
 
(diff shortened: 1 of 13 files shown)


emilia-friedberg (commented):
Your general approach here looks good, but I think we should stick with the `<Link>` elements where possible here -- using `<a>` tags instead will cause full reloads, which could make loading slower because we lose information from the reducer. Also, where you're nesting a `<div>` inside an `<a>` tag, it would be best to use spans so we aren't nesting block elements inside of inline elements.

cissyxyu:
> Your general approach here looks good, but I think we should stick with the `<Link>` elements where possible here -- using `<a>` tags instead will cause full reloads, which could make loading slower because we lose information from the reducer. Also, where you're nesting a `<div>` inside an `<a>` tag, it would be best to use spans so we aren't nesting block elements inside of inline elements.

Got it. I didn't realize that about the <a></a> tag reloading. I've changed everything to use the <Link> tag instead!

emilia-friedberg (approved):
LGTM! Thank you for making that change!

PR #11573, opened 2024-03-06
Create Evidence::PromptResponse model

brendanshean:
## WHAT
Create the Evidence::PromptResponse model

## WHY
We would like a resource to store embeddings for student responses to evidence prompts.  These embeddings will be leveraged in downstream tasks involving generative AI.

## HOW
Create a resource with the vector datatype.  Use the neighbor gem `has_neighbors` helper on the model to enable useful similarity functionality.

### Screenshots

### Notion Card Links
https://www.notion.so/quill/Backend-for-Embeddings-Caching-for-Evidence-2f735c61f06848eeb8fc5c2130390c1f?pvs=4

I created a record in the console on staging and ensured that the embedding was properly assigned via the callback.

PR Checklist | Your Answer
------------ | -------------


services/QuillLMS/engines/evidence/app/models/evidence/prompt_response.rb
```diff
@@ -0,0 +1,37 @@
...
+#
+# Indexes
+#
+#  index_evidence_prompt_responses_on_text  (text) UNIQUE
+#
+
+require 'neighbor'
+
+module Evidence
+  class PromptResponse < ApplicationRecord
```
dandrabik, line 16: I have some questions about the modeling. So this class has no connection to an `Evidence::Prompt` and seems to be a generic storage of text and an embedding. So a few questions:
1. If you'd like to keep this model generic, we might want to change the model name to something generic.
2. Are you adding a connection to an `Evidence::Prompt` in a different model? Or should a `prompt_id` be added here (this may be the simpler route for now)?

dandrabik (changes requested):
Overall the code looks good and is well tested. I have some questions about the modeling and its connection to an `Evidence::Prompt` that I think we should talk through before merging (small change for either approach)

services/QuillLMS/engines/evidence/app/models/evidence/prompt_response.rb
brendanshean: After discussing with Dan and noting that the lever for adjusting dimension, I think it makes sense to include `prompt_id` with the model since that will some transparency on the use case for these records and possibly for the selection of model (and dimension).

dandrabik (approved):
LGTM, thanks for putting this together!

services/QuillLMS/engines/evidence/lib/evidence/open_ai/embedding_fetcher.rb
```diff
@@ -0,0 +1,31 @@
...
+      def initialize(dimension:, input:, model:)
+        @dimension = dimension
+        @input = input
+        @model = model
+      end
+
+      def cleaned_results
+        response
+          .parsed_response['data']
+          &.first
```
happythenewsad, line 21: Do we really want safe nav operators here? This could result in some records having `nil` or `[]` embeddings without us easily knowing. 

If you prefer not throwing an exception on unparsed input, we could add a column to the new table in order to earmark records that need to be run through OpenAI again.

brendanshean: The `Evidence::PromptResponse` table has non-null constraint as well as a validation for presence on `embedding` so creation attempts with `embedding: nil` should raise      
`ActiveRecord::RecordInvalid: Validation failed: Embedding can't be blank`

Also, the neighbor gem will raise `Neighbor::Error: Expected 1536 dimensions, not 0` when `embedding: []`.

Nonetheless, you are correct in raising this because there is a tbd question of what is be done with cases where this does fail other than raise exceptions.

PR #11581, opened 2024-03-08
Add closest feedback method

brendanshean:
## WHAT
1. Add `closest_feedback` method to `Evidence::PromptResponse`
2.  Rename `PromptResponse` text field to `response_text`

## WHY
1.  The method will be useful when populating `Evidence::PromptResponseFeedback`
2.  `text` feels kind of ambiguous since there is also text associated with the prompt.

## HOW
1.  Utilize the neighbor gem's `nearest_neighbors` method and return the top result.
2.  Add migration that renames column

### Screenshots

### Notion Card Links
https://www.notion.so/quill/Backend-for-Embeddings-Caching-for-Evidence-2f735c61f06848eeb8fc5c2130390c1f?pvs=4

I created some records in the console on staging and verified that the nearest neighbor worked.

PR Checklist | Your Answer
------------ | -------------

### DIFF
--- a/services/QuillLMS/engines/evidence/app/models/evidence/prompt_response.rb
+++ b/services/QuillLMS/engines/evidence/app/models/evidence/prompt_response.rb
@@ -18,6 +18,8 @@ module Evidence
     DIMENSION = 1536
     MODEL = 'text-embedding-3-small'
 
+    DISTANCE_METRIC = 'cosine'
+
     belongs_to :prompt
     has_one :prompt_response_feedback, dependent: :destroy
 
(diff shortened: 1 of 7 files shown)


services/QuillLMS/engines/evidence/app/models/evidence/prompt_response.rb
```diff
@@ -18,20 +18,38 @@ class PromptResponse < ApplicationRecord
...
 
-    validates :text, presence: true
+    validates :response_text, presence: true
     validates :embedding, presence: true
     validates :prompt, presence: true
 
     before_validation :set_embedding
 
+    def closest_prompt_response
+      @closest_prompt_response ||= begin
```
dandrabik: This is probably fine to memoize give our use case, but this is a method whose value isn't constant and can change over time.

dandrabik (approved):
This looks good. One small comment about memoization that is probably fine either way.

services/QuillLMS/engines/evidence/app/models/evidence/prompt_response.rb
brendanshean: That's a good point.  I was thinking of memoization to prevent duplicate calls for closest feedback, but I can just use a temp variable there.

PR #11583, opened 2024-03-08
lms/allow-multiple-view-refreshes

anathomical:
## WHAT
Update materialized view refresh controller to take multiple views
## WHY
We're only allowed to make one HTTP request when Airbyte finishes syncing a dataset, so that call needs to be able to refresh multiple materialized views.
## HOW
Parse the URL `param` and split it on commas.  NOTE: This means that we can't use commas in our view names, but I think that may be syntactically invalid already.

Deployed code to my staging environment, manually dropped the view, manually hit the URL to refresh both views on my staging environment, confirmed that the view got created in BigQuery.

PR Checklist | Your Answer
------------ | -------------


happythenewsad (approved):

dandrabik (approved):
:+1: LGTM

PR #11588, opened 2024-03-11
Lms/post dianogstic from materialized view

anathomical:
## WHAT
- Pass "growth percentage" from the back-end rather than trying to live-calculate it on the front-end
- Replace the back-end query for Post Diagnostic Completion with one based on our new materialized view
- Update the materialized view to not have a cut-off date that got left in by accident
- Add a new shared context in our specs to support testing around the new materialized view

NOTE: Spec queries in this current approach are designed to fail once (since the underlying materialized view doesn't exist in the test environment), then succeed on the retry.  This means we're running twice as many BigQuery queries as strictly necessary, and there's certainly a cleaner way of handling that, but I wanted to get this out the door quicker so I'm skipping over that issue here for now.  Open to discussion on the subject, though.
## WHY
- The growth percentage calculation is quite complicated, and at this scale can't be easily done on the front-end
- The complexity of the growth percentage is such that we significantly benefit from using a materialized view that stores that calculated value
- While that date was in the future during the development of the view, so always had all our data in it, the lingering date has fallen into the past and was thus cutting some data out of reports
- So that it's easy to pull in all the required base data when writing tests for the new approach
## HOW
- Add additional values to the API response for this call and update the front-end code to parse and display those values
- Add a new `DiagnosticAggregateViewQuery` base class that is mostly code from the `DiagnosticAggregateQuery` class, but it inherits from `QuillBigQuery::MaterializedViewQuery` instead of `QuillBigQuery::Query` so that we get the materialized view fallback and retry logic.  Then replace the existing code for this API call with new code that relies on the new inheritance chain and materialized view.
- Change some of the date filtering from `BETWEEN` to `>=`.
- Created a new shared context file that instantiates all the data models required for the fallback query for the materialized view to work

### Notion Card Links
https://www.notion.so/quill/Admin-Diagnostic-BigQuery-enhancements-d578edc7e5ec487899ff8bd633588295?pvs=4

I repeated the process we used in our initial testing of this feature:
1. Filter the admin report to a specific teacher and group by classroom
1. In another tab, log into that teacher's account in production and look at their Diagnostic Reports page
1. Compare all the values shown on the admin report to all the values shown on the teacher report to confirm that both sides are getting the same numbers.  (Specifically tested in this round, for this specific API: student assigned and completed counts for the Post diagnostic and overall skill growth percentage.)

PR Checklist | Your Answer
------------ | -------------


eadams17 (approved):
LGTM!

services/QuillLMS/app/queries/admin_diagnostic_reports/post_diagnostic_completed_view_query.rb
```diff
@@ -0,0 +1,54 @@
...
+          FROM (#{super})
+          GROUP BY diagnostic_id, diagnostic_name, aggregate_id, name, group_by, post_students_completed
+      SQL
+    end
+
+    def specific_select_clause
+      <<-SQL
+        COUNT(DISTINCT performance.post_activity_session_id) AS post_students_completed,
+        /*
+        The value below is used to duplicate growth percentage calculations from teacher reports:
```
happythenewsad: Let's yank the comments outside of the HEREDOC. They're difficult to read and to differentiate from the SQL.

happythenewsad (approved):
Looks good, except for a cosmetic change

services/QuillLMS/app/queries/admin_diagnostic_reports/diagnostic_aggregate_view_query.rb
```diff
@@ -0,0 +1,207 @@
...
+    def school_ids_where_clause = "AND schools_users.school_id IN (#{school_ids.join(',')})"
+    def teacher_ids_where_clause = ("AND schools_users.user_id IN (#{teacher_ids.join(',')})" if teacher_ids.present?)
+
+    def group_by_clause = "GROUP BY performance.activity_id, activities.name, aggregate_id, #{aggregate_sort_clause}"
+
+    def active_classroom_stubs_view = materialized_view('active_classroom_stubs_view')
+    def active_user_names_view = materialized_view('active_user_names_view')
+    def performance_view = materialized_view('pre_post_diagnostic_skill_group_performance_view')
+
+    def materialized_views = [active_classroom_stubs_view, active_user_names_view, performance_view]
```
dandrabik: Small request: Can you move this method to the top of the file below initialize (or below run). Personal preference, but it's a method the subclass is required to implement that powers a lot of the logic. I think putting it up top makes it a little easier to parse.

dandrabik (approved):
Overall looks good. I pushed some changes to queries.thor and left a small comment about file organization. Looks to go and ready for the performance updates.

PR #11592, opened 2024-03-12
Call search API only once when admin update search criteria

cissyxyu:
## WHAT
On Diagnostic, Grammar, and Connect admin grading panels, the responses/search API is being called twice, sometimes three times at once when admin do something like toggle a filter, change the page they're viewing, or enter search text. This fixes assorted bugs so that the search function is called ONLY once each time admin performs a sorting/search action.

## WHY
The search function is a costly and time-consuming API call that takes a long time to return. Calling it three times is a waste of resources and confusing for admin, because sometimes a call will fail and the front end will not update with expected data, will update with no data.

## HOW
There are a few things going on here that cause the API to get double- and triple-called:

1. When a Status is toggled (e.g. hints selected or focus points/incorrected sequence toggled), it's registered as two updates, one is the Status itself and the other is resetting the page number to be 1. This triggers two search calls. Consolidate these two calls into one action so that the search is only triggered once.

2. When admin type into the search bar, it triggers an extra API call if they type a non-empty string. Remove this extra API call because we don't want to search unless they specifically click the Search button.

3. When admin click Search, it both triggers a searchResponses() call and then, once the page is refreshed, the search text in the props has changed which triggers another search API call in the response component. Remove the text key from the props component when detecting a props change, because we don't care about the text changing -- we only care if the admin has expressly clicked on the search bar.

4. When admin change the page they're viewing, two function are called -- updatePageNumber and search responses. Remove the extra searchResponses call here, and modify the filterReducers to register a page number change as a change in formattedFilterData which will trigger a searchResponses() call when the response component is updated. Essentially, this treats a page number change the same way as a Sort order change or a toggle Status change. None of our handleChange functions should ever call searchResponses() on their own, since this is taken care of in the componentDidUpdate() check.

5. In the filterReducer, two keys were not being added to the formattedFilterData -- text and pageNumber. However these keys are technically part of the filters payload and sometimes appear in the props passed into responseComponent.filters.formattedFilterData. It's important to have these props in the filterReducer because otherwise, the existence or non-existence of these two keys might be registered as a change in filters, even if it's not intended as such.

### Screenshots

### Notion Card Links
[(Please provide links to any relevant Notion card(s) relevant to this PR.)](https://www.notion.so/quill/The-filters-in-the-Connect-Diagnostic-CMS-are-not-working-as-expected-10466ac87020487e87eed7f72eff4681?pvs=4)

We don't have unit tests in this part of the code base, but I've deployed everything to staging and have manually tested clicking each of the sorting and filtering UI options:
-Order by ascending/descending
-Toggling various filter buttons 
-Changing page numbers 
-Searching for text

I've monitored the Network tab to confirm that for each of these actions, one and only one API call is made to CMS/responses/search.

PR Checklist | Your Answer
------------ | -------------
Have you added and/or updated tests? |  NO, no unit or snapshot tests here. I've manually tested (See above).


services/QuillLMS/client/app/bundles/Connect/components/questions/responseComponent.jsx
```diff
@@ -63,7 +63,12 @@ class ResponseComponent extends React.Component {
 
   componentDidUpdate(prevProps) {
     const { filters, states, questionID, dispatch, } = this.props
-    if (!_.isEqual(filters.formattedFilterData, prevProps.filters.formattedFilterData)) {
+    // remove text field when comparing, since sometimes the text search can change without necessarilly requiring
+    // a new search (e.g when admin is typing)
+    const { ["text"]: unused, ...prevFiltersWithoutText } = prevProps.filters.formattedFilterData;
+    const { ["text"]: unused2, ...currFiltersWithoutText } = filters.formattedFilterData;
```
anathomical, line 69: Hadn't seen this pattern before.  It's cool.

anathomical (approved):

brendanshean (approved):
Great on work figuring all of these issues out.

PR #11596, opened 2024-03-13
update how we save grammar concept results to cms

emilia-friedberg:
## WHAT
Fix bug where Grammar concept results were not saving on first submission of new responses.

## WHY
We want these to be stored correctly.

## HOW
Rachel mentioned, and I confirmed, that this was not working on first submission, but did work when rematched. I looked to see what difference there was between the rematching code and the regular submission, and we were transforming the concept results for rematching but not regular submission, so I pulled out and renamed the transform function and applied it in both instances.

Note: in both Connect and Diagnostic, we are similarly doing this transform when we rematch but not regularly submit, but I tested and we don't seem to be having the same issue there. I don't want to make a wide-reaching change unnecessarily, so I've left the code there alone for now.

### Screenshots

### Notion Card Links

I made a test question and test activity, and played through that to make sure that new questions and responses are getting concept results saved correctly. Then I played through an existing activity and confirmed that their concept results were also still getting saved as we'd expect.

PR Checklist | Your Answer
------------ | -------------

### DIFF
--- a/services/QuillLMS/client/app/bundles/Grammar/actions/responses.ts
+++ b/services/QuillLMS/client/app/bundles/Grammar/actions/responses.ts
@@ -6,6 +6,8 @@ import { ActionTypes } from './actionTypes';
 
 import { requestDelete, requestGet, requestPost, requestPut, } from '../../../modules/request/index';
 import objectWithSnakeKeysFromCamel from '../libs/objectWithSnakeKeysFromCamel';
+import convertConceptResultsArrayToHash from '../libs/convertConceptResultsArrayToHash';
+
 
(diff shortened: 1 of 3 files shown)


anathomical (approved):
Breaking this into its own lib is nice.  It's another place that makes me think about how great it would be if we could combine all the different apps into a single app.  One day!

emilia-friedberg:
> Breaking this into its own lib is nice. It's another place that makes me think about how great it would be if we could combine all the different apps into a single app. One day!

I'm thinking about proposing this as a Q2 OKR because I feel like such a high volume of our support tickets recently has had to do with issues related to this.

PR #11600, opened 2024-03-14
update student results page

emilia-friedberg:
## WHAT
Update the student results page to show score-specific data if the teacher has that enabled for their students. Otherwise, shift to showing key target concept skills rather than all concept results.

## WHY
This is part of the project to show students their scores.

## HOW
Update the backend to format the data the way we need it, then update the frontend to display in accordance with the new designs.

### Screenshots
Screenshot 2024-03-14 at 10 23 06 AM
Screenshot 2024-03-14 at 10 21 21 AM
Screenshot 2024-03-14 at 10 17 07 AM
Screenshot 2024-03-14 at 10 10 47 AM

### Notion Card Links
https://www.notion.so/quill/Product-Board-30f97e4eb01246dbb8264253fb385073?p=e9d92e7e51f2457fa456193fa624b14e&pm=c

I played through multiple activities as a student, both with the "show exact scores" setting for their teacher toggled on and off. Jamie, Peter Sharkey, and Jack also reviewed. We found an unrelated bug that we nonetheless want to fix before this goes out because the question-based report makes it more obvious, so forthcoming PR for that.

PR Checklist | Your Answer
------------ | -------------


services/QuillLMS/app/models/activity_session.rb
```diff
@@ -324,38 +324,6 @@ def grade
     percentage
   end
 
-  # rubocop:disable Metrics/CyclomaticComplexity
```
brendanshean, line 327: 😎

services/QuillLMS/app/controllers/activity_sessions_controller.rb
```diff
@@ -56,6 +54,17 @@ def activity_session_from_classroom_unit_and_activity
     redirect_to "/activity_sessions/#{started_activity_session_id}/play"
   end
 
+  private def set_variables_and_kick_off_worker_for_partner_session
```
brendanshean: I like this extraction to cut down on complexity.  You could refactor further and pull out some assignments (which I'm noticing I missed when I did the determine_layout method below).
```
def result
  ...
  run_partner_session_worker
  ...
end

private def partner_session? = session[:partner_session].present? && partner_name && partner_session_id
private def partner_name = session[:partner_session]["partner_name"]
private def partner_session_id = session[:partner_session]["session_id"]

private def run_partner_session_worker
  return unless partner_session?

  activity_url = @activity_session.activity.anonymous_module_url.to_s
  results_url = url_for(action: 'result', uid: @activity_session.uid)
  
  AmplifyReportActivityWorker.perform_async(
    partner_session_id, 
    @activity_session.activity.name, 
    @activity_session.activity.description, 
    @activity_session.percentage, 
    activity_url, 
    results_url
  )
end
```

Then at the bottom refactor determine layout:
```
  private def determine_layout = "integrations" if partner_session?
```

services/QuillLMS/spec/models/activity_session_spec.rb
```diff
@@ -501,51 +501,6 @@
     end
   end
 
-  describe '#parse_for_results' do
```
brendanshean, line 504: 😎

brendanshean (approved):
I'm excited to see this feature go out.  Nice work!

services/QuillLMS/client/app/bundles/Teacher/containers/ResultsPage.jsx
```diff
@@ -1,95 +1,109 @@
...
+      <div className="activity-results">
+        <div className="activity-name-and-overall-score">
+          <div>
+            <h2>{activityName}</h2>
+            {showExactScore && <p>Total Score: {numberOfCorrectQuestions} of {numberOfQuestions} target skills correct ({Math.round(percentage * 100)}%)</p>}
+          </div>
+          <ResultsIcon activityType={activityType} percentage={percentage} />
+        </div>
+        {this.renderResultsSection('Proficient', proficientSkills)}
+        {this.renderResultsSection('Keep Practicing', keepPracticingSkills)}
```
eadams17, line 71: should we make these first arguments consts?

eadams17 (approved):

services/QuillLMS/client/app/bundles/Teacher/containers/ResultsPage.jsx
emilia-friedberg: I don't think it's necessary here because they're just what gets displayed, not being used for business logic or referenced in more than one place.

PR #11599, opened 2024-03-13
lms/fix-admin-diagnostic-recommendation-count

anathomical:
## WHAT
Rewrite the Diagnostic Overview Query to use a materialized view that rolls up all the data
## WHY
This is way faster and cheaper to run that the old complicated query
## HOW
Create a rollup materialized view in BigQuery, then build a Query object that pulls data from that view instead of from raw tables

### Notion Card Links
https://www.notion.so/quill/ADGR-QA-February-22-bb480e0255e542909ecee020b9a97742?pvs=4#f16d8adae0a0480fb683ddedebd942a6

### What have you done to QA this feature?
- First pass look: confirm that "Completed Activities" column never has a number that's higher than the "Pre Diagnostics Completed" column across a number of users and filter options
- Deep dive on a couple of specific classrooms:
- - Aggregate on classrooms
- - Log into a the teacher account for that classroom and look at the recommendations page for the diagnostic
- - Count-by-eye the number of students on that page who have completed at least one recommendation, and confirm that that number matches the number shown in the admin report

PR Checklist | Your Answer
------------ | -------------
Have you added and/or updated tests? |  Yes
Have you deployed to Staging? | Yes
Self-Review: Have you done an initial self-review of the code below on Github? | Yes

PR #11602, opened 2024-03-15
Lms/admin diagnostic overview performance

anathomical:
## WHAT
- Modify the existing pre-post comparison rollup view to include a couple of extra values (specifically around post diagnostic assignment date) and to use a `FULL JOIN` so that it includes data for edge cases where a student is assigned the post diagnostic, but not the pre diagnostic, in a single classroom.  (I'm not sure how that happens, I only know that it does)
- Add a new materialized view that puts all the filter values in one table so that we can eliminate a bunch of JOINs when filtering admin queries
- Create new modified queries that rely on the rollup view for `PostDiagnosticAssigned`, `PreDiagnosticAssigned`, and `PreDiagnosticCompleted`
- Use the new view-based queries in the worker so that they get used to populate cache
- Run benchmarking on the non-view and the view-based version of the queries for performance comparison

NOTE: This branch also contains #11599 in order to consolidate all performance improvements on this page into one place
## WHY
We want to speed up the performance of the Admin Diagnostic Overview Report, so we want the component queries to be fast.  Secondarily, we'd like them to run as much in BI Engine as possible to keep costs down.
## HOW
- Use a similar structure as we use for Pre Diagnostic calculation to include cases where a student has been assigned a Post, but hasn't completed it yet
- Join all of the various filter values into a single materialized view that can be joined onto things we want to filter
- Start with the base queries, then modify them to work with the new views rather than the base tables
- Modify the `QUERIES` hash in the worker
- Update the `queries` Thor task and run it for the new and old versions of the queries.

### Notion Card Links
https://www.notion.so/quill/BigQuery-Performance-Investigation-and-Improvements-adfb12edb23d4f3f9a9724bc74df685f?pvs=4

Given that we've extensively tested this page in its old state, and this update should not change behavior, I cycled through a number of admin users in both a staging environment using the new branch, and in production (using the non-view version of the code).  I then compared the two views ensuring that each one had the same values for all records.  I tested this at all three levels of aggregation (grade, teacher, and classroom) to make sure that none of the aggregation was different either.

PR Checklist | Your Answer
------------ | -------------


services/QuillLMS/app/workers/admin_diagnostic_reports/diagnostic_overview_worker.rb
```diff
@@ -14,10 +14,10 @@ def message
     end
 
     QUERIES = {
-      'pre-diagnostic-assigned' => PreDiagnosticAssignedQuery,
-      'pre-diagnostic-completed' => PreDiagnosticCompletedQuery,
+      'pre-diagnostic-assigned' => PreDiagnosticAssignedViewQuery,
+      'pre-diagnostic-completed' => PreDiagnosticCompletedViewQuery,
       'recommendations' => DiagnosticRecommendationsQuery,
```
dandrabik, line 19: Wasn't there a rewrite of the `DiagnosticRecommendationsQuery`? Is that in a different branch or does the base class refactor speed that up?

dandrabik (approved):
Those times look much better than before. I had one question about whether there should be updates to the `DiagnosticRecommendationsQuery` (I thought you had refactored that last week). Otherwise, looks good to me.

services/QuillLMS/app/workers/admin_diagnostic_reports/diagnostic_overview_worker.rb
anathomical: @dandrabik That's in #11599 , but I've just pulled that branch into this one (with attendant benchmarking) to keep things consolidated.  It may make sense for you to review it on that PR since it has only the changes relevant to that query.

dandrabik: @anathomical I approved it there, so consider this PR with the changes approved as well.

happythenewsad (approved):

PR #11614, opened 2024-03-18
batch concept results save workers and use pusher to notify frontend

emilia-friedberg:
## WHAT
Batch our Concept Results save workers so that we can send a Pusher notification when they're all saved to the frontend.

## WHY
We were seeing a bug where students would sometimes wind up at the results page before all the concept results for the session have actually been saved. This prevents that by waiting until the concept results are saved before sending them there.

## HOW
Just batch the workers following the structure we use for `BatchAssignRecommendationsWorker`, and update the frontends to listen with Pusher for these jobs to complete. Note: neither Lessons, Diagnostics, or Evidence activities send students to the Results page, so it wasn't necessary to update those apps here.

### Screenshots

### Notion Card Links
https://www.notion.so/quill/RFC-for-saving-concept-results-037c580b64ce4ca5a8ab5d47caf97169

I played through a Grammar, Connect, and Proofreader activity as a logged-in student and confirmed that I was successfully sent to the activity session results page with saved concept results; then I did the same thing as a logged-out user (testing the anonymous path).

PR Checklist | Your Answer
------------ | -------------

### DIFF
--- a/services/QuillLMS/app/workers/batch_save_activity_session_concept_results_worker.rb
+++ b/services/QuillLMS/app/workers/batch_save_activity_session_concept_results_worker.rb
@@ -1,0 +1,35 @@
+# frozen_string_literal: true
+
+class BatchSaveActivitySessionConceptResultsWorker
+  include Sidekiq::Worker
+
+  sidekiq_options queue: SidekiqQueue::CRITICAL
+
+  def perform(concept_results, activity_session_id, activity_session_uid)
+    batch_runner(activity_session_uid) do
+      concept_results
+        .map { |cr| concept_results_hash(cr, activity_session_id) }
+        .reject(&:empty?)
+        .each { |crh| SaveActivitySessionConceptResultsWorker.perform_async(crh) }
+    end
+  end
+
+  def on_success(_status, options)
+    PusherTrigger.run(options['activity_session_uid'], 'concept-results-saved', "Concept results saved for activity session with uid: #{options['activity_session_uid']}")
+  end
+
+  private def batch_runner(activity_session_uid, &save_concept_results)
+    batch = Sidekiq::Batch.new
+    batch.description = 'Saving Concept Results for Activity Session'
+    batch.callback_queue = SidekiqQueue::CRITICAL
+    batch.on(:success, self.class, activity_session_uid: activity_session_uid)
+    batch.jobs { save_concept_results.call }
+  end
+
+  private def concept_results_hash(concept_result, activity_session_id)
+    concept = Concept.find_by(uid: concept_result["concept_uid"])
+    return {} if concept.blank?
+
+    concept_result.merge(concept_id: concept.id, activity_session_id: activity_session_id)
+  end
+end
--- a/services/QuillLMS/app/workers/batch_save_activity_session_concept_results_worker.rb
+++ b/services/QuillLMS/app/workers/batch_save_activity_session_concept_results_worker.rb
@@ -14,15 +14,19 @@ class BatchSaveActivitySessionConceptResultsWorker
     end
   end
 
-  def on_success(_status, options)
-    PusherTrigger.run(options['activity_session_uid'], 'concept-results-saved', "Concept results saved for activity session with uid: #{options['activity_session_uid']}")
+  def on_complete(status, options)
+    if status.failures == 0
+      PusherTrigger.run(options['activity_session_uid'], 'concept-results-saved', "Concept results saved for activity session with uid: #{options['activity_session_uid']}")
+    else
+      PusherTrigger.run(options['activity_session_uid'], 'concept-results-partially-saved', "Concept results partially saved for activity session with uid: #{options['activity_session_uid']}, #{status.total - status.failures} succeeded, #{status.failures} failed")
+    end
   end
 
   private def batch_runner(activity_session_uid, &save_concept_results)
     batch = Sidekiq::Batch.new
     batch.description = 'Saving Concept Results for Activity Session'
     batch.callback_queue = SidekiqQueue::CRITICAL
-    batch.on(:success, self.class, activity_session_uid: activity_session_uid)
+    batch.on(:complete, self.class, activity_session_uid: activity_session_uid)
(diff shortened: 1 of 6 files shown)


services/QuillLMS/client/app/bundles/Proofreader/components/proofreaderActivities/container.tsx
```diff
@@ -300,7 +301,21 @@ export class PlayProofreaderContainer extends React.Component<PlayProofreaderCon
...
+    initializeSubscription(activitySessionUid) {
+      if (process.env.NODE_ENV === 'development') {
+        Pusher.logToConsole = true;
+      }
+      if (!window.pusher) {
+        window.pusher = new Pusher(process.env.PUSHER_KEY, { cluster: process.env.PUSHER_CLUSTER });
+      }
+      const channel = window.pusher.subscribe(activitySessionUid);
+      channel.bind('concept-results-saved', () => {
+        document.location.href = `${process.env.DEFAULT_URL}/activity_sessions/${activitySessionUid}`;
```
brendanshean, line 313: I realize this is refactoring, but do you know why `this.setState({ saved: true, });`
isn't used here but is with Connect and Grammar?

emilia-friedberg: @brendanshean this is a little bit circular, but that line exists in Connect and Grammar but not in Proofreader because there is no reference to a `saved` state value in Proofreader, whereas in Grammar we use it for some business logic in `componentDidUpdate` and in Connect it looks like there's a display state dependent on it.

brendanshean (approved):
Nice job pulling this together so quickly.

dandrabik (approved):
Nice work putting this together quickly! Well done! A small question about the callback, but that's about it.

One thing I'll note is that you should pay special attention to monitoring the sidekiq queue when deploying this. I don't see anything here that looks like it will be an issue, but when we edit a job that is ~30% of our background processing time, it's especially important to monitor performance, e.g. if the batch process isn't performant for some reason, it could cause a huge backlog. I'd monitor the sidekiq queues for at least 30 minutes after the deploy.

services/QuillLMS/app/workers/batch_save_activity_session_concept_results_worker.rb
```diff
@@ -0,0 +1,35 @@
...
+  def perform(concept_results, activity_session_id, activity_session_uid)
+    batch_runner(activity_session_uid) do
+      concept_results
+        .map { |cr| concept_results_hash(cr, activity_session_id) }
+        .reject(&:empty?)
+        .each { |crh| SaveActivitySessionConceptResultsWorker.perform_async(crh) }
+    end
+  end
+
+  def on_success(_status, options)
```
dandrabik: This is more of a product question, but the [batch](https://github.com/sidekiq/sidekiq/wiki/Batches#callbacks) has a `complete` and a `success` callback. If one of the 15 jobs fails, do you want to show nothing (`success`) or show the results of the 14 (`complete`). We might want `complete` here, but it's more of a product question than an engineering one.

emilia-friedberg: @dandrabik that's a good callout. I suspect we want complete rather than success (which would be closer to the existing behavior, anyway, since currently we don't rely at all on the concept results actually *being* saved). I'll update this.

emilia-friedberg:
@dandrabik and @brendanshean I'm re-requesting a review here because I've changed the logic slightly following @dandrabik's question and discussion with Peter Sharkey. If any of the concept result jobs fail now, the student will still get redirected (so they're not hanging in a waiting state forever), but they'll skip the results page and just go to their dashboard, per Peter Sharkey's request.

services/QuillLMS/app/workers/batch_save_activity_session_concept_results_worker.rb
```diff
@@ -0,0 +1,39 @@
...
+        .reject(&:empty?)
+        .each { |crh| SaveActivitySessionConceptResultsWorker.perform_async(crh) }
+    end
+  end
+
+  def on_complete(status, options)
+    if status.failures == 0
+      PusherTrigger.run(options['activity_session_uid'], 'concept-results-saved', "Concept results saved for activity session with uid: #{options['activity_session_uid']}")
+    else
+      PusherTrigger.run(options['activity_session_uid'], 'concept-results-partially-saved', "Concept results partially saved for activity session with uid: #{options['activity_session_uid']}, #{status.total - status.failures} succeeded, #{status.failures} failed")
```
brendanshean, line 21: We might want to add some logging here for failures, such as a ChangeLog (e.g. see [recalculate_locks](https://github.com/empirical-org/Empirical-Core/blob/develop/services/QuillLMS/app/models/change_log.rb#L104)).  Otherwise the student gets redirected but we have no idea how much this might be happening.

brendanshean (approved):
I think the switch from on_success to on_complete is a good one for this use case.

services/QuillLMS/app/workers/batch_save_activity_session_concept_results_worker.rb
emilia-friedberg: We'll still see it [here](https://quillorg-5s.sentry.io/issues/?groupStatsPeriod=auto&page=0&project=11238&query=is%3Aunresolved+saveactivitysessionconceptresults&referrer=issue-list&statsPeriod=14d), though, right? I am personally inclined to feel like our existing error tracking is enough if we want to actually deal with the root cause here, but that seems like a separate concern to me.

dandrabik (approved):
:+1: Thanks for the quick iteration on this. Looks good to me.

Carrying over my previous review note: Let's monitor sidekiq for 30 minutes post-deploy, just in case.

services/QuillLMS/app/workers/batch_save_activity_session_concept_results_worker.rb
brendanshean: Good point.   Carry on!

PR #11619, opened 2024-03-18
Lms/admin diagnostic skill api

anathomical:
## WHAT
Add a view version of the Admin Diagnostic Skills query
## WHY
We want this to be as performant as possible, and this version runs under 4 seconds in benchmarking.
## HOW
Rewrite the query around the new aggregate views.  Then stub out a spec file to confirm that the new code works the same as the old code.

### Notion Card Links
https://www.notion.so/quill/BigQuery-Performance-Investigation-and-Improvements-adfb12edb23d4f3f9a9724bc74df685f?pvs=4

Repeat the process we used to QA this page originally: use filters to get data aggregated to a specific set of classrooms.  Then log in as the teachers of those classrooms and compare the values in the teacher reports with the values in the admin report to ensure that they match.

PR Checklist | Your Answer
------------ | -------------


dandrabik (approved):
LGTM, thanks for making these changes.

PR #11621, opened 2024-03-19
Update Premium banners and footers to use new PostNavigationBanner

cissyxyu:
## WHAT
Update all the Premium banners, and the Premium footer, to use the new PostNavigationBanner component instead of custom coded HTML elements.

## WHY
Now that Eric has finished this banner component, we can start showing teachers the new updated banner.

## HOW
For the three premium banners, just update the existing React components to use the new PostNavigationBanner instead of custom coded HTML. I had to refactor some button component code into a function instead because the new PostNavigationBanner doesn't accept Button components as props, but we can just pass it a click handler function instead of a button.

For the footer, I had to make a new PremiumFooterAppClient so the premium_footer .erb file could pull this React component in directly. It has to be its own new AppClient because the footer is displayed separately from all other page content, so it can't be attached to any of the other AppClients because it needs to be shown on virtually all the pages. It also needs its own bundle and entrypoint because many separate bundles need to access the footer (Login, Home page, Public pages), and some pages that don't otherwise load any React, such as the "About Us" and "Team" pages, still need to load the footer. So I made a Footer bundle with the React Component that all other bundles can import, and a clientRegistration file that the shared/ entrypoint can import to use only the Footer and no other React.

### Screenshots
Screenshot 2024-03-19 at 12 38 23 PM
Screenshot 2024-03-18 at 4 39 21 PM
Screenshot 2024-03-18 at 4 39 12 PM
Screenshot 2024-03-18 at 4 39 00 PM

### Notion Card Links
https://www.notion.so/quill/Product-Board-30f97e4eb01246dbb8264253fb385073?p=44c320acad824af6a728d533f58771cf&pm=c

In the staging site, created a new teacher and went to /premium to see the first NoPremium banner. Then, clicked "Try Premium" to see the Premium Trying banner. Finally, clicked another page and went back to the /premium page to see the Ongoing Premium banner. On that banner, clicked "Upgrade to Premium Now" to ensure that the Stripe payment is triggered. On all the banners, I clicked all the buttons to verify the correct links. 

Then, log out of the teacher account and create a new teacher account to see the new footer. Click buttons and verify all links working.

PR Checklist | Your Answer
------------ | -------------


services/QuillLMS/app/views/application/_premium_footer.html.erb
```diff
@@ -1,12 +1,7 @@
-<% if !current_user || (current_user.teacher? && !current_user.is_premium?) %>
-  <div class="premium-footer">
-    <div class="content-container">
-      <div class="text-section">
-        <h2>Learn More About Quill Premium</h2>
-        <p>Premium subscriptions for schools and districts interested in priority technical support, additional reporting, and support from Quill's professional learning team--plus an option for individual teachers</p>
-      </div>
+<script> const global = globalThis; </script>
```
brendanshean: What is the purpose of this assignment?

cissyxyu: On the home page, public page, and session login pages, I was unable to load the React component and getting a console error "global is not defined". After digging around it seemed like this solution worked:

https://github.com/jpuri/react-draft-wysiwyg/issues/1317

happythenewsad: @cissyxyu It's unlikely that this directive is necessary, since no other pages require it (although see _head for the vite shim). Would you mind poking at it a bit more? If you don't have a solution by Wednesday EST, ping me, and we can look into it together.

services/QuillLMS/app/assets/stylesheets/shared/banner-with-button.scss
```diff
@@ -74,8 +74,8 @@
     }
   }
   .explore-premium-button {
-    color: $quill-gold;
-    background-color: white;
+    color: white;
```
emilia-friedberg: We should really use $quill-white here (even though they're the same).

services/QuillLMS/app/views/application/_premium_footer.html.erb
emilia-friedberg: Agreed that this feels like something we should be able to get rid of.

emilia-friedberg (approved):
Other than @happythenewsad's point about the global shim, which feels weird, this looks good to me. I like the approach you went with here -- at first I was thinking it should just be in the Shared bundle but I think ultimately because we do need it to stand on its own as an app putting it in its own unencumbered bundle makes sense. Thank you for your work on this!

services/QuillLMS/app/views/application/_premium_footer.html.erb
cissyxyu: @happythenewsad I dug around a bit and it looks like I just need to include the same vite shim that _head already has. This is because _head and _twenty_seventeen_home (which loads the current home page plus connect, grammar, etc. overview pages) and _activity (which loads teacher pages) are currently three different loading paths. So I copied the vite shim and added it to the two other root html files. Do you think this would be an appropriate solution?

eadams17 (approved):
This looks good to me. The only thing I want to flag is the possible performance implications. We removed React from all of the public pages for speed reasons, so it's probably worth checking that we won't fall below a failing score again with these changes. [This is the card](https://www.notion.so/quill/Speed-up-Minimize-js-on-public-pages-v1-6f12681cfdc24d959a96696055f33b47?pvs=4) for that project for reference (the link for the speed test tool is in there as well)

happythenewsad:
> This looks good to me. The only thing I want to flag is the possible performance implications. We removed React from all of the public pages for speed reasons, so it's probably worth checking that we won't fall below a failing score again with these changes. [This is the card](https://www.notion.so/quill/Speed-up-Minimize-js-on-public-pages-v1-6f12681cfdc24d959a96696055f33b47?pvs=4) for that project for reference (the link for the speed test tool is in there as well)

Good call out, @eadams17

services/QuillLMS/app/views/application/_premium_footer.html.erb
happythenewsad: Yeah, I think it's fine and necessary to put the shim into `_twenty_seventeen_home` home (although we should collapse this layout into `application.html.erb` in a different PR). 

However, note that the `activity.html.erb` layout includes the `_head` partial. `_head` has the shim, so `activity.html.erb` shouldn't need to add it a second time. 

I checked your branch, removed the shim from `activity.html.erb`, and loaded an Evidence activity. Looks fine:
Screenshot 2024-03-25 at 10 28 15 AM

happythenewsad (approved):
Approving, providing that you remove the shim from `activity.html.erb`

PR #11623, opened 2024-03-19
Rules analysis speedup

happythenewsad:
## WHAT
- ports the core Rules Analysis query to BigQuery
- parameterize the `activity_versions` API, call, to offer a fast version that doesn't include feedback history counts

## WHY
- Rules Analysis queries time out regularly, due to operations on large quantities of data. 
- the `activity_versions` API is called on 6 different pages, including the Rules Analysis page. Some of these pages require counts, while other pages ignore the counts completely. Making the counts generation opt-in via a query param allows us to quickly and independently speed up all of these pages, if and when we want to. Jamie currently is specing a comprehensive solution for this slow API issue.

## PERFORMANCE:
Note: This table also includes `Sessions View` improvements, which are coming in a subsequent PR 

Screenshot 2024-03-21 at 9 52 25 AM

### Notion Card Links
https://www.notion.so/quill/Product-Board-30f97e4eb01246dbb8264253fb385073?p=2a3bf6bd2c8f498fb7c4335635e5d92f&pm=s

- generally clicked around several Rules Analysis pages, using the various input fields.}
- Sent to Jamie, who has also QA'd
- Compared API results side by side, to check for accuracy. Example:

side by side rules analysis comparison

PR Checklist | Your Answer
------------ | -------------


services/QuillLMS/app/controllers/api/v1/rule_feedback_histories_controller.rb
```diff
@@ -5,7 +5,7 @@ def by_conjunction
     raise ArgumentError unless params.include?('activity_id') && params.include?('conjunction')
 
     options = params.permit(:conjunction, :activity_id, :start_date, :end_date).to_h.symbolize_keys
-    report = RuleFeedbackHistory.generate_report(**options)
```
happythenewsad, line 8: Note: I will remove the old `RuleFeedbackHistory.generate_report` method (which used Postgres) after we're happy with the performance of this PR in production.

services/QuillLMS/app/queries/staff/rules_analysis_query.rb
```diff
@@ -0,0 +1,104 @@
...
+
+    def end_date_where_clause
+      end_date ? "AND feedback_histories.time <= '#{DateTime.parse(end_date).to_fs(:db)}'" : ""
+    end
+
+    def activity_version_where_clause
+      activity_version ? "AND feedback_histories.activity_version = #{activity_version}" : ""
+    end
+
+    # rubocop:disable Metrics/CyclomaticComplexity
```
happythenewsad: This method was ported directly from `RuleFeedbackHistory`. The old method used AREL `includes` to access feedbacks. Since BigQuery does not have the `includes` feature, we must write our own. 

Note that calling from Postgres and BigQuery for a single data assemblage could be problematic due to sync lag. However, in practice, rules and feedbacks are not updated often, so this should be fine.

services/QuillLMS/spec/queries/staff/rules_analysis_query_spec.rb
```diff
@@ -0,0 +1,175 @@
...
+        expect(
+          described_class.new(activity_id: 1, conjunction: 'so').post_query_transform(example_bigquery_result)
+        ).to eq expected
+      end
+    end
+
+    context 'big_query_snapshot', :big_query_snapshot do
+      context 'without post query transforms' do
+        before do
+          described_class.class_eval { def post_query_transform(data) = data }
```
happythenewsad, line 36: To test BigQuery before running post query transforms, we replace `post_query_transform` with the identity function, for this rspec block.

services/QuillLMS/engines/evidence/lib/evidence/change_log.rb
```diff
@@ -59,17 +59,18 @@ def log_change(user_id, action, changed_record, changed_attribute = nil, previou
       Evidence.change_log_class.create(change_log)
     end
 
-    def activity_versions
+    def activity_versions(ignore_count: false)
```
anathomical: The param being `ignore_count` instead of `include_count` feels a little awkward, but in the context of "not passing the new param should result in the same behavior as the old code" makes it a reasonable approach here.

services/QuillLMS/app/queries/staff/rules_analysis_query.rb
```diff
@@ -0,0 +1,104 @@
...
+      start_date ? "AND feedback_histories.time >= '#{DateTime.parse(start_date).to_fs(:db)}'" : ""
+    end
+
+    def end_date_where_clause
+      end_date ? "AND feedback_histories.time <= '#{DateTime.parse(end_date).to_fs(:db)}'" : ""
+    end
+
+    def activity_version_where_clause
+      activity_version ? "AND feedback_histories.activity_version = #{activity_version}" : ""
+    end
```
anathomical, line 59: This is purely stylistic, and given the ternaries may not actually be visually better, but I've really been liking the one-liner format for one-line methods: `def start_date_where_clause = start_date ? "AND feedback_histories.time >= '#{DateTime.parse(start_date).to_fs(:db)}'" : ""`, and so on.

anathomical (approved):

services/QuillLMS/engines/evidence/lib/evidence/change_log.rb
dandrabik: I agree with the original comment and think we should use a positive param name like include_count. If you want to match old behavior, you can use the default for that, e.g. `include_count: true`

services/QuillLMS/app/queries/staff/rules_analysis_query.rb
```diff
@@ -0,0 +1,104 @@
...
+        first_feedback, second_feedback = rules_with_feedbacks.find{|x| x.id == r[:id]}&.feedbacks&.sort_by(&:order)
+        {
+            rule_uid: r[:rules_uid],
+            api_name: r[:rule_type],
+            rule_order: r[:rule_suborder],
+            first_feedback: first_feedback&.text || '',
+            second_feedback: second_feedback&.text || '',
+            rule_note: r[:rule_note],
+            rule_name: r[:rule_name],
+            avg_confidence: r[:avg_confidence] ? (r[:avg_confidence] * 100).round : nil,
```
dandrabik: Could you do this calc in the SQL?

services/QuillLMS/app/queries/staff/rules_analysis_query.rb
```diff
@@ -0,0 +1,104 @@
...
+
+    def activity_version_where_clause
+      activity_version ? "AND feedback_histories.activity_version = #{activity_version}" : ""
+    end
+
+    # rubocop:disable Metrics/CyclomaticComplexity
+    def post_query_transform(query_result)
+      rules_with_feedbacks = Evidence::Rule.where(id: query_result.pluck(:id)).includes(:feedbacks)
+      query_result.map do |r|
+        first_feedback, second_feedback = rules_with_feedbacks.find{|x| x.id == r[:id]}&.feedbacks&.sort_by(&:order)
```
dandrabik: Consider adding a `.map {|f| f&.text || ''}` to the end of this to get it in its ultimate form.

services/QuillLMS/app/queries/staff/rules_analysis_query.rb
```diff
@@ -0,0 +1,104 @@
...
+        {
+            rule_uid: r[:rules_uid],
+            api_name: r[:rule_type],
```
dandrabik: A lot of your query column names are subtly different than the return hash, which makes this mapping required. This could be simplified if you just used the same column names. e.g. If you used the same names, added that map above, and did the average in SQL (or define it above and add `avg_confidence:` to the merge). These 15 lines convert to:
```
result.merge(first_feedback:, second_feedback:)

File: services/QuillLMS/app/queries/staff/rules_analysis_query.rb [hunk 9]
```diff
@@ -0,0 +1,104 @@
...
+            rule_name: r[:rule_name],
+            avg_confidence: r[:avg_confidence] ? (r[:avg_confidence] * 100).round : nil,
+            total_responses: r[:total_responses],
+            strong_responses: r[:total_strong],
+            weak_responses: r[:total_weak],
+            repeated_consecutive_responses: r[:repeated_consecutive],
+            repeated_non_consecutive_responses: r[:repeated_non_consecutive],
+        }
+      end
+      # rubocop:enable Metrics/CyclomaticComplexity
```
dandrabik: I think we should avoid new code with these rubocop exceptions. Look to refactor this. Some suggestions above.

dandrabik (changes requested):
This overall looks good, but I'd like to see a refactor of a few things I mentioned in my comments (cleaning up a complex method, using positive params)

services/QuillLMS/spec/queries/staff/rules_analysis_query_spec.rb
brendanshean: I think this is an interesting approach.

brendanshean (approved):
Great work on these performance improvements!

dandrabik (approved):
:+1: Looks good to me. Thanks for making the requested changes and for moving this along. 🚢

PR #11625, opened 2024-03-19
lms/multi-count-students-in-filter-scope

anathomical:
## WHAT
- Modify filter scope query to count students once for each classroom that they're in (to match the behavior of the report itself)
- Also update the filter scope query to use the new, more efficient JOIN structure from our performance pass on other queries
## WHY
- We want admins to get an accurate addressable size calculation for this report
- We want all of our queries to be as performant as possible (and to use the same patterns when possible)
## HOW
- Count on `DISTINCT CONCAT` of the two ids that we care about 
- Update the JOIN chain to use the pattern established in our performance pass: use the filter materialized view

### Notion Card Links
https://quill.slack.com/archives/C06HG01B2BS/p1710785291565199

Using a known user that has repeated students, run the scope size query and compare it against the length of the Student report.  Confirm that in old code, the Student report has more entries than the count indicates, then confirm that the count and the length of the Student report are identical in this branch.

PR Checklist | Your Answer
------------ | -------------

### DIFF
--- a/services/QuillLMS/app/queries/admin_diagnostic_reports/diagnostic_performance_by_student_view_query.rb
+++ b/services/QuillLMS/app/queries/admin_diagnostic_reports/diagnostic_performance_by_student_view_query.rb
@@ -68,6 +68,7 @@ module AdminDiagnosticReports
         aggregate_id,
         #{aggregate_sort_clause},
         student_id,
+        student_name,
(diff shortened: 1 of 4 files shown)


dandrabik (approved):
:+1: LGTM.

PR #11630, opened 2024-03-19
lms/stop-double-counting-post-assigned

anathomical:
## WHAT
Only count one Pre or Post assignment per classroom

NOTE: This does address one issue with the calculation of Growth Percentages that Peter S flagged, but also surfaced a more complex issue that is not addressed in this PR: https://www.notion.so/quill/ADGR-QA-February-22-bb480e0255e542909ecee020b9a97742?pvs=4#cd2f38c11c71431f8574461704187699
## WHY
Apparently it's possible to assign two instances of the Post diagnostic to a classroom even if it only has a single Pre diagnostic assigned.  Not sure how that works, but it does exist in the wild.  In cases like this, we don't want to count that as two assignments.

Also, a performance improvement we made resulted in the inclusion of multiple assignments within the same classroom for Pre diagnostics getting into the rollup view, so we need to account for that.
## HOW
Change our `DISTINCT` value from `classroom_unit_id` and `user_id` to `classroom_id` and `user_id`

### Notion Card Links
https://www.notion.so/quill/ADGR-QA-February-22-bb480e0255e542909ecee020b9a97742?pvs=4#9bed3921cb94450c9e9b6d251c1e5b34

- Update tests
- Run query in console to confirm that the query run in the QA pass that found this error returns the correct value
- Re-test the user from the QA pass that surfaced the issue

PR Checklist | Your Answer
------------ | -------------

### DIFF
--- a/services/QuillLMS/app/queries/admin_diagnostic_reports/post_diagnostic_assigned_view_query.rb
+++ b/services/QuillLMS/app/queries/admin_diagnostic_reports/post_diagnostic_assigned_view_query.rb
@@ -2,7 +2,7 @@
 
 module AdminDiagnosticReports
   class PostDiagnosticAssignedViewQuery < DiagnosticAggregateViewQuery
-    def specific_select_clause = "COUNT(DISTINCT CONCAT(performance.post_classroom_unit_id, ':', performance.student_id)) AS post_students_assigned"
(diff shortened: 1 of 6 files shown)


brendanshean (approved):
Thanks for fixing this quickly.

dandrabik (approved):
:+1:

PR #11638, opened 2024-03-20
call for exact scores when completed tab is clicked and not before

emilia-friedberg:
## WHAT
Call for exact scores when the "Completed" tab is clicked, and not before. Also update the page so that the loading spinner does not take over the page, but shows up in individual rows next to the percentage until all the data is loaded.

Note: we are planning to wait to see if caching helps significantly reduce our load, and if not, deploy after that.

## WHY
This endpoint is slow so it's adding a lot of load to the site. Students are visiting the "To-do activities" tab more than the "Completed activities" tab, so we're currently hitting it more than is necessary. I did this initially to reduce visible load time for users, but we don't want that to make the site slower for everyone.

## HOW
Wait til the user actually clicks the "Completed activities" tab to load the exact score data.

### Screenshots

### Notion Card Links

Tested out the page to ensure that loading a) waits until you actually click "Completed activities" but then b) successfully completes once that happens.

PR Checklist | Your Answer
------------ | -------------

### DIFF
--- a/services/QuillLMS/app/assets/stylesheets/pages/student_dashboard/student_profile_unit.scss
+++ b/services/QuillLMS/app/assets/stylesheets/pages/student_dashboard/student_profile_unit.scss
@@ -80,6 +80,9 @@
     .score {
       display: flex;
       align-items: center;
+      .button-loading-spinner {
+        margin-right: 8px;
+      }
       div {
         display: inline-block;
         width: 24px;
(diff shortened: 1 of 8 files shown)


eadams17 (approved):

dandrabik (approved):
LGTM, thanks for putting this together quickly. If caching doesn't reduce the load enough, we'll deploy this next.

PR #11640, opened 2024-03-20
Cache Student Scores

dandrabik:
## WHAT
We are displaying scores to students. This is somewhat static data, but hits our two largest tables to calculate (`activity_sessions` and `concept_results`). Adding a cache to the controller endpoint and an cache.delete when a new activity is completed by a student.
## WHY
Performance reasons. This new controller endpoint is now 50% of our controller processing time.
## HOW
Cache.fetch in the controller and delete in the batch worker on activity completion.
### Screenshots
Screenshot 2024-03-20 at 12 14 50 PM

### Notion Card Links

Having Emilia QA this for time reasons.

PR Checklist | Your Answer
------------ | -------------

### DIFF
--- a/services/QuillLMS/app/controllers/profiles_controller.rb
+++ b/services/QuillLMS/app/controllers/profiles_controller.rb
@@ -63,28 +63,10 @@ class ProfilesController < ApplicationController
 
   def student_exact_scores_data
     exact_scores_data = params[:data].map do |ua|
-      activity_id = ua['activity_id']
-      classroom_unit_id = ua['classroom_unit_id']
-      cache_key = "#{Student::EXACT_SCORES_CACHE_KEY}/#{current_user.id}/#{activity_id}/#{classroom_unit_id}"
-
-      Rails.cache.fetch(cache_key, expires_in: 8.hours) do
-        activity_sessions = ActivitySession
-          .includes(:concept_results, :activity, :unit)
-          .where(
-            user_id: current_user.id,
-            activity_id: activity_id,
-            classroom_unit_id: classroom_unit_id,
-            state: ActivitySession::STATE_FINISHED
-          )
-
-        ua['sessions'] = activity_sessions.map { |as| format_activity_session_for_tooltip(as, current_user) }
-        ua['completed_attempts'] = activity_sessions.length
-
-        ua
-      end
+      student_exact_scores(current_user, ua['activity_id'], ua['classroom_unit_id'])
     end
 
-    render json: { exact_scores_data: exact_scores_data}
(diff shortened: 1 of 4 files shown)


emilia-friedberg (approved):
LGTM. Thanks for getting that up so fast.

services/QuillLMS/app/controllers/profiles_controller.rb
```diff
@@ -99,6 +87,29 @@ def staff
...
+        .includes(:concept_results, :activity, :unit)
+        .where(
+          user_id:,
+          activity_id:,
+          classroom_unit_id:,
+          state: ActivitySession::STATE_FINISHED
+        )
+
+      {
+        'sessions' => activity_sessions.map { |as| format_activity_session_for_tooltip(as, user) },
```
dandrabik, line 109: @emilia-friedberg One thing to confirm with you, I'm explicitly only returning these keys, and not the original `ua`. Rails parameters can't be cached if they are unpermitted, so we'd have to be explicit about the keys either in the params or in the return hash.

emilia-friedberg: Yeah, we need to include at least the the `ua_id` and the `classroom_unit_id` at the top-level here. In the production version of the code, we're sending: `ua_id`, `classroom_unit_id`, `unit_id`, and `activity_id`.

services/QuillLMS/app/workers/batch_save_activity_session_concept_results_worker.rb
```diff
@@ -15,13 +15,27 @@ def perform(concept_results, activity_session_id, activity_session_uid)
...
+      PusherTrigger.run(activity_session_uid, 'concept-results-partially-saved', "Concept results partially saved for activity session with uid: #{options['activity_session_uid']}, #{status.total - status.failures} succeeded, #{status.failures} failed")
     end
   end
 
+  private def delete_student_score_cache(activity_session_uid)
+    session = ActivitySession.find_by(uid: activity_session_uid)
+
+    return unless session
+
+    cache_key = "#{Student::EXACT_SCORES_CACHE_KEY}/#{session.user_id}/#{session.activity_id}/#{session.classroom_unit_id}"
```
anathomical, line 34: Looks like we're generating this cache key in two separate places.  Given the limited scope here, that's probably fine, but I did want to flag that in general we've tried to put the key generation logic in shared functions.

dandrabik: That's fair, will look at better ways to do this after deploying this (since it's somewhat time sensitive).

PR #11644, opened 2024-03-20
Cache student scores at student level.

dandrabik:
## WHAT
Edit: Also optimize queries to run one group query, instead of looping through many queries.
Adjust student score caching to be at the student level (not student/activity/classroom_unit level). 
## WHY
This controller action is still slow. We're spending a lot of time in Redis per controller action and making an average of 30 calls. This should reduce that time drastically and make subsequent calls faster. The downside is this cache will be flushed more often (when ever a new activity is finished, the whole cache for a student is flushed, vs. just a section of it), but I think it will be a net gain.
## HOW
Move the caching outside of the loop. Do one main query, put the results in a hash and look them up.
### Screenshots
Screenshot 2024-03-20 at 4 47 28 PM

### Notion Card Links

I QA'd this feature on my staging and will ask Emilia to do a quick QA as well.

PR Checklist | Your Answer
------------ | -------------

### DIFF
--- a/services/QuillLMS/app/controllers/profiles_controller.rb
+++ b/services/QuillLMS/app/controllers/profiles_controller.rb
@@ -94,25 +94,25 @@ class ProfilesController < ApplicationController
   private def student_score_cache_key = User.student_scores_cache_key(current_user.id, params[:classroom_id])
 
   private def exact_scores_data_all(user, data, classroom_id)
-    user_id = user.id
-    classroom_unit_id = data.map{|h| h[CLASSROOM_UNIT_ID]}
-    activity_id = data.map{|h| h[ACTIVITY_ID]}
+    activity_sessions_grouped = student_activity_sessions_grouped(user.id, data)
 
-    activity_sessions_grouped = ActivitySession
+    data.map do |user_activity|
+      key = [user_activity[ACTIVITY_ID]&.to_i, user_activity[CLASSROOM_UNIT_ID]&.to_i]
+      activity_sessions = activity_sessions_grouped[key] || []
+      student_exact_scores(user, user_activity, activity_sessions)
+    end
+  end
+
+  private def student_activity_sessions_grouped(user_id, data)
+    ActivitySession
       .includes(:unit, concept_results: :concept, activity: :classification)
       .where(
-        user_id:,
-        activity_id:,
-        classroom_unit_id:,
+        user_id: user_id,
+        activity_id: data.map{|h| h[ACTIVITY_ID]},
+        classroom_unit_id: data.map{|h| h[CLASSROOM_UNIT_ID]},
         state: ActivitySession::STATE_FINISHED
       )
       .group_by {|as| [as.activity_id, as.classroom_unit_id]}
-
-    data.map do |ua|
-      key = [ua[ACTIVITY_ID]&.to_i, ua[CLASSROOM_UNIT_ID]&.to_i]
-      activity_sessions = activity_sessions_grouped[key] || []
-      student_exact_scores(user, ua, activity_sessions)
-    end
   end
 
   private def student_exact_scores(user, unit_activity_params, activity_sessions)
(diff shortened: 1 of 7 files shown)


dandrabik:
@emilia-friedberg If you get a chance tomorrow morning, could you do a quick QA on staging (it's deployed to my env).

emilia-friedberg (changes requested):
Hey Dan -- I'm happy to QA this, but I think we need to be caching by at least the classroom id in addition to the student. Right now, this call is made for every classroom (that's how the "scores" query that this is built on is scoped). Otherwise, changing classrooms would result in the wrong data showing. We can either update the frontend to pass back the classroom id as one of the params (let me know if you'd like me to work on that), or find the first unit activity's classroom unit and grab the classroom id from there.

dandrabik:
> Hey Dan -- I'm happy to QA this, but I think we need to be caching by at least the classroom id in addition to the student. Right now, this call is made for every classroom (that's how the "scores" query that this is built on is scoped). Otherwise, changing classrooms would result in the wrong data showing. We can either update the frontend to pass back the classroom id as one of the params (let me know if you'd like me to work on that), or find the first unit activity's classroom unit and grab the classroom id from there.

Ah, ok, good catch. Let me take a pass at this. It's not super urgent since the page is stable now, but will look at this later today. Thanks for the review.

dandrabik:
> > Hey Dan -- I'm happy to QA this, but I think we need to be caching by at least the classroom id in addition to the student. Right now, this call is made for every classroom (that's how the "scores" query that this is built on is scoped). Otherwise, changing classrooms would result in the wrong data showing. We can either update the frontend to pass back the classroom id as one of the params (let me know if you'd like me to work on that), or find the first unit activity's classroom unit and grab the classroom id from there.
> 
> Ah, ok, good catch. Let me take a pass at this. It's not super urgent since the page is stable now, but will look at this later today. Thanks for the review.

@emilia-friedberg I've adjusted the caching on this page and also added some query optimization (one group query put into a hash that the rest pull from). This PR is ready for re-review. I've also done a basic QA on my staging env, but would love a second set of eyes since I'm not too familiar with this feature.

services/QuillLMS/app/controllers/profiles_controller.rb
```diff
@@ -87,34 +87,44 @@ def staff
...
-      {
-        'sessions' => activity_sessions.map { |as| format_activity_session_for_tooltip(as, user) },
-        'completed_attempts' => activity_sessions.length,
-        'activity_id' => activity_id,
-        'classroom_unit_id' => classroom_unit_id,
-        'ua_id' => ua_id
-      }
+  ACTIVITY_ID = 'activity_id'
+  CLASSROOM_UNIT_ID = 'classroom_unit_id'
+  UA_ID = 'ua_id'
```
emilia-friedberg: I feel like we usually group constants at the top of a file. I'm not sure it really matters either way because I'm normally cmd+f-ing anyway, but we might want to move these for consistency.

services/QuillLMS/app/controllers/profiles_controller.rb
```diff
@@ -87,34 +87,44 @@ def staff
...
+
+  private def student_score_cache_key = User.student_scores_cache_key(current_user.id, params[:classroom_id])
+
+  private def exact_scores_data_all(user, data, classroom_id)
+    activity_sessions_grouped = student_activity_sessions_grouped(user.id, data)
```
emilia-friedberg, line 97: Yay for the efficiency gain here! Good call.

emilia-friedberg (approved):
Deployed the most recent version to my staging environment, QAed, and all looks well. One tiny note about constant placement but otherwise LGTM!

services/QuillLMS/app/controllers/profiles_controller.rb
dandrabik: That's fair. I've starting putting constants close to their use, but I think that breaks from the pattern we use everywhere else, so I'll move them to the top.

PR #11651, opened 2024-03-21
updates to research diagnostic query

happythenewsad:
## WHAT
We're pausing work on this research query for now, but I want to check in the code in case we need it in the future. It's not bundled into any production bundles. 

1. send data concatenated as strings to the UDF, rather than an array of arrays. This is because, according to Google's documentation, ARRAY_AGG order is not guaranteed when multiple ARRAY_AGGs are used in a single query. 
2. update activity recommendations lookup data structure 
3. stop rolling up concept_result scores by student 
4. Remove the tierUDF function

## WHY
1. I initially moved away from ARRAY_AGG on the concern that data was being misordered. In practice, it wasn't, but it's a risk, according to the documentation.
2. activity IDs in this data structure were incorrect 
3. I misunderstood how concept_results related to activity_sessions. We should differentiate concept_result scores by student and skill group, rather than rolling them up by student and skill group.
4. Curriculum requested fractions for now. And also, we can now do tiering in the main rollup, if we want.

### Screenshots

### Notion Card Links

no QA - not production code

PR Checklist | Your Answer
------------ | -------------
Have you deployed to Staging? | n/a 

### DIFF
--- a/services/QuillLMS/client/app/modules/BigQueryUDFs.js
+++ b/services/QuillLMS/client/app/modules/BigQueryUDFs.js
@@ -76,26 +76,26 @@ export function studentwiseSkillGroupUDF(elements) {
   //source: https://docs.google.com/spreadsheets/d/1JFey0UpMkmPzkQtZKsr_FdXRXnNEDFXZe52H7dUMg9E/edit#gid=0
   const skillGroupsByActivity = {
     1161: [
-      { name: 'Sentences with To Be' },
-      { name: 'Sentences With Have' },
-      { name: 'Sentences With Want' },
-      { name: 'Listing Adjectives and Nouns' },
-      { name: 'Writing Questions' }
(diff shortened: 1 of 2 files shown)


services/QuillLMS/client/app/modules/BigQueryUDFs.js
```diff
@@ -55,55 +94,77 @@ export function studentwiseSkillGroupUDF(scores, activityIds, completedAts, skil
     return row.score
   }
 
+  function deduplicateAndAverageScores(arr) {
```
happythenewsad, line 97: Note how, like previously, `deduplicateAndAverageScores` is defined as an exportable function above, but redefined within the parent UDF here.  This allows unit testing. Not DRY at all. Can be fixed by writing a special-purpose JS bundler, should we continue with this pattern in the future.

brendanshean (approved):
This feels like useful code for the future whether we use it now or not.

PR #11654, opened 2024-03-21
Add Experiment Runner

brendanshean:
## WHAT
1. Add a Gemini Pro api wrapper similar to the OpenAI one.
2. Add `RunExperimentWorker` job that will be queued via a controller
3. Add a bunch of attr_readonly to our models.
4. Add `run` method to Experiment

## WHY
1. We need a way to get feedback from Gemini
2. This will allow for easy queuing of experiments
3. When experiments are reported, it's important that the attributes of the relevant models cannot be changed.
4. This will be the entry point to running an experiment

## HOW
1.  Model this after the OpenAI API concern
2.  Worker is defined here and calls `.run` on the relevant Experiment record
3.  Add `attr_readonly` to relevant models
4.  Iterate over PassagePromptResponses and create new LLM Responses, with crux being:
```
def self.save_llm_feedback!(passage_prompt_response:, llm_prompt:, llm_client:)
  feedback = llm_client.run(prompt: llm_prompt.feedback_prompt(passage_prompt_response.response))
  create!(feedback:, passage_prompt_response:)
end
```

### Screenshots

### Notion Card Links

Testing locally via experiments

PR Checklist | Your Answer
------------ | -------------


services/QuillLMS/engines/evidence/app/models/evidence/research/gen_ai/experiment.rb
```diff
@@ -17,11 +18,44 @@ module Evidence
...
+
+        delegate :llm_client, to: :llm_config
+
+        def completed! = update!(status: COMPLETED)
+        def running! = update!(status: RUNNING)
+        def failed! = update!(status: FAILED)
+
+        def run
+          running!
+          create_llm_prompt_responses_feedbacks
```
dandrabik, line 46: This is probably fine for now, but we might consider making this a batch of jobs that runs async.

services/QuillLMS/engines/evidence/app/models/evidence/research/gen_ai/experiment.rb
```diff
@@ -17,11 +18,44 @@ module Evidence
...
+          completed!
+        rescue StandardError => e
+          experiment_errors << e.message
+          failed!
+        end
+
+        # TODO: limit should be abstracted out for real experiments
+        private def create_llm_prompt_responses_feedbacks
+          passage_prompt_responses.limit(3).each do |passage_prompt_response|
+            LLMPromptResponseFeedback.save_llm_feedback!(passage_prompt_response:, llm_prompt:, llm_client:)
```
dandrabik: Reading the name `.save_llm_feedback!` I was surprised to see it both hit the API and then save. Consider renaming or breaking those actions up into two steps here.

services/QuillLMS/engines/evidence/app/models/evidence/research/gen_ai/llm_config.rb
```diff
@@ -14,8 +14,21 @@ module Evidence
...
+        VENDORS = [
+          GOOGLE = 'google'
+        ]
+
         validates :vendor, presence: true
         validates :version, presence: true
+
+        attr_readonly :vendor, :version
+
+        def llm_client
```
dandrabik: Consider converting this to a hash lookup instead of a case statement.
```ruby
VENDOR_MAP = {
  GOOGLE => Evidence::Gemini::Completion
}
def llm_client = VENDOR_MAP.fetch(vendor) { raise UnsupportedVendorError }

File: services/QuillLMS/engines/evidence/lib/evidence/gemini/concerns/api.rb [hunk 4]
```diff
@@ -0,0 +1,81 @@
...
+
+        private def body = request_body.merge(safety_settings:).to_json
+
+        private def headers = { "Content-Type" => "application/json" }
+
+        private def safety_settings = SAFETY_SETTINGS
+
+        private def timeout = TIMEOUT
+
+        # TODO: remove puts statements once this is stable
```
dandrabik, line 62: If we are implementing backoff and retry logic, we might consider making this action a job instead, and just adjust sidekiq configs to make it fit our run conditions.

dandrabik (approved):
I added some general thoughts about making this a batch of jobs to avoid building a backoff/retry setup and some small refactoring suggestions.

services/QuillLMS/engines/evidence/app/models/evidence/research/gen_ai/experiment.rb
brendanshean: The backoff/retry was a first pass at making sure this works and getting a feel for the rate limiting of the API, but I agree that this should eventually be a a background job.

services/QuillLMS/engines/evidence/lib/evidence/gemini/concerns/api.rb
brendanshean: I agree.  I think for this first pass, I wanted the simplicity of keeping the entire computation in one thread due to the rate limit of 60 requests / min.

PR #11662, opened 2024-03-22
Lms/fix growth aggregation

anathomical:
## WHAT
Fix both growth percentage aggregation and completion count aggregation in the same query
## WHY
Data needs to be split out by skill for the weird way we calculate aggregate growth rates, but that duplicates completion counts.  This code resolves that tension, allowing both calculations to be run in the same query.
## HOW
Allow growth rate calculation to to be the primary way we organize things, then divide the duplication back out of the completion count calculation.

Take a handful of admins, filter to a specific teacher, group by classroom.  Then log into those teachers' accounts on prod and compare the admin report values to what the teachers have in their individual reports.

PR Checklist | Your Answer
------------ | -------------
Have you added and/or updated tests? | I propose to defer specs for this.  The specific pre-conditions needed to really capture this edge case in a spec are pretty involved, and given the timeline we're working on here, I'm inclined to move forward for now without them.  Open to discussion.

### DIFF
--- a/services/QuillLMS/app/queries/admin_diagnostic_reports/post_diagnostic_completed_view_query.rb
+++ b/services/QuillLMS/app/queries/admin_diagnostic_reports/post_diagnostic_completed_view_query.rb
@@ -10,7 +10,7 @@ module AdminDiagnosticReports
             aggregate_id,
             name,
             group_by,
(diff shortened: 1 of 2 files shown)


dandrabik (approved):
:+1:

happythenewsad (approved):

PR #11673, opened 2024-03-26
lms/aggregate-then-average-growth-score

anathomical:
## WHAT
- Use the same rounding logic for growth score as teacher reports.  That is: average the growth scores for each skill, THEN round to the nearest whole percentage instead of rounding then averaging.
- Temporarily remove the ELL diagnostics from the report while we deal with a score-calculating issue.
## WHY
- So that we don't get occasional edge cases where the different rounding approaches give us off-by-one results when comparing the teacher and admin reports.
- Turns out that the ELL Starter and ELL Intermediate diagnostics have the same scoring issue as the Advanced diagnostic.  While the ELL Advanced diagnostic does not have this issue, it would be weird to have only our least-used ELL diagnostic in the report without the others.
## HOW
- Instead of rounding Pre and Post scores and then subtracting to figure out growth, subtract un-rounded Pre and Post scores, and round the resulting difference.  This was `10.5 - 5.4` doesn't end up being rounded to `11 - 5` and giving us `6`.  Instead it gets subtracted, then rounded: `10.5 - 5.4 = 5.1`, rounded to `5`.
- Remove the ids for the ELL diagnostics from the back-end whitelist of diagnostics to report on, and also remove the ELL diagnostics from the front-end dropdown for selecting specific diagnostics on the Student and Skills report pages

### Screenshots
OLD
image
NEW
image

### Notion Card Links
https://www.notion.so/quill/ADGR-QA-February-22-bb480e0255e542909ecee020b9a97742?pvs=4#c242f37a4988412fa9d7fc190794fe9a

Starting with the users that caused us to first notice the issue, compare specific classroom-level stats in the admin report to the numbers being shown to the Teachers in the teacher reports.  Confirm that the data matches between the two.  Then sanity check by looking at another two or three admins worth of data to confirm everything matches up.

PR Checklist | Your Answer
------------ | -------------
Have you added and/or updated tests? |  Some small updates around the `attempt_number = NULL` expectation, but no coverage for the rounding change.  The pre-conditions for a spec are relatively complex, and we'd like to get this fix out the door quickly.  I've added a TODO to the spec file, and propose deferring writing it for now.


dandrabik (approved):
👍 LGTM

eadams17 (approved):

PR #11676, opened 2024-03-26
start sending cues to lms to store in concept result metadata

emilia-friedberg:
## WHAT
Send cues to LMS as part of the `concept_results` hash to store in the concept result metadata.

## WHY
Right now, we only store cues in the concept results as part of a string attached to the directions (ex: ""Combine the sentences. (After, Even though, Since)"). This works fine for the teacher activity analysis, where we just display this string as it's currently encoded, but for the student view, we want to break it out. To that end, it feels like a better longterm structure to me to have cues be stored independently, in case we were to change how this string is formatted at a later date. When I write the frontend implementation for this project, my plan is to first check for this `cues` attribute, and then parse the existing `directions` string if it's absent, so both past and future data will render cues correctly. 

## HOW
Just add in the logic to pass the cues in all the question types we do or could use cues.

### Screenshots

### Notion Card Links
https://www.notion.so/quill/Start-saving-cues-in-the-metadata-for-concept-results-9d7cc40dae7f490695b90efbe5d99983?pvs=4

Played through activity sessions with cued questions for each of the possible types (grammar, connect sentence combining, connect fill in the blank, diagnostic sentence combining, diagnostic fill in the blank) and looked at the saved concept results to confirm this is getting passed through correctly.

PR Checklist | Your Answer
------------ | -------------


eadams17 (approved):

brendanshean:
Screenshot 2024-03-27 at 7 11 28 AM
😎

brendanshean (approved):
This sounds like a sensible change.

PR #11677, opened 2024-03-26
lms/increase-adgr-worker-priority

anathomical:
## WHAT
Put Admin Diagnostic cache workers into the CRITICAL_EXTERNAL queue
## WHY
So that they don't get trapped behind other lower-priority jobs in the DEFAULT queue
## HOW
Set the `queue` Sidekiq option in the job classes

I don't see an obvious way of proving that this will improve experienced performance without worker traffic like we have in production.  Open to suggestions.

PR Checklist | Your Answer
------------ | -------------
Have you added and/or updated tests? |  No tests on config values


dandrabik (approved):
:+1:

PR #11680, opened 2024-03-26
Proofreader/ grammar follow-up bug

eadams17:
## WHAT
fix issue where follow up grammar activities for Proofreader sessions were not being supplied the correct amount of questions

## WHY
we want the activity to have the expected number of questions

## HOW
check by number of incorrect concepts rather than number of incorrect concept results

### Screenshots

### Notion Card Links
https://www.notion.so/quill/Proofreader-not-linking-to-Grammar-correctly-0e7c02de2a5b4a2886bb5063526c0750?pvs=4

I tested some activities on staging and also had Rachel verify that the number of questions matched the number of concepts

PR Checklist | Your Answer
------------ | -------------

### DIFF
--- a/services/QuillLMS/client/app/bundles/Grammar/actions/session.ts
+++ b/services/QuillLMS/client/app/bundles/Grammar/actions/session.ts
@@ -166,7 +166,7 @@ export const startListeningToFollowUpQuestionsForProofreaderSession = (proofread
   }
 }
 
-const handleProofreaderSession = (proofreaderSession, state) => {
+export const handleProofreaderSession = (proofreaderSession, state) => {
   return dispatch => {
 
     // if there is no proofreader session or concept results for that session, we can't do anything here
@@ -194,7 +194,7 @@ const handleProofreaderSession = (proofreaderSession, state) => {
         concepts[cr.concept_uid] = { quantity }
       }
     })
-    debugger
+
     dispatch(saveProofreaderSessionToReducer(proofreaderSession))
     dispatch(getQuestionsForConcepts(concepts, 'production'))
   }
--- a/services/QuillLMS/client/app/bundles/Grammar/actions/session.ts
+++ b/services/QuillLMS/client/app/bundles/Grammar/actions/session.ts
@@ -174,12 +174,19 @@ const handleProofreaderSession = (proofreaderSession, state) => {
     // that all gets processed by the normal `setSessionReducerToSavedSession` function and we don't need to set new ones
     if (!proofreaderSession || !proofreaderSession.conceptResults || proofreaderSession.answeredQuestions || state.session.proofreaderSession) { return }
 
-    const concepts: { [key: string]: { quantity: 1|2|3 } } = {}
-    const incorrectConcepts = proofreaderSession.conceptResults.filter(cr => cr.metadata.correct === 0)
+    const concepts: { [key: string]: any } = {}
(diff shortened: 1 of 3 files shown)


services/QuillLMS/client/app/bundles/Grammar/actions/session.ts
```diff
@@ -166,27 +166,35 @@ export const startListeningToFollowUpQuestionsForProofreaderSession = (proofread
...
   return dispatch => {
 
     // if there is no proofreader session or concept results for that session, we can't do anything here
     // if there are already answered questions or the session is already in the state
     // that all gets processed by the normal `setSessionReducerToSavedSession` function and we don't need to set new ones
     if (!proofreaderSession || !proofreaderSession.conceptResults || proofreaderSession.answeredQuestions || state.session.proofreaderSession) { return }
 
-    const concepts: { [key: string]: { quantity: 1|2|3 } } = {}
-    const incorrectConcepts = proofreaderSession.conceptResults.filter(cr => cr.metadata.correct === 0)
+    const concepts: { [key: string]: any } = {}
```
brendanshean: I found the changing of type for `concepts` value to be confusing.  I would introduce an intermediate variable to hold the incorrectConceptUIDs:

```
const incorrectConceptUIDs = proofreaderSession.conceptResults.reduce((results, cr) => {
  if (metadata.correct === 0 && !results.includes(cr.concept_uid)) {  results.push(cr.concept_uid); }
  return results;
}, []);


const quantityMap = { '1-4': 3, '5-9': 2, '10+': 1 };  // These magic might be worth some explanation:
const quantity = incorrectConceptUIDs.length > 9 ? quantityMap['10+'] 
               : incorrectConceptUIDs.length > 4 ? quantityMap['5-9']
               : quantityMap['1-4']
             
 const concepts: { [key: string]: { quantity: 1|2|3 } } = {}
 incorrectConceptUIDs.forEach(concept_uid => { concepts[concept_uid] = { quantity }; } } 
 ```

brendanshean (approved):
One refactoring comment, but otherwise looks good.

services/QuillLMS/client/app/bundles/Grammar/actions/session.ts
emilia-friedberg: I like this revision, though I'd probably be inclined to stick with the existing quantity logic and just add a comment explaining it.

emilia-friedberg (approved):
Love the tests!!!

services/QuillLMS/client/app/bundles/Grammar/actions/session.ts
eadams17: I like this idea-- I do just want to note that concepts results can have the same concept so that's why I opted for a hash. I could just put `incorrectConceptUIDs` inside of a Set though to remove duplicates

eadams17: I like this idea-- I'll add a comment for some further explanation

brendanshean: I might be mixing something up here, but I think `incorrectConceptUIDs` should not have duplicates due to `!results.includes(cr.concept_uid)`

eadams17: you're right! I missed that initially

PR #11691, opened 2024-03-27
Change automatic searching in Admin panel to manual

cissyxyu:
## WHAT
Admin have complained about very slow/non-responsive filters and sorting in the grading panels for a while. This PR attempts to fix this problem by removing the "one-search-per-click" UI that we currently implement, and instead shows a "Show Results" button to let admin choose when they want to actually perform a search.

## WHY
Searches are slow/non-responsive because right now, admin often click many filters very quickly when they're trying to get to a specific set of filters on and off. This often triggers 5-6 search calls to the CMS in the span of seconds, which are very CPU-expensive and ultimately cause the LMS search workers to time out and return 500 errors. Our current UI is faulty because admin actually don't even want to perform these extra search actions -- they just want to get results for one very specific set of filters. Changing automatic search to manual search helps them only send search queries when they actually have the filter state they desire.

Also -- I added a loading spinner when responses are loading so admin can tell when a search has returned, since sometimes the search takes >5 seconds and admin think the search didn't work, and they click more buttons, sending more requests and slowing down the CMS even more.

## HOW
In Connect, Grammar and Diagnostic filter panels, add a "Show Results" button. Change the logic so search is only performed when this button is clicked. Also, show a spinner instead of results when a search is still being performed.

### Screenshots
Screenshot 2024-03-28 at 12 39 39 AM
Screenshot 2024-03-28 at 12 39 03 AM

### Notion Card Links
https://www.notion.so/quill/The-filters-in-the-Connect-Diagnostic-CMS-are-not-working-as-expected-10466ac87020487e87eed7f72eff4681?pvs=4

Rachel and I have tested using filters, sorting, and searching on my staging environment. We also tested pre-existing features like changing page number and updating responses to make sure nothing existing is broken.

PR Checklist | Your Answer
------------ | -------------
Have you added and/or updated tests? |  no tests in this part of the code base

### DIFF
--- a/services/QuillLMS/client/app/bundles/Connect/components/questions/responseComponent.jsx
+++ b/services/QuillLMS/client/app/bundles/Connect/components/questions/responseComponent.jsx
@@ -7,6 +7,7 @@ import {
   QuestionBar,
   ResponseSortFields,
   ResponseToggleFields,
+  Spinner,
   hashToCollection,
   responsesWithStatus,
 } from '../../../Shared/index';
@@ -54,6 +55,7 @@ class ResponseComponent extends React.Component {
       health: {},
       gradeBreakdown: {},
       enableRematchAllButton: true,
+      isLoadingResponses: false,
     };
   }
 
@@ -90,6 +92,10 @@ class ResponseComponent extends React.Component {
   unsubscribeToQuestion = (questionID) => {
     const { dispatch, } = this.props
     dispatch(questionActions.removeSubscription(questionID));
+  }
+
+  setResponsesLoaded = () => {
+    this.setState({isLoadingResponses: false})
   }
 
   getBoundsForCurrentPage = length => {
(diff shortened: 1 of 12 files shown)


services/QuillLMS/client/app/bundles/Connect/components/questions/responseComponent.jsx
```diff
@@ -485,7 +488,17 @@ class ResponseComponent extends React.Component {
...
+
+    if (isLoadingResponses && viewingResponses) {
+      return (
+        <div className="loading-spinner-container">
+          <Spinner />
+        </div>
+      );
+    }
+
+    if (viewingResponses) {
```
anathomical, line 501: Not a huge deal, but you could re-arrange your if conditionals with an early return:
```javascript
if (!viewingResponses) return
if (isLoadingResponses) { return ( ... spinner stuff ) }
return <ResponseList ...
```

anathomical (approved):
A note on `if` organization, but this looks good to me overall.

happythenewsad (approved):

PR #11698, opened 2024-03-28
lms/admin-diagnostic-growth-tooltip-data

anathomical:
## WHAT
Add an additional `pre_score_completed_post` value to the API.  We want a rollup of the pre scores of only students who have also completed the post diagnostic so that we can help explain growth percentage to users.

Also: using a pattern similar to the one established for benchmarking to lock down an expected payload to catch regressions moving forward.
## WHY
We think that the specific way that we calculate growth percentages could lead to confusion on pages where we show an average score for all students on the Pre diagnostic and the Post diagnostic, but the growth percentage isn't just the difference between the two.  The intention is to add a tool-tip that goes into more detail, and this new value will be put in that tooltip.
## HOW
Calculate a Pre diagnostic score that includes only the students who have completed an associated Post diagnostic and pass that data back in the API

### Notion Card Links
https://www.notion.so/quill/ADGR-Improve-the-UX-of-the-No-growth-state-on-the-Performance-by-Skill-tab-fcce1690908c4eb6a0abbdecb5d79c7c

- In the Rails console, run the Query and sanity check the returned values
- - confirm that this matches base pre score if all students have completed both pre and post
- - confirm that this number is higher than post score when post scores exist and growth percentage is 0
- - confirm that this value is `NULL` when no students have completed the post
- Deploy to staging and ensure that the new value is included in the API calls even though we aren't using it in the front-end yet

PR Checklist | Your Answer
------------ | -------------


services/QuillLMS/lib/tasks/queries.thor
```diff
@@ -118,12 +114,66 @@ class Queries < Thor
...
+
+    student_queries = {
+      'diagnostic-students-view' => ::AdminDiagnosticReports::DiagnosticPerformanceByStudentViewQuery,
+      'student-recommendation' => ::AdminDiagnosticReports::DiagnosticRecommendationsByStudentQuery,
+      'filter-scope' => ::AdminDiagnosticReports::StudentCountByFilterScopeQuery
+    }
+
+    timeframe_start = DateTime.parse(DEFAULT_START)
+    timeframe_end = DateTime.parse(DEFAULT_END)
+    school_ids = [38811,38804,38801,38800,38779,38784,38780,38773,38765,38764]
```
brendanshean, line 144: Maybe some explanation for the magic numbers is warranted?

brendanshean (approved):
I like this approach for dealing with regressions in the query results.  Nice work!

dandrabik (approved):
:+1: Nice additional script for testing regressions!

PR #11699, opened 2024-03-28
Make sure to avoid returning more than the 500 students from spec

anathomical:
Because of the way we do the data roll-ups with aggregate rows by skill, and with the number of skills for each diagnostic being potentially different, there's not an easy way to get this limit accurate in SQL, so we pull potentially extra records in SQL and need to cut off any extras here.

## WHAT
Make sure to avoid returning more than the 500 students from spec
## WHY
Giant tables get very unruly and unusable in a browser, so we want to set a maximum size on the table and have settled on 500 as a reasonable limit.  In an ideal world, we'd handle this limit in SQL, but the way we do aggregation (where a given student "row" may have multiple rows in the SQL data representing performance on different skills, or a student "row" may only have one row if the student has only been assigned the diagnostic but have never completed it) doesn't give us the precision we want.  We've been using `LIMIT 5000` on the SQL to make sure we don't pull way more data than we need, but this isn't sufficient to ensure only 500 records are available to the front-end.
## HOW
Add a `slice(0, 500)` call to the data after we've finished post-processing it into its final form where each record represents one student.

### Screenshots
OLD
image
NEW
image

### Notion Card Links
https://www.notion.so/quill/ADGR-Display-a-download-link-on-the-Performance-by-Student-tab-when-an-admin-loads-the-maximum-n-838ccf65f35f4f4d9edcf17f430c5ab3

- Log into prod and find a user who is getting more than 500 records back on the Student report
- Log in as the same user on staging with the new code
- - Confirm that the UI shows only 500 records available
- - Using the browser tools Network tab, confirm that only 500 records are coming back from the API

PR Checklist | Your Answer
------------ | -------------

### DIFF
--- a/services/QuillLMS/app/queries/admin_diagnostic_reports/diagnostic_performance_by_student_view_query.rb
+++ b/services/QuillLMS/app/queries/admin_diagnostic_reports/diagnostic_performance_by_student_view_query.rb
@@ -7,6 +7,7 @@ module AdminDiagnosticReports
     attr_reader :diagnostic_id
 
     AGGREGATE_COLUMN = :student_id
+    MAX_STUDENTS_TO_RETURN = 500
 
     def initialize(diagnostic_id:, **options)
       raise InvalidDiagnosticIdError, "#{diagnostic_id} is not a valid diagnostic_id value." unless DIAGNOSTIC_ORDER_BY_ID.include?(diagnostic_id.to_i)
@@ -95,6 +96,7 @@ module AdminDiagnosticReports
       result.group_by { |row| "#{row[:classroom_id]}:#{row[self.class::AGGREGATE_COLUMN]}" }
         .values
         .map { |group_rows| build_diagnostic_aggregates(group_rows) }
+        .slice(0, MAX_STUDENTS_TO_RETURN)
     end
 
     private def valid_aggregation_options = ['student']
--- a/services/QuillLMS/spec/queries/admin_diagnostic_reports/diagnostic_performance_by_student_view_query_spec.rb
+++ b/services/QuillLMS/spec/queries/admin_diagnostic_reports/diagnostic_performance_by_student_view_query_spec.rb
@@ -7,24 +7,38 @@ module AdminDiagnosticReports
     include_context 'Admin Diagnostic Aggregate CTE'
     include_context 'Pre Post Diagnostic Skill Group Performance View'
 
+    let(:query_args) do
+      {
+        timeframe_start: timeframe_start,
+        timeframe_end: timeframe_end,
+        school_ids: school_ids,
+        grades: grades,
+        teacher_ids: teacher_ids,
+        classroom_ids: classroom_ids,
+        diagnostic_id: pre_diagnostic.id
+      }
+    end
+
+    context 'when finding more than 500 records' do
+      let(:runner_results) do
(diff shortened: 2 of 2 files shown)


brendanshean (approved):

dandrabik (approved):
I'm approving to get this out since it seems like a reasonable stop-gap. It would be really nice to get this limit into SQL if possible, since we could end up querying/compile/ parsing huge amounts of data then throwing most it away. Let's chat on Monday and if there's anything we can do (in the meantime, let's ship this).

anathomical:
> I'm approving to get this out since it seems like a reasonable stop-gap. It would be really nice to get this limit into SQL if possible, since we could end up querying/compile/ parsing huge amounts of data then throwing most it away. Let's chat on Monday and if there's anything we can do (in the meantime, let's ship this).

Glad to hop on a call about this, but I wanted to make sure I laid out the issue in a public place for posterity.  The reason this is hard to do completely in SQL is that the number of rows we need is variable.  When we query for this student data, for each student we expect to either get 1 row or X rows (depending on whether they've only been assigned the diagnostic, or if they've been assigned and completed the diagnostic).  Where X is the number of skills that the specific diagnostic we're getting data for has.  X can be 5, 6, 7, or 8 depending on the diagnostic in question.  (It could theoretically be even higher, but all existing diagnostics have 5-8 skills associated with them.)  Currently, there is a SQL `LIMIT 5000` clause on this query, which theoretically could be lowered to `LIMIT 4000` since the current maximum number of rows we'd want back is 500 * 8 for some theoretical set of students where they've all completed the diagnostic.  Further wrinkle: because we must do a SQL side `ORDER BY`, the full number of un-LIMITed rows has to get processed, so the `LIMIT` is only saving us payload size (which is worth doing, of course), not total data processed size.

dandrabik:
> > I'm approving to get this out since it seems like a reasonable stop-gap. It would be really nice to get this limit into SQL if possible, since we could end up querying/compile/ parsing huge amounts of data then throwing most it away. Let's chat on Monday and if there's anything we can do (in the meantime, let's ship this).
> 
> Glad to hop on a call about this, but I wanted to make sure I laid out the issue in a public place for posterity. The reason this is hard to do completely in SQL is that the number of rows we need is variable. When we query for this student data, for each student we expect to either get 1 row or X rows (depending on whether they've only been assigned the diagnostic, or if they've been assigned and completed the diagnostic). Where X is the number of skills that the specific diagnostic we're getting data for has. X can be 5, 6, 7, or 8 depending on the diagnostic in question. (It could theoretically be even higher, but all existing diagnostics have 5-8 skills associated with them.) Currently, there is a SQL `LIMIT 5000` clause on this query, which theoretically could be lowered to `LIMIT 4000` since the current maximum number of rows we'd want back is 500 * 8 for some theoretical set of students where they've all completed the diagnostic. Further wrinkle: because we must do a SQL side `ORDER BY`, the full number of un-LIMITed rows has to get processed, so the `LIMIT` is only saving us payload size (which is worth doing, of course), not total data processed size.

Ok, thanks for that context, that's helpful. I think the fact that we are limiting to 5000 in the query makes me feel a little better. 

There is likely some optimization we could do here with an additional, purposely simple query, e.g. query students who have any data and take the first 500 (or potentially the 500 most active if that's easy to calc), then use the student ids in the next query to pull all of the data.  For now, if there already is some limit, I don't want to add that extra complication. Let's leave this as is for now. Thanks for following up.

PR #11700, opened 2024-03-28
Add Chain-of-Thought to Example Feedback

brendanshean:
## WHAT
1.  Add `chain_of_thought` (i.e. CoT) attribute to `Evidence::Research::GenAI::ExampleFeedback`
1.  Currently, the end of our LLMPrompt contain a label (e.g. `Feedback:` or `Evaluation`).  Let's parameterize this label as `coda` and add it as an attribute on `LLMPromptTemplate`

## WHY
1.  We'd like to run experiments to see if CoT helps with accuracy.
1.   The use of CoT is related to the coda we want to make this dependency explicit.

## HOW
1.  Add a migration with the field and then update the PromptBuilder substitution to encode `examples_chain_of_thought`
1.  Add a migration with the field and then update the Experiment#run method to use the `prompt_response_coda` method.

### Screenshots

### Notion Card Links

Tested out some creation via the UI on staging.

PR Checklist | Your Answer
------------ | -------------

### DIFF
--- a/services/QuillLMS/db/migrate/20240328190319_add_coda_to_evidence_research_gen_ai_llm_prompt_templates.evidence.rb
+++ b/services/QuillLMS/db/migrate/20240328190319_add_coda_to_evidence_research_gen_ai_llm_prompt_templates.evidence.rb
@@ -1,0 +1,12 @@
+# frozen_string_literal: true
+
+# This migration comes from evidence (originally 20240328185342)
+class AddCodaToEvidenceResearchGenAILLMPromptTemplates < ActiveRecord::Migration[7.0]
+  def change
+    add_column :evidence_research_gen_ai_llm_prompt_templates,
+      :coda,
+      :string,
+      null: false,
+      default: Evidence::Research::GenAI::LLMPromptTemplate::FEEDBACK_CODA
+  end
+end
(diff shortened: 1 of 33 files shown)


services/QuillLMS/db/migrate/20240328190319_add_coda_to_evidence_research_gen_ai_llm_prompt_templates.evidence.rb
```diff
@@ -0,0 +1,12 @@
+# frozen_string_literal: true
+
+# This migration comes from evidence (originally 20240328185342)
+class AddCodaToEvidenceResearchGenAILLMPromptTemplates < ActiveRecord::Migration[7.0]
+  def change
+    add_column :evidence_research_gen_ai_llm_prompt_templates,
+      :coda,
```
dandrabik, line 7: I'm assuming you are using `coda` since it's added at the end (or is this an ML term I'm not aware of). We might consider using plainer language. It's not a particularly common word (I only know it from crossword puzzles).

dandrabik (commented):
Logic-wise seems fine. I'd like to chat through to see if there is a more common name than `coda`. I think that might be confusing to some.

services/QuillLMS/db/migrate/20240328190319_add_coda_to_evidence_research_gen_ai_llm_prompt_templates.evidence.rb
brendanshean: It's not an ML word, I borrowed it from linguistics, (I thought of it by pronouncing CoT), but in retrospect, I agree with your point.

PR #11704, opened 2024-04-01
Show custom Max Attempts Feedback for Connect fill in blank questions 

cissyxyu:
## WHAT
When students play through a Connect fill in the blank question and get to 5 attempts without an optimal answer, they should see feedback that tells them Nice Try... and shows the correct answer.

## WHY
Right now students are just getting regular wrong answer feedback, which is confusing because they may not realize that the question is at max attempts and they should move on to the next question.

## HOW
Implement the same logic we currently use for Grammar questions when a student arrives at max attempt #. Show the actual correct answer in the feedback component.

### Screenshots
Screenshot 2024-04-01 at 3 30 49 PM

### Notion Card Links
https://www.notion.so/quill/Implement-max-attempt-feedback-for-Fill-in-the-Blank-Questions-e690d7964aff45f6a9ee4ac160f03678?pvs=4

Played through a grammar question in staging to max attempts to check behavior. Played through a grammar activity with correct answer to confirm the behavior only happens when max attempts is not optimal.

PR Checklist | Your Answer
------------ | -------------


services/QuillLMS/client/app/bundles/Connect/components/studentLessons/fillInBlank.tsx
```diff
@@ -370,7 +389,18 @@ export class PlayFillInTheBlankQuestion extends React.Component<PlayFillInTheBla
     const { previewMode, question } = this.props;
     const { responses, inputErrors } = this.state
 
-    if (inputErrors && _.size(inputErrors) !== 0) {
+    const maxAttemptsSubmitted = question.attempts && question.attempts.length === ALLOWED_ATTEMPTS;
+    const latestAttempt = getLatestAttempt(question.attempts);
+
+    if (maxAttemptsSubmitted && !latestAttempt.response.optimal) {
+      const finalAttemptFeedback = `<b>Good try!</b> Compare your response to the strong response, and then go on to the next question.<br><br><b>Your response</b><br>${latestAttempt.response.text}<br><br><b>A strong response</b><br>${this.correctResponse()}`
```
emilia-friedberg: Since this code duplicates the `question.tsx` file, I think it would be best to extract the `ALLOWED_ATTEMPTS` constant and turn `finalAttemptFeedback` into a function that just takes the student's response and the correct response, and then import them here. That allows us to update fewer places in the event of a copy change.

emilia-friedberg (changes requested):
Approach LGTM, but I think we want to extract this code from both places it's used rather than duplicating it.

happythenewsad (commented):
Please request my review once Emilia's refactor is pushed up, thank you @cissyxyu

anathomical (approved):

services/QuillLMS/client/app/bundles/Connect/components/studentLessons/fillInBlank.tsx
```diff
@@ -78,6 +80,23 @@ export class PlayFillInTheBlankQuestion extends React.Component<PlayFillInTheBla
     }
   }
 
+  correctResponse = () => {
+    const { responses } = this.state
+    const { question } = this.props
+    let text
+    if (Object.keys(responses).length) {
+      const responseArray = hashToCollection(responses).sort((a: Response, b: Response) => b.count - a.count)
+      const correctResponse = responseArray.find((r: Response) => r.optimal)
```
happythenewsad: Variable name shadows the function name

happythenewsad (approved):
Approving, while mentioning 2 optional suggestions.

services/QuillLMS/client/app/bundles/Shared/libs/finalAttemptFeedback.tsx
```diff
@@ -0,0 +1,13 @@
+import React from "react";
```
emilia-friedberg: Looking good but if this is going to return JSX, I would just make it a component rather than a function, and put it in the `../components` folder. Then you can add a test for it too.

emilia-friedberg (approved):

PR #11705, opened 2024-04-01
Fix highlighting cursor bug on draggable TextEditor

cissyxyu:
## WHAT
Fix a bug where, if the page has just loaded and nothing is clicked, highlighting a section of text to edit it would place the cursor in the wrong location.

## WHY
This was causing inconvenience for Staff when trying to edit focus point and incorrect sequence feedback.

## HOW
This bug is happening because the drag and drop feature in SortableList is wrongly being triggered when the user places the cursor on the text area. This is because the first time the page loads, a text area is not being recognized as a ContentEditable section since the active area on first page load is not the text area, but the body of the page itself.

Add a check in `SortableList` to not only check if the activeTarget of the page is `contentEditable`, but also check if the source of the event (the page element that was just clicked) is `contentEditable`. And if that is true, then do not engage the drag-and-drop feature of sortable list.

### Screenshots

### Notion Card Links
https://www.notion.so/quill/Grammar-Highlighting-Typing-Bug-a2f2ffa10f2f429c8ec98fff1e3a5465?pvs=4

Deploy to staging and verify in the Connect, Grammar, and Diagnostic panels that highlighting text and changing it in Focus Points and Incorrect Sequence containers is not causing the cursor to be misplaced. Then, to confirm that we have not broken other functionality, go into the uses of this component on Teacher-facing pages (Active Classes list, custom activity pack page, and activity category filters) and confirm that functionality for teachers on drag-and-drop lists is still working as intended.

PR Checklist | Your Answer
------------ | -------------
Have you added and/or updated tests? |  NO, tested manually.


brendanshean (approved):
Thank you for fixing this!

emilia-friedberg (approved):

eadams17 (approved):

PR #11708, opened 2024-04-01
Fix timeframe_spec bug

brendanshean:
## WHAT
Fix a spec [failure](https://app.circleci.com/pipelines/github/empirical-org/Empirical-Core/28241/workflows/e5c80aec-ac37-45ab-8385-377dbf3834c6) involving timeframes_spec.

## WHY
The code involving timeframes has some edge cases involving February leap years and end of the month calculation.  I think there's an off by one error here for non February, March months that hadn't manifest until April 1 occurred.

## HOW
Change the threshold value from 1 to 2.

### Screenshots

Using timecop, I tested this over different months to make sure prior months tests still work:
 
```
    before do
      travel_to 1.month.from_ago
      allow(DateTime).to receive(:current).and_return(now)
    end

    after { travel_back }
```

PR Checklist | Your Answer
------------ | -------------

### DIFF
--- a/services/QuillLMS/spec/lib/snapshots/timeframes_spec.rb
+++ b/services/QuillLMS/spec/lib/snapshots/timeframes_spec.rb
@@ -108,8 +108,8 @@ module Snapshots
 
     it_behaves_like 'snapshots period length, previous within threshold', 'this-school-year', 1
     it_behaves_like 'snapshots period length, previous within threshold', 'last-school-year', 1
-    it_behaves_like 'snapshots period length, previous within threshold', 'this-month', 1, 3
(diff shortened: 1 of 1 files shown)


anathomical (approved):

dandrabik (commented):
Thanks for jumping in here and fixing the failing spec. There's actually a small bug in the underlying context that I missed (using `today` when it should be `yesterday`, so I think the better fix is to fix that: PR for that here: https://github.com/empirical-org/Empirical-Core/pull/11710 It's failing because the 2 periods being compared are Feb/March and Feb/Jan.

PR #11709, opened 2024-04-01
hardcode activity analysis feedback

emilia-friedberg:
## WHAT
Hardcode activity analysis feedback for Connect, Grammar, and Proofreader sessions. Update how we find activity analysis feedback for suboptimal final Evidence submissions to reflect what we actually show to students.

## WHY
We want to show fifth attempt feedback here.

## HOW
  Add a function that calls other classification-specific functions (to reduce cyclomatic complexity) to get `finalAttemptFeedback` for the various question types. Split off Evidence-specific logic, which is more complicated than the other apps, into its own concern. Update the frontend to use `finalAttemptFeedback` instead of `feedback`, which appears inconsistently across the various apps.

### Screenshots

Evidence Suboptimal
Screenshot 2024-04-04 at 9 27 31 AM

Evidence Suboptimal - Spelling and Grammar
Screenshot 2024-04-04 at 9 26 42 AM

Evidence Optimal
Screenshot 2024-04-04 at 9 26 34 AM

Grammar Suboptimal
Screenshot 2024-04-04 at 9 26 02 AM

Grammar Optimal
Screenshot 2024-04-04 at 9 24 58 AM

Proofreader Suboptimal
Screenshot 2024-04-04 at 9 24 45 AM

Proofreader Optimal
Screenshot 2024-04-04 at 9 24 41 AM

Connect Suboptimal
Screenshot 2024-04-04 at 9 24 15 AM

Connect Optimal
Screenshot 2024-04-04 at 9 24 06 AM

### Notion Card Links
https://www.notion.so/quill/Version-1-5-Part-B-Introduce-the-Activity-Analysis-report-for-students-that-lets-them-access-their-ec0794cc0c464c3990f5515a6fdb1454

Push to staging, check reports for all relevant activity classifications with both optimal and suboptimal final attempts (and the different types of suboptimal, for Evidence), and have Jamie QA as well.

PR Checklist | Your Answer
------------ | -------------


services/QuillLMS/app/models/concerns/public_progress_reports.rb
```diff
@@ -255,11 +264,13 @@ def format_concept_results(activity_session, concept_results)
...
       directfirst = cr.first.concept_result_directions&.text || cr.first.concept_result_instructions&.text || ""
       prompt_text = cr.first.concept_result_prompt&.text
+      score = get_score_for_question(cr)
+      question_uid = cr.first.extra_metadata ? cr.first.extra_metadata['question_uid'] : nil
       hash = {
         directions: directfirst.gsub(/(<([^>]+)>)/i, "").gsub("()", "").gsub("&nbsp;", ""),
         prompt: prompt_text,
         answer: cr.first.answer,
-        score: get_score_for_question(cr),
+        score: score,
```
brendanshean: This can be shortened: `score:`,

services/QuillLMS/app/models/concerns/public_progress_reports.rb
```diff
@@ -270,13 +281,14 @@ def format_concept_results(activity_session, concept_results)
...
             lastFeedback: crs.concept_result_previous_feedback&.text,
+            finalAttemptFeedback: get_final_attempt_feedback(activity_session, question_uid, score, prompt_text, attempt_number),
             attempt: attempt_number || 1,
             answer: crs.answer,
             directions: direct.gsub(/(<([^>]+)>)/i, "").gsub("()", "").gsub("&nbsp;", "")
           }
         },
         question_number: cr.first.question_number,
-        question_uid: cr.first.extra_metadata ? cr.first.extra_metadata['question_uid'] : nil
+        question_uid: question_uid
```
brendanshean: This can be shortened: `question_uid:,`

services/QuillLMS/app/models/concerns/public_progress_reports.rb
```diff
@@ -469,4 +479,24 @@ def generic_questions_for_report(activity)
...
+  private def proofreader_final_attempt_feedback(question, score)
+    if question&.question_type == Question::TYPE_GRAMMAR_QUESTION
+      grammar_final_attempt_feedback(score)
+    else
+      score > 0 ? PROOFREADER_OPTIMAL_FINAL_ATTEMPT_FEEDBACK : PROOFREADER_SUBOPTIMAL_FINAL_ATTEMPT_FEEDBACK
+    end
+  end
+
+  private def connect_final_attempt_feedback(question, score)
+    if score > 0
```
brendanshean:
```
return CONNECT_OPTIMAL_FINAL_ATTEMPT_FEEDBACK if score > 0
return CONNECT_SUBOPTIMAL_FINAL_ATTEMPT_SENTENCE_COMBINING_FEEDBACK if question&.connect_sentence_combining?

CONNECT_SUBOPTIMAL_FINAL_ATTEMPT_FILL_IN_BLANKS_FEEDBACK
```

Then extract a method to question.rb:

`def connect_sentence_combining? = question_type == TYPE_GRAMMAR_QUESTION`

services/QuillLMS/app/models/concerns/evidence_reports.rb
```diff
@@ -0,0 +1,63 @@
...
+
+  def get_evidence_prompt_from_activity_and_prompt_text(activity_session, prompt_text)
+    prompt_ids = Evidence::Activity.find_by(parent_activity_id: activity_session.activity_id)&.prompt_ids
+    Evidence::Prompt.where(id: prompt_ids, text: prompt_text)&.first
+  end
+
+  def get_suboptimal_final_attempt_evidence_feedback_from_activity_session_and_prompt_text(activity_session, prompt_text)
+    feedback_history = get_feedback_history_from_activity_session_prompt_text_and_attempt_number(activity_session, prompt_text, EVIDENCE_FINAL_ATTEMPT_NUMBER)
+
+    if [FeedbackHistory::GRAMMAR, FeedbackHistory::RULES_BASED_THREE, FeedbackHistory::SPELLING].include?(feedback_history&.feedback_type)
```
brendanshean: Let's extract a method into feedback_history.rb:

`def foo_method? = feedback_type.in?[GRAMMAR, RULES_BASED_THREE, SPELLING]`

Then this conditional block becomes
`return EVIDENCE_SUBOPTIMAL_SPELLING_OR_GRAMMAR_FINAL_ATTEMPT_FEEDBACK if feedback_history&.foo_method`

services/QuillLMS/app/models/concerns/evidence_reports.rb
```diff
@@ -0,0 +1,63 @@
...
+
+  def format_max_attempts_feedback(max_attempts_feedback)
+    return nil unless max_attempts_feedback
+
+    # remove HTML, split to take everything before the split-off text, remove leading and trailing whitespace, and make sure any punctuation is followed by a space
+    ActionView::Base.full_sanitizer.sanitize(max_attempts_feedback)
+      .split(EVIDENCE_SUBOPTIMAL_SPLIT_TEXT)
+      .first
+      .strip
+      .gsub(/(\.|,|;|!|\?)(?=\S)/, '\1 ')
```
brendanshean: For clarity, let's add a REGEX constant for the punctuation here.

services/QuillLMS/app/models/concerns/public_progress_reports.rb
```diff
@@ -255,11 +264,13 @@ def format_concept_results(activity_session, concept_results)
       cr.sort!{|x,y| (x.attempt_number || 0) <=> (y.attempt_number || 0)}
       directfirst = cr.first.concept_result_directions&.text || cr.first.concept_result_instructions&.text || ""
       prompt_text = cr.first.concept_result_prompt&.text
+      score = get_score_for_question(cr)
+      question_uid = cr.first.extra_metadata ? cr.first.extra_metadata['question_uid'] : nil
```
brendanshean: This can be shortened with `dig`:  `question_uid = cr.first.extra_metadata&.dig('question_uid')`

services/QuillLMS/app/models/concerns/evidence_reports.rb
```diff
@@ -0,0 +1,63 @@
+# frozen_string_literal: true
+
+module EvidenceReports
+  extend ActiveSupport::Concern
+
+  # this constant is also used on the frontend to generate hardcoded feedback for students to see (see services/QuillLMS/client/app/constants/evidence.ts) and should be updated there if it's updated here.
+  EVIDENCE_SUBOPTIMAL_SPELLING_OR_GRAMMAR_FINAL_ATTEMPT_FEEDBACK = "You completed four revisions! You’ve found the right piece of evidence, but there may still be spelling or grammar changes you could make to improve your sentence. Read your sentence one more time and think about what changes you could make. Then move on to the next prompt."
```
anathomical, line 7: Does it make sense to use the same name for this constant that we use on the front-end?  I recognize that this name is more descriptive, and your comment here may be enough to communicate what's going on, but I could imagine someone not recognizing which two constants need to be kept in sync even with the comment.  Not a huge deal, but figured I'd call it out.

brendanshean (commented):
Some refactoring

services/QuillLMS/app/models/concerns/evidence_reports.rb
```diff
@@ -0,0 +1,63 @@
...
+    return nil if feedback_histories.empty? || prompt_text.blank? || attempt_number.blank?
+
+    prompt = get_evidence_prompt_from_activity_and_prompt_text(activity_session, prompt_text)
+
+    feedback_histories.find {|fh| fh.attempt == attempt_number.to_i && fh.prompt_id == prompt&.id }
+  end
+
+  def get_evidence_prompt_from_activity_and_prompt_text(activity_session, prompt_text)
+    prompt_ids = Evidence::Activity.find_by(parent_activity_id: activity_session.activity_id)&.prompt_ids
+    Evidence::Prompt.where(id: prompt_ids, text: prompt_text)&.first
```
anathomical: `where(...)&.first` is secretly identical to `find_by(...)` in Rails, so you can simplify slightly on that front.  Also, I don't know if it's actually more readable, but you can do this with only one database call instead of two:
```ruby
Evidence::Prompt.joins(:activity)
  .find_by(text: prompt_text, activity: {parent_activity_id: activity_session.activity_id})
```

brendanshean (approved):
Great work on this!  Looks good to go once the refactoring comments are addressed.

anathomical (approved):
Overall, this looks reasonable to me.  One thing I want to call out as a high-level concern is that it feels weird to hard-code a bunch of weird conditional logic here that tries to communicate to teachers what we're showing to their students.  That data really should live on the `ConceptResult` records themselves so that we are recording what students actually see and showing that to teachers.

But that feels like it's a bigger project, and we can get the same effect with this.  I'm just worried about future changes to feedback where we end up having to update this code with some sort of "before this date, show these constants, after this date, show these other constants".

emilia-friedberg:
> Overall, this looks reasonable to me. One thing I want to call out as a high-level concern is that it feels weird to hard-code a bunch of weird conditional logic here that tries to communicate to teachers what we're showing to their students. That data really should live on the `ConceptResult` records themselves so that we are recording what students actually see and showing that to teachers.
> 
> But that feels like it's a bigger project, and we can get the same effect with this. I'm just worried about future changes to feedback where we end up having to update this code with some sort of "before this date, show these constants, after this date, show these other constants".

@anathomical I totally agree with you, and I brought up the concerns at the start of this project. I think the idea is to actually start storing this stuff in concept results over the summer, and the curriculum team has agreed not to change any default feedback til then, but I think the ahistoricity here (even in what's currently present) is weird. Just noting that this was discussed and this was the decision reached by various members of the product and curriculum team.

services/QuillLMS/app/models/concerns/evidence_reports.rb
emilia-friedberg: Good call -- instead of changing this to match the frontend I'm changing the frontend to match this.

PR #11712, opened 2024-04-01
Bug fix/ student metrics activity total time count

eadams17:
## WHAT
tweak student metrics to exclude hidden activity sessions from the total time spent calculation

## WHY
we want the time spent to match the time spent shown on the Activity Summary teacher reports

## HOW
add a where clause to the `completed_sessions` method to include `visible: true`
### Screenshots

### Notion Card Links
https://www.notion.so/quill/Is-a-discrepancy-between-the-Total-time-spent-on-a-student-s-Activity-Summary-report-and-the-Time-sp-bb4b8777e4784ef995d39e1b8b60abf8

hooked into production locally to verify that both time spent counts match for these two users

PR Checklist | Your Answer
------------ | -------------

### DIFF
--- a/services/QuillLMS/app/queries/student_dashboard_metrics.rb
+++ b/services/QuillLMS/app/queries/student_dashboard_metrics.rb
@@ -34,6 +34,7 @@ class StudentDashboardMetrics
         .joins(:classroom_unit)
         .where(user:, classroom_unit: {classroom_id:})
         .where.not(completed_at: nil)
+        .where(visible: true)
     end
   end
 
--- a/services/QuillLMS/spec/queries/student_dashboard_metrics_spec.rb
+++ b/services/QuillLMS/spec/queries/student_dashboard_metrics_spec.rb
@@ -64,7 +64,7 @@ describe StudentDashboardMetrics do
   describe 'completed_sessions' do
     it 'only includes sessions with completed_at not nil and visible true' do
       create(:activity_session, :started, user: student, classroom_unit:)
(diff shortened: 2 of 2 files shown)


services/QuillLMS/app/queries/student_dashboard_metrics.rb
```diff
@@ -34,6 +34,7 @@ def completed_sessions
         .joins(:classroom_unit)
         .where(user:, classroom_unit: {classroom_id:})
         .where.not(completed_at: nil)
+        .where(visible: true)
```
emilia-friedberg, line 37: I think if you just remove `unscoped`, rather than specifying `visible: true`, it'll be automatically scoped to include only visible activities -- I think this is how it works in most places in the codebase now, but I can certainly see an argument that we actually do want to be explicit about this.

emilia-friedberg (approved):
I'm approving, but curious about other's thoughts about explicitly vs implicitly scoping to visible: true.

services/QuillLMS/app/queries/student_dashboard_metrics.rb
brendanshean: Initially, I was going to say get rid of the `unscoped` and remove the `where(visible: true)`.  However, on closer inspection, I noticed `ActivitySession` has another `default_scope` for `joins(:activity)`, so I would opt for being explicit here with the `where(visible: true)`.

brendanshean (approved):

PR #11719, opened 2024-04-02
Add GenAI Experiment Metrics Part 1: Optimal/Suboptimal match

brendanshean:
## WHAT
1.  Add `HasOptimalAndSubOptimal` concern
1.  Add `experiment_id` to `LLMFeedback` migration
1. Add `OptimalAndSubOptimalResultsBuilder` service object 
1. Update experiment#show view to include eval table

## WHY
1.  We'd like a way of determining optimal-suboptimal based on a specific prefix which applies to both `ExampleFeedback` and `LLMFeedback` for use in calculating the ConfusionMatrix
1. `LLMFeedback` belongs_ to `PassagePromptResponse` but we need to add additional experiment_id information to differentiate the feedback from other experiments that use the same `PassagePromptResponse`
1.  We'd like something to calculate the ConfusionMatrix matrix as well as the accuracy for optimal and suboptimal.
1.  It's easy to quickly compare example and llm feedback via color coding.

## HOW
1.  Create a concern that defines `optimal?` and `sub_optimal?` and have `ExampleFeedback` and `LLMFeedback` include it.
1.  Create a migration with ` add_column :evidence_research_gen_ai_llm_feedbacks, :experiment_id, :integer, null: false`. 
1.  After running an experiment, iterate through the LLM Feedback and decide which of the four cases each item belongs to: (optimal, optimal), (optimal, sub-optimal), (sub-optimal, optimal), and (sub-optimal, sub-optimal)
 1. Add table to the erb template.

### Screenshots
Screenshot 2024-04-02 at 11 18 12 AM

### Notion Card Links

Tested on staging.

PR Checklist | Your Answer
------------ | -------------


services/QuillLMS/engines/evidence/app/models/concerns/evidence/research/gen_ai/has_optimal_and_sub_optimal.rb
```diff
@@ -0,0 +1,22 @@
...
+        extend ActiveSupport::Concern
+
+        OPTIMAL_PREFIXES = [
+          'Nice work',
+          'Good job',
+          'Great job',
+          'Excellent job'
+        ].freeze
+
+        def optimal? = OPTIMAL_PREFIXES.any? { |prefix| text.start_with?(prefix) }
```
dandrabik: I think we want to downcase before this comparison on both sides, just to be a little more accurate in case it randomly capitalizes the text, e.g. Nice Work!. We also should strip any leading whitespace from the text before comparison.

dandrabik (approved):
Overall looks good. I added a note about downcasing and stripping whitespace to make the comparison a little more foolproof.

services/QuillLMS/engines/evidence/app/models/concerns/evidence/research/gen_ai/has_optimal_and_sub_optimal.rb
brendanshean: Good idea on downcasing.

PR #11722, opened 2024-04-03
Add GenAI Experiment Metrics Part 2: Identical match

brendanshean:
## WHAT
Add in evaluation for identical matches between llm feedback and corresponding example feedback

## WHY
It's another datapoint for evaluating accuracy of LLM feedback

## HOW
Iterate through the LLM feedback text and note how many responses have the exact same text as the training example.

### Screenshots
Screenshot 2024-04-03 at 12 42 09 PM

### Notion Card Links

I've checked that these new calculations show up on staging experiments.

PR Checklist | Your Answer
------------ | -------------

### DIFF
--- a/services/QuillLMS/engines/evidence/app/models/evidence/research/gen_ai/llm_feedback.rb
+++ b/services/QuillLMS/engines/evidence/app/models/evidence/research/gen_ai/llm_feedback.rb
@@ -29,7 +29,9 @@ module Evidence
 
         delegate :example_optimal?, :example_feedback, to: :passage_prompt_response
 
-        def example_sub_optimal? = !example_optimal?
+        def identical_feedback? = example_feedback.text == text
+
+        def optimal_or_sub_optimal_match? = example_optimal? ? optimal? : sub_optimal?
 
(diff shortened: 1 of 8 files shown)


services/QuillLMS/engines/evidence/app/models/evidence/research/gen_ai/llm_feedback.rb
```diff
@@ -29,7 +29,9 @@ class LLMFeedback < ApplicationRecord
 
         delegate :example_optimal?, :example_feedback, to: :passage_prompt_response
 
-        def example_sub_optimal? = !example_optimal?
+        def identical_feedback? = example_feedback.text == text
```
dandrabik: Note sure if we are doing this anywhere, but we might want to [strip](https://apidock.com/ruby/v2_5_5/String/strip) whitespace from the start and end of **both** strings. Whitespace could cause some false negative matches.

dandrabik (approved):
Overall looks good. I left a note about stripping out whitespace from each of the strings before the comparison to get slightly more accurate results.

PR #11726, opened 2024-04-03
AGDR/ sort Classroom and Teacher aggregate row data for Overview alphabetically

eadams17:
## WHAT
for the Overview tab, sort by name alphabetically when grouping by "teacher" or "classroom"

## WHY
we want this data to be returned alphabetically

## HOW
add an `order` property to the object literals used to aggregate the various APIs so that we can return the records in the order in which they were received from the backend

### Screenshots
Screen Shot 2024-04-03 at 5 56 26 PM

### Notion Card Links
https://www.notion.so/quill/ADGR-Sort-teachers-and-classrooms-alphabetically-when-grouping-by-them-on-the-Performance-Overvie-b025b1a9615945a583bcdb19a0abcef5

tested on staging with various users to ensure that the aggregate row data is being displayed in alphabetically order when grouping by teacher or classroom

PR Checklist | Your Answer
------------ | -------------


services/QuillLMS/client/app/bundles/PremiumHub/containers/diagnosticGrowthReports/helpers.tsx
```diff
@@ -93,8 +96,17 @@ function overallSkillGrowthValue({diagnosticId, overallSkillGrowth, handleGrowth
   return 'No Growth';
 }
 
-function createAggregateRowData({ aggregateRowsDataForDiagnostic, diagnosticId, handleGrowthChipClick, handlePreDiagnosticChipClick }) {
-  return Object.keys(aggregateRowsDataForDiagnostic).map(key => {
+export function createAggregateRowData({ aggregateRowsDataForDiagnostic, diagnosticId, handleGrowthChipClick, handlePreDiagnosticChipClick, groupByValue }) {
+  let keys
+  if (groupByValue === 'teacher' || groupByValue === 'classroom') {
```
anathomical, line 101: I think that the back-end should also have the intended order when sorted by "grade" as well, so you can probably get away with skipping this conditional and just always calling the `sort` function?  I suppose that if grade is already in the correct order that would actually involve more work on the front-end by sorting when it's not fully necessary, but the code would be easier to read without the extra conditional, so I think it's just a judgment call.

anathomical (approved):
I like this as a solution for devolving sort order to the back-end rather than re-writing the sorting algorithms on the front-end.  Good call.

services/QuillLMS/client/app/bundles/PremiumHub/containers/diagnosticGrowthReports/helpers.tsx
emilia-friedberg, line 101: Just wondering if we have constants already for `teacher` or `classroom` since they're being used for business logic here.

emilia-friedberg (approved):

PR #11728, opened 2024-04-04
Lms/diagnostic question optimal concepts

anathomical:
## WHAT
- Create new model to store Optimal Concepts for Diagnostic Questions
- Add ActiveRecord relationships to connect the new model to Questions and to Concepts
- Write a Rake task to:
- - Validate that we can parse all of the data out of a CSV (sourced from a Notion document maintained by Curriculum)
- - Import records into the new table from that CSV

CSV generated from: https://www.notion.so/quill/f937b1cc8d2d4ed8943a36ce35dca177?v=f3ea279a0aac49f2af8589ea8e83ac49
## WHY
- There are a number of Diagnostics which have multiple possible Concepts that can be returned in scoring, but that only "count" a specific Concept (or two) rather than all of them.  So we need a way to store which Concepts "matter" for scoring those questions.
- We want to be able to use all of our ActiveRecord conveniences with the new model
- We want to confirm that a Notion document that was populated by hand for non-automated purposes can be fully parsed so that we can flag any issues that might need to be tweaked by hand to get all the data popluated
- The core goal of this project is to give ourselves access to these Optimal Concepts within the code, so importing this data to the model is really important
## HOW
- Create new model and migration
- Add relationships between the new model and related existing models
- Write a script to run through the entire provided CSV to validate two major things and print out any discrepancies for a human to investiate:
- - Does the Question in the Notion document (i.e. Starter Pre Diagnostic Question 10) match the Question in LMS (so that we can sanity check that we're going to apply the Optimal Concepts to the right question)
- - Does the full list of Optimal Concepts in the Notion document correspond to Concepts in the LMS (to sanity check that we're actually going to mirror the data that Curriculum operates off of without missing anything or adding anything extra)

### Notion Card Links
https://www.notion.so/fcce1690908c4eb6a0abbdecb5d79c7c?pvs=25
https://www.notion.so/quill/RFC-Key-Concepts-for-Diagnostic-Questions-ef54770b1b4847ef85e5a80a12d1565f?d=d6f92002ceb34daf907e0395e8befdf1
https://www.notion.so/quill/2023-2024-School-Year-Diagnostic-Documentation-1fff737635bc4e2780930ba7cca996b8

In local environment and in staging: run migration, run CSV validation task to confirm no data discrepancies, run CSV data import task, validate that all questions in Diagnostic activities have at least one Optimal Concept record.

PR Checklist | Your Answer
------------ | -------------


services/QuillLMS/lib/tasks/diagnostic_question_optimal_concepts.rake
```diff
@@ -0,0 +1,156 @@
...
+        total_count += 1
+      end
+    end
+    puts "Total rows in CSV processed: #{total_count}"
+    puts "Duplicate question + concept data rows: #{duplicate_count}"
+    puts "Records with failures to match: #{missing_concept_count}"
+  end
+
+  desc "Validate the input CSV to ensure that all data that we expect to find in the LMS using the CSV is found.  This will flag cases where there's some kind of mis-match so that we can manually confirm that any discrepancies are just typos, but that the correct data is being found."
+  task validate_csv_data: :environment do
```
happythenewsad: Is there any reason not to call the validate task within the main task?

services/QuillLMS/lib/tasks/diagnostic_question_optimal_concepts.rake
```diff
@@ -0,0 +1,156 @@
...
+          .where.not(parent: {parent_id: nil})
+        concepts = concepts.where(parent: {name: level1}) if level1
+        concepts = concepts.where(grandparent: {name: level2}) if level2
+        concept = concepts.where(name: level0)
+        if concept.length > 1
+          puts "Matched too many Concepts: #{concept_name}"
+        elsif concept.length == 1
+          # A number of questions are used in multiple diagnostics, but should have the same Optimal Concepts for each question.  Since each question has a row in the CSV for every diagnostic it's in, we expect to have about 50 duplicates which we can skip over
+          begin
+            DiagnosticQuestionOptimalConcept.create!(concept_id: concept.first.id, question_uid: question.uid)
```
happythenewsad, line 63: If you use `create_or_update!` here, this script will be more reusable in the future - if Curriculum sends us an edited CSV, we can just run the script again and the updates will persist.

services/QuillLMS/app/models/diagnostic_question_optimal_concept.rb
```diff
@@ -0,0 +1,20 @@
...
+#  question_uid :string           not null
+#  created_at   :datetime         not null
+#  updated_at   :datetime         not null
+#  concept_id   :integer          not null
+#
+# Indexes
+#
+#  unique_diagnostic_question_optimal_concepts_uid_id  (question_uid,concept_id) UNIQUE
+#
+class DiagnosticQuestionOptimalConcept < ApplicationRecord
```
happythenewsad, line 17: Add a `dependent: destroy` directive on Question, or some other directive, to avoid this new table triggering a `ActiveRecord::InvalidForeignKey` error when a Question is deleted (it's an edge case, but might as well anticipate it now).

services/QuillLMS/lib/tasks/diagnostic_question_optimal_concepts.rake
anathomical: Currently the validation doesn't return a "clean" response when doing an automated comparison.  It flags a number of questions as "non-identical" between the CSV and the Notion doc that, when examined manually, are revealed to be minor spacing or wording issues.  We could insist that Curriculum update their Notion doc, but I figured it'd be better to just "sign off" on the discrepancies and move on.  But this means that there's a human step that has to be done when analyzing the validation results.

services/QuillLMS/lib/tasks/diagnostic_question_optimal_concepts.rake
anathomical: I'm not sure that `create_or_update!` is actually different from `create!`.  Because the index is on the combo-key of `concept_id` and `question_uid`, if there's a record with both of those values in a later update, it's just the same value.  There's nothing to update, right?  Either the record with that pair exists or it doesn't.  Maybe I'm missing something, though?

services/QuillLMS/app/models/diagnostic_question_optimal_concept.rb
anathomical: Good call!  I'll update this for both `Question` and `Concept`.

services/QuillLMS/lib/tasks/diagnostic_question_optimal_concepts.rake
happythenewsad: You're right, given that the only data persisted is the data within the index keys.

happythenewsad (approved):
LGTM

services/QuillLMS/app/models/diagnostic_question_optimal_concept.rb
```diff
@@ -0,0 +1,20 @@
...
+  belongs_to :question, class_name: 'Question', foreign_key: :question_uid, primary_key: :uid
```
dandrabik, line 18: I don't think you need `class_name` since that should be the default.

services/QuillLMS/app/models/question.rb
```diff
@@ -50,6 +50,7 @@ class Question < ApplicationRecord
     TYPE_GRAMMAR_QUESTION => 'grammar_questions',
   }
 
+  has_many :diagnostic_question_optimal_concepts, class_name: 'DiagnosticQuestionOptimalConcept', foreign_key: :question_uid, primary_key: :uid, dependent: :destroy
```
dandrabik, line 53: I think this `class_name` can also be dropped.

services/QuillLMS/lib/tasks/diagnostic_question_optimal_concepts.rake
```diff
@@ -0,0 +1,156 @@
...
+    CSV.parse(pipe_data, headers: true).each do |row|
+      activity = diagnostics[row[DIAGNOSTIC_INDEX]]
+      questions = activity.data['questions'].filter {|d| d['questionType'] != 'titleCards' }
+      question_number = row[QUESTION_NUMBER_INDEX].to_i
+      question_uid = questions[question_number - 1]['key']
+      question = Question.find_by(uid: question_uid)
+
+      # Flag any cases where the question from the CSV doesn't exactly match the text of the same question number in the LMS for auditing to ensure that we're actually looking at data for the right question
+      # Note: from the most recent Notion CSV there are 5 questions with discrepancies, but they're all minor typos in Notion that aren't worth dealing with.  Confirmed that all questions in Notion refer to the correct question in LMS.
+      sanitized_question = question.data['prompt']
```
dandrabik: Are these sanitations all supposed to be subtly different?

services/QuillLMS/lib/tasks/diagnostic_question_optimal_concepts.rake
```diff
@@ -0,0 +1,156 @@
...
+    "ELL Intermediate Post" => 1814,
+    "ELL Advanced Pre" => 1590,
+    "ELL Advanced Post" => 1818,
+    "Springboard" => 1432,
+    "PreAP 1" => 1229,
+    "PreAP 2" => 1230,
+    "AP" => 992
+  }
+
+  desc 'Populate diagnostic_question_optimal_concepts from CSV extracted from Curriculum Notion database: https://www.notion.so/quill/f937b1cc8d2d4ed8943a36ce35dca177?v=7dd55ee1f04e4b5daf79645aae960aa1'
```
dandrabik, line 30: I think it's worth doing a refactor pass on these rake tasks. They are very long and complicated, and a bit hard to review. The validation script doesn't share any code with the live script. You can use modules to define sub methods for use in Rake tasks, which might help clean this up a bit.

services/QuillLMS/lib/tasks/diagnostic_question_optimal_concepts.rake
dandrabik: Hmm, I have to agree that it is a little odd that the validation/dry_run step uses entirely different code than the live step. How are we sure it's validating the live code?

services/QuillLMS/lib/tasks/diagnostic_question_optimal_concepts.rake
dandrabik: You should also consider having a `dry_run` or `live` flag` on the main script that you could use for validation purposes (so that your validation uses the same code as your live script).

dandrabik (changes requested):
The data model looks good. I think it's worth taking a small refactor pass at the rake task to ensure that is coded properly.

dandrabik (approved):
:+1: Thanks for cleaning this up. Looks good!

PR #11730, opened 2024-04-06
Compute Gen AI Experiment metrics

brendanshean:
## WHAT
1.  Add new worker queue `gen_ai_eval_worker`
1.  Add num_examples field to `Evidence::Research::GenAI::Experiment` model
1.  Change Experiment#new UI from dropdowns to checkboxes
1.  Extract out evaluations into: `CalculateResultsWorker`
1.  Create ResultsFetcher class
1.  Add STOP_ALL_GEN_AI_EXPERIMENTS environment variable

## WHY
1.  This work is computed on a separate server so it can be decoupled from Experiment Queue.  I'd like to keep it's configuration isolated from other queues since I'm not sure yet what concurrency it can handle.
1.  When running experiments, we might not want to run the entire training set, so this will allow us to persist the selected amount for ease of reference.  This can be computed via (llm_feedbacks.count), but I'd rather just cache it on the model too.
1.  Checkboxes will allow for multiple configurations to be created in tandem and easy comparison across different LLMConfigs
1.  Evaluation calculations are becoming more complex and also computed elsewhere so this should be extracted.
1.  Create a new class to aggregate calculations over the LLM feedback
1.  Multiple experiments can quickly get queued.  If for some reason we don't want to run them, we can add this ENV key to short circuit any remaining ones

## HOW
1.  Update Procfile and Procfile.dev and add config/sidekiq_gen_ai_eval.yml
1.  Add migration file and update ExperimentsController
1.  Replace dropdown elements with checkboxes.
1.  Enqueue the worker within the Experiment#run
1.  Use a similar pattern as gemini/api.rb 
1.  Add a guard clause that checks for this ENV variable setting

### Screenshots
Screenshot 2024-04-07 at 3 52 21 PM

### Notion Card Links

I've tested this on staging

PR Checklist | Your Answer
------------ | -------------


dandrabik (approved):
:+1:

PR #11733, opened 2024-04-10
Expire Concept and Concept Feedback caches on create/update actions

cissyxyu:
## WHAT
Change the 1 hour caching on the Concept and ConceptFeedback indexing endpoints to flexible redis caching that refreshes on record creation/updates. Also show an alert window when ConceptFeedback are created or updated.

## WHY
Admin have to wait one hour right now to see any changes they make on these records which slows down their grading time considerably.

## HOW
Right now we're using the Rails `caches_action` on this endpoint, which is not flexible enough for this use case. Change the caching to redis caching, and manually expire the cache each time a Concept or ConceptFeedback record is created or updated.

### Screenshots
Screenshot 2024-04-10 at 4 42 39 PM
Screenshot 2024-04-10 at 4 41 42 PM

### Notion Card Links
https://www.notion.so/quill/ea241f31a84c4b79ad56e83e99f7ee93?v=a9d41c7da6794f429933c6369ee1e230&p=277b024d38df42df93c83eebe99d81e5&pm=c

Deploy to staging, then test both Concept and ConceptFeedback creation flows that admin typically use:

Create Concept in Concept CMS -> go into Connect CMS and click Concepts -> Check if Concept appears in the list. Try adding the Concept to a Connect question.

Create new ConceptFeedback in Concept CMS. Verify that alert pops up showing conceptFeedback successfully saved. Attempt to add Concept to a new question and verify that ConceptFeedback is saved.

PR Checklist | Your Answer
------------ | -------------

### DIFF
--- a/services/QuillLMS/app/controllers/api/v1/concept_feedback_controller.rb
+++ b/services/QuillLMS/app/controllers/api/v1/concept_feedback_controller.rb
@@ -6,7 +6,7 @@ class Api::V1::ConceptFeedbackController < Api::ApiController
 
   def index
     all_concept_feedbacks = $redis.get(ConceptFeedback::ALL_CONCEPT_FEEDBACKS_KEY)
-    all_concept_feedbacks ||= get_all_concept_feedbacks_and_cache
+    all_concept_feedbacks ||= fetch_all_concept_feedbacks_and_cache
 
     render json: all_concept_feedbacks
   end
@@ -50,7 +50,7 @@ class Api::V1::ConceptFeedbackController < Api::ApiController
     params.require(:concept_feedback).except(:uid)
   end
 
-  private def get_all_concept_feedbacks_and_cache
+  private def fetch_all_concept_feedbacks_and_cache
     all_concept_feedbacks = ConceptFeedback
       .where(activity_type: params[:activity_type])
       .all
--- a/services/QuillLMS/app/controllers/api/v1/concept_feedback_controller.rb
+++ b/services/QuillLMS/app/controllers/api/v1/concept_feedback_controller.rb
@@ -4,12 +4,11 @@ class Api::V1::ConceptFeedbackController < Api::ApiController
   before_action :activity_type, except: [:index]
   before_action :concept_feedback_by_uid, except: [:index, :create, :update]
 
-  caches_action :index, format: 'json', expires_in: 1.hour
+  ALL_CONCEPT_FEEDBACKS_KEY = 'all_concept_feedbacks'
+
   def index
-    all_concept_feedbacks = ConceptFeedback
-      .where(activity_type: params[:activity_type])
-      .all
-      .reduce({}) { |agg, q| agg.update({q.uid => q.as_json}) }
+    all_concept_feedbacks = $redis.get(ALL_CONCEPT_FEEDBACKS_KEY)
+    all_concept_feedbacks ||= get_all_concept_feedbacks_and_cache
 
     render json: all_concept_feedbacks
   end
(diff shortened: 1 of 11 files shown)


services/QuillLMS/app/controllers/api/v1/concept_feedback_controller.rb
```diff
@@ -4,12 +4,9 @@ class Api::V1::ConceptFeedbackController < Api::ApiController
   before_action :activity_type, except: [:index]
   before_action :concept_feedback_by_uid, except: [:index, :create, :update]
 
-  caches_action :index, format: 'json', expires_in: 1.hour
   def index
-    all_concept_feedbacks = ConceptFeedback
-      .where(activity_type: params[:activity_type])
-      .all
-      .reduce({}) { |agg, q| agg.update({q.uid => q.as_json}) }
+    all_concept_feedbacks = $redis.get(ConceptFeedback::ALL_CONCEPT_FEEDBACKS_KEY)
```
brendanshean: When I fetch `ConceptFeedback::ALL_CONCEPT_FEEDBACKS_KEY` I would expect to get all concept feedbacks.  However, the logic filters all concept results by an activity_type.  Do we instead need a different cache key per activity type?

happythenewsad (commented):
1. Notice that, outside of specs, we're setting `$redis.del(Concept::ALL_CONCEPTS_KEY)` 4 different times. I'm concerned that so much caching logic is going to cause difficult-to-diagnose behavior in the future. 

You should be able to leverage `Concept`'s [active record life cycle callbacks](https://guides.rubyonrails.org/active_record_basics.html#callbacks) to collapse this logic into a single line. 

2. Before doing this, note that only 927 Concepts exist. So it's possible that simply removing `caches_action` will achieve your goal, with a reasonable performance hit. It's worth trying, since the best code is no code.

cissyxyu:
> 1. Notice that, outside of specs, we're setting `$redis.del(Concept::ALL_CONCEPTS_KEY)` 4 different times. I'm concerned that so much caching logic is going to cause difficult-to-diagnose behavior in the future.
> 
> You should be able to leverage `Concept`'s [active record life cycle callbacks](https://guides.rubyonrails.org/active_record_basics.html#callbacks) to collapse this logic into a single line.
> 
> 2. Before doing this, note that only 927 Concepts exist. So it's possible that simply removing `caches_action` will achieve your goal, with a reasonable performance hit. It's worth trying, since the best code is no code.

Thanks for the tip! I went and did some testing and it seems like actually this endpoint is fairly fast even without caching. So I changed my approach and instead just removed all of the caching from both endpoints. Much cleaner and easier!

services/QuillLMS/app/controllers/api/v1/concept_feedback_controller.rb
cissyxyu: I've decided to change my approach and get rid of caching on both of these endpoints instead of using redis (see Peter's comment below). So we won't need to use these keys anymore!

brendanshean (approved):

happythenewsad (approved):
🚢

anathomical (commented):
I'm okay with the new approach, but did want to flag that the `concepts#index` action is about 1.25% of our total web traffic, so if this gets significantly slower without caching, this could be a non-trivial performance hit.  It also might not, so I think trying this out is fine, but we'd want to monitor this endpoint to understand how expensive the lack of caching becomes.

(Basically, we turned on caching for a reason, but maybe it isn't that big a deal to turn it off.)
image

PR #11737, opened 2024-04-10
Fix Rails Admin Slowness.

dandrabik:
## WHAT
We’re seeing performance issues in the logs from RailsAdmin that have recently cause site issues for teachers and students. It seems to be misconfigured. This attempts to fix that.
## WHY
We don’t want our admin panel to be slow for our staff, and we especially don’t want using it to accidentally cause issues for students and teachers.
## HOW
The changes are mainly removing some things. Here is the summary:

- Large tables such as `activity_sessions` and `concept_results` have been excluded. Other tables have also been excluded if they had issues or likely aren’t needed.
    - Table Exclusions
        - ActiveActivitySession
        - ActivitySession
        - CanvasConfig
        - ClassroomActivity
        - ClassroomUnit
        - ClassroomUnitActivityState
        - ClassroomsTeacher
        - ConceptResult
        - ConceptResultDirections
        - ConceptResultInstructions
        - ConceptResultPreviousFeedback
        - ConceptResultPrompt
        - ConceptResultQuestionType
        - FeedbackHistory
        - FeedbackHistoryFlag
        - FeedbackHistoryRating
        - FeedbackSession
        - FirebaseApp
        - IpLocation
        - PackSequence
        - PackSequenceItem
        - StudentsClassrooms
        - TeacherSavedActivity
        - ThirdPartyUserId
        - UnitActivity
        - UserActivityClassification
        - UserLogin
        - UserMilestone
        - UserPackSequenceItem
        
- “Have many” relationships were removed from the list view to speed it up
    - e.g. A `unit` can have many `activities` . When we list the `units` we no longer list all `activities` for each `unit` in the list (they were mostly cut off anyway).
        - Click into the info or edit views for an individual record to see these items.
- For large table, show truncated pagination (only show next button, not count of records)
- Some items were hidden in `edit` and `info` views since they were very slow.
### Screenshots

### Notion Card Links
https://www.notion.so/quill/RailsAdmin-Performance-8e0eb00ff2454639b23151ef2f6790b8

I've clicked through a lot of packages locally while connected to the staging DB, and fix slowness by monitoring the logs and editing the config. I'm reaching out to the team for QA for features that they might need (since I removed some).

PR Checklist | Your Answer
------------ | -------------
Have you added and/or updated tests? |  No, this is a config.


services/QuillLMS/app/models/user.rb
```diff
@@ -142,6 +142,8 @@ class User < ApplicationRecord
   has_many :user_subscriptions
   has_many :subscriptions, through: :user_subscriptions
   has_many :activity_sessions
+  has_many :started_activities, through: :activity_sessions, source: :activity
```
dandrabik, line 145: Move this from the student concern since it needs to be defined after activity_sessions when used in user. There's probably more cleanup with redundant associations, but I was trying to be as light touch as possible.

services/QuillLMS/app/models/user.rb
```diff
@@ -194,7 +196,7 @@ class User < ApplicationRecord
     through: :administered_school_canvas_instance_schools,
     source: :canvas_instance
```
dandrabik, line 198: I believe this is just a typo, but it was raising errors in RailsAdmin

services/QuillLMS/config/initializers/rails_admin.rb
```diff
@@ -1,146 +1,432 @@
...
+    end
+  end
 
-    config.model School do
-      list do
-        field :zipcode do
-          searchable true
-        end
-      end
+  config.model 'ActivityClassification' do
```
dandrabik, line 61: Note, I exclude a lot of fields from the `list` view, so I debated extracting this into a config, e.g. 
```ruby
LIST_VIEW_EXCLUSIONS = {
  'ActivityClassification' => [:activities],
  ...
}
```
Or writing some code to [exclude all `has_many` relationships](https://github.com/railsadminteam/rails_admin/issues/1574#issuecomment-1787111163)

But there a good number of models with other configs and some edge_cases in which I want `has_manys` but  `eager_load` them instead. I worried that might be too confusing to mix multiple ways of setting the configs, so I left it verbose. Open to feedback on this.

services/QuillLMS/config/initializers/rails_admin.rb
```diff
@@ -1,146 +1,432 @@
...
 
-    config.excluded_models.push(*MODELS_TO_EXCLUDE)
+  config.model 'UserSubscription' do
+    field :user
+    field :subscription do
+      searchable [:id, :account_type]
+    end
   end
+
+  config.included_models = [
```
dandrabik, line 329: This is an exported list of `ActiveRecord `descendants. I decided to use an `allow list` instead of a `block list`, since I think it's better to be explicit about the models we want to show, so if we add a new large model, it doesn't cause performance issues that surprise us. The downside is we'd have to add models here as they are added. Open to feedback.

Also, while editing this a few times, I found having the full list of everything with the unused commented out we easier for me to parse, but others may not like this. Open to feedback.

services/QuillLMS/app/models/change_log.rb
```diff
@@ -196,10 +196,15 @@ def conjunctions
...
   end
 
+  # TODO: Fix this overwrite of :user association
   def user
     User.find_by(id: user_id)&.name
   end
 
+  def user_name
+    User.find_by(id: user_id)&.name
+  end
```
anathomical, line 206: These are now identical.  Should one be calling the other, is one of them no longer necessary?

services/QuillLMS/config/initializers/rails_admin.rb
anathomical: I'm on board with the verbose version here.  It's also less surprising if someone comes back in two or three years and has to update this.  They can just use RailsAdmin documentation rather than having to figure out the special thing we do with the config.

services/QuillLMS/config/initializers/rails_admin.rb
anathomical: I like the white-list approach over the black-list approach here.  Agree that maintenance may be an ongoing issue, but I half-suspect that even as we add more models, it will turn out that very few of them will need to be accessible through RailsAdmin.  So even if we never add another model to this list, it might turn out not to matter.

I'm okay with the commenting-out thing, but think it may be worth having an explanatory comment at the top of the list so that we don't come back in six months and just delete them.  The fact that it serves as a record that "we made a decision about this model" will also probably make it easier if we do need to come back later and add new models to the list.  The distinction between "this model hasn't been added because we haven't looked at it yet" and "this model is intentionally excluded" seems valuable to me.

anathomical (approved):

services/QuillLMS/app/models/change_log.rb
dandrabik: So the frontend is still using `user` which is a `name` in the `json` generated by the model. I'd have to track down and test that changing `user` -> `user_name` doesn't break anything. Searching for `user` is a little hard in our codebase since it's used so much. I'm instead punting on that and [using this](https://github.com/empirical-org/Empirical-Core/pull/11737/files#diff-d26be7099bbffc96b58f911750f7623d98dcfc7b3aab58db42781b62554a12f4R111-R114) to make RailsAdmin workaround. I'll make a card to fix this separately.

services/QuillLMS/config/initializers/rails_admin.rb
happythenewsad: As long as we maintain this list, I like seeing everything in one place, as you mention

happythenewsad (approved):
Looks good! 

Have you encountered any resources that you were on the fence about, in terms of performance? Wondering if we anticipate follow-on work.

brendanshean (approved):
Thank you for the painstaking effort in cleaning up this configuration.

dandrabik:
> Looks good!
> 
> Have you encountered any resources that you were on the fence about, in terms of performance? Wondering if we anticipate follow-on work.

There were a few, but I used my best judgement. I believe everything here that is live is performant enough not to cause issues, so hopefully this is it for a bit.

The only on-going work might be if I removed something that a staff member uses. I sent this to Partnerships and Curriculum for feedback (a few marked as viewed, but no comments about missing needed things). If we need to add a model back in, we'd just add a config that looks like this

```ruby
config.model 'Model' do
  list do
    exclude_fields model.has_many_relationships
    limited_pagination true
  end
 # test edit and show for problematic/slow fields
end
# exclude model from other model's list pages where needed.

PR #11740, opened 2024-04-11
Add more information to experiment results page

brendanshean:
## WHAT
1. Add LLMPrompts#show view
1. Add `experiment_duration` and `evaluation_duration` values to experiment#show 
1. Add progress and percent complete to experiment#show
1. Add scope `order(:id)` for various has_many associations in experiment.rb
1. Correct the calculation of num_examples
1. Add BertScore columns and add RogueL and BLEU scores
1. Add `Resolver` service object
1. Add some error handling to Resolver
1. Add in Retry functionality for failed experiments

## WHY
1. It's useful to be able to see the actual LLMPrompt sent to the LLM.
1. These values will give transparency on how long these experiments take.
1. It's useful to see progress for the experiment since it regularly seems to idle.
1. When the example feedback is sent for evaluation we would like to have it sent in a consistent order
1. Before experiment records took in `nil` as a value here (since `relation.limit(nil)` will give all records).  Instead lets store the actual computed value since this can be reported in the experiments#index
1. They were previously reported in a large hash
1. Experiments are giving "Response: .... Feedback:....."  We'd like to parse out just the feedback
1. Some Resolver output is blank and we'd like to know what the input feedback and passage prompt response id.
1. Some experiments fail due to timeouts, rate limiting, etc.  Being able to retry them with the click of one button streamlines user error.

## HOW
1.  Add a LLMPromptsController and the corresponding route and template.
1.  Add migration file and update the view with these two values, leveraging `ExperimentsHelper`
1.  Leverage `ExperimentsHelper` to compute these values
1.  Add `-> { order(:id) }` to the relations 
1. Check for nil value when passed from the controller and set num_examples equal to total number of examples.  Otherwise choose the specified value.
1. Update the experiment#show template
1.  Add Resolver processing into the experiment pipeline.
1. Add some error classes and include the relevant passage_prompt_id and feedback.  Also continue the experiment since the remaining examples can still be tried.
1.  Add retry link in the experiments#index.  Only show failed experiments in the index that don't have retried experiments that have completed.

### Screenshots

Percent Progress:
Screenshot 2024-04-11 at 12 13 40 PM

Various metrics
Screenshot 2024-04-11 at 12 14 46 PM

### Notion Card Links

Tested on staging

PR Checklist | Your Answer
------------ | -------------

### DIFF
--- a/services/QuillLMS/config/initializers/zeitwerk.rb
+++ b/services/QuillLMS/config/initializers/zeitwerk.rb
@@ -14,6 +14,7 @@ Rails.autoloaders.each do |autoloader|
     'llm_configs_controller' => 'LLMConfigsController',
     'llm_feedback' => 'LLMFeedback',
     'llm_prompt' => 'LLMPrompt',
+    'llm_prompts_controller' => 'LLMPromptsController',
     'llm_prompt_builder' => 'LLMPromptBuilder',
     'llm_prompt_template' => 'LLMPromptTemplate',
     'llm_prompt_templates_controller' => 'LLMPromptTemplatesController',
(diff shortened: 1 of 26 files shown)


dandrabik (approved):
:+1: LGTM. I like these additional touches

PR #11746, opened 2024-04-12
student activity report

emilia-friedberg:
## WHAT
Backend and frontend of the new student activity report.

Note: this is just the report itself, which is currently orphaned and inaccessible to any users without a direct link. Since the diff was large already, I thought it would be best to split this out on its own.

## WHY
We wanted students to be able to see the same granular information about their sessions that their teachers can view.

## HOW
Created a new endpoint and component for this report, refactoring out logic from the teacher version of this report where possible to share across components.

### Screenshots
Screenshot 2024-04-12 at 11 07 36 AM
Screenshot 2024-04-12 at 11 08 00 AM
Screenshot 2024-04-12 at 11 09 05 AM
Screenshot 2024-04-12 at 11 09 15 AM

### Notion Card Links
https://www.notion.so/quill/Version-1-5-Part-B-Introduce-the-Activity-Analysis-report-for-students-that-lets-them-access-their-ec0794cc0c464c3990f5515a6fdb1454?pvs=4

Looked at sessions on staging for every activity type we're going to use for this report, making sure to view some that belonged to each proficiency category and of a variety of structures. I will have Jamie QA this when it's not an orphaned report, so there may be some additional changes down the line.

PR Checklist | Your Answer
------------ | -------------


services/QuillLMS/client/app/bundles/Shared/libs/findFeedbackForReport.ts
brendanshean, line 1: I found the`let feedback = false` surprising here since it could eventually be assigned as `String`.  What about `null` instead?  Also, the logic here is a little complex and I think refactoring into a couple methods would help with readability:

```
export function findFeedbackForReport(attemptNum, conceptsByAttempt) {
  const currentAttempt = conceptsByAttempt[attemptNum];
  const nextAttempt = conceptsByAttempt[attemptNum + 1];

  return findFeedbackInAttempt(nextAttempt) || currentAttempt[0]?.finalAttemptFeedback;
}

function findFeedbackInAttempt(attempt) {
  if (!attempt) return null;

  return attempt.find(activity => activity.lastFeedback || activity.directions) ?? null;
}
```

brendanshean (approved):
Great work on this!  I really appreciate the thorough testing too.

services/QuillLMS/app/models/concerns/public_progress_reports.rb
```diff
@@ -193,7 +193,7 @@ def results_for_classroom(unit_id, activity_id, classroom_id)
     :not_completed_names
   end
 
-  def formatted_score_obj(final_activity_session, classification, student, average_score_on_quill)
+  def formatted_score_obj(final_activity_session, classification, student, average_score_on_quill=0)
```
anathomical, line 196: I like moving this into a default argument instead of doing the `||` below.

anathomical (approved):

PR #11748, opened 2024-04-15
add results page for diagnostic

emilia-friedberg:
## WHAT
Update the existing results page with a special view for the diagnostic (which shows no scoring information), and then update the ELL and standard diagnostic components to actually send the user there instead of just back to their dashboard.

## WHY
This was a feature request folded into the student activity reports project.

## HOW
Just update the existing component, and pull out the save-to-LMS logic in both Diagnostic components into a shared file so I could just add the Pusher/redirection logic in one place.

### Screenshots
Screenshot 2024-04-15 at 10 32 41 AM

### Notion Card Links
https://www.notion.so/quill/Diagnostic-results-page-e36e32f179be4c65822045aa59c7413c?pvs=4

Played through both a standard and an ELL diagnostic as a student to confirm the redirection and new page work correctly. I'm also going to have Jamie and Jack review it on emilia.quill.org.

PR Checklist | Your Answer
------------ | -------------


services/QuillLMS/client/app/bundles/Diagnostic/utils/saveSession.tsx
```diff
@@ -0,0 +1,83 @@
...
+    {
+      state: 'finished',
+      activity_uid: lessonID,
+      concept_results: results,
+      percentage: score,
+      data
+    },
+    (body) => {
+      initializeSubscription(body.activity_session.uid, isTurkSession, updateState)
+    }
```
anathomical, line 83: I don't know of a better way to do this, but I can imagine some weird edge case where this would result in a race condition.  If the controller from this POST call kicks off a job that somehow finishes and triggers Pusher before the controller returns the UID, the Pusher even will be missed because the subscription won't be in place yet.

Again, not sure there's a better way to do this, but I did want to flag this possibility in case someone thought it was something more than just an unlikely edge case that we don't need to worry about.

services/QuillLMS/client/app/bundles/Diagnostic/utils/saveSession.tsx
```diff
@@ -0,0 +1,83 @@
+import * as React from 'react'
```
anathomical, line 1: I love consolidating this logic into a shared util.  Nice to clean up the duplication.

anathomical (approved):

services/QuillLMS/client/app/bundles/Diagnostic/utils/saveSession.tsx
emilia-friedberg: Yeah, this occurred to me too. This is the same pattern we follow with Connect/Proofreader/Grammar, and as far as I know it hasn't been a problem there, but it is possible in theory. I'm open to changing it if someone has an alternate approach, but I couldn't think of anything since we do need the session uid in order to key the subscription off of.

brendanshean: Since we're already following the same pattern with connect/proofreader/grammar, I'd probably be inclined to keep as-is for now.

If you wanted to guard from the aforementioned race conditions, I think you could create the uid client side first, then initialize pusher with the value, and finally create the activity session with the uid:

```
const createAnonActivitySession = (lessonID, results, score, data, isTurkSession, updateState) => {
  const activitySessionUid = createUID(). // either use external library or make some method
  initializeSubscription(activitySessionUid, isTurkSession, updateState);

  requestPost(
    `${process.env.DEFAULT_URL}/api/v1/activity_sessions/`,
    {
      activity_uid: lessonID,
      concept_results: results,
      percentage: score,
      data,
      activity_session_uid: activitySessionUid  // Include the UUID in the POST data
    },
```

brendanshean (approved):
Nice work on this. I also really appreciate the refactoring with saveSession

services/QuillLMS/client/app/bundles/Diagnostic/utils/saveSession.tsx
emilia-friedberg: Yeah, this is a good point. I'll make a backlog task to pull this logic out further at some point so we can use it across all of the apps, and then maybe it makes sense to add that functionality then so we introduce it everywhere at once.

services/QuillLMS/client/app/bundles/Diagnostic/utils/saveSession.tsx
```diff
@@ -0,0 +1,83 @@
...
+
+  const relevantAnsweredQuestions = playDiagnostic.answeredQuestions.filter(q => q.questionType !== TITLE_CARD_TYPE)
+  const results = getConceptResultsForAllQuestions(relevantAnsweredQuestions);
+  const data = { time_tracking: roundValuesToSeconds(timeTracking), }
+
+  if (sessionID) {
+    finishActivitySession(sessionID, results, 1, data, isTurkSession, updateState);
+  } else {
+    createAnonActivitySession(diagnosticID, results, 1, data, isTurkSession, updateState);
+  }
```
eadams17, line 26: I love all the consolidation! Thank you for that. I think it might be worth adding a comment to explain the need for the `1` argument?

eadams17 (approved):

PR #11749, opened 2024-04-15
Lms/use optimal concepts to score diagnostics

anathomical:
## WHAT
- Update Materialized Views to use `diagnostic_question_optimal_concepts` data when calculating whether or not a student answered a given Diagnostic question "right" or not
- Add query snapshots for all of our Admin Diagnostic queries that didn't have one already
- Add all the removed diagnostic IDs to the activity whitelist so that the data for these activities comes back in reports
- Add the removed diagnostics to the front-end drop-down menus for selecting specific diagnostics
## WHY
- The accurate calculation needed to identify whether or not a student demonstrated a skill on a Diagnostic question was more complicated than we were accounting for for some of the diagnostics.  Updating the calculation allows us to accurately show these Diagnostics in the report.
- This allows us to better identify potential regressions in our queries, and was required for the testing process used for this code
- Now that we can accurately "score" these questions, we should include data for the Diagnostics containing them in the reports
- Now that we support these Diagnostics in reports, for reports that only look at a specific diagnostic we want to allow users to select these diagnostics
## HOW
- Modify the logic that parses `concept_results` records into the materialized view that utilizes data from `diagnostic_question_optimal_concepts` in order to ignore "irrelevant" `concept_results` and count only those that are for Concepts that Curriculum has identified as relevant demonstration of key skills
- Add configuration to the snapshot generating `thor` task that includes all of the queries that we use in the Admin Diagnostic reports, run the task, and then commit the new snapshot files
- Revert the change that removed these IDs from the whitelist when we realized we had this issue
- Revert the change that removed these vlaues from the dropdown when we realized we had this issue, and then add "Pre-AP Writing Skills Survey 2" to the drop-down after realizing it was missing for some reason.

### Screenshots
OLD
image
NEW
image

### Notion Card Links
https://www.notion.so/quill/ADGR-Address-issue-with-diagnostics-that-have-questions-with-multiple-concepts-associated-with-the-d08336d8af1f4e24968c3b9d0a9f3683?pvs=4

Two major approaches:
- Run new snapshot generation on the original code (after adding the expanded diagnostic IDs to the whitelist) in order to get a full view of what our current reports produce
- Create a new materialized view in the namespace `lms_staging` in BigQuery to test modifications on
- Point all queries at the staging view in a branch
- Run snapshot-generating code
- Ensure that the snapshot values DO NOT CHANGE for Starter and Intermediate diagnostics since we've signed off on those as correct and don't want a regression.  (Note, there were changes, but they were rounding differences in the 16th decimal place, so I consider those to be the "same" value.  Additional note: because these snapshots were not previously committed, the diffs are not visible in this PR.  Open to discussion about a good way of making said diffs easily available for review if someone wants to look over them.)

- Because the Teacher reports for Advanced diagnostics have a similar bug as the Admin reports, we can't just compare Admin reports to Teacher reports for validation
- Instead, I did deep dives into all student scoring for a specific classroom (`1448764`), hand-scoring each student's `ConceptResults` records and comparing those scores to what was coming back in the Student report for the updated Admin Diagnostic reports.  (Approximately 30 students.)

PR Checklist | Your Answer
------------ | -------------


services/QuillLMS/db/big_query/views/pre_post_diagnostic_skill_group_performance_view.sql
```diff
@@ -51,9 +51,21 @@ SELECT
...
+            concept_results.question_number,
+            (
+              COUNT(diagnostic_question_optimal_concepts.id) > 0 AND
+              LOGICAL_AND(CASE WHEN diagnostic_question_optimal_concepts.id IS NOT NULL THEN concept_results.correct END)
+            ) AS correct
+          FROM special.concept_results AS concept_results
+          LEFT OUTER JOIN lms.diagnostic_question_optimal_concepts ON STRING(PARSE_JSON(concept_results.extra_metadata).question_uid) = diagnostic_question_optimal_concepts.question_uid AND concept_results.concept_id = diagnostic_question_optimal_concepts.concept_id
+        WHERE concept_results.attempt_number IS NULL
+        GROUP BY concept_results.activity_session_id, concept_results.extra_metadata, concept_results.question_number
+      ) AS concept_results ON activity_sessions.id = concept_results.activity_session_id
```
dandrabik: Take or leave, I found it a little surprising that this subquery is aliased `concept_results` which is an existing table name.

services/QuillLMS/db/big_query/views/pre_post_diagnostic_skill_group_performance_view.sql
```diff
@@ -13,7 +13,7 @@ SELECT
     pre.classroom_unit_id AS pre_classroom_unit_id,
     post.classroom_unit_id AS post_classroom_unit_id,
     IFNULL(pre.activity_name, post.pre_activity_name) AS activity_name,
-    pre.questions_correct AS pre_questions_correct,
+    IFNULL(pre.questions_correct, 0) AS pre_questions_correct,
```
dandrabik: Do any of the other total fields need this `IFNULL`? Why just this one?

dandrabik: If you didn't rename the view, is this new view accidentally currently being used in production (since we don't have a staging env for BigQuery)?

I was going to leave a note to make sure to refresh the view, but I think you might have already overwritten it.

dandrabik (approved):
Overall looks good. I have one concern that the view has already been "deployed" to prod that I think we'll want to talk through in the future.

services/QuillLMS/db/big_query/views/pre_post_diagnostic_skill_group_performance_view.sql
anathomical: This probably isn't obvious in this PR, but I created a new "lms_staging" view to test all of this with, and that code is what I deployed to my test environment.  Once I finished testing, I moved that code to the view in the branch.  So this version of the view is not yet in prod and will need a refresh when it is.  Open to other approaches for safe work around this sort of thing, though.

eadams17 (approved):

services/QuillLMS/db/big_query/views/pre_post_diagnostic_skill_group_performance_view.sql
anathomical: Good call.  I think I defaulted to using the same name to avoid re-working the JOIN structure, but I believe that you're right that we should prioritize clarity and minimize surprise in a query this complicated.

services/QuillLMS/db/big_query/views/pre_post_diagnostic_skill_group_performance_view.sql
anathomical: I think this is a leftover from testing a different solution.  We don't actually need the `IFNULL` here, so I'm removing it.

PR #11751, opened 2024-04-16
move quill-marking-logic into main app

emilia-friedberg:
## WHAT
- Move Quill Marking Logic into the `Shared` bundle in the main app
- Remove module-specific files from that directory (`package.json` etc)
- Rewrite tests to use Jest instead of Karma/Mocha/Chai
- Redirect all references to the package in the app to the local files
- Remove the package from the LMS's package.json
- Update the lambda so it also uses the local path, which required bundling it with vite 

Note: I realize the diff here is massive, but it's almost all path changes/test rewrites. I think it's okay to skim those files, because the tests pass and compilation doesn't break.

## WHY
It's pretty annoying right now to make changes to anything in the QML, because testing it locally is tricky and it's frustrating to have to keep publishing a package and installing it to iterate on a piece of code. This will speed up time and reduce developer frustration in tracking down bugs and making grading changes.

## HOW
See "What"

### Screenshots

### Notion Card Links
https://www.notion.so/quill/Quill-Marking-Logic-Remove-from-package-d688e80aa1e2453e91bfe67c34e43a24?pvs=4

Published to staging and played at least one of each activity type that uses the QML, making sure to include all relevant question types, then looked at the reports for those activities to make sure the correct concept results/algorithms seem to have been applied. Also checked that rematching all and rematching one response still works.

PR Checklist | Your Answer
------------ | -------------


lambdas/rematching/package.json
```diff
@@ -10,10 +10,7 @@
   "license": "ISC",
   "dependencies": {
     "lodash": "^4.17.21",
-    "pg": "^7.7.1",
-    "quill-marking-logic": "0.15.14",
     "request": "^2.88.0",
-    "sequelize": "^4.44.3",
-    "underscore": "^1.12.1"
+    "vite": "^4.2.1"
```
anathomical, line 14: Really appreciate the cleanup on this.

brendanshean (approved):
Great work!  This refactoring is a huge win for developer happiness.  Also, I really appreciate migrating tests to jest.

anathomical (approved):
I'm excited to get more of our tests unified into a more modern framework.

PR #11753, opened 2024-04-17
Add order to new incorrect sequences & button to save incorrect sequence order

cissyxyu:
## WHAT
Incorrect Sequences currently aren't being saved with any order value, so the order will change around and confuse admins. We also save incorrect sequence order on every time the list order is changed, when instead we should save only when a "Save Order" button is pressed.

## WHY
This behavior was causing some bugs on admin end because they thought one order was being saved when in fact the sequences would show up in a different order. The "save on drag" behavior was also causing problems with admin because of race condition with order updates on every drag.

## HOW
Use the same behavior we currently have on focus points where we send incorrect sequences to the API with a designated "order" variable. Move ordering API calls to a button click rather than a callback on sorting.

### Screenshots
Screenshot 2024-04-18 at 12 19 49 AM

### Notion Card Links
[(Please provide links to any relevant Notion card(s) relevant to this PR.)](https://www.notion.so/quill/Incorrect-Sequences-in-Grammar-currently-have-inconsistent-saving-behavior-a2308d4636244e94917cf72b907039d6?pvs=4)

Deploy to staging and create new incorrect sequences on each type of question. Then reorder some questions and click "save" to verify saving behavior. I'm also asking Rachel to do additional QA run on all the apps to double check the behavior.

PR Checklist | Your Answer
------------ | -------------

### DIFF
--- a/services/QuillLMS/app/controllers/api/v1/incorrect_sequences_controller.rb
+++ b/services/QuillLMS/app/controllers/api/v1/incorrect_sequences_controller.rb
@@ -27,9 +27,8 @@ class Api::V1::IncorrectSequencesController < Api::ApiController
   end
 
   def destroy
+    @question.delete_incorrect_sequence(params[:id])
     render json: {}, status: 200
-    # @question.delete_incorrect_sequence(params[:id])
-    # render(plain: 'OK')
   end
 
   def update_all
--- a/services/QuillLMS/app/controllers/api/v1/incorrect_sequences_controller.rb
+++ b/services/QuillLMS/app/controllers/api/v1/incorrect_sequences_controller.rb
@@ -27,8 +27,9 @@ class Api::V1::IncorrectSequencesController < Api::ApiController
   end
 
   def destroy
-    @question.delete_incorrect_sequence(params[:id])
-    render(plain: 'OK')
+    render json: {}, status: 200
+    # @question.delete_incorrect_sequence(params[:id])
+    # render(plain: 'OK')
   end
 
   def update_all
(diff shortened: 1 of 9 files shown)


anathomical (approved):
I'm not super-familiar with this part of the code, but I wonder if, since we're moving to use the pattern we already use for Focus Points we should just build some sort of shared component to unify the code?

emilia-friedberg (commented):
I agree with @anathomical -- this looks good to me but I think it's a really good opportunity to combine all of these files and the respective focus point files into one, since they share so much code already. If that feels like a way bigger project than makes sense to do right now, let me know and I'll review again.

cissyxyu:
> I agree with @anathomical -- this looks good to me but I think it's a really good opportunity to combine all of these files and the respective focus point files into one, since they share so much code already. If that feels like a way bigger project than makes sense to do right now, let me know and I'll review again.

Hi, I tried porting over those files today, and it ended up being more complicated than I anticipated. Basically we wouldn't just have to port over the couple of IncorrectSequence and FocusPoint components in each app -- each of these components imports a bunch of subcomponents -- Response, ConceptSelectorInputs, ConceptFeedback with and without checkboxes, and so forth. And then these subcomponents import a bunch of reducers for concepts and conceptFeedbacks... I'd be happy to investigate how to do this, but I think it would basically be one big project that imports all of our shared components from Connect, Diagnostic, and Grammar over, including response and admin panels. Everything in this part of the code is very intertwined because we reuse a bunch of tiny pieces, e.g. different selectors and inputs.

For this one bug fix, I'd like to just get this functionality in for incorrect sequences so that Rachel can proceed with some admin work she's doing on the Grammar activities. So I think it would make more sense to split up these two PRs if possible? Again, I'm happy to do the refactoring work if split into a separate PR with its own spec!

emilia-friedberg (approved):

emilia-friedberg:
Thanks for the investigation, @cissyxyu!

PR #11755, opened 2024-04-17
Lms/update activity session timestamps in teacher fix

anathomical:
## WHAT
- Find `update_columns` calls throughout our code and make sure that they also set `updated_at` so that the `updated_at` value stays accurate.  (`update_columns` calls directly to SQL, bypassing callbacks and automatic timestamps)
- Update Vitally sync logic to use `completed_at` instead of `updated_at` when determining which school year an ActivitySession counts toward
## WHY
- We want this value to be accurate, but it's also important because we use `updated_at` to keep values up to date in BigQuery, so when the values on the model change without `updated_at` changing, the changes aren't propagated through Airbyte.
- Sometimes an ActivitySession may be updated (like when we merge student accounts), and `updated_at` should change, but we shouldn't change what school year we count the activity toward.  (Note that in 99+% of cases, these two values will be the same, but we do want to handle the edge cases accurately.)
## HOW
- Just make sure to add `updated_at: DateTime.current` to `update_columns` calls throughout the codebase.
- Update query to filter on `completed_at` instead of `updated_at` and update spec conditions to test behavior properly

### Notion Card Links
https://www.notion.so/quill/Admin-Diagnostic-Growth-Report-Not-Counting-All-Completed-Diagnostics-One-School-53371cb83b4b4df2b7a12fa8939bafb6

This can't be tested outside of production, but once this gets merged and we reset the data in BigQuery we should trigger one of these update types to make sure that the data makes it to BigQuery.

PR Checklist | Your Answer
------------ | -------------

### DIFF
--- a/services/QuillLMS/app/models/activity_session.rb
+++ b/services/QuillLMS/app/models/activity_session.rb
@@ -237,17 +237,17 @@ class ActivitySession < ApplicationRecord
 
     # mark all finished anonymous sessions as final score.
     if user.nil?
-      update_columns(is_final_score: true)
+      update_columns(is_final_score: true, updated_at: DateTime.current)
       return
     end
 
     a = ActivitySession.where(classroom_unit: classroom_unit, user: user, is_final_score: true, activity: activity)
                        .where.not(id: id).first
     if a.nil?
-      update_columns is_final_score: true
+      update_columns is_final_score: true, updated_at: DateTime.current
     elsif a.percentage.nil? || percentage >= a.percentage
-      update_columns is_final_score: true
-      a.update_columns is_final_score: false
+      update_columns is_final_score: true, updated_at: DateTime.current
+      a.update_columns is_final_score: false, updated_at: DateTime.current
     end
     # return true otherwise save will be prevented
     true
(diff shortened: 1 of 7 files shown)


brendanshean (approved):
Thank you for cleaning up these date time calculations.

services/QuillLMS/app/models/activity_session.rb
```diff
@@ -237,17 +237,17 @@ def determine_if_final_score
 
     # mark all finished anonymous sessions as final score.
     if user.nil?
-      update_columns(is_final_score: true)
+      update_columns(is_final_score: true, updated_at: DateTime.current)
```
dandrabik, line 240: So it might be better to use something like [`.touch`](https://apidock.com/rails/ActiveRecord/Persistence/touch) here, since that uses the default Rails timestamp setting. I'm not sure if using DateTime.current will be off because of time zones, i.e. our DB timestamps are in UTC. Not sure if this would set this to ET by accident.

dandrabik (changes requested):
I think let's discuss this. I think if we use .touch, this is an approve. I don't want to be off by the timezone by accident.

dandrabik (approved):
:+1:

PR #11759, opened 2024-04-19
Combine HTML-stripping and space-adding logic

anathomical:
## WHAT
Update logic for formatting strings while ensuring spaces after periods
## WHY
- Activities sometimes write values to ConceptResult records without spaces between the end of one sentence and start of the next.  (This looks like it usually happens when it's collapsing two paragraphs into one string.)
- While we generally strip tags from strings before displaying them, we weren't doing it everywhere, and we want to.
## HOW
Chain `formatString` and `formatStringAndAddSpacesAfterPeriods` so that when the latter is called, it also applies the formatting from the former.

### Screenshots
OLD
Screenshot 2024-04-19 at 1 01 01 PM
NEW
Screenshot 2024-04-19 at 1 03 03 PM

### Notion Card Links
https://www.notion.so/quill/Formatting-weirdness-in-Proofreader-teacher-reports-2be885a917794ea09f70f2749659c8b6

Deployed to staging and loaded the report from the original Support ticket to compare formatting between production and new code to confirm that the issues flagged in the original report have been resolved.

PR Checklist | Your Answer
------------ | -------------
Have you added and/or updated tests? | No changes to snapshot expectations

### DIFF
--- a/services/QuillLMS/client/app/bundles/Teacher/components/progress_reports/diagnostic_reports/formatString.ts
+++ b/services/QuillLMS/client/app/bundles/Teacher/components/progress_reports/diagnostic_reports/formatString.ts
@@ -3,7 +3,7 @@ export const formatString = (str: string) => {
 }
 
 export const formatStringAndAddSpacesAfterPeriods = (str: string) => {
-  return stringifiedString(str).replace(/\.(?=[^ ])/g, '. ')
+  return formatString(str).replace(/\.(?=[^ ])/g, '. ')
 }
 
 export const stringifiedString = (str: string) => {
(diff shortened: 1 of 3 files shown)


services/QuillLMS/client/app/bundles/Teacher/components/progress_reports/diagnostic_reports/formatString.ts
```diff
@@ -3,7 +3,7 @@ export const formatString = (str: string) => {
 }
 
 export const formatStringAndAddSpacesAfterPeriods = (str: string) => {
-  return stringifiedString(str).replace(/\.(?=[^ ])/g, '. ')
+  return formatString(str).replace(/\.(?=[^ ])/g, '. ')
```
anathomical, line 6: This seemed the cleanest solution, but open to discussion if we're worried about edge cases where we want the spaces-after-periods logic, but none of the other formatting.

emilia-friedberg (approved):
This looks good to me -- it looks like the only place we currently use that function is for the prompt, where presumably if there *is* HTML it's also going to get rendered literally right now. We can always adjust if we realize we want things to work differently down the line.

PR #11763, opened 2024-04-23
Send optimal response concept result data with new activity creation 

cissyxyu:
## WHAT
In Connect/Grammar, send the concept_results object with the Optimal Response payload so that new questions get created with the correct concept attached.

## WHY
Admin are confused why optimal response questions get created without the concept attached. This is expected behavior whenever they create a new prompt with an optimal response.

## HOW
For connect questions, we just weren't sending the concept_results object with the payload. For Grammar, we were sending it in the wrong place (we should send it after question creation so we can attach the newly created question_uid).

### Screenshots
Screenshot 2024-04-23 at 5 25 13 PM

### Notion Card Links
https://www.notion.so/quill/The-first-optimal-you-assign-when-creating-a-Connect-Grammar-or-FITB-question-displays-as-though-i-b5aef8227414431bb55fd4126f65ac32?pvs=4

Deployed to staging and created new questions for each activity category, then verified that the prompts have concept results attached. Also asked Rachel to test this flow as well for sanity check.

PR Checklist | Your Answer
------------ | -------------
Have you added and/or updated tests? |  No, tested manually

### DIFF
--- a/services/QuillLMS/client/app/bundles/Connect/components/fillInBlank/fillInBlankForm.jsx
+++ b/services/QuillLMS/client/app/bundles/Connect/components/fillInBlank/fillInBlankForm.jsx
@@ -113,7 +113,8 @@ class FillInBlankForm extends Component {
           text: newQuestionOptimalResponse.trim(),
           optimal: true,
           count: 0,
-          feedback: "That's a strong sentence!"
+          feedback: "That's a strong sentence!",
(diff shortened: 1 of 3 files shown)


emilia-friedberg (approved):
LGTM! FYI, Thomas is out for this whole sprint -- I think this PR is small enough to go out with just my approval but for other PRs you might want to tag an additional person.

PR #11764, opened 2024-04-23
add icons and typography to backpack

emilia-friedberg:
## WHAT
Add icons and typography scale from Figma to our variables file (for typography) and quill.org/backpack for reference.

Note: this PR is huge, but it's almost all image files.

## WHY
To standardize design styles and icon imports across the codebase going forward.

## HOW
Add `mixin` definitions for each of the scaled font definitions (since they comprise multiple properties) and show examples of their use in Backpack. For the icons, I organized them all into folders, uploaded them to AWS, and then built a gallery for them in Backpack.

Note: in the past, we've tried adding icons to a `Shared/images` file. I haven't personally found this super useful since we didn't have a view to see each of the icons as they render, and while it saves having to type out the whole HTML element `` each type we use it, there's still the export boilerplate and less flexibility with alt text, and we also have to come up with names for each of the images that are clearly differentiated. I personally feel like just being able to see the gallery in Backpack will be more useful to my workflow, but if other devs feel strongly that we should export these all from a single file let me know and we can discuss.

### Screenshots
Screenshot 2024-04-23 at 9 43 09 AM
Screenshot 2024-04-23 at 9 43 39 AM

### Notion Card Links
https://www.notion.so/quill/Quill-Backpack-Icons-1728728b19c64e4fa3bae9717b8d7384?pvs=4

https://www.notion.so/quill/Quill-Backpack-Typography-Scale-f2399be85738476589c124e0519e2caf?pvs=4

Uploaded to staging and reviewed on the Backpack page to make sure everything renders correctly.

PR Checklist | Your Answer
------------ | -------------
Have you added and/or updated tests? | NO - we can't really test CSS


brendanshean (approved):
Thanks for putting this all in backpack!

PR #11777, opened 2024-04-29
Start caching Concepts endpoint again

cissyxyu:
## WHAT
Start caching the Concepts endpoint again because we determined the endpoint is taking up too much server traffic.

## WHY
We removed this last month because we posited that the ~1000 records being returned would not be too large a load to call uncached. However, this endpoint is called frequently enough that removing the cache did significantly increase server load, so we want to put in caching.

## HOW
Put in Rails.cache caching on Concept and ConceptFeedback indexes.

### Screenshots

### Notion Card Links
https://www.notion.so/quill/Put-Caching-back-in-place-for-Concept-endpoints-c604d9af6dfc48a0b4e6bb2c28f3e70d?pvs=4

Deploy to staging, make sure both endpoint caches are invalidated when a new record is created, and it shows up in the index call right after creation.

PR Checklist | Your Answer
------------ | -------------

### DIFF
--- a/services/QuillLMS/app/models/concept.rb
+++ b/services/QuillLMS/app/models/concept.rb
@@ -107,6 +107,6 @@ class Concept < ApplicationRecord
   end
 
   private def clear_concept_cache
-    Rails.cache.delete("#{ALL_CONCEPTS_KEY}")
+    Rails.cache.delete(ALL_CONCEPTS_KEY.to_s)
   end
 end
--- a/services/QuillLMS/app/models/concept.rb
+++ b/services/QuillLMS/app/models/concept.rb
@@ -107,6 +107,6 @@ class Concept < ApplicationRecord
   end
 
   private def clear_concept_cache
-    Rails.cache.delete(ALL_CONCEPTS_KEY.to_s)
+    Rails.cache.delete(ALL_CONCEPTS_KEY)
   end
 end
--- a/services/QuillLMS/app/models/concept.rb
+++ b/services/QuillLMS/app/models/concept.rb
@@ -106,6 +106,6 @@ class Concept < ApplicationRecord
   end
 
   private def clear_concept_cache
-    Rails.cache.delete("#{ALL_CONCEPTS_KEY}")
+    Rails.cache.delete(ALL_CONCEPTS_KEY.to_s)
   end
 end
--- a/services/QuillLMS/app/models/concept.rb
+++ b/services/QuillLMS/app/models/concept.rb
@@ -107,6 +107,6 @@ class Concept < ApplicationRecord
   end
 
   private def clear_concept_cache
-    Rails.cache.delete(ALL_CONCEPTS_KEY)
+    $redis.del(ALL_CONCEPTS_KEY)
   end
 end
--- a/services/QuillLMS/app/models/concept.rb
+++ b/services/QuillLMS/app/models/concept.rb
@@ -22,6 +22,7 @@ class Concept < ApplicationRecord
   validates :name, presence: true
   has_many :concept_results
   has_many :change_logs, as: :changed_record
+  has_many :diagnostic_question_optimal_concepts, dependent: :destroy
 
   ALL_CONCEPTS_KEY = "all_concepts_with_level"
 
@@ -106,6 +107,6 @@ class Concept < ApplicationRecord
   end
 
   private def clear_concept_cache
-    Rails.cache.delete(ALL_CONCEPTS_KEY.to_s)
+    Rails.cache.delete("#{ALL_CONCEPTS_KEY}")
   end
 end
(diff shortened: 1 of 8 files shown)


services/QuillLMS/app/models/concept.rb
```diff
@@ -101,4 +105,8 @@ def self.visible_level_zero_concept_ids
       SQL
     ).values.flatten
   end
+
+  private def clear_concept_cache
+    Rails.cache.delete(ALL_CONCEPTS_KEY.to_s)
```
brendanshean: I don't think `to_s` here is needed.

services/QuillLMS/app/models/concept_feedback.rb
```diff
@@ -21,17 +21,26 @@ class ConceptFeedback < ApplicationRecord
...
   validates :data, presence: true
   validates :uid, presence: true, uniqueness: { scope: :activity_type }
   validates :activity_type, presence: true, inclusion: {in: TYPES}
   validate :data_must_be_hash
 
+  after_commit :clear_concept_feedbacks_cache
+
   def as_json(options=nil)
     data
   end
```
brendanshean, line 39: The cache key seems to be used in several places, so I would extract it into its own method:

```
def foo = "#{ALL_CONCEPT_FEEDBACKS_KEY}_#{activity_type}"
```

services/QuillLMS/app/models/concept_feedback.rb
```diff
@@ -21,17 +21,26 @@ class ConceptFeedback < ApplicationRecord
...
 
   private def data_must_be_hash
     errors.add(:data, "must be a hash") unless data.is_a?(Hash)
   end
+
+  private def clear_concept_feedbacks_cache
```
brendanshean, line 44: Then you can update:
```
private def clear_concept_feedbacks_cache = Rails.cache.delete(foo)
```

services/QuillLMS/spec/controllers/api/v1/concept_feedback_controller_spec.rb
```diff
@@ -16,6 +16,17 @@
       get :index, params: { activity_type: concept_feedback.activity_type }, as: :json
       expect(JSON.parse(response.body).keys.first).to eq(concept_feedback.uid)
     end
+
+    it 'resets the rails cache for all concept feedbacks if not set already' do
+      Rails.cache.delete("#{ConceptFeedback::ALL_CONCEPT_FEEDBACKS_KEY}_#{concept_feedback.activity_type}")
```
brendanshean: Then you can update:
```
      Rails.cache.delete(concept_feedback.foo)
```

services/QuillLMS/spec/controllers/api/v1/concept_feedback_controller_spec.rb
```diff
@@ -40,6 +51,11 @@
       post :create, params: { activity_type: concept_feedback.activity_type, concept_feedback: data }, as: :json
       expect(ConceptFeedback.count).to eq(pre_create_count + 1)
     end
+
+    it "should expire the redis cache for concept feedbacks with that activity type" do
+      expect(Rails.cache).to receive(:delete).with("#{ConceptFeedback::ALL_CONCEPT_FEEDBACKS_KEY}_#{concept_feedback.activity_type}")
```
brendanshean: Then you can update:
```     
expect(Rails.cache).to receive(:delete).with(concept_feedback.foo)
```

services/QuillLMS/spec/controllers/api/v1/concept_feedback_controller_spec.rb
```diff
@@ -71,5 +87,17 @@
 
       expect(ConceptFeedback.find_by(uid: uid)).to be
     end
+
+    it "should expire the redis cache for concept feedbacks with that activity type" do
+      expect(Rails.cache).to receive(:delete).with("#{ConceptFeedback::ALL_CONCEPT_FEEDBACKS_KEY}_#{concept_feedback.activity_type}")
```
brendanshean: Then you can update:
```
expect(Rails.cache).to receive(:delete).with(concept_feedback.foo)
```

services/QuillLMS/spec/models/concept_feedback_spec.rb
```diff
@@ -58,4 +58,23 @@
       expect(concept_feedback.as_json).to eq(concept_feedback.data)
     end
   end
+
+  describe '#callbacks' do
+    let!(:concept_feedback) { create(:concept_feedback) }
+
+    context 'after update' do
+      it 'calls Rails.cache.delete on concept feedback with activity type' do
+        expect(Rails.cache).to receive(:delete).with("#{ConceptFeedback::ALL_CONCEPT_FEEDBACKS_KEY}_#{concept_feedback.activity_type}")
```
brendanshean: Then you can update:
```
expect(Rails.cache).to receive(:delete).with(concept_feedback.foo)
```

brendanshean (approved):
Some minor refactoring comments but otherwise looks good to go.  Thank you for implementing both of these solutions.

happythenewsad (approved):
Approving, post-Brendan's change requests

PR #11785, opened 2024-04-30
Update Connect and Diagnostic incorrect sequence panels to new format

cissyxyu:
## WHAT
Update the Connect and Diagnostic incorrect sequence panels to use the new behavior: read incorrectSequences as objects instead of reading them as arrays and send "Save" behavior one time only instead of saving each individual sequence separately.

## WHY
This new behavior is what admins prefer and fixes some common bugs seen when using these panels.

## HOW
Update the function calls to use this new format.

### Screenshots
Screenshot 2024-04-30 at 11 11 15 PM

### Notion Card Links
[(Please provide links to any relevant Notion card(s) relevant to this PR.)](https://www.notion.so/quill/Incorrect-Sequences-in-Grammar-currently-have-inconsistent-saving-behavior-a2308d4636244e94917cf72b907039d6?pvs=4)

I deployed to staging and tried creating new sequences, updating them, sorting them, and using the new save all feature in both Connect and Diagnostic panels.

PR Checklist | Your Answer
------------ | -------------
Have you added and/or updated tests? |  No, no tests here.

### DIFF
--- a/services/QuillLMS/client/app/bundles/Connect/components/incorrectSequence/incorrectSequenceContainer.jsx
+++ b/services/QuillLMS/client/app/bundles/Connect/components/incorrectSequence/incorrectSequenceContainer.jsx
@@ -25,17 +25,6 @@ class IncorrectSequencesContainer extends Component {
     this.state = { orderedIds: orderedIds, questionType, actionFile, questionTypeLink, incorrectSequences };
   }
 
-  // componentDidMount() {
-  //   const { actionFile } = this.state
-  //   const { getUsedSequences } = actionFile
-  //   const { dispatch, match } = this.props
-  //   const { params } = match
-  //   const { questionID } = params
-  //   if (getUsedSequences) {
-  //     dispatch(getUsedSequences(questionID))
-  //   }
-  // }
-
   UNSAFE_componentWillReceiveProps(nextProps) {
     const { incorrectSequences } = this.state
     const { match, questions } = nextProps
--- a/services/QuillLMS/client/app/bundles/Connect/components/incorrectSequence/incorrectSequenceContainer.jsx
+++ b/services/QuillLMS/client/app/bundles/Connect/components/incorrectSequence/incorrectSequenceContainer.jsx
@@ -22,7 +22,7 @@ class IncorrectSequencesContainer extends Component {
     const sequencesCollection = hashToCollection(incorrectSequences)
     const orderedIds = sequencesCollection.sort((a, b) => a.order - b.order).map(b => b.key);
 
-    this.state = { orderedIds: orderedIds, questionType, actionFile, questionTypeLink, incorrectSequences };
+    this.state = { orderedIds, questionType, actionFile, questionTypeLink, incorrectSequences };
   }
 
   UNSAFE_componentWillReceiveProps(nextProps) {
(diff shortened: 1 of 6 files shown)


services/QuillLMS/client/app/bundles/Connect/components/incorrectSequence/incorrectSequenceContainer.jsx
```diff
@@ -17,35 +17,24 @@ class IncorrectSequencesContainer extends Component {
...
+    const orderedIds = sequencesCollection.sort((a, b) => a.order - b.order).map(b => b.key);
 
-  componentDidUpdate(previousProps, previousState) {
-    const question = this.props[this.state.questionType].data[this.props.match.params.questionID]
-    const incorrectSequences = this.getSequences(question)
-
-    if (!_.isEqual(previousState.incorrectSequences, incorrectSequences)) {
-      this.setState({ incorrectSequences, orderedIds: null });
-    };
+    this.state = { orderedIds: orderedIds, questionType, actionFile, questionTypeLink, incorrectSequences };
```
emilia-friedberg: Tiny but this can just be `orderedIds, questionType, ...` because the value is the same as the key.

services/QuillLMS/client/app/bundles/Connect/components/incorrectSequence/incorrectSequenceContainer.jsx
```diff
@@ -17,35 +17,24 @@ class IncorrectSequencesContainer extends Component {
...
   }
 
-  getSequenceIndexByKey = (key) => {
-    const { incorrectSequences, } = this.state
+  UNSAFE_componentWillReceiveProps(nextProps) {
```
emilia-friedberg, line 28: It feels like bad practice to add a new `UNSAFE` lifecycle method here. I'm sure this is what we're doing elsewhere, and if we were sharing code it would be one thing, but it's just going to be more work to clean up down the road. I'm okay if we want to ship with this since I know the curriculum team really needs this functionality, but then I think the next card you work on should be a refactor here.

services/QuillLMS/client/app/bundles/Diagnostic/components/incorrectSequence/incorrectSequenceContainer.jsx
```diff
@@ -29,28 +32,16 @@ class IncorrectSequencesContainer extends Component {
...
-    const incorrectSequences = this.getSequences(question)
-
-    if (_.isEqual(previousState.incorrectSequences, incorrectSequences)) return;
-
-    this.setState({ incorrectSequences, orderedIds: null, });
-  }
-
-  getSequenceIndexByKey = (key) => {
-    const { incorrectSequences, } = this.state
+  UNSAFE_componentWillReceiveProps(nextProps) {
```
emilia-friedberg, line 35: Ideally this would be rewritten as well -- I know we already had an `UNSAFE_` function in this file but if you're already doing the work of adding new functionality and having to QA it and all that, it's a good time to clean up stuff like this.

emilia-friedberg (approved):
Approving because the logic itself looks good, but I do think there's cleanup here we should do as an immediate follow-up (to avoid it getting kicked down the road).

eadams17 (approved):
Agree with Emilia's comments, but otherwise looks good to me!

PR #11794, opened 2024-05-01
AGDR/ save user selections for "group by" and "diagnostic type" dropdowns

eadams17:
## WHAT
save user selections for "group by" and "diagnostic type" dropdowns so they are pre-loaded when a user returns to that tab

## WHY
this will be better UX for these reports

## HOW
store values in `admin_report_filter_selection` table, render on load or tab change; update props accordingly

### Screenshots

### Notion Card Links
https://www.notion.so/quill/ADGR-Remember-a-user-s-selections-on-the-Group-by-and-Diagnostic-controls-b065cbd8fb6d443180cbc04011652810

tested on staging with several different users that dropdown options selections were persisted across both:
- changing tabs to another report
- closing the page and opening a new one

PR Checklist | Your Answer
------------ | -------------


services/QuillLMS/client/app/bundles/PremiumHub/containers/diagnosticGrowthReports/index.tsx
```diff
@@ -58,10 +60,29 @@ export const DiagnosticGrowthReportsContainer = ({
...
+    if (!passedData) {
+      getFilterSelections(activeTab)
+    }
+  }, [])
+
+  React.useEffect(() => {
+    if (!passedData) {
+      getFilterSelections(activeTab)
+    }
+  }, [activeTab])
```
emilia-friedberg: You shouldn't need both of these -- the one with `[activeTab]` should run on the initial render as well as when the value changes.

services/QuillLMS/client/app/bundles/PremiumHub/containers/diagnosticGrowthReports/index.tsx
```diff
@@ -78,27 +99,62 @@ export const DiagnosticGrowthReportsContainer = ({
...
+        setSelectedGroupByValue(group_by_value)
+        setSelectedDiagnosticType(diagnostic_type_value)
+      } else {
+        setSelectedGroupByValue(groupByDropdownOptions[0])
+        setSelectedDiagnosticType(diagnosticTypeDropdownOptions[0])
+      }
+    })
+  }
+
+  function saveFilterSelections(tab, successCallback = () => { }) {
```
emilia-friedberg, line 132: NBD but I might define the default `successCallback` as its own function and then call the variable name here instead of putting the whole function definition in the parameter just to keep things clean.

services/QuillLMS/client/app/bundles/PremiumHub/containers/diagnosticGrowthReports/overviewSection.tsx
```diff
@@ -127,7 +125,7 @@ export const OverviewSection = ({
 
   React.useEffect(() => {
     // this is for testing purposes; this value will always be null in a non-testing environment
-    if (!passedData) {
+    if (!passedData && groupByValue) {
```
emilia-friedberg, line 128: Why don't we want to call `getData` if there's a `groupByValue`?

emilia-friedberg (approved):

emilia-friedberg (approved):
LGTM overall! Just two minor comments and a clarifying question.

services/QuillLMS/client/app/bundles/PremiumHub/containers/diagnosticGrowthReports/index.tsx
eadams17: oops-- meant to take that out, thanks for catching!

services/QuillLMS/client/app/bundles/PremiumHub/containers/diagnosticGrowthReports/overviewSection.tsx
eadams17: I'm not sure I follow... do you mean why don't we want to call `getData` if there's **no** `groupByValue`? We need this value for the `aggregateOverviewData` function which is called after `getData`

services/QuillLMS/client/app/bundles/PremiumHub/containers/__tests__/__snapshots__/diagnosticGrowthReports.test.tsx.snap
```diff
@@ -130,7 +130,7 @@ exports[`DiagnosticGrowthReports full state it should render the expected header
       </button>
     </div>
     <div
-      class="dropdown-container input-container inactive has-text  group-by-dropdown searchable "
+      class="dropdown-container input-container inactive   group-by-dropdown searchable "
```
anathomical, line 133: Do we need the extra whitespace between `interactive` and `group-by-dropdown`?

services/QuillLMS/spec/controllers/admin_report_filter_selections_controller_spec.rb
```diff
@@ -76,6 +78,27 @@
...
+        updated_filter_selections = {
+          'group_by_value' => classroom_option,
+          'diagnostic_type_value' => ell_starter_diagnostic_option
+        }
+        updated_attributes = { report: report, filter_selections: updated_filter_selections }
+
+        expect do
+          post :create_or_update, params: { admin_report_filter_selection: updated_attributes }
+        end.to change { admin_report_filter_selection.reload.filter_selections['group_by_value']['value'] }.from('grade').to('classroom')
+          .and change { admin_report_filter_selection.reload.filter_selections['diagnostic_type_value']['value'] }.from('1663').to('1161')
```
anathomical: I'd rather see the variable names than their literal values in the `from` and `to` calls here.  That way we won't break this by accident if we update values somewhere else.

anathomical (approved):

services/QuillLMS/client/app/bundles/PremiumHub/containers/__tests__/__snapshots__/diagnosticGrowthReports.test.tsx.snap
eadams17: I think this is just a weird snapshot syntax thing? if you look at the component, there's no extra space in the class definition

PR #11799, opened 2024-05-05
Match incorrect sequences in sequential order in marking logic

cissyxyu:
## WHAT
Now that all incorrect sequences have an "order" attribute, we want to match incorrect sequences in sequential order from lowest order to highest order, rather than just simply selecting the first incorrect sequence that matches on the response.

## WHY
Order is important when admin create incorrect sequences so we want this to be hard coded rather than depending on javascript objects' inherent ordering.

## HOW
Instead of using _.any to match incorrect sequences, sort the sequences by order, then use a for loop to check each incorrect sequence until we find one that matches on the response.

### Screenshots

### Notion Card Links

Deploy to staging and test using two incorrect sequences that both match a response in Connect, Diagnostic, and Grammar. Make sure that after reordering the incorrect sequences, the first ordered sequence always matches onto the response.

PR Checklist | Your Answer
------------ | -------------

### DIFF
--- a/services/QuillLMS/client/app/bundles/Shared/quill-marking-logic/src/libs/matchers/incorrect_sequence_match.ts
+++ b/services/QuillLMS/client/app/bundles/Shared/quill-marking-logic/src/libs/matchers/incorrect_sequence_match.ts
@@ -11,12 +11,19 @@ export function incorrectSequenceMatchHelper(responseString:string, incorrectSeq
 }
 
 export function incorrectSequenceMatch(responseString: string, incorrectSequences:Array<IncorrectSequence>):IncorrectSequence {
-  return _.find(incorrectSequences, (incSeq) => {
+  const sortedIncorrectSequences = incorrectSequences.sort((a, b) => (a.order || 0) - (b.order || 0));
+
+  for (let i = 0; i < sortedIncorrectSequences.length; i++) {
+    let incSeq = sortedIncorrectSequences[i];
     const options = incSeq.text.split('|||');
     const caseInsensitive = incSeq.caseInsensitive ? incSeq.caseInsensitive : false
     const anyMatches = _.any(options, particle => incorrectSequenceMatchHelper(responseString, particle, caseInsensitive));
-    return anyMatches;
-  });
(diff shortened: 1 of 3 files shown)


services/QuillLMS/client/app/bundles/Shared/quill-marking-logic/src/libs/matchers/incorrect_sequence_match.ts
```diff
@@ -11,12 +11,19 @@ export function incorrectSequenceMatchHelper(responseString:string, incorrectSeq
 }
 
 export function incorrectSequenceMatch(responseString: string, incorrectSequences:Array<IncorrectSequence>):IncorrectSequence {
-  return _.find(incorrectSequences, (incSeq) => {
+  const sortedIncorrectSequences = incorrectSequences.sort((a, b) => (a.order || 0) - (b.order || 0));
+
+  for (let i = 0; i < sortedIncorrectSequences.length; i++) {
```
emilia-friedberg: You can still use underscore's `_.find` here -- [under the hood](https://underscorejs.org/docs/modules/find.html) it is also just doing a `for` loop, so it should always be iterating in order.

emilia-friedberg (approved):
Love the tests and that we'll actually be using this property now! I left a comment but I think you can still use underscore's `_.find` function on the sorted sequences.

eadams17 (approved):
Looks good to me! Appreciate the test coverage

happythenewsad (commented):
We resolved in the Engineering Meeting in January to privilege lodash over underscore going forward: https://www.notion.so/quill/Engineering-Meeting-187fbbd0975c415a8d40804ab17bb870#043faa6ebab146cdae947c4569896356

There's also a work ticket for this: https://www.notion.so/quill/Remove-underscore-and-underscore-string-packages-7dfab273adc942c999e2de805e561788

Furthermore, Emilia's suggestion for using `lodash.find` will result in cleaner code here.

happythenewsad (approved):
Thanks for making that change, Cissy - LGTM!

PR #11800, opened 2024-05-06
Turn on spell check for all TextEditors in Connect/Grammar/Diagnostic 

cissyxyu:
## WHAT
Turn on spell check for the TextEditor component everywhere it appears in Connect, Diagnostic or Grammar admin panels (besides text boxes where students are typing their answer).

## WHY
So admin have an additional spell check step before they make edits to any student-facing text.

## HOW
Simply pass the shouldCheckSpelling flag to the React component when initialized.

### Screenshots
Screenshot 2024-05-06 at 3 52 01 PM
Screenshot 2024-05-06 at 3 51 50 PM
Screenshot 2024-05-06 at 3 51 42 PM
Screenshot 2024-05-06 at 3 51 31 PM
Screenshot 2024-05-06 at 3 51 22 PM
Screenshot 2024-05-06 at 3 51 10 PM

### Notion Card Links
[(Please provide links to any relevant Notion card(s) relevant to this PR.)](https://www.notion.so/quill/Add-spell-check-to-the-Grammar-Diagnostic-and-Connect-CMSs-34aa46374067403184367163e906d3b6?pvs=4)

Deploy to staging, go through the checklist on the Notion card and test all the flows listed, making sure that spell check is activated.

PR Checklist | Your Answer
------------ | -------------
Have you added and/or updated tests? |  manually tested

### DIFF
--- a/services/QuillLMS/client/app/bundles/Connect/components/feedback/feedbackForm.jsx
+++ b/services/QuillLMS/client/app/bundles/Connect/components/feedback/feedbackForm.jsx
@@ -52,6 +52,7 @@ export default class FeedbackForm extends React.Component {
             EditorState={EditorState}
             handleTextChange={this.handleChange.bind(null, part)}
             key={part}
+            shouldCheckSpelling={true}
             text={this.state[part]}
           />)
         ]
--- a/services/QuillLMS/client/app/bundles/Connect/components/feedback/feedbackForm.jsx
+++ b/services/QuillLMS/client/app/bundles/Connect/components/feedback/feedbackForm.jsx
@@ -1,4 +1,3 @@
-import { ContentState, EditorState } from 'draft-js';
 import React from 'react';
 
 import { ConceptExplanation, TextEditor } from '../../../Shared/index';
@@ -48,11 +47,8 @@ export default class FeedbackForm extends React.Component {
         return [
           (<label className="label">{part}</label>),
           (<TextEditor
-            ContentState={ContentState}
-            EditorState={EditorState}
             handleTextChange={this.handleChange.bind(null, part)}
             key={part}
-            shouldCheckSpelling={true}
             text={this.state[part]}
           />)
         ]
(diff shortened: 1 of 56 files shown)


emilia-friedberg (commented):
Are there any places this editor is used where we don't want spellcheck? I'm just wondering if it makes sense to just change the default behavior of the editor so we don't have to remember to apply this prop if we use it in another place.

brendanshean (commented):
It looks like some snapshot tests need to be updated

cissyxyu:
> Are there any places this editor is used where we don't want spellcheck? I'm just wondering if it makes sense to just change the default behavior of the editor so we don't have to remember to apply this prop if we use it in another place.

Yes, we don't want to use spell check in any boxes where the student is typing a response, because we want them to correct spelling in their response based on our feedback.

emilia-friedberg:
> Yes, we don't want to use spell check in any boxes where the student is typing a response, because we want them to correct spelling in their response based on our feedback.

@cissyxyu but isn't that a different text editor? I think we're just using `react-contenteditable` for those, not the shared draftjs editor you're passing this prop to now.

brendanshean:
I'm also curious about boilerplate Emilia mentioned.  Maybe even one step further with abstracting out: 

```
ContentState={ContentState} 
EditorState={EditorState}
```
as it seems to be in all of these examples.

cissyxyu:
> > Yes, we don't want to use spell check in any boxes where the student is typing a response, because we want them to correct spelling in their response based on our feedback.
> 
> @cissyxyu but isn't that a different text editor? I think we're just using `react-contenteditable` for those, not the shared draftjs editor you're passing this prop to now.

I checked it out and you're right -- it looks like the TextEditor we use for student responses is actually a different component. I've updated the TextEditor component itself to have spellcheck, ContentState, and EditorState as inherent attributes rather than passed props, and changed all its references accordingly.

cissyxyu:
> I'm also curious about boilerplate Emilia mentioned. Maybe even one step further with abstracting out:
> 
> ```
> ContentState={ContentState} 
> EditorState={EditorState}
> ```
> 
> as it seems to be in all of these examples.

Good catch -- I've updated the component now to use ContentState, EditorState, and spellCheck inside the component rather than passed props! And I've removed those attributes from the outside references.

brendanshean (approved):
Screenshot 2024-05-13 at 9 29 09 AM !!

Much better!  Nice work on this.

emilia-friedberg (approved):

PR #11805, opened 2024-05-07
Add data_partition to example feedback

brendanshean:
## WHAT
1.  Add `data_partition` field to `ExampleFeedback`
1.  Add passages controller, views, routes
1.  Move data_importer code to automl_data_importer
1.  Add new data_importer code

## WHY
1.  The experiment runner needs to know what examples are for prompt_engineering, fine_tuning and testing.
1.  It's easier to add new passages in than command line
1.  The old data_importer code was for our automl experiments
1.  The format of new passage data is in CSV and requires different parameters

## HOW
1.  Add migration file and model validations for `data_partition` attribute
1.  Use same pattern as passage_prompt, llm_config, etc.
1.  Rename file.
1.  Since the CSV sheets are per passage and conjunction, update data importer to take in passage_prompt_id parameter.  The rest of the data importing is simplified from the automl datasets. 

### Screenshots

### Notion Card Links
https://www.notion.so/quill/Separate-Test-Data-from-Example-data-1d373fd2d8e5461a9a2403438d2a9676?pvs=4

I've tested locally and on staging.

PR Checklist | Your Answer
------------ | -------------

### DIFF
--- a/services/QuillLMS/db/migrate/20240425125302_add_data_partition_to_example_feedbacks.evidence.rb
+++ b/services/QuillLMS/db/migrate/20240425125302_add_data_partition_to_example_feedbacks.evidence.rb
@@ -7,6 +7,6 @@ class AddDataPartitionToExampleFeedbacks < ActiveRecord::Migration[7.0]
       :data_partition,
       :string,
       null: false,
-      default: Evidence::Research::GenAI::ExampleFeedback::TEST_DATA
+      default: Evidence::Research::GenAI::ExampleFeedback::TESTING_DATA
   end
 end
--- a/services/QuillLMS/db/migrate/20240425125302_add_data_partition_to_example_feedbacks.evidence.rb
+++ b/services/QuillLMS/db/migrate/20240425125302_add_data_partition_to_example_feedbacks.evidence.rb
@@ -3,10 +3,6 @@
 # This migration comes from evidence (originally 20240425125151)
 class AddDataPartitionToExampleFeedbacks < ActiveRecord::Migration[7.0]
   def change
-    add_column :evidence_research_gen_ai_example_feedbacks,
-      :data_partition,
-      :string,
-      null: false,
-      default: Evidence::Research::GenAI::ExampleFeedback::TESTING_DATA
+    add_column :evidence_research_gen_ai_example_feedbacks, :data_partition, :string
   end
 end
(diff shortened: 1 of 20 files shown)


services/QuillLMS/db/migrate/20240425125302_add_data_partition_to_example_feedbacks.evidence.rb
```diff
@@ -0,0 +1,12 @@
+# frozen_string_literal: true
+
+# This migration comes from evidence (originally 20240425125151)
+class AddDataPartitionToExampleFeedbacks < ActiveRecord::Migration[7.0]
+  def change
+    add_column :evidence_research_gen_ai_example_feedbacks,
+      :data_partition,
```
dandrabik: I didn't expect this to be called `data_partition` but I also don't think I have a better name (`group`, `bucket`, `data_type`, `tag` are all possibilities, but not sure I like any of them enough to suggest a rename).

services/QuillLMS/db/migrate/20240425125302_add_data_partition_to_example_feedbacks.evidence.rb
```diff
@@ -0,0 +1,12 @@
...
+      :string,
+      null: false,
```
dandrabik: I think `null:false` is probably fine for now, but I could see having datasets in which we don't break the data into groups and just randomly assign (that feels like a separate feature, so we can probably ignore that for now).

dandrabik (approved):
:+1: LGTM. I left two comments that are more food for thought than suggestions for changes. 🚢

services/QuillLMS/db/migrate/20240425125302_add_data_partition_to_example_feedbacks.evidence.rb
brendanshean: I don't love data_partition, but I couldn't think of anything better either.

services/QuillLMS/db/migrate/20240425125302_add_data_partition_to_example_feedbacks.evidence.rb
brendanshean: I was kind of on the fence about this one.  I think I will turn it off for now.

PR #11806, opened 2024-05-07
lms/fix-snapshot-interpolation-bug

anathomical:
## WHAT
Use double-quotes to ensure that string interpolation works
## WHY
We don't want to pass the variable name into SQL, we want the value
## HOW
Replace `'` with `"` for strings that are intended to be interpolated

- Deploy code to staging
- Ensure that Admin Diagnostic reports load
- Modify report filters to explicitly include "No grade set"
- Check Network tab to ensure that "null" is in the "grades" key of the request payload
- Confirm that data loads without error, and that there are no retries in the staging Sidekiq queue

PR Checklist | Your Answer
------------ | -------------
Have you added and/or updated tests? |  No.  This file does not currently have coverage, and this is a small enough fix that it didn't strike me as reasonable to build that infrastructure out for it

### DIFF
--- a/services/QuillLMS/app/queries/snapshots/reporting_session_query.rb
+++ b/services/QuillLMS/app/queries/snapshots/reporting_session_query.rb
@@ -53,7 +53,7 @@ module Snapshots
       <<-SQL
         AND (
           #{reporting_sessions_view.name}.grade IN (#{grades.map { |g| "'#{g}'" }.join(',')})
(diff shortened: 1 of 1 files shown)


dandrabik (approved):
:+1:

happythenewsad (approved):

brendanshean (approved):

PR #11817, opened 2024-05-09
Integrate OpenAI completion with Experiments        

brendanshean:
## WHAT
1.  Rename `Evidence::OpenAI::Completion` to  `Evidence::OpenAI::SentenceCompletion`
1.  Connect open ai completion wrapper to experiments framework
1.  Run experiments over testing data only

## WHY
1.  This api access uses an older endpoint and deals with a specific use case of sentence completions.
1.  Experiments should allow for different models
1.  Fine tuning and prompt engineering example data is for other use cases

## HOW
1.  Rename file and test file.
1.  Update the endpoint and add newer model keys
1.  Iterate over testing data in the `experiment#create_llm_prompt_responses_feedbacks`

### Screenshots

### Notion Card Links

Tested on staging

PR Checklist | Your Answer
------------ | -------------

### DIFF
--- a/services/QuillLMS/engines/evidence/lib/evidence/open_ai/completion.rb
+++ b/services/QuillLMS/engines/evidence/lib/evidence/open_ai/completion.rb
@@ -6,33 +6,39 @@ module Evidence
       include Evidence::OpenAI::Concerns::Api
       include Evidence::OpenAI::Concerns::SentenceResults
 
-      ENDPOINT = '/completions'
+      ENDPOINT = '/chat/completions'
 
-      MAX_TOKENS = 24
+      MAX_TOKENS = 500
 
+      # rubocop:disable Naming::VariableNumber
       MODELS = {
-        ada: 'text-ada-001',
-        babbage: 'text-babbage-001',
-        curie: 'text-curie-001',
-        davinci: 'text-davinci-002',
-        turbo35_instruct: 'gpt-3.5-turbo-instruct'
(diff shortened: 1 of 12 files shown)


services/QuillLMS/engines/evidence/lib/evidence/open_ai/completion.rb
```diff
@@ -6,45 +6,53 @@ class Completion < Evidence::ApplicationService
       include Evidence::OpenAI::Concerns::Api
       include Evidence::OpenAI::Concerns::SentenceResults
```
dandrabik: You'd want to remove the `include Evidence::OpenAI::Concerns::SentenceResults` since that's used for synthetic data and not this.

dandrabik (changes requested):
So this completion class is used by the [synthetic data generators](https://github.com/empirical-org/Empirical-Core/blob/develop/services/QuillLMS/engines/evidence/lib/evidence/synthetic/generators/paraphrase.rb#L41-L45) and has the expectation that the results are parsed into less sentences (note, I should have isolated the code better because, it does two things 1) call the API and 2) parse the result for synthetic data usage).

It might be easiest to leave the existing class/tests and rename it to `SentenceCompletion` and edit the synthetic generator code to use that.  Then use this model code for the generic calls to OpenAI.

dandrabik (approved):
:+1: Thanks for cleaning this up!

PR #11818, opened 2024-05-09
script to deduplicate question uids

emilia-friedberg:
## WHAT
Add a script that:

- identifies pairs of questions with (case-insensitive) identical uids in the LMS database
- identifies, for each pair, which question has fewer responses in the CMS
- create duplicates of all responses for the identified questions in the CMS with a new uid
- creates duplicates of the questions themselves in the LMS, with the new uid we established in the CMS
- replaces the old question key with the new question key for any activities in the LMS
- archives the old questions

## WHY
This issue was causing a weird experience for curriculum team members, who were seeing responses for two different questions on the afflicted questions in Grammar, and ultimately led to some issues where students got stuck because a curriculum team member thought they were just cleaning up old data and deleted optimal answers. 

## HOW
More or less followed the steps in the RFC, with a couple of deviations due to the fact that I didn't realize just how many responses would be impacted if we updated both questions in the duplicated pair (almost 6 million). Instead, I opted to replace only the question and respective responses that had fewer responses in the database -- some of these are still very large but in total this script will create just over 1 million responses. I also decided to create duplicate responses, rather than just update the existing ones, in the interests of being able to go back to using the old ones more easily if something were to go wrong. That does mean that we'll want to go back and delete the old responses at some point after this script has been run, though, because otherwise for the half of the questions we aren't replacing, curriculum team members would still see irrelevant responses (though nothing bad would happen if they deleted them). By creating these duplicates instead of updating the old ones, we also don't need to put all the questions in alpha before we start this process, so we no longer run the risk of interrupting the student experience.

Most of my time working on this was spent trying to figure out how to make the response creation not take a million years. `insert_all` is very useful in this case!

### Screenshots

### Notion Card Links
https://www.notion.so/quill/Deduplicate-Question-UIDs-in-Grammar-aa23d0b06e15447e9c77abd6ede5f71e?pvs=4

Run the script on my local database

PR Checklist | Your Answer
------------ | -------------
Have you deployed to Staging? | NO - saving staging testing for after this has passed code review because it's annoying, though possible, to reverse


scripts/rails_console/deduplicate_question_uids.rb
```diff
@@ -0,0 +1,397 @@
...
+        q['key'] = new_uid if q['key'] == old_question.uid
+      end
+      activity.save!
+    end
+
+    # archive the old question
+    old_question.data['flag'] = Question::FLAG_ARCHIVED
+    old_question.save!
+  rescue => e
+    puts e.message
```
brendanshean, line 394: Not that you're expecting failure, but I would put the the `lower_count_question` payload for this rescue so that you have everything you need to inspect.  (Not sure if `e` would be providing that already).

scripts/rails_console/deduplicate_question_uids.rb
```diff
@@ -0,0 +1,397 @@
...
+start_time = Time.now
+lower_count_questions.each do |question|
+  Response.where(question_uid: question[:uid]).find_in_batches(batch_size: 100) do |batch|
+    begin
+      responses_for_insertion = []
+      batch.each do |response|
+        duplicated_response = response.attributes.slice('author', 'text', 'feedback', 'count', 'first_attempt_count', 'child_count', 'optimal', 'weak', 'concept_results', 'spelling_error')
+
+        duplicated_response['question_uid'] = question[:new_uid]
+
```
brendanshean, line 338: I think you want a temp variable for Time.now so that `created_at` and `updated_at` are exactly the same on creation:

```
  right_now = Time.now
  duplicated_response['created_at'] = right_now
  duplicated_response['updated_at'] =  right_now
```

brendanshean (approved):
Nice work on carefully building out this script for deduplication.

scripts/rails_console/deduplicate_question_uids.rb
```diff
@@ -0,0 +1,397 @@
...
+        # set created and updated at to Time.now because insert_all doesn't autopopulate
+        duplicated_response['created_at'] = Time.now
+        duplicated_response['updated_at'] = Time.now
+
+        responses_for_insertion.push(duplicated_response.symbolize_keys)
+      end
+      Response.insert_all!(responses_for_insertion)
```
dandrabik: The doc for [insert_all!](https://api.rubyonrails.org/classes/ActiveRecord/Persistence/ClassMethods.html#method-i-insert_all-21) says it sets timestamps. Might be worth trying:


```ruby
insert_all!(record_timestamps: true)

File: scripts/rails_console/deduplicate_question_uids.rb [hunk 4]
```diff
@@ -0,0 +1,397 @@
...
+    rescue => e
+      puts "Failed to process response batch for question UID #{question[:uid]}: #{e.message}"
+      next
+    end
+  end
+  puts 'Processed question uid', question[:uid]
+end; puts Time.now - start_time
+
+UpdateElasticsearchWorker.perform_async(start_time.to_s, Time.now.to_s)
```
dandrabik: This worker is a bit flaky since this time-based query times out a lot (we likely want to remove it, I probably should have done that in a previous refactor). I believe you can get the `ids` from the return of `insert_all!` and loop through them and call this instead:
```ruby
UpdateIndividualResponseWorker.perform_async(response.id)

File: scripts/rails_console/deduplicate_question_uids.rb (hunk 2 above)
Comment (anathomical) [reply scripts/rails_console/deduplicate_question_uids.rb:338]: Additional item for consideration (I haven't fully thought this through, so take it with a grain of salt): should we set the `created_at` of the new response to the same value as the original response?  Just so that we maintain the understanding of when this text actually entered our system as a response?  I could see this going either way.  Even if we do copy the `created_at`, I do kind of like the idea of setting `updated_at` to `Time.now`.

File: scripts/rails_console/deduplicate_question_uids.rb [hunk 5]
```diff
@@ -0,0 +1,397 @@
...
+    old_question = Question.find_by_uid(lower_count_question[:uid])
+
+    duplicated_question = old_question.dup
+
+    # update the UID and the data->key with the new UID
+    duplicated_question.uid = new_uid
+    duplicated_question.data['key'] = new_uid
+
+    # set created and updated at to nil so they get autopopulated correctly
+    duplicated_question.created_at = nil
```
anathomical, line 375: Again, I wonder if we should keep the original `created_at`, but I'm okay with using auto-values here.  It probably doesn't matter much either way.

anathomical (approved):
Nice job working through all the moving parts here.  Looks good to me.

scripts/rails_console/deduplicate_question_uids.rb
```diff
@@ -0,0 +1,397 @@
...
+lower_count_questions = duplicate_groups_with_counts.filter { |dg| dg.length > 1 }.map do |duplicate_group|
+  question_to_change = duplicate_group[0][:count] > duplicate_group[1][:count] ? duplicate_group[1] : duplicate_group[0]
+  question_to_change[:new_uid] = SecureRandom.uuid
+  question_to_change
+end
+
+# COPY LOWER COUNT QUESTIONS SOMEWHERE AND SAVE IT
+
+=begin
+lower_count_questions = [{:uid=>"-Jzw0qjJeHLqN6eDQTJk", :count=>349, :new_uid=>"66586fd5-30c5-4359-9ba8-cbfba8b3ec16"},
```
dandrabik, line 36: Take or leave, since this may not be worth the effort (I'm on the fence about it). You could consider hosting this as a csv file and downloading parsing the CSV so you don't have to copy/paste this data in multiple places. Even if the file is just in github, you can download the raw version as a csv. Anyway, take or leave.

```ruby
require 'open-uri'
require 'csv'

csv_data = CSV.readlines(open(download_url), headers: true)

dandrabik (approved):
Overall looks good. Some small comments about using a different job in the CMS and a possible way to simplify the timestamp code.

scripts/rails_console/deduplicate_question_uids.rb
emilia-friedberg: I think this is a good question -- I added it to the agenda for tomorrow's engineering meeting because I think it's worth the team agreeing on a general approach to similar issues.

scripts/rails_console/deduplicate_question_uids.rb
emilia-friedberg: @dandrabik I actually tried that in an earlier iteration of this project, but I was hitting memory limits due to the number of workers enqueued (at least on staging) and when I tried to stagger them, the script itself would time out, so I think we do need the jobs themselves to get enqueued by another worker. I think historically `UpdateElasticsearchWorker` was used to update 24-hour sets of responses, which is a much broader timeframe than what I'd be doing here -- I wonder if that mitigates the risk of timeout, or if I should just go ahead and add another worker to the codebase that can take the array of ids you mention. Let me know what you think.

scripts/rails_console/deduplicate_question_uids.rb
emilia-friedberg: @dandrabik that was a good thought, but unfortunately that's a Rails 7 addition (docs for Rails 6 here: https://apidock.com/rails/v6.0.0/ActiveRecord/Persistence/ClassMethods/insert_all).

PR #11828, opened 2024-05-13
Add raw_text to LLMFeedback and use json for structured output

brendanshean:
## WHAT
* Add `raw_text` attribute to `LLMFeedback` model
* When injecting prompt engineering examples into the LLM prompt, use JSON instead of unstructured text.
* Append the desired JSON format of feedback from the LLM
* Update the resolver to parse JSON strings instead of unstructured text

## WHY
* It's useful for auditing to store the LLM Feedback before it gets transformed in the Resolver
* JSON provides the LLM with structure that can be reused in the output.
* The LLM tends to give different formats for answers, this _should_ help to keep variability restricted to JSON
* This will return just the feedback and no paraphrase / response data.

## HOW
* Write migration and update validations
* Update `ExampleFeedback#response_and_feedback` to use JSON strings
* Update `LLMPrompt#feedback_prompt` to conclude with the specified JSON format
* Use JSON.parse and dig out the desired value.

### Screenshots

### Notion Card Links
https://www.notion.so/quill/Get-rid-of-metadata-in-LLM-Output-c96208f8223e4763ad71c6dd18002bf2?pvs=4

I'v tested this out locally with Gemini 1.0 and 1.5.

PR Checklist | Your Answer
------------ | -------------

### DIFF
--- a/services/QuillLMS/db/migrate/20240513162849_add_raw_text_to_llm_feedback.evidence.rb
+++ b/services/QuillLMS/db/migrate/20240513162849_add_raw_text_to_llm_feedback.evidence.rb
@@ -1,0 +1,20 @@
+# frozen_string_literal: true
+
+# This migration comes from evidence (originally 20240513162557)
+class AddRawTextToLLMFeedback < ActiveRecord::Migration[7.0]
+  def change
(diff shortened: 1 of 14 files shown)


dandrabik (approved):
:+1: Looks good to me!

PR #11842, opened 2024-05-16
Cleanup malformed JSON from LLM feedback

brendanshean:
## WHAT
1.  Add custom config for Gemini 1.5-{Pro, Flash}-latest
1.  Add `parslet` gem to evidence
1.  Add some another case to `Resolver`
1.  Add `MalformedJSONFixer` service object
1.  Add `RawTextPreprocessor` service object

## WHY
1. These LLMs have a JSON format mode
1.  LLM feedback occasionally appears as malformed JSON missing quotes. Regex, being a regular grammar (type 3) tool is inadequate for parsing JSON, as JSON requires a context-free grammar (type 2) parser.  [Parslet](https://kschiess.github.io/parslet/) has a simple DSL to write grammars for parsing and transforming strings. 
1. The LLM JSON comes in different structures, this time, a new one is `{ properties: { feeback: 'Feedback' } }`
1. It will parse and then transform bad JSON
1. Some of the LLM feedback has simple formatting issues like ````json: [feedback]` or ```Feedback: [feedback]``` where the a simple regex will suffice.  Instead of complicating the parsing grammar, handle these separately.

## HOW
1.  Add `{ "generationConfig": { "response_mime_type": "application/json" }` to POST request
1.  Add to `parslet` to Gemfile.  Use the tool to parse and transform the invalid JSON.
1.  Add method for `properties_feedback`
1.  The service object has two main pieces: `JSONParser` and JSONCleaner` both of which have a grammar that specify rules.  While the real allows for almost any codepoint except quote `"` and control characters I simplified the grammar by allowing for alpha numeric plus underscores in the key, and only try to parse JSON objects.

Here's an example:
Consider the string ` "{ key1\":\"value1\",\"key2\":\"value2\"}"`

`JSONParser` is passed this string which is a pseudo-JSON object and it will try to fix the missing quotes.  It will run the rules and parse this object:
```
 {
  :object=> [
    {
      :key=> { :string=>"key1"@2 },
      :val=>. { :string=>"value1"@9}
    }, 
    {
      :key=> { :string=>"key2"@18 }, 
      :val=>  { :string=>"value2"@25} }
    ]
  }
```
The grammar specifies `.as(name)` and those names get used in keys for the parse. 

Once the string is parsed, it is passed to the transformer, `JSONCleaner`.  The keys are used by the transformer grammar to reconstruct the JSON string yielding:
```
"{\"key1\":\"value1\",\"key2\":\"value2\"}"
```

NB:
Before using Parslet and building a grammar, I attempted to repair the strings via error messages using column and line number of the error.  

JSON.parse (doesn't give line, column information at all, simply returns the string)
OJ.load (sometimes gives line, column information but it's not actually where the repair needs to be made)
Parslet.parse (ditto OJ notes)
Parslet docs have a JSON parser [example](https://jmettraux.wordpress.com/2011/05/11/parslet-and-json/
) but this ended up being since it parses _valid_ JSON and what we have is invalid.

### Screenshots

### Notion Card Links

PR Checklist | Your Answer
------------ | -------------


dandrabik (approved):
:+1: This looks good. Nice work on the parser and thanks for walking me through it. Glad you found a very clean solution to a complicated problem! 🚢

PR #11856, opened 2024-05-21
Add project to automl model

brendanshean:
## WHAT
Add a non-null string attribute 'project' to `Evidence::AutomlModel`

## WHY
We are adding another project to our VertexAI configuration and we want to be able to specify which project is associated with a given `AutomlModel`.  The `project` value will serve as a key to the appropriate configuration.

## HOW
Add a migration and backfill existing data

Ran migration on staging and it completed.  Also inspected values to make sure backfill was performed properly.

PR Checklist | Your Answer
------------ | -------------

### DIFF
--- a/services/QuillLMS/db/migrate/20240521201204_add_project_to_evidence_automl_models.evidence.rb
+++ b/services/QuillLMS/db/migrate/20240521201204_add_project_to_evidence_automl_models.evidence.rb
@@ -5,7 +5,7 @@ class AddProjectToEvidenceAutomlModels < ActiveRecord::Migration[7.0]
   def up
     add_column :evidence_automl_models, :project, :string, null: true
 
-    # Backfill the location column with the original location
+    # Backfill the project column with the original location
     Evidence::AutomlModel.update_all(project: 'comprehension')
 
     change_column_null :evidence_automl_models, :project, false
--- a/services/QuillLMS/db/migrate/20240521201204_add_project_to_evidence_automl_models.evidence.rb
+++ b/services/QuillLMS/db/migrate/20240521201204_add_project_to_evidence_automl_models.evidence.rb
@@ -1,0 +1,17 @@
+# frozen_string_literal: true
+
+# This migration comes from evidence (originally 20240521200827)
+class AddProjectToEvidenceAutomlModels < ActiveRecord::Migration[7.0]
+  def up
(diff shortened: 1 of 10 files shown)


services/QuillLMS/db/migrate/20240521201204_add_project_to_evidence_automl_models.evidence.rb
```diff
@@ -0,0 +1,17 @@
...
+
+# This migration comes from evidence (originally 20240521200827)
+class AddProjectToEvidenceAutomlModels < ActiveRecord::Migration[7.0]
+  def up
+    add_column :evidence_automl_models, :project, :string, null: true
+
+    # Backfill the location column with the original location
+    Evidence::AutomlModel.update_all(project: 'comprehension')
+
+    change_column_null :evidence_automl_models, :project, false
```
dandrabik, line 11: Oh this is a nice method for migrations like this. Thanks for introducing me to it.

dandrabik (approved):
Thanks for getting this in place quickly!

services/QuillLMS/db/migrate/20240521201204_add_project_to_evidence_automl_models.evidence.rb
```diff
@@ -0,0 +1,17 @@
+# frozen_string_literal: true
+
+# This migration comes from evidence (originally 20240521200827)
+class AddProjectToEvidenceAutomlModels < ActiveRecord::Migration[7.0]
+  def up
+    add_column :evidence_automl_models, :project, :string, null: true
+
+    # Backfill the location column with the original location
+    Evidence::AutomlModel.update_all(project: 'comprehension')
```
dandrabik, line 9: Note, I think we might want the project-if here, not project name, I.e. comprehension-27654 or whatever it is.

dandrabik (approved):
One small comment about whether to store project name or id

services/QuillLMS/db/migrate/20240521201204_add_project_to_evidence_automl_models.evidence.rb
brendanshean: I'm using project here and then having a project_id lookup in a separate PR.  I think this will be more readable & predictable.

PR #11858, opened 2024-05-22
Update VertexAI client access to per project calls.

brendanshean:
## WHAT
When making requests to the VertexAI, configure locally with each API call based on the project associated with the model.

## WHY
There are `Evidence::AutomlModel` records that exist in different projects.  Accessing associated information requires configuring the call with different credentials based on the project.

## HOW
* Create a `Client` base class and subclasses `ModelClient`,`EndpointClient`, and `PredictionClient` that take in a project name.  This wrapper will act as an isolated, local configuration
* Update `ParamsBuilder` to use the `project` since the `AutomlModelsController` will now need that piece of information when creating new records.
* Update `TextClassifier` to use `project` since calls to the `PredictionClient` should utilize the correct project credentials.
* Update the associated frontend UI to list `model_name,project` instead of the previous dropdown that just listed the name.

NB:  There's also a change to evidence rails_helper where I moved up the loading of Dotenv since the .env.test variables we're loading after the config/initializers and not being found.

### Screenshots
instructions for curriculum

### Notion Card Links

Assigned a new model locally.
Going to do more testing on staging once model training is completed in the new project.

PR Checklist | Your Answer
------------ | -------------


services/QuillLMS/engines/evidence/spec/rails_helper.rb
```diff
@@ -3,6 +3,9 @@
 require 'spec_helper'
 ENV['RAILS_ENV'] ||= 'test'
 
+require 'dotenv'
+Dotenv.load('.env.test')
+
```
brendanshean, line 8: We want the .env.test variables to load before config/initializers does otherwise ENV variables used in those files won't be found.

services/QuillLMS/engines/evidence/app/services/evidence/vertex_ai/client_fetcher.rb
```diff
@@ -0,0 +1,19 @@
+# frozen_string_literal: true
+
+module Evidence
+  module VertexAI
+    class ClientFetcher
```
dandrabik: Take to leave, but you might consider refactoring this service-object-structured code to a base `Client` class and some subclasses, i.e. `EndpointClient`, `PredictionClient`, etc. so you can organize all of the appropriate code as instance methods on the subclasses, with `credentials`, `project_id`,  and `connection` in the base class. Mentioning because need a `client_class` on each initial setup, you are currently passing around `project` a lot into class methods, and the api-endpoint-specific implementation is spread throughout the other services, but has functions that aren't universal to any `ClientFetcher`, e.g. `list_endpoints`, `list_models`, etc. 

I think you might be able to define the client once per service and have simple methods on that ,e.g. `client`, `client.model_names_and_endpoints`, etc.

dandrabik (approved):
Overall this code seems reasonable, thanks for putting this together. There is a refactor (from service-objects to classes/subclasses) that I think you should consider that I detail in a comment. I'm happy to jump on a call to discuss, but it is ultimately your decision.

services/QuillLMS/engines/evidence/app/services/evidence/vertex_ai/client_fetcher.rb
brendanshean: Good idea.  I've refactored the clients using that pattern.

dandrabik (approved):
Nice cleanup, this looks good!

PR #11862, opened 2024-05-23
Fix delete concept feedback feature, add new default instruction

cissyxyu:
## WHAT
The delete concept feedback feature wasn't working as expected. Also added a new "fill in the blank to..." default instruction for Connect Fill in Blank questions.

## WHY
Admin want these features to be working as expected to smoothly edit activities.

## HOW
Add a popup confirmation wrapper around deleting concept feedback. Fix the back end so it returns an empty json object instead of "OK" because the string "OK" can't be properly read by the response parser. Also just add the desired string constant to fill in blank directions.

### Screenshots
Screenshot 2024-05-23 at 10 07 27 PM
Screenshot 2024-05-23 at 8 21 01 PM

### Notion Card Links
https://www.notion.so/quill/Add-validation-to-Delete-Concept-Feedback-button-b64e2cfcee634191ae025d0b9f1fe410?pvs=4
https://www.notion.so/quill/Add-Fill-in-the-blank-with-the-correct-option-to-the-default-FITB-instructions-drop-down-options-0e30b69e943848888d9ab5a525459a38?pvs=4

Deploy to staging and test out the features on all three pertinent apps.

PR Checklist | Your Answer
------------ | -------------
Have you added and/or updated tests? |  NO, small changes

### DIFF
--- a/services/QuillLMS/client/app/bundles/Diagnostic/components/feedback/concept-feedback.jsx
+++ b/services/QuillLMS/client/app/bundles/Diagnostic/components/feedback/concept-feedback.jsx
@@ -10,7 +10,9 @@ class ConceptFeedback extends React.Component {
     const { dispatch, match } = this.props;
     const { params } = match;
     const { conceptFeedbackID } = params;
-    dispatch(actions.deleteConceptsFeedback(conceptFeedbackID))
+    if (confirm('⚠️ Are you sure you’d like to delete this concept feedback?')) {
+      dispatch(actions.deleteConceptsFeedback(conceptFeedbackID, () => window.location.reload()))
+    }
   };
 
   toggleEdit = () => {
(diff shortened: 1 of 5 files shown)


services/QuillLMS/client/app/bundles/Diagnostic/components/feedback/concept-feedback.jsx
```diff
@@ -10,7 +10,9 @@ class ConceptFeedback extends React.Component {
     const { dispatch, match } = this.props;
     const { params } = match;
     const { conceptFeedbackID } = params;
-    dispatch(actions.deleteConceptsFeedback(conceptFeedbackID))
+    if (confirm('⚠️ Are you sure you’d like to delete this concept feedback?')) {
+      dispatch(actions.deleteConceptsFeedback(conceptFeedbackID, () => window.location.reload()))
```
anathomical: Is there a reason that we're doing a `reload` callback for Diagnostic, but not Connect or Grammar?  Not a problem or anything, but I wanted to understand the discrepancy.

anathomical (approved):
One question for clarification, but this looks fine to me.

services/QuillLMS/client/app/bundles/Diagnostic/components/feedback/concept-feedback.jsx
cissyxyu: whoops, this is outdated (from an earlier implementation that I decided not to go with). I removed this!

PR #11870, opened 2024-05-29
Lms/admin diagnostic csv download

anathomical:
## WHAT
- Update Admin Diagnostic Report UI to add Download buttons (including one on the Students tab when a user has "loaded more" enough times to show 500 students, the current cap on that page)
- Add a new API controller action that allows a user to request download links for CSVs of their report
- Add a worker to extract the user's saved filters so that the CSV filters the same way that the user's web reports do
- Add a worker to run all the Admin Diagnostic queries, roll up the data, put it into CSVs, stash those CSVs in S3, and then send an email to the user with links to those CSVs
- Add adapters that can convert the results of Admin Diagnostic queries into CSVs
- Add a Mailer to assemble the email with links to the CSVs to be sent to requesting users
## WHY
We want admins to be able to take their Diagnostic data and plug it into their own data analysis tools for maximum flexibility.  We also want to allow users to get access to Student data for more than 500 students, which is the hard limit in the web-interface.  So giving them access to a CSV with all filtered student data should help with that.
## HOW
- Styling for the front-end was already in place and just commented out until we were ready, so I uncommented it.  I also used the Data Export front-end code as a template to add new functions to handle download requests.
- Add a simple action that takes the `current_user.id` and enqueues the worker to send the email report to that user
- Add an Enqueue worker that takes a `user_id` and looks up that user's saved filters so that we can build CSVs that match those selected filters.  This worker also provided default filter values in cases where there are no saved values.
- Add a Send worker that takes the filter data from the Enqueue worker, uses it to run all the various Admin Diagnostic queries and combine the data in cases where a single report uses multiple API endpoints.  Then feed that data through the CSV data adapters to get CSV versions of the report, upload those CSVs to S3, and then send out an email to the user with links to those CSVs
- Add query adapters to convert each Admin Diagnostic tab (Overview, Skills, and Students) into CSVs
- Add a Mailer that takes links to the CSVs and provides them to the requesting user

### Notion Card Links
https://www.notion.so/quill/CSV-layout-for-Admin-Diagnostic-Growth-Report-384848b258eb41f3979e24c5a76f52b4#09a27a1d9c7d480aa0485ba770773fdf

- On staging, with two different users (ids `4226515` and `5785736`), test that when a user requests a Download, that a CSV is actually generated and that it matches the web version of the reports under the following conditions:
- - No filters selected
- - Filtering by grade, teacher, and classroom
- - Selecting non-default "Group by" values
- - Selecting different "Group by" values on the Overview page vs the Skills page
- - Selecting non-default "Diagnostic" values
- - Selecting different "Diagnostic" values on the Skills page vs the Students page
- Confirm that when a user is on the Students tab and has clicked "load more" until 500 students are showing, that the helper link to download a CSV report for more data appears, and that clicking it successfully downloads the data.
- Confirm that the number of records in the Student CSV has a number of records equal to the total number of students relevant to the selected filters (rather than maxing out at 500 like the web interface does)

PR Checklist | Your Answer
------------ | -------------


services/QuillLMS/app/workers/admin_diagnostic_reports/enqueue_csv_email_worker.rb
```diff
@@ -0,0 +1,70 @@
...
+    BASE_REPORT_NAME = 'diagnostic_growth_report'
+    OVERVIEW_REPORT_NAME = 'diagnostic_growth_report_overview'
+    SKILL_REPORT_NAME = 'diagnostic_growth_report_skill'
+    STUDENT_REPORT_NAME = 'diagnostic_growth_report_student'
+
+    DEFAULT_TIMEFRAME = 'this-school-year'
+    DEFAULT_AGGREGATION = 'grade'
+    DEFAULT_DIAGNOSTIC_ID = 1663
+
+    def perform(user_id)
```
happythenewsad: I haven't finished reviewing, but my initial question is: why break the async work into two workers? I get that both workers are doing a lot of stuff, but that complexity can be managed with plain Ruby classes, rather than multiple workers. 

Related: If we're keeping two workers, the name should be more expansive. As in,  `PreprocessFiltersAndEnqueue*` vs. `Enqueue*`

services/QuillLMS/app/queries/admin_diagnostic_reports/diagnostic_performance_by_student_view_query.rb
```diff
@@ -9,10 +9,11 @@ class InvalidDiagnosticIdError < StandardError; end
     AGGREGATE_COLUMN = :student_id
     MAX_STUDENTS_TO_RETURN = 500
 
-    def initialize(diagnostic_id:, **options)
+    def initialize(diagnostic_id:, limited: true, **options)
       raise InvalidDiagnosticIdError, "#{diagnostic_id} is not a valid diagnostic_id value." unless DIAGNOSTIC_ORDER_BY_ID.include?(diagnostic_id.to_i)
 
       @diagnostic_id = diagnostic_id
+      @limited = limited
```
dandrabik, line 16: You should add an `attr_reader :limited` so you can reference it without the `@`. It's added protection against accidentally changing the var.

services/QuillLMS/app/queries/admin_diagnostic_reports/diagnostic_performance_by_student_view_query.rb
```diff
@@ -81,7 +82,7 @@ def group_by_clause
     end
 
     def order_by_clause = "ORDER BY TRIM(SUBSTR(TRIM(student_name), STRPOS(student_name, ' ') + 1)), student_name, student_id, skill_group_name"
-    def limit_clause = " LIMIT 5000"
+    def limit_clause = " LIMIT 5000" if @limited
```
dandrabik: e.g. `if limited`

services/QuillLMS/app/controllers/admin_diagnostic_reports_controller.rb
```diff
@@ -3,17 +3,23 @@
...
+  before_action :set_query, only: [:report]
+  before_action :validate_request, only: [:report]
+  before_action :authorize_request, only: [:report]
 
   def report
     render json: retrieve_cache_or_enqueue_worker(WORKERS_FOR_ACTIONS[action_name])
   end
 
+  def download
+    WORKERS_FOR_ACTIONS[action_name].perform_async(current_user.id)
```
dandrabik: This might just be better as: ` AdminDiagnosticReports::EnqueueCsvEmailWorker.perform_async(current_user.id)`. I'm not sure much is gained by this abstraction since each endpoint is custom code.

services/QuillLMS/app/queries/adapters/csv/admin_diagnostic_overview_data_export.rb
```diff
@@ -0,0 +1,73 @@
...
+
+require 'csv'
+
+module Adapters
+  module Csv
+    class AdminDiagnosticOverviewDataExport < CsvDataExport
+      class UnhandledColumnError < StandardError; end
+      class BigQueryResultMissingRequestedColumnError < StandardError; end
+
+      def self.ordered_columns
```
dandrabik: Is there are reason these are all class variables? This would fit well into a Service Object pattern, e.g. You could create an `ApplicationService` that returns a csv. The `to_csv_string` would be the `run` method. I think that would make the code intentions clearer.

dandrabik: Also, you might consider making this a constant even if you reference as a var, e.g.
```ruby
ORDER_COLUMNS = {....}

def ordered_columns = ORDERED_COLUMNS

File: services/QuillLMS/app/queries/adapters/csv/admin_diagnostic_skills_summary_data_export.rb [hunk 6]
```diff
@@ -0,0 +1,63 @@
...
+      def self.format_lambdas
+        {
+          skill_group_name: format_all_caps,
+          pre_score: format_percent_as_integer,
+          post_score: format_percent_as_integer,
+          growth_percentage: format_percent_as_integer
+        }
+      end
+
+      def self.format_all_caps = ->(x) { x.upcase }
```
dandrabik: Seems like a lot of this are repeated. Maybe they should be in the base class so they don't drift from each other.

Alternatively, I think you might want to break off a separate Service Object for the formatter (general code structure), e.g.

```ruby
class Formatter < ApplicationService

  FORMATTERS = {
    all_caps: ->(x) { x.upcase },
    rounded_integer: ->(x) { x.present? ? x.round : format_blank_as_zero.call(x) },
    ...
  }

  def initialize(format, value)
    ...
  end

  def run = FORMATTERS[format]&.call(value)

end

## Then you can use
def formats
  {
   skill_group_name: :all_caps,
  }
end

Formatter.run(format, value)

File: services/QuillLMS/app/queries/adapters/csv/csv_data_export.rb [hunk 7]
```diff
@@ -0,0 +1,65 @@
...
+      def self.second_row_tooltips = raise NotImplementedError
+      def self.format_lambdas = raise NotImplementedError
+
+      def self.to_csv_string(bigquery_result, columns = ordered_columns.keys)
+        sym_columns = columns.map(&:to_sym)
+        processed_result = pre_process_result(bigquery_result)
+        validate_input!(processed_result, sym_columns)
+
+        CSV.generate do |csv|
+          csv << human_displayable_csv_headers(sym_columns)
```
dandrabik: Do we need `human_displayable` as part of the var names? Would `csv_headers` and `csv_tooltips` be clear enough?

services/QuillLMS/app/queries/adapters/csv/csv_data_export.rb
```diff
@@ -0,0 +1,65 @@
...
+        row[:aggregate_rows].each do |agg_row|
+          csv << ([:name] + sym_columns.slice(1..)).map { |key| format_cell(key, agg_row[key]) }
+        end
+      end
+
+      def self.human_displayable_csv_headers(sym_columns) = sym_columns.map{|col| ordered_columns[col][0] }
+      def self.human_displayable_csv_tooltips(sym_columns) = sym_columns.map{|col| ordered_columns[col][1] }
+
+      def self.format_cell(sym_column, value)
+        return (format_lambdas[sym_column].call(value)) if format_lambdas.include?(sym_column)
```
dandrabik: Take or leave, but you might consider making a `default` formatter which does nothing but return the value, and require the code the specify the value of each column (even if it's `default`). That extra explicitness might be a nice check that you've made a formatting decision about every column and then you could remove this `if`.

services/QuillLMS/app/workers/admin_diagnostic_reports/enqueue_csv_email_worker.rb
```diff
@@ -0,0 +1,70 @@
...
+    private def base_filters = @base_filters ||= rename_base_filter_keys(extract_filter_selection(BASE_REPORT_NAME) || {})
+
+    private def rename_base_filter_keys(processed_filters)
+      processed_filters['teacher_ids'] = processed_filters.delete('teachers') if processed_filters['teachers']
+      processed_filters['classroom_ids'] = processed_filters.delete('classrooms') if processed_filters['classrooms']
+
+      processed_filters
+    end
+
+    private def overview_filters
```
dandrabik: Could these all be methods on the `AdminReportFilterSelection` instance?

services/QuillLMS/app/workers/admin_diagnostic_reports/send_csv_email_worker.rb
```diff
@@ -0,0 +1,146 @@
...
+        timeframe_end: parse_datetime_string(timeframe['timeframe_end']),
+        school_ids:,
+        grades: filters.fetch('grades', nil),
+        teacher_ids: filters.fetch('teacher_ids', nil),
+        classroom_ids: filters.fetch('classroom_ids', nil),
+        user: @user
+      }
+    end
+
+    private def overview_link
```
dandrabik: A lot of these submethods are pretty complicated, e.g. this one parses filters, runs 5 queries, does 4 merges of data, generates a csv string, and then uploads that csv string. This could be its own service object (possibly several). I wonder if you sliced this code in verticals by report, you might be able to break it into smaller, one responsibility services, i.e. instead of generating all of the filters for the reports in one job, and then generating all of the reports in one job, you could have, for `one` report, a service to generate the filters, a service to run the queries and generate the data in the right shape, a service to generate the csv (you have this more or less) and upload the csv (this could be a generic service). I think you'd end up with much smaller classes that have one clear responsibility. It doesn't have to be that exact setup, but reading this code, each method seems like a service unto itself.

dandrabik (changes requested):
Nice work on getting this running. I left several refactor comments that I think it's worth do a refactor pass on this to make sure this is maintainable long term. I think you've done a good job getting the logic and functionality working, and I think the code would now benefit from some reorganization. I'm happy to jump on a call to chat through any of these comments.

services/QuillLMS/app/workers/admin_diagnostic_reports/enqueue_csv_email_worker.rb
anathomical: This is a fair question.  I don't think it has to be split into two parts, but I'm partially anticipating a future feature (which may be a mistake at this point).  Specifically, the plan is to eventually add the ability for admins to subscribe to regular emails of these reports.  I expect that if we do that, then we'll actually be using a different set of saved filters, so separating the logic of extracting and formatting filters from the actual CSV generation made sense to me.  Open to the idea that we should just wait until building that future feature before worrying about the split.

services/QuillLMS/app/mailers/admin_diagnostic_reports/report_mailer.rb
```diff
@@ -0,0 +1,21 @@
...
+    def csv_download_email(user_id, overview_url, skills_url, students_url)
+      user = User.find(user_id)
+      @addressee_name = user.first_name
+      @overview_url = overview_url
+      @skills_url = skills_url
+      @students_url = students_url
+      @human_date = DateTime.current.strftime('%B %d, %Y')
+
+      #email = user.email
+      email = 'thomas@quill.org'
```
happythenewsad: Better address these two lines before production :-)

happythenewsad: In the past, I've used an app setting, so that I can mail myself things during development, but also have the same code deployable to prod.

services/QuillLMS/app/queries/adapters/csv/csv_data_export.rb
```diff
@@ -0,0 +1,58 @@
+# frozen_string_literal: true
```
happythenewsad: This class shares a lot of similar method implementations and signatures to AdminPremiumDataExport. 

- `row_contains_requested_columns?`, identical
- `BigQueryResultMissingRequestedColumnError`, identical 
- `format_cell`, `to_csv_string`, `validate_input!` similar 

Please consider refactoring this class so that code is less duplicated, via inheritance, module inclusion, or some other method.

services/QuillLMS/app/workers/admin_diagnostic_reports/enqueue_csv_email_worker.rb
happythenewsad: I think it's best to not split it for now. Reasoning:

- Outside of the context of this PR diff, the intention of the code is harder to follow when it's split across two jobs
- every job adds additional overhead of managing idempotency and failure/retry states

services/QuillLMS/app/queries/adapters/csv/csv_data_export.rb
anathomical: I like this idea a lot, and have implemented it.

services/QuillLMS/app/workers/admin_diagnostic_reports/enqueue_csv_email_worker.rb
anathomical: Good call.

services/QuillLMS/app/workers/admin_diagnostic_reports/send_csv_email_worker.rb
anathomical: Refactored into a number of smaller services for clarity.

services/QuillLMS/app/mailers/admin_diagnostic_reports/report_mailer.rb
anathomical: I've followed your pattern.

services/QuillLMS/app/queries/adapters/csv/csv_data_export.rb
anathomical: I've refactored this adapter, and also refactored your upload worker to use the new `UploadToS3` service.

services/QuillLMS/app/mailers/admin_diagnostic_reports/report_mailer.rb
```diff
@@ -0,0 +1,20 @@
+# frozen_string_literal: true
+
+module AdminDiagnosticReports
+  class ReportMailer < ::UserMailer
+    default from: "The Quill Team <hello@quill.org>"
```
brendanshean: `default from:` is already inherited from `::UserMailer`

services/QuillLMS/app/controllers/admin_diagnostic_reports_controller.rb
```diff
@@ -6,14 +6,24 @@ class AdminDiagnosticReportsController < ApplicationController
...
-  before_action :validate_request
-  before_action :authorize_request
+  BASE_REPORT_FILTER_NAME = 'diagnostic_growth_report'
+  OVERVIEW_REPORT_FILTER_NAME = 'diagnostic_growth_report_overview'
+  SKILL_REPORT_FILTER_NAME = 'diagnostic_growth_report_skill'
+  STUDENT_REPORT_FILTER_NAME = 'diagnostic_growth_report_student'
+
+  before_action :set_query, only: [:report]
+  before_action :validate_request, only: [:report]
+  before_action :authorize_request, only: [:report]
```
brendanshean, line 16: Does download action need to be authorized?

services/QuillLMS/spec/services/admin_diagnostic_reports/construct_overview_query_payload_spec.rb
```diff
@@ -0,0 +1,25 @@
...
+    let(:specific_filter_name) { AdminReportFilterSelection::DIAGNOSTIC_GROWTH_REPORT_OVERVIEW }
+    let(:specific_filters) { create(:diagnostic_report_overview_selection, user:, report: specific_filter_name) }
+
+    it { expect(specific_filters.aggregation).to eq(subject[:aggregation]) }
+    it { expect(subject.keys).not_to include(:diagnostic_id) }
+
+    context 'no saved specific filters' do
+      let(:specific_cilters) { nil }
+
+      it { expect(described_class::DEFAULT_AGGREGATION).to eq(subject[:aggregation]) }
```
brendanshean: `subject` should be in the expect clause:

```
      it { expect(subject[:aggregation]).to eq(described_class::DEFAULT_AGGREGATION) }
```

services/QuillLMS/app/services/upload_to_s3.rb
```diff
@@ -0,0 +1,33 @@
+# frozen_string_literal: true
+
+class UploadToS3 < ApplicationService
```
brendanshean, line 3: I really like this extraction.  I think it needs to be renamed to `UploadCsvToS3` or something with CSV since it's not a generic uploader but a `CSV` one.

services/QuillLMS/app/services/admin_diagnostic_reports/construct_query_payload.rb
```diff
@@ -0,0 +1,59 @@
...
+    DEFAULT_AGGREGATION = 'grade'
+    DEFAULT_DIAGNOSTIC_ID = 1663
+
+    def initialize(user_id, shared_filters_name, specific_filters_name)
+      @user_id = user_id
+      @shared_filters_name = shared_filters_name
+      @specific_filters_name = specific_filters_name
+    end
+
+    def run
```
brendanshean: `def run = payload`

services/QuillLMS/app/queries/admin_diagnostic_reports/diagnostic_recommendations_by_student_query.rb
```diff
@@ -51,7 +52,7 @@ def relevant_diagnostic_where_clause = "AND activity_id = #{diagnostic_id}"
 
     def group_by_clause = ""
     def order_by_clause = "ORDER BY TRIM(SUBSTR(TRIM(students.name), STRPOS(students.name, ' ') + 1))"
-    def limit_clause = "LIMIT 500"
+    def limit_clause = "LIMIT 500" if @limited
```
brendanshean: Does this need attr_reader?

services/QuillLMS/spec/services/formatter_spec.rb
```diff
@@ -0,0 +1,138 @@
...
+require 'rails_helper'
+
+describe Formatter do
+  subject { described_class.run(selected_format, value) }
+
+  context 'default' do
+    let(:selected_format) { described_class::DEFAULT }
+    let(:value) { 'Should not be modified' }
+
+    it { expect(subject).to eq(value) }
```
brendanshean: The spec file switches between `it { expect(subject)... }` and `it { is_expected }`. For consistency, what about using `it { is_expected... }` throughout?

brendanshean (commented):
I've left some general comments.  I think the CSV Data Export piece should use a Service Object pattern, but short of that, I think this is good to go. 

Overall, this is quite an impressive PR pulling in a lot of moving parts.

services/QuillLMS/app/queries/adapters/csv/admin_diagnostic_overview_data_export.rb
brendanshean: I would agree with this point.  I found the class variables surprising, especially in the inheritance hierarchy.

services/QuillLMS/app/services/admin_diagnostic_reports/construct_query_payload.rb
happythenewsad: This comment should be tagged [Optional], since it's not currently enforced by our Rubocop or our style guide. 

Rubocop's style guide considers it "opt-in": https://github.com/rubocop/ruby-style-guide?tab=readme-ov-file#endless-methods

There's also a Cop available: https://www.rubydoc.info/gems/rubocop/RuboCop/Cop/Style/EndlessMethod. We have the Cop enabled in LMS, but not to the specificity that it *requires* endless methods.

FWIW I like the Endless Method syntax in most cases, but want to allow for developer discretion in cases we don't necessarily anticipate right now.

happythenewsad (approved):
Big lift 🏋️‍♂️ 

Thanks for doing the refactors

services/QuillLMS/app/controllers/admin_diagnostic_reports_controller.rb
anathomical: Good question.  I think that in this case, because the code logic generates and sends a report to `current_user`, we don't need auth.  You can never get access to data that's not yours.  Open to discussion, though.

services/QuillLMS/app/services/admin_diagnostic_reports/construct_query_payload.rb
brendanshean: As we've discussed in engineering meetings, the endless syntax is **always** optional.  I only point it out as a stylistic suggestion.

brendanshean: But I think your point about explicitly tagging as optional is correct.

services/QuillLMS/app/queries/admin_diagnostic_reports/diagnostic_recommendations_by_student_query.rb
anathomical: Yes.  Good catch.  I thought I'd cleaned all those up, but missed this one.

services/QuillLMS/app/services/upload_to_s3.rb
```diff
@@ -0,0 +1,47 @@
...
+    EXTENSION_UPLOADERS.fetch(extension)
+  rescue KeyError
+    raise UndefinedExtensionError, "'#{extension}' is not a defined extension for the uploader"
+  end
+
+  private def upload_csv
+    # store! returns nil on failure, rather than raising an exception.
+    # We address this gotcha by manually raising an exception.
+    upload_status = uploader.store!(local_tempfile)
+    raise CloudUploadError, "Unable to upload CSV for user #{user.id}" unless upload_status
```
brendanshean: The extension should be extracted out of this string: `Unable to upload #{extension} for user #{user.id}`

services/QuillLMS/app/services/upload_to_s3.rb
```diff
@@ -0,0 +1,47 @@
...
+  def initialize(user, payload, extension)
+    @user = user
+    @payload = payload
+    @extension = extension
+  end
+
+  def run
+    local_tempfile << payload
+
+    upload_csv
```
brendanshean: Maybe just `upload` (or `upload_file`) here?

services/QuillLMS/app/queries/adapters/csv/admin_diagnostic_overview_data_export.rb
anathomical: I wrote these "adapters" following an established pattern from `app/queries/adapters/csv/admin_premium_data_export.rb` that we deployed last year for the premium data export report.  I'm not opposed to refactoring them into a service, but figured that there was some value to maintaining consistency of approach.  If we do refactor these, I imagine that we'd also want to refactor the historical example, too, for consistency sake.

brendanshean (commented):
A couple more comments on UploadToCSV

services/QuillLMS/app/queries/adapters/csv/admin_diagnostic_overview_data_export.rb
brendanshean: I think refactoring them would improve the readability of the code.  Initially there were just a couple of class methods on that file, but now there are a bunch.  I should have called that out in the earlier PR, so my mistake there.

dandrabik (approved):
This looks much cleaner with more focused classes. Thanks for cleaning this up. Looks good to me.

brendanshean (approved):
Thank you for doing the service object extraction.  This looks great! :shipit:

PR #11874, opened 2024-05-30
Update the width on hints to 100%

samanthamjohn:
## WHAT
Made the hint box full width. 

## WHY
The hints box should always be full width, but in the case where the hint text was too short, it was not coming up as full width. 

## HOW
Edited the CSS to add width: 100% 

### Screenshots
Production: 
2024-05-30 at 11 31 AM

New: 
2024-05-30 at 11 31 AM

### Notion Card Links
https://www.notion.so/quill/Ensure-all-Concept-Feedback-spans-the-width-of-the-activity-20d47a7b07c3439a8641a9ec49a68d3e?pvs=4

Looked at it on my local machine. 

PR Checklist | Your Answer
------------ | -------------
Have you added and/or updated tests? |  NO, because it was just a style change. 

### DIFF
--- a/services/QuillLMS/client/app/bundles/Diagnostic/styles/components/_student-question.scss
+++ b/services/QuillLMS/client/app/bundles/Diagnostic/styles/components/_student-question.scss
@@ -137,54 +137,6 @@
   transform: rotate(180deg);
 }
 
-.concept-explanation {
-  font-family: $family-monospace;
-  margin: 1em 0;
-  font-size: 20px;
-  width: 100%;
-  .concept-explanation-title {
-    font-weight: bold;
-    padding-left: 15px;
-    padding-right: 15px;
-    display: flex;
-    align-items: center;
-    img {
-      margin-right: 8px;
-    }
-  }
-  .concept-explanation-description {
-    padding-left: 15px;
-    padding-right: 15px;
-    margin-bottom: 15px;
-  }
-
-  .concept-explanation-see-write {
-    display: flex;
-    background-color: #fefefe;
-    padding: 15px 25px;
-    border: 1px dashed #cecece;
-    // border-radius: 5px;
-    margin-bottom: 0.5em;
-    .concept-explanation-see {
-      flex-grow: 1;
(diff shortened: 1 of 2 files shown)


services/QuillLMS/client/app/bundles/Diagnostic/styles/components/_student-question.scss
```diff
@@ -141,6 +141,7 @@
   font-family: $family-monospace;
   margin: 1em 0;
   font-size: 20px;
+  width: 100%;
```
samanthamjohn: This change was not strictly needed to fix the bug as reported but it has the same selector as the one that I had to change. 
But, this is in the diagnostic styles, and the diagnostic does not seem to have hints. I'd propose deleting this selector from the CSS, but it seems a little dangerous on my third day. Would love to hear more context/opinions from the team.

emilia-friedberg: I believe you're correct that the diagnostic will never show hints (because there are no revisions). Connect and Diagnostic used to be one app, and when they were split the code was duplicated (deduplication and recombining the various apps into one is an ongoing topic of conversation), which accounts for the unused CSS here. It should be safe to delete this selector, since it's a) not being used and b) styles in the Shared bundle will apply to all apps, so even if it were to be used it would get the CSS from there, but I would defer to your comfort level -- my inclination is that, on a bug fix, it is always nice but never necessary to do clean-up you come across along the way.

emilia-friedberg (approved):
LGTM!

services/QuillLMS/client/app/bundles/Diagnostic/styles/components/_student-question.scss
```diff
@@ -137,53 +137,6 @@
   transform: rotate(180deg);
 }
 
-.concept-explanation {
```
samanthamjohn, line 140: @emilia-friedberg I deleted it! Hopefully nothing weird happens :)

PR #11875, opened 2024-05-30
Add G-Eval to experiment configuration

brendanshean:
## WHAT
Add [g-eval]() to the GenAI experiment configuration.

## WHY
We'd like to use a SOTA evaluation tool for LLM feedback

## HOW
1.  Add GEval model to store different G-eval configurations
1.  Add AutoChainOfThoughtController.   G-Eval generation requires an intermediate step of generating a "Evaluation Steps" (see Auto CoT in screenshot below) from an LLM.  This controller handles the intermediate step and redirects to the GEvalController, specifically the new template.
1.  Add GEval runner to evaluate LLM Feedback from Ideal Feedback using a g-eval configuration.

### Screenshots
Figure from [arxiv](https://arxiv.org/abs/2303.16634) paper:
Screenshot 2024-05-30 at 3 47 10 PM

Screenshot 2024-05-30 at 3 59 45 PM

Screenshot 2024-05-30 at 3 14 43 PM

### Notion Card Links
https://www.notion.so/quill/Host-G-Eval-endpoint-45eef4c938e445d78875e4f9bf8d941d?pvs=4

I tested this functionality out both locally and on staging.

PR Checklist | Your Answer
------------ | -------------
Self-Review: Have you done an initial self-review of the code below on Github? | YS

### DIFF
--- a/services/QuillLMS/config/initializers/zeitwerk.rb
+++ b/services/QuillLMS/config/initializers/zeitwerk.rb
@@ -8,6 +8,8 @@ Rails.autoloaders.each do |autoloader|
     'concept_replacement_lms_worker' => 'ConceptReplacementLMSWorker',
     'ell_starter_diagnostic_email_job' => 'ELLStarterDiagnosticEmailJob',
     'gen_ai' => 'GenAI',
+    'g_eval' => 'GEval',
+    'g_evals_controller' => 'GEvalsController',
     'graphiql' => 'GraphiQL',
     'html_tag_remover' => 'HTMLTagRemover',
     'llm_config' => 'LLMConfig',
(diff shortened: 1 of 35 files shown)


dandrabik (approved):
Overall, this looks good. I think I was picturing just having a single G-Eval config, but I see how this is a more accurate recording of a snapshot of the config at the time it was run. I'm curious how many configs we go through to find one that works well.

Anyway, thanks for putting this together! Loving the progress on this!

brendanshean:
> Overall, this looks good. I think I was picturing just having a single G-Eval config, but I see how this is a more accurate recording of a snapshot of the config at the time it was run. I'm curious how many configs we go through to find one that works well.
> 
> Anyway, thanks for putting this together! Loving the progress on this!

Initially, I was thinking a single G-Eval config too, but then I realized, much like the LLM Prompt, it makes more sense to iterate on configs until we're happy with one, then use that.

I could also see there being one config per conjunction.  This framework will give us flexibility to defer on that question until later.

PR #11882, opened 2024-06-03
Update question flags

samanthamjohn:
## WHAT
Make the grammar module update questions when you updated an activity's flag to production. 

## WHY
It's quite time-consuming to have to go in and update all the question's flags individually when you update the activity's flag. Plus this feature was already working in the Connect and Diagnostic modules. 

## HOW
- Added logic to the server API for updating an activity to make it also update the questions' flags if you were updating it to the production flag.
- Removed the client-side logic to update each questions individually in the Connect and Diagnostic modules. 
- Added logic to reload the questions when an activity was updated and had the production flag. 

## Extras
- Renamed a folder from `tests` to `__tests__` for consistency with other JS tests. 
- Fixed a small typescript type warning

### Notion Card Links
https://www.notion.so/quill/Ensure-that-when-a-Grammar-activity-is-moved-to-production-the-prompts-within-that-activity-are-aut-06ce91a5356a429aa3450a7dd71d47dc?pvs=4

(Provide enough detail that a reviewer could assess whether additional QA should be done. For larger projects, additionally use the Engineer Feature Testing Notion template. Review Guidelines if needed: https://www.notion.so/quill/Github-PR-QA-Guidelines-49e99fc965654ceeb8c6249bd9d181d7

PR Checklist | Your Answer
------------ | -------------


samanthamjohn:
I still need to fix some stuff, so not quite ready to review

samanthamjohn:
And net negative lines of code FTW!

services/QuillLMS/client/app/bundles/Shared/actions/questions.ts
```diff
@@ -0,0 +1,22 @@
...
+import connect from "../../Connect/actions/questions";
+import diagnostic from "../../Diagnostic/actions/questions";
+import { CONNECT, DIAGNOSTIC, GRAMMAR } from "../utils/constants";
+
+export const reloadQuestionsIfNecessary = (flag: String, activityType: String) => {
+  if (flag !== PRODUCTION) { return null; }
+  return (dispatch: Function) => {
+    switch (activityType) {
+      case GRAMMAR:
+        dispatch(grammarListenToQuestions());
```
samanthamjohn, line 12: I created a shared action for this new method so that it could be used by all three modules. Emilia and I discussed passing the functions directly into the code rather than importing them which could also be done if folks have strong feelings. The advantage of this method is that it sets us up for moving the `loadQuestions`/`startListeningToQuestions` functions into shared logic as well.

services/QuillLMS/app/models/activity.rb
```diff
@@ -313,6 +313,18 @@ def serialize_with_topics_and_publication_date
     serializable_hash.merge({topics: topics&.map(&:genealogy), publication_date: publication_date})
   end
 
+  def update_questions_flag_status_if_necessary!
+    return unless flags.include?(:production)
+
+    question_ids = data["questions"].map{|q| q["key"]}
+    questions = Question.where("uid in (?)", question_ids)
```
emilia-friedberg: Tiny, but I'd rename this variable `question_uids` to avoid ambiguity (since questions do also have ids).

emilia-friedberg (approved):
One tiny variable name tweak suggestion, but otherwise LGTM! Thank you for all the clean-up here.

eadams17 (approved):
Appreciate all the clean up here!

services/QuillLMS/app/controllers/api/v1/questions_controller.rb
```diff
@@ -28,11 +28,6 @@ def destroy
     render(plain: 'OK')
   end
 
-  def update_flag
-    @question.update_flag(valid_params[:flag])
-    render_question
-  end
-
```
anathomical, line 35: I'm not positive that this is a use-case (might be worth checking with Jamie about it), but what if someone wants to change the Flag of a Question from Production to something else (assuming you've already changed the Activity's flag)?  It looks to me like we're disabling that functionality entirely?  That might be reasonable, but I'd be hesitant to make it impossible to make those updates.

services/QuillLMS/app/models/activity.rb
```diff
@@ -313,6 +313,18 @@ def serialize_with_topics_and_publication_date
     serializable_hash.merge({topics: topics&.map(&:genealogy), publication_date: publication_date})
   end
 
+  def update_questions_flag_status_if_necessary!
+    return unless flags.include?(:production)
+
+    question_uids = data["questions"].map{|q| q["key"]}
+    questions = Question.where("uid in (?)", question_uids)
```
anathomical: This is fine as-is, but it feels like these could be two methods on the model that could be useful in other places:  (Also note the `&` null-safe syntax so that this won't crash if somehow `data` doesn't have a "questions" key.)
```
def question_uids = data["questions"]&.map{|q| q["key"]}

def questions = Question.where(uid: question_uids)
```

services/QuillLMS/spec/models/activity_spec.rb
```diff
@@ -769,4 +769,41 @@
...
+        question.data["flag"] = "beta"
+        question.save
+        activity = create(:grammar_activity,
+        flags: ["alpha"],
+        data: {questions: [{"key" => question.uid, "question_type" => "questions"}]}
+        )
+        activity.update_questions_flag_status_if_necessary!
+        expect(question.reload.data["flag"]).to eq("beta")
+      end
+    end
```
anathomical: The test coverage itself is fine, but I'd do these quite a bit differently stylistically:
- I'd do my data setups (like the `build`s) in `let` calls
- I often like to use `expect{ }.to change()`any time I'm tempted to use `reload`
- One thing we're moving more and more toward is concentrating code into `context` and trying to make `it` blocks small enough that descriptions are unnecessary

Something like this, maybe:
```
describe "#update_questions_flag_status_if_necessary!" do
  let(:q1) { build(:question, flag: "beta") }
  let(:q2) { build(:question, flag: "beta") }
  let(:questions) { [{"key" => q1.uid, "question_type" => "questions"}, {"key" => q2.uid, "question_type" => "questions"}] }
  let(:data) { {questions: } } # note that the unmatched `x:` syntax in Ruby will try to set the value by finding a variable with the name of the key, so in this case "questions"
  let(:flags) { ["production"] }
  let(:activity) { create(:grammar_activity, flags:, data:) }

  context 'activity flag set to production' do
    it do
      expect do
        activity.update_questions_flag_status_if_necessary!
      end.to change(q1, :flag).to("production")
        .and change(q2, :flag).to("production")
    end
  end

  context 'activity flag is not set to production' do
    let(:flags) { ["alpha"] }

    it do
      expect do
        activity.update_questions_flag_status_if_necessary!
      end.not_to change(q1, :flag)
    end
  end
end
```
Philosophically, this approach tries to reduce the `it` descriptions as much as possible because we've found that it's one of the most likely things to drift: tests get updated without their descriptions, so descriptions often end up being wrong.  By describing contexts and reducing the amount of code in the `it` blocks we attempt to make the `it` blocks kind of close to "self-documenting" as possible.

anathomical (changes requested):
Generally speaking, this looks good.  Two stylistic things that I'd like to resolve before proceeding.  (These could be resolved by just discussing, no necessarily through code changes.)
1. I'm a little worried about removing the API endpoint to update Question flag values since it feels like there should be cases where we want to update that value that's not just the case where they get upgraded to "production".  But I could be wrong about that.
2. The Ruby spec itself seems good, but I've got some comments on style.

services/QuillLMS/app/models/activity.rb
samanthamjohn: I agree! I felt like it would be nice to be able to just call `activity.questions` but didn't want to go too into the weeds.

services/QuillLMS/spec/models/activity_spec.rb
samanthamjohn: i updated this to use the `let` and `expect` syntax. I did have to use `create` instead of build, and unfortunately still had to reload the questions. Getting into that `data` blob is a little bit tricky.

services/QuillLMS/app/models/activity.rb
samanthamjohn: updated to use the `?`

services/QuillLMS/app/controllers/api/v1/questions_controller.rb
samanthamjohn: The way that you change the question flag is by editing the question itself--there's no UI to change just the flag, and the old way of changing a question by using the update path still works.

samanthamjohn: So this endpoint was only used in the old code path where we would update the question flags in javascript after you had updated the activity. We should be safe!

anathomical (approved):

PR #11883, opened 2024-06-03
Backpack/ lists

eadams17:
## WHAT
add lists to Backpack

## WHY
we are expanding our component library

## HOW
created new `List` component with 3 style types and add to style guide

### Screenshots
Screen Shot 2024-06-03 at 6 58 14 PM

### Notion Card Links
https://www.notion.so/quill/Quill-Backpack-Buttons-Menus-Text-Fields-Segmented-Controls-Lists-85b0fa6c8cee4a108a015f307befb7a7?pvs=4

testing on staging that the different lists behave and display as expected

PR Checklist | Your Answer
------------ | -------------


services/QuillLMS/client/app/bundles/Shared/components/shared/list.tsx
```diff
@@ -0,0 +1,61 @@
...
+const SECONDARY = 'secondary'
+const CHECKBOX = 'checkbox'
+
+export const List = ({ items, style }: ListProps) => {
+  if(style === PRIMARY) {
+    return(
+      <ul className="list quill-list single-line">
+        {items.map(({ label, onClick }) => (
+          <li>
+            <button className="interactive-wrapper" onClick={onClick}>{label}</button>
```
emilia-friedberg: I think this button probably needs a `type` to pass the linter, right? I also wonder if we want to give it a `focus-on-light` class.

services/QuillLMS/client/app/bundles/Shared/components/shared/list.tsx
```diff
@@ -0,0 +1,61 @@
...
+        ))}
+      </ul>
+    )
+  }
+  if(style === SECONDARY) {
+    return (
+      <ul className="list quill-list double-line">
+        {items.map(({ primaryText, secondaryText, onClick }) => (
+          <li>
+            <button className="interactive-wrapper" onClick={onClick}>
```
emilia-friedberg: See comment on button above.

emilia-friedberg (commented):
I left a couple of notes about the buttons, but more generally, I'm wondering if there is always on `onClick` handler getting passed in here? Maybe I'm wrong but the design comp to me looks like maybe sometimes it's just a static list, in which case I think we want to do some logic to check to see if an `onClick` is getting passed in and if not, render a static element rather than a button (to avoid weird UX for keyboard users and others).

services/QuillLMS/client/app/bundles/Shared/components/shared/list.tsx
eadams17: I think that's a reasonable approach-- I just assumed by the focus and active states that these would always be interactive elements. I'll add logic to check for a click handler and if none is present, just render a regular `li` element

emilia-friedberg (approved):
Looks great! Thank you for making that change.

PR #11885, opened 2024-06-04
adding multiple sessions for demo account

emilia-friedberg:
## WHAT
Update the demo account generation data and script to allow for repeated sessions for activities.

## WHY
So we can show off in the demo account that teachers and students can view repeated attempts.

## HOW
Update the generation yml files with multiple sessions from production; update the script to allow for finding multiple sessions and storing them all.

### Screenshots
Screenshot 2024-06-04 at 11 26 43 AM

### Notion Card Links
https://www.notion.so/quill/Add-activities-with-multiple-attempts-to-the-Quill-demo-accounts-student-teacher-5e515c32e0654f3fbbf5e9762888db4e?pvs=4

Updated the demo account generation script and ran it to confirm new sessions show up where they should.

PR Checklist | Your Answer
------------ | -------------


services/QuillLMS/app/services/demo/report_demo_creator.rb
```diff
@@ -362,16 +359,16 @@ def self.create_classroom_units(classroom, units)
     end
   end
 
-  def self.create_replayed_activity_session(student, classroom_unit, session_data)
-    clone_activity_session(student.id, classroom_unit.id, REPLAYED_SAMPLE_USER_ID, REPLAYED_ACTIVITY_ID, session_data)
-  end
+  def self.clone_activity_sessions(student_id, classroom_unit_id, clone_user_id, clone_activity_id, session_data)
```
brendanshean, line 362: You could make this more functional which would get rid of the guard clause and temporary variable: 

```
 def self.clone_activity_sessions(student_id, classroom_unit_id, clone_user_id, clone_activity_id, session_data)
    session_data
      .activity_sessions
      .filter {|session| session.activity_id == clone_activity_id && session.user_id == clone_user_id}
      .each { |session| clone_activity_session(student_id, classroom_unit_id, clone_activity_id, session, session_data) }
  end
```

brendanshean (approved):
Thanks for adding this functionality.

I think it's funny that n=1 code tends to look more complicated than n = many 
(n here is number of replayed activity_sessions)

services/QuillLMS/app/services/demo/session_data.rb
```diff
@@ -87,8 +84,13 @@ def generate_files
...
 
-        ActivitySession.unscoped.find_by(activity_id: activity_id, user_id: user_id, is_final_score: true)
-      end.compact
+        # we only want to find the first one if it's a diagnostic because there should be no replays
+        if Activity.diagnostic_activity_ids.include?(activity_id)
+          ActivitySession.unscoped.find_by(activity_id: activity_id, user_id: user_id, state: ActivitySession::STATE_FINISHED)
+        else
+          ActivitySession.unscoped.where(activity_id: activity_id, user_id: user_id, state: ActivitySession::STATE_FINISHED)
+        end
+      end.flatten.compact
```
anathomical: This is purely stylistic, but we can use an early return pattern here inside a loop block the same way we do in methods except we use `next` instead of `return`:
```
ACTIVITY_USER_PAIRS.map do |id_pair|
  activity_id, user_id = id_pair

  # we only want to find the first one if it's a diagnostic because there should be no replays
  next ActivitySession.unscoped.find_by(...) if Activity.diagnostic_activity_ids.include?(activity_id)

  ActivitySession.unscoped.where(...)
end.flatten.compact
```

anathomical (approved):
Minor stylistic note, but this looks good to me.

PR #11886, opened 2024-06-04
Lw tags

happythenewsad:
## WHAT / WHY
- add Sidekiq workers to manage daily tag sync from LMS to Learnworlds
- rename and move existing LW files so that they are minimally dispersed around the filesystem and all share the `LearnWorldsIntegration` module 
- refactor SSORequest so that it subclasses from UserRequest, so that LearnWorlds headers are not duplicated in various subclasses (more functionality may be shared via inheritance in future PRs)

### Screenshots

### Notion Card Links
[(Please provide links to any relevant Notion card(s) relevant to this PR.)](https://www.notion.so/quill/Product-Board-30f97e4eb01246dbb8264253fb385073?p=bd19e778527a48e58ecdaabe8b46cd30&pm=s)

1. Verify SSO for both new and existing users on staging (which is now hooked up to the staging LW instance)
2. Run LearnWorldsIntegration::SyncOrchestratorWorker.perform_async manually from a staging instance and observe the tags appearing in the LW UI.

PR Checklist | Your Answer
------------ | -------------


services/QuillLMS/app/workers/learn_worlds_integration/sync_orchestrator_worker.rb
```diff
@@ -0,0 +1,40 @@
...
+    # NOTE: this worker 'smears' API calls over time to avoid hitting
+    # the LearnWorlds 30 requests / 10 seconds rate limit.
+    SMEAR_RATE_IN_SECONDS = 1
+
+    def perform
+      userwise_subject_areas_relation = LearnWorldsAccount.all.includes(user: { teacher_info: [:subject_areas] } )
+      userwise_subject_areas_relation.find_each.with_index do |row, idx|
+        next unless row&.user
+
+        SyncUserTagsWorker.perform_in((idx * SMEAR_RATE_IN_SECONDS).seconds, *marshal(row))
```
happythenewsad, line 16: Testing on staging with > 1000 records indicated that this smear rate is effective.

services/QuillLMS/.env-sample
```diff
@@ -58,6 +58,10 @@ OPINION_API_DOMAIN=https://opinion-api.quill.org
 GRAMMAR_API_DOMAIN=https://quill.spell.services/Quill/grammar/predict
 
 
+# LEARNWORLDS - copy the secret values from a Heroku staging instance
+LEARN_WORLDS_BASE_URI=https://quillacademy-staging.learnworlds.com
+LEARN_WORLDS_CLIENT_ID=
+LEARN_WORLDS_ACCESS_TOKEN=
```
anathomical, line 64: Oh, always love it when someone remembers to add new ENV variables to this file.  I always forget.

services/QuillLMS/app/models/cron.rb
```diff
@@ -86,6 +86,7 @@ def self.run_at_17_hour_mark
   # 4/5PM depending on daylight savings
   def self.run_at_21_hour_mark
     AdminDiagnosticReports::PerformanceBenchmarkWorker.perform_async
+    LearnWorldsIntegration::SyncOrchestratorWorker.perform_async
```
anathomical: No objection to this, but is there a reason to run this at a particular time of day, or should we put this with the rest of the "overnight" jobs in `interval_1_day`?  Mostly I think we should reserve `run_at_X` for things we have strong reason to run at that time.  Which may be this task, I just wanted to cofnirm.

services/QuillLMS/app/services/learn_worlds_integration/sso_request.rb
```diff
@@ -0,0 +1,47 @@
...
+
+    def initialize(user)
+      @user = user
+    end
+
+    def run
+      raise NilUserError if user.nil?
+      raise NilEmailError if email.nil?
+
+      HTTParty.post(endpoint, body: body, headers: headers)
```
anathomical: I'm pretty sure that newer Ruby syntactic sugar allows you to implicitly pull values if they are stored in a variable with the same name as the key.  So I think you can write this as `HTTParty.post(endpoint, body:, headers:)` which avoids a bit of unnecessary boilderplate.

services/QuillLMS/app/services/learn_worlds_integration/user_tags_request.rb
```diff
@@ -0,0 +1,24 @@
...
+    end
+
+    def initialize(external_id, tags)
+      raise ArgumentError, "Invalid tags: #{tags}" unless tags.respond_to?(:each)
+
+      @external_id = external_id
+      @tags = tags
+    end
+
+    def run = HTTParty.put(endpoint, body: data.to_json, headers: headers)
```
anathomical: Again, you can shortcut the `headers:` part of this since the value has the same name as the key.  So: `HTTParty.put(endpoint, body: data.to_json, headers:)`

Additional thought: in order to take advantage of that, thoughts on renaming `def data` to `def body` and including the `to_json` call in that definition?  Feel free to not do that, but I do like naming consistency when it's easy.

services/QuillLMS/app/workers/learn_worlds_integration/sync_orchestrator_worker.rb
anathomical: The way that I understand `perform_in` to work is that the job is not necessarily run on the specified interval, but it is enqueued on the specified interval.  This job were to run at a time when there are (say) 20 minutes of other jobs already in the queue, a bunch of these jobs will hit their `perform_in` times and end up appended to the queue.  This means that they'll all run at once without the intended time spacing.

If we're going to use this smear approach, we'll need to do our best to ensure that this job runs at a time when the queue is very unlikely to have many pending jobs in it so that these don't stack up on the end of said queue.  If a rate limit issue causes the `SyncUserTagsWorker` to retry, that may cover us in instances where this approach breaks (we use this approach in `SyncVitallyOrganizationWorker` which gets a `perform_in` call from the `SyncVitallyWorker`).

anathomical (approved):
I'm approving this as I think that we can deploy this as-is, but I've also flagged a potential concern around `perform_in`.  I think it's likely that this code could already handle the need to retry when running into rate-limiting, but if not there is a need for modification.

services/QuillLMS/app/workers/learn_worlds_integration/sync_orchestrator_worker.rb
happythenewsad: Yes, the "job bunching" issue is a legitimate concern. I considered several ways to get around the rate limit; none of them are perfect.

If we see more retries than are acceptable in prod, we could 
1) adjust the smear rate
2) move the jobs into the `experiment` queue, which has concurrency = 1

services/QuillLMS/app/models/cron.rb
happythenewsad: Overnight is fine

services/QuillLMS/app/services/learn_worlds_integration/user_tags_request.rb
```diff
@@ -0,0 +1,26 @@
...
+
+      @external_id = external_id
+      @tags = tags
+    end
+
+    def run = HTTParty.put(endpoint, body:, headers:)
+
+    def body = data.to_json
+
+    def data = { tags: tags }
```
brendanshean: `def data = { tags: }`

services/QuillLMS/spec/services/learn_worlds_integration/user_tags_request_spec.rb
```diff
@@ -0,0 +1,26 @@
+# frozen_string_literal: true
+
+require 'rails_helper'
+
+RSpec.describe LearnWorldsIntegration::UserTagsRequest do
```
brendanshean, line 5: Some refactoring suggestions:
1). Extract out subject: If the method changes you only have to update in one place.
2). Parameterize `example_tags` in the same way for each spec.  This allows for easy scanning of what changes in each spec.

```
RSpec.describe LearnWorldsIntegration::UserTagsRequest do
  subject { described_class.run(example_external_id, example_tags) } 

  let(:example_external_id) { 123 }

  context 'invalid parameters' do
    let(:example_tags) { 'i dont belong here' }
      
    it { expect { subject }.to raise_error ArgumentError }
  end
    
  context 'valid parameters'
    let(:example_tags) { ['math'] }

    it 'invokes a PUT request' do
      expect(HTTParty).to receive(:put)
      subject
    end
  end
end
```

services/QuillLMS/spec/workers/learn_worlds_integration/sync_orchestrator_worker_spec.rb
```diff
@@ -0,0 +1,52 @@
...
+  describe '#string_to_subject_area_tag' do
+    it 'converts a string to a LearnWorlds tag' do
+      expect(worker.string_to_subject_area_tag('History / Social Studies'))
+        .to eq('subject_area_history_social_studies')
+      expect(worker.string_to_subject_area_tag('English as a New Language'))
+        .to eq('subject_area_english_as_a_new_language')
+    end
+  end
+
+  describe '#tags' do
```
brendanshean, line 17: Some refactoring suggestions

```
 describe '#tags' do
    subject { worker.tags(user) }  # extract out subject

    let(:user) { create(:user) }
    
    # Extract out authorization higher up so in future tests I will be reminded to specify `admin` boolean
    before { allow(user).to receive(:admin?).and_return(admin) }

    context 'when the user is an admin' do
      # specify authorization setting
      let(:admin) { true }

      # one liner is just as clear and avoid comment which might become stale.
      it { is_expected.to include('admin') }

      it 'ignores nil values' do
        teacher_info_mock = double
        mock_subject_areas = ['math', 'Poly Sci'].map {|x| create(:subject_area, name: x) }

        allow(teacher_info_mock).to receive(:subject_areas).and_return mock_subject_areas
        allow(user).to receive(:teacher_info).and_return(teacher_info_mock)

        expect(subject).to_match_array(['subject_area_math', 'admin', 'subject_area_poly_sci'])
      end
    end

    context 'when the user is not an admin' do
      let(:admin) { false }

      it { is_expected.not_to include('admin') } 
    end
  end
```

services/QuillLMS/app/workers/learn_worlds_integration/sync_orchestrator_worker.rb
```diff
@@ -0,0 +1,40 @@
...
+    def string_to_subject_area_tag(str)
+      "subject_area_#{str.downcase.gsub(%r{[\s/]+}, '_')}"
+    end
+
+    def tags(user)
+      subjects_taught = user&.teacher_info&.subject_areas || []
+
+      user_account_type = user.admin? ? 'admin' : 'teacher'
+
+      (subjects_taught.compact.map{ |x| string_to_subject_area_tag(x.name)} << user_account_type)
```
brendanshean: Function chaining might be more readable here:

```
subjects_taught
  .compact
  .map(&:name)
  .map(&method(:string_to_subject_area_tag))
  .append(user_account_type)
```

services/QuillLMS/app/services/learn_worlds_integration/user_tags_request.rb
```diff
@@ -0,0 +1,26 @@
+# frozen_string_literal: true
+
+module LearnWorldsIntegration
+  class UserTagsRequest < UserRequest
+
+    attr_reader :external_id, :tags
+
+    def endpoint
```
brendanshean: I would use one-liner here:
`def endpoint =  "#{USER_TAGS_ENDPOINT}/#{external_id}"`

Also, `def initialize` should be the first instance method in a class

brendanshean (approved):
Some minor refactoring suggestions, but overall this looks good. I like UserRequest refactoring.

PR #11890, opened 2024-06-05
Rename Evidence::Research::GenAI::Passage to Activity

brendanshean:
## WHAT
Rename the `Evidence::Research::GenAI::Passage` to `Evidence::Research::GenAI::Activity`

## WHY
Curriculum has [requested](https://www.notion.so/quill/DRAFT-Curriculum-Team-GenAI-Experiment-Naming-Preferences-18763e556a644230a28d7151cd281e06?pvs=4#eeb58c2bfd274ee99e9952986f92d1f0) a different model name than Passage.  Activity seems like a sensible transition since an activity has text and name. 

Also, add some new fields to activity that contain relevant_evidence for each conjunction.  Initially, these values were stored at the PassagePrompt level, but Curriculum has requested they be available for all PassagePrompts for a given Activity so I'm pulling them up a level.

## HOW
Add migration and then rename the relevant fields.

### Screenshots

### Notion Card Links

I have tested this on staging

PR Checklist | Your Answer
------------ | -------------


dandrabik (approved):
This looks good for now. Thanks for putting this together.

One small caveat, if we find ourselves adding more and more attributes of activities and prompts, we should just reach outside of the `research` namespace and connect to the actual `activity` or `prompt` (this might even be enough related info now to justify doing that). That would eliminate any info onboarding needed to experiment on new activities AND will make porting any of this code to production endpoints easier (since it will use the same connections). Something to consider for the future.

PR #11893, opened 2024-06-05
Update the sample environment to point to the stagigng CMS

samanthamjohn:
## WHAT
Added a new key to the sample env so that we can get data from the staging CMS database. 
## WHY
The default development CMS has no data, we don't use this often, so rather than seeding it we rely on the staging version. 
## HOW
Added the key to the sample env

Used it on my own setup. 

PR Checklist | Your Answer
------------ | -------------
Have you added and/or updated tests? | No, because this code is never run, you'd have to copy it into your env file to see a change


happythenewsad (approved):
🚢

emilia-friedberg (approved):

PR #11907, opened 2024-06-10
better check for valid regex and lots of consolidation

emilia-friedberg:
## WHAT
Add a check for the validity of incorrect sequences/focus points that will actually make sure the grading logic runs successfully before allowing saves to production; lots of file clean up to consolidate `ConceptSelector`, `ConceptSelectorWithCheckbox`, `IncorrectSequencesInputAndConceptSelector`, and `FocusPointsInputAndConceptSelector` into files in the `Shared` component library.

## WHY
Sometimes the curriculum team would enter regexes that were technically valid, but failed in the actual algorithmic application because of our preprocessing logic. This PR makes sure that the same logic we run in the matchers runs on the sequences before they get saved, and also cleans up a lot of duplicate code.

## HOW
1. Extract the preprocessing logic from the matchers into `generateRegexMatchList`
2. Use `generateRegexMatchList` in new `isValidFocusPointOrIncorrectSequence` function.
3. Consolidate all versions of `IncorrectSequencesInputAndConceptSelector` and `FocusPointsInputAndConceptSelector` across the different apps, as well as `ConceptSelector` and `ConceptSelectorWithCheckbox` (because they import it) into one shared file in the `Shared` bundle, then update that file to use the new `isValidFocusPointOrIncorrectSequence` function.
 
### Screenshots

### Notion Card Links
https://www.notion.so/quill/Improve-incorrect-sequence-regex-validation-so-all-incorrect-regex-is-caught-before-going-to-product-a36554025839428186d9615b04c70df6?pvs=4

https://www.notion.so/quill/Invalid-Regex-code-getting-in-d5d6bb7f16ee47e99c6a1b5467327153?pvs=4

Tested each instance of `ConceptSelector`, `ConceptSelectorWithCheckbox`, and the incorrect sequence/focus point forms on staging.

PR Checklist | Your Answer
------------ | -------------


happythenewsad (approved):
We all appreciate the grunt work of DRY'ing up `conceptSelector` et al.

One step closer to the Great Convergence 🙏

eadams17 (approved):
So much cleaner! Thanks for fixing this issue and all of the refactoring

PR #11922, opened 2024-06-12
Hints (Concept Feedback) manual translations using Gengo V1

samanthamjohn:
## WHAT
Use the Gengo API to request translations of hint descriptions (`ConceptFeedback`).  

## WHY
This is a proof of concept in beginning to translate instructions for activities in our database. 

## HOW
Created a script / model methods  to request translations via the Gengo API and then download them once they are ready. 

## How Gengo Works 
1. send an API call with one or more strings to be translated
2. receive back an order number. On Gengo's end, they have put up a group of "jobs" on their site for a human translator to do. Think of it like task rabbit or mechanical turk, or an uber driver deciding to pick you up. **This is not deterministic** we don't know when someone will pick it up or when they will finish.  
3. with the order number, we can request the list of individual jobs so that we have a reference to the translations we have asked for. 
4. Usually within 24 hours someone will pick up the "job" and translate our text. At that point, Gengo can send us a callback (we're not yet implementing this, but have a [notion card](https://www.notion.so/quill/Use-gengo-callbacks-to-update-the-translated_texts-729ee4be44dc40e4a7977f16442ea597?pvs=4) for it). 
5. Right now we have a method to manually pull the jobs--you can log into the Gengo dashboard to see the status of your order so with this workflow you'd be able to pull the translation into our database. 

### Notion Card Links
[Translation for hints](https://www.notion.so/quill/Add-translations-for-hints-descriptions-aa534571553c4666a3119be4662e8b25?pvs=4)
[RFC for translations](https://www.notion.so/quill/RFC-for-storage-of-Translation-Strings-21ce3b870e08401faf3933d6b5bfc06a?pvs=4)
[adding a sidekiq job](https://www.notion.so/quill/Use-a-sidekiq-job-to-get-translation_jobs-from-gengo-and-save-them-to-the-translated_text-table-2a87fd5b961a4f3f95e12c4c2f84e76b?pvs=4)

Have run the translations locally using the Gengo Sandbox. The staging app will also use the sandbox, so we'll have to go to production before we start using real translations. 

PR Checklist | Your Answer
------------ | -------------


services/QuillLMS/.env-sample
```diff
@@ -73,3 +73,10 @@ ORTTO_API_KEY=
 
 # BIGQUERY
 BIGQUERY_CREDENTIALS=
+
+
+# GENGO
+# Note, you can get these from the heroku quill-staging app.
+# Add single quotes around the private key to prevent $ from being interpolated.
+GENGO_PUBLIC_KEY=
```
brendanshean, line 81: Good idea adding these.

services/QuillLMS/app/models/concept_feedback.rb
```diff
@@ -37,6 +41,31 @@ def as_json(options=nil)
...
+    translated_texts.each(&:fetch_translation!)
+  end
+
+  private def create_english_text_and_return_payload
+    english = EnglishText.create(text: desc)
+    translation_mappings.create(english_text: english)
+    english.gengo_payload
+  end
+
+  private def desc
```
brendanshean: one liner seems fitting here:  `private def desc = data['description']`

services/QuillLMS/app/models/concept_feedback.rb
```diff
@@ -37,6 +41,31 @@ def as_json(options=nil)
...
+    return unless translation_mappings.empty?
+
+    english = EnglishText.find_by(text: desc)
+    return create_english_text_and_return_payload unless english.present?
+
+    translation_mappings.create(english_text: english)
+    nil
+  end
+
+  def fetch_translation!
```
brendanshean: Why not `fetch_translations!` plural?  

Also, one liner seems readable:
`def fetch_translations! = translated_texts.each(&:fetch_translation!)`

services/QuillLMS/app/models/concept_feedback.rb
```diff
@@ -37,6 +41,31 @@ def as_json(options=nil)
...
+    translated_texts.each(&:fetch_translation!)
+  end
+
+  private def create_english_text_and_return_payload
+    english = EnglishText.create(text: desc)
```
brendanshean: `english_text` seems more appropriate for variable name

services/QuillLMS/db/migrate/20240610155753_create_translation_mappings.rb
```diff
@@ -0,0 +1,13 @@
+# frozen_string_literal: true
+
+class CreateTranslationMappings < ActiveRecord::Migration[7.0]
+  def change
+    create_table :translation_mappings do |t|
+      t.integer :english_text_id, null: false
+      t.string :source_type, null: false
+      t.integer :source_id, null: false
+      t.string :source_key
```
brendanshean: What is source_key?

services/QuillLMS/app/models/concept_feedback.rb
```diff
@@ -37,6 +41,31 @@ def as_json(options=nil)
     data
   end
 
+  def queue_translation
+    return if data["description"].nil?
+    return unless translation_mappings.empty?
+
+    english = EnglishText.find_by(text: desc)
```
brendanshean: For clarity, I would set variable as `english_text`.

services/QuillLMS/app/models/concept_feedback.rb
```diff
@@ -37,6 +41,31 @@ def as_json(options=nil)
     data
   end
 
+  def queue_translation
```
brendanshean: I think `queue` here is an implementation detail and probably not needed for the method name.  Instead, something like `translate_description` or `map_description_translation`.

brendanshean (commented):
I have a few minor comments so far, but I feel like the job stuff should be in a sidekiq worker.  Let's talk offline about it.

samanthamjohn:
> I have a few minor comments so far, but I feel like the job stuff should be in a sidekiq worker. Let's talk offline about it.

Agree, I had made a [notion card](https://www.notion.so/quill/Use-a-sidekiq-job-to-get-translation_jobs-from-gengo-and-save-them-to-the-translated_text-table-2a87fd5b961a4f3f95e12c4c2f84e76b?pvs=4) for it and just added it to this PR now

services/QuillLMS/app/models/concept_feedback.rb
samanthamjohn: for the moment it's really a one-to-one relationship, but the relationship is a has_many so there *could* be more than one. I'll change it to be plural.

services/QuillLMS/db/migrate/20240610155753_create_translation_mappings.rb
samanthamjohn: It's not used yet, so I could remove it. The idea is that hints have 3 fields that need to be translated and the key would tell you which field the translation mapped to. In the process of defining this we realized that two of the fields were a little more complicated to translate (most of it would be english, except for a small part) so I made this PR just for translating the one.  I will remove it for now.

services/QuillLMS/app/models/english_text.rb
```diff
@@ -0,0 +1,70 @@
...
+      hash.merge({(job[:slug]).to_s => job })
+    end
+  end
+
+  def gengo_payload
+    {
+      type: "text",
+      body_src: text,
+      lc_src: "en",
+      lc_tgt: "es-la",
```
brendanshean: Let's extract a constant here.

services/QuillLMS/app/models/concept_feedback.rb
```diff
@@ -29,6 +29,10 @@ class ConceptFeedback < ApplicationRecord
   validates :activity_type, presence: true, inclusion: {in: TYPES}
   validate :data_must_be_hash
 
+  has_many :translation_mappings, as: :source
+  has_many :english_texts, through: :translation_mappings
+  has_many :translated_texts, through: :english_texts
+
```
dandrabik, line 36: If you add `store_accessor :data, :description` to this file, you can access `description` like a normal attribute. [Reference](https://api.rubyonrails.org/classes/ActiveRecord/Store.html)

services/QuillLMS/app/models/english_text.rb
```diff
@@ -0,0 +1,70 @@
...
+#
+class EnglishText < ApplicationRecord
+  has_many :translated_texts
+  has_many :translation_mappings
+
+  STANDARD_COMMENT = "
+  We are translating the instructions for an English-language grammar activity. The content of the activity itself is not translated. Therefore, please leave words that sound like they are part of the activity in the original english. Often they will between an HTML tag such as in <em>english word</em> or <ul>english  word</ul>.
+"
+
+  def self.translate!(jobs_list:)
```
dandrabik: As discussed, all of these class methods could be extracted as service objects. I don't think they belong in this class.

services/QuillLMS/app/models/concept_feedback.rb
dandrabik: Small style thing for method/variable names: We prefer full words over shortened ones, e.g. `description` instead of `desc`, `response` instead of `resp`. It's more characters, but we feel it makes the code more readable long-term.

services/QuillLMS/app/models/english_text.rb
```diff
@@ -0,0 +1,70 @@
...
+    resp = GengoAPI.postTranslationJobs(jobs: create_payload(jobs_list:))
+    return unless resp.present?
+
+    SaveTranslatedTextWorker
+    .perform_in(1.minute, resp.dig("response", "order_id"))
```
dandrabik: Style, indent for chained calls, e.g.

```suggestion
      .perform_in(1.minute, resp.dig("response", "order_id"))
```

services/QuillLMS/app/models/english_text.rb
```diff
@@ -0,0 +1,70 @@
...
+class EnglishText < ApplicationRecord
+  has_many :translated_texts
+  has_many :translation_mappings
+
+  STANDARD_COMMENT = "
+  We are translating the instructions for an English-language grammar activity. The content of the activity itself is not translated. Therefore, please leave words that sound like they are part of the activity in the original english. Often they will between an HTML tag such as in <em>english word</em> or <ul>english  word</ul>.
+"
+
+  def self.translate!(jobs_list:)
+    resp = GengoAPI.postTranslationJobs(jobs: create_payload(jobs_list:))
```
dandrabik: Use full words for variables, e.g. `response`

services/QuillLMS/app/models/english_text.rb
```diff
@@ -0,0 +1,70 @@
...
+    return unless resp.present?
+
+    SaveTranslatedTextWorker
+    .perform_in(1.minute, resp.dig("response", "order_id"))
+  end
+
+  def self.save_translated_text!(order_id:)
+    response = GengoAPI.getTranslationJobs({order_id:})
+    response&.dig("response")&.each do |job|
+      create_translated_text!(job_id: job["job_id"])
```
dandrabik: Any key that is used for logic in more than one place should be extracted as a constant at the top of the file, e.g.
```ruby
JOB_ID = "job_id"
...
create_translated_text!(job_id: job[JOB_ID])
```
Lots of logical strings/keys in this file to do this for.

services/QuillLMS/app/models/english_text.rb
```diff
@@ -0,0 +1,70 @@
...
+    end
+  end
+
+  def self.create_translated_text!(job_id:)
```
dandrabik: This method does 3 things. Makes a request, parses the response, and creates the `translated_text`. Consider breaking this up into smaller methods that have one responsibility.

services/QuillLMS/app/models/translated_text.rb
```diff
@@ -0,0 +1,31 @@
...
+#  created_at         :datetime         not null
+#  updated_at         :datetime         not null
+#  english_text_id    :integer          not null
+#  translation_job_id :string           not null
+#
+class TranslatedText < ApplicationRecord
+  belongs_to :english_text
+  scope :pending_translation, -> { where(translation: nil) }
+
+  def self.fetch_pending!
```
dandrabik: `fetch` is often used for pulling records, consider `translate_pending!` to make the action clearer (I was surprised to see this translate records).

services/QuillLMS/app/models/translated_text.rb
```diff
@@ -0,0 +1,31 @@
...
+    pending_translation.each(&:fetch_translation!)
+  end
+
+  def fetch_translation!
+    return unless translation.nil?
+
+    response = GengoAPI.getTranslationJob({id: translation_job_id})
+    return unless response
+
+    update_attribute(:translation, response.dig("response", "job", "body_tgt"))
```
dandrabik: Use constants for keys.

services/QuillLMS/app/models/english_text.rb
```diff
@@ -0,0 +1,70 @@
...
+#  id         :bigint           not null, primary key
+#  text       :text             not null
+#  created_at :datetime         not null
+#  updated_at :datetime         not null
+#
+class EnglishText < ApplicationRecord
+  has_many :translated_texts
+  has_many :translation_mappings
+
+  STANDARD_COMMENT = "
```
dandrabik: I think you want to use [Squiggly heredoc](https://ruby-doc.org/core-2.5.0/doc/syntax/literals_rdoc.html#label-Here+Documents) here to extract remove whitespace and indentation.., e.g.

```ruby
STANDARD_COMMENT = <<~STRING
     We are translating the instructions for an English-language grammar activity. The content of the activity itself 
     is not translated. Therefore, please leave words that sound like they are part of the activity in the original 
     english. Often they will between an HTML tag such as in <em>english word</em> or <ul>english  word</ul>.
STRING

**Review by dandrabik — CHANGES_REQUESTED** [review]
This is a great start! I flagged some refactors to tighten this up a bit and fit the style of our codebase. Ping me if you have any questions, otherwise re-request a review when you feel the code is in a good place.

File: services/QuillLMS/app/models/english_text.rb (hunk 11 above)
Comment (samanthamjohn) [reply services/QuillLMS/app/models/english_text.rb]: Updated that file, but will change it elsewhere. I think we should be able to configure rubocop to enforce that.

File: services/QuillLMS/app/models/translated_text.rb (hunk 15 above)
Comment (samanthamjohn) [reply services/QuillLMS/app/models/translated_text.rb]: It pulls the records from Gengo, so I think `fetch` is the right thing here. This method was moved to `app/services/gengo/save_translated_text.rb`

**Review by brendanshean — APPROVED** [review]
Thank you for making the updates we discussed.  This looks great.  I also really like the you added the description of how the Gengo API is used.

File: services/QuillLMS/app/models/translated_text.rb (hunk 15 above)
Comment (dandrabik) [reply services/QuillLMS/app/models/translated_text.rb]: To rephrase my original comment, as a fresh set of eyes, it was surprising to see an action called `fetch_pending` create/update records in the DB. The chain of this method goes `fetch_pending` -> `fetch_translations` -> `SaveTranslatedText`. That last `Save` is a switch of intention/naming that wasn't obvious from the first method. Consider `fetch_and_save_pending!` or something like that.

File: services/QuillLMS/app/models/translated_text.rb [hunk 18]
```diff
@@ -0,0 +1,26 @@
...
+  belongs_to :english_text
+  scope :pending_translation, -> { where(translation: nil) }
+
+  def self.fetch_pending!
+    pending_translation.each(&:fetch_translation!)
+  end
+
+  def fetch_translation!
+    Gengo::SaveTranslatedText.run(translation_job_id)
+  end
```
dandrabik: This is a new Ruby thing, so you can take or leave, but there are a few methods in this PR that could be one-liners. 

Note, if you are not used to one line methods from other languages, the new syntax takes a little getting used to. Once you start using it, though, it's pretty nice for making code shorter and more parsable.
```suggestion
  def fetch_translation! = Gengo::SaveTranslatedText.run(translation_job_id)
```

dandrabik (approved):
Overall, looks good! Thanks for making the requested changes. Approving with one request for a method name change for clarity.

Nice work on this!

PR #11929, opened 2024-06-13
Add order to payload object in Regex Input and Concept Selector form

cissyxyu:
## WHAT
Bugfix on this form -- the "order" attribute wasn't being read into the React class's state and therefore not being sent on the submission callback. This caused the "order" attribute of the object to get lost when admin use this page and then the overall order of incorrect sequences and focus points is scrambled after they submit their edits.

## WHY
The order needs to be sent in the payload in order for the overall order of sequences to stay in the same order after the admin uses this page.

## HOW
Just read the order into the initial state and then send it through on the submit action.

### Screenshots

### Notion Card Links
https://www.notion.so/quill/Incorrect-Sequence-Shuffling-1d58145f28264fc9bd2bff7fb0788845?pvs=4

Deploy to staging and test editing the incorrect sequences and focus points on Connect, Grammar and Diagnostic to determine that order does not reshuffle after editing the sequences.

PR Checklist | Your Answer
------------ | -------------


emilia-friedberg (approved):
I juuuuuust consolidating all six versions of this file into one place so I feel happy that this is now a two-line code change! Thanks for fixing this.

PR #11935, opened 2024-06-13
better LW API error handling

happythenewsad:
## WHAT
- handles case where the LW `GET suspended users` API returns a 404

## WHY
- The API has the unexpected behavior of returning a 404, rather than an empty list, when there are zero suspended users. This scenario occurred during the initial SyncOrchestrator rollout. It's unlikely to occur again, but it's worth fixing now to avoid a possible confusing error in the future. 

## HOW
- in the 0 suspended users scenario, return early with `[]`

### Screenshots

### Notion Card Links
none

Run `SuspendedUsersRequest` in staging, when the suspend list is zero. Observe the output.

PR Checklist | Your Answer
------------ | -------------

### DIFF
--- a/services/QuillLMS/app/services/learn_worlds_integration/suspended_users_request.rb
+++ b/services/QuillLMS/app/services/learn_worlds_integration/suspended_users_request.rb
@@ -10,13 +10,21 @@ module LearnWorldsIntegration
 
     def fetch_page(page_number)
       response = HTTParty.get("#{endpoint}&page=#{page_number}", headers:)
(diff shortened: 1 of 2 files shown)


anathomical (approved):
Nice catch.  Any thoughts about using status `code` instead of `message`?  They should always be in sync, so it probably doesn't matter, but in my head status code is more "stable".

Anyway, perfectly happy for things to go out as-is.

emilia-friedberg (approved):

PR #11936, opened 2024-06-13
fix route to unsubscribe from usage snapshot emails

emilia-friedberg:
## WHAT
Make it so that the link users receive to unsubscribe from usage snapshot emails works as expected.

## WHY
The link was taking users to a 404 page and failing to unsubscribe them.

## HOW
I changed the verb to GET from DELETE, since following the link was always being treated as a GET request, resulting in the 404. I couldn't figure out a way to generate the email link with a DELETE method, but I recognize that this is not the most RESTful approach, so please let me know if you have other suggestions. I'm not clear on how this worked at one point or when it broke.

### Screenshots

### Notion Card Links
https://www.notion.so/quill/Email-unsubscribe-option-for-the-Usage-Snapshot-report-is-not-working-c15257596a074f67af9b37d19aec4a92

Confirmed that following a link with this pattern no longer results in the 404 page.

PR Checklist | Your Answer
------------ | -------------

### DIFF
--- a/services/QuillLMS/client/app/bundles/Proofreader/tests/helpers/determineUnnecessaryEditType.spec.ts
+++ b/services/QuillLMS/client/app/bundles/Proofreader/tests/helpers/determineUnnecessaryEditType.spec.ts
@@ -1,4 +1,3 @@
-import * as expect from 'expect';
 import {
   MULTIPLE_UNNECESSARY_ADDITION,
   MULTIPLE_UNNECESSARY_DELETION,
--- a/services/QuillLMS/config/routes.rb
+++ b/services/QuillLMS/config/routes.rb
@@ -774,7 +774,7 @@ EmpiricalGrammar::Application.routes.draw do
     collection do
       post :create_or_update
       get :current
-      delete 'unsubscribe/:token', to: 'pdf_subscriptions#unsubscribe', as: :unsubscribe
+      get 'unsubscribe/:token', to: 'pdf_subscriptions#unsubscribe', as: :unsubscribe
     end
   end
 
--- a/services/QuillLMS/spec/controllers/pdf_subscriptions_controller_spec.rb
+++ b/services/QuillLMS/spec/controllers/pdf_subscriptions_controller_spec.rb
@@ -96,8 +96,8 @@ RSpec.describe PdfSubscriptionsController, type: :controller do
     end
   end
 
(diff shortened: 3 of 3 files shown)


happythenewsad (approved):
This is the right call. HTML forms only support GET and POST. Since we can't rely on having JS in an email environment, we cannot be RESTful.

https://www.w3.org/TR/html401/interact/forms.html#h-17.3

That said, POST is better, since GET is not expected to change state on the server.

emilia-friedberg:
> This is the right call. HTML forms only support GET and POST. Since we can't rely on having JS in an email environment, we cannot be RESTful.
> 
> https://www.w3.org/TR/html401/interact/forms.html#h-17.3
> 
> That said, POST is better, since GET is not expected to change state on the server.

@happythenewsad but I think if we're navigating to the url (ie, via the email link), it has to be a GET request, right? I can't figure out how to format the link so that we'd be sending anything else, and if the route isn't defined as a GET visiting that url returns a 404.

happythenewsad:
> > This is the right call. HTML forms only support GET and POST. Since we can't rely on having JS in an email environment, we cannot be RESTful.
> > https://www.w3.org/TR/html401/interact/forms.html#h-17.3
> > That said, POST is better, since GET is not expected to change state on the server.
> 
> @happythenewsad but I think if we're navigating to the url (ie, via the email link), it has to be a GET request, right? I can't figure out how to format the link so that we'd be sending anything else, and if the route isn't defined as a GET visiting that url returns a 404.

Good point - I think the way to do this RESTfully (if you want the user to navigate to a page, and not just click a link and stay in the email client) would be to have the initial URL be a GET, then have a unsubscribe button at the destination page that makes a POST. But in this case I suspect we both prefer simplicity over REST dogma. :-)

anathomical (approved):
Good catch.  I think that making compromises around the weird behavior of emails is just something we have to resign ourselves to.  I certainly don't know of a way to force a link in an email to use a specific HTTP method.

brendanshean (approved):
This is weird.  I definitely checked this DELETE request when I built this out and it worked.  I'm wondering if browsers tightened this practice up for security reasons.  In any event, GET seems fine for now.

If we really wanted this to be RESTful, you could have the GET request route to a landing page with a button and then have that button be the final action to delete it, but I don't think that should be a priority.

Thank you for fixing this.

PR #11941, opened 2024-06-17
Add translations to the hints admin interface

samanthamjohn:
## WHAT
In the admin interface, add a button to show translations for hints if they are available. 
## WHY
So that the admin can see how a hint will look once it's in the app. 
## HOW
Added the translations to the api for hints, and then show them if the toggle is on. 
### Screenshots

2024-06-17 at 5 49 PM
2024-06-17 at 5 49 PM

### Notion Card Links
[Show translations to admins](https://www.notion.so/quill/Show-hints-translations-in-the-admin-interface-6b3ba14ea6a6472ba9c981a45c58004a?pvs=4)

PR Checklist | Your Answer
------------ | -------------

### DIFF
--- a/services/QuillLMS/app/models/concept_feedback.rb
+++ b/services/QuillLMS/app/models/concept_feedback.rb
@@ -39,7 +39,10 @@ class ConceptFeedback < ApplicationRecord
   def cache_key = "#{ALL_CONCEPT_FEEDBACKS_KEY}_#{activity_type}"
 
   def as_json(options=nil)
-    data
+    translation = translation(locale: "es-la")
+    return data unless translation.present?
+
+    data.merge({"translatedDescription" => translation(locale: "es-la")})
   end
 
   def create_translation_mappings
@@ -50,6 +53,9 @@ class ConceptFeedback < ApplicationRecord
     translation_mappings.create(english_text: )
   end
 
+  def translation(locale:) = translated_texts.find_by(locale:)&.translation
+
+  def translate! = Gengo::RequestTranslations.run(english_texts)
   def fetch_translations! = translated_texts.each(&:fetch_translation!)
 
   private def data_must_be_hash
--- a/services/QuillLMS/client/app/bundles/Grammar/components/conceptsFeedback/conceptFeedback.tsx
+++ b/services/QuillLMS/client/app/bundles/Grammar/components/conceptsFeedback/conceptFeedback.tsx
@@ -23,9 +23,12 @@ class ConceptFeedbackComponent extends React.Component<ConceptFeedbackComponentP
 
     this.deleteConceptsFeedback = this.deleteConceptsFeedback.bind(this)
     this.toggleEdit = this.toggleEdit.bind(this)
+    this.toggleTranslation = this.toggleTranslation.bind(this)
     this.submitNewFeedback = this.submitNewFeedback.bind(this)
     this.cancelEdit = this.cancelEdit.bind(this)
     this.concept = this.concept.bind(this)
+
+    this.state = {translated: false}
   }
 
   deleteConceptsFeedback() {
@@ -41,6 +44,16 @@ class ConceptFeedbackComponent extends React.Component<ConceptFeedbackComponentP
     const conceptFeedbackID = this.props.match.params.conceptFeedbackID
     if (conceptFeedbackID) {
       this.props.dispatch(actions.startConceptsFeedbackEdit(conceptFeedbackID))
+    }
+  }
+
+  toggleTranslation() {
+    const translated = !this.state.translated
+    this.setState({...this.state, ...{translated: translated}})
+    if (translated) {
+      document.getElementById('toggle-translation').innerHTML = "Hide translation"
+    } else {
+      document.getElementById('toggle-translation').innerHTML = "Show translation"
     }
   }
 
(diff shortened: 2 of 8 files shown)


services/QuillLMS/client/app/bundles/Grammar/components/conceptsFeedback/conceptFeedback.tsx
```diff
@@ -44,6 +47,16 @@ class ConceptFeedbackComponent extends React.Component<ConceptFeedbackComponentP
     }
   }
 
+  toggleTranslation() {
+    const translated = !this.state.translated
+    this.setState({...this.state, ...{translated: translated}})
+    if (translated) {
```
samanthamjohn: I think this is not the react way, but wanted to get something working and figured we could fix it in post :)

services/QuillLMS/app/models/concept_feedback.rb
```diff
@@ -39,7 +39,10 @@ class ConceptFeedback < ApplicationRecord
   def cache_key = "#{ALL_CONCEPT_FEEDBACKS_KEY}_#{activity_type}"
 
   def as_json(options=nil)
-    data
+    translation = translation(locale: "es-la")
```
samanthamjohn, line 42: It's possible that this will slow things down a little since we're now pulling from the translation_mappings table each time we grab the as_json on the concept_feedback. Right now those tables are tiny so it shouldn't matter (plus the API call is cached), but I'm open to pulling this data another way from the get-go just in case.

services/QuillLMS/client/app/bundles/Shared/components/feedback/conceptExplanation.tsx
```diff
@@ -1,16 +1,17 @@
 import * as React from 'react'
 
-function getClassName(description, leftBox, rightBox) {
+function getClassName(description, leftBox, rightBox, translatedDescription) {
   if (description && leftBox && rightBox) {
```
eadams17: do we need to be passing `translatedDescription` here or can we just remove that from the function arguments?

services/QuillLMS/client/app/bundles/Grammar/components/conceptsFeedback/conceptFeedback.tsx
eadams17: so I think all we are worried about is toggling the text change on button click, right? If so, the following would be preferable:

```
toggleTranslation() {
  const { translated } = this.state
  this.setState({ translated: !translated })
}

renderTranslationButton(translatedDescription) {
  const { translated } = this.state
  if(translatedDescription) {
    const buttonText = translated ? "Hide translation" : "Show translation"
    return <button className="button is-info" id='toggle-translation' onClick={this.toggleTranslation}>{buttonText}</button>
  }
}

return {
  const { translatedDescription } = data[conceptFeedbackID]
  ...
  <button className="button is-info" onClick={this.toggleEdit}>Edit Feedback</button>
  <button className="button is-danger" onClick={this.deleteConceptsFeedback}>Delete Concept Feedback</button>
  {renderTranslationButton(translatedDescription)}
}
```

I know a lot of these older files are not quite uniform, but we try to use desctructuring whenever possible, especially for props and state. (Also, we don't need to use spread for state, we can just update an individual piece as written above)

eadams17 (changes requested):
Thanks for putting this together-- this is a great first pass! I added some comments on some refactoring suggestions.

services/QuillLMS/client/app/bundles/Shared/styles/conceptFeedback.scss
```diff
@@ -1,3 +1,10 @@
+.control {
+  margin-top: 10px;
+}
+
+.control button {
+  margin-right: 10px;
+}
```
emilia-friedberg: This is going to change the CSS for anywhere we use the `.control` class -- maybe we can nest this rule inside of `.concept-explanation`?

services/QuillLMS/client/app/bundles/Grammar/components/conceptsFeedback/conceptFeedback.tsx
emilia-friedberg: I agree with Eric's proposal here for this and for the corresponding file in Connect. One note: for the `toggleTranslation` function, you can actually just do:

``` 
toggleTranslation() {
  this.setState(prevState => { translated: !prevState.translated }) }
```

emilia-friedberg (changes requested):
Very close! I agree with Eric's comments, and added one more note about scoping a CSS rule.

services/QuillLMS/client/app/bundles/Shared/components/feedback/conceptExplanation.tsx
samanthamjohn: good catch!

services/QuillLMS/app/models/concept_feedback.rb
anathomical: I think that your point about the tables being small, but especially the one about us caching the output of this call, means that the extra overhead is totally reasonable.

emilia-friedberg (approved):
LGTM!

eadams17 (approved):
Thanks for making those changes!

PR #11943, opened 2024-06-18
Use a constant for SPANISH_LOCALE (es-la) everywhere

samanthamjohn:
## WHAT
Code cleanup to use es-la everywhere in the app as a constant 
## WHY
So we don't type it in all the time which is error prone.
## HOW
Added it to the gengo.rb initializer

PR Checklist | Your Answer
------------ | -------------
Have you added and/or updated tests? | Yes, updated test to use the new constant


anathomical (approved):
Always love constants which let us avoid accidentally typoing something and breaking it.

PR #11944, opened 2024-06-18
Don’t re-translate the same english_text

samanthamjohn:
## WHAT
If we already have a translation pending for an english text, don't re-send it to Gengo
## WHY
If we have a bunch of translated text records for the same word, our data gets messy and it's hard to see why the translation isn't showing up in the interface if say, one translated text has pulled its translation and another has not. 
## HOW
Put a guard on the `RequestTranslations` service to check the `translated_texts` table before sending the request to gengo

### Notion Card Links
https://www.notion.so/quill/Don-t-re-translate-an-already-translated-english_text-9e5ea315395b4811b38c5607dfb7a62f?pvs=4

PR Checklist | Your Answer
------------ | -------------


services/QuillLMS/app/models/english_text.rb
```diff
@@ -12,4 +12,8 @@
 class EnglishText < ApplicationRecord
   has_many :translated_texts
   has_many :translation_mappings
+
+  def needs_translation(locale: Gengo::SPANISH_LOCALE)
```
brendanshean: Since this is returning a boolean, rename `needs_translation?`

services/QuillLMS/app/services/gengo/request_translations.rb
```diff
@@ -24,7 +26,9 @@ def run
     end
 
     def gengo_payload
-      english_texts.reduce({}) do |hash, english_text|
```
brendanshean, line 27: `each_with_object` is a nice ruby pattern for reduce/merge; also, begin/end block will help with getting formatting with memoized variable:

```
   @gengo_payload ||= begin
      english_texts
        .select(&:needs_translation?)
        .each_with_object({}) do |english_text, hash|
          hash[english_text.id.to_s] = {
            type: "text",
            body_src: english_text.text,
            lc_src: "en",
            lc_tgt: Gengo::SPANISH_LOCALE,
            tier: "standard",
            slug: english_text.id,
            group: true,
            auto_approve: true,
            comment: STANDARD_COMMENT
          }
      end
    end
```

services/QuillLMS/app/services/gengo/request_translations.rb
brendanshean, line 1: I noticed in this file:
`attr_accessor :english_texts`

Should this be `attr_reader :english_text`?

services/QuillLMS/spec/models/english_text_spec.rb
```diff
@@ -16,4 +16,32 @@
...
+
+      it "defaults to Gengo::SPANISH_LOCALE" do
+        expect(english_text.needs_translation).to be false
+      end
+    end
+
+    context 'there is not a translated_text record with that locale' do
+      let(:other_locale) { "jp" }
+
+      it { expect(subject).to be true}
```
brendanshean: Consider `it { is_expected.to be true }`

services/QuillLMS/spec/models/english_text_spec.rb
```diff
@@ -16,4 +16,32 @@
...
+    let!(:translated_text) {
+      create(:translated_text,
+      english_text: english_text,
+      locale: other_locale)
+    }
+
+    context 'there is a translated_text record associated with that locale' do
+      let(:other_locale) { locale }
+
+      it { expect(subject).to be false}
```
brendanshean: Consider `it { is_expected.to be false }`

brendanshean (approved):
Some minor stylistic comments, but otherwise good to go.

PR #11949, opened 2024-06-18
Code clean up for concept feedback and delete them from diagnostic 

samanthamjohn:
## WHAT
- moved the `FeedbackForm` to edit concept feedback to a shared component that both grammar and connect can use
- fixed console error in the `FeedbackForm`
- Deleted `conceptFeedback` from diagnostic in the admin interface since we do not use it there. 

## WHY
- this will make our code easier to maintain as we had the same thing duplicated in three places, and now it is in one places. 
- fixing the console error makes it easier to debug future features. 

## HOW
- deleted unused code and combined code that was the same 

Checked that the diagnostics still worked. Checked that editing a conceptFeedback works properly

PR Checklist | Your Answer
------------ | -------------
Have you added and/or updated tests? | No, all frontend 


emilia-friedberg (approved):
LGTM! Thank you for this cleanup!

PR #11963, opened 2024-06-20
GenAI Reorg Part 8:  Update Trial and LLMPrompt schemas

brendanshean:
## WHAT
1.  Remove num_examples from `Trials`
1.  Add `dataset_id` to `Trials`
1.  Add `version` to `Dataset`
1.  Change `LLMFeedback` to `LLMExample`
1.  Change `text` in `LLMExample` to `llm_feedback`
1.  Add `llm_assigned_status` to `LLMExample`
1.  Rename `category` in `Guideline` to `staff_assigned_status`
1.  Add DatasetsController
1.  Simplify `TrialsController` to use dropdowns instead of checkboxes for `LLM`, `LLMPromptTemplate` and `GEval` 
1.  Add `{suboptimal, optimal}_{guideline, examples}_count` to LLMPrompt
1.  Add HasAssignedStatus
1.  Connect `DatasetImporter` with `DatasetsController#create`
1.  Update `llm.create_from_template`
1.  Add locks to `LLMPromptGuideline` and `LLMPromptPromptExample`
1.  Remove StudentResponse model
1.   Break `Resolver` up into `LLMFeedbackResolver` and `LLMAssignedStatusResolver`
1.  Update `LLMPromptBuilder` to incorporate guidelines and prompt examples

## WHY
1.  This number was used to initially use a sample of a dataset.  Now that we can upload small datasets its no longer needed
1.  Trials are now linked to Datasets
1.  We need a way to distinguish datasets within a StemVault
1.  This naming is consistent with `TestExample` and `PromptExample`
1.  `LLMFeedback` has multiple text fields.
1.  We know store the optimal boolean value for each `LLMExample`
1.  This naming is consistent with other fields using optimal: true|false
1.  Used to create Datasets
1.  Dropdowns will simplify the UX for future users
1.  These values will allow quick scanning of different variables chosen for a prompt
1.  This concern collects the optimal/suboptimal designation that is applied to `Guideline`, `{Prompt, Test, LLM}Example` models
1.  DataImporter was initially used to import CSV via command-line but it is now done through controller
1.  When creating a prompt, we also need to take into account which guidelines and prompt examples were selected and then create the requisite join records
1.  Once a LLMPrompt is created, we want to prevent additional Guidelines and PromptExamples from being attached to it. 
1.  StudentResponse is now bundled with `TestExample`
1.  Having two Resolvers helps to manage the separate tasks of parsing out the LLMFeedback as well as the [Sub]Optimal status
1.  New substitution rules need to be added to handle Guidelines and PromptExamples selected for inclusion in a LLMPrompt

### Screenshots

### Notion Card Links
https://www.notion.so/quill/Internal-Tool-Organizing-Trials-into-Datasets-8dbc188caf194157a2a28395e907ddc0?pvs=4

Tested on staging.

PR Checklist | Your Answer
------------ | -------------


services/QuillLMS/db/migrate/20240620123600_add_various_fields_to_trial_and_llm_prompt.evidence.rb
```diff
@@ -0,0 +1,29 @@
+# frozen_string_literal: true
+
+# This migration comes from evidence (originally 20240620122344)
+class AddVariousFieldsToTrialAndLLMPrompt < ActiveRecord::Migration[7.0]
+  def change
+    remove_column :evidence_research_gen_ai_trials, :num_examples, :integer
+    rename_column :evidence_research_gen_ai_trials, :stem_vault_id, :dataset_id
```
brendanshean: This a hack because stem_vault_id points to a fundamentally different object than dataset_id.  This change is about injecting a new parent relationship.

Old chain: StemVault -> Trial 
New chain: StemVault -> Dataset -> Trial

services/QuillLMS/engines/evidence/app/controllers/evidence/research/gen_ai/datasets_controller.rb
```diff
@@ -0,0 +1,38 @@
...
+        def show
+          @dataset = Dataset.find(params[:id])
+          @trials = @dataset.trials.order(id: :desc)
+        end
+
+        private def dataset_params
+          params
+            .require(:research_gen_ai_dataset)
+            .permit(:file)
+            .merge(optimal_count: 0, suboptimal_count: 0, locked: false)
```
dandrabik: Take or leave. Maybe these are better as database or model defaults rather than injected params.

services/QuillLMS/engines/evidence/app/controllers/evidence/research/gen_ai/llm_prompts_controller.rb
```diff
@@ -4,7 +4,10 @@ module Evidence
   module Research
     module GenAI
       class LLMPromptsController < ApplicationController
-        def show = @llm_prompt = LLMPrompt.find(params[:id])
+        def show
+          @llm_prompt = LLMPrompt.find(params[:id])
+          @dataset = @llm_prompt.trial.dataset
```
dandrabik: Could `trial` be `nil` here? Consider using `&.`

services/QuillLMS/engines/evidence/app/controllers/evidence/research/gen_ai/trials_controller.rb
```diff
@@ -6,84 +6,57 @@ module GenAI
...
-        end
-
         def new
-          @trial = Trial.new
+          @trial = dataset.trials.new
           @llms = LLM.all
-          @stem_vaults = StemVault.all
           @llm_prompt_templates = LLMPromptTemplate.all
           @g_evals = GEval.selectable
+          @guidelines = dataset.stem_vault.guidelines
```
dandrabik, line 14: Could stem_vault be nil here? Might be worth doing a `nil` audit for these nested relationships.

services/QuillLMS/engines/evidence/app/views/evidence/research/gen_ai/datasets/show.html.erb
```diff
@@ -0,0 +1,55 @@
...
+          <th>Actions</th>
+        </tr>
+      </thead>
+      <tbody>
+        <% @trials.each do |trial| %>
+          <tr>
+            <td><%= trial.id %></td>
+            <td><%= trial.created_at.strftime("%m/%d/%y, %H:%M") %></td>
+            <td><%= trial.llm %></td>
+            <td><%= trial.stem_vault.name %></td>
```
dandrabik: nil check?

dandrabik (approved):
Small note about doing a sweep for potential nil errors, but otherwise, LGTM. Thanks for all of your hard work on this!

services/QuillLMS/engines/evidence/app/controllers/evidence/research/gen_ai/trials_controller.rb
brendanshean: dataset has a `belongs_to  :stem_vault`, so this one should not require safe navigation.

services/QuillLMS/engines/evidence/app/views/evidence/research/gen_ai/datasets/show.html.erb
brendanshean: trial belongs_to :dataset, and dataset belongs_to :stem_vault so these should not require safe navigation.

PR #11966, opened 2024-06-21
nces sync updates

emilia-friedberg:
## WHAT
Overhaul our NCES import rake task to use a new API, rather than having to manually download and upload CSVs. Also add a new attribute to the schools table: direct certification. 

## WHY
This was a request from Peter Gault and the Partnerships team -- there were some inaccuracies in our data, so we needed to do a fresh import anyway (and hoping to get one scheduled more regularly going forward), and this new API makes that process simpler as well as offering various other data fields we may make use of in the future.

## HOW
Rewrite the rake task to use the API, then move that code out into workers because I did hit a timeout on the successive school imports (there are 11 pages of data with 10k records each). 

### Screenshots

### Notion Card Links
https://www.notion.so/quill/Data-Project-for-Quill-DB-Vitally-Use-New-Source-of-Data-For-Free-Reduced-Lunch-Status-of-School-85d1a6ae6f904b39b02ef40a9c86c3c7?pvs=4

https://www.notion.so/quill/District-Data-Issues-Districts-with-Same-Name-Treated-as-One-District-in-Quill-CMS-Need-To-Be-Uni-833039225648442bb24190b1f3dd709c?pvs=4

I reviewed the data on staging, and asked Shannon to as well to catch any issues before we run it on production.

PR Checklist | Your Answer
------------ | -------------


services/QuillLMS/app/workers/sync_district_data_from_url_worker.rb
```diff
@@ -0,0 +1,38 @@
...
+        zipcode: district_data["zip_location"],
+        phone: district_data["phone"],
+        total_schools: district_data["number_of_schools"],
+        total_students: district_data["enrollment"],
+        grade_range: "#{district_data['lowest_grade_offered']} - #{district_data['highest_grade_offered']}"
+      }
+
+      district = District.find_by(nces_id: attributes_hash[:nces_id])
+
+      if district.present?
```
brendanshean: Ternary seems readable here:

```district.present? ? district.update!(attributes_hash) : District.create!(attributes_hash)```

File: services/QuillLMS/app/workers/sync_school_data_from_url_worker.rb [hunk 2]
```diff
@@ -0,0 +1,59 @@
...
+
+    schools_data = response['results']
+
+    schools_data.each do |school_data|
+      attributes_hash = build_attribute_hash(school_data)
+
+      district = District.find_by(nces_id: school_data['leaid'].to_i)
+      attributes_hash[:district_id] = district&.id
+
+      # we have to do this complicated search because historically nces ids, which are largely but not always numerals stored as strings in our database, have sometimes been stored with leading zeroes and sometimes without
```
brendanshean, line 19: You could extract a constant out for the SQL query:
`NCES_ID_BLAH_QUERY = "CAST(CASE WHEN nces_id ~ '^[0-9]+$' THEN nces_id ELSE '0' END AS BIGINT) = ?"`

Then some refactoring here: (I think `find_by` should work here in place of `.where(...).first`)
```
nces_id = attributes_hash[:nces_id]
school = School.find_by(nces_id:) || School.find_by(NCES_ID_BLAH_QUERY, nces_id) 
school.present? ? school.update!(attributes_hash.except(:name)) : School.create!(attributes_hash)
```

brendanshean (approved):
Nice work on figuring out how to update the syncing of these ids.

anathomical (approved):
Super happy to see the sync logic factored out into workers.  This will make it way easier to hook up regular syncs to cron jobs in the future.

One minor thought, that I'm not sure is worth dealing with at the moment, is that it might be worth making each of these workers into two workers: one to pull the data from the API and parse each district/school, and then a second one to update each district/school which is enqueued from the first job.  This follows an existing pattern that we use sometimes, and also means that if there's somehow an error updating some specific record it doesn't crash the job and prevent any other records from being updated.

I don't think we need to address this now, though.

PR #11969, opened 2024-06-24
add view and button for admin usage pdf snapshot unsubscribe

emilia-friedberg:
## WHAT
Update how unsubscribing from the Admin Usage Snapshot PDF works by adding a button to the page that calls the DELETE action.

## WHY
I tried to fix this by just turning the whole thing into a GET request, but then we were getting [this Sentry error](https://quillorg-5s.sentry.io/issues/5512423280/?project=11238&query=is%3Aunresolved+issue.priority%3A%5Bhigh%2C+medium%5D&referrer=issue-stream&stream_index=4) on production. 

## HOW
Update the logic for the #unsubscribe and #destroy controller actions so that unsubscribe doesn't actually try to write to the database, just renders a page with a button that will link to destroy, which in turn handles HTML requests by destroying the subscription if it exists and redirecting with a notice. Note: the JSON version of this endpoint will error if the user is not the user associated with the subscription, while the HTML version will not. This preserves the previously-implemented logic of the unsubscribe endpoint, and since the idea is that we get here via a link from an email, I think it makes the UX of unsubscribing less cumbersome (I hate it when websites make me log in in order to unsubscribe). However, if people feel strongly, I can change this logic.

### Screenshots
Screenshot 2024-06-24 at 9 12 47 AM

### Notion Card Links
https://www.notion.so/quill/Email-unsubscribe-option-for-the-Usage-Snapshot-report-is-not-working-c15257596a074f67af9b37d19aec4a92?pvs=4

Deployed to staging and tested both HTM and JSON versions of this endpoint. Also having Peter Sharkey check it out.

PR Checklist | Your Answer
------------ | -------------

### DIFF
--- a/services/QuillLMS/app/controllers/pdf_subscriptions_controller.rb
+++ b/services/QuillLMS/app/controllers/pdf_subscriptions_controller.rb
@@ -47,9 +47,7 @@ class PdfSubscriptionsController < ApplicationController
   def unsubscribe
     @pdf_subscription = PdfSubscription.find_by(token:)
 
-    if @pdf_subscription.nil?
-      redirect_to root_path, flash: { error: 'Subscription not found' }
-    end
+    redirect_to root_path, flash: { error: 'Subscription not found' } if @pdf_subscription.nil?
   end
 
   private def admin_report_filter_selection_id = pdf_subscription_params[:admin_report_filter_selection_id]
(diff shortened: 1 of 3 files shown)


brendanshean (approved):
This was the code I should have written the first time 😄 

Since the unsubscribe email link contains a token, I would agree with your sentiment of not requiring the user to have to login.

happythenewsad (approved):

anathomical (approved):

PR #11970, opened 2024-06-24
Lms/restrict admin reporting to subscribed schools

anathomical:
## WHAT
- Add a new scope to the `School` model to select only schools with active subscriptions (either direct or through a District)
- Add a new method on the `User` model to find all schools that a user administers that are actively premium
- Use the new `User` method in the `SnapshotsController#options` when determining which schools to show in the admin's filters
## WHY
Admins can have a mix of premium and non-premium schools that they administer, and we only want admins to be able to access premium reports (like Snapshots and Admin Diagnostic Reports) for schools that have active subscriptions.
## HOW
- Add a new scope to `School` that joins and filters for non-expired Subscriptions attached to the school either directly or through a District.
- Use the new scope when calculating which Schools to use to populate the Admin Report Filters.

### Notion Card Links
https://www.notion.so/quill/Only-allow-admins-access-to-the-new-admin-reports-for-schools-which-have-a-premium-subscription-199992ab5853422381fdcf540939931a?pvs=4

- Deploy to staging
- Find a known admin with premium schools, give them administrator privileges for a school that does not have premium (there are no known admins who this particular edge case currently impacts, but that could easily change after contract rollover into the new school year)
- Confirm that their admin filters now only contain premium schools that they admin

PR Checklist | Your Answer
------------ | -------------


services/QuillLMS/app/models/school.rb
```diff
@@ -72,6 +72,15 @@ class School < ApplicationRecord
            :lower_grade_greater_than_upper_grade
   validates :zipcode, length: { minimum: 5 }, allow_blank: true
 
+  # Lambda has to be wrapped in parens to avoid syntax error in this construction.
+  # Also, this is slightly magical:
+  scope :premium, (lambda do
```
happythenewsad, line 77: [Optional - style]

This syntax looks cleaner to me:
```ruby
premium_lambda =  lambda do
    left_outer_joins(:subscriptions)
      .left_outer_joins(district: :subscriptions)
      # Below is a slightly weird construction, but Rails magic doesn't work so we had to get explicit
      .where('(subscriptions.expiration >= ? OR subscriptions_districts.expiration >= ?)', DateTime.current, DateTime.current)
  end

  scope :premium, premium_lambda```

**Review by happythenewsad — APPROVED** [review]

File: services/QuillLMS/app/models/school.rb (hunk 1 above)
Comment (brendanshean) [reply services/QuillLMS/app/models/school.rb:77]: I think you could rewrite with Rails methods only:

```
scope :premium, (lambda do
  left_outer_joins(:subscriptions)
    .left_outer_joins(district: :subscriptions)
    .where(
      subscriptions: { expiration: DateTime.current.. },
    )
    .or(
      left_outer_joins(:subscriptions)
        .left_outer_joins(district: :subscriptions)
        .where(districts: { subscriptions: { expiration: DateTime.current.. } })
    )
end)
```

Not sure if that's more readable though.

File: services/QuillLMS/spec/models/school_spec.rb [hunk 2]
```diff
@@ -58,6 +58,28 @@
 
   it_behaves_like 'a subscriber'
 
+  context 'scopes' do
+    describe 'premium' do
+      subject { School.premium.all }
```
brendanshean: Is `.all` needed here?

brendanshean (approved):
Looks good to me.

PR #11973, opened 2024-06-25
Set role selected at sign up when creating TeacherInfo record

cissyxyu:
## WHAT
We've been creating new `TeacherInfo` records without the "`role_selected_at_sign_up`" attribute, because of a change last year that accidentally began omitting this field when we switched to the account creation callback flow.

## WHY
This attribute is useful for admin to see when viewing a teacher's behavior and we should have it filled in.

## HOW
Since only the controllers have access to the session object, I created a virtual attribute `role_selected_at_sign_up` on the User object to store this field temporarily until the after-save callback creates the `TeacherInfo` object. We can't use the "User.role" field for this because it disallows the `individual-contributor` option, which we want to allow because we want to differentiate teachers, admins, and individual contributors.

### Screenshots

### Notion Card Links
https://www.notion.so/quill/Role-selected-on-signup-mostly-blank-in-teacher_infos-eba4f00c2d4547b18b6c640fa4849d63?pvs=4

Deployed to staging and tested the sign up flow for teachers, admins, individual contributors, and students to make sure the TeacherInfo role field was set correctly for all four (students should not have a record created at all).

PR Checklist | Your Answer
------------ | -------------

### DIFF
--- a/services/QuillLMS/app/models/user.rb
+++ b/services/QuillLMS/app/models/user.rb
@@ -125,6 +125,8 @@ class User < ApplicationRecord
     :skip_capitalize_names_callback,
     :validate_username
 
+  attribute :role_selected_at_signup, :string
+
   has_secure_password validations: false
   has_one :admin_info, dependent: :destroy
   has_one :auth_credential, dependent: :destroy
@@ -969,7 +971,7 @@ class User < ApplicationRecord
 
   private def generate_default_teacher_info
     # in addition to setting the notification_email_frequency here, show_students_exact_score is also being set to true automatically on creation
-    create_teacher_info(notification_email_frequency: TeacherInfo::DAILY_EMAIL)
+    create_teacher_info(notification_email_frequency: TeacherInfo::DAILY_EMAIL, role_selected_at_signup: role_selected_at_signup)
   end
 
   private def generate_default_teacher_notification_settings
--- a/services/QuillLMS/app/models/user.rb
+++ b/services/QuillLMS/app/models/user.rb
@@ -971,7 +971,7 @@ class User < ApplicationRecord
 
   private def generate_default_teacher_info
     # in addition to setting the notification_email_frequency here, show_students_exact_score is also being set to true automatically on creation
-    create_teacher_info(notification_email_frequency: TeacherInfo::DAILY_EMAIL, role_selected_at_signup: role_selected_at_signup)
+    create_teacher_info(notification_email_frequency: TeacherInfo::DAILY_EMAIL, role_selected_at_signup:)
   end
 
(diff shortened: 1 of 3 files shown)


services/QuillLMS/app/models/user.rb
```diff
@@ -969,7 +971,7 @@ def google_student_set_password?
 
   private def generate_default_teacher_info
     # in addition to setting the notification_email_frequency here, show_students_exact_score is also being set to true automatically on creation
-    create_teacher_info(notification_email_frequency: TeacherInfo::DAILY_EMAIL)
+    create_teacher_info(notification_email_frequency: TeacherInfo::DAILY_EMAIL, role_selected_at_signup: role_selected_at_signup)
```
emilia-friedberg: This is a good use case for the syntax where you just do `role_selected_at_signup:)`

emilia-friedberg (approved):
Thanks for getting to the bottom of this!

services/QuillLMS/spec/models/user_spec.rb
```diff
@@ -2007,12 +2007,21 @@
...
-    it 'should create new TeacherInfo record' do
-      teacher = build(:teacher)
+    context 'should create new TeacherInfo record' do
+
+      it 'should create a record' do
+        teacher = build(:teacher)
+
+        expect do
+          teacher.save
+        end.to change(TeacherInfo, :count).by(1)
```
anathomical: This is purely stylistic, but I'd be tempted to one-line this:
```
expect{teacher.save}.to change(TeacherInfo, :count).by(1)
```

services/QuillLMS/spec/models/user_spec.rb
```diff
@@ -2007,12 +2007,21 @@
...
+      end
+
+      it 'should save the role_selected_at_signup attribute to TeacherInfo record' do
+        teacher = create(:teacher, role_selected_at_signup: 'admin')
+        expect(teacher.teacher_info.role_selected_at_signup).to eq(teacher.role_selected_at_signup)
+      end
```
anathomical: I think there's some (very vague) risk of this not behaving as expected.  If the `create` call was somehow bugged and set `role_selected_at_signup` on the model to `nil`, this could pass even if it's not doing what we want.  I think I'd set the and check the constant directly:
```ruby
let(:role_selected_at_signup) { 'admin' }
let(:teacher) { create(:teacher, role_selected_at_signup:)

it { expect(teacher.teacher_info.role_selected_at_signup).to eq(role_selected_at_signup) }
```

anathomical (approved):
One minor flag for a potential edge case in one of the specs, but nice find and I think this is a good way to solve the issue.

PR #11979, opened 2024-06-26
fill in blank input text alignment

emilia-friedberg:
## WHAT
Always show centered text in fill in blank inputs.

## WHY
Rachel requested this change.

## HOW
Rename the `fillInBlankInputWidth` file and function to `fillInBlankInputStyle` and update it so that the input text has extra properties applied when it has not yet started to expand the input itself.  These styles can't be applied to the input in general, because when it is actively expanding things get flickery if it's center-justified. I also added an extra test-only function to allow me to test different `offsetWidths`, because I wasn't able to stub that to be different for different elements.

### Notion Card Links
https://www.notion.so/quill/Off-centered-fill-in-the-blank-414cabf160ea4ed7829d053af67cfd5a?pvs=4

Tested on various questions.

PR Checklist | Your Answer
------------ | -------------


happythenewsad (changes requested):
I don't think increased testability is justification for adding extra params/logic to non-test functions. The tests are nice though. Can we not use spyOn or mock?

eadams17 (approved):
LGTM! And I'm fine with the extra params for testing as we already follow this pattern in other areas and it's been pretty useful for certain cases (i.e. https://github.com/empirical-org/Empirical-Core/blob/develop/services/QuillLMS/client/app/bundles/PremiumHub/containers/diagnosticGrowthReports/overviewSection.tsx#L127)

emilia-friedberg:
> I don't think increased testability is justification for adding extra params/logic to non-test functions. The tests are nice though. Can we not use spyOn or mock?

@happythenewsad unfortunately, I spent several hours yesterday trying to get mocks and spies to work here and wasn't able to come up with a successful solution. I could mock it so that `offsetWidth` always returns a given value, but not a different one for different elements, which this logic relies on. I wasn't able to get anything that worked out of ChatGPT, either. Eric is right that we use a similar strategy sometimes in React components, where we pass in props that are only used in tests to get around the fact that you can't really stub hooks in tests. I think our choices here are to follow this pattern or leave this code untested -- let me know what you think is best.

happythenewsad:
> > I don't think increased testability is justification for adding extra params/logic to non-test functions. The tests are nice though. Can we not use spyOn or mock?
> 
> @happythenewsad unfortunately, I spent several hours yesterday trying to get mocks and spies to work here and wasn't able to come up with a successful solution. I could mock it so that `offsetWidth` always returns a given value, but not a different one for different elements, which this logic relies on. I wasn't able to get anything that worked out of ChatGPT, either. Eric is right that we use a similar strategy sometimes in React components, where we pass in props that are only used in tests to get around the fact that you can't really stub hooks in tests. I think our choices here are to follow this pattern or leave this code untested -- let me know what you think is best.

Got it... because jest can't support mocking an inner function that is implemented in the same module as the outer function. 

My preference is to either move getOffsetWidth into a helper/utils module, so that it is testable, or to not include tests that require parameterizing getOffsetWidth. 

Reasoning:
1. using `isTest` parameters in prod code is an antipattern, even if we have done so previously
2. getOffsetWidth is simple enough that I don't feel the need for testing it
3. If you really want to test it, there is a way (separate module) 

Just re-request with whatever option you go with and I'll approve. I'm glad we talked it out - it's a good reminder of a very particular jest blind spot

emilia-friedberg:
@happythenewsad yeah, good call about moving into a different function. I don't care about `offsetWidth` itself but I did want to be able to test how things changed based on its outcome, so this new setup solves that. Thank you for thinking this through with me!

happythenewsad (approved):
🚢  Nice work!

PR #9288, opened 2022-06-21
Wrap query methods in Arel.sql

brendanshean:
## WHAT
For some `order` queries that use raw sql with values that aren't attributes on the model, wrap in Arel.sql.

## WHY
Methods that use raw sql for certain methods will be disallowed in [Rails 6](https://github.com/rails/rails/pull/27947)

## HOW
Locate offending queries and wrap in `Arel.sql`

### Screenshots
(If applicable. Also, please censor any sensitive data)

### Notion Card Links
(Please provide links to any relevant Notion card(s) relevant to this PR.)

PR Checklist | Your Answer
------------ | -------------
Have you added and/or updated tests? |  No
Have you deployed to Staging? | Not yet! Deploying now. 
Self-Review: Have you done an initial self-review of the code below on Github? | YES
Spec Review: Have you reviewed the spec and ensured this meets requirements and/or matches design mockups? | N/A

PR #11981, opened 2024-06-26
Add OpenAI translation option

samanthamjohn:
- co-authored by @anathomical 

## WHAT
- Added a way to request translations via the OpenAI API.

## WHY
- OpenAI is much faster and cheaper than Gengo and could potentially be an option for getting translations out to all 16 languages. 

## HOW
- Added an option to request OpenAI translations.
- Default to showing the OpenAI translations in the hints admin interface. 
- The `translated_texts` table now can have multiple translations for the same word from different sources. 
- Added a new table to store Gengo job_ids, and use its corresponding model to fetch job info from Gengo. The `TranslatedText` is now more neutral. 

### Notion Card Links
[Allow the app to use either OpenAI or Gengo](https://www.notion.so/quill/Allow-the-app-to-use-either-OpenAI-or-Gengo-for-translations-d7d57c77892540d4bf0d486b08e18c0e?pvs=4)

Have tested the feature locally in the admin interface by running many of the hints through the OpenAI translation

PR Checklist | Your Answer
------------ | -------------


samanthamjohn:
Note, this is currently not on staging because a few more branches have been deployed on top of it.

services/QuillLMS/app/services/gengo/request_translations.rb
```diff
@@ -28,7 +29,7 @@ def run
     def gengo_payload
       @gengo_payload ||= begin
         english_texts
-          .select(&:needs_translation?)
+          .select{|e| e.needs_gengo_translation?(locale:)}
```
brendanshean: I noticed the formatting on this method needs to be adjusted.  Basically the ends should look weird since there's a block within the chain:

```
    def gengo_payload
      @gengo_payload ||= begin
        english_texts
          .select{|e| e.needs_gengo_translation?(locale:)}
          .each_with_object({}) do |english_text, hash|
            hash[english_text.id.to_s] = {
              type: "text",
              body_src: english_text.text,
              lc_src: "en",
              lc_tgt: Gengo::SPANISH_LOCALE,
              tier: "standard",
              slug: english_text.id,
              group: true,
              auto_approve: true,
              comment: STANDARD_COMMENT
            }
          end
      end
    end
```

services/QuillLMS/app/models/concept_feedback.rb
```diff
@@ -32,17 +32,20 @@ class ConceptFeedback < ApplicationRecord
...
   has_many :english_texts, through: :translation_mappings
   has_many :translated_texts, through: :english_texts
+  has_many :gengo_jobs, through: :english_texts
   store_accessor :data, :description
 
   after_commit :clear_concept_feedbacks_cache
 
   def cache_key = "#{ALL_CONCEPT_FEEDBACKS_KEY}_#{activity_type}"
 
   def as_json(options=nil)
```
anathomical, line 42: I recognize that the signature here with `options=nil` isn't part of this PR, but I think this would be cleaner if you just made this take `def as_json(source_api: TranslatedText::OPEN_AI_SOURCE)` rather than dealing with `dig` when there's only one option available to be set.

services/QuillLMS/app/models/concept_feedback.rb
```diff
@@ -53,10 +56,23 @@ def create_translation_mappings
     translation_mappings.create(english_text: )
   end
 
-  def translation(locale:) = translated_texts.find_by(locale:)&.translation
+  def translation(locale:, source_api: TranslatedText::OPEN_AI_SOURCE)
+    translations(locale:, source_api:)&.first&.translation
+  end
+
+  private def translations(locale:, source_api:)
+    translated_texts.where(locale:).ordered_by_source_api(source_api)
```
anathomical, line 63: Ordering by `source_api` feels awkward to me here.  I believe that what you're trying to account for is a preference for a specific `source_api` (such as OpenAI), but with built-in allowance for a fallback to any other source if the preferred source isn't available.

This is a reasonable concern, but I think that it might be beyond the scope of what we actually need at this point.  After all, all of our initial testing looks like it's going to be with OpenAI, anyway.  I think I'd prefer to skip this, or if we really need it just replace it with `where(source_api:)` or something.  That allows us to defer figuring out how we want to handle fallback stuff until we actually have a real use-case for it and thus a better idea of the best approach.

services/QuillLMS/app/services/open_ai/save_translated_text.rb
```diff
@@ -0,0 +1,28 @@
+# frozen_string_literal: true
+
+module OpenAI
+  class SaveTranslatedText < ApplicationService
+
+    class OpenAITranslationError < StandardError; end
+    attr_accessor :english_text
+
+    def initialize(english_text)
```
anathomical: Feel free to take or leave this suggestion as I don't feel particularly strongly about it:

While I recognize that we don't currently support multiple locales, I think it's probably worth building into the method signature at this time.  You can just pass the `locale` into the `Translate` service, and then your `PROMPT_START` value could be a hash keyed on `locale` with only one entry.  That establishes a pattern for the future, and also lets OpenAI and Gengo translation requests have the same signature (`english_text:, locale:`).

services/QuillLMS/app/models/english_text.rb
```diff
@@ -12,8 +12,9 @@
 class EnglishText < ApplicationRecord
   has_many :translated_texts
   has_many :translation_mappings
+  has_many :gengo_jobs
 
-  def needs_translation?(locale: Gengo::SPANISH_LOCALE)
-    translated_texts.where(locale:).empty?
+  def needs_gengo_translation?(locale: Gengo::SPANISH_LOCALE)
```
anathomical: This is purely a naming concern, but I think I'd inver the logic here and call this `has_gengo_translation?`.  `needs` is opinionated, suggesting that if it's false we need to do something, while `has` is more neutral, allowing us to say "it doesn't have one, but also doesn't need one".  If you take this suggestion, remember that you'll need to invert any conditions that use it.  `do_something if e.needs_gengo_translation?` would need to become `do_something unless e.has_gengo_translation?`

services/QuillLMS/app/models/gengo_job.rb
```diff
@@ -0,0 +1,36 @@
...
+#  english_text_id    :integer          not null
+#  translated_text_id :integer
+#  translation_job_id :string           not null
+#
+class GengoJob < ApplicationRecord
+  belongs_to :translated_text
+  belongs_to :english_text
+  scope :pending_translation, -> { where(translated_text_id: nil) }
+
+  def self.fetch_and_save_pending! = pending_translation.each(&:fetch_translation!)
```
anathomical, line 20: Take or leave:

I wonder if this should trigger an async worker.  It's an unpredictable / potentially large enough series of external API calls that wait times might be especially surprising?

services/QuillLMS/app/models/gengo_job.rb
```diff
@@ -0,0 +1,36 @@
...
+  def fetch_translation! = Gengo::SaveTranslatedText.run(translation_job_id)
+  def update_or_create_translated_text!(translation)
+    return update_translated_text!(translation) if translated_text.present?
+
+    self.translated_text = TranslatedText.create(
+      locale: locale,
+      english_text_id: english_text_id,
+      translation: translation
+    )
+  end
```
anathomical: I think that this might be surprising in it does not persist the `translated_text` relationship in the `create` path.  `self.translated_text = ...` saves the relationship to memory, but doesn't write the relationship to the database.  I suspect that the intention here is to persist the relationship.

services/QuillLMS/app/models/gengo_job.rb
```diff
@@ -0,0 +1,36 @@
...
+
+  private def update_translated_text!(translation)
+    translated_text.update(translation:)
+  end
```
anathomical: Take or leave:

I don't know if this is better, but you could also `delegate :translation, to: :translated_text` to avoid having to write a method for this and just call `instance.update(translation: 'Translated text')`.  I sometimes hesitate to use delegation because it can be surprising, but it feels more "Rails-y" to do it here than write a custom method to basically do the same thing.

services/QuillLMS/app/models/translated_text.rb
```diff
@@ -4,19 +4,21 @@
...
-  scope :pending_translation, -> { where(translation: nil) }
-
-  def self.fetch_and_save_pending! = pending_translation.each(&:fetch_translation!)
-
-  def fetch_translation! = Gengo::SaveTranslatedText.run(translation_job_id)
+  scope :ordered_by_source_api, lambda { |source_api = OPEN_AI_SOURCE|
+    order(
+      Arel.sql("CASE WHEN source_api = '#{source_api}' THEN 0 ELSE 1 END, source_api ASC")
+    )
+  }
```
anathomical: I mentioned this above, but this feels like a weird piece of code.  I think I recognize what we're trying to do here, but suspect that this may be a premature solution.  We don't actually have the problem of multiply-sourced translation strings yet, so maybe we can just skip trying to solve for that issue for now.

services/QuillLMS/app/services/concerns/open_ai/api.rb
```diff
@@ -0,0 +1,51 @@
...
+
+    def run
+      return cleaned_results if response.present?
+
+      @response = post_request
+
+      cleaned_results
+    rescue *Evidence::HTTP_TIMEOUT_ERRORS
+      []
+    end
```
anathomical, line 38: Take or leave:

I don't know if this is better, but I think this reads a bit cleaner with `||=` notation and some one-line methods.
```ruby
def response = @response ||= post_request
def run = cleaned_results
```
Assuming that `cleaned_results` implementations reference `response`, the `||=` lets you ensure that `post_request` will only run once, and that if the response has already been saved `response` will contain the cached data.

services/QuillLMS/app/services/concerns/open_ai/api.rb
```diff
@@ -0,0 +1,51 @@
...
+
+    private def headers
+      {
+        "Content-Type" => "application/json",
+        "Authorization" => "Bearer #{Evidence::OpenAI::API_KEY}"
```
anathomical: I wonder if this should reference the `ENV` variable directly, or set a duplicate constant on this concern rather than relying on a constant in an engine.  (While this won't ever happen in practice, in theory you could deploy the LMS application without including the Evidence engine, which would break this).

services/QuillLMS/app/services/concerns/open_ai/api.rb
```diff
@@ -0,0 +1,51 @@
...
+      }
+    end
+
+    private def post_request
+      self.class.post(endpoint, body: request_body.to_json, headers: headers, timeout: TIMEOUT)
```
anathomical, line 48: Take or leave:

I'm probably just overly enamored with the ability to not have to declare values when they are stored in a variable with the same name as the key, but I'd likely write this as:
```ruby
def body = request_body.to_json
def timeout = TIMEOUT
def post_request = self.class.post(endpoint, body:, headers:, timeout:)
```

services/QuillLMS/app/services/gengo/request_translations.rb
anathomical: I mentioned earlier that I like `has_gengo_translation?` over `needs`, and if you make that change you'll need to invert the conditional here.

services/QuillLMS/app/services/gengo/save_translated_text.rb
```diff
@@ -20,26 +20,39 @@ def initialize(job_id)
...
+        locale:,
+        translation: new_translation,
+        source_api: TranslatedText::GENGO_SOURCE
+      )
+    end
+
+    private def create_gengo_job
+      GengoJob.create(
         english_text_id: job[SLUG],
         translation_job_id: job[JOB_ID],
```
anathomical: We made took `job[LC_TGT]` out of this `create` call by making a one-liner method, it makes sense to me to follow the same pattern for `english_text_id` and `translation_job_id`:
```ruby
private def english_text_id = job[SLUG]
private def translation_job_id = job[JOB_ID]
private def create_gengo_job = GengoJob.create(english_text_id:, translation_job:, locale:)
```

services/QuillLMS/app/models/translated_text.rb
```diff
@@ -4,19 +4,21 @@
...
-#  id                 :bigint           not null, primary key
-#  locale             :string           not null
-#  translation        :text
-#  created_at         :datetime         not null
-#  updated_at         :datetime         not null
-#  english_text_id    :integer          not null
-#  translation_job_id :string           not null
+#  id              :bigint           not null, primary key
+#  locale          :string           not null
+#  source_api      :string
```
anathomical, line 9: I think it's probably worth adding some validators to this file for the values we know have to be present.  And also worth validating the values for `source_api`
```ruby
SOURCES = [
  GENGO_SOURCE = "gengo"
  OPEN_AI_SOURCE = "open_ai"
]
validates :source_api, presence: true, inclusion: { in: SOURCES }
validates_presence_of :english_text_id # I think this could also be validates_presence_of :english_text
```

services/QuillLMS/app/services/open_ai/save_translated_text.rb
```diff
@@ -0,0 +1,28 @@
...
+    def run
+      raise OpenAITranslationError unless response.present?
+
+      translated_text.update(translation: response)
+    end
+
+    private def response = @response ||= Translate.run(english_text: english_text.text)
+
+    private def translated_text
+      @translated_text ||= TranslatedText.find_or_initialize_by(
```
anathomical: I think we need to be setting a `source_api` here.

services/QuillLMS/app/services/open_ai/save_translated_text.rb
```diff
@@ -0,0 +1,28 @@
...
+        english_text_id: english_text.id,
+        locale: Gengo::SPANISH_LOCALE
```
anathomical: Feels weird to define our locales in a module for another third-party service.  I wonder if we should move the locales to `TranslatedText` or something.

services/QuillLMS/app/services/open_ai/save_translated_text.rb
```diff
@@ -0,0 +1,28 @@
...
+      )
```
anathomical: Take or leave, but stylistically I prefer:
```ruby
private def locale = Gengo::SPANISH_LOCALE
private def source_api = TranslatedText::OPEN_AI_SOURCE
# note that setting english_text instead of english_text_id lets us pass the object that's already set on the service object
private def trasnlated_text = @translated_text ||= TranslatedText.find_or_initialize_by(english_text:, locale:, source_api:)
```

services/QuillLMS/app/services/open_ai/save_translated_text.rb
```diff
@@ -0,0 +1,28 @@
+# frozen_string_literal: true
+
+module OpenAI
+  class SaveTranslatedText < ApplicationService
```
anathomical: Take or leave:

The name here implies to me that the translation already exists, but I'm pretty sure that this class generates the translation on run.  So maybe this should be `TranslateAndSaveText` or something?

services/QuillLMS/app/services/open_ai/translate.rb
```diff
@@ -0,0 +1,57 @@
+# frozen_string_literal: true
+
+module OpenAI
+  class Translate < ApplicationService
+    include OpenAI::Api
+    PROMPT_START = <<~STRING
```
anathomical, line 6: Take or leave:

I mentioned this elsewhere, but I like the idea of storing prompts in a hash by locale, even if we only have one locale in it now.

services/QuillLMS/app/services/open_ai/translate.rb
```diff
@@ -0,0 +1,57 @@
...
+    KEY_CONTENT = "content"
+    ROLE_SYSTEM = "system"
+    ROLE_USER = "user"
+    ROLE_ASSISTANT = "assistant"
+    RESPONSE_FORMAT = { "type" => "json_object" }
+    attr_reader :english_text, :system_prompt
+
+    def initialize(english_text:)
+      @english_text = english_text
+      prompt = PROMPT_START + english_text
```
anathomical: I would make `prompt` a one-liner:
```ruby
private def prompt = PROMPT_START + english_text
```
Or, given that you just immediately set the value to `@system_prompt`, maybe we don't need `attr_accessor` for `:system_prompt`, and we just have `private def system_prompt = PROMPT_START + english_text`

services/QuillLMS/app/models/translated_text.rb
```diff
@@ -4,19 +4,21 @@
...
 class TranslatedText < ApplicationRecord
+  GENGO_SOURCE = "gengo"
+  OPEN_AI_SOURCE = "open_ai"
   belongs_to :english_text
-  scope :pending_translation, -> { where(translation: nil) }
-
-  def self.fetch_and_save_pending! = pending_translation.each(&:fetch_translation!)
-
-  def fetch_translation! = Gengo::SaveTranslatedText.run(translation_job_id)
+  scope :ordered_by_source_api, lambda { |source_api = OPEN_AI_SOURCE|
```
brendanshean, line 24: While our codebase has some [cases](https://github.com/empirical-org/Empirical-Core/pull/9288) of Arel, those involved a requirement to get around order clauses using raw sql, we tend to avoid its use.

Also, I think the current query is susceptible to SQL injection since the string is being interpolated directly.  Using '?'  placeholder instead with plain SQL would be preferred here: 
```
scope :ordered_by_source_api, lambda { |source_api = OPEN_AI_SOURCE|
  order("CASE WHEN source_api = ? THEN 0 ELSE 1 END, source_api ASC", source_api)
}
```

services/QuillLMS/app/models/concept_feedback.rb
```diff
@@ -53,10 +56,23 @@ def create_translation_mappings
...
+  end
 
-  def translate! = Gengo::RequestTranslations.run(english_texts)
-  def fetch_translations! = translated_texts.each(&:fetch_translation!)
+  def translate!(locale:, source_api:)
+    case source_api
+    when TranslatedText::GENGO_SOURCE
+      Gengo::RequestTranslations.run(english_texts, locale)
+    when TranslatedText::OPEN_AI_SOURCE
+      OpenAI::SaveTranslatedTexts.run(english_texts)
```
brendanshean: I would get rid of the wrapper object and just call this directly:
`      english_texts.each{ |text| OpenAI::SaveTranslatedText.run(text) }`

services/QuillLMS/app/services/open_ai/save_translated_text.rb
```diff
@@ -0,0 +1,28 @@
...
+
+    class OpenAITranslationError < StandardError; end
+    attr_accessor :english_text
```
brendanshean: I think you want `attr_reader` here.

services/QuillLMS/app/services/open_ai/save_translated_text.rb
brendanshean: I would pull locale up to the initializer that way you are making the dependency more visible:
````
def initialize(english_text:, locale: Gengo::SPANISH_LOCALE)
  @english_text = english_text
  @locale = locale
end

# Yielding a simpler construction below
private def translated_text = @translated_text ||= TranslatedText.find_or_initialize_by(english_text:, locale:)
```

services/QuillLMS/app/services/open_ai/save_translated_texts.rb
brendanshean, line 1: I think this class can be removed.  See my point earlier in the concept_feedback.rb

services/QuillLMS/app/services/open_ai/save_translated_text.rb
brendanshean, line 1: I tend to name service objects as a noun that end in 'or' or 'er' (as opposed to a verb).
That way when call them, the `.run` is more natural:`TranslatedTextSaver.run` vs `SaveTranslatedText.run`

services/QuillLMS/spec/factories/gengo_jobs.rb
```diff
@@ -0,0 +1,20 @@
...
+#  id                 :bigint           not null, primary key
+#  created_at         :datetime         not null
+#  updated_at         :datetime         not null
+#  english_text_id    :integer          not null
+#  translated_text_id :integer
+#  translation_job_id :string           not null
+#
+FactoryBot.define do
+  factory :gengo_job do
+    english_text_id { 1 }
```
anathomical: I'd like to see this done with an association so that the underlying model gets created in case it matters for a spec:
```
association :english_text
```

services/QuillLMS/spec/factories/translated_texts.rb
```diff
@@ -4,18 +4,24 @@
...
+#  source_api      :string
+#  translation     :text             not null
+#  created_at      :datetime         not null
+#  updated_at      :datetime         not null
+#  english_text_id :integer          not null
 #
 FactoryBot.define do
   factory :translated_text do
     locale {Gengo::SPANISH_LOCALE}
     english_text_id { 1 }
```
anathomical: Again, I think it's worth making this an `association`

services/QuillLMS/spec/models/concept_feedback_spec.rb
```diff
@@ -74,39 +75,116 @@
       it 'adds the translations to the data' do
         expect(subject["translatedDescription"]).to eq(translation)
       end
+
+      context 'there is only a gengo translation' do
+        let(:translated_text) { create(:gengo_translated_text, translation: translation)}
```
anathomical, line 80: Could be `create(:gengo_translated_text, translation:)`

services/QuillLMS/spec/models/concept_feedback_spec.rb
```diff
@@ -74,39 +75,116 @@
...
+
+        it 'adds the translations to the data' do
+          expect(subject["translatedDescription"]).to eq(translation)
+        end
+      end
+
+      context 'there is a gengo and open ai translation' do
```
anathomical, line 87: Take or leave:

I mentioned this in the relevant files, but I'm not sure this is a case we need to try to solve right now.

services/QuillLMS/spec/models/concept_feedback_spec.rb
```diff
@@ -74,39 +75,116 @@
...
+        end
+
+      end
     end
   end
 
-  describe '#translation(locale:)' do
+  describe '#translation(locale:, source:)' do
     let(:locale) {"es-la"}
+    let(:translation_locale) { locale }
```
anathomical, line 112: Do we need to separate `locale` from `translation_locale`?  Can we just override `locale` in specs where we want a different one?

services/QuillLMS/spec/models/concept_feedback_spec.rb
```diff
@@ -74,39 +75,116 @@
...
+    let(:source) { TranslatedText::OPEN_AI_SOURCE }
```
anathomical: If we call this `source_api` it lets us simplify creation calls below with `create(:translated_text, source_api:, ...)`

services/QuillLMS/spec/models/concept_feedback_spec.rb
```diff
@@ -74,39 +75,116 @@
...
+
+    context 'source is not passed in' do
+      subject { concept_feedback.translation(locale: locale)}
+
+      context 'a translation exists' do
+        let(:translation) { "test translation" }
+        let(:translated_text) do
+          create(:translated_text,
+           translation: translation,
+           locale: translation_locale,
```
anathomical, line 123: I think we can just use the keys here if we use `locale` instead of `translation_locale`

services/QuillLMS/spec/models/english_text_spec.rb
```diff
@@ -17,24 +17,27 @@
...
     let(:locale) { Gengo::SPANISH_LOCALE }
-    let!(:translated_text) {
-      create(:translated_text,
-      english_text: english_text,
-      locale: other_locale)
-    }
+    let!(:gengo_job) do
+      create(
+        :gengo_job,
+        english_text: english_text,
```
anathomical: Can omit the variable value here.

services/QuillLMS/app/models/concept_feedback.rb
brendanshean, line 1: I think the translation logic belongs outside of concept_feedback.rb, especially given that we've abstracted out translation_mappings with a `as: :source`

Maybe something like:

```
# app/models/concerns/translatable.rb
module Translatable
  extend ActiveSupport::Concern

  included do
    has_many :translation_mappings, as: :source
    has_many :english_texts, through: :translation_mappings
    has_many :translated_texts, through: :english_texts
    has_many :gengo_jobs, through: :english_texts
  end

  def create_translation_mappings
  def translation(locale:, source_api: TranslatedText::OPEN_AI_SOURCE)...
  def translations(locale:, source_api:)...
  def translate!(locale:, source_api:)...
  def fetch_translations!  
end
```

services/QuillLMS/app/models/concept_feedback.rb
```diff
@@ -32,17 +32,20 @@ class ConceptFeedback < ApplicationRecord
...
   store_accessor :data, :description
 
   after_commit :clear_concept_feedbacks_cache
 
   def cache_key = "#{ALL_CONCEPT_FEEDBACKS_KEY}_#{activity_type}"
 
   def as_json(options=nil)
-    translation = translation(locale: "es-la")
+    source_api = options&.dig(:source_api) || TranslatedText::OPEN_AI_SOURCE
+    locale = Gengo::SPANISH_LOCALE
```
brendanshean: I think that `locale = Gengo::SPANISH_LOCALE` can be removed and put as a default parameter in the `translate!` method.

services/QuillLMS/spec/models/gengo_job_spec.rb
```diff
@@ -0,0 +1,54 @@
...
+#  translation_job_id :string           not null
+#
+require 'rails_helper'
+
+RSpec.describe GengoJob, type: :model do
+  describe "self.pending_translation" do
+    subject { GengoJob.pending_translation }
+
+    let!(:untranslated) { create(:gengo_job) }
+    let!(:complete) { create(:gengo_job, translated_text_id: 2) }
```
anathomical, line 22: These two `let!` calls are repeated both here and in the next `describe` block which makes me think that there should be some `context` that wraps both `describe`s where the `let!`s are declared.

services/QuillLMS/app/models/gengo_job.rb
```diff
@@ -0,0 +1,36 @@
...
+#
+class GengoJob < ApplicationRecord
+  belongs_to :translated_text
+  belongs_to :english_text
+  scope :pending_translation, -> { where(translated_text_id: nil) }
+
+  def self.fetch_and_save_pending! = pending_translation.each(&:fetch_translation!)
+
+  def fetch_translation! = Gengo::SaveTranslatedText.run(translation_job_id)
+  def update_or_create_translated_text!(translation)
```
brendanshean: I found the flow for this function a little confusing.  What about getting rid of helper method

```
  def update_or_create_translated_text!(translation)
    return if translated_text&.update(translation:)

    self.translated_text = TranslatedText.create(locale:, english_text_id:, translation:)
  end
```

services/QuillLMS/app/services/gengo/save_translated_text.rb
```diff
@@ -20,26 +20,39 @@ def initialize(job_id)
...
-    private def job = response.dig(RESPONSE, JOB)
+    private def job = response&.dig(RESPONSE, JOB)
     private def active_job? = ![DELETED, CANCELED].include?(job[STATUS])
     private def new_translation = job[BODY_TGT]
+    private def locale = job[LC_TGT]
 
+    private def gengo_job
+      # Not using find_or_create_by because the find_by doesn't make an external API request
+      @gengo_job ||= GengoJob.find_by(translation_job_id: job_id) || create_gengo_job
+    end
```
brendanshean, line 40: Let's add a space between the methods here.

anathomical (commented):
I've got a ton of comments, but the vast majority are purely stylistic.  Feel free to take or leave those as you like.  A handful are more substantive, like whether we even need to worry about ordering of translations, all of which I'm open to discussion of.

services/QuillLMS/app/services/gengo/save_translated_text.rb
brendanshean, line 1: There's a `    attr_accessor :job_id` at the top of this file that should be a `attr_reader`.  In general, `attr_reader` should be the default attr choice for initializers.  (Think of immutable data coming into the object)

brendanshean (changes requested):
It looks like you've made some good progress with the OpenAI translations.  I've left some comments for cleaning up the code a little bit.

services/QuillLMS/spec/factories/gengo_jobs.rb
brendanshean: Even better, just write `english_text`

services/QuillLMS/app/models/translated_text.rb
samanthamjohn: Hah, this is what I originally had, and I think Rubocop changed it! I will switch it back

services/QuillLMS/app/models/concept_feedback.rb
samanthamjohn: Yeah, that is my plan too, I'm waiting to pull it out until we add more models though.

services/QuillLMS/app/models/concept_feedback.rb
samanthamjohn: I worry that it's a little unconventional to do that--it would still work, but `as_json(options=nil)` is such a standard rails things to do that I sort of prefer to leave it that way, even id it's a bit messier. But open to what other folks think! Maybe @brendanshean  can tiebreak?

services/QuillLMS/app/models/gengo_job.rb
samanthamjohn: that makes sense, I'm thinking it should go in the `GengoJob` class. i.e. `fetch_translation!` should kick off a worker, rather than calling the service directly.

services/QuillLMS/app/models/gengo_job.rb
samanthamjohn: oh good catch--we're not using this anyways, so I'll delete it.

services/QuillLMS/app/services/concerns/open_ai/api.rb
samanthamjohn: This one I'm gonna leave since it's a straight copy of the version in the `Evidence` engine, and if we want to bring them back together later better if they're more identical.

services/QuillLMS/spec/models/concept_feedback_spec.rb
samanthamjohn: `subject` uses `locale`, while `translated_text` uses `translated_locale` so we want to be able to override one but not the other.

services/QuillLMS/spec/models/concept_feedback_spec.rb
samanthamjohn: this one is so that it's overwritable so we do need two diff names

services/QuillLMS/app/models/concept_feedback.rb
brendanshean: I would agree with Thomas' general sentiment about  `method(options=nil)` being opaque and why not just put what we're expecting, but `as_json` is documented with that method signature so I think you should leave it.

services/QuillLMS/app/models/gengo_job.rb
samanthamjohn: that makes sense, AND I realized this is another orphan method that is unused, so even better than delegate is delete!

services/QuillLMS/app/services/concerns/open_ai/api.rb
samanthamjohn: Agree, but this one I am going to leave for consistency with its twin for now.

services/QuillLMS/app/models/translated_text.rb
samanthamjohn: The `english_text_id` is enforced at the database level. 
I like the sources thing

services/QuillLMS/app/services/open_ai/save_translated_text.rb
samanthamjohn: done. Though now that I did it, I kinda think we should have a locales file. For now I'm gonna leave it translatedTExt though.

services/QuillLMS/app/services/open_ai/translate.rb
samanthamjohn: Let's think about when we have more languages. I think the prompt will be largely the same, just with a different language name each time.

services/QuillLMS/spec/models/concept_feedback_spec.rb
samanthamjohn: I'd like to find a linter that would do this for us automatically!

samanthamjohn: I will make another PR to add this to Rubocop
```
Style/HashSyntax:
  EnforcedShorthandSyntax: consistent
```

services/QuillLMS/spec/models/concept_feedback_spec.rb
samanthamjohn: This is a case that I run into a lot on my local machine because I am testing them both. I think we can take it out later if it turns out we get rid of Gengo, but since the code is already there, and it's useful for dev purposes, I'm gonna leave it in for now.

services/QuillLMS/spec/models/gengo_job_spec.rb
samanthamjohn: hmm, I don't love adding the same context around two describe blocks though, that feels weird. I know it's a little reudundant but since it's a test, I don't mind so much.

services/QuillLMS/app/models/translated_text.rb
samanthamjohn: Hmm, I can't do your version because rspec complains: 
```
ActiveRecord::UnknownAttributeReference:
       Dangerous query method (method whose arguments are used as raw SQL) called with non-attribute argument(s): "CASE WHEN source_api = ? THEN 0 ELSE 1 END, source_api ASC".This method should not be called with user-provided values, such as request parameters or model attributes. Known-safe values can be passed by wrapping them in Arel.sql().
```

We can sanitize ourselves like this: 
```
scope :ordered_by_source_api, lambda { |source_api = OPEN_AI_SOURCE|
    source_api = OPEN_AI_SOURCE unless SOURCES.include? source_api
    order (
      Arel.sql("CASE WHEN source_api = '#{source_api}' THEN 0 ELSE 1 END, source_api ASC")
    )
  }
```

services/QuillLMS/app/services/open_ai/save_translated_text.rb
samanthamjohn: Thomas suggested changing it to `TranslateAndSaveText`, and I think it would make the noun version pretty awkward. Maybe we punt on the naming til the next PR.

samanthamjohn:
I incorporated most of the changes. The one thing that might be a sticking point is that I did want to leave in the thing to order by a specified translation type, because it's useful for testing locally. We could take it out later, as this part of the code is still in pretty active flux!

services/QuillLMS/app/models/translated_text.rb
brendanshean: Let's add a 
```
SOURCE_APIS = [GENGO_SOURCE, OPEN_AI_SOURCE]

validates :source_api, inclusion: { in: SOURCE_APIS }
```

services/QuillLMS/spec/models/concept_feedback_spec.rb
brendanshean: Good idea.  Please add this

services/QuillLMS/spec/models/concept_feedback_spec.rb
```diff
@@ -192,19 +270,64 @@
...
+
+    it { expect(subject).to include(gengo_job)}
+  end
+
   describe "#fetch_translations!" do
-    context "there is a translated_text associated" do
-      let(:t1) { create(:translated_text)}
-      let(:t2) { create(:translated_text)}
+    context "there is a gengo_job associated" do
+      let(:t1) { create(:gengo_job)}
```
brendanshean: Let's use full names instead of abbreviations for variables:  `translation1`,` translation2`

brendanshean (approved):
This looks great.  Nice work on this.

jtarchie:
I’ve written a golang cli to aid in translations. https://github.com/jtarchie/rails-translator Sharing if you are looking for alternative implementations.

PR #11982, opened 2024-06-26
fix zero width user input bug

happythenewsad:
## WHAT
Sanitizes zero width spaces from user input 

## WHY
These are not detectable with the naked eye, leading to confusing bugs. Also, zero width chars are not removed via common methods like `String#strip` 

## HOW
Create a convenience strip method for these characters. Make note to discuss a default solution (i.e. running the sanitizing method on all user input, as middleware) in the next eng meeting. 

### Screenshots

### Notion Card Links
https://www.notion.so/quill/Teacher-cannot-invite-a-co-teacher-2d6a2eaca9494312825664e2427ec2ac

tested in rails console

PR Checklist | Your Answer
------------ | -------------

### DIFF
--- a/services/QuillLMS/app/controllers/invitations_controller.rb
+++ b/services/QuillLMS/app/controllers/invitations_controller.rb
@@ -2,7 +2,7 @@
 
 class InvitationsController < ApplicationController
   before_action :verify_current_user_owns_classrooms, only: :create_coteacher_invitation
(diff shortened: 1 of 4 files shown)


brendanshean (approved):
Thanks for catching this very tricky bug

emilia-friedberg (approved):
Good catch!

PR #11984, opened 2024-06-26
Show hint translations in the student view

samanthamjohn:
## WHAT
If you add `?showTranslations=true` to the URL (before the # mark) you will see the translated version of the hints. 

## WHY
So that we can start looking at the translations in the app 

## HOW
Added a flag to the hints JS to show the translations

### Screenshots
2024-06-24 at 10 24 AM

### Notion Card Links
[Show users translations for hints](https://www.notion.so/quill/Show-users-the-translations-for-hints-f2702f0426bd44b1bed732a633eec41a?pvs=4)

Looked at it in development. We will have to run some translations in staging in order to see it there.
PR Checklist | Your Answer
------------ | -------------
Have you added and/or updated tests? | No, it was a frontend only change, not sure we have tests for that component. 

### DIFF
--- a/services/QuillLMS/client/app/bundles/Shared/components/feedback/conceptExplanation.tsx
+++ b/services/QuillLMS/client/app/bundles/Shared/components/feedback/conceptExplanation.tsx
@@ -7,16 +7,22 @@ function getClassName(description, leftBox, rightBox) {
   return "concept-explanation empty"
 }
 
-const ConceptExplanation = ({ description, leftBox, rightBox, translatedDescription, translated}) => (
-  <div className={getClassName(description, leftBox, rightBox)}>
-    <div className="concept-explanation-title"><img alt="Light Bulb Icon" src="https://assets.quill.org/images/icons/hint.svg" /><span>Hint</span></div>
-    <div className="concept-explanation-description" dangerouslySetInnerHTML={{__html: description}} />
-    { translated && translatedDescription && <div className="concept-explanation-translation" dangerouslySetInnerHTML={{__html: translatedDescription}} /> }
-    <div className="concept-explanation-see-write">
-      <div className="concept-explanation-see" dangerouslySetInnerHTML={{__html: leftBox}} />
-      <div className="concept-explanation-write" dangerouslySetInnerHTML={{__html: rightBox}} />
+const ConceptExplanation = ({ description, leftBox, rightBox, translatedDescription, translated}) => {
+  const urlParams = new URLSearchParams(window.location.search)
+  const translatedOverride = urlParams.get('showTranslations')
+  const showsTranslation = translatedDescription && (translated || translatedOverride)
+  return (
+    <div className={getClassName(description, leftBox, rightBox)}>
+      <div className="concept-explanation-title"><img alt="Light Bulb Icon" src="https://assets.quill.org/images/icons/hint.svg" /><span>Hint</span></div>
+      <div className="concept-explanation-description" dangerouslySetInnerHTML={{__html: description}} />
+      { showsTranslation && <div className="concept-explanation-translation" dangerouslySetInnerHTML={{__html: translatedDescription}} /> }
+
+      <div className="concept-explanation-see-write">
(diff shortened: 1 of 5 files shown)


services/QuillLMS/client/app/bundles/Shared/components/feedback/conceptExplanation.tsx
```diff
@@ -7,16 +7,22 @@ function getClassName(description, leftBox, rightBox) {
...
-const ConceptExplanation = ({ description, leftBox, rightBox, translatedDescription, translated}) => (
-  <div className={getClassName(description, leftBox, rightBox)}>
-    <div className="concept-explanation-title"><img alt="Light Bulb Icon" src="https://assets.quill.org/images/icons/hint.svg" /><span>Hint</span></div>
-    <div className="concept-explanation-description" dangerouslySetInnerHTML={{__html: description}} />
-    { translated && translatedDescription && <div className="concept-explanation-translation" dangerouslySetInnerHTML={{__html: translatedDescription}} /> }
-    <div className="concept-explanation-see-write">
-      <div className="concept-explanation-see" dangerouslySetInnerHTML={{__html: leftBox}} />
-      <div className="concept-explanation-write" dangerouslySetInnerHTML={{__html: rightBox}} />
+const ConceptExplanation = ({ description, leftBox, rightBox, translatedDescription, translated}) => {
+  const urlParams = new URLSearchParams(window.location.search)
```
emilia-friedberg, line 11: This is new to me. Cool!

services/QuillLMS/client/app/bundles/Shared/components/feedback/conceptExplanation.tsx
```diff
@@ -7,16 +7,22 @@ function getClassName(description, leftBox, rightBox) {
...
+  const translatedOverride = urlParams.get('showTranslations')
+  const showsTranslation = translatedDescription && (translated || translatedOverride)
```
emilia-friedberg: One thing to note is that I think `translatedOverride` here is going to evaluate to `true` even if the querystring is `?showTranslations=false`, because the value is always a string. That probably doesn't matter in practice but it may be worth explicitly checking that it is equal to true?

emilia-friedberg: Also, a more general question here: do we always want the translation to show if the query string is present, even if `translated` is false? Maybe I'm misinterpreting something but it seems to me like maybe we want the query string to create the condition where it is possible to show the translations, but we should still respect the user's toggle?

emilia-friedberg (commented):
Looks good overall, just some questions about the logic here. Also, you're right that there is no existing test for this component, but I think since it's quite small it would be good to add at least a snapshot for the translated state and the non-translated state -- example of very short snapshot test here: https://github.com/empirical-org/Empirical-Core/blob/develop/services/QuillLMS/client/app/bundles/Shared/components/studentLessons/__tests__/finalAttemptFeedback.test.tsx#L17.

services/QuillLMS/client/app/bundles/Shared/components/feedback/conceptExplanation.tsx
samanthamjohn: The `translated` flag only comes in on the admin interface. So this is a different override to show translations for students. The `translatedDescription` is the thing that figures out if there is an actual translation to show. Your comment indicates to me that these things need better naming!

samanthamjohn:
Added the test, and clarified the language around when to show the translation. I started on the path of adding an admin interface to show the hints, but there are _so_ many places we show the hints, that I think it's better to do something more global from the URL, or I've been thinking that the admin interface for the translations perhaps ought to be separate from the normal admin interface, something more like this (with editing) https://quill-translation-test.vercel.app/

emilia-friedberg (approved):
Thank you for the tests, and the renaming is definitely helpful! This isn't part of the PR, but in response to your comment, I like the idea of just having a master page where we show all the hints in translations rather than translating them everywhere they appear (which sounds both onerous for you and also like potentially cluttered UX).

PR #11994, opened 2024-06-28
Rubocop enforce hash syntax

samanthamjohn:
Add `EnforcedShorthandSyntax: always` to rubocop. 

Note, this will catch instances of this: 
``` 
{ foo: foo } 
``` 

But it will not catch instances in named arguments for method calls: 
```
def foo(bar:)
...
end

foo(bar: bar)
```


brendanshean (approved):
Awesome.  Thanks for finding this configuration.

PR #11996, opened 2024-07-01
Fix wrong date formats in jest tests

cissyxyu:
## WHAT
Fix jest test date formatting errors.

## WHY
They were causing warnings when running jest tests because npm now requires all dates to be formatted in standard UTC format.

## HOW
Fix jest test date strings to be standard UTC format.

### Screenshots

### Notion Card Links
https://www.notion.so/quill/Console-log-cleanup-value-provided-is-not-in-a-recognized-RFC2822-or-ISO-format-9-0abd6e81e3eb420ab72ca03a04b4d591?pvs=4

Not a feature, just some test fixes. Now when I run `npx jest --watch` and run all tests I'm not getting the console errors.

PR Checklist | Your Answer
------------ | -------------
Have you deployed to Staging? | No - not a feature change.


emilia-friedberg (commented):
Hey Cissy! The test fixes here look good, but I think the idea of this card (which evidently should have been clearer) is to address console.log errors that are surfacing in Jest tests. Sometimes these are false positives (Jest-only issues), but in this case it looks to me like we actually are seeing this console errors in the browser as well, so I think we need to address the errors in the actual component.

cissyxyu:
> Hey Cissy! The test fixes here look good, but I think the idea of this card (which evidently should have been clearer) is to address console.log errors that are surfacing in Jest tests. Sometimes these are false positives (Jest-only issues), but in this case it looks to me like we actually are seeing this console errors in the browser as well, so I think we need to address the errors in the actual component.

Hi Emilia! I was able to find one actual browser instance of this error in the assigning activities flow (the publication date was being sent from the backend in the wrong format). That one bug was causing all of the console errors in the activity assignment flow. I fixed this and pushed it to this PR. 

However I wasn't able to replicate the other jest bugs in the actual console after clicking around on the front end a bunch (in their respective components that were throwing the old Jest errors). Do you recall which of the components or pages seemed to be throwing the console errors?

emilia-friedberg (approved):
Hey Cissy -- I think I just checked the first instance in the PR (assign_new_activities), which it looks like is the one you found the other fix for? If you checked everything else in the browser and they look good I think this is good to go! Thanks for your work on this.

eadams17 (approved):

PR #11998, opened 2024-07-01
Lms/adgr email subscription model controller

anathomical:
## WHAT
Build the backend infrastructure to allow admins to subscribe to regular emails with the Admin Diagnostic Growth Report data.
- A back-end model to store which admins are subscribed to these emails
- A controller to create, update, and destroy these models for admins
- A small view to help facilitate admin unsubscribe actions so that a link can be included in the report emails

It seemed worth cutting out this bit of complete code even though on its own it won't change any user-experience.  This is reviewable as a slice even if we know additional front-end code and back-end logic for actually sending these emails will be required in follow-ons.
## WHY
We offer an email subscription to the Admin Snapshot reports, and want to provide the same experience with the Admin Diagnostic Growth Reports.
## HOW
(Modeled largely on the Admin Snapshot `PdfSubscription` model and controller.)
- create migration
- define model
- create controller
- add small view

### Notion Card Links
https://www.notion.so/quill/Email-subscription-for-Admin-Diagnostic-Growth-Report-d7d578f870104d77ac9e95b23ee0ab51

QA for this has been done entirely in specs.  Additional QA of this feature will be done in follow-on code that makes use of this infrastructure.

PR Checklist | Your Answer
------------ | -------------
Have you deployed to Staging? | No, this is purely setup for follow-on code


services/QuillLMS/config/routes.rb
```diff
@@ -779,6 +779,14 @@
     end
   end
 
+  resources :email_subscriptions, param: :type, only: [:destroy] do
+    post :create_or_update
+    get :current
+    collection do
+      get 'unsubscribe/:cancel_token', to: 'email_subscriptions#unsubscribe', as: :unsibscribe
```
brendanshean: typo with `unsibscribe`

services/QuillLMS/spec/factories/email_subscriptions.rb
```diff
@@ -0,0 +1,26 @@
...
+#  subscription_type :string           not null
+#  created_at        :datetime         not null
+#  updated_at        :datetime         not null
+#  user_id           :integer          not null
+#
+FactoryBot.define do
+  factory :email_subscription do
+    frequency { EmailSubscription::MONTHLY }
+    subscription_type { EmailSubscription::ADMIN_DIAGNOSTIC_REPORT }
+    association :user
```
brendanshean: I tend to prefer the shorter: `user`
I only use `association` when it's a non-standard factory that I need to refer to.

brendanshean (approved):
Nice work on this.  I thought the layout of the controller was very easy to follow.

services/QuillLMS/config/routes.rb
anathomical: Good catch.

services/QuillLMS/spec/factories/email_subscriptions.rb
anathomical: Seems reasonable to me.  I'll make this change.

services/QuillLMS/app/controllers/email_subscriptions_controller.rb
```diff
@@ -0,0 +1,76 @@
...
+
+  private def subscription_params
+    params.require(:subscription)
+      .permit(:frequency, :params)
+  end
+
+  # Using three different param patterns to find the relevant EmailSubscription
+  # feels like it should be some sort of anti-pattern, but these are genuinely
+  # different workflows, and it's a pattern we've established in the
+  # PdfSubscriptionsController
```
happythenewsad, line 64: 👍

services/QuillLMS/app/controllers/email_subscriptions_controller.rb
```diff
@@ -0,0 +1,76 @@
...
+  private def email_subscription_by_id = EmailSubscription.find_by(id:)
+  private def email_subscription_by_cancel_token = EmailSubscription.find_by(cancel_token:)
+  private def email_subscription_by_type = EmailSubscription.find_by(user_id:, subscription_type:)
+
+  # This param is set on the route
+  private def cancel_token = params[:cancel_token]
+  # This param is set on the route
+  private def id = params[:type]
+  # This param is set on the route (some routes name it one thing, some the other)
+  private def subscription_type = params[:email_subscription_type] || params[:type]
```
happythenewsad, line 74: Nit: `params[:type]` should be `id`

happythenewsad (approved):

PR #12011, opened 2024-07-02
Make activities and questions translatable

samanthamjohn:
## WHAT
Allow you to create translations for activities and questions.
## WHY
So that ELL students can read instructions in their own language. 
## HOW
Added a concern called `Translatable` that encapsulates the logic needed to translate a quill model. 

### Notion Card Links
[Activites](https://www.notion.so/quill/Add-translations-for-activity-instructions-26c26760eb564ac1a68d0713f228d2e5?pvs=4)
[Questions](https://www.notion.so/quill/Add-translations-for-Questions-e5d87594978848ed974a4052cc201283?pvs=4)
Ran translate!  on all the models in the staging console. 

PR Checklist | Your Answer
------------ | -------------


samanthamjohn:
Note: the test failures will be fixed after https://github.com/empirical-org/Empirical-Core/pull/12013 is merged.

services/QuillLMS/app/models/concerns/translatable.rb
```diff
@@ -0,0 +1,92 @@
+# frozen_string_literal: true
+
+# app/models/concerns/translatable.rb
```
brendanshean: Let's delete this comment.

services/QuillLMS/app/models/concerns/translatable.rb
```diff
@@ -0,0 +1,92 @@
...
+  extend ActiveSupport::Concern
+
+  SPANISH_LOCALE = "es-la"
+  DEFAULT_LOCALE = SPANISH_LOCALE
+  GENGO_SOURCE = "gengo"
+  OPEN_AI_SOURCE = "open_ai"
+  SOURCES = [OPEN_AI_SOURCE, GENGO_SOURCE]
+
+  included do
+    has_many :translation_mappings, as: :source
```
brendanshean, line 13: `:example_filename` and `:translatable_attribute` values don't appear to change across instances within a class so I think `class_attribute` might make more sense here.


I'd also like to suggest a couple of renames: 
:translatable_attribute -> :translatable_field_name.  # distinguishes column name vs the value
:example_filename -> :translatable_examples.    # prepend translatable to help differentiate it from non concern methods

```
included do
  class_attribute :translatable_field_name, default: nil
  class_attribute :translatable_examples, default: nil
  ...
end
```

Then in a class the configuration seems more natural:

```
class Question < ApplicationRecord
  include Translatable

  translatable_field_name :instructions
  translation_examples "questions.json"
  ...
```

services/QuillLMS/app/models/question.rb
```diff
@@ -163,6 +165,17 @@ def connect_sentence_combining?
     question_type == TYPE_CONNECT_SENTENCE_COMBINING
   end
 
+  # Translatable
+  private def custom_prompt
```
brendanshean: This field also feels like a class variable in that it doesn't appear to change across instances, but it's also configuration so maybe create a general config file and pull in the prompt examples with it?

You could create a new folder:  translatable_configs/{activities, questions, ...}.json

where the file would have two top-level keys:

examples: {
...
}
custom_prompt: {
... 
}

samanthamjohn: ok, I like that!

services/QuillLMS/spec/models/concept_feedback_spec.rb
brendanshean, line 1: Maybe add something about the concern being included:

`it { expect(ConceptFeedback.ancestors).to include(Translatable) }`

brendanshean (commented):
Something else I forgot to mention, let's add a couple of tests to check that the concern is included:

it { expect(Activity.ancestors).include(Translatable) } 
it { expect(ConceptFeedback.ancestors).to include(Translatable) }

brendanshean (approved):
This looks great!  Nice work on extracting out the concern.  I wasn't thinking of yaml when I suggested config but its way easier to parse than JSON.

PR #12012, opened 2024-07-02
Add translations to the admin interface

samanthamjohn:
## WHAT
Add a translations page so that admins can see what has been translated. 
## WHY
We've started translating a bunch of stuff which we'll want to QA before showing to students. This gives admins a way to see what has been translated without having to add new custom fields everywhere in the admin interface. 
## HOW
Added a new page called Translations to the staff admin interface. 
### Screenshots
2024-07-02 at 4 23 PM

2024-07-02 at 4 22 PM

### Notion Card Links
[First admin page](https://www.notion.so/quill/Add-an-admin-interface-that-shows-all-translated-items-in-one-big-list-313b879b9fdc46e78676f09a429c98fb?pvs=4)

Checked it out on staging. 

PR Checklist | Your Answer
------------ | -------------


services/QuillLMS/app/controllers/pages_controller.rb
```diff
@@ -503,6 +503,12 @@ def locker
     @style_file = "#{ApplicationController::STAFF}.scss"
   end
 
+  def translations
+    return redirect_to profile_path if !staff?
```
anathomical: The linter might catch this, but I prefer `unless condition` over `if !condition`

services/QuillLMS/app/views/pages/translations.html.erb
```diff
@@ -0,0 +1,18 @@
+<h1>Translations</h1>
+<table class="table">
+  <tr class="header-row">
+    <th>Module</th>
+    <th>English</th>
+    <th>Spanish</th>
+  </tr>
+    <%@translations.each do |translation|%>
+      <tr class="data-row">
+        <td><%= translation.source.class.name %></td>
```
anathomical, line 10: This is not a new problem in your code, but you've probably noticed that we often have Model names that don't match what Curriculum calls something.  `ConceptFeedback` vs "Hint" for instance.  Not a problem with this early version, but I wonder if it makes sense to consider including a way to show what Curriculum calls these values somewhere instead of making them learn the underlying model names.

I think solving this issue is outside the scope of this PR, though, so not something you need to solve here.

anathomical (approved):
I like you adding this page to the lockers.  Good call.

services/QuillLMS/app/views/pages/translations.html.erb
```diff
@@ -0,0 +1,18 @@
...
+        <td><%= raw(translation.english_text.text)%></td>
+        <td><%= raw(translation.translated_texts.ordered_by_source_api.first.translation) %></td>
+      </tr>
+    <%end%>
+  </tbody>
+</table>
+
+<p> <%= paginate @translations %> </p>
```
emilia-friedberg, line 18: I've never seen this before -- this is very cool!

services/QuillLMS/app/views/pages/translations.html.erb
emilia-friedberg: FWIW concept feedback is referred to that way in other parts of the admin UI, so they do know what that means in this instance. If we do need to use something different for display at a later date, I think we could just create a variable to store a key value pair for this.

emilia-friedberg (approved):
LGTM! Nice work!

services/QuillLMS/app/controllers/pages_controller.rb
samanthamjohn: Ah, I copied it from the controller action above. I can change to `unless`, I like that better too.

services/QuillLMS/client/app/bundles/Staff/helpers/locker/lockerItems.ts
```diff
@@ -269,7 +275,7 @@ export const lockerItems: LockerItemsInterface = {
     emojiLabel: 'stack of books',
     lockers: {
       [TOOLS]: [EVIDENCE, CONNECT, DIAGNOSTIC, GRAMMAR, LESSONS, PROOFREADER],
-      [ACTIVITY_MANAGEMENT]: [ACTIVITIES, ACTIVITY_PACKS, ACTIVITY_PACKS_CATEGORIES, AUTHORS, RAILS_ADMIN],
+      [ACTIVITY_MANAGEMENT]: [ACTIVITIES, ACTIVITY_PACKS, ACTIVITY_PACKS_CATEGORIES, AUTHORS, TRANSLATIONS, RAILS_ADMIN],
```
brendanshean: Lines 278, 332 should be in the same order.

brendanshean (approved):
Coincidentally, I'm adding something to the locker so reviewing this was very helpful.  😄

eadams17 (approved):
LGTM!

PR #12014, opened 2024-07-02
If the user writes in Dit or Van as their name, don’t lowercase it for them

samanthamjohn:
## WHAT
Our tests were sporadically failing when faker generated the name "Van" as a first name because it would not get capitalized as expected. 
## WHY
We are using a [capitalization gem](https://github.com/infiton/capitalize-names) that by default lowercases Van and Dit, but maybe those could be your first name! That would be weird for that person if their first name was not capitalized. 
## HOW
Passed in an option to skip the van and dit rules on the capitalize gem. 

### Notion Card Links

PR Checklist | Your Answer
------------ | -------------
Have you deployed to Staging? | NO, small change

### DIFF
--- a/services/QuillLMS/app/models/user.rb
+++ b/services/QuillLMS/app/models/user.rb
@@ -473,7 +473,7 @@ class User < ApplicationRecord
   end
 
   def capitalize_name
-    self.name = ::CapitalizeNames.capitalize(name)
+    self.name = ::CapitalizeNames.capitalize(name, skip_van_space: true, skip_dit_space: true)
   end
 
   def admin?
--- a/services/QuillLMS/spec/models/user_spec.rb
+++ b/services/QuillLMS/spec/models/user_spec.rb
@@ -389,16 +389,19 @@ RSpec.describe User, type: :model do
 
     context 'van is the first name' do
       let(:name) { 'van nieman'}
+
       it { expect { subject }.to change(user, :name).from(name).to('Van Nieman') }
     end
 
     context 'dit is the first name' do
       let(:name) { 'dit ditsy'}
+
       it { expect { subject }.to change(user, :name).from(name).to('Dit Ditsy') }
     end
 
     context 'van is in the middle' do
       let(:name) { 'betsy van foo'}
+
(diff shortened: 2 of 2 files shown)


services/QuillLMS/spec/models/user_spec.rb
```diff
@@ -386,6 +386,25 @@
...
+    context 'dit is the first name' do
+      let(:name) { 'dit ditsy'}
+
+      it { expect { subject }.to change(user, :name).from(name).to('Dit Ditsy') }
+    end
+
+    context 'van is in the middle' do
+      let(:name) { 'betsy van foo'}
+
+      it { expect { subject }.to change(user, :name).from(name).to('Betsy Van Foo') }
```
brendanshean, line 405: Our system doesn't distinguish between middle and last names.  In all likelihood, this should most likely be 'Betsy van Foo'.

services/QuillLMS/app/models/user.rb
```diff
@@ -473,7 +473,7 @@ def self.sorting_name_sql
   end
 
   def capitalize_name
-    self.name = ::CapitalizeNames.capitalize(name)
+    self.name = ::CapitalizeNames.capitalize(name, skip_van_space: true, skip_dit_space: true)
```
brendanshean, line 476: I think these settings will give undesired behavior
```
1] pry(main)> CapitalizeNames.capitalize("Alexander Van Humboldt")
=> "Alexander van Humboldt"
[2] pry(main)> CapitalizeNames.capitalize("Alexander Van Humboldt", skip_van_space: true)
=> "Alexander Van Humboldt"
```
Instead, keep the original setting
`self.name = ::CapitalizeNames.capitalize(name)`

and then do some post-processing that checks for the first word in the name for "mac, dit, etc...." and capitalizes them.

anathomical:
Just a broad comment: I'm not sure that we should be making name capitalization decisions without consulting Partnerships.  We don't know WHAT rules we should be supporting here, and I'm not sure that there's an easy way for us to get to a good answer without further consultation.

PR #12020, opened 2024-07-05
Add more premium attributes to Vitally school and district stats

cissyxyu:
## WHAT
Add data about total months a district/school has had premium, and earliest premium subscription start date.

## WHY
This data will allow sales staff to make more informed decisions about schools and districts.

## HOW
Add a function `length_in_months` to the subscription model to calculate the length a premium subscription is from start date to expiration. Add functions in the Vitally serializers to total this number for all subscriptions. Add another function to calculate the earliest start date a school or district has for all its premium subscriptions.

### Screenshots

### Notion Card Links
https://www.notion.so/quill/Add-New-Vitally-Date-1-School-District-Subscription-Length-2-Historical-Subscription-First--24361ba4e073490c8845ad07c8bb2628?pvs=4

deploy to staging and confirm through console actions that serializing a new district and school will have the accurate data for these two new stats.

PR Checklist | Your Answer
------------ | -------------


services/QuillLMS/app/models/subscription.rb
```diff
@@ -327,6 +327,10 @@ def subscription_status
     )
   end
 
+  def length_in_months
```
brendanshean, line 330: I would add a guard clause `return 0 if expiration.nil? || start_date.nil?`

services/QuillLMS/app/services/vitally_integration/serialize_vitally_sales_organization.rb
```diff
@@ -108,12 +108,14 @@ def diagnostic_rollups
 
     def subscription_rollups
       {
-        premium_start_date: premium_start_date,
```
brendanshean, line 111: It looks like all of these can use the shorthand for parameters:
```
{
  earliest_premium_start_date:,
  current_premium_start_date:,
  premium_expiry_date:,
  district_subscription:,
  annual_revenue_current_contract:,
  stripe_invoice_id_current_contract:,
  purchase_order_number_current_contract:
  total_premium_months:,
}
```

services/QuillLMS/app/services/vitally_integration/serialize_vitally_sales_organization.rb
brendanshean, line 1: At the top of this class, let's add a
`attr_reader :district`

Then replace all `@district` instance variables with `district`.

Instance variables like `@district` will not raise an error with misspelling whereas `district` does.

brendanshean (approved):
I've left a few comments for cleanup but good to go after that.  Thank you for updating these analytics.

services/QuillLMS/app/services/vitally_integration/serialize_vitally_sales_account.rb
```diff
@@ -129,6 +133,14 @@ def data
       "https://www.quill.org/cms/schools/#{@school.id}"
     end
 
+    private def first_subscription
+      @school&.subscriptions&.order(start_date: :asc)&.first
+    end
+
+    private def earliest_premium_start_date
+      first_subscription&.start_date || NOT_APPLICABLE
+    end
```
anathomical, line 142: Purely stylistic, and you can skip this suggestion, but these could be one-liners:
```ruby
private def first_subscription = @school&.subscriptions&.order(start_date: :asc)&.first
private def earliest_premium_start_date = first_subscription&.start_date || NOT_APPLICABLE
```

services/QuillLMS/app/services/vitally_integration/serialize_vitally_sales_organization.rb
```diff
@@ -207,11 +209,19 @@ def activities_finished_query(district)
         .where('activity_sessions.state = ?', 'finished')
     end
 
+    private def first_subscription
```
anathomical, line 214: Again, this is just style, and we don't have a requirement to use the shorter style, I just like it:
```ruby
private def first_subscription = @district.subscriptions.order(start_date: :asc).first
private def earliest_premium_start_date = first_subscription&.start_date || VITALLY_NOT_APPLICABLE
private def current_premium_start_date = @district.subscription&.start_date || VITALLY_NOT_APPLICABLE
private def local_premium_months = @district.subscriptions.map(&:length_in_months).sum
```

services/QuillLMS/spec/models/subscription_spec.rb
```diff
@@ -202,6 +202,15 @@
     end
   end
 
+  describe '#length_in_months' do
+    it 'returns the length of the subscription in months' do
+      length_in_months = 45
+      subscription = create(:subscription, start_date: Time.zone.today, expiration: Time.zone.today + (length_in_months).months)
```
anathomical, line 208: I don't think the parens around `length_in_months` is necessary here, but I actually like it as a way to emphasize where the `months` call will apply.  (I often get confused about how Ruby is going to distribute some of those method calls.)

anathomical (approved):

PR #12024, opened 2024-07-08
Move logic for translations admin into the model and out of the view

samanthamjohn:
Plus catch some edge cases

## WHAT
Use model methods for finding translated and english text for a given translation mapping.

## WHY
So that we could test the edge cases and so that the view code would be cleaner.

## HOW
in the admin interface for the translations, we are pulling the translated and english text into a table, I added model methods to the `TranslationMapping` class and used them in the admin view.

Checked the translation admin on staging

PR Checklist | Your Answer
------------ | -------------


brendanshean (approved):
Nice refactoring!

PR #12029, opened 2024-07-08
Don't re-translate texts that already have a translation associated with them. 

samanthamjohn:
## WHAT
Don't retranslate something if we already have a translation saved in our DB. Also updated the translate rake task to take the UIDs of all the translatable activities. 
## WHY
Many of the question instructions along with other things, have identical english text to one another. So when we go to translate all the things, we don't want to spend all our time on duplicates. 
## HOW
Check if something is translated before sending out to OpenAI. 

### Notion Card Links
[Don't retranslate](https://www.notion.so/quill/Don-t-re-translate-the-same-word-with-OpenAI-unless-you-add-a-force-keyword-1293575d5fa44b24acebefb474a59189?pvs=4)

Tested it locally and on staging by running `rake translate:activities[2]` 

PR Checklist | Your Answer
------------ | -------------

### DIFF
--- a/services/QuillLMS/lib/tasks/temporary/translate.rake
+++ b/services/QuillLMS/lib/tasks/temporary/translate.rake
@@ -7,106 +7,315 @@ namespace :translate do
     count = hints.count
     hints.each_with_index do |hint, index|
       puts "translating #{index + 1}/#{count}..."
-      res = Evidence::OpenAI::Translate.run(english_text: hint.description, prompt: hint.prompt)
-      hint.create_translation_mappings
-      hint.english_texts.first.openai_translated_texts.create(
-        translation: res,
-        locale: Translatable::DEFAULT_LOCALE)
+      hint.translate!
     end
 
   end
 
   task activities: :environment do
-    activities = Activity.where("id in (?)", activity_ids).limit(50)
-    CSV.open("open_ai_spanish_activities.csv", "wb") do |csv|
-      csv << ["english", "spanish"]
-      activities.each_with_index do |activity, index|
-        puts "translating #{index + 1}/50..."
-        html = activity.data["landingPageHtml"]
-        next unless html.present?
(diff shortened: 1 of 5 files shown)


services/QuillLMS/lib/tasks/temporary/translate.rake
```diff
@@ -7,106 +7,318 @@ namespace :translate do
...
-  task activities: :environment do
-    activities = Activity.where("id in (?)", activity_ids).limit(50)
-    CSV.open("open_ai_spanish_activities.csv", "wb") do |csv|
-      csv << ["english", "spanish"]
-      activities.each_with_index do |activity, index|
-        puts "translating #{index + 1}/50..."
-        html = activity.data["landingPageHtml"]
-        next unless html.present?
+  desc 'translate activities and their questions'
+  task :activities, [:limit] => :environment do |t, args|
```
anathomical: Not worth dealing with in this PR, but it may be worth considering writing some of our future tasks in `thor` instead of `rake`.  We've been using more and more `thor` recently as it generally has much cleaner syntax for taking command line args.  That said, this is totally reasonable, and no reason to change approaches in the middle of a project here.

services/QuillLMS/lib/tasks/temporary/translate.rake
```diff
@@ -7,106 +7,318 @@ namespace :translate do
...
+    limit = args[:limit] ? args[:limit].to_i : nil
+    activities = Activity.where("uid in (?)", activity_uids).limit(limit)
```
anathomical: I think that you should just be able to use `where(uid: activity_uids)` and have ActiveRecord automatically figure out the `IN` syntax.

anathomical (approved):
Looks good to me.  I appreciate the way you've got the tests structured here, too.

PR #12034, opened 2024-07-09
lms/update-mat-view-for-new-column-type

anathomical:
## WHAT
Update queries to match new Airbyte behavior
## WHY
Airbyte has changed the way that it ETLs SQL `Array<Integer>` values so that they're now landing in BigQuery as JSON data.  We want our queries to work with the way data is actually getting put into BigQuery, and our spec behavior matches what Airbyte is actually doing when transferring data.
## HOW
- Explicitly parse the value from JSON into a scalar before calling `UNNEST` in queries.
- Update the spec test runner to explicitly make this JSON data in spec setup so that specs match real life expectations around data shape

- Use new materialized view fallback in local environment to ensure that it runs and that data looks reasonable

PR Checklist | Your Answer
------------ | -------------
Have you deployed to Staging? | No.  Staging using production BigQuery

### DIFF
--- a/services/QuillLMS/app/queries/admin_diagnostic_reports/post_diagnostic_assigned_query.rb
+++ b/services/QuillLMS/app/queries/admin_diagnostic_reports/post_diagnostic_assigned_query.rb
@@ -10,7 +10,7 @@ module AdminDiagnosticReports
 
     def from_and_join_clauses
       super + <<-SQL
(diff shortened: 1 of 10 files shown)


happythenewsad (approved):
Nice work!

brendanshean (approved):
Thanks for the quick fix!

PR #12036, opened 2024-07-09
Use sidekiq to run translations for OpenAI

samanthamjohn:
## WHAT
Make a Sidekiq job to do OpenAI translations so that we can translate a lot of things at once. 
## WHY
The translations were too slow to run everything all at once from the console, so to do a lot of translations we'll need to split them out. 
## HOW
Added Sidekiq job to run the translations. 

### Notion Card Links
https://www.notion.so/quill/Translate-via-sidekiq-jobs-for-OpenAI-7c5ab4b9256f422e988b02c37a64aeb4?pvs=4

Ran the translations on local computer

PR Checklist | Your Answer
------------ | -------------


services/QuillLMS/app/workers/open_ai/translate_worker.rb
```diff
@@ -0,0 +1,23 @@
+# frozen_string_literal: true
+
+require_dependency '../../services/concerns/open_ai/api'
+
+module OpenAI
+  class TranslateWorker
+    include Sidekiq::Worker
+
+    sidekiq_options queue: SidekiqQueue::LOW
```
brendanshean, line 9: Let's replace the case statement with a hash.

You could also maybe infer this mapping via the concern, but this seems fine for now.

```
    TRANSLATABLE_TYPES = {
      'Activity' => Activity,
      'Question' => Question,
      'ConceptFeedback' => ConceptFeedback
    }.freeze

    def perform(translatable_id, translatable_type)
      TRANSLATABLE_TYPES[translatable_type]
        &.find_by(id: translatable_id)
        &.translate!
    end
```

services/QuillLMS/app/workers/open_ai/translate_worker.rb
```diff
@@ -0,0 +1,23 @@
+# frozen_string_literal: true
+
+require_dependency '../../services/concerns/open_ai/api'
```
brendanshean, line 3: `require_dependency` should not be needed anymore due to [zeitwerk](https://guides.rubyonrails.org/classic_to_zeitwerk_howto.html#delete-require-dependency-calls).  Let me know if you have issues with removing it.

services/QuillLMS/app/workers/open_ai/translate_activity_and_questions_worker.rb
```diff
@@ -0,0 +1,18 @@
...
+  class TranslateActivityAndQuestionsWorker
+    include Sidekiq::Worker
+
+    sidekiq_options queue: SidekiqQueue::LOW
+
+    def perform(activity_id)
+      activity = Activity.find_by(id: activity_id)
+      return unless activity.present?
+
+      TranslateActivityAndQuestions.run(activity)
```
brendanshean: Take or leave comment.

Since the `TranslateActivityAndQuestions` is pretty short, I would probably just leave the two lines 
```
activity.translate!
activity.questions.each(&:translate!)
```
and get rid of the service object. (unless you anticipate the service object getting bigger)

services/QuillLMS/spec/workers/open_ai/translate_worker_spec.rb
brendanshean, line 1: Using the 
```
  TRANSLATABLE_TYPES = {
      'Activity' => Activity,
      'Question' => Question,
      'ConceptFeedback' => ConceptFeedback
    }.freeze
```
mentioned earlier, I think you can refactor this as for loop over those elements.

services/QuillLMS/spec/services/open_ai/translate_activity_and_questions_spec.rb
```diff
@@ -0,0 +1,28 @@
+# frozen_string_literal: true
+
+require 'rails_helper'
+RSpec.describe OpenAI::TranslateActivityAndQuestions, type: :service do
+  describe "run" do
+    subject { described_class.run(activity)}
+
+    let(:activity) { create(:activity)}
+
+    before do
```
brendanshean: RSpec [recommends](https://rspec.info/features/3-12/rspec-mocks/old-syntax/any-instance/) not using 'allow_any_instance_of`.

Instead you will want to use explicit calls for each object.

services/QuillLMS/spec/models/activity_spec.rb
```diff
@@ -802,4 +801,50 @@
...
+      let(:questions) { create_list(:question, 2)}
+
+      before do
+        activity.data["questions"] = questions.map{|q| {"key" => q.uid}}
+        activity.save
+      end
+
+      it { expect(subject).to be_a(ActiveRecord::Relation) }
+
+      it 'returns all the questions' do
```
brendanshean: Instead of questions.first,questions.last use 
`expect(subject).to match_array(questions)`

services/QuillLMS/spec/models/activity_spec.rb
```diff
@@ -802,4 +801,50 @@
...
+        expect(subject.last).to eq(questions.last)
+      end
+
+      context 'one of the questions does not exist' do
+        before do
+          activity.data["questions"] << {"key" => "124"}
+          activity.save
+        end
+
+        it "returns only the existing questions" do
```
brendanshean: Instead of questions.first,questions.last use 
`expect(subject).to match_array(questions)`

services/QuillLMS/app/models/activity.rb
```diff
@@ -322,14 +322,16 @@ def serialize_with_topics_and_publication_date
...
 
-    question_uids = data["questions"]&.map{|q| q["key"]}
-    questions = Question.where("uid in (?)", question_uids)
-
     questions.each do |question|
       question.update_flag("production")
     end
   end
 
+  def questions
```
brendanshean: A one-liner seems appropriate here.  `def questions = Questions.where(uid: question_uids)`

services/QuillLMS/app/models/activity.rb
```diff
@@ -322,14 +322,16 @@ def serialize_with_topics_and_publication_date
   def update_questions_flag_status_if_necessary!
     return unless flags.include?(:production)
 
-    question_uids = data["questions"]&.map{|q| q["key"]}
```
brendanshean: I like these two extractions.

brendanshean (approved):
Some refactoring comments, but nothing major that requires a re-review.

services/QuillLMS/app/workers/open_ai/translate_worker.rb
samanthamjohn: it was required in this case--maybe we can look into a better config for zeitwork though. It seemed weird to me that workers were not able to autoload files from the `services` directory. I played around a bit and this was the one I found that worked.

services/QuillLMS/app/workers/open_ai/translate_activity_and_questions_worker.rb
samanthamjohn: Good catch, it started with more logic and got smaller as I factored more into `Activity`.

services/QuillLMS/spec/workers/open_ai/translate_worker_spec.rb
samanthamjohn: In this case I'm going to leave it explicit in the tests. I could see it getting more cumbersome later and we can revisit, but for me it feels too meta for testing.

PR #12038, opened 2024-07-09
Use single quotes everywhere and enforce with rubocop

samanthamjohn:
## WHAT
Make rubocop enforce using single quotes

## WHY
So we don't have to think about whether to use single or double quotes at any given moment. 

## HOW
Added the stringLiteral rule to rubocop.yml and ran rubocop with autocorrect

Making sure tests pass

PR Checklist | Your Answer
------------ | -------------
Have you deployed to Staging? | Not yet, deploying now


services/QuillLMS/.rubocop.yml
brendanshean, line 1: Let's add this instead to rubocop base so that it also applies to QuillCMS:

https://github.com/empirical-org/Empirical-Core/blob/develop/services/.rubocop_base.yml

brendanshean: At the top of .rubocop.yml, note the inherits:  https://github.com/empirical-org/Empirical-Core/blob/a2a441de0cbc209289fd7d16df12becf369255a6/services/QuillLMS/.rubocop.yml#L2

brendanshean (approved):
Thank you for standardizing this!

PR #12058, opened 2024-07-11
Add a UID to identify feedback for translations

samanthamjohn:
## WHAT
Add a UID to `incorrectSequences` and `focusPoints` that are stored as arrays within the `questions.data` jsonb column. 

## WHY
So that they can be uniquely identified and mapped to a translation. 

## HOW
Add the UIDs all at once in a migration script, and then whenever you add a new record, we also add a uid. 

### Notion Card Links
[RFC for data migration](https://www.notion.so/quill/RFC-Give-all-incorrectSequences-unique-keys-8d492736b76e45c787fd3fdaa5cb880f)

Ran the script locally to ensure that it works properly. Then tested out updating sequences on questions that have an array of incorrectSequences 

PR Checklist | Your Answer
------------ | -------------


services/QuillLMS/app/models/question.rb
```diff
@@ -178,6 +210,12 @@ def self.translatable_field_name = 'instructions'
     SecureRandom.uuid
   end
 
+  private def time_based_uuid
```
emilia-friedberg: Is this getting used?

services/QuillLMS/app/models/question.rb
```diff
@@ -127,35 +116,38 @@ def get_incorrect_sequence(incorrect_sequence_id)
...
+    add_data_for(type: 'incorrectSequences', new_data:)
   end
 
-  def set_incorrect_sequence(incorrect_sequence_id, new_data)
-    data['incorrectSequences'] ||= {}
-    incorrect_sequence_id = incorrect_sequence_id.to_i if stored_as_array?('incorrectSequences')
-    data['incorrectSequences'][incorrect_sequence_id] = new_data
-    save
+  def add_focus_point(new_data)
+    add_data_for(type: 'focusPoints', new_data:)
```
emilia-friedberg: I think we should add a constant for 'focusPoints' and 'incorrectSequences' since they're being used architecturally here, even though the previous version of this code didn't do that.

services/QuillLMS/lib/tasks/migrate_incorrect_sequences_and_focus_points.thor
```diff
@@ -0,0 +1,17 @@
+# frozen_string_literal: true
+
+require_relative '../../config/environment'
+
+class MigrateIncorrectSequencesAndFocusPoints < Thor
```
emilia-friedberg, line 5: I think we generally prefer Rake to Thor, though I'm not sure if there's a rationale for this or if it's just convention. I think it might be worth considering using Rake here just to not have this outlier.

services/QuillLMS/app/models/question.rb
```diff
@@ -168,6 +160,46 @@ def connect_sentence_combining?
...
+  private def add_data_for(type:, new_data:)
+    if stored_as_array?(type)
+      id = data[type].length
+    else
+      id = new_uuid
+    end
+    return id if set_data_for(type:, id:, new_data:)
+  end
+
+  private def add_uid(data_array:)
```
emilia-friedberg: Maybe not worth doing anything about, but naming this `data_array` when it's not always an array (as evidenced by the next line) confused me a little. Maybe just `data`?

services/QuillLMS/spec/models/question_spec.rb
```diff
@@ -189,7 +189,8 @@
     end
 
     it 'should remove the specified incorrectSequence if stored in an array' do
-      question.update_incorrect_sequences([{'text'=>'text', 'feedback'=>'bar'}, {'text'=>'text', 'feedback'=>'bar'}, {'text'=>'text', 'feedback'=>'bar'}])
+      data = [{'text'=>'text', 'feedback'=>'bar'}, {'text'=>'text', 'feedback'=>'bar'}, {'text'=>'text', 'feedback'=>'bar'}]
+      question.data['incorrectSequences'] = data
```
emilia-friedberg: Why did these need to be rewritten to stop using the `update_incorrect_sequences` function?

emilia-friedberg (approved):
This is a good, much simpler solution to a thorny problem. I left a couple of minor comments but nothing I need to re-review if updating and adding sequences and focus points looks good in QA. Thank you for your work on this!

services/QuillLMS/app/models/question.rb
samanthamjohn: I'd like to switch to time-based UIDs in general, but it's a larger convo and not needed to do this PR!

services/QuillLMS/lib/tasks/migrate_incorrect_sequences_and_focus_points.thor
samanthamjohn: @anathomical told me we prefer `thor`, haha! I'll put it as a topic in the engineering meeting.

services/QuillLMS/spec/models/question_spec.rb
samanthamjohn: `update_incorrect_sequences` now adds a UID if there isn't one, so the data no longer matches in the test expectation. That said, I could manually add UIDs to the data here instead or have the matcher ignore the UIDs.

samanthamjohn: this was the easiest change, but probably adding the UIDs is more correct.

PR #12060, opened 2024-07-11
Fix unit template ordering

brendanshean:
## WHAT
Fix the an ordering [bug](https://www.notion.so/quill/Order-of-recommended-activities-shuffling-b4096d912d794a3ca6186916707f6e98?pvs=4) involving activities associated with a unit template.

## WHY
The existing code doesn't actually create any new ActivitiesUnitTemplate records or store ordering.  Instead, it only reorders existing ActivitiesUnitTemplate records that were created via a [seed file](https://github.com/empirical-org/Empirical-Core/blob/develop/services/QuillLMS/db/seeds/activities_unit_templates.sql).

## HOW
Add a loop in the create/update actions that iterates over the activity ids and creates the necessary record.  In the case of update, first delete all existing ActivitiesUnitTemplate recrords.

### Screenshots

### Notion Card Links
https://www.notion.so/quill/Order-of-recommended-activities-shuffling-b4096d912d794a3ca6186916707f6e98?pvs=4

I tried this out on staging and had Rachel confirm that the functionality is working properly.

PR Checklist | Your Answer
------------ | -------------

### DIFF
--- a/services/QuillLMS/app/controllers/cms/unit_templates_controller.rb
+++ b/services/QuillLMS/app/controllers/cms/unit_templates_controller.rb
@@ -28,8 +28,12 @@ class Cms::UnitTemplatesController < Cms::CmsController
   end
 
   def create
-    @unit_template = UnitTemplate.new(unit_template_params)
+    @unit_template = UnitTemplate.new(unit_template_params.except(:activity_ids))
     if @unit_template.save!
+      unit_template_params[:activity_ids].each_with_index do |activity_id, order_number|
+        ActivitiesUnitTemplate.create(activity_id:, order_number:, unit_template_id: @unit_template.id)
+      end
+
       render json: @unit_template
     else
       render json: {errors: @unit_template.errors}, status: 422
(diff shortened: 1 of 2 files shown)


services/QuillLMS/app/controllers/cms/unit_templates_controller.rb
```diff
@@ -28,23 +28,26 @@ def new
...
   def update
-    params = unit_template_params
-    @unit_template.activities = []
-    @unit_template.save
-    if @unit_template.update(params)
-      @unit_template.activities_unit_templates.each do |aut|
-        order_number = params['activity_ids'].index(aut.activity_id)
-        aut.update(order_number: order_number)
+    if @unit_template.update(unit_template_params.except(:activity_ids))
+      @unit_template.activities_unit_templates.destroy_all
```
anathomical, line 45: Destroying and recreating records like this feels a little awkward, but given how rarely staff touch this data, I can't imagine that this is worth being particularly worried about.

anathomical (approved):

PR #12064, opened 2024-07-12
learn worlds - username updates

happythenewsad:
## WHAT
- update LearnWorlds sync API to use 'name' instead of 'email' for the LW username
- backfill script for existing users 

## WHY
- so that user emails are not publicized within our LW instance 

### Screenshots

### Notion Card Links
https://www.notion.so/quill/LearnWorlds-Update-Change-process-to-use-Quill-name-instead-of-Quill-email-for-LearnWorlds-Username-85799021ff52407b8b58829355fd58df

1. Sign up as new LW user; confirm that email is not not displayed as the username within LW
2. run backfill task on staging

PR Checklist | Your Answer
------------ | -------------


services/QuillLMS/lib/tasks/learn_worlds.rake
```diff
@@ -0,0 +1,22 @@
+# frozen_string_literal: true
```
happythenewsad: This rake task is deletable after a single run

anathomical: Is it work putting in `lib/tasks/temporary` since it's intended to run once?

anathomical (approved):
One thought about location for the rake task code, but this looks reasonable to me as-is.

services/QuillLMS/app/lib/utils/string.rb
```diff
@@ -10,6 +10,8 @@ def self.last_name_or_name(first_name_last_name_string)
     last || first
   end
 
+  def self.to_username(str) = str.downcase.strip.gsub(/\s/, '.')
```
brendanshean: I think this method belongs in a LearnWorlds namespace.  app/lib suggests this is a general method that applies to all strings.  I don't think there's a generally accepted to_username method.

brendanshean (approved):
Looks good to me!

PR #12071, opened 2024-07-14
GenAI Updates Part 7: Create a “Prompt Template Library” Page

brendanshean:
## WHAT
* Add an index page for LLM Prompt Templates
* Add editable notes attribute to `LLMPromptTemplate`
* Add 'Preview' mode for `LLMPromptTemplate` contents

## WHY
* It's useful to see what templates are available
* While the prompt contents are read-only, it's useful to add notes over time
* Toggling between template view and the filled in prompt with colors can help elucidate the pattern

## HOW
* Add route, template, and controller action
* Add migration
* Iterate through contents for substitutions and replace with `"<span style='color: #{color};'>#{substitution}</span>")`.  Then use the LLMPromptBuilder to fill in content with a generic activity and its attributes.

### Screenshots

### Notion Card Links
https://www.notion.so/quill/Internal-Tool-Organizing-Trials-into-Datasets-8dbc188caf194157a2a28395e907ddc0?pvs=4#080cdc9162c843cdbccd9b91d9ff8fc0

Tested on staging

PR Checklist | Your Answer
------------ | -------------


services/QuillLMS/engines/evidence/app/models/evidence/research/gen_ai/llm_prompt.rb
```diff
@@ -75,7 +75,7 @@ def prompt_with_student_response(student_response)
 
         def guidelines_count = optimal_guidelines_count + suboptimal_guidelines_count
 
-        def to_s = description.truncate(40)
+        def to_s = name.truncate(40)
```
dandrabik: Just a note, this is probably fine for now, but might be something that causes issues if long names prove helpful for distinguishing between subtly different prompts (but this truncate cuts off the part that distinguishes it, e.g. `"- V2"`. Might be better to wrap this in the UI rather than cut it off.

dandrabik (approved):
Looks good. Small note about potential issues with truncation that we could always deal with later.

services/QuillLMS/engines/evidence/app/models/evidence/research/gen_ai/llm_prompt.rb
brendanshean: Good point.  I will remove.

PR #12072, opened 2024-07-15
Add editing to translated texts admin

samanthamjohn:
## WHAT
Updated the admin interface for translations and allow you to edit them. 
## WHY
So that when we check over the translations we can make little edits as needed. 
## HOW
Used the `TranslatedText` model as the base of the admin translation table and added an update action to the controller. Wrote some js to allow you to edit in place. 

### Notes
- I chose to use pretty vanilla rails and JS since this is a one-off admin page that isn't connected to anything else. In the future we could write something in react if allowing admins to edit translations becomes a priority. 
- We originally talked about doing this via a CSV upload and download. But it was easier to build it through rails and it makes for a better UX for the staff. 

### Screenshots

Viewing Translations | Editing Translations |
-- | -- |
2024-07-15 at 9 27 AM |  2024-07-15 at 9 27 AM |

### Notion Card Links
[Allow admins to edit translations](https://www.notion.so/quill/Allow-admins-to-edit-translations-869e5f4fb05b4912a8f209b97b2c1a54?pvs=4)

Tested it locally by viewing and editing translations as a staff member.

PR Checklist | Your Answer
------------ | -------------


services/QuillLMS/app/controllers/translated_texts_controller.rb
```diff
@@ -0,0 +1,28 @@
+# frozen_string_literal: true
+
+class TranslatedTextsController < ApplicationController
+  def index
+    return redirect_to profile_path unless staff?
```
brendanshean: Let's remove this guard clause and use `before_action :staff!` at the top of this controller.

services/QuillLMS/app/controllers/translated_texts_controller.rb
```diff
@@ -0,0 +1,28 @@
...
+      .all
+      .includes(:english_text, :translation_mappings)
+      .page(params[:page])
+      .per(200)
+
+    @js_file = "entrypoints/snippets/translated_text.ts"
+  end
+
+  def update
+    return redirect_to profile_path unless staff?
```
brendanshean: Let's remove this guard clause and use `before_action :staff!` at the top of this controller.

services/QuillLMS/app/controllers/translated_texts_controller.rb
anathomical: Purely stylistic, so take or leave: This being in both the `index` and `update` action suggest to me that we might just want a `before_action` to handle the auth stuff.

services/QuillLMS/app/models/translated_text.rb
```diff
@@ -14,12 +14,19 @@
...
 
   scope :ordered_by_source_api, lambda { |source_api = Translatable::OPEN_AI_SOURCE|
     source_api = Translatable::OPEN_AI_SOURCE unless Translatable::SOURCES.include? source_api
     order(
       Arel.sql("CASE WHEN source_api = '#{source_api}' THEN 0 ELSE 1 END, source_api ASC")
     )
   }
+
+  def source
+    translation_mappings.first&.source_type
```
anathomical, line 30: I doubt that this is likely to come up in real-world scenarios, so may not be something we have to worry about, but should we consider possible edge cases where a `translated_text` record will have multiple `translation_mappings`, but different `sources` from them?

services/QuillLMS/config/routes.rb
```diff
@@ -84,6 +84,12 @@
 
   resources :student_feedback_responses, only: [:create]
 
+  resources :translated_texts, only: [:index, :update] do
+    collection do
+      get :download_csv
```
anathomical: Is this still a real resource?  I don't remember seeing it in the controller.

services/QuillLMS/client/app/entrypoints/snippets/translated_text.ts
```diff
@@ -0,0 +1,78 @@
...
+        headers: {
+          'Content-Type': 'application/json',
+          'X-CSRF-Token': document.querySelector('meta[name="csrf-token"]').content
+        },
+        body: JSON.stringify({ translated_text: { translation: newText } })
+      })
+        .then(response => response.json())
+        .then(data => {
+          if (data.success) {
+            row.querySelector('.text-content').innerHTML = newText;
```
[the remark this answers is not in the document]
brendanshean: I think sanitizing the string will get rid of this warning:

```
.then(response => response.json())
.then(data => {
  if (data.success) {
    const escapeHTML = (str) => {
      const div = document.createElement('div');
      div.appendChild(document.createTextNode(str));
      return div.innerHTML;
    };

    const sanitizedText = escapeHTML(newText);
    row.querySelector('.text-content').innerHTML = sanitizedText;
  }
})
```

services/QuillLMS/client/app/entrypoints/snippets/translated_text.ts
```diff
@@ -0,0 +1,78 @@
+function adjustTextareaHeight(textarea) {
+  textarea.style.height = 'auto';
+  textarea.style.height = (textarea.scrollHeight) + 'px';
+}
+
+function initializeTranslatedTexts() {
```
brendanshean: This method feels like its doing to many things and probably should be refactored.  I'll be curious to hear what others think.

services/QuillLMS/client/app/entrypoints/snippets/translated_text.ts
```diff
@@ -0,0 +1,78 @@
...
+          }
+        })
+        .catch(error => {
+          alert('An error occurred while updating the translation');
+        });
+    }
+  });
+}
+
+// Try to run the initialization immediately
```
brendanshean: I feel like this comment is unnecessary

services/QuillLMS/client/app/entrypoints/snippets/translated_text.ts
```diff
@@ -0,0 +1,78 @@
...
+initializeTranslatedTexts();
+
+// Also run on turbolinks:load if it exists
```
brendanshean: I don't think we're [using](https://github.com/empirical-org/Empirical-Core/pull/12081) turbolinks so this probably isn't needed.

services/QuillLMS/client/app/entrypoints/snippets/translated_text.ts
```diff
@@ -0,0 +1,78 @@
...
+document.addEventListener('turbolinks:load', initializeTranslatedTexts);
+
+// For cases where neither immediate execution nor turbolinks work,
+// fall back to DOMContentLoaded
+document.addEventListener('DOMContentLoaded', initializeTranslatedTexts);
```
anathomical, line 111: Since this is our final fallback, is there a reason not to just use this listener and skip instant initialization?  There might be a good reason, but I feel like we should aim to support one approach or the other rather than trying to do both.

services/QuillLMS/client/app/entrypoints/snippets/translated_text.ts
```diff
@@ -0,0 +1,78 @@
+function adjustTextareaHeight(textarea) {
```
anathomical, line 2: I don't object to skipping React for this, but wonder how the front-end folks (@eadams17, @emilia-friedberg) feel about it.  Not sure if we want to be long-term committed to supporting this, but maybe it's a reasonable short-term solution and if we ever need to enhance or fix it we'd just replace it with something more robust.

services/QuillLMS/client/app/entrypoints/snippets/translated_text.ts
anathomical: I think I agree.  I'd be inclined to take the code in each of the `if` blocks and make it separate function so that you just end up with a bunch of `if` calls that simply pass `target` through to a function to actually handle the case.

services/QuillLMS/spec/models/translated_text_spec.rb
```diff
@@ -51,4 +52,33 @@
...
+  end
+
+  describe '#source' do
+    subject { translated_text.source }
+
+    let(:translated_text) { create(:translated_text)}
+
+    context 'translation mapping exists' do
+      it 'is the source of the first translation_mapping' do
+        translation_mapping = create(:translation_mapping, english_text: translated_text.english_text)
```
anathomical, line 73: Small stylistic nit-pick that you can take or leave: you could move this `create` call into a `let!` declaration in the containing context.

services/QuillLMS/spec/controllers/translated_texts_controller_spec.rb
```diff
@@ -0,0 +1,101 @@
...
+require 'rails_helper'
+
+RSpec.describe TranslatedTextsController, type: :controller do
+  let(:staff_user) { create(:user, role: 'staff') }
+  let(:regular_user) { create(:user) }
+
+  describe 'GET #index' do
+    context 'when user is staff' do
+      before do
+        allow(controller).to receive(:signed_in?) { true }
```
anathomical: This `allow` looks like it's declared in every context, so I think you might want to declare a `before` block with this `allow` higher up in the hierarchy so you only need it once.  There's some similarities with the `current_user` allow as well, but I note that it's not always the same user, so maybe that doesn't need to be moved up.

anathomical (approved):
A couple of style-focused comments, but generally speaking I think this is a solid approach.  I also appreciate the thoroughness of the tests.

services/QuillLMS/app/controllers/translated_texts_controller.rb
samanthamjohn: perfect! I was looking for something like that and did not find it. Also gonna use it in the `pages_controller`

services/QuillLMS/app/models/translated_text.rb
samanthamjohn: yeah, I was thinking about that. I think for a first pass it's ok, and also that the source_type isn't super important information here.

services/QuillLMS/config/routes.rb
samanthamjohn: nope, forgot to remove it.

services/QuillLMS/client/app/entrypoints/snippets/translated_text.ts
emilia-friedberg: @anathomical yeah, I don't think this is a pattern we want to follow in the future -- I don't really like the idea of having a bunch of vanilla Javascript files floating around, which force us to do things like the native implementation of fetch Samantha has here rather than the shared `requestX` functions we use elsewhere -- but as effectively a prototype I think it's okay given the short timetable Samantha has for this project. My preference would be to rewrite it whenever we next need to make a change.

samanthamjohn: sounds good. Something I've been thinking is that the react app is actually quite separate from the backend and could even eventually be taken out of rails and deployed on its own. I think there might be a few advantages to that, and we could simplify our pipeline. But obviously a much longer-term project that we'd have to discuss as a team.

PR #12083, opened 2024-07-16
Translate feedback from LMS + CMS

samanthamjohn:
## WHAT
When you translate a `Question`, also translate its `focusPoints` and `incorrectSequences` and responses from the CMS

## WHY
so we can also have translations for feedback! 

## HOW
Add `translation_mappings` for `focusPoints`, `incorrectSequences`, and 'cms_responses' in the `questions`

### Notion Card Links
[Add translations for feedback](https://www.notion.so/quill/Add-translations-for-feedback-from-LMS-789789716c574163adfbfe1c53e31ef2?pvs=4)
[Add translations for CMS feedback ](https://www.notion.so/quill/Add-translations-for-feedback-from-CMS-1cb233374f464c2b820a28b03130e9c3?pvs=4)

Ran the `translate!` command on a question to check that it translates all the feedback. 

PR Checklist | Your Answer
------------ | -------------


services/QuillLMS/app/models/activity.rb
```diff
@@ -254,7 +254,7 @@ def data_as_json
     data
   end
 
-  def self.translatable_field_name = 'landingPageHtml'
+  def self.default_field_name = 'landingPageHtml'
```
anathomical: This new name feels rather unspecific.  If you see that it comes from the `Translatable` concern, it makes sense, but in a case like this where it's declared hundreds of lines after the concern inclusion, it seems unlikely that most people would realize what this declaration even does.  We don't need the old name exactly, but something that lets someone reading this file understand what this declaration might be doing would be nice.

services/QuillLMS/app/models/concerns/translatable.rb
```diff
@@ -15,23 +15,27 @@ module Translatable
...
+    class_attribute :default_field_name, default: nil
   end
 
   def translated_json(options = {})
     source_api = options&.dig(:source_api) || Translatable::OPEN_AI_SOURCE
     translation_text = translation(source_api: source_api)
     return data unless translation_text.present?
 
-    data.merge({"translated#{translatable_field_name.capitalize}" => translation_text})
+    data.merge({"translated#{default_field_name.capitalize}" => translation_text})
```
anathomical: This is a longer-term concern, and not something we need to fix at this moment, but I read this as making the `Translatable` concern as only really working with models that have a `data` JSON column, which I imagine will not always be the case.  I suppose we can deal with that if/when it happens.

services/QuillLMS/app/models/question.rb
```diff
@@ -165,7 +165,38 @@ def connect_sentence_combining?
...
+    data[type]&.transform_values { |v| v['feedback'] }
+      &.compact
+      &.transform_keys { |k| "incorrectSequences.#{k}" } || {}
+  end
+
+  private def translatable_for_array(type)
+    data[type].each_with_object({}) do |item, hash|
+      hash["incorrectSequences.#{item['uid']}"] = item['feedback']
+    end
+  end
```
anathomical: Take or leave this, but I wonder if there's an easier-to-read approach here where there's a method to force the shape of the array version of the data to be the same as the hash version of the data, and then a second method that does the translation-focused transform.  That way if you tweak the logic of the transform you don't have to do it in two places.

Also, I note that this looks like it hard-codes `incorrectSequences` into the result key names.  Does this work with Focus Points?

services/QuillLMS/spec/models/question_spec.rb
```diff
@@ -526,4 +526,105 @@
...
+      expect(question).to receive(:create_data_translation_mappings).with(type: Question::FOCUS_POINTS)
+      expect(question).to receive(:create_data_translation_mappings).with(type: Question::INCORRECT_SEQUENCES)
+      subject
+    end
+
+    it 'creates translation_mappings for instructions' do
+      subject
+      instructions_mapping = question.translation_mappings.find_by(field_name: "instructions")
+      expect(instructions_mapping.text).to eq(instructions)
+    end
```
anathomical: This may not actually need a change, but I did want to point out that it feels a little weird to test for mappings in two different ways.  For FocusPoints and IncorrectSequences we confirm that stubbed methods are called, but we check for database artifacts to be created for Instructions.  Given that you have coverage for `create_data_translation_mappings` below, that may be just fine, but I wanted to call out that this tweaked a sort of "that seems weird" moment from me.

services/QuillLMS/spec/models/question_spec.rb
```diff
@@ -526,4 +526,105 @@
...
+    subject {question.create_data_translation_mappings(type: )}
+
+    let(:type) { Question::FOCUS_POINTS }
+    let(:question) { create(:question)}
+    let(:translatable_data) { { 'focusPoints.uid1' => 'f1', 'focusPoints.uid2' => 'f2'}}
+
+    before do
+      allow(question).to receive(:translatable_data)
+        .with(type:)
+        .and_return(translatable_data)
```
anathomical: Stubbing the method that's under test here feels like it may be hiding bugs.  For instance, I flagged the hard-coding of `incorrectSequence` into the method in the model, which means I'm not sure that you can actually get the code as-written to produce the `translatable_data` shape that you're declaring in this spec.  If I'm right in that assumption, that failure is hidden by stubbing the call here.

This may actually be reasonable since you have an explicit set of specs below focused on `translatable_data` below, so this may not need to actually be "fixed", but again I wanted to call out that this felt risky to me.

anathomical (commented):
The general structure of this code feels right to me.  I've called out a couple of places where naming feels like it could be improved, or where there are some minor places where structure feels a little weird.

I'm holding off on approving this until you get a chance to either make updates, or until you just tell me why you think some things don't need changing.

services/QuillLMS/app/models/concerns/translatable.rb
samanthamjohn: I think I'm gonna change this eventually because I'm redoing the APIs. Is it ok if we hold off on changing this method til I do [this card](https://www.notion.so/quill/Add-an-API-endpoint-to-get-all-concept_feedback-for-each-language-b4da66f858844d1f9990f21ba3f64122?pvs=4)? ConceptFeedback is the only place that uses it right now anyways.

services/QuillLMS/app/models/question.rb
samanthamjohn: Good catch! will fix.

services/QuillLMS/app/models/concerns/translatable_question.rb
```diff
@@ -0,0 +1,66 @@
+# frozen_string_literal: true
```
anathomical, line 1: I like moving this code into its own module.  Good call.

services/QuillLMS/app/models/concerns/translatable_question.rb
```diff
@@ -0,0 +1,66 @@
...
+  end
+
+  private def cms_responses
+    response_body = JSON.parse(cms_data.body)
+
+    transform_translatable_array(response_body, CMS_RESPONSES, 'id')
+  end
+
+  private def cms_url = "#{ENV['CMS_URL']}/questions/#{uid}/responses"
+  private def cms_data
```
anathomical: This is purely stylistic, so don't feel like you have to do this, but I like one-liners even for memoized methods like this one.  That said, I know some people find the `= ... ||=` construction to be awkward, so leaving this as-is is fine.
```ruby
private def cms_data = @cms_data ||= HTTParty.get(cms_url)
```

services/QuillLMS/app/models/question.rb
```diff
@@ -52,8 +52,9 @@ class Question < ApplicationRecord
     TYPE_GRAMMAR_QUESTION => 'grammar_questions',
   }
 
-  INCORRECT_SEQUENCES = 'incorrectSequences'
-  FOCUS_POINTS = 'focusPoints'
+  INCORRECT_SEQUENCES = TranslatableQuestion::INCORRECT_SEQUENCES
+  FOCUS_POINTS = TranslatableQuestion::FOCUS_POINTS
+  CMS_RESPONSES = TranslatableQuestion::CMS_RESPONSES
```
anathomical: I imagine that you have these here because something doesn't work without them, but I would have expected these constants to be set via the inclusion of `TranslatableQuestion`.  Weird that you have to declare them again.

anathomical (approved):
Thanks for the quick turn-around on feedback.

services/QuillLMS/app/models/concerns/translatable_question.rb
samanthamjohn: I like it !

services/QuillLMS/app/models/question.rb
samanthamjohn: They are set, but I thought it looked nicer to not have to prefix them since they're used so much in both files. That said, CMS_RESPONSES is not needed here.

PR #12084, opened 2024-07-16
Use data accessors for questions

samanthamjohn:
## WHAT
Use `data_accessor` so that we can access various properties that are inside `question.data` by name rather than relying on strings. 
## WHY
This is more robust and less error-prone than using strings. Plus the code looks nicer. 

Checked that the site worked on dev

PR Checklist | Your Answer
------------ | -------------


anathomical (approved):
Huge fan of this update.  Much more readable.

PR #12085, opened 2024-07-16
adgr-configure-materialized-view-for-fallback

anathomical:
## WHAT
Properly register materialized view with query class
## WHY
So that it can automatically trigger fallback code in cases where the view is erroring
## HOW
Register the materialized view with the class so that it understands how to construct a fallback query when a failure is caused by a missing materialized view.

- Run query in console to ensure that it does not error
- Deploy to staging and confirm queries do not error

PR Checklist | Your Answer
------------ | -------------
Have you added and/or updated tests? | No tests on this type of fallback code


happythenewsad (approved):
Thanks for addressing 🚢

dandrabik (approved):
Thanks for fixing this.

PR #12086, opened 2024-07-16
tweak demo account generation script

emilia-friedberg:
## WHAT
Tweak the admin demo account generation script to run more quickly and fix what looked like a potential bug.

## WHY
Alex reported a different issue with the admin demo account today, and as I was investigating it I noticed the admin demo account only had one school associated with it. When I tried to replicate this on staging, I found that it was a) really slow and b) on at least one run it wasn't correctly associating new schools with the most recent instance of the admin demo user. I think that was due to some weird ghost in the machine because the previous user had the same email, but I made these tweaks to a) run more quickly, because I was no longer seeing the Redis memory errors we'd previously experienced on staging that necessitated these long delays in the first place, and b) use the instance variable that it seems like we should have been using in the first place, which appears to have solved the issue.

## HOW
Update default time, add bang operator to school creation so we can see errors more easily, 

### Screenshots

### Notion Card Links
https://www.notion.so/quill/Make-sure-all-schools-are-consistently-created-on-admin-demo-account-1630a1b7976041d992e1b949a580a477?pvs=4

Tested on staging and confirmed that it runs in about 45 minutes and the data looks as it should.

PR Checklist | Your Answer
------------ | -------------

### DIFF
--- a/services/QuillLMS/app/services/demo/create_admin_report.rb
+++ b/services/QuillLMS/app/services/demo/create_admin_report.rb
@@ -3,7 +3,7 @@
 class Demo::CreateAdminReport
 
   RANGE_OF_NUMBER_OF_SESSIONS_TO_DESTROY = 1..20
-  BATCH_DELAY = 1.minute
+  BATCH_DELAY = 10.seconds
   NUMBER_OF_STUDENTS_PER_CLASSROOM = 25
 
   GRADE_MIN = 5
@@ -36,7 +36,7 @@ class Demo::CreateAdminReport
   end
 
   private def find_or_create_school(school_name)
-    School.find_or_create_by(name: school_name, mail_street: school_street)
+    School.find_or_create_by!(name: school_name, mail_street: school_street)
   end
 
   private def find_or_create_teacher_data(teacher_name, school, subscription)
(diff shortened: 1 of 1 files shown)


services/QuillLMS/app/services/demo/create_admin_report.rb
```diff
@@ -68,7 +68,7 @@ def reset
     admin_teacher = User.create!(
       name: 'Toni Morrison',
       role: User::ADMIN,
-      email: teacher_email,
+      email: @teacher_email,
```
anathomical: This change shouldn't be necessary because we have an `attr_accessor` declared for `:teacher_email`.  That should make both the old and new versions of this line the same.  If you're making this change to fix some sort of issue that you were observing in testing, this may be worth an additional look.

anathomical (approved):
I have a minor concern about one of the changes in that I don't think it does anything, and thus if you were running into an issue I can't imagine it actually fixing it.  The other changes look great to me.

happythenewsad (approved):

services/QuillLMS/app/services/demo/create_admin_report.rb
emilia-friedberg: Thank you for calling that out. I'm going to put this back the way it was and see if I can replicate that issue on staging.

PR #12092, opened 2024-07-17
Add the activity translations to the lessons API

samanthamjohn:
## WHAT
Update the API for an individual activity to include translation. 
## WHY
This data will be used to display the translated activity landing page when a user selects a new language. It will also be used to determine which languages are available for a given activity. 
## HOW
Return the `translated_json` for the activity in the `api/v1/lessons/[uid].json` api call. Updated the method to return a hash of languages. 

### Notion Card Links
[Add hash of languages to activity API](https://www.notion.so/quill/Add-a-hash-of-languages-to-the-landingPageHTML-for-each-activity-json-payload-269f2024a320485ea59ffc59f9262c4c?pvs=4)

looked at the network tab when loading activities with and without translations to see that the expected payload was there. 

PR Checklist | Your Answer
------------ | -------------

### DIFF
--- a/services/QuillLMS/app/models/activity.rb
+++ b/services/QuillLMS/app/models/activity.rb
@@ -254,7 +254,15 @@ class Activity < ApplicationRecord
     data
   end
 
+  # translatable
   def self.translatable_field_name = 'landingPageHtml'
+
+  def translated_json(options = {})
+    translations = translated_texts.pluck(:locale, :translation).to_h
+    return data unless translations.present?
+
+    data.merge('translations' => translations)
+  end
 
   def add_question(question)
     return if !validate_question(question)
--- a/services/QuillLMS/spec/models/activity_spec.rb
+++ b/services/QuillLMS/spec/models/activity_spec.rb
@@ -840,4 +840,33 @@ describe Activity, type: :model, redis: true do
       it { expect(subject.count).to eq(0) }
     end
   end
+
+  describe 'translated_json' do
+    subject {activity.translated_json}
+
+    let(:activity) { create(:activity)}
+
+    context 'there are translations' do
+      it 'is data + all available languages' do
+        activity.data.merge!({'landingPageHtml' => "html"})
+        activity.create_translation_mappings
+        chinese_locale = "zh-CN"
+        english_text = activity.english_texts.first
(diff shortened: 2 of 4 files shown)


services/QuillLMS/app/models/activity.rb
```diff
@@ -254,8 +254,16 @@ def data_as_json
     data
   end
 
+  # translatable
   def self.translatable_field_name = 'landingPageHtml'
 
+  def translated_json(options = {})
```
anathomical, line 260: This is purely stylistic, but I wonder if this should be called `json_with_translations` or something, to make it clear that it's additive?  That's just a naming thing, though, so feel free to keep the existing name, too.

services/QuillLMS/spec/models/activity_spec.rb
```diff
@@ -840,4 +840,33 @@
...
+        chinese_locale = "zh-CN"
+        english_text = activity.english_texts.first
+        chinese = create(:translated_text, english_text:, locale: chinese_locale)
+        spanish = create(:translated_text, english_text:, locale: Translatable::SPANISH_LOCALE)
+        expected_data = activity.data.merge(
+          'translations' => {
+            chinese_locale => chinese.translation,
+            Translatable::SPANISH_LOCALE => spanish.translation
+          }
+        )
```
anathomical, line 862: It feels like a lot of this could be moved into `let` and `let!` calls in the context to simplify the `it` block down to just the `expect` call.

anathomical (approved):
A couple of take-or-leave style suggestions, but this is a good solution.

services/QuillLMS/app/models/activity.rb
samanthamjohn: the only thing is that this is also part of the `translatable` module. I like the new name though! I think I'll change it in a later PR when I switch up the API for the concept_feedback.

services/QuillLMS/spec/models/activity_spec.rb
samanthamjohn: I tried it, but unfortunately they all sort of need to happen in order (i.e. can't make the chinese text until we've created the translation mappings) and the let syntax doesn't give you the ordering you'd want.

PR #12101, opened 2024-07-18
Refactor Vitally functions to use global attributes instead of parameters

cissyxyu:
## WHAT
Admin requested a lot of new Vitally attributes to be added to the sync. Before we do that, I want to refactor our Vitally-related functions to stop passing parameters back and forth between functions, and instead use global variables accessible by all the functions. 

## WHY
This is a much cleaner way to calculate data because we no longer need to pass parameters back and forth between functions. The code is more readable and easier to use. The cleanup will make the upcoming Vitally workload much easier.

## HOW
Stop passing these attributes back and forth between functions - `district, school, user, school_year_start, school_year_end`. Instead, make these variables global variables that each subclass must set upon initialization. 

Each shared function module defines only the global variables necessary for their scope. So for example, `VitallySharedFunctions` defines the `vitally_entity` variable which represents either schools, districts or users depending on the subclass. While `VitallySchoolStats` defines only the `school` variable because its implementation assumes that the entity is a `School` record.

### Screenshots

### Notion Card Links
[(Please provide links to any relevant Notion card(s) relevant to this PR.)](https://www.notion.so/quill/Update-Vitally-Data-Sync-New-Diagnostic-Usage-Data-2fb92d4ba89847af9da661fc78710556?pvs=4)

Ran all the rspec tests to ensure everything is working properly. 

PR Checklist | Your Answer
------------ | -------------

### DIFF
--- a/services/QuillLMS/app/models/concerns/vitally_school_stats.rb
+++ b/services/QuillLMS/app/models/concerns/vitally_school_stats.rb
@@ -3,7 +3,9 @@
 module VitallySchoolStats
   extend ActiveSupport::Concern
 
-  private def active_students_query(school)
+  attr_accessor :school
+
+  private def active_students_query
     # use raw SQL to bypass scope limits (visible: true) on classrooms
     ActivitySession.unscoped.select(:user_id)
       .distinct
@@ -15,7 +17,7 @@ module VitallySchoolStats
       .where(state: 'finished').where('schools.id = ?', school.id)
   end
 
-  private def activities_finished_query(school)
+  private def activities_finished_query
     ClassroomsTeacher.joins('JOIN users ON users.id=classrooms_teachers.user_id')
       .joins('JOIN schools_users ON schools_users.user_id=users.id')
       .joins('JOIN schools ON schools.id=schools_users.school_id')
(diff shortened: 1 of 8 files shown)


services/QuillLMS/app/models/concerns/vitally_school_stats.rb
```diff
@@ -3,7 +3,9 @@
 module VitallySchoolStats
   extend ActiveSupport::Concern
 
-  private def active_students_query(school)
+  attr_accessor :school
```
samanthamjohn: This could be an `attr_reader` I believe.

services/QuillLMS/app/models/concerns/vitally_shared_functions.rb
```diff
@@ -3,6 +3,8 @@
 module VitallySharedFunctions
   extend ActiveSupport::Concern
 
+  attr_accessor :vitally_entity, :school_year_start, :school_year_end
```
samanthamjohn: These could be `attr_reader`s I believe.

samanthamjohn (approved):
It looks good to me! 

Just flagged a couple of `attr_accessor`s that could be `attr_reader`s since we never set them outside of the initializer. 

Also, a small thing--in the PR description you describe the new variables as _global_ variables, but in fact, you added a bunch of _instance_ variables.  Let's edit the description to correctly reflect that for posterity, especially because to me, adding a lot of global variables is a code smell, whereas making these instance variables seems very reasonable.

anathomical (approved):
I like moving to instance variables set during `initialize`.  You're right that this feels a lot more parseable.

PR #12108, opened 2024-07-22
content hubs

emilia-friedberg:
## WHAT
Update the Assign Activities and My Reports index pages with cards that link off to the new Content Hub pages (not built yet) and update the Explore Curriculum dropdown with the same link. Also update Backpack with new icons and get the helper in place for the rest of the content hub work.

## WHY
This delivers the first set of cards for the content hubs project.

## HOW
Add the new cards to the various pages and, on the backend, add a controller action and a helper with the relevant data so that we can check to see if any relevant activities have been assigned. Note: I did think about storing this information in the database, but since courseview is coming along which will be a much more complex feature and supplant what we're building here, Jamie and I agreed it probably made sense to just hardcode everything for now.

### Screenshots
Screenshot 2024-07-17 at 9 43 14 AM
Screenshot 2024-07-22 at 9 44 02 AM
Screenshot 2024-07-22 at 9 44 21 AM
Screenshot 2024-07-22 at 9 50 56 AM
Screenshot 2024-07-22 at 9 51 19 AM

### Notion Card Links
https://www.notion.so/quill/Assign-Activities-Page-Add-Cards-for-World-History-Building-AI-Knowledge-Offerings-007eea2e444b4a449974780d83c3a602?pvs=4
https://www.notion.so/quill/My-Reports-Add-Cards-Navigation-Links-for-Social-Studies-Interdisciplinary-Science-Dashboards-418ced232efd44ff95b33dc9f4680223?pvs=4
https://www.notion.so/quill/Explore-Curriculum-Drop-Down-Navigation-Bar-Logged-Out-6c5efbe546f7420498370d2ba5808856?pvs=4

Tested them out on staging in the demo account both before and after assigning a relevant featured activity pack.

PR Checklist | Your Answer
------------ | -------------


services/QuillLMS/app/helpers/content_hubs_helper.rb
```diff
@@ -0,0 +1,298 @@
...
+  def unit_activities_include_content_activities?(unit_activities, content_activity_ids)
+    unit_activities.where(activity_id: content_activity_ids).any?
+  end
+
+  def unit_activities_include_social_studies_activities?(unit_activities)
+    activity_ids = world_history_1200_to_present.map { |unit_template| unit_template[:activities].map { |act| act[:activity_id] } }.flatten
+    unit_activities_include_content_activities?(unit_activities, activity_ids)
+  end
+
+  def unit_activities_include_science_activities?(unit_activities)
```
happythenewsad, line 14: this is equivalent to 

`def unit_activities_include_science_activities? = false`

happythenewsad (changes requested):
I think this is OK for now, except for the one refactor request

happythenewsad (approved):
LGTM!

eadams17 (approved):
LGTM!

services/QuillLMS/app/helpers/content_hubs_helper.rb
```diff
@@ -0,0 +1,295 @@
+# frozen_string_literal: true
+
+module ContentHubsHelper
+
+  def unit_activities_include_content_activities?(unit_activities, content_activity_ids)
+    unit_activities.where(activity_id: content_activity_ids).any?
+  end
+
+  def unit_activities_include_social_studies_activities?(unit_activities)
+    activity_ids = world_history_1200_to_present.map { |unit_template| unit_template[:activities].map { |act| act[:activity_id] } }.flatten
```
anathomical: `world_history_1200_to_present` is not nested, so I don't think that `flatten` does anything here.  Though it doesn't hurt anything, so leaving it might be fine?  I half-suspect that you might have meant to call `compact` instead, though, since that's the one that eliminates `nil` values from the array, and we know there will be some.  Though, again, I don't think that having `nil` values in this array hurts anything.  So maybe nothing needs to be changed here.

services/QuillLMS/app/helpers/content_hubs_helper.rb
```diff
@@ -0,0 +1,295 @@
...
+    unit_activities_include_content_activities?(unit_activities, activity_ids)
+  end
+
+  def unit_activities_include_science_activities?(unit_activities) = false
```
anathomical: I know that this is just a placeholder until we have content, but I sort of love this for some reason anyway.

services/QuillLMS/app/helpers/content_hubs_helper.rb
```diff
@@ -0,0 +1,295 @@
...
+        activities: globalization_activities
+      }
+    ]
+  end
+
+  def global_tapestry_activities
+    [
+      {
+        activity_id: 2750,
+        display_name: 'Wokou Pirates',
```
anathomical: This is a higher level issue, but do we really want to be hard-coding these display names?  This will mean that they don't match the actual activity names, and thus may result in discrepancies in Teacher reporting.  Maybe that's fine, and like I said, it's probably a question beyond the scope of this code, but I notice, for instance, that this activity is actually named "How did the Wokou Pirates Disrupt Sea Routes in East Asia?" in the database.

services/QuillLMS/client/app/bundles/Staff/components/styleGuide/icons.tsx
```diff
@@ -299,6 +304,7 @@ const IconGallery = () => {
     <div id="icons">
       <h2 className="style-guide-h2">Icons</h2>
       <p>All paths follow: <code>{process.env.CDN_URL}/images/icons</code></p>
+      <p>If you need to change the color of an image for use in a design, rather than downloading a new version of the image, please use a <a href="https://codepen.io/sosuke/pen/Pjoqqp" rel="noopener noreferrer" target="_blank">CSS filter generator</a> to apply via CSS and make a mixin for that filter in the variables file if it doesn't already exist. See: `@mixin quill-green-filter`</p>
```
anathomical: Nice!

anathomical (approved):
A few minor items, but I think this could go out as-is.

services/QuillLMS/app/helpers/content_hubs_helper.rb
emilia-friedberg: @anathomical There's a double map, through the unit templates and their nested activities, so we have to call `flatten` to avoid getting an array of arrays of activity ids back here. I'll separate those onto a different lines in an upcoming PR so that's easier to parse. You're right that it does contain some `nil` values, but you're also correct that it doesn't make a difference for what we're doing here.

services/QuillLMS/app/helpers/content_hubs_helper.rb
emilia-friedberg: Good question. I'll double check with Jamie. I think Wokou Pirates is actually just a placeholder here (the description also isn't ready yet for the curriculum team). A lot of the other activities have names that include things like "[Beta - Early Access]", which is part of why I hardcoded, but maybe it would be better to just go ahead and strip that string rather than actually hardcoding a name for each one. I do kind of like that it's easier to see which hash belongs to which activity for the developer, but that doesn't trump the discrepancy possibility you pointed out.

PR #12111, opened 2024-07-22
Concept feedback api

samanthamjohn:
## WHAT
Add a new API to get translations for `concept_feedback`. Also removed now unneeded code for showing hint translations. 

## WHY
We want to pull the translated descriptions to the front-end if the user requests another language. 

## HOW
Add a new route and controller action for translations

### Notion Card Links
[Add concept_feedback API](https://www.notion.so/quill/Add-an-API-endpoint-to-get-all-concept_feedback-for-each-language-b4da66f858844d1f9990f21ba3f64122?pvs=4)

Tested the API route. 

PR Checklist | Your Answer
------------ | -------------

### DIFF
--- a/services/QuillLMS/app/controllers/api/v1/concept_feedback_controller.rb
+++ b/services/QuillLMS/app/controllers/api/v1/concept_feedback_controller.rb
@@ -1,14 +1,20 @@
 # frozen_string_literal: true
 
 class Api::V1::ConceptFeedbackController < Api::ApiController
-  before_action :activity_type, except: [:index]
-  before_action :concept_feedback_by_uid, except: [:index, :create, :update]
+  before_action :activity_type, except: [:index, :translations]
+  before_action :concept_feedback_by_uid, except: [:index, :create, :update, :translations]
 
   CACHE_EXPIRY = 24.hours
 
   def index
-    concept_feedbacks = $redis.get("#{ConceptFeedback::ALL_CONCEPT_FEEDBACKS_KEY}_#{params[:activity_type]}")
+    concept_feedbacks = $redis.get(concepts_cache_key)
     concept_feedbacks ||= fetch_all_concept_feedbacks_and_cache
+    render json: concept_feedbacks
+  end
+
+  def translations
+    concept_feedbacks = $redis.get(translation_cache_key)
+    concept_feedbacks ||= fetch_all_concept_feedback_translations_and_cache
     render json: concept_feedbacks
   end
 
@@ -53,13 +59,26 @@ class Api::V1::ConceptFeedbackController < Api::ApiController
 
   private def fetch_all_concept_feedbacks_and_cache
     concept_feedbacks = ConceptFeedback
-      .includes(:translation_mappings, :english_texts, :translated_texts)
       .where(activity_type: params[:activity_type])
       .all
       .reduce({}) { |agg, q| agg.update({q.uid => q.as_json}) }
       .to_json
 
-    $redis.set("#{ConceptFeedback::ALL_CONCEPT_FEEDBACKS_KEY}_#{params[:activity_type]}", concept_feedbacks)
+    $redis.set(concepts_cache_key, concept_feedbacks)
     concept_feedbacks
   end
+
+  private def fetch_all_concept_feedback_translations_and_cache
+    concept_feedbacks = ConceptFeedback
+      .includes(:translation_mappings, :english_texts, :translated_texts)
+      .where(activity_type: params[:activity_type])
+      .reduce({}) { |acc, cf| acc.merge!(cf.translations_json(locale: params[:locale])) }
+      .to_json
+
+    $redis.set(translation_cache_key, concept_feedbacks)
+    concept_feedbacks
+  end
+
+  private def translation_cache_key = "#{ConceptFeedback::ALL_CONCEPT_FEEDBACKS_KEY}_#{params[:activity_type]}_#{params[:locale]}"
+  private def concepts_cache_key = "#{ConceptFeedback::ALL_CONCEPT_FEEDBACKS_KEY}_#{params[:activity_type]}"
 end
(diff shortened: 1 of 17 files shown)


services/QuillLMS/spec/models/concept_feedback_spec.rb
```diff
@@ -84,4 +74,25 @@
     end
   end
 
+  describe '#translations_json(language:)' do
+    subject { concept_feedback.translations_json(locale: )}
```
anathomical, line 78: Style nitpick: I think you want to move the space inside the final parens to outside the parens and inside the curly brace.  Interesting that we don't have a linting rule for consistent spacing like this.

services/QuillLMS/spec/models/concept_feedback_spec.rb
```diff
@@ -84,4 +74,25 @@
...
+
+    let(:locale) { Translatable::DEFAULT_LOCALE}
+
+    context 'there is a translation for the language' do
+      let(:concept_feedback) { create(:concept_feedback, :with_translated_text)}
```
anathomical, line 83: Again, I'd probably have a space at the end here so that you have a space inside both ends of the curly brace.  Same in the next context.

services/QuillLMS/spec/factories/translation_mappings.rb
```diff
@@ -18,4 +18,8 @@
     english_text { create(:english_text) }
     field_name { 'description' }
   end
+
+  factory :translation_mapping_with_translation, parent: :translation_mapping do
+    english_text { create(:english_text, :with_translated_text)}
```
anathomical, line 23: Again, this doesn't really matter, but you've got a space after the left curly brace, but not before the right curly brace.

services/QuillLMS/spec/controllers/api/v1/concept_feedback_controller_spec.rb
```diff
@@ -27,6 +26,89 @@
...
+      context 'there is no cache set' do
+        before do
+          $redis.del(cache_key)
+        end
+
+        it 'returns all the concept feedback translated fields for that language' do
+          get :translations, params: {
+              locale:,
+              activity_type: concept_feedback1.activity_type
+          }, as: :json
```
anathomical, line 49: Take or leave this, but since this `get` call is repeated in both `it` blocks, I think you might just want to put the `get` call in the `before` block in this context.

services/QuillLMS/spec/controllers/api/v1/concept_feedback_controller_spec.rb
```diff
@@ -27,6 +26,89 @@
...
+          expect($redis.get(cache_key)).to eq(response.body)
+        end
+      end
+
+      context 'the cache is set' do
+        let(:cache_value) { {foo: 'bar'}.to_json }
+
+        it 'returns the cache value' do
+          $redis.set(cache_key, cache_value)
```
anathomical, line 68: This is purely stylistic especially given that there's only the one `it` block here, but it might be worth standardizing whether these `$redis` calls are all going to be in `before` blocks (as they are above and below), or if we want them in the `it` blocks in these cases where there's only the one `it` block.

services/QuillLMS/app/models/concerns/translatable.rb
```diff
@@ -39,7 +31,7 @@ def create_translation_mappings_with_text(translatable_text:, field_name: defaul
   end
 
   def translation(locale: DEFAULT_LOCALE, source_api: OPEN_AI_SOURCE)
-    translations(locale: locale, source_api: source_api)&.first&.translation
+    @translation ||= translations(locale: locale, source_api: source_api)&.first&.translation
```
anathomical, line 34: I really dig this memoization.  Good call.

services/QuillLMS/app/controllers/api/v1/concept_feedback_controller.rb
```diff
@@ -1,15 +1,17 @@
 # frozen_string_literal: true
 
 class Api::V1::ConceptFeedbackController < Api::ApiController
-  before_action :activity_type, except: [:index]
-  before_action :concept_feedback_by_uid, except: [:index, :create, :update]
+  before_action :activity_type, except: [:index, :translations]
+  before_action :concept_feedback_by_uid, except: [:index, :create, :update, :translations]
```
anathomical, line 5: This is not your code, but we have moved to generally preferring `only` over `except` so that you don't have to go in and add your new actions to lists like this if the `before_actions` are irrelevant.  Might be worth cleaning up here, but, again, this isn't your code so feel free to skip it.

anathomical (approved):
A couple of super-minor style notes, but I like this approach in general.  It's exciting to see the translation code expand to new models so cleanly.

services/QuillLMS/spec/models/concept_feedback_spec.rb
samanthamjohn: I will make a rubocop for this and put it in another PR

services/QuillLMS/spec/models/concept_feedback_spec.rb
samanthamjohn: I will make a rubocop for this and put it in another PR

services/QuillLMS/spec/factories/translation_mappings.rb
samanthamjohn: I will make a rubocop for this and put it in another PR

services/QuillLMS/app/controllers/api/v1/concept_feedback_controller.rb
samanthamjohn: I will make another PR for this, as I am a tiny bit nervous about fat-fingering something and screwing us up. Would rather have that isolated.

PR #12112, opened 2024-07-22
Adgr report new diagnostics

anathomical:
## WHAT
- Add new diagnostic IDs to BigQuery materialized views used in reporting Admin Diagnostics
- Tweak the "Diagnostic" selection drop-downs on the "Performance by Skill" and "Performance by Student" tabs so that it (invisibly to the user) populates the drop-downs with the appropriate Diagnostic ID values based on the currently-selected Timeframe filter
## WHY
- We want the new diagnostics that Curriculum is launching to be included in the data being reported on
- We want the complexity of the diagnostic switchover to be effectively invisible to admins, and for everything to work seamlessly.
## HOW
- Add the new diagnostic IDs to the root admin diagnostic query
- Update the underlying materialized view definition so that BigQuery includes the new diagnostics in its data rollups
- Update the logic in the code that sets Diagnostic drop-down selections to change the option values based on the the selected timeframe

### Screenshots
Screenshot 2024-07-22 at 3 34 25 PM

### Notion Card Links
https://www.notion.so/quill/New-Diagnostic-Engineering-Tasks-ef849d3f473a4a11a675081a89556a75?pvs=4#18c3f8030c5e4269bc80b49090bb2af6

BACKEND
- Used a "STAGING" version of the materialized view to test functionality without impacting production (note that this code is not visible in "Files changed" because once functionality was confirmed, those changes were moved into the production materialized view definition)
- - Created a copy of the `pre_post_diagnostic_skill_group_performance_view.sql`
- - Modified it with the new logic
- - Updated queries to point at the STAGING materialized view in the branch
- - Tested all functionality in staging
FRONTEND
- Select "This school year" as the timeframe
- Select Skills tab.  Confirm that the `diagnostic_id` requested for the report was was one of the new IDs
- Select a different diagnostic from the drop-down, confirm that the `diagnostic_id` requested for the report is one of the new ones
- Repeat for Student tab.
- Select "Last school year" as the timeframe and repeat the earlier steps, but confirm that the `diagnostic_id` requested for reports are old ones
- On the Skill tab, with "This school year" selected in filters, select a non-default Diagnostic from the drop-down (Intermediate Pre), confirm that a new ID is submitted for reporting
- - Change the filter to select "Last school year"
- - Confirm that the selected Diagnostic does not change
- - Confirm that the updated report requested by the timeframe filter looks for the old ID rather than the new one
- Repeat steps for the Student tab

PR Checklist | Your Answer
------------ | -------------
Have you added and/or updated tests? | Minor tweak to keep existing tests passing, but no new tests


services/QuillLMS/client/app/bundles/PremiumHub/shared.tsx
```diff
@@ -57,6 +71,12 @@ export const restrictedPage = (
 
 export const RESTRICTED_TEXT = "Sorry, only verified admins with a subscription to School or District Premium can access this feature."
 
+export function getDiagnosticTypeDropdownOptionsByTimeframe(timeframe) {
```
dandrabik: Take or leave. You are passing `timeframe` as a required var, so you could call this `getDiagnosticTypeDropdownOptions` which will always read as `getDiagnosticTypeDropdownOptions(selectedTimeframe)` so the `ByTimeframe` is somewhat redundant. If you think it's clearer to leave it to distinguish from other methods, it's fine to leave as is.

services/QuillLMS/client/app/bundles/PremiumHub/shared.tsx
```diff
@@ -22,9 +22,11 @@ export const STUDENT = 'student'
 export const FETCH_ACTION = 'fetch'
 export const RESET_ACTION = 'reset'
 
+export const OLD_DIAGNOSTIC_TIMEFRAME = 'last-school-year'
```
emilia-friedberg, line 25: Is the idea that we'll just remove this next year?

services/QuillLMS/db/big_query/views/pre_post_diagnostic_skill_group_performance_view.sql
```diff
@@ -56,22 +56,31 @@ SELECT
...
             concept_results.question_number,
-            (
-              COUNT(diagnostic_question_optimal_concepts.id) > 0 AND
-              LOGICAL_AND(CASE WHEN diagnostic_question_optimal_concepts.id IS NOT NULL THEN concept_results.correct END)
-            ) AS correct
+            /*
+              Old diagnostics use a more complicated way of scoring questions than new ones
+              Some old style diagnostics have multiple concepts that they test for, and thus need to account for cases where there are multiple ConceptResults for a single answer, but only some of them count toward whether or not the question as a whole counts as "correct" or not
+              New style diagnostics expect to always have only one ConceptResult for each student response, so we're effectively expecting to be doing LOGICAL_AND on a set of one item for each of them
+            */
```
emilia-friedberg, line 63: Thank you for the comment here, and it's nice that we'll no longer have to worry about this.

dandrabik (approved):
:+1: LGTM

emilia-friedberg (approved):
This looks good to me, but I'm curious about the fact that some of this code is necessarily temporary (as I think we'll have to remove the switching based on timeframe once all timeframes use the new diagnostics). Also, I just want to flag that once this is live and the demo account data has been updated, we'll also need to rerun the admin demo generation script, as the data currently in that account is for the old diagnostics.

services/QuillLMS/client/app/bundles/PremiumHub/shared.tsx
anathomical: This is a good call out.  My understanding is that the intention is to add a new timeframe every Summer to extend an additional year.  So, for instance, there will be something like "Two years ago" added next Summer.  This value will need to constantly be updated to match whatever that oldest timeframe is.  Which means that there may be a cleaner way of doing this than using a constant in this way.  I'm open to suggestions on that subject.

services/QuillLMS/client/app/bundles/PremiumHub/containers/diagnosticGrowthReports/skillSection.tsx
```diff
@@ -122,7 +122,18 @@ export const SkillSection = ({
   React.useEffect(() => {
     if (!passedData && groupByValue && diagnosticTypeValue) {
       // this is for testing purposes; this value will always be null in a non-testing environment
-      getData()
+
+      // If the timeframe has changed, we may be re-populating the selectedDiagnosticType drop-downs
+      // In these cases, we need to re-select the appropriate value from the drop-down for the new timeframe
+      // The re-selection will re-trigger this effect by changing the value of diagnosticTypeValue
```
eadams17, line 128: appreciate the comment here!

eadams17 (approved):
LGTM! Thanks for putting this together

services/QuillLMS/client/app/bundles/PremiumHub/shared.tsx
happythenewsad: We could punt on this til next Summer and see what the usage patterns are.

happythenewsad (approved):

PR #12118, opened 2024-07-23
Questions api

samanthamjohn:
## WHAT
Added a new API to get questions for a single activity--either in english, or translated
## WHY
So that we can display the question and feedback in a new language
## HOW
added a new controller and action to get all questions. 

### Notion Card Links
[Add API endpoint for questions translation](https://www.notion.so/quill/Add-an-API-endpoint-to-get-translated-question-feedback-for-an-activity-c82111c419d649c1b2efac810e132265?pvs=4)

Tested the API URL on local to see if it works for translated and untranslated questions. 
 
PR Checklist | Your Answer
------------ | -------------

### DIFF
--- a/services/QuillLMS/app/controllers/api/v1/activity_questions_controller.rb
+++ b/services/QuillLMS/app/controllers/api/v1/activity_questions_controller.rb
@@ -2,7 +2,7 @@
 
 class Api::V1::ActivityQuestionsController < ApplicationController
   def index
-    activity = Activity.find_by(uid: params[:activity_id])
+    activity = Activity.find_by(uid: permitted_params[:activity_id])
 
     if activity.nil?
       render json: { error: 'Activity not found' }, status: :not_found
@@ -15,12 +15,16 @@ class Api::V1::ActivityQuestionsController < ApplicationController
   private def cached_questions(activity)
     Rails.cache.fetch(cache_key(activity), expires_in: 1.hour) do
       activity.questions.each_with_object({}) do |question, hash|
-        hash[question.uid] = question.as_json(locale: params[:locale])
+        hash[question.uid] = question.as_json(locale: permitted_params[:locale])
       end
     end
   end
 
   private def cache_key(activity)
-    "activity_questions/#{activity.uid}/#{params[:locale]}"
+    "activity_questions/#{activity.uid}/#{permitted_params[:locale]}"
+  end
+
+  private def permitted_params
+    params.permit(:activity_id, :locale)
   end
 end
--- a/services/QuillLMS/app/controllers/api/v1/activity_questions_controller.rb
+++ b/services/QuillLMS/app/controllers/api/v1/activity_questions_controller.rb
@@ -1,0 +1,26 @@
+# frozen_string_literal: true
+
+class Api::V1::ActivityQuestionsController < ApplicationController
+  def index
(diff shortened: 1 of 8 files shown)


samanthamjohn:
Note: There might be a little bit of funky white space in some of the tests, but I just added the [rubocops to enforce whitespace rules here](https://github.com/empirical-org/Empirical-Core/pull/12123), so I think we don't need to worry about it.

eadams17 (approved):
LGTM!

services/QuillLMS/app/controllers/api/v1/activity_questions_controller.rb
```diff
@@ -0,0 +1,26 @@
+# frozen_string_literal: true
+
+class Api::V1::ActivityQuestionsController < ApplicationController
+  def index
+    activity = Activity.find_by(uid: params[:activity_id])
```
anathomical: This probably doesn't matter, but I generally like the pattern of running `params` through a `permit` call.  So something like
```ruby
private def permitted_params
  params.permit(:activity_id, :locale)
end```
And then call `permitted_params` instead of `params`.  It provides a good framework for expansion and potential validation.

File: services/QuillLMS/app/models/question.rb [hunk 2]
```diff
@@ -79,7 +79,8 @@ class Question < ApplicationRecord
   scope :production, -> { where("data->>'flag' = ?", FLAG_PRODUCTION) }
 
   def as_json(options=nil)
-    data
+    locale = options&.[](:locale)
```
anathomical: This is purely stylistic, but I find `.[]` to be weird enough that it takes me a second to parse.  I prefer `.fetch(key, nil)`.  So `locale = options&.fetch(:locale, nil)`

services/QuillLMS/spec/controllers/api/v1/activity_questions_controller_spec.rb
```diff
@@ -0,0 +1,92 @@
...
+    before do
+      allow(Activity).to receive(:find_by).with(uid: activity.uid).and_return(activity)
+      allow(activity).to receive(:questions).and_return(questions)
+      questions.each do |question|
+        allow(question).to receive(:as_json).with(locale: locale).and_return({ "id" => question.id, "content" => "Question content" })
+      end
+    end
+
+    it "returns http success" do
+      get :index, params: { activity_id: activity.uid, locale: locale }
```
anathomical: Purely stylistic, but you can infer the value part of `locale` here: `get :index, params: { activity_id: activity.uid, locale: }`

anathomical (approved):
Some minor style stuff I noted, but this looks good to me.

services/QuillLMS/app/models/question.rb
samanthamjohn: this is a way nicer syntax! I didn't know it.

services/QuillLMS/spec/controllers/api/v1/activity_questions_controller_spec.rb
samanthamjohn: whoops! I think that rubocop I made is not working right...

services/QuillLMS/app/controllers/api/v1/activity_questions_controller.rb
samanthamjohn: good call

PR #12127, opened 2024-07-24
Scope questions for grammar

samanthamjohn:
## WHAT
Don't load *all* the grammar questions every time somebody does a grammar activity. 
## WHY
The payload is very large leading to slow response times. It's one of our top most time-consuming requests. 

## HOW
In the grammar module switch to use the new activity-scoped question api. 

### Notion Card Links
[Scope Question to a single activity](https://www.notion.so/quill/Scope-question-json-API-to-the-current-activity-rather-than-loading-all-questions-for-grammar-50d2bbb6ae46491faf88551d8df993b6?pvs=4)

Tested that grammar activities still work properly in admin and for students.

PR Checklist | Your Answer
------------ | -------------

### DIFF
--- a/services/QuillLMS/client/app/bundles/Grammar/actions/session.ts
+++ b/services/QuillLMS/client/app/bundles/Grammar/actions/session.ts
@@ -6,6 +6,7 @@ import * as responseActions from './responses';
 
 import { hashToCollection } from '../../Shared/index';
 import { permittedFlag } from '../helpers/flagArray';
+import getParameterByName from '../helpers/getParameterByName';
 import { shuffle } from '../helpers/shuffle';
 import { Question } from '../interfaces/questions';
 import { GRAMMAR_QUESTION_TYPE, QuestionApi } from '../libs/questions_api';
@@ -15,7 +16,7 @@ import { SessionState } from '../reducers/sessionReducer';
 export const allQuestions = {};
 let questionsInitialized = false;
 
-export const populateQuestions = (questions: object, forceRefresh: boolean) => {
+export const populateQuestions = (questions: object, forceRefresh?: boolean) => {
   if (questionsInitialized && !forceRefresh) return;
   Object.keys(questions).forEach((uid) => {
     questions[uid].uid = uid
@@ -126,11 +127,12 @@ const normalizeQuestionWithAttempts = (question: object): object => {
     attempts: question.attempts,
   }
 }
+const activityUID = getParameterByName('uid', window.location.href)
 
 const handleGrammarSession = (session) => {
   return dispatch => {
     if (session && Object.keys(session).length > 1 && !session.error) {
-      QuestionApi.getAll(GRAMMAR_QUESTION_TYPE).then((allQuestions) => {
+      QuestionApi.getAll(activityUID).then((allQuestions) => {
         if (session.currentQuestion) {
           if (!session.currentQuestion.prompt || !session.currentQuestion.answers) {
             const currentQuestion = allQuestions[session.currentQuestion.uid]
@@ -209,7 +211,7 @@ export const getQuestionsForConcepts = (concepts: any, flag: string) => {
   return (dispatch, getState) => {
     dispatch(setSessionPending(true))
     const conceptUIDs = Object.keys(concepts)
-    QuestionApi.getAll(GRAMMAR_QUESTION_TYPE).then((questions) => {
+    QuestionApi.getAllForType(GRAMMAR_QUESTION_TYPE).then((questions) => {
       const questionsForConcepts = {}
       Object.keys(questions).map(q => {
         if (conceptUIDs.includes(questions[q].concept_uid) && questions[q].prompt && questions[q].answers && permittedFlag(flag, questions[q].flag)) {
@@ -250,7 +252,7 @@ export const getQuestionsForConcepts = (concepts: any, flag: string) => {
 export const getQuestions = (questions: any, flag: string) => {
   return dispatch => {
     dispatch(setSessionPending(true))
-    QuestionApi.getAll(GRAMMAR_QUESTION_TYPE).then((allQuestions) => {
+    QuestionApi.getAll(activityUID).then((allQuestions) => {
       const arrayOfQuestions = questions.map(q => {
         const question = allQuestions[q.key]
         question.uid = q.key
--- a/services/QuillLMS/client/app/bundles/Grammar/actions/session.ts
+++ b/services/QuillLMS/client/app/bundles/Grammar/actions/session.ts
@@ -127,7 +127,7 @@ const normalizeQuestionWithAttempts = (question: object): object => {
     attempts: question.attempts,
   }
 }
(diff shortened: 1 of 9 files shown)


samanthamjohn:
**Notes**
- we should probably merge the [questions API PR](https://github.com/empirical-org/Empirical-Core/pull/12118) first. This one is set to merge into that so you can see the difference. 
- I wasn't sure how to get the `activity_uid` in react, so I did a little bit of hacky code with the URLs. A better method would be welcome!

services/QuillLMS/client/app/bundles/Grammar/actions/session.ts
```diff
@@ -127,10 +127,16 @@ const normalizeQuestionWithAttempts = (question: object): object => {
   }
 }
 
+const getUID = () => {
+  const url = window.location.href;
+  const uidMatch = url.match(/[?&]uid=([^&]+)/);
+  return uidMatch ? uidMatch[1] : null;
+}
+
```
eadams17, line 131: There's a `getParameterByName` function you can use from the helpers (`client/app/bundles/Grammar/helpers/getParameterByName.ts`)

eadams17 (approved):
There's a shared function you can use for fetching the UID, but otherwise looks good to me!

services/QuillLMS/client/app/bundles/Grammar/actions/session.ts
emilia-friedberg: Eric is right -- example of how this being used here: https://github.com/empirical-org/Empirical-Core/blob/bfb249482a998f2e29a51c6343f76c9b0745f159/services/QuillLMS/client/app/bundles/Grammar/components/grammarActivities/container.tsx#L90.

services/QuillLMS/client/app/bundles/Grammar/actions/session.ts
```diff
@@ -209,7 +215,7 @@ export const getQuestionsForConcepts = (concepts: any, flag: string) => {
   return (dispatch, getState) => {
     dispatch(setSessionPending(true))
     const conceptUIDs = Object.keys(concepts)
-    QuestionApi.getAll(GRAMMAR_QUESTION_TYPE).then((questions) => {
+    QuestionApi.getAll(getUID()).then((questions) => {
```
emilia-friedberg: This function is used when the session is either a) a follow-up to a Proofreader activity, or b) when the activity itself has assigned concepts rather than assigned questions ([example activity here](https://www.quill.org/grammar/#/admin/lessons/eUTXuBfu8EpXOq2pSfBpFg/)). In both instances, the questions themselves aren't associated with the activity directly, but instead there are set concepts where questions that are associated with that concept are pulled at random (which is what's happening below, in this method). So I don't think it'll work to pull questions by activity here, because you'll get nothing. We probably need to either move that logic to the backend or leave it alone for now.

services/QuillLMS/client/app/bundles/Grammar/components/admin/admin.tsx
```diff
@@ -39,7 +39,12 @@ class AdminContainer extends React.Component<AdminContainerProps> {
       }
     }
     );
-    this.props.dispatch(questionsActions.startListeningToQuestions());
+    function getUIDFromURL(url: string): string {
+      const parts = url.split('/');
+      return parts[parts.length - 2]
+    }
+    const activityUID = getUIDFromURL(this.props.location.pathname)
+    this.props.dispatch(questionsActions.startListeningToQuestions(activityUID));
```
emilia-friedberg: Does this component normally have an activity uid parameter? It looks to me like this is the top-level parent component for the whole admin interface, so I wonder if we want to keep this questions call generic so we have them all in the reducer for the `/grammar/#/admin/questions` page and other places we might actually want all of the questions.

emilia-friedberg (changes requested):
Overall I think this will be a big improvement and I'm excited to see it go out, but I agree with Eric's suggestion about using the existing `getParameterByName` method for the activity uid here. In addition, I have some concerns about places we may actually want or need to use the existing `getAll` functionality -- we should make sure we do a thorough QA once those concerns have been addressed.

services/QuillLMS/client/app/bundles/Grammar/actions/session.ts
samanthamjohn: good catch!

services/QuillLMS/client/app/bundles/Grammar/components/admin/admin.tsx
samanthamjohn: you're right, it seems the grammar admin interface is set up so that it expects to have the list of *all* the grammar questions. That's probably why the student-facing one does the same.

services/QuillLMS/client/app/bundles/Grammar/actions/session.ts
```diff
@@ -126,11 +127,12 @@ const normalizeQuestionWithAttempts = (question: object): object => {
     attempts: question.attempts,
   }
 }
+const activityUID = getParameterByName('uid', window.location.href)
```
emilia-friedberg: I'm not totally sure when this gets instantiated, or if it'll change when the URL changes. It might be safest to make it into a function (ex `const activityUID = () => getParameterByName...`) so you can be confident it's getting defined when it's called and not holding onto something stale.

services/QuillLMS/client/app/bundles/Grammar/libs/questions_api.ts
```diff
@@ -4,12 +4,17 @@ import { FocusPoint, IncorrectSequence, Question } from '../interfaces/questions
...
 const questionApiBaseUrl = `${process.env.DEFAULT_URL}/api/v1/questions`;
+const activityApiBaseUrl = `${process.env.DEFAULT_URL}/api/v1/activities`;
 
 class QuestionApi {
-  static getAll(questionType: string): Promise<Array<Question>> {
+  static getAllForType(questionType: string): Promise<Array<Question>> {
     return requestGet(`${questionApiBaseUrl}.json?question_type=${questionType}`, null, (error) => {throw(error)});
   }
 
+  static getAll(activity_uid: string): Promise<Array<Question>> {
```
emilia-friedberg: I think it would be clearer to name this `getAllQuestionsForActivity`.

emilia-friedberg (approved):
Two final comments, but nothing I need to look at again. Thank you for making these changes!

services/QuillLMS/client/app/bundles/Grammar/libs/questions_api.ts
samanthamjohn: switching it to `getAllForActivity` since it's already in `QuestionsApi`

services/QuillLMS/client/app/bundles/Grammar/actions/session.ts
samanthamjohn: yep, it causes problems in admin if you don't make it a function, good catch.

PR #12134, opened 2024-07-24
Add a filter for languages on the admin translations page.

samanthamjohn:
## WHAT
Add a list of available locales to the admin translations page 

## WHY
So that someone checking over translations doesn't get bogged down by all the languages once we've added them

## HOW
Added a section to the top of the admin page with a list of locales. 

### Screenshots
2024-07-24 at 1 32 PM

### Notion Card Links
[Admins can filter by language](https://www.notion.so/quill/Admins-can-filter-by-language-on-the-translations-page-05c823eb9952420fb281855442490aac?pvs=4
)
Tested the filter on development 

PR Checklist | Your Answer
------------ | -------------
Have you deployed to Staging? | NO, small change (and I am waiting on another PR that has a staging deploy) 

### DIFF
--- a/services/QuillLMS/app/controllers/translated_texts_controller.rb
+++ b/services/QuillLMS/app/controllers/translated_texts_controller.rb
@@ -5,12 +5,14 @@ class TranslatedTextsController < ApplicationController
 
   def index
     @translated_texts = TranslatedText
-      .all
       .includes(:english_text, :translation_mappings)
+      .order(created_at: :asc)
       .page(params[:page])
       .per(200)
 
+    @translated_texts = @translated_texts.where(locale: params[:locale]) if params[:locale].present?
     @js_file = "entrypoints/snippets/translated_text.ts"
+    @locales = TranslatedText.select(:locale).distinct.pluck(:locale)
   end
 
   def update
--- a/services/QuillLMS/spec/controllers/translated_texts_controller_spec.rb
+++ b/services/QuillLMS/spec/controllers/translated_texts_controller_spec.rb
@@ -39,7 +39,7 @@ RSpec.describe TranslatedTextsController, type: :controller do
         locales = ["es-la", "zh-cn"]
         locales.each { |locale| create(:translated_text, locale:) }
         get :index
-        expect(assigns(:locales).to match_array(locales)
+        expect(assigns(:locales)).to match_array(locales)
       end
 
       context 'a locale parameter is passed in' do
--- a/services/QuillLMS/spec/controllers/translated_texts_controller_spec.rb
+++ b/services/QuillLMS/spec/controllers/translated_texts_controller_spec.rb
@@ -36,10 +36,10 @@ RSpec.describe TranslatedTextsController, type: :controller do
       end
 
       it 'assigns @locales' do
-        create(:translated_text, locale: "es-la")
-        create(:translated_text, locale: "zh-cn")
+        locales = ["es-la", "zh-cn"]
+        locales.each { |locale| create(:translated_text, locale:) }
         get :index
-        expect(assigns(:locales)).to match_array(["es-la", "zh-cn"])
+        expect(assigns(:locales).to match_array(locales)
       end
 
       context 'a locale parameter is passed in' do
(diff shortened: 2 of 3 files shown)


services/QuillLMS/app/controllers/translated_texts_controller.rb
```diff
@@ -5,12 +5,14 @@ class TranslatedTextsController < ApplicationController
 
   def index
     @translated_texts = TranslatedText
-      .all
       .includes(:english_text, :translation_mappings)
+      .order(created_at: :asc)
       .page(params[:page])
       .per(200)
 
+    @translated_texts = @translated_texts.where(locale: params[:locale]) if params[:locale].present?
```
happythenewsad, line 13: This method is calling to the DB twice, when it only needs one call. One way to fix this, while staying in the AREL syntax, would be:

```ruby
    where_clause = params[:locale].present? ? ['local IN ?', params[:locale]] : ['1 = ?', '1']

    @translated_texts = TranslatedText
      .includes(:english_text, :translation_mappings)
      .where(*where_clause)
      .order(created_at: :asc)
      .page(params[:page])
      .per(200)
```

services/QuillLMS/app/controllers/translated_texts_controller.rb
```diff
@@ -5,12 +5,14 @@ class TranslatedTextsController < ApplicationController
...
     @js_file = "entrypoints/snippets/translated_text.ts"
+    @locales = TranslatedText.select(:locale).distinct.pluck(:locale)
```
happythenewsad, line 15: Optional: depending on the anticipated traffic on this page, consider caching this value in redis or declaring it inline statically, as it should eventually converge on a stable array value

happythenewsad (changes requested):

services/QuillLMS/app/controllers/translated_texts_controller.rb
samanthamjohn: Rails will lazy load this, so it only does one query, even though it looks like two!

services/QuillLMS/app/controllers/translated_texts_controller.rb
samanthamjohn: I think the traffic will be quite low since it's only a staff page, but you're right that it could get a little slow as we add more languages. Maybe cross that bridge if folks complain?

samanthamjohn: I don't think we'll get beyond a couple hundred thousand records in `translated_text` and I think Postgres has this stuff pretty optimized so I doubt it'll be an issue.

services/QuillLMS/app/controllers/translated_texts_controller.rb
happythenewsad: My bad; my rails console inspect was causing the non-lazy load; doing this fixed the issue:

`console$ <query> ; nil`

happythenewsad (approved):

services/QuillLMS/spec/controllers/translated_texts_controller_spec.rb
```diff
@@ -34,6 +34,31 @@
       it 'assigns @js_file' do
         expect(assigns(:js_file)).to eq("entrypoints/snippets/translated_text.ts")
       end
+
+      it 'assigns @locales' do
+        create(:translated_text, locale: "es-la")
+        create(:translated_text, locale: "zh-cn")
+        get :index
+        expect(assigns(:locales)).to match_array(["es-la", "zh-cn"])
```
anathomical: Feel free to take or leave this, but I'm always a bit concerned about cases where a string is typed twice.  In this case, it's fine, but I could imagine a typo in a spec update breaking this at some point.  I think I'd prefer something like:
```
it 'assigns @locales' do
  locales = ["es-la", "zh-cn"]
  locales.each { |locale| create(:translated_text, locale:) }
  get :index
  expect(assigns(:locales).to match_array(locales)
end
```
Or some other way of making sure the strings only get typed one time.

anathomical (approved):
A small take-or-leave style comment, but I like the approach here.  Agree that we can worry about solving any locale caching at some future date if it ends up being a problem.  (Even if the dataset gets large, this page shouldn't see a ton of traffic, so it may be okay for a very long time.)

PR #12144, opened 2024-07-25
Use the language options from the API to populate the language picker

samanthamjohn:
## WHAT
Make the language header populate based on the available languages available in the activity. If there is only english available, don't show it. 

## WHY
So we don't tempt students with languages we don't actually support 
## HOW
Use the `translations` key in the `activity` payload to determine which languages to show. 

### Screenshots
2024-07-25 at 3 19 PM

### Notion Card Links
https://www.notion.so/quill/Add-language-picker-to-the-top-bar-on-activities-where-there-are-translations-for-Connect-and-Gramma-0c449f70511b4ae599beb67294cce102?pvs=4

Tested it on development
PR Checklist | Your Answer
------------ | -------------
Have you added and/or updated tests? |  NO-frontend change

### DIFF
--- a/services/QuillLMS/client/app/bundles/Connect/components/home.tsx
+++ b/services/QuillLMS/client/app/bundles/Connect/components/home.tsx
@@ -29,17 +29,16 @@ export const Home = ({playLesson, lessons, dispatch}) => {
   const [skippedToQuestionFromIntro, setSkippedToQuestionFromIntro] = React.useState<boolean>(false);
   const [languageOptions, setLanguageOptions] = React.useState<any>(null);
   React.useEffect(() => {
-    if (lessons?.hasreceiveddata) {
-      const translations = lessons.data?.[lessonUid]?.translations ?? {};
-      const languageOptions = [
-        { value: ENGLISH, label: ENGLISH },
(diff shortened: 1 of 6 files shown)


services/QuillLMS/client/app/bundles/Connect/components/home.tsx
```diff
@@ -25,6 +27,20 @@ export const Home = ({playLesson, dispatch}) => {
   const [questionToPreview, setQuestionToPreview] = React.useState<any>(null);
   const [switchedBackToPreview, setSwitchedBackToPreview] = React.useState<boolean>(false);
   const [skippedToQuestionFromIntro, setSkippedToQuestionFromIntro] = React.useState<boolean>(false);
+  const [languageOptions, setLanguageOptions] = React.useState<any>(null);
+  React.useEffect(() => {
+    if (lessons?.hasreceiveddata) {
```
emilia-friedberg: Very minor but I would do an early return here, so `if (!lessons?.hasreceiveddata) { return }` and then the rest of the effect.

emilia-friedberg (approved):
LGTM!

PR #12147, opened 2024-07-25
Add array indentation rule to rubocop  and remove an exception that wasn't doing anything. 

samanthamjohn:
## WHAT
Make arrays always indent based on the opening parenthesis. https://rubydoc.info/gems/rubocop/RuboCop/Cop/Layout/FirstArrayElementIndentation
## WHY
To make our code style more consistent so we don't have to think about it. 
## HOW
Remove exceptions from `rubocop_todo.yml`


happythenewsad (commented):
Let's discuss in the next engineering meeting whether or not we want this rule. Without additional formatting, it's leading to cases like this, which look worse than pre-rule:
Screenshot 2024-07-26 at 9 39 05 AM

samanthamjohn:
yeah, I saw a few of these. I think it's a good thing to discuss at engineering meeting! The nice thing about having the rule is that it sort of forces you to format your lines to look good *with* the rule, and that line was probably way too long to start with. 

If you see something like that you could fix it up like this: 
2024-07-26 at 9 47 AM
Or like this: 
2024-07-26 at 9 50 AM

anathomical (approved):

samanthamjohn:
I think it looks pretty good now

happythenewsad (approved):
🚢

PR #12149, opened 2024-07-26
Activity landing page translation

samanthamjohn:
## WHAT
Translate the activity landing page if you select a language from the dropdown. 
## WHY
So that people for whom english is a second language can see instructions in their own language. 
## HOW
Pull the translation from the activity API call and conditionally add it to the page
### Screenshots
2024-07-26 at 9 36 AM

### Notion Card Links
https://www.notion.so/quill/If-you-pick-a-language-show-connect-activity-landing-page-in-english-your-second-language-6777392ce1464a139fec5048104613e9?pvs=4

Tested it on local and staging

PR Checklist | Your Answer
------------ | -------------


services/QuillLMS/client/app/bundles/Connect/styles/components/_student-container.scss
```diff
@@ -26,3 +26,7 @@
   width: 830px;
   box-shadow: 0 14px 28px rgba(0,0,0,0.25), 0 10px 10px rgba(0,0,0,0.22);
 }
+
+.student-container hr {
+  background-color: $hr-gray;
```
emilia-friedberg: This is actually defined in our outermost variables.scss file as `$quill-grey-10`, so you can just use that instead of having to define it independently in the Connect variable scss file.

services/QuillLMS/client/app/bundles/Shared/components/studentLessons/register.tsx
```diff
@@ -75,13 +76,43 @@ class Register extends React.Component<any, any> {
     );
   }
 
+  translatedText = (): string => {
+    function getTranslation( translations, targetLanguage): string | undefined {
```
emilia-friedberg: I would move this inner function outside of `translatedText` (and maybe even outside of the component?) for readability and theoretical performance concerns about re-defining `getTranslation` every time `translatedText` gets called.

services/QuillLMS/client/app/bundles/Shared/components/studentLessons/register.tsx
```diff
@@ -75,13 +76,43 @@ class Register extends React.Component<any, any> {
...
+      const targetLocale = Object.entries(localeToLanguageMap).find(
+        ([_, language]) => language === targetLanguage
+      )?.[0];
+
+      if (targetLocale && targetLocale in translations) {
+        return translations[targetLocale];
+      }
+
+      // If the exact locale isn't found, try to find a matching language
```
emilia-friedberg: I'm confused about why we need both of these iterations. It seems like we could just cut the first one and always rely on the second? I think I'm missing how the first one offers greater specificity since we're always just grabbing the first item returned, unless it's the ordering of the `localeToLanguageMap` that matters?

emilia-friedberg (approved):
Looks good overall, just have one small suggestion for keeping variables consistent in CSS and a question about some logic.

services/QuillLMS/client/app/bundles/Shared/components/studentLessons/register.tsx
samanthamjohn: looking at this again, I think it's better if the ruby API brings back the languages by name and the JS doesn't have to worry about all these mappings back and forth

PR #12158, opened 2024-07-29
fix tooltip clickability

emilia-friedberg:
## WHAT
Fix the tooltip so that it's possible to click links within the tooltip text and it's more accessible all around, and also add back some css in Evidence so the modal works.

## WHY
Jamie noticed that this wasn't working on the new course page, so I broke it out into a separate card because the file diff was so big due to all the tests.

## HOW
Update the tooltip itself so that all the `onX` handlers apply to the tooltip container, not just the tooltip trigger, so that we don't always blur when the user tries to move the mouse over the tooltip itself. Then update a big suite of tests and also the Evidence modal css, because once this was fixed it became apparent that that was also broken.

Screenshot 2024-07-29 at 2 10 07 PM
Screenshot 2024-07-29 at 2 10 17 PM

### Screenshots
Screenshot 2024-07-29 at 2 10 07 PM
Screenshot 2024-07-29 at 2 10 17 PM

### Notion Card Links
https://www.notion.so/quill/Fix-tooltip-links-9f69025c613a4f96ad2af1d49a227190?pvs=4

Tested tooltips with links in Evidence and on a merge branch for the new content hub page, plus spot-checked various other places around the website to make sure tooltips in general haven't changed behavior.

PR Checklist | Your Answer
------------ | -------------

### DIFF
--- a/services/QuillLMS/client/app/bundles/Evidence/styles/style.scss
+++ b/services/QuillLMS/client/app/bundles/Evidence/styles/style.scss
@@ -110,6 +110,33 @@ button {
   }
 }
 
+// TODO: move modal scss from here https://github.com/empirical-org/Empirical-Core/blob/develop/services/QuillLMS/app/assets/stylesheets/shared/modals.scss#L415-L453 into Shared bundle and use that instead 
+.modal, .modal-background {
+  width: 100%;
+  height: 100%;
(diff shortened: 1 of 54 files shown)


eadams17 (approved):
thanks for fixing this!

PR #12160, opened 2024-07-29
Update connect to pull all questions for the activity in one request 

samanthamjohn:
## WHAT
Make only one API call to get all the questions in one request

## WHY
This sets us up to use the same API to pull the translations. It is also more efficient than pulling each question individually--both in terms of server load (the `question.get` call is one of our most time-consuming transactions) and in terms of code complexity.  

## HOW
Use the new `QuestionAPI.getAllForActivity` call to pull all the question data. 

### Notion Card Links
https://www.notion.so/quill/Load-all-questions-for-a-connect-activity-together-rather-than-one-at-a-time-ec0f8306767d4963a0747f509c7108e4?pvs=4

Tested it on local and student questions still work fine

PR Checklist | Your Answer
------------ | -------------

### DIFF
--- a/services/QuillLMS/app/models/question.rb
+++ b/services/QuillLMS/app/models/question.rb
@@ -79,7 +79,8 @@ class Question < ApplicationRecord
 
   def as_json(options=nil)
     locale = options&.fetch(:locale, nil)
-    locale.present? ? translated_data(locale:) : data
+    json = locale.present? ? translated_data(locale:) : data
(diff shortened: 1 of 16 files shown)


eadams17 (approved):
I think this seems reasonable to me. I would just make sure that you've played through several different activities to make sure this is thoroughly tested (on staging)

emilia-friedberg (approved):
LGTM!

PR #12170, opened 2024-07-30
Evidence GenAI - Add Experiments for secondary feedback

dandrabik:
## WHAT
Add a system prompt for secondary feedback and a script with accompanying data to run experiments on it.
## WHY
We want to test converting a piece of feedback generated by GenAI to "secondary" feedback which is stronger and points to a highlighted portion of the text. The secondary feedback is supposed to be more helpful to the study who may be stuck 
## HOW
1. Make `PromptBuilder` a base class since there will be several. Build off of that
2. Create all, train, test datasets from a Metabase Query. Make a service to parse the csv
3. Create a thor task to run a prompt against the test set. Save results in a csv.
4. Create a thor task to run an individual entry to see the prompt.

### Notion Card Links
https://www.notion.so/quill/GenAI-Highlight-and-Secondary-feedback-Experiments-fb38b9399af14269b371fb0cf0ee09ed

I've run the task to build a report on the performance several times (and used that to iterate on the prompt). I ran the script to view the secondary prompt output and used that to iterate as well.

PR Checklist | Your Answer
------------ | -------------
Have you deployed to Staging? |  NO - non-live change.


brendanshean (approved):
This is great!  Nice work on building out this secondary feedback testing.

services/QuillLMS/engines/evidence/app/services/evidence/gen_ai/secondary_feedback_prompt_builder.rb
```diff
@@ -0,0 +1,37 @@
...
+
+      private def feedback_data_tuples
+        feedback_data
+          .select {|fd| fd.conjunction == prompt.conjunction }
+          .first(EXAMPLE_LIMIT)
+          .map {|f| [f.primary, f.secondary]}
+      end
+
+      private def feedback_data
+        @feedback_data ||= Evidence::GenAI::SecondaryFeedbackDataFetcher.run
```
brendanshean: This seems odd not having any arguments.  I see a default FILE_TRAIN in the file, but what about other args?

services/QuillLMS/engines/evidence/app/services/evidence/gen_ai/prompt_builder.rb
```diff
@@ -0,0 +1,57 @@
...
+
+      private def template_folder
+        raise NotImplementedError, 'Must implement template_folder'
+      end
+
+      private def default_template
+        raise NotImplementedError, 'Must implement default_template'
+      end
+
+      # Template priority order
```
brendanshean: These comments feel unnecessary given the line `@template_file = template_file || default_template`

dandrabik: Good point, this had a more complicated hierarchy before, but I simplified it. This comment isn't necessary in this state, removing.

services/QuillLMS/engines/evidence/app/services/evidence/gen_ai/secondary_feedback_prompt_builder.rb
dandrabik: This service was originally just pulling the entire training file (or whatever file) and then the user would filter what they want. Looking at this with fresh eyes and based on this comment, I've refactored it so the `DataFetcher` does the fetching AND filtering, which I think makes more sense (and what you are driving at). Thanks for the feedback!

PR #12180, opened 2024-07-31
Translate hard coded feedback

samanthamjohn:
## WHAT
Translate the the hard-coded feedback strings. 

## WHY
So that things not in the database can still be shown in a second language. 

## HOW
Moved feedback strings to a JSON file. 
Also added a new rake task to translate the file. 

### Notion Card Links
https://www.notion.so/quill/Make-a-script-to-translate-the-hard-coded-feedback-f4b59e6675494849a438230d03add510?pvs=4

Ran the tests, tested that nothing broke in the front end

PR Checklist | Your Answer
------------ | -------------


services/QuillLMS/app/models/concerns/prompts.rb
brendanshean, line 1: `Prompts` is an overloaded term in our codebase.  

What about `TranslatablePrompts` or namespace within `Translatable::Prompts`

samanthamjohn: I went with TranslationPrompts, since it's used by things that are not translatable, and they are prompts to help you do translations

brendanshean: Great.  Thank you for updating

anathomical (approved):
It feels a bit weird to maintain YAML files in `app/models` and JSON files in `config` that are both related to translations, but I get how they are different use-cases, and don't feel like I have a better suggestion anyway.  So this seems reasonable to me.

services/QuillLMS/app/models/concerns/translation_prompts.rb
```diff
@@ -0,0 +1,52 @@
+# frozen_string_literal: true
+
+module TranslationPrompts
+  extend ActiveSupport::Concern
+  module ClassMethods
+    def prompt_start(locale:)
```
happythenewsad: What is the motivation for non-privatizing the refactored methods?

happythenewsad (approved):

services/QuillLMS/app/models/concerns/translation_prompts.rb
samanthamjohn: They ended up being class methods so they couldn't be private anymore.

samanthamjohn: updated this to just use a class for TranslationPrompts because there was a lot of weirdness with trying to call methods directly from a module in the rake task

PR #12183, opened 2024-08-01
Enforce making methodCallBraces symmetrical

samanthamjohn:
Enable this cop: 
https://www.rubydoc.info/gems/rubocop/RuboCop/Cop/Layout/MultilineMethodCallBraceLayout

> Checks that the closing brace in a method call is either on the same line as the last method argument, or a new line.
> 
> If a method call’s opening brace is on the same line as the first argument of the call, then the closing brace should be on the same line as the last argument of the call.
> 
> If an method call’s opening brace is on the line above the first argument of the call, then the closing brace should be on the line below the last argument of the call.


services/QuillLMS/app/lib/rule_feedback_history.rb
```diff
@@ -51,22 +51,29 @@ def self.feedback_history_to_json(f_h)
   end
 
   def self.generate_rulewise_report(rule_uid:, prompt_id:, start_date: nil, end_date: nil)
```
samanthamjohn, line 53: Updated this code to make it safer. (Even though I think it's just staff running it, but we got a security issue)

anathomical (approved):

PR #12187, opened 2024-08-01
guard-proofreader-from-missing-translation-error

anathomical:
## WHAT
Add a guard clause to protect against missing activity on props
## WHY
Because Proofreader doesn't actually set this value
## HOW
Check if `activity` exists before trying to destructure it, and bail early if it isn't there

### Screenshots
image

### Notion Card Links
https://www.notion.so/quill/Student-sees-a-blank-page-when-they-click-the-Resume-button-for-the-activity-The-Supremes-106e439473f643a3bdafd3c010fe6304

- Complete a Proofreader activity on staging
- Click "Save and exit"
- Confirm that when Resuming the activity, the page crashes when trying to load the follow-up Grammar activity
- Fix code and test
- Confirm by copying `data` blob from a known erroring production `ActiveActivitySession` to staging and confirm that Resuming does not crash the follow-up activity

PR Checklist | Your Answer
------------ | -------------
Have you added and/or updated tests? | No, there's no test coverage of the code with the edge case

### DIFF
--- a/services/QuillLMS/client/app/bundles/Grammar/actions/session.ts
+++ b/services/QuillLMS/client/app/bundles/Grammar/actions/session.ts
@@ -132,7 +132,16 @@ const activityUID = (): string => { return getParameterByName('uid', window.loca
 const handleGrammarSession = (session) => {
   return dispatch => {
     if (session && Object.keys(session).length > 1 && !session.error) {
-      QuestionApi.getAllForActivity(activityUID()).then((allQuestions) => {
+
+      // activityUID will not be set in cases where a user is resuming a
+      // Proofreader activity (that feeds into a Grammar activity).  We need
+      // to provide a fallback to load all questions in these cases so that
+      // the activity will fetch data properly
+      const fetchQuestions = activityUID()
+        ? QuestionApi.getAllForActivity(activityUID())
+        : QuestionApi.getAllForType(GRAMMAR_QUESTION_TYPE);
+
+      fetchQuestions.then((allQuestions) => {
         if (session.currentQuestion) {
           if (!session.currentQuestion.prompt || !session.currentQuestion.answers) {
             const currentQuestion = allQuestions[session.currentQuestion.uid]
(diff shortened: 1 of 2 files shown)


services/QuillLMS/client/app/bundles/Grammar/actions/session.ts
```diff
@@ -132,7 +132,16 @@ const activityUID = (): string => { return getParameterByName('uid', window.loca
 const handleGrammarSession = (session) => {
   return dispatch => {
     if (session && Object.keys(session).length > 1 && !session.error) {
-      QuestionApi.getAllForActivity(activityUID()).then((allQuestions) => {
+
+      // activityUID will not be set in cases where a user is resuming a
```
samanthamjohn, line 136: Question: did we confirm that not having questions populated was causing our error? i.e. could we return an empty object, or maybe just return one question? It seems that proofreader activities do not use questions at all, so it seems wasteful to pull all the questions here.

services/QuillLMS/client/app/bundles/Grammar/components/grammarActivities/intro.tsx
```diff
@@ -45,6 +45,10 @@ export default class Intro extends React.Component<IntroProps, IntroState> {
 
   translatedText = () => {
     const { activity, language } = this.props;
+
+    // Proofreader does not actually set the activity on props, so bail
+    if (!activity) return false
```
samanthamjohn, line 50: looks good!

samanthamjohn (changes requested):
I think the fetchQuestions could be done in another PR, because it's related to a different potential bug.

happythenewsad (approved):

services/QuillLMS/client/app/bundles/Grammar/actions/session.ts
anathomical: What's happening is that if you don't get all the questions right on the Proofreader activity, the system attempts to start the follow-up Grammar activity.  But the uid is still not set in the URL, so the Grammar activity is what's crashing.

samanthamjohn (approved):

PR #12220, opened 2024-08-07
GenAI Prototype - Connect Secondary Feedback and Highlighting

dandrabik:
## WHAT
Update GenAI prototype to check if a response needs secondary feedback AND if it does, generate that feedback.
## WHY
When students are stuck, we want to give them more direct feedback and highlight the text, so they can answer without getting frustrated.
## HOW
1. Added a `RepeatFeedbackChecker`
2. Connected the pieces together in the GenAI Check class to make the algorithm.

Logic is like this:
```mermaid

flowchart TD
    A[Make first LLM call for feedback and optimal] --> B{Is entry optimal?}
    B -- No --> D{Is this the first student attempt in session?}
    B -- Yes --> N[Return response]
    D -- No --> F{Is this the fifth student attempt in session, i.e. last attempt?}
    D -- Yes --> N[Return response]
    F -- No --> H[Make repeated_feedback LLM call]
    F -- Yes --> N[Return response]
    
    H --> I{Is the feedback a repeat of previoius feedback in this session?}
    I -- No --> N[Return response]
    I -- Yes --> M[Make secondary_feedback and highlighting LLM call]
    M --> O[Merge secondary_feedback and highlighting into response]
    O --> N[Return response]
```

### Notion Card Links
https://www.notion.so/quill/GenAI-Feedback-Algorithm-v1-9caa63caba7942ca8e2c2456240e3bad

I've tested throughly using a local script. I've also deployed this to my env where it is testable and is working both hitting the correct LLMs and highlighting items in the passage.

PR Checklist | Your Answer
------------ | -------------
Have you added and/or updated tests? |  Yes.


anathomical (approved):
I appreciate the logic diagram you included in this PR.  I also think that the approach of flagging for repetition on the generation of the primary feedback to be an interesting one.  It makes sense to do that in the prompt rather than try to write a heuristic that tries to figure out in Ruby whether two pieces of primary feedback are similar to each other.

brendanshean (commented):
Great work on getting these pieces connected.

Also really like the mermaid diagram

dandrabik:
> I appreciate the logic diagram you included in this PR. I also think that the approach of flagging for repetition on the generation of the primary feedback to be an interesting one. It makes sense to do that in the prompt rather than try to write a heuristic that tries to figure out in Ruby whether two pieces of primary feedback are similar to each other.

Yeah, we're starting with an LLM call. This is a modular piece, so we could always swap in different logic. We may use not use an LLM for this if we can get faster performance that is still accurate from something else.

PR #12224, opened 2024-08-08
Use Sidekiq batch for trial runs and g-eval scores

brendanshean:
## WHAT
* Remove `experiment` and `gen_ai_eval` queues
* Add `number` to Trial model
* Add GEvalScore model
* Parallelize the individual LLM calls during a Trial run
* Parallelize GEVal score fetching for an individual trial

## WHY
* These added complexity to the sidekiq configuration and now with increased capacity with API calls, these jobs can be moved to `low`
* It's useful to have a relative number for the Trial within the context of a dataset
* Each GEVal score is a third party call to an LLM so persisting this value facilitates keeping track of these calls when split across individual jobs.
* A Trial with 200 examples can take ~10-12 minutes, parallelizing the individual LLM will make jobs idempotent and speed up the trial run to under a minute
* Computing GEVal scores via individual jobs speeds up the score fetching from several minutes to under a minute.

## HOW
* Update Procfile{.dev} and remove `config/sidekiq_{experiment, gen_ai_eval}.yml`
* Add migration file and then add after_create callback.
* Create migration and model for GEVal score.
* Create `BuildLLMExampleWorker` that uses `Sidekiq::Batch job` in trial.rb 
* Create `GEvalScoresFetcher` that uses `Sidekiq::Batch` job to provision individual `RunGEvalWorker` jobs.

### Screenshots

### Notion Card Links

Tested on staging

PR Checklist | Your Answer
------------ | -------------


services/QuillLMS/engines/evidence/app/models/evidence/research/gen_ai/trial.rb
```diff
@@ -76,22 +78,25 @@ def average_g_eval_score
...
           return unless pending?
 
+          update!(trial_start_time: Time.zone.now)
           update!(status: RUNNING)
-          query_llm
-          CalculateResultsWorker.perform_async(id)
-          trial_errors.empty? ? update!(status: COMPLETED) : update!(status: FAILED)
+
+          batch = Sidekiq::Batch.new
+          batch.on(:complete, self.class, trial_id: id)
```
dandrabik, line 87: I like the use of `complete` rather than `success`.

services/QuillLMS/engines/evidence/app/models/evidence/research/gen_ai/trial.rb
```diff
@@ -104,24 +109,11 @@ def retry_params = { llm_id:, llm_prompt_id:, dataset_id: }
...
-              trial_errors << { error: e.message, test_example_id: test_example.id, raw_text: }.to_json
-              next
-            end
-            update_results(api_call_times:)
-          end
+        private def on_complete(_status, options)
+          trial = Trial.find(options['trial_id'])
+          trial.update!(trial_duration: Time.zone.now - Time.zone.parse(trial.trial_start_time))
+          CalculateResultsWorker.perform_async(options['trial_id'])
+          trial.trial_errors.empty? ? trial.update!(status: COMPLETED) : trial.update!(status: FAILED)
```
dandrabik, line 116: If I'm reading this correctly, if there were 99 successful jobs and 1 errored job, this trial would be labeled `Failed` but the final results of the job would still be calculated at that moment. I think that's reasonable and I like that the final status is calculated.

In the future, we might consider a way to retry the errors (kinda like CircleCI's 'run from failed'). Probably over-engineering at this point, but if we have a lot of partial failures, we should consider it.

services/QuillLMS/engines/evidence/app/models/evidence/research/gen_ai/trial.rb
```diff
@@ -76,22 +78,25 @@ def average_g_eval_score
         def scores = g_evals&.values&.flatten&.compact&.map(&:to_f)
 
         def run
```
dandrabik, line 80: You might consider extracting this `run` method and its support methods out of the model and into a service or worker.

services/QuillLMS/engines/evidence/app/workers/evidence/research/gen_ai/build_llm_example_worker.rb
```diff
@@ -0,0 +1,33 @@
+# frozen_string_literal: true
+
+module Evidence
+  module Research
+    module GenAI
+      class BuildLLMExampleWorker
+        include Evidence.sidekiq_module
+
+        sidekiq_options retry: 0, queue: 'low'
```
dandrabik: I think we want these in the `default` queue. if we are considering staff to be "users" who are waiting on info in the UI, you could even make a case for `critical_external`, but I think since it's staff, `default` is a nice middle ground.

services/QuillLMS/engines/evidence/app/workers/evidence/research/gen_ai/calculate_results_worker.rb
```diff
@@ -6,16 +6,15 @@ module GenAI
       class CalculateResultsWorker
         include Evidence.sidekiq_module
 
-        sidekiq_options retry: 3, queue: 'gen_ai_eval'
+        sidekiq_options retry: 3, queue: 'low'
```
dandrabik: To be clear, I think these all should be `default` queue.

dandrabik (approved):
Overall looks good. I think we should use the `default` queue in sidekiq and a couple of other small suggestions for refactor (one is a future item that I don't think we should do now).

Thanks for your work on this.

services/QuillLMS/engines/evidence/app/models/evidence/research/gen_ai/trial.rb
brendanshean: Yeah.  I kind of debated on adding something like this in, but in general, the trials seem to be going through without errors so I punted on it.

PR #12239, opened 2024-08-13
evidence student pop up

emilia-friedberg:
## WHAT
Add a pop-up on the student dashboard when a user has completed an Evidence activity prior to this year and is now starting a new one.

## WHY
We want to make sure students realize the activities are now being graded.

## HOW
Just add information about whether the student has previously completed Evidence activities to the JSON response, and use that on the frontend to determine whether or not to show the modal.

### Screenshots
Screenshot 2024-08-14 at 10 14 40 AM

### Notion Card Links
https://www.notion.so/quill/Evidence-Scoring-Project-Part-7-Student-Dashboard-Includes-a-new-Custom-Pop-Up-for-Students-who-Us-10c10ad1aa0c4aae98b39c3176a1e67e?pvs=4

PR Checklist | Your Answer
------------ | -------------
Have you deployed to Staging? | NO - will deploy all at once after next PR


services/QuillLMS/app/controllers/profiles_controller.rb
```diff
@@ -190,6 +191,13 @@ def staff
     @act_sesh_records.first
   end
 
+  protected def completed_evidence_activity_prior_to_scoring
+    ActivitySession
+      .where(user_id: current_user.id, activity_id: Activity.evidence.ids)
+      .where("completed_at < ?", DateTime.new(2024, School::SCHOOL_YEAR_START_MONTH, 1))
+      .present?
```
anathomical: This isn't super-obvious from the Rails code, but this particular approach makes two database calls: one to get a list of Activity IDs, and then a second one to look up ActivitySessions.  Given how often this looks like it might get called, it's probably worth making this slightly more efficient if we can.  An alternative version of this query which removes that first query looks like this:
```
ActivitySession
  .joins(activity: :classification)
  .where(activity: {classification: ActivityClassification.evidence})
  .where(user_id: current_user.id)
  .where("completed_at < ?", DateTime.new(2024, School::SCHOOL_YEAR_START_MONTH, 1))
  .present?
```

anathomical (approved):
One suggestion on a slightly different approach to the database lookup, but this looks fine and could go out as-is.

happythenewsad (approved):

PR #12240, opened 2024-08-13
Fix Connect resume bugs

anathomical:
## WHAT
Update the logic used for starting Connect activities so that we don't bypass `resume` calls when generating translation text.

Additionally, remove some dead code that can never be triggered.  Specifically, `showIntro` is always true for Connect activities in production, so there's no need for the fallbacks.
## WHY
`showIntro` is always true, so this code was resulting in all clicks to start a Connect activity from the landing page starting the activity from the beginning even if there was an in-flight session available.
## HOW
Move the translation logic into an existing no-in-flight-session logic flow.

### Notion Card Links
https://www.notion.so/quill/Unable-to-save-progress-when-pressing-the-Save-and-Exit-button-and-resuming-a-Quill-Connect-activity-b1808666a2134afead2fad6aa17ece8a

- Start a new Connect activity
- Answer one questions
- Save and Exit
- Resume activity
- Confirm that the activity starts on the second question (instead of the first, which it has been doing)

PR Checklist | Your Answer
------------ | -------------

### DIFF
--- a/services/QuillLMS/client/app/bundles/Shared/components/studentLessons/register.tsx
+++ b/services/QuillLMS/client/app/bundles/Shared/components/studentLessons/register.tsx
@@ -40,11 +40,6 @@ class Register extends React.Component<any, any> {
     }
   }
 
-  handleStartLessonClick = () => {
-    const { startActivity, } = this.props
-    startActivity();
-  }
-
   resume = () => {
     const { resumeActivity, session, } = this.props
     resumeActivity(session);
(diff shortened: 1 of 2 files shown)


services/QuillLMS/client/app/bundles/Shared/components/studentLessons/register.tsx
```diff
@@ -96,56 +87,19 @@ class Register extends React.Component<any, any> {
...
-            {this.renderButton(showIntro)}
-            <br />
-          </div>
-        </div>
-      );
-    }
+
+    return (
+      <div className="container">
+        <div className="landing-page-html" dangerouslySetInnerHTML={{ __html: lesson.landingPageHtml, }} />
```
emilia-friedberg: We might still want some kind of fallback here, just because even though it seems like we do always have `landingPageHtml`, it would be possible to create a lesson without it and we wouldn't want that to break.

emilia-friedberg (approved):
One comment about adding a fallback for an unlikely scenario, but otherwise LGTM!

services/QuillLMS/client/app/bundles/Shared/components/studentLessons/register.tsx
anathomical: Fair point.  I'll put that condition back in.

PR #12244, opened 2024-08-14
GenAI - Gemini for GenAI production path

dandrabik:
## WHAT
Add the ability to use Gemini for the GenAI check. This adds a new `Evidence::Gemini::Chat` API wrapper
## WHY
We want to test with Gemini since we are demoing for Google and our Gemini research has had good results. Also like we've been doing with OpenAI, we want to use the system prompt and send the chat history when this is live (which the current API wrapper doesn't do) to (maybe) give enough context so it has a more natural chat.
## HOW
Create a `Evidence::Gemini::Chat` which is different than the `Evidence::Gemini::Completion` used by the `Research` namespace. Following the pattern of before and creating a different API rather than trying to make on class work for both use cases. Also, I made this match the protocol of the OpenAI version of `Chat`, to make switching back and forth easy (change one constant).

### Notion Card Links
https://www.notion.so/quill/Use-Gemini-for-GenAI-checks-3cdcf8001c4849b8916171bdaac4bd53

I have a thor task `bundle exec thor gen_a_i_tasks:example_check` that runs through the whole pipeline and hits the primary, repeated, and secondary feedback calls. I've been using this as an integration test of sorts, since it prints all the relevant information to the screen (prompts, response, timing). I could convert this to an external_api rspec as well.

PR Checklist | Your Answer
------------ | -------------
Have you deployed to Staging? | Not yet - deploying soon.

### DIFF
--- a/services/QuillLMS/engines/evidence/app/services/evidence/gen_ai.rb
+++ b/services/QuillLMS/engines/evidence/app/services/evidence/gen_ai.rb
@@ -1,3 +1,5 @@
+# frozen_string_literal: true
+
 module Evidence
   module GenAI
     HistoryItem = Struct.new(:user, :assistant, keyword_init: true)
--- a/services/QuillLMS/engines/evidence/app/services/evidence/gen_ai.rb
+++ b/services/QuillLMS/engines/evidence/app/services/evidence/gen_ai.rb
@@ -2,6 +2,6 @@
 
 module Evidence
   module GenAI
-    HistoryItem = Struct.new(:user, :assistant, keyword_init: true)
+    HistoryItem = Data.define(:user, :assistant)
   end
 end
--- a/services/QuillLMS/engines/evidence/app/services/evidence/gen_ai.rb
+++ b/services/QuillLMS/engines/evidence/app/services/evidence/gen_ai.rb
@@ -1,0 +1,5 @@
+module Evidence
+  module GenAI
+    HistoryItem = Struct.new(:user, :assistant, keyword_init: true)
+  end
+end
--- a/services/QuillLMS/engines/evidence/app/services/evidence/gen_ai/repeated_feedback_checker.rb
+++ b/services/QuillLMS/engines/evidence/app/services/evidence/gen_ai/repeated_feedback_checker.rb
@@ -4,14 +4,14 @@ module Evidence
   module GenAI
     class RepeatedFeedbackChecker < ApplicationService
       KEY_REPEAT = 'repeat_feedback'
-      MODEL = 'gpt-4o-mini'
 
-      attr_reader :feedback, :history, :optimal
+      attr_reader :feedback, :history, :optimal, :chat_api
 
-      def initialize(feedback:, history: [], optimal: false)
+      def initialize(feedback:, history: [], optimal: false, chat_api: Evidence::OpenAI::Chat)
         @feedback = feedback
         @history = history
         @optimal = optimal
+        @chat_api = chat_api
       end
 
       def run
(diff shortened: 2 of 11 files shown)


services/QuillLMS/engines/evidence/app/services/evidence/gen_ai/repeated_feedback_checker.rb
```diff
@@ -24,7 +24,7 @@ def run
 
       private def system_prompt = Evidence::GenAI::RepeatedFeedbackPromptBuilder.run(prompt: nil, history:)
 
-      private def llm_response = Evidence::OpenAI::Chat.run(system_prompt:, entry: feedback, model: MODEL)
+      private def llm_response = chat_api.run(system_prompt:, entry: feedback, model: chat_api::SMALL_MODEL)
```
anathomical, line 27: I really like the idea of using the same interface/expectations for both OpenAI and Gemini Chat services.

services/QuillLMS/engines/evidence/lib/evidence/open_ai/chat.rb
```diff
@@ -20,7 +19,7 @@ class Chat < Evidence::ApplicationService
 
       attr_reader :system_prompt, :entry, :history, :temperature, :model
 
-      def initialize(system_prompt:, entry:, history: [], temperature: 0.5, model: DEFAULT_MODEL)
+      def initialize(system_prompt:, entry:, history: [], temperature: 1, model: DEFAULT_MODEL)
```
anathomical, line 22: Is it worth setting up a `DEFAULT_TEMPERATURE` constant to match the pattern we're using for `DEFAULT_MODEL`?

anathomical (approved):

services/QuillLMS/engines/evidence/app/services/evidence/gen_ai.rb
```diff
@@ -0,0 +1,7 @@
+# frozen_string_literal: true
+
+module Evidence
+  module GenAI
+    HistoryItem = Struct.new(:user, :assistant, keyword_init: true)
```
brendanshean: Maybe an opportunity to use the new value object interface: `Data.define(:user,...)`

services/QuillLMS/engines/evidence/lib/evidence/gemini/chat.rb
```diff
@@ -0,0 +1,79 @@
...
+          [
+            { KEY_ROLE => ROLE_USER, KEY_PARTS => [{ KEY_TEXT => h.user }] },
+            { KEY_ROLE => ROLE_ASSISTANT, KEY_PARTS => [{ KEY_TEXT => h.assistant }] }
+          ]
+        end.flatten
+      end
+
+      private def generation_config = { temperature:, response_mime_type: }
+
+      private def response_mime_type = 'application/json'
```
brendanshean: I would extract a constant here:
`JSON_MIME_TYPE = 'application/json'`

brendanshean (approved):
Nice work on connecting these pieces!  It's exciting to see this coming together.

services/QuillLMS/engines/evidence/lib/evidence/open_ai/chat.rb
dandrabik: Will do!

services/QuillLMS/engines/evidence/app/services/evidence/gen_ai.rb
dandrabik: Definitely, updating!

services/QuillLMS/engines/evidence/lib/evidence/gemini/chat.rb
dandrabik: Good call!

PR #12256, opened 2024-08-19
building ai knowledge course page

emilia-friedberg:
## WHAT
Build out the Building AI Knowledge course view page.

## WHY
The social studies half of this project is complete, but now the material is ready for the science half. This is likely the second-to-last PR for this project (subject page forthcoming).

## HOW
Largely mirror the structure of the World History page and routing, with component logic further extended/abstracted where appropriate to cover both sets of data. 

### Screenshots
Screenshot 2024-08-19 at 1 31 01 PM

### Notion Card Links
https://www.notion.so/quill/Building-AI-Knowledge-Course-Page-852801de37604f27a7fcc0b92cad0e52?pvs=4

Established that it works as expected for all the paths in which it needs to be accessible, and looks good on all screen sizes. Jack and Jamie will QA on staging.

PR Checklist | Your Answer
------------ | -------------
Have you deployed to Staging? | Not yet -- waiting for goahead to copy production data to staging so we can test with the actual unit templates


services/QuillLMS/spec/controllers/teachers/progress_reports_controller_spec.rb
```diff
@@ -243,4 +243,17 @@
...
     end
   end
+
+  describe '#building_ai_knowledge_unit_templates' do
+    before do
+      allow(controller).to receive(:course_with_assignment_data).and_return([])
+    end
+
+    it 'renders json with a unit_templates key' do
+      get :building_ai_knowledge_unit_templates
```
anathomical: This is pretty minor, but you could move this `get` call into the `before` block and get the same behavior.  I think that's slightly easier to expand if we ever added more specs here.  Still, it doesn't really change anything, so take or leave it.

services/QuillLMS/spec/helpers/navigation_helper_spec.rb
```diff
@@ -438,4 +438,38 @@
...
     end
   end
+
+  describe '#show_science_dashboard_tab?' do
+    subject { helper.show_science_dashboard_tab?(current_user) }
+
+    context 'when user has unit activities including social studies activities' do
+      before do
+        allow(helper).to receive(:unit_activities_include_science_activities?).and_return(true)
+      end
```
anathomical: This is also organizational, so take or leave it.  The `before` blocks in this context and the following two are very similar, only really differing in the value we want to return.  You could simplify this by doing something like:
```
before do
  allow(helper).to receive(:unit_activities_include_science_activities?).and_return(helper_result)
end

context 'when user has unit activities including social studies activities' do
  let(:helper_result) { true }
  ...
end

context  'when user does not have unit activities including social studies activities' do
  let(:helper_result) { false }
  ...
end
```

anathomical (approved):
A couple of take-or-leave comments about spec organization, but this looks good to me.

services/QuillLMS/app/helpers/content_hubs_helper.rb
```diff
@@ -14,7 +14,14 @@ def unit_activities_include_social_studies_activities?(unit_activities)
     unit_activities_include_content_activities?(unit_activities, activity_ids)
   end
 
-  def unit_activities_include_science_activities?(unit_activities) = false
+  def unit_activities_include_science_activities?(unit_activities)
+    activity_ids = building_ai_knowledge_data
```
brendanshean: I think you can use `flat_map` and `pluck`
```
  activity_ids = building_ai_knowledge_data
    .flat_map { |template| template[:activities].pluck(:activity_id) }
```

services/QuillLMS/client/app/bundles/Teacher/containers/WorldHistory1200ToPresent.tsx
```diff
@@ -14,99 +13,40 @@ const LOCAL_STORAGE_KEY = 'worldHistory1200ToPresentExpandedUnitTemplateIds'
 const SLUG = 'world-history-1200-to-present'
 
 const WorldHistory1200ToPresent = ({ backlinkPath, }) => {
```
brendanshean, line 15: I like the refactoring here!

brendanshean (approved):
Nice work on getting this new page setup!

PR #12258, opened 2024-08-19
fix-teacher-diagnostic-student-report-ordering

anathomical:
## WHAT
Explicitly sort Skill Groups by order_number
## WHY
So that Curriculum specification controls the displayed order of skills, since those are sequenced
## HOW
Add an explicit `order` clause to the code that fetches diagnostic activity Skill Groups

### Screenshots
OLD
image
NEW
image

### Notion Card Links
https://www.notion.so/quill/2024-Long-Term-Diagnostic-Bugs-Fixes-For-Mid-August-f7440092d6924dfcb8ae2f28c03f1475?pvs=4#b9ec1338a58b4032bd34a58b161ba996

- Ensure that Student Report for Teacher Diagnostics shows skill columns in the intended order (defined by `order_by` on `skill_group_activities`)

PR Checklist | Your Answer
------------ | -------------

### DIFF
--- a/services/QuillLMS/app/controllers/concerns/results_summary.rb
+++ b/services/QuillLMS/app/controllers/concerns/results_summary.rb
@@ -8,7 +8,7 @@ module ResultsSummary
 
   def results_summary(activity_id, classroom_id, unit_id)
     activity = Activity.find(activity_id)
-    @skill_groups = activity.skill_groups
+    @skill_groups = activity.skill_groups.order(:order_number)
     set_activity_sessions_and_assigned_students_for_activity_classroom_and_unit(activity_id, classroom_id, unit_id, hashify_activity_sessions: true)
     @skill_group_summaries = @skill_groups.map do |skill_group|
       {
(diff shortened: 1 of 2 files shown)


emilia-friedberg (approved):
This LGTM. I almost wonder if it's worth doing a `default_scope` on the `skill_groups` model to order them this way everywhere automatically? I know `default_scope` can be iffy but I can't really think of a case where we wouldn't want these ordered.

anathomical:
> This LGTM. I almost wonder if it's worth doing a `default_scope` on the `skill_groups` model to order them this way everywhere automatically? I know `default_scope` can be iffy but I can't really think of a case where we wouldn't want these ordered.

I think that makes a ton of sense.  I'll update to take that approach instead.

anathomical:
> > This LGTM. I almost wonder if it's worth doing a `default_scope` on the `skill_groups` model to order them this way everywhere automatically? I know `default_scope` can be iffy but I can't really think of a case where we wouldn't want these ordered.
> 
> I think that makes a ton of sense. I'll update to take that approach instead.

Actually, after further discussion with the team, we'll skip the `default_scope`, but I do like the idea of making this a regular `scope`, so I'll refactor for that.

brendanshean (approved):

PR #12259, opened 2024-08-19
pluralize-skills-properly

anathomical:
## WHAT
- Add a pluralize utility method
- Use it for student response reporting around the word "Skill(s)"
## WHY
It's a nice readability feature we should provide if we can
## HOW
- Create a new `Shared/lib` module for pluralization (I modeled the signature off of Rails' `pluralize`)
- Use new module in existing template

### Screenshots
Screenshot 2024-08-19 at 4 56 32 PM
Screenshot 2024-08-19 at 4 59 13 PM

### Notion Card Links
https://www.notion.so/quill/New-Diagnostic-Engineering-Tasks-ef849d3f473a4a11a675081a89556a75?pvs=4#673308dd75f04dc0ba79485810c44238

Find an example teacher who has a student with only 1 skill to practice, confirm that text is singular as expected.

PR Checklist | Your Answer
------------ | -------------
Have you added and/or updated tests? | No test coverage here

### DIFF
--- a/services/QuillLMS/client/app/bundles/Shared/index.tsx
+++ b/services/QuillLMS/client/app/bundles/Shared/index.tsx
@@ -153,7 +153,8 @@ export {
   getlanguageOptions,
   renderSaveAndExitButton,
   hasTranslationFlag,
-  showTranslations
+  showTranslations,
+  pluralize
 } from './libs/index'
 
 export {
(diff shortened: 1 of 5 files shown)


emilia-friedberg (approved):
LGTM, but I think it would be nice to have a test for the new lib function.

anathomical:
> LGTM, but I think it would be nice to have a test for the new lib function.

Good call.  I've added one and re-requested your review just to make sure I'm not missing any obvious test cases that I should include.

emilia-friedberg (approved):
Thanks for doing that!

PR #12269, opened 2024-08-22
interdisciplinary science subject page

emilia-friedberg:
## WHAT
Add the subject page for Interdisciplinary Science (aiEDU offerings), and some small copy changes that Jamie found in project QA. 

## WHY
This is the last big piece of the content hub project!

## HOW
Largely use the same structure I did for the social studies subject page, but extract shared code where possible.

### Screenshots
Screenshot 2024-08-22 at 8 13 28 AM

### Notion Card Links
https://www.notion.so/quill/Interdisciplinary-Science-Subject-Page-abd90f0406f441658191e4ec95e903fa?pvs=4

Tested all routes to this page, the page itself on various browser sizes, and had Jack and Jamie QA.

PR Checklist | Your Answer
------------ | -------------


brendanshean (approved):
This looks great!  Nice work on this

services/QuillLMS/app/assets/stylesheets/pages/content_hub/subject_page.scss
```diff
@@ -188,4 +195,17 @@
...
+      background-image: url("https://assets.quill.org/images/pages/social_studies_subject_page/header-backdrop.svg");
+    }
+  }
+  &.interdisciplinary-science-subject-page {
+    header {
+      background-image: url("https://assets.quill.org/images/pages/interdisciplinary_science_subject_page/header_backdrop.svg");
+    }
+    .partners {
+      border-top: 1px solid $quill-grey-15;
+    }
```
anathomical, line 209: No strong feelings about this so take or leave it, but I assume that this sub-selector is applied only to the science section because of the way these are listed next to each other.  I wonder if it'd make sense to do something more generic like adding a `:first-child` selector to the general `.partners` selector further up so that if/when we add more of these, or re-order them, we don't have to tweak the CSS for this.

services/QuillLMS/client/app/bundles/Teacher/components/content_hubs/subject_page_teacher_center_articles.tsx
```diff
@@ -0,0 +1,72 @@
...
+  return renderPreviewContent(blogPost)
+}
+
+const SubjectPageTeacherCenterArticles = () => (
+  <section className="teacher-center-articles">
+    <div className="container">
+      <h2>Learn more with these Teacher Center articles</h2>
+      <div className="articles-wrapper">
+        <TeacherCenterArticleCard blogPostId={GETTING_STARTED_COLLECTION_ID} />
+        <TeacherCenterArticleCard blogPostId={INTEGRATING_QUILL_EVIDENCE_ARTICLE_ID} />
```
anathomical, line 66: This is purely stylistic so take or leave it, but maybe it'd be cleaner to expand this if we used some sort of `map` call to build each of these `TeacherCenterArticleCard` instances rather than just copy/pasting the line with new constants in each of them.  Given that there are just two of them, this probably doesn't matter here, I'm just imagining a future where we had three or four more, which may not be likely.

anathomical (approved):
I like the way that you've refactored a lot of the re-used bits into separate components.

happythenewsad (approved):

services/QuillLMS/app/assets/stylesheets/pages/content_hub/subject_page.scss
emilia-friedberg: This is actually just a temporary measure while we wait to get permission to use some teacher quotes. In the final design, we won't have this line, because there will be a different block here. So I agree with your recommendation in general, but I think in this instance since it'll be removed soon I'll leave it as is.

services/QuillLMS/client/app/bundles/Teacher/components/content_hubs/subject_page_teacher_center_articles.tsx
emilia-friedberg: I agree that having just the two makes me feel like it's okay to leave for now, but I will definitely keep this in mind if we expand later (or start using different articles for different pages).

PR #12273, opened 2024-08-22
Evidence GenAI - Environment flag for GenAI (and bug fix)

dandrabik:
## WHAT
Add an ENV var to turn on `GenAI` feedback for an app (will use for staging). When making this change and running tests, I found a bug that need fixing (using memoization in a class method) which caused spurious test failures since the memoized var wasn't
## WHY
To use this for demoing on a staging environment. It's gotten a little cumbersome to make and then revert a hardcoded swap of `GenAI` and `AutoML`
## HOW
1. Make a small env var change
2. Copy current tests with a `GenAI` variation
3. See that those tests fail
4. Fix tests, remove tests that actually don't work, refactor because of lint complexity flag.
I've tested this locally in the app and in the console to see that the ENV was picked up. Deploying to staging and will confirm there as well. 

PR Checklist | Your Answer
------------ | -------------
Have you deployed to Staging? | deploying now!


services/QuillLMS/engines/evidence/lib/evidence/check.rb
```diff
@@ -4,12 +4,15 @@ module Evidence
...
+    USE_GEN_AI = ENV.fetch('EVIDENCE_USE_GEN_AI', 'false') == 'true'
+    AI_CHECK = USE_GEN_AI ? GenAI : AutoML
+
     ALL_CHECKS = [
       Prefilter,
       RegexSentence,
       Opinion,
       Plagiarism,
-      AutoML,
+      AI_CHECK,
```
dandrabik, line 15: This is the full code change (not including bug fixes to get tests to pass).

services/QuillLMS/engines/evidence/lib/evidence/check.rb
```diff
@@ -4,12 +4,15 @@ module Evidence
...
       RegexPostTopic,
       Grammar,
       Spelling,
```
dandrabik, line 18: Everything above is the full code change (not including bug fixes to get tests to pass).

services/QuillLMS/engines/evidence/lib/evidence/check.rb
dandrabik, line 15: This is the full code change (not including bug fixes to get tests to pass).

services/QuillLMS/engines/evidence/spec/lib/check_with_genai_spec.rb
brendanshean, line 1: I noticed many `allow_any_instance` and `expect_any_instance` instances in these tests.  The RSpec maintainers have labeled this old syntax and recommend [against](https://rspec.info/features/3-12/rspec-mocks/old-syntax/any-instance/) using these constructs and I would agree with them.

I think for purposes of the demo they are fine, but long term we probably want to move away from them.

brendanshean (approved):
There's one comment about {expect, allow}_any_instance worth thinking about, otherwise looks good.

anathomical (approved):

services/QuillLMS/engines/evidence/spec/lib/check_with_genai_spec.rb
dandrabik: I agree. This is somewhat temporary code (we will likely store `GenAI` or `AutoML` config at the `evidence_activity` level vs globally) and it's just copies of existing tests (changing `AutoML` to `GenAI`), so I'm going to leave for now for speed of getting this out, but we should clean these up in the future.

PR #12278, opened 2024-08-23
Back apply evidence question scores

anathomical:
## WHAT
- Add a worker to look at `ConceptResults` for a given `ActivitySession` and update the `question_score` values based on updated scoring logic
- Add a temporary Rake task to enqueue all old Evidence ActivitySessions for evaluation by this new worker
## WHY
The "new" scoring logic was implemented in November 2023, and better reflects our philosophy of what constitutes student success than the old scoring logic (which we implemented when we never thought we'd "use" scoring).  While it's unlikely that many, if any, users will be looking at scoring data this old, it's still worth updating in case we change our minds later.
## HOW
- Using an `ActivitySession` id, pull all of its `ConceptResults`, group them by `question_number`, calculate a score for that question based on the logic of 100% if the student got an optimal response in 5 attempts, and 0% if they had a non-optimal response on their fifth attempt.  Then save that updated score to the `question_score` value for the relevant `ConceptResults`
- Create a Rake task to find all `ActivitySessions` completed prior to 2023-11-15, a few days after we deployed the new logic for scoring to production.  This should catch all of the records scored with the old style, and no-op on the items that already have the new style assigned.

### Notion Card Links
https://www.notion.so/quill/Scoring-Change-Tag-Evidence-Sessions-As-Optimal-When-5th-Attempt-is-Optimal-Backfill-5d9a8b503ee544bdb9b789e212b6ae62?pvs=4

- Deploy to staging
- Find five example Evidence ActivitySessions with the old scoring applied (139626058,139659873,139707729,139714240,139717600)
- Run Rake task (~10 minutes to enqueue all records)
- Confirm queued jobs (to ballpark queue size: 456,822)
- Turn on `migration` workers in the staging environment
- Benchmark job process rate to estimate total completion time (2 1x workers, ~50 minutes)
- Confirm that the five example ActivitySessions have new scoring applied

PR Checklist | Your Answer
------------ | -------------

### DIFF
--- a/services/QuillLMS/app/workers/evidence_update_scoring_worker.rb
+++ b/services/QuillLMS/app/workers/evidence_update_scoring_worker.rb
@@ -1,0 +1,51 @@
+# frozen_string_literal: true
+
+class EvidenceUpdateScoringWorker
+  include Sidekiq::Worker
+  sidekiq_options queue: SidekiqQueue::MIGRATION
+
+  attr_reader :concept_results
+
+  def perform(activity_session_id)
+    @concept_results = ConceptResult.where(activity_session_id:)
+
+    return if concept_results.empty?
+    return if unscored?
+
+    process_grouped_by_question
+  end
+
+  private def unscored? = concept_results.first.question_score.nil?
+  private def updated_score(question_results) = last_attempt_correct?(question_results) ? 1.0 : 0.0
+
+  private def process_grouped_by_question
+    concept_results.group_by(&:question_number).values.each do |question_results|
+      # All cases in the old scoring system that were 100s are still 100s in the updated system
+      next if question_results.first.question_score == 1
+
+      update_scores(question_results)
+    end
+  end
+
+  private def update_scores(question_results)
+    question_score = updated_score(question_results)
+
+    question_results.each do |concept_result|
+      next if concept_result.question_score == question_score
+
+      concept_result.update(question_score:)
+    end
+  end
+
+  private def last_attempt_correct?(question_results)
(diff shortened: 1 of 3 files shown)


happythenewsad (approved):
Approving the code in isolation, but FYI, the Airbyte instance uses `created_at` as a cursor, so these updates will not be visible from BigQuery queries as-is.

brendanshean (approved):
LGTM.  Thank you for the detailed explanation in the PR description.

services/QuillLMS/app/workers/evidence_update_scoring_worker.rb
```diff
@@ -0,0 +1,51 @@
...
+class EvidenceUpdateScoringWorker
+  include Sidekiq::Worker
+  sidekiq_options queue: SidekiqQueue::MIGRATION
+
+  attr_reader :concept_results
+
+  def perform(activity_session_id)
+    @concept_results = ConceptResult.where(activity_session_id:)
+
+    return if concept_results.empty?
```
dandrabik, line 12: Consider adding a check that this is an `Evidence` activity in case this is re-used or there's a bug in how the jobs get queued.

dandrabik (approved):
Small comment about confirming it's an `evidence` activity, but otherwise looks good!

anathomical:
Per [discussion in this Notion card](https://www.notion.so/quill/BLOCKED-Scoring-Change-Tag-Evidence-Sessions-As-Optimal-When-5th-Attempt-is-Optimal-Backfill-5d9a8b503ee544bdb9b789e212b6ae62), we've decided not to proceed with this backfill.

PR #12304, opened 2024-08-28
Evidence GenAI - Only use previous feedback for prompt (not session)

dandrabik:
## WHAT
A single `FeedbackSession` keeps track of history for ALL three prompts (one session -> many prompts). For "previous feedback" we only want the focus on the current prompt, so use the browser tracked `previous_feedback` which has all we need.
## WHY
Feeding this `FeedbackSession` "chat history" into the GenAI model was causing feedback related from previous prompts (e.g. the `because` stem) to influence the current prompt (e.g. the `but` stem). Even though the system prompt had changed to the system prompt for the current conjunction, the "chat history" seems to have influenced it to continue the conversation from the previous prompt.
## HOW
Use the previous feedback from the browser which is more efficient and resets on each prompt.
Remove all feedback_session code from Evidence since it's not needed anymore.
### Notion Card Links
https://www.notion.so/quill/Issue-with-prompt-feedback-bleeding-across-prompts-b3f0c38551774bba81c26c85f0fcbedf

Running the app locally and log the history to confirm the proper history and feedback is happening. Since this is a bug, I also, I also tested `develop` locally to confirm that the cross-prompt issue was happening (but went away on this branch).

PR Checklist | Your Answer
------------ | -------------
Have you added and/or updated tests? |  'YES


services/QuillLMS/app/models/feedback_history.rb
```diff
@@ -99,7 +99,7 @@ class FeedbackHistory < ApplicationRecord
   scope :autoML, -> { where(feedback_type: AUTO_ML) }
   scope :confidence_greater_than, ->(lower_limit) { where("CAST(metadata->'api'->'confidence' AS DOUBLE PRECISION) > ?", lower_limit) }
   scope :entry_shorter_than, ->(length) { where('LENGTH(entry) < ?', length) }
-  scope :for_prompt, ->(prompt_id) { where(prompt_id: prompt_id) }
+  scope :for_prompt, ->(prompt_id) { where(prompt_id: prompt_id, prompt_type: DEFAULT_PROMPT_TYPE) }
```
brendanshean: I think you can write `prompt_id:`.

I thought we had a cop for this.

services/QuillLMS/engines/evidence/lib/evidence/check/gen_ai.rb
```diff
@@ -38,23 +42,29 @@ def use_for_optimal_feedback? = true
...
-        @history ||= session
-          .feedback_history
-          .sort_by(&:id)
-          .map { |fh| Evidence::GenAI::HistoryItem.new(user: fh.entry, assistant: fh.feedback_text) }
-      end
+      private def history_feedback_only = previous_feedback.map { |f| f[KEY_FEEDBACK] }
 
-      private def session
-        @session ||= Evidence.feedback_session_class.find_by(activity_session_uid: session_uid)
+      private def with_logging
```
brendanshean, line 58: Is this logging temporary?
Do we want to restrict to staging environments?

brendanshean (approved):
Nice work on figuring out this subtle effect.

services/QuillLMS/engines/evidence/lib/evidence/check/gen_ai.rb
dandrabik: This is temporary, but I'd like to leave it in for a bit for debugging purposes on beta, but will leave a comment to that effect. 

This GenAI check should be off in production, so the additional `Rails.production?` check I'm planning to add there should, in effect, restrict this to non-prod.

services/QuillLMS/app/models/feedback_history.rb
dandrabik: Good call, updating.

PR #12309, opened 2024-08-28
Add parent_id to Dataset model

brendanshean:
## WHAT
Add parent_id `Evidence::Research::GenAI::DataSlice` model

## WHY
This feature will help distinguish complete datasets from partial datasets (i.e. data subsets)

## HOW
Write a migration to add the `parent_id` attribute to the Dataset model.

### Screenshots

### Notion Card Links

This is just a migration.

PR Checklist | Your Answer
------------ | -------------


services/QuillLMS/engines/evidence/app/models/evidence/research/gen_ai/dataset.rb
```diff
@@ -21,6 +21,7 @@ class Dataset < ApplicationRecord
         has_many :prompt_examples, dependent: :destroy
         has_many :trials, dependent: :destroy
         has_many :comparisons, dependent: :destroy
+        has_many :dataslices, dependent: :destroy
```
dandrabik: Should this be `data_slices`? Also, do we want the dependent destroy? What if experiments have been run.

services/QuillLMS/db/migrate/20240828204236_create_evidence_research_gen_ai_data_slices.evidence.rb
```diff
@@ -0,0 +1,11 @@
+# frozen_string_literal: true
+
+# This migration comes from evidence (originally 20240828195901)
+class CreateEvidenceResearchGenAIDataSlices < ActiveRecord::Migration[7.0]
```
dandrabik: Naming-related, so take or leave, but consider `DataSet` and `DataSubset` instead of `DataSet` and `DataSlice`.

services/QuillLMS/engines/evidence/app/models/evidence/research/gen_ai/data_slice.rb
```diff
@@ -0,0 +1,27 @@
...
+#  id                :bigint           not null, primary key
+#  created_at        :datetime         not null
+#  updated_at        :datetime         not null
+#  child_dataset_id  :integer          not null
+#  parent_dataset_id :integer          not null
+#
+module Evidence
+  module Research
+    module GenAI
+      class DataSlice < ApplicationRecord
```
dandrabik: Take or leave, but if this is always a one-to-many parent/children relationship, you could put a `parent_id` on `DataSet`, so everything would still be a `DataSet`. It would leave mostly `null` values there, but it would be less code and less models.

I honestly don't have a strong opinion either way, just think it's worth examining.

dandrabik (approved):
Overall, this looks good. I think it's worth considering whether this should just be a `parent_id` on DataSet.

services/QuillLMS/db/migrate/20240828204236_create_evidence_research_gen_ai_data_slices.evidence.rb
brendanshean: Yeah. I agree. After some more code with this, I found it to be an annoying inconsistency.

services/QuillLMS/engines/evidence/app/models/evidence/research/gen_ai/data_slice.rb
brendanshean: I think you're right to point this out.  Initially, this was the design I was going with, but I didn't really like all the nulls.  I feel like the abstraction were modeling here is the relationship between two datasets.

services/QuillLMS/engines/evidence/app/models/evidence/research/gen_ai/dataset.rb
brendanshean: That's correct.  Good eye.

services/QuillLMS/engines/evidence/app/models/evidence/research/gen_ai/data_slice.rb
brendanshean: Actually, on second thought I'm going back to original table.

services/QuillLMS/db/migrate/20240828204236_create_evidence_research_gen_ai_data_slices.evidence.rb
brendanshean: What's unfortunate about this is in Data Science, the term 'dataset' is actually the preferred term of 'data set'.  But, 'data slice' and 'data subset' are preferred over 'dataslice' and 'datasubset'.  Not sure what I'm going to do here.  Also, 'dataset' is the more common form in our codebase already.

PR #12312, opened 2024-08-29
Evidence GenAI - Improvements to RepeatFeedback::Checker

dandrabik:
## WHAT
An attempt to improve the performance of the RepeatFeedback LLM call. There

The change is this:
1. Use Gemini instead of OpenAI
2. Strip HTML and extra spaces from the history list and entry to focus comparison on the content
3. Use an ordered list for history to make it more obvious when multi-line entries start/stop
4. Use temperature of `0` instead of `1` and give the `RepeatChecker` its own independent temperature setting.
5. Edit the text of the prompt

## WHY
The current `RepeatedFeedback::Checker` was a good start but I hadn't spent much time improving it.

**Baseline (what's on develop)**
Differences: 83.5%
Similar: 93.79%
**This code**
Differences: 82.0%
Similar: 100%

So flagging repeats (which is the most important), got stronger, with too often flagging differences only going up slightly (less of a concern).
## HOW
I created a test runner that records the test results and puts them in a csv in the repo. Iterated on temp, prompt, api, model, and code fixes until I was getting higher numbers.

### Notion Card Links
https://www.notion.so/quill/Repeated-Feedback-Performance-024e87c9ad9e45769549066f6bc6e0b4

I ran lots of tests locally using the script. I'm doing some more testing in the browser to see if there are any issues in typical sessions.

PR Checklist | Your Answer
------------ | -------------
Have you deployed to Staging? | Doing some more testing locally then deploying to beta.


services/QuillLMS/engines/evidence/app/services/evidence/gen_ai/repeated_feedback/prompts/2024_08_28_gemini2.md
```diff
@@ -0,0 +1,6 @@
+One at a time, compare the entry to each numbered entry in the list. Is the entry identical to any item in the list? If so, return {'repeat_feedback' : true}. Otherwise, return {'repeat_feedback' : false}
```
dandrabik, line 1: This is the template used. Note, including examples in the prompt gave much worse performance.

services/QuillLMS/engines/evidence/app/services/evidence/gen_ai/repeated_feedback/checker.rb
```diff
@@ -26,8 +28,12 @@ def run
         private def last_attempt? = history.size >= 4
 
         private def system_prompt = Evidence::GenAI::RepeatedFeedback::PromptBuilder.run(prompt: nil, history:)
+        private def entry = HTMLTagRemover.run(feedback.chomp)
+        private def model = chat_api::SMALL_MODEL
```
brendanshean, line 32: Were larger models giving similar accuracy 82%/100%?

services/QuillLMS/engines/evidence/app/services/evidence/gen_ai/repeated_feedback/prompts/2024_08_28_gemini2.md
anathomical: The language does make me wonder how identical "identical" is.  I presume we don't want actual identical comparisons because if we did this could just be a pure Ruby comparison, but I figured I'd drop a note just in case this really should be a Ruby comparison because we mean "identical" literally.

brendanshean: That is surprising

brendanshean (approved):
100% is hard to beat!  Nice work

anathomical (approved):
Sometimes I think that prompts are self-documenting, but then I read one and realize that some of the language is probably being used because of a quirk in how the model interprets them instead of because their literal meaning actually describes the behavior we want.

services/QuillLMS/engines/evidence/app/services/evidence/gen_ai/repeated_feedback/checker.rb
dandrabik: I didn't try with Gemini-pro yet, but early on `gpt-4o` and `gpt-4` performed roughly identically to `gpt-4o-mini` (some times slightly better, some times slightly worse). My hypothesis is that for the super short prompt, the size of the model may not matter that much.

Given the latency concerns, I stuck to the smaller models.

services/QuillLMS/engines/evidence/app/services/evidence/gen_ai/repeated_feedback/prompts/2024_08_28_gemini2.md
dandrabik: We don't want identical, but 'identical' seems to work well for the prompt to do a 'semantic match'. It's definitely inaccurate from a strict definition perspective though.

Long-term, this might be a something that doesn't need an LLM (but for not it's performing well).

PR #12328, opened 2024-09-04
Add manifest.json file for PWA installs

cissyxyu:
## WHAT
In order for our website to be "installable" as a web "app" that can be then opened through the Mac/Windows application dock and phone home screens, we need to add a manifest.json files that designates some properties and icons.

## WHY
This way, a portion of users who like this functionality can "install" Quill to quickly open from their home screens without going through an extra web browser.

## HOW
Add a manifest.json files with required attributes, and serve it from a GET request to a static controller.

### Screenshots
Screenshot 2024-09-04 at 2 45 35 PM
screenshot

### Notion Card Links
https://www.notion.so/quill/Product-Board-30f97e4eb01246dbb8264253fb385073?p=9c9a9512c6e44585907cbdde6c35f6db&pm=c

I've tested it by downloading my own staging site as an app through Mac Desktop, using my iPhone to download it as an app, and asking Thomas to verify downloading on his Android phone.

PR Checklist | Your Answer
------------ | -------------


anathomical (changes requested):
I believe that you may be able to simplify this by simply stashing your `manifest.json` file in the `public/` folder where `robots.txt` is stored.  I think this will allow you to skip the need to define a route since routing should fall through to looking for static files in `public/` if there's no defined route.

I could be wrong, but I'm pretty sure that should work.

cissyxyu:
> I believe that you may be able to simplify this by simply stashing your `manifest.json` file in the `public/` folder where `robots.txt` is stored. I think this will allow you to skip the need to define a route since routing should fall through to looking for static files in `public/` if there's no defined route.
> 
> I could be wrong, but I'm pretty sure that should work.

Yes, that worked! Thanks for the tip, it saves a lot of work.

emilia-friedberg (approved):
So cool to have this working!

anathomical (approved):

PR #12344, opened 2024-09-09
"for administrators" page

emilia-friedberg:
## WHAT
Create new For Administrators page.

## WHY
From Notion: "The purpose of this page is to attract administrators in the “awareness” phase of the sales funnel. When visiting this page, administrators should be able to learn what Quill is and its unique value to teachers and students."

## HOW
Lots of HTML and CSS, some updates so the embedded parts of the page work. I built this in ERB instead of React in keeping with our pattern for other public-facing pages, because it doesn't have any custom interactions.

### Screenshots
Screenshot 2024-09-09 at 1 28 06 PM
Screenshot 2024-09-09 at 1 28 15 PM
Screenshot 2024-09-09 at 1 28 26 PM
Screenshot 2024-09-09 at 1 28 33 PM

### Notion Card Links
https://www.notion.so/quill/Spec-Quill-For-Schools-Districts-Admin-Landing-Page-bda46df222354287842ad5cc279f4408

Viewed on staging on multiple screen sizes, and having Jack and Katie Mcneil review as well.

PR Checklist | Your Answer
------------ | -------------
Have you added and/or updated tests? |  Manually tested -- no tests for ERB


services/QuillLMS/app/views/pages/admins.html.erb
```diff
@@ -1,2 +1,181 @@
+<% for_administrators_img_base = "#{ENV['CDN_URL']}/images/pages/for_administrators" %>
```
happythenewsad, line 1: Lines 1-6 could be in a helper or controller

happythenewsad (approved):
Looks good

services/QuillLMS/app/views/pages/admins.html.erb
emilia-friedberg: They could, but I feel like they're view-specific enough that to me it feels easier to keep them in this file. If you or other feel strongly, though, I can make a change.

eadams17 (approved):
LGTM!

services/QuillLMS/app/views/pages/admins.html.erb
happythenewsad: My impetus for the suggestion is not that these lines aren't view-specific, but that any logic that can be reasonably moved out of views is generally a good thing - easier unit testing, for example. But variable assignment is arguably not logic, so I was on the fence. Thus, I approved the PR and consider this an optional enhancement.

anathomical (approved):
This shouldn't block deploy, but I note that there are two new video files added to the CDN, but they're in different formats (`mov` and `mp4`).  Not a huge deal, but there probably is some value to moving toward a "standard" file type for videos moving forward.

PR #12365, opened 2024-09-12
Pre post data linking

anathomical:
## WHAT
- Add a new model to organize `StudentActivitySequence`s
- Add a new model to connect individual `ActivitySession`s to `StudentActivitySequence`s via `StudentActivitySequenceActivity`s
- Add a service to create necessary records when a student is assigned a new activity
- Add a service to create/update necessary records when a student completes an activity session
- Add workers to handle both of the services
- Hook the workers into existing assignment and completion code paths
## WHY
We want to explicitly record sequences of activities: Pre Diagnostic -> Recommendations -> Post Diagnostic so that we can stop building reports on inference moving forward.
## HOW
- Create migrations and Model definitions for two new tables
- Add a module `StudentActivitySequences` to organize new services and workers
- Create single-path `HandleAssignment` service to build all new records when a student is assigned to a `ClassroomUnit`
- Create a single-path `HandleCompletion` service to update records when a student completes an  `ActivitySession`

### Notion Card Links
https://www.notion.so/quill/RFC-for-Pre-Post-data-storage-ade408cd92724dec9807be31bd3be12b?pvs=4
https://www.notion.so/quill/Implement-Pre-Post-data-storage-43aa79b3fee94e28a6ac0003002ca9c1?pvs=4

- Assign a new pre diagnostic to a classroom
- - Confirm that `StudentActivitySequence` and `StudentActivitySequenceActivity` records are created for assigned students
- - Unassign student, confirm that records remain
- - Re-assign student, confirm that no new records are created (we want to re-use the already-existing ones)
- Assign a second copy of the same pre-diagnostic to the same students
- - Confirm that new records are created for the new sequence
- Complete the first assigned pre-diagnostic as a student, confirm that the `StudentActivitySequenceActivity` record gets updated upon activity completion
- Assign recommendations as a teacher
- - Confirm that the `StudentActivitySequenceActivity` records for each recommended activity assigned gets created
- - Confirm that, in the case of multiple pre-diagnostic assignments, that the recommendations are attached to the most recently-created `StudentActivitySequence`
- Complete a recommendation as a student
- - Confirm that the `StudentActivitySequenceActivity` record for the completed activity is updated with a `completed_at` and `activity_session_id` value
- Assign the post-diagnostic as a teacher
- - Confirm that a new `StudentActivitySequenceActivity` record is created for the new assignment
- Complete the post-diagnostic as a student
- - Confirm that the `StudentActivitySequenceActivity` recor for the post-diagnostic gets updated

PR Checklist | Your Answer
------------ | -------------


services/QuillLMS/app/controllers/api/v1/activity_sessions_controller.rb
```diff
@@ -42,6 +43,7 @@ def create
     @activity_session.user = current_user if current_user
 
     if @activity_session.save
+      handle_student_activity_sequence_activity_completion if @activity_session.completed_at
```
brendanshean: Since there's a guard clause for this on line 85 I would remove the `if @activity_session.completed_at` to keep consistent with line 22.

services/QuillLMS/spec/factories/student_activity_sequence.rb
```diff
@@ -0,0 +1,10 @@
+# frozen_string_literal: true
+
+FactoryBot.define do
```
brendanshean: Factories that have the same name as the association can use the shorthand:

```
factory :student_activity_sequence do
    classroom
    association :initial_activity, factory: :pre_diagnostic_activity
    association :initial_classroom_unit, factory: :classroom_unit
    user
  end
```

services/QuillLMS/spec/factories/student_activity_sequence_activity.rb
```diff
@@ -0,0 +1,9 @@
+# frozen_string_literal: true
+
+FactoryBot.define do
+  factory :student_activity_sequence_activity do
```
brendanshean: You can use the shorthand on all three of these
```
factory :student_activity_sequence_activity do
    activity
    classroom_unit
    student_activity_sequence
  end
```

services/QuillLMS/app/services/student_activity_sequences/find_start_activity.rb
```diff
@@ -0,0 +1,41 @@
+# frozen_string_literal: true
+
+module StudentActivitySequences
+  class FindStartActivity < ApplicationService
```
brendanshean: Is StartActivity the same as initial_activity in `StudentActivitySequence`?  If so, for consistency I would rename this to `FindInitialActivity`

services/QuillLMS/spec/services/student_activity_sequences/find_start_activity_spec.rb
```diff
@@ -0,0 +1,60 @@
...
+    end
+
+    context 'activity is post_diagnostic' do
+      let(:activity) { create(:diagnostic_activity) }
+      let!(:pre_diagnostic) { create(:pre_diagnostic_activity, follow_up_activity_id: activity.id) }
+
+      it { expect(subject).to eq(pre_diagnostic) }
+    end
+
+    context 'activity is part of a recommendation' do
```
brendanshean: Nice specs here.  They cleared up the questions I had on `for_recommended_activity`

services/QuillLMS/app/models/student_activity_sequence.rb
```diff
@@ -0,0 +1,27 @@
...
+#  id                        :bigint           not null, primary key
+#  created_at                :datetime         not null
+#  updated_at                :datetime         not null
+#  classroom_id              :integer          not null
+#  initial_activity_id       :integer          not null
+#  initial_classroom_unit_id :integer          not null
+#  user_id                   :integer          not null
+#
+class StudentActivitySequence < ApplicationRecord
+  belongs_to :classroom
```
brendanshean: The [RFC](https://www.notion.so/quill/RFC-for-Pre-Post-data-storage-ade408cd92724dec9807be31bd3be12b?pvs=4#7a08786044864950a246d72fdac7f2f0) listed
`student_id`
`classroom_id` 
`initial_activity_id`

But now there is a `initial_classroom_unit_id`.  A couple of questions,
1) Do we need classroom if we now have classroom_unit?
2) Why is this called `initial_classroom_unit`?  Could there be a change in the course of the student activity sequence?  (If not, I would probably just call it `classroom_unit`)

anathomical: Fair questions.  I added the new value when I got into testing the weird edge case where a teacher can assign the same diagnostic to their class more than once.  Differentiating simply on activity and classroom resulted in ambiguity there.  This does make `classroom_id` strictly unnecessary, so I'm open to arguments that it should be cleaned up, but it felt like the de-normalization in this case would likely actually be useful for reporting purposes (as I expect us to often want to filter on this value).

The `initial` naming did feel a bit odd to me, but I wanted to ensure that future engineers didn't believe that the value on this record should necessarily match the record on the associated `StudentActivitySequenceActivity` record.  It's possible that this is more confusing, and i'm open to a name change on that as well.

brendanshean: Thank you for that explanation.  I misunderstood the data model here, so I think `initial_classroom_unit` makes sense.

As far as `classroom_id` I'll be curious to hear what others think but I think your de-normalization argument for reporting seems reasonable.

brendanshean (approved):
Great work on this.  The service objects and their corresponding specs made this very clear.

happythenewsad (approved):
LGTM!

dandrabik (approved):
Overall this looks good. My only comment, which we discussed, is the names are a little confusing, particularly `student_activity_sequence_activity`. Let's find a better name, and 🚢

PR #12369, opened 2024-09-13
Add two new Queue Latency endpoints for NewRelic Pings 

cissyxyu:
## WHAT
Add two new endpoints to check for queue latency exceeding the limit, for both the critical_external queue and the low queue. Remove the old endpoint.

## WHY
We need to convert some NewRelic alerts to pings, so we need the data from the Sidekiq Queue Latency check endpoints to be more specific -- one endpoint should return 400 only if the critical external queue is too slow, and the other endpoint should return 400 only if the low priority queue is too slow.

## HOW
Split up the existing endpoint into two endpoints, one checking for the critical external queue only and the other checking for the low priority queue only.

### Screenshots

### Notion Card Links
https://www.notion.so/quill/Convert-New-Relic-Synthetic-Scripts-to-Pings-71240e4e89cf41729bf137335bc01b06?pvs=4

Deploy to staging and check hitting both endpoints at the default state with nothing in either queue (should return 200 on both endpoints). Then set the latency limits artificially low via environment variables on staging and verify that both endpoints return a 400.

PR Checklist | Your Answer
------------ | -------------

### DIFF
--- a/services/QuillLMS/app/controllers/statuses_controller.rb
+++ b/services/QuillLMS/app/controllers/statuses_controller.rb
@@ -34,7 +34,6 @@ class StatusesController < ApplicationController
     render plain: 'OK'
   end
 
-
   def sidekiq_critical_external_latency
     queues = Sidekiq::Queue.all
     latency_hash = queues.to_h { |q| [q.name, q.latency] }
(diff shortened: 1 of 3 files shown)


brendanshean (approved):
Thank you for splitting this monitor up.

anathomical (approved):

happythenewsad (approved):

PR #12371, opened 2024-09-13
Evidence GenAI - Gemini for Secondary Feedback and basic similarity test.

dandrabik:
## WHAT
Add a `cosine_similarity` test for secondary feedback, update the prompt, and port to Gemini. There's iteration and testing to be done here, but prompt/example count/api is meeting / mildly beating the OpenAI prompt.
## WHY
We've found Gemini Flash to be much faster, porting this mainly for the speed bump.
## HOW
Wrote a test harness in `thor`. Iterated on setups and prompts.

I've run test scripts to hit this code a lot. This is 'OFF' in production so not much to worry about for deploy. Functionality-wise is a relatively small change.

PR Checklist | Your Answer
------------ | -------------
Have you deployed to Staging? | Not yet - deploying to Beta now.


anathomical (approved):

services/QuillLMS/app/lib/core_extensions/array.rb
```diff
@@ -17,5 +17,26 @@ def median
...
       (dup[mid] + dup[mid - 1]) / 2.0
     end
+
+    def average
+      return 0 if empty?
+
+      sum.to_f / size
+    end
+
+    def dot_product(other_array) = zip(other_array).map { |a, b| a * b }.sum
```
brendanshean: What about error handling for arrays are of unequal length:
```
def dot_product(other_array)
  return 0 if size != other_array.size

  zip(other_array).map { |a, b| a * b }.sum
end
```

brendanshean (approved):

services/QuillLMS/app/lib/core_extensions/array.rb
dandrabik: That's a good idea, I'll add that to these methods.

PR #12383, opened 2024-09-17
allow-translation-updates

anathomical:
## WHAT
Update translatable concern to translate changes to source text
## WHY
Without this we were basically bailing early on a translation if the object had ever been translated, even if what we're trying to translate is a change in its original text
## HOW
Instead of just exiting early from the translation process if the underlying `field_name` has already been translated, check to see if it's been translated for the existing source text.  If it hasn't, delete the old (now no longer relevant) `translation_mapping` and re-translate the object.

### Notion Card Links
https://www.notion.so/quill/Investigate-whether-the-existing-Rake-task-will-handle-updating-translations-after-re-grading-4d50020dc9b94af2aa1ebe33987c673a?pvs=4
https://www.notion.so/quill/Translations-changes-6dc4b121ceb94109bcd13c07c913ad96

- Deploy to staging
- Examine an existing question's translations in the API
- Update the feedback for that question
- Re-run the translation rake task in staging
- Manually clear staging cache (these API calls are cached)
- Confirm that the question's translations in the API are updated

PR Checklist | Your Answer
------------ | -------------

### DIFF
--- a/services/QuillLMS/app/models/concerns/translatable.rb
+++ b/services/QuillLMS/app/models/concerns/translatable.rb
@@ -67,7 +67,7 @@ module Translatable
     return unless translatable_text.is_a?(String) && translatable_text.present?
     return unless translation_mappings.joins(:english_text)
       .where(field_name:)
-      .where(english_text: {text: translatable_text})
+      .where(english_text: { text: translatable_text })
       .empty?
 
     clean_deprecated_translations(translatable_text:, field_name:)
@@ -131,7 +131,7 @@ module Translatable
   private def clean_deprecated_translations(translatable_text:, field_name:)
     translation_mappings.joins(:english_text)
       .where(field_name:)
-      .where.not(english_text: {text: translatable_text})
+      .where.not(english_text: { text: translatable_text })
       .destroy_all
   end
 end
--- a/services/QuillLMS/app/models/concerns/translatable.rb
+++ b/services/QuillLMS/app/models/concerns/translatable.rb
@@ -65,7 +65,12 @@ module Translatable
 
   def create_translation_mappings_with_text(translatable_text:, field_name: default_translatable_field)
     return unless translatable_text.is_a?(String) && translatable_text.present?
-    return unless translation_mappings.where(field_name:).empty?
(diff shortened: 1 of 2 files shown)


eadams17 (approved):

services/QuillLMS/app/models/concerns/translatable.rb
```diff
@@ -65,7 +65,12 @@ def create_translation_mappings
 
   def create_translation_mappings_with_text(translatable_text:, field_name: default_translatable_field)
     return unless translatable_text.is_a?(String) && translatable_text.present?
-    return unless translation_mappings.where(field_name:).empty?
+    return unless translation_mappings.joins(:english_text)
```
dandrabik, line 68: Let's remove this double negative here (`unless finder_code.empty?`, and do `return if finder_code.exists?`

happythenewsad (approved):

services/QuillLMS/app/models/concerns/translatable.rb
```diff
@@ -65,7 +65,12 @@ def create_translation_mappings
...
+      .where(field_name:)
+      .where(english_text: { text: translatable_text })
+      .empty?
+
+    clean_deprecated_translations(translatable_text:, field_name:)
```
dandrabik, line 73: I think using `destroy_` instead of `clean_` in the name is more accurate.

services/QuillLMS/app/models/concerns/translatable.rb
```diff
@@ -122,4 +127,11 @@ def open_ai_prompt(locale:)
   private def translatable_text
     data[default_translatable_field]
   end
+
+  private def clean_deprecated_translations(translatable_text:, field_name:)
+    translation_mappings.joins(:english_text)
+      .where(field_name:)
+      .where.not(english_text: { text: translatable_text })
+      .destroy_all
```
dandrabik, line 135: Take or leave, it is a little surprising to see methods called `create_` and `create_` and then see a `destroy` side effect. I wonder if it would be better to move this up in the chain.

dandrabik (commented):
Suggesting a few changes to make this clearer.

PR #12386, opened 2024-09-18
Restrict Vitally This Year calculations to live activities

cissyxyu:
## WHAT
Curriculum team asked if we could only calculate currently live diagnostic activities in the statistics for this year. So, instead of considering all pre-diagnostic and post-diagnostic activities viable in the count, we need to restrict calculations for this year to only live, unarchived pre-diagnostic and post-diagnostic activities using the constant `Activity::PRE_TEST_DIAGNOSTIC_IDS` which is always updated to only contain live activities.

## WHY
They want the number here to match other numbers they're using in their metrics, so they don't want archived activities in the mix. We do however need to keep the functions for archived activities for calculating previous year metrics.

## HOW
Add some additional functions to pull from live diagnostics only.

### Screenshots

### Notion Card Links
https://www.notion.so/quill/Update-Vitally-Data-Sync-New-Diagnostic-Usage-Data-2fb92d4ba89847af9da661fc78710556?pvs=4

Deploy to staging and test it on a few schools to ensure that the numbers match up to expected counts.

PR Checklist | Your Answer
------------ | -------------


happythenewsad (approved):

services/QuillLMS/app/models/concerns/vitally_shared_functions.rb
```diff
@@ -6,7 +6,9 @@ module VitallySharedFunctions
   attr_reader :vitally_entity, :school_year_start, :school_year_end
 
   PRE_DIAGNOSTIC_IDS = Activity.where(activity_classification_id: ActivityClassification.diagnostic&.id).where.not(follow_up_activity: nil).pluck(:id)
+  LIVE_PRE_DIAGNOSTIC_IDS = Activity.where(id: Activity::PRE_TEST_DIAGNOSTIC_IDS).pluck(:id)
```
anathomical: This isn't a huge deal, but doesn't this literally just go to the database to re-produce `Activity::PRE_TEST_DIAGNOSTIC_IDS`?  Is there a need to do a query and `pluck(:id)`?  Can we just use the constant here?

anathomical (approved):
One minor take-or-leave comment about querying for constants we already have, but this looks like a reasonable solution to the issue to me.

PR #12392, opened 2024-09-18
stop-error-on-completing-unassigned-student-learning-sequences

anathomical:
## WHAT
Early return instead of raising an error
## WHY
Not every ActivitySession is expected to be associated with a `StudentLearningSequenceActivity` record, so we want to make sure that there's an early return path for cases where that happens instead of erroring
## HOW
Instead of raising an error when `HandleCompletion` can't find a record to update, just return early

- Deploy to staging
- Find a non-recommendation Connect ActivitySession
- Pass that ActivitySession id to `StudentLearningSequences::HandleCompletion.run`
- Confirm that there's an early return with no error

PR Checklist | Your Answer
------------ | -------------


eadams17 (commented):
looks good to me!

PR #12406, opened 2024-09-24
Fix bug in Vitally stats where classrooms with multiple teachers are double counted

cissyxyu:
## WHAT
For schools and districts, we currently have a bug in the Vitally data aggregation. We're counting data twice when we have a classroom with multiple teachers because each teacher connected to the classroom gets aggregated into the query. To remedy this, we need to add a WHERE clause to filter for only classrooms where the teacher is the owner.

I also found this same bug in the Schools.students function, which I also fixed.

## WHY
We're currently double-counting some classrooms, which causes the Vitally numbers overall to be inflated for all schools and districts when it comes to activities and active students.

## HOW
Filter for classrooms where the teacher is the owner only.

### Screenshots

### Notion Card Links
https://www.notion.so/quill/Audit-Vitally-sync-calculations-to-make-sure-we-re-not-multi-counting-records-when-joining-on-classr-59019fb038a042a7b472c75dbfc564b1

Deployed to staging, I also tested the queries generated by the new code in Metabase to ensure that the numbers returned by the new query match up to the number in the Usage Snapshots reports.

PR Checklist | Your Answer
------------ | -------------

### DIFF
--- a/services/QuillLMS/app/models/school.rb
+++ b/services/QuillLMS/app/models/school.rb
@@ -190,7 +190,7 @@ class School < ApplicationRecord
   end
 
   def students
-    User.joins(student_in_classroom: { teachers: :school }).where(schools: { id: id }).distinct
+    User.joins(student_in_classroom: { teachers: :school }).where(schools: { id: id }).where(classrooms_teachers: {role: ClassroomsTeacher::ROLE_TYPES[:owner]}).distinct
   end
 
   def detach_from_existing_district_admins(district)
--- a/services/QuillLMS/app/models/school.rb
+++ b/services/QuillLMS/app/models/school.rb
@@ -190,7 +190,10 @@ class School < ApplicationRecord
   end
 
   def students
-    User.joins(student_in_classroom: { teachers: :school }).where(schools: { id: id }).where(classrooms_teachers: { role: ClassroomsTeacher::ROLE_TYPES[:owner] }).distinct
+    User.joins(student_in_classroom: { teachers: :school })
+      .where(schools: { id: id })
+      .where(classrooms_teachers: { role: ClassroomsTeacher::ROLE_TYPES[:owner] })
+      .distinct
   end
 
   def detach_from_existing_district_admins(district)
--- a/services/QuillLMS/app/models/school.rb
+++ b/services/QuillLMS/app/models/school.rb
@@ -190,7 +190,7 @@ class School < ApplicationRecord
   end
 
   def students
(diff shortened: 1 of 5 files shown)


happythenewsad (approved):

services/QuillLMS/app/models/school.rb
```diff
@@ -190,7 +190,7 @@ def generate_leap_csv(activities_since = Date.parse('2010-01-01'))
   end
 
   def students
-    User.joins(student_in_classroom: { teachers: :school }).where(schools: { id: id }).distinct
+    User.joins(student_in_classroom: { teachers: :school }).where(schools: { id: id }).where(classrooms_teachers: { role: ClassroomsTeacher::ROLE_TYPES[:owner] }).distinct
```
anathomical: This has gotten long enough that it might be worth breaking up over multiple lines the way we do in `vitally_school_stats.rb`

services/QuillLMS/spec/services/vitally_integration/serialize_vitally_sales_account_spec.rb
```diff
@@ -276,6 +276,24 @@
 
       it { expect(results[:traits][:active_students]).to eq(2) }
     end
+
+    context 'when teachers are merely coteachers and not owners of the classroom' do
+      it 'does not double count activities finished and active students' do
+        classroom_teachers[0].update(role: 'coteacher')
+        classroom_teachers[1].update(role: 'coteacher')
```
anathomical: Take or leave this, but I'd probably do the coteacher change in a `let` call inside the context:
```
let!(:classroom_teachers) { create_list(:classrooms_teacher, 2, classroom:, role: 'coteacher')

it 'does not double count activities finished and active students' do
  ...
```
Just to keep the `it` block super simple.

services/QuillLMS/spec/services/vitally_integration/serialize_vitally_sales_organization_spec.rb
```diff
@@ -404,6 +404,24 @@
           expect(described_class.new(district).data[:traits][:active_students_all_time]).to eq(2)
         end
       end
+
+      context 'excludes classrooms where teacher is coteacher' do
```
anathomical, line 408: Again, take or leave this, but I'd just override the `let!` call instead of using a `before` block:
```
let!(:classroom_teacher1) { create(:classrooms_teacher, user: teacher2, classroom: classroom1, role: 'coteacher') }
```

anathomical (approved):
A couple of minor take-or-leave style approaches, but this looks good to me.

PR #12409, opened 2024-09-24
Student learning sequence enhancements

anathomical:
## WHAT
- Add `unscoped` to lookups of `ActivitySession`
- Tweak the callback that enqueues `HandleAssignmentWorker` jobs to only enqueue newly-assigned student_ids rather than all of them
- Add indexes to `student_learning_sequences` and `student_learning_sequence_activities`
## WHY
- Apparently `ActivitySession` records are occasionally created with `visible: false` which, in the default scope, can't be found by follow-up workers
- This should cut down on unnecessary job runs
- Speed up look-ups during assignment and completion workflows
## HOW
- Add `unscoped` to `ActivitySession.find`
- Compare the old array of `assigned_student_ids` to the new one, and only enqueue workers for new values in the array
- Create multi-column indexes for the two queries that are used to find existing records in the assignment and completion services

- Assign new activity pack in staging
- Confirm new `StudentLearningSequence` and `StudentLearningSequenceActivity` records are created on assignment
- Complete activity as student
- Confirm that `StudentLearningSequenceActivity` is updated as expected

PR Checklist | Your Answer
------------ | -------------

### DIFF
--- a/services/QuillLMS/app/models/classroom_unit.rb
+++ b/services/QuillLMS/app/models/classroom_unit.rb
@@ -151,7 +151,9 @@ class ClassroomUnit < ApplicationRecord
   end
 
   private def assign_student_learning_sequences
-    assigned_student_ids.each do |student_id|
+    old_ids, new_ids = saved_change_to_assigned_student_ids
+    newly_assigned = new_ids - old_ids
+    newly_assigned.each do |student_id|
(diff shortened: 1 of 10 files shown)


brendanshean (approved):
Thanks for making these improvements.

PR #12422, opened 2024-09-25
Add LabeledEntry model

brendanshean:
## WHAT
Create a `LabeledEntry` model.

## WHY
Currently, embeddings are used in`PromptResponse` which connect to `PromptResponseFeedback` models.  For purposes of classification models and RAG it makes more sense to have this object as one record, so `LabeledEntry` is a combination of these two.

## HOW
Write a migration that combines the two models.  Rename `feedback` to `label` since this particular task concerns classification.  Also, `response` becomes `entry` which is consistent with `FeedbackHistory`

NB.  I will add indexing to the embedding vector later following neighbor gem [advice](https://github.com/pgvector/pgvector?tab=readme-ov-file#index-build-time)
> Like other index types, it’s faster to create an index after loading your initial data

Also, while `neighbor` gem already exists in the engine, not having it in the main app results in weird loading errors locally as well as the warning: unknown OID 386956: failed to recognize type of 'embedding'. It will be treated as String. when running migrations.

### Screenshots

### Notion Card Links

This was tested on staging

PR Checklist | Your Answer
------------ | -------------


services/QuillLMS/Gemfile
brendanshean, line 1: While neighbor already exists in the engine, not having it in the main app results in weird loading errors locally as well as the warning: `unknown OID 386956: failed to recognize type of 'embedding'. It will be treated as String.` when running migrations.

services/QuillLMS/db/migrate/20240925185730_create_prompt_response_labels.evidence.rb
brendanshean, line 1: pgvector [suggests](https://github.com/pgvector/pgvector?tab=readme-ov-file#index-build-time)

> Like other index types, it’s faster to create an index after loading your initial data

so I will add an index to the embedding field in a later PR

services/QuillLMS/engines/evidence/app/models/evidence/labeled_entry.rb
```diff
@@ -0,0 +1,60 @@
...
+    MODEL = 'text-embedding-3-small'
+
+    DISTANCE_METRIC = 'cosine'
+
+    belongs_to :prompt
+
+    has_neighbors :embedding
+
+    validates :embedding, presence: true
+    validates :label, presence: true
```
dandrabik, line 40: This could come later, but we likely do want the `label_transformed` in the `metadata` field.

services/QuillLMS/engines/evidence/app/models/evidence/labeled_entry.rb
```diff
@@ -0,0 +1,60 @@
...
+
+# == Schema Information
+#
+# Table name: evidence_labeled_entries
+#
+#  id         :bigint           not null, primary key
+#  embedding  :vector(1536)     not null
+#  entry      :text             not null
+#  label      :text             not null
+#  metadata   :jsonb
```
dandrabik: So, I think we'll also want another nullable boolean field in here called `approved` or something like that. 

Basically, it would be nice to have a place to store the entry/embedding from the incoming student responses (so we only need to fetch from the API once), but only use the embeddings that have been flagged as `approved` (as whatever name) for the nearest neighbor analysis. Thinking a nullable boolean since 'null' would be "unlooked at", 'false' would "this is incorrect, ignore it for now". Happy to discuss, especially if you have different ideas of what to do with incoming data (I don' t have strong opinions here).

services/QuillLMS/engines/evidence/app/models/evidence/labeled_entry.rb
```diff
@@ -0,0 +1,60 @@
...
+#  created_at :datetime         not null
+#  updated_at :datetime         not null
+#  prompt_id  :integer          not null
+#
+# Indexes
```
dandrabik, line 18: I believe we want a unique index constraint on [prompt_id, entry].

dandrabik (approved):
Overall this looks good, nice work on this. I have a few requested changes that I'd like either made or we should discuss to find a solution. Marking as 'approved' since those changes don't really need to be re-reviewed after they are made.

PR #12443, opened 2024-10-02
Stop opening links in browser when user is using PWA App

cissyxyu:
## WHAT
When the user has installed Quill.org as a PWA app opened separately in a new window, we don't want any links to open in a target="_blank" style in a new tab because that will redirect them to their normal browser. So we want to remove the target=_blank directive if the user is inside a PWA app.

## WHY
So the user has a seamless experience of using the standalone app and all navigation is contained in the app only.

## HOW
We can add a script after page load that detects if the user is in a standalone PWA app. If they are, we can remove target="_blank" from all links.

### Screenshots

### Notion Card Links
https://www.notion.so/quill/PWA-Manifest-Links-open-within-da22b29e5aa048499b90afe5ac6d13eb?pvs=4

Deploy to staging and verify that links in the app open in the app only.

PR Checklist | Your Answer
------------ | -------------
Have you added and/or updated tests? |  No, manually tested

### DIFF
--- a/services/QuillLMS/app/views/layouts/activity.html.erb
+++ b/services/QuillLMS/app/views/layouts/activity.html.erb
@@ -32,3 +32,5 @@
   </body>
 
 </html>
+
+<%= render partial: 'pages/shared/pwa_app_script' %>
--- a/services/QuillLMS/app/views/layouts/application.html.erb
+++ b/services/QuillLMS/app/views/layouts/application.html.erb
@@ -53,3 +53,5 @@
   </body>
 
 </html>
+
+<%= render partial: 'pages/shared/pwa_app_script' %>
--- a/services/QuillLMS/app/views/layouts/home.html.erb
+++ b/services/QuillLMS/app/views/layouts/home.html.erb
@@ -56,3 +56,5 @@
     <%= yield %>
   </body>
 </html>
+
+<%= render partial: 'pages/shared/pwa_app_script' %>
--- a/services/QuillLMS/app/views/layouts/integrations.html.erb
+++ b/services/QuillLMS/app/views/layouts/integrations.html.erb
@@ -20,3 +20,5 @@
   </body>
 
 </html>
+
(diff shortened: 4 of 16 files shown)


eadams17 (approved):
This generally looks good to me. Just a question though-- what about pages that don't use the footer? (i.e. during assignment flow, https://www.quill.org/assign/diagnostic)

emilia-friedberg (commented):
I agree with Eric -- it might be better to store this as a partial and include it in the layout files, rather than the footer.

anathomical:
> I agree with Eric -- it might be better to store this as a partial and include it in the layout files, rather than the footer.

I don't believe we have a global "include this script partial in all views" partial, but this seems like a good use-case for one.  Should we consider introducing one?  If so, are there other scripts that it would make sense to include there?

cissyxyu:
> I agree with Eric -- it might be better to store this as a partial and include it in the layout files, rather than the footer.

Just made it into a partial view that I included in all the pages I could think of -- I'm not sure if I might have missed any though, let me know if you think of any edge pages that might also need it!

emilia-friedberg:
@cissyxyu I think if you just include it in every file under `views/layouts`, that should be exhaustive, because those are the highest-level view files.

emilia-friedberg:
> > I agree with Eric -- it might be better to store this as a partial and include it in the layout files, rather than the footer.
> 
> Just made it into a partial view that I included in all the pages I could think of -- I'm not sure if I might have missed any though, let me know if you think of any edge pages that might also need it!

@cissyxyu I think if you just include it in every file under views/layouts, that should be exhaustive, because those are the highest-level view files.

emilia-friedberg (commented):
Just linking to this comment as a review for better visibility: https://github.com/empirical-org/Empirical-Core/pull/12443#issuecomment-2429540500

emilia-friedberg (approved):

PR #12444, opened 2024-10-02
fix-translated-question-instructions

anathomical:
## WHAT
Filter by field_name when sourcing the base translation value
## WHY
When originally written, this code assumed that a single model instance could only have a single translation associated with it.  This is not the case with questions, which was resulting in an arbitrary associated translation being pulled.  Adding this logic should consistently pull the intended translation.
## HOW
When fetching the `translation` for a translatable model, join to the `translation_mappings` table where we store the `field_name` value.  This value signals which column from the translatable model the translation is for, and each translatable model has a default field name that its `translation` call is expected to fetch.  Add a `where` clause to filter for the default `field_name` when seeking translations so that the default field is always pulled while we rely on supplemental code to fetch translations for non-default fields (see `TranslatableQuestion` concern).

### Notion Card Links
https://www.notion.so/quill/Students-seeing-code-in-Spanish-translations-of-instructions-in-ELL-activities-112d42e6f941803681d3f039ddabfbc0?pvs=4

- Using old code, fetch a `Question` object that is known to provide the incorrect value for `translation`
- Confirm that it consistently does return the incorrect value
- Using the new code, fetch the same object, confirm that its `translation` value is correct

PR Checklist | Your Answer
------------ | -------------

### DIFF
--- a/services/QuillLMS/app/models/concerns/translatable.rb
+++ b/services/QuillLMS/app/models/concerns/translatable.rb
@@ -77,11 +77,13 @@ module Translatable
   end
 
   def translation(locale: DEFAULT_LOCALE, source_api: OPEN_AI_SOURCE)
-    @translation ||= translations(locale: locale, source_api: source_api)&.first&.translation
+    @translation ||= translations(locale: locale, source_api: source_api, field_name: default_translatable_field)&.first&.translation
   end
 
-  def translations(locale:, source_api:)
-    translated_texts.where(locale: locale).ordered_by_source_api(source_api)
+  def translations(locale:, source_api:, field_name:)
+    translated_texts.joins(:translation_mappings)
+      .where(translation_mappings: { field_name: })
+      .where(locale: locale).ordered_by_source_api(source_api)
   end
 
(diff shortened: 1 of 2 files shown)


happythenewsad (approved):

services/QuillLMS/app/models/concerns/translatable.rb
```diff
@@ -77,11 +77,13 @@ def create_translation_mappings_with_text(translatable_text:, field_name: defaul
   end
 
   def translation(locale: DEFAULT_LOCALE, source_api: OPEN_AI_SOURCE)
-    @translation ||= translations(locale: locale, source_api: source_api)&.first&.translation
+    @translation ||= translations(locale: locale, source_api: source_api, field_name: default_translatable_field)&.first&.translation
```
brendanshean: It feels a little redundant to have the default set in two places here.  I'd probably just have it in the method definition).

brendanshean (approved):
Thanks for fixing this!

eadams17 (approved):
Thanks for tracking this down!

PR #12446, opened 2024-10-02
add ai type migration

emilia-friedberg:
## WHAT
Add an `ai_type` migration to the Evidence::Activity model and validation for which `ai_type`s are allowed.

## WHY
We are going to use this field to differentiate on the frontend.

## HOW
Just add the migration, validations, and requisite tests.

### Screenshots

### Notion Card Links
https://www.notion.so/quill/Backend-for-AI-Type-update-113d42e6f94180228db9dd9c49accc6a?pvs=4

Ran the migration on staging, then updated all the existing `Evidence::Activity`s in the console to have the `AutoML` type. 

PR Checklist | Your Answer
------------ | -------------


services/QuillLMS/engines/evidence/app/models/evidence/activity.rb
```diff
@@ -52,6 +53,12 @@ class Activity < ApplicationRecord
     DEFAULT_REPEAT_STEM_RULE_NAME = 'Repeating the stem'
     DEFAULT_REPEAT_STEM_RULE_CONCEPT = 'N5VXCdTAs91gP46gATuvPQ'
 
+    AI_TYPES = [
```
brendanshean, line 56: For more consistency with other parts of the codebase, can you update these to:
``` 
AI_TYPES = [
  AUTO_ML = 'AutoML',
  GEN_AI = 'GenAI', 
  RAG = 'RAG'
]
```

brendanshean (approved):
One minor request, but other than that, looks good!

happythenewsad (approved):

services/QuillLMS/engines/evidence/spec/models/evidence/activity_spec.rb
```diff
@@ -40,6 +45,24 @@ module Evidence
...
+        end
+
+        it 'null should be invalid' do
+          expect { activity.update!(ai_type: nil) }.to raise_error(ActiveRecord::RecordInvalid)
+        end
+
+        it 'any other string should be invalid' do
+          expect { activity.update!(ai_type: 'something else') }.to raise_error(ActiveRecord::RecordInvalid)
+        end
+      end
```
anathomical: I think that you can actually capture these cases with:
```
it { should validate_presence_of(:ai_type) } # this should handle the not-null case
it { should validate_inclusion_of(:ai_type).in?(Activity::AI_TYPES) } # this should handle the valid types and "other strings" cases
```

anathomical (approved):
A note on possible ways to simplify the specs, but this looks good to me.

services/QuillLMS/engines/evidence/spec/models/evidence/activity_spec.rb
emilia-friedberg: So weirdly, `validate_presence_of` isn't testing true here -- I think just because that validation doesn't literally exist in this case. I'm going to take your advice and just use the second line, though, since the way the tests are written now they're basically testing the internals of `validate_inclusion_of`.

PR #12457, opened 2024-10-15
handle invalid plagiarism texts

emilia-friedberg:
## WHAT
Handle invalid plagiarism texts similarly to how we handle invalid highlights.

## WHY
This was a request from the curriculum team.

## HOW
Update the backend to identify bad texts, and the frontend to display them. Also move some duplicated string formatting rules into a helper. Note: there's probably more deduplication we could do here, but since I'm going to be adding similar (but not identical) functionality for "relevant texts" as part of the AI project, I'm punting on that for now.

### Screenshots
Screenshot 2024-10-15 at 3 20 59 PM
Screenshot 2024-10-15 at 3 21 06 PM
Screenshot 2024-10-15 at 3 21 21 PM

### Notion Card Links
https://www.notion.so/quill/The-CMS-is-not-showing-an-error-message-when-the-plagiarism-text-doesn-t-match-the-associated-text-i-3519ee8c595646cdb8accdf179a184a1?pvs=4

Tested that a) the plagiarism check still functions, since I made a small change to that function, and that b) bad texts get highlighted on the activity settings page, the plagiarism rules index page, and the plagiarism rules show page.

PR Checklist | Your Answer
------------ | -------------

### DIFF
--- a/services/QuillLMS/client/app/bundles/Staff/components/evidence/configureRules/rulePlagiarismAttributes.tsx
+++ b/services/QuillLMS/client/app/bundles/Staff/components/evidence/configureRules/rulePlagiarismAttributes.tsx
@@ -97,12 +97,15 @@ const RulePlagiarismAttributes = ({
     if (plagiarismText._destroy) { return <span /> }
 
     return (
-      <PlagiarismTextEditor
-        index={i}
-        key={i}
-        setPlagiarismText={onHandleSetPlagiarismText}
-        text={plagiarismText.text}
-      />
+      <React.Fragment>
+        <PlagiarismTextEditor
+          index={i}
+          key={i}
+          setPlagiarismText={onHandleSetPlagiarismText}
+          text={plagiarismText.text}
+        />
+        {!plagiarismText.valid_in_all_targets && <p className="all-errors-message">This plagiarism text does not match the associated activity text. This means that it will not trigger the plagiarism feedback as intended. Please update the text above.</p>}
+      </React.Fragment>
     )
   })
 
--- a/services/QuillLMS/client/app/bundles/Staff/components/evidence/configureRules/rulePlagiarismAttributes.tsx
+++ b/services/QuillLMS/client/app/bundles/Staff/components/evidence/configureRules/rulePlagiarismAttributes.tsx
@@ -104,7 +104,7 @@ const RulePlagiarismAttributes = ({
           setPlagiarismText={onHandleSetPlagiarismText}
           text={plagiarismText.text}
         />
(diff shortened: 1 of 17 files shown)


services/QuillLMS/client/app/bundles/Staff/components/evidence/configureRules/rulePlagiarismAttributes.tsx
```diff
@@ -97,12 +97,15 @@ const RulePlagiarismAttributes = ({
...
-        text={plagiarismText.text}
-      />
+      <React.Fragment>
+        <PlagiarismTextEditor
+          index={i}
+          key={i}
+          setPlagiarismText={onHandleSetPlagiarismText}
+          text={plagiarismText.text}
+        />
+        {!plagiarismText.valid_in_all_targets && <p className="all-errors-message">This plagiarism text does not match the associated activity text. This means that it will not trigger the plagiarism feedback as intended. Please update the text above.</p>}
```
anathomical: One subtle difference between bad highlights and bad plagiarism text is that a badly defined highlight won't display at all, but a badly defined plagiarism text will still catch a lot of plagiarism as long as it's close.  I don't think that means we need to change the language here or anything, but it may be worth making sure that Curriculum understands that while highlight text that is close-but-not exact doesn't work at all, they may see that plagiarism mostly gets caught even with out-of-date texts.

We still want to flag this so that they can update them, especially in cases where the plagiarism text changes significantly, so I don't think I'm suggesting code changes, but we do want Curriculum to understand the distinction.

services/QuillLMS/engines/evidence/app/models/evidence/activity.rb
```diff
@@ -135,7 +136,7 @@ def flag=(flag)
     end
 
     def invalid_highlights
-      invalid_feedback_highlights
+      invalid_feedback_highlights.concat(invalid_plagiarism_texts)
     end
```
anathomical: I don't think that this matters that much, but it feels like maybe there should be a new name for the combined thing that's highlights plus plagiarism.  Maybe `invalid_related_texts` or something?

anathomical (approved):
A couple of minor thoughts about naming things and messaging, but extending the existing pattern makes sense to me.  I also appreciate moving the text manipulation code into a helper module.  That's smart.

eadams17 (approved):
Thanks for putting this together!

services/QuillLMS/client/app/bundles/Staff/components/evidence/configureRules/rulePlagiarismAttributes.tsx
emilia-friedberg: This is a good point. I'm going to change this to say "may not trigger", and will drop an extra note to the curriculum team that it may or may not be necessary to update all of these. Some of the ones I noticed in sampling the data were the difference of a word or two.

PR #12463, opened 2024-10-17
SelectAClassroom fetch optimization

happythenewsad:
## WHAT / HOW
Identifies an unnecessary, high traffic network call on the /classrooms view, removes it, and stitches various logics back together. 

## WHY
`profiles#student_profile_data` is called ~1MM times per day. Any reduction in calls or optimization for this transaction will result in serious compute savings. 

`/classrooms` is requested about 320k times per day, and `/classrooms` will no longer call `profiles#student_profile_data` after this patch

### Screenshots

### Notion Card Links
https://www.notion.so/quill/Product-Board-30f97e4eb01246dbb8264253fb385073?p=10dd42e6f941801db084d924df5e4718&pm=s

1. Login as student. 
2. Observe successful Select A Classroom render, and no `profiles#student_profile_data` fetch 
3. Click a class. Observe successful render. 
4. Refresh browser. 
5. Navigate back to Select a Classroom. 
6. Repeat steps 1-5 as teacher previewing their student

PR Checklist | Your Answer
------------ | -------------
Have you added and/or updated tests? |  manual testing


eadams17 (approved):
Nice! Hopefully we see significant computational savings with this

services/QuillLMS/client/app/bundles/Teacher/containers/StudentProfile.jsx
```diff
@@ -29,11 +29,8 @@ class StudentProfile extends React.Component {
     if (classroomId) {
       handleClassroomClick(classroomId);
       fetchStudentProfile(classroomId)
```
anathomical, line 31: Do we know if this `fetchStudentProfile` call in cases where a `classroomId` is passed in is used?  It's probably not worth a deep-dive investigation since I suspect that the fix you have here resolves most calls, but I wonder if we can eliminate this entirely.

anathomical (approved):
Nice catch on this issue.

services/QuillLMS/client/app/bundles/Teacher/containers/StudentProfile.jsx
happythenewsad: Yeah, that's something we could look in to in the future. The `classroomId` code path is used, but it could be cleaned up a bit

PR #12468, opened 2024-10-21
Fix incorrect pathing bug on manifest.json

cissyxyu:
## WHAT
Fix a bug where the link to `manifest.json` file was 404ing on some pages.

## WHY
We don't want any 404 errors on the site and a missing manifest file prevents the user from downloading our PWA app.

## HOW
Use an absolute path here instead of relative path, because some of the relative paths such as login page and teacher home page are on a different root path, so the manifest file is not found.

### Screenshots

### Notion Card Links

Deploy to staging and verify that the pages where this manifest is currently 404ing are not broken after the update.

PR Checklist | Your Answer
------------ | -------------

### DIFF
--- a/services/QuillLMS/app/views/application/_head.html.erb
+++ b/services/QuillLMS/app/views/application/_head.html.erb
@@ -38,7 +38,7 @@
   <meta name="og:description" content="<%= @description || @content %>">
   <link rel="apple-touch-icon" href="<%= ENV['CDN_URL'].gsub(/[‘’]/, '') + "/images/logos/apple-touch-icon.png" %>">
   <link rel="canonical" href="https://www.quill.org<%= request.path %>">
-  <link rel="manifest" href="<%= ENV['DEFAULT_URL'] + "/manifest.json" %>" />
+  <link rel="manifest" href="/manifest.json" />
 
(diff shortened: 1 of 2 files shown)


happythenewsad (approved):

brendanshean (approved):
Thank you for fixing this bug.

services/QuillLMS/app/views/application/_head.html.erb
```diff
@@ -38,7 +38,7 @@
   <meta name="og:description" content="<%= @description || @content %>">
   <link rel="apple-touch-icon" href="<%= ENV['CDN_URL'].gsub(/[‘’]/, '') + "/images/logos/apple-touch-icon.png" %>">
   <link rel="canonical" href="https://www.quill.org<%= request.path %>">
-  <link rel="manifest" href="manifest.json" />
+  <link rel="manifest" href="<%= ENV['DEFAULT_URL'] + "/manifest.json" %>" />
```
anathomical: I think there's a slightly cleaner way of doing this that relies on the subtleties of `href`.  `filename.json` will build a link to a path that's the same as the current page with only the section after the last `/` changed.  `host/session/new` would become `host/session/filename.json`, which is what I assume you have been running into here.  But if you specify a `href` value starting with `/`, it creates a link to a path that's just the host and the `href` value.  So `quill.org/session/new` gets changed to `quill.org/filename.json`.

All of which is a long-winded way of saying that I think that instead of including `ENV['DEFAULT_URL']`, you can just make this `href="/manifest.json"`

anathomical (changes requested):
I think this would be simpler if we didn't have to reference `ENV`.  I've proposed an approach that should allow that.

anathomical (approved):

PR #12484, opened 2024-10-24
Change StemVaultId to DatasetId in Guideline model

brendanshean:
## WHAT
Change the foreign key on Guideline model from `stem_vault_id` to `dataset_id`

## WHY
Curriculum would like to have these vary at the dataset level rather than stem vault one.

## HOW
Write a migration for Guideline that 
1.  Adds the `dataset_id`
2.  For each existing guideline, create new guideline records for each dataset associated with the existing guidelines' stem vault_id.  
3.  Update any existing LLMPromptGuidelines with the appropriate guideline_id for each of these newly created records
4.  Remove stem_vault_id from the Guideline table
5. Update various pieces of the UI.

### Screenshots

### Notion Card Links

I have verified:
Optimal and Suboptimal Guideline indices are no longer present on the StemVault show page
Optimal and Suboptimal Guideline indices are now present on Dataset show pages
Guidelines can be created through the UI.
The hide guideline functionality works and only applies under the specific Dataset.

PR Checklist | Your Answer
------------ | -------------

### DIFF
--- a/services/QuillLMS/db/migrate/20241024190034_change_stem_vault_id_to_dataset_id_in_evidence_research_gen_ai_guidelines.evidence.rb
+++ b/services/QuillLMS/db/migrate/20241024190034_change_stem_vault_id_to_dataset_id_in_evidence_research_gen_ai_guidelines.evidence.rb
@@ -1,0 +1,53 @@
+# frozen_string_literal: true
+
+# This migration comes from evidence (originally 20241024174949)
+class ChangeStemVaultIdToDatasetIdInEvidenceResearchGenAIGuidelines < ActiveRecord::Migration[7.1]
+  def up
+    add_column :evidence_research_gen_ai_guidelines, :dataset_id, :integer
+
+    Evidence::Research::GenAI::Guideline.reset_column_information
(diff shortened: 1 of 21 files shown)


emilia-friedberg (commented):
This looks good overall, but I'm wondering if you wanted to add `notes` to the table at the same time? Also, maybe I'm just missing something, but is there a mechanism in place here to copy over guidelines to new datasets when they're created? I think we need that so curriculum doesn't have to write them from scratch for each dataset, but maybe it's coming later.

brendanshean:
> This looks good overall, but I'm wondering if you wanted to add `notes` to the table at the same time? Also, maybe I'm just missing something, but is there a mechanism in place here to copy over guidelines to new datasets when they're created? I think we need that so curriculum doesn't have to write them from scratch for each dataset, but maybe it's coming later.

Thank you for the reminder on notes.  I will add that.  

I am adding the mechanism for copying over guidelines to new datasets along with similar functionality for relevant texts in a separate PR.

emilia-friedberg (approved):

PR #12518, opened 2024-11-01
trial runner page pt 3

emilia-friedberg:
## WHAT
Add functionality to view and create prompt examples, and actually run the trial from the `TrialRunner` page. This is the last of the work for this page.

## WHY
We want to be able to run trials in the Evidence CMS.

## HOW
Mostly following React/Rails paradigms established elsewhere in the project. I did notice, in the course of testing this, that the structure I was using for `requestPut` and `requestPost` handling wasn't handling errors correctly (due to the backend returning a 404 instead of a 200), so I updated some of the functions in that file to fix that.

Bonus: noticed the `RadioButton` component had the wrong type set, so I updated that, which updated some other snapshots around the site.

### Screenshots
Screenshot 2024-11-04 at 12 23 28 PM
Screenshot 2024-11-04 at 12 23 53 PM
Screenshot 2024-11-04 at 12 23 37 PM

### Notion Card Links
https://www.notion.so/quill/Trial-Runner-Page-12ed42e6f9418002b2f0e2e7917a8117?pvs=4

Tested on staging, and having Jamie test on staging.

PR Checklist | Your Answer
------------ | -------------


brendanshean (approved):
This is really coming along quickly!  I really like all the thorough testing in here.  Good catch on the radio buttons.

services/QuillLMS/client/app/bundles/Staff/components/evidence/llmPromptDatasets/trialRunner/promptExamplesSection.tsx
```diff
@@ -0,0 +1,140 @@
...
+import { OPTIMAL, SUBOPTIMAL, } from '../../../../../../constants/evidence'
+
+function getRandomSample(array) {
+  // Get a random number of items to sample, between 1 and the length of the array
+  const sampleSize = Math.floor(Math.random() * array.length) + 1;
+
+  // Shuffle the array and slice it to get the sample
+  const shuffledArray = array.sort(() => Math.random() - 0.5);
+  return shuffledArray.slice(0, sampleSize);
+}
```
eadams17, line 18: it might be worth moving this to the Shared bundle in case we want to use it in the future elsewhere?

eadams17 (approved):
LGTM!

services/QuillLMS/client/app/bundles/Staff/components/evidence/llmPromptDatasets/trialRunner/promptExamplesSection.tsx
emilia-friedberg: I think this makes sense as a call-out, but I prefer not to abstract stuff until we have a second use case for it.

PR #12524, opened 2024-11-06
UI updates/ Manage Classes modals

eadams17:
## WHAT
update all of the modals used throughout the Manage Classes page

## WHY
this is part of a large design overhaul of this section of the dashboard

## HOW
a lot of React and CSS changes

just a note: I refactored some of the modal files from class components into functional components whenever it was relatively straightforward

### Screenshots
Screen Shot 2024-11-05 at 9 47 41 PM
Screen Shot 2024-11-05 at 9 48 42 PM
Screen Shot 2024-11-05 at 9 48 58 PM

### Notion Card Links
https://www.notion.so/quill/Update-UI-Manage-Classes-Modal-Dialogs-f6283bcb39934012b98ad8a39405b662?pvs=4

tested on staging that all modals render as defined by the design specs

PR Checklist | Your Answer
------------ | -------------


services/QuillLMS/client/app/bundles/Shared/components/shared/input.tsx
```diff
@@ -123,11 +123,11 @@ export class Input extends React.Component<InputProps, InputState> {
     }
   }
 
-  renderCancelSymbol = () => {
+  renderCancelSymbol = (id) => {
     const { inactive } = this.state
     const { handleCancel, value } = this.props
```
anathomical: Minor thought, so take or leave.  It looks like the `id` value that gets passed into this function always come from `this.props`.  Does it make sense to just extract `id` from `this.props` in this function instead of passing it around?  Doing the extraction here puts the value closer to the call so it's more obvious what value is going into the `id` attribute.

services/QuillLMS/client/app/bundles/Shared/styles/button.scss
```diff
@@ -44,69 +44,80 @@ a.quill-button {
...
     font-weight: 700;
     &.disabled {
-      background-color: $quill-grey-5 !important;;
+      background-color: $quill-grey-5 !important;
       color: $quill-grey-30 !important;;
-      border: 1px solid $quill-grey-15 !important;;
+      border: 1px solid $quill-grey-15 !important
     }
     &.gold {
+      border: 1px solid $quill-gold-dark-50;
```
anathomical, line 57: Given that we're not changing the border width or style, I think all of these color modifiers should just use the `border-color` directive.  That'll make it easier to update size and style for all color variations in the future without having to modify each one individually.

services/QuillLMS/client/app/bundles/Teacher/components/classrooms/__tests__/create_a_class_form.test.jsx
```diff
@@ -16,7 +16,7 @@ describe('CreateAClassForm component', () => {
     })
 
     it('should have a disabled button', () => {
-      expect(wrapper.find('.quill-button-archived').hasClass('disabled')).toBe(true)
+      expect(wrapper.find('.quill-button').at(2).hasClass('disabled')).toBe(true)
```
anathomical: Another take-or-leave comment: Using `at` to index the button's position in the wrapper feels a bit brittle (this could break just by adding another button).  I note that the relevant button looks to be the only one in the component with the `contained` class, so maybe a more focused selector would be a little cleaner here?  It probably doesn't really matter that much, but I figured I'd call it out.

services/QuillLMS/client/app/bundles/Teacher/components/classrooms/__tests__/create_a_class_form.test.jsx
```diff
@@ -28,7 +28,7 @@ describe('CreateAClassForm component', () => {
     wrapper.setState({ name: 'Classroom', code: 'happy-day', grade: 7 })
 
     it('should not have a disabled button', () => {
-      expect(wrapper.find('.quill-button-archived').hasClass('disabled')).toBe(false)
+      expect(wrapper.find('.quill-button').at(2).hasClass('disabled')).toBe(false)
```
anathomical: Same as my above comment if we decide to go with `find('.quill-button.contained')` instead of `find(...).at(2)`

services/QuillLMS/client/app/bundles/Teacher/components/classrooms/edit_student_account_modal.tsx
```diff
@@ -11,115 +11,118 @@ interface EditStudentAccountModalProps {
...
+    const { id, value } = target
+    if (id === FIRST_NAME) {
+      setFirstName(value)
+    }
+    if (id === LAST_NAME) {
+      setLastName(value)
+    }
+    if (id === USERNAME) {
+      setUsername(value)
+    }
```
anathomical: This is purely stylistic, so take or leave, but I prefer objects with variable keys for these sorts of patterns.  It mostly just reads better to me than a sequence of `if` blocks.  I think that's because it makes it clear up-front that all of these are mutually exclusive, I don't have to go through each condition to understand that:
```javascript
const functionsById = {
  [FIRST_NAME]: setFirstName,
  [LAST_NAME]: setLastName,
  [USERNAME]: setUsername
}
functionsById[id](value)
```

anathomical (approved):
I had a couple of take-or-leave style suggestions, but overall this looks good.  I appreciate the refactor that you've done here, especially moving so many things to constants and bringing the code in line with our more modern patterns.

services/QuillLMS/app/assets/stylesheets/pages/classrooms.scss
```diff
@@ -270,16 +273,6 @@
...
-
-  // this is stuff that should be coming from the component library but is not working right now
-  .character-limit {
-    color: #7f7f7f;
-  }
-
-  .has-character-limit:not(.error) {
-    height: 65px;
-  }
-
```
emilia-friedberg, line 282: Yay!

services/QuillLMS/client/app/bundles/Teacher/components/classrooms/class_code_link.tsx
```diff
@@ -9,59 +9,32 @@ interface ClassCodeLinkProps {
...
-}
-
-export default class ClassCodeLink extends React.Component<ClassCodeLinkProps, ClassCodeLinkState> {
-  constructor(props) {
-    super(props)
-
-    this.showSnackbar = this.showSnackbar.bind(this)
-  }
+export const ClassCodeLink = ({ next, back, showSnackbar, classroom }: ClassCodeLinkProps) => {
+  const link = `quill.org/join/${classroom?.code}`
```
emilia-friedberg: Is it possible for `classroom` to be null here? That seems like confusing UX since this link won't work.

emilia-friedberg (approved):
These look great. I appreciate all the extra refactoring here.

services/QuillLMS/client/app/bundles/Shared/components/shared/input.tsx
eadams17: yeah, that's a good point-- I missed that while I was adding this functionality. I'll update

services/QuillLMS/client/app/bundles/Teacher/components/classrooms/__tests__/create_a_class_form.test.jsx
eadams17: that's a good suggestion, I'll add the `contained` selector

PR #12526, opened 2024-11-06
Evidence GenAI Label Retriever

dandrabik:
## WHAT
Create a v1 of the GenAI Label Retriever
## WHY
We want this to replace AutoML, so starting the process of writing the code that will be used by the production endpoint.
## HOW
**I made some decisions here for production use:**
1. Use a string instead of a file read for the template since it is small.
2. Don't store the embeddings in the DB for student entries (I think we want to just throw this away to prevent it from making our DB huge).
3. Only query close neighbors once for speed.
4. Add a Proposed Logic to the algorithm

Algorithm
1. Find X number of close examples in our DB
2. If they are all the same label, return that label
3. If the first label is very, very close, use that label
4. If no examples are anywhere close to the label, return miscellaneous feedback (I want to try this out, easy to remove)
5. Otherwise make LLM Call with Prompt and examples 

### Notion Card Links
https://www.notion.so/quill/Build-a-GenAI-LabelRetriever-135d42e6f94180feae37ffc7eec7cb2a

I've done extension testing in the console for now. This isn't hooked up to anything on the site yet (that code is coming next), so safe to deploy.

PR Checklist | Your Answer
------------ | -------------
Have you deployed to Staging? | No, this code isn't used by the site yet.


services/QuillLMS/engines/evidence/app/services/evidence/gen_ai/label_feedback/retriever.rb
```diff
@@ -0,0 +1,66 @@
...
+        MISC_LABEL = Evidence::Label::MISC_LABEL
+
+        attr_reader :prompt, :entry
+
+        def initialize(prompt:, entry:)
+          @prompt = prompt
+          @entry = entry
+        end
+
+        def run
```
dandrabik, line 22: This is the main method to review, everything stems from it.

services/QuillLMS/engines/evidence/app/services/evidence/gen_ai/label_feedback/retriever.rb
```diff
@@ -0,0 +1,66 @@
...
+        MISC_THRESHOLD = 0.4
+        MISC_LABEL = Evidence::Label::MISC_LABEL
+
+        attr_reader :prompt, :entry
+
+        def initialize(prompt:, entry:)
+          @prompt = prompt
+          @entry = entry
+        end
+
```
dandrabik, line 21: The `run` method is the main method to review. Everything else stems from it.

services/QuillLMS/engines/evidence/app/models/evidence/labeled_entry.rb
```diff
@@ -84,5 +93,9 @@ def nearest_label
         self.label_transformed = label
       end
     end
+
+    def self.embedding_for(entry:)
```
brendanshean, line 97: I ilke this extraction.

brendanshean (approved):
Looks good to me!

PR #12528, opened 2024-11-07
individual trial page part one

emilia-friedberg:
## WHAT
Part one of the individual trial page, featuring all the top-level dataset and trial information and modals to update the trials' notes and view the LLM Prompt and GEval information.

## WHY
This is the first section of work for the individual trial page. Next up is the evaluation table and ability to create a data subset.

## HOW
Updating `trials_controller` actions to get and update the data I need, and then build out the views in React and CSS. I ported over effectively the same HTML that Brendan used in the ERB version of this page for the histogram and confusion matrix.

Bonus: some QA fixes for the previous PR, and I added a classname to the div that wraps the `TextArea`, in order to be able to apply CSS to it, so that updated some additional snapshot tests.

### Screenshots
Screenshot 2024-11-07 at 11 26 15 AM
Screenshot 2024-11-07 at 11 26 07 AM
Screenshot 2024-11-07 at 11 23 21 AM

### Notion Card Links
https://www.notion.so/quill/Individual-Trial-Page-135d42e6f941806ba3bccf4e3b951b50?pvs=4

Tested locally, will have Jamie test on staging when the next part is complete.

PR Checklist | Your Answer
------------ | -------------
Have you deployed to Staging? |  NOT YET


services/QuillLMS/client/app/bundles/Staff/components/evidence/llmPromptDatasets/individualTrial/__tests__/histogram.test.tsx
```diff
@@ -0,0 +1,45 @@
...
+  });
+
+  it('calculates the correct maximum value for bar heights', () => {
+    const { container } = render(<Histogram histogramData={mockHistogramData} />);
+    const maxValue = Math.max(...Object.values(mockHistogramData));
+    const bars = container.querySelectorAll('.bar');
+
+    bars.forEach((bar, index) => {
+      const frequency = mockHistogramData[Object.keys(mockHistogramData)[index]];
+      const expectedHeight = `${(frequency / maxValue) * 100}px`; // 100 is HISTOGRAM_HEIGHT
```
anathomical: I'm not sure this is actually better, but is it worth just importing the constant to the test file and using it instead of hard-coding `100`?  The upside is that any intentional changes to the constant will not require anyone to update the tests.  The downside is that if someone changes the constant without realizing that it might break this functionality, the test will catch it.  I'm inclined to think the former is slightly better since it's a constant definition, but I'll defer to your better understanding of the code.

services/QuillLMS/client/app/bundles/Staff/components/evidence/llmPromptDatasets/individualTrial/__tests__/trialNotesModal.test.tsx
```diff
@@ -0,0 +1,115 @@
...
+    jest.clearAllMocks();
+  });
+
+  it('renders the modal with initial trial notes', () => {
+    render(
+      <TrialNotesModal {...defaultProps} />
+    );
+
+    expect(screen.getByText('Trial Notes')).toBeInTheDocument();
+    expect(screen.getByLabelText('Notes')).toHaveValue('Initial trial note');
```
anathomical: Small style thing to take or leave: I'm generally inclined to avoid risks of typo-ing string constants that are defined in the test by referencing them by variable name in `expect`.  I think this works with `expect(...).toHaveValue(trial.notes)`, which makes it explicit that we have this output expectation because of the value of the input.

services/QuillLMS/client/app/bundles/Staff/components/evidence/llmPromptDatasets/individualTrial/__tests__/trialNotesModal.test.tsx
```diff
@@ -0,0 +1,115 @@
...
+
+  it('updates the notes when input changes', async () => {
+    render(
+      <TrialNotesModal {...defaultProps} />
+    );
+
+    const notesInput = screen.getByLabelText('Notes');
+    await userEvent.clear(notesInput);
+    await userEvent.type(notesInput, 'Updated note');
+    expect(notesInput).toHaveValue('Updated note');
```
anathomical: Similar to my above note, another take-or-leave style thing: I like to set the input/output string to a const to use in both places in order to make it clear that there's an explicit relationship:
```javascript
const updatedNote = 'Updated note'
await useEvent.type(notesInput, updatedNote)
expect(notesInput).toHaveValue(updatedNote)
```

services/QuillLMS/client/app/bundles/Staff/components/evidence/llmPromptDatasets/individualTrial/__tests__/trialNotesModal.test.tsx
```diff
@@ -0,0 +1,115 @@
...
+      successCallback({ ...trial, notes });
+    });
+
+    render(
+      <TrialNotesModal {...defaultProps} />
+    );
+
+    const saveButton = screen.getByText('Save');
+    await userEvent.clear(screen.getByLabelText('Notes'));
+    await userEvent.type(screen.getByLabelText('Notes'), 'New trial note');
```
anathomical: Again, I like to put repeated strings into a const to make it clear that the input/output are connected.

services/QuillLMS/client/app/bundles/Staff/components/evidence/llmPromptDatasets/individualTrial/__tests__/viewGEvalModal.test.tsx
```diff
@@ -0,0 +1,60 @@
...
+    expect(screen.getByText('Metric')).toBeInTheDocument();
+    expect(screen.getByText('Accuracy')).toBeInTheDocument();
+    expect(screen.getByText('Version')).toBeInTheDocument();
+    expect(screen.getByText('1')).toBeInTheDocument();
+    expect(screen.getByText('Task Introduction')).toBeInTheDocument();
+    expect(screen.getByText('This is a task introduction for testing.')).toBeInTheDocument();
+    expect(screen.getByText('Evaluation Criteria')).toBeInTheDocument();
+    expect(screen.getByText('These are the evaluation criteria.')).toBeInTheDocument();
+    expect(screen.getByText('Evaluation Steps')).toBeInTheDocument();
+    expect(screen.getByText('These are the evaluation steps.')).toBeInTheDocument();
```
anathomical: With this many values, my preferred approach of making a bunch of `const`s feels a bit heavy-handed.  That said, comparing the input and output, it may not be entirely obvious in this case whether the `expect(screen.getByText('1')).toBeInTheDocument();` refers to the `1` in `GEvalInterface.id` or the `1` in `GeEvalInterface.version`.  I'm pretty sure it's the latter based on context, but making that explicit by either using a `const` for at least that value, or changing the `version` or `id` value so that there aren't any duplicates in the input would be a bit clearer.

services/QuillLMS/client/app/bundles/Staff/components/evidence/llmPromptDatasets/individualTrial/__tests__/viewLLMPromptModal.test.tsx
```diff
@@ -0,0 +1,55 @@
...
+    expect(asFragment()).toMatchSnapshot();
+  });
+
+  it('displays the correct LLM Prompt information', () => {
+    render(<ViewLLMPromptModal {...defaultProps} />);
+
+    expect(screen.getByText('LLM Prompt Template')).toBeInTheDocument();
+    expect(screen.getByText('Sample Prompt')).toBeInTheDocument();
+    expect(screen.getByText('Contents')).toBeInTheDocument();
+    expect(screen.getByText('This is the prompt content for testing.')).toBeInTheDocument();
```
anathomical: If you do decide to update to `const` declarations for these comparisons, just a quick flag that you'd want one for `Sample Prompt` and `This is the prompt content for testing.`.  Which would also make it a bit more explicit that the `LLM Prompt Template` and `Contents` text are not values that we expect because of testing setup, but are expected because of the root component itself.

services/QuillLMS/client/app/bundles/Staff/utils/evidence/genAIAPIs.ts
```diff
@@ -31,6 +31,19 @@ export const fetchDatasetAndAssociatedRecords = async ({ queryKey }) => {
...
+  const [key, datasetId, trialId]: [string, string, string] = queryKey
+
+  const response = await researchGenAiApiFetch(`datasets/${datasetId}/trials/${trialId}`);
+
+  let data = await response.json();
+
+  return {
+    error: handleApiError('Failed to fetch trial, please refresh the page.', response),
+    data
+  };
```
anathomical, line 57: I don't actually have a strong thought here, but this return pattern seems pretty common for our error handling.  I wonder if there's some way to avoid the repetition.  Of course, these aren't actually identical, so maybe it's more trouble than it's worth.

anathomical (approved):
I've got a number of style thoughts, but they're all take-or-leave items.  Overall, I wanted to call out that I've actually enjoyed reading over components written entirely from scratch as they showcase our modern style approaches to front-end development, and it's just so readable and understandable.

emilia-friedberg:
@anathomical these are all really helpful notes, thank you for the thorough review. I'm updating in accordance with your suggestions and will use those patterns going forward.

services/QuillLMS/engines/evidence/app/controllers/evidence/research/gen_ai/trials_controller.rb
```diff
@@ -64,15 +64,37 @@ def create
           end
         end
 
+        def update
+          if trial.update(trial_params)
```
brendanshean: Since many of the attributes on trial are read-only, I like to use separate params methods to make transparent the difference.  For example,  the LLMs [controller](https://github.com/empirical-org/Empirical-Core/blob/develop/services/QuillLMS/engines/evidence/app/controllers/evidence/research/gen_ai/llms_controller.rb#L27) has `llm_create_params` and `llm_update_params`.  While protection does exist on the model via `attr_readonly` I think the different methods highilight that the params should be different objects.

brendanshean (approved):
I think all of these modals will make for a better UX when running trials.  Nice work!

services/QuillLMS/client/app/bundles/Staff/components/evidence/llmPromptDatasets/individualTrial/confusionMatrix.tsx
```diff
@@ -0,0 +1,55 @@
+import React from 'react';
+
+const ConfusionMatrix = ({ confusionMatrix }) => {
```
eadams17, line 3: welcome to the.... Matrix?

services/QuillLMS/client/app/bundles/Staff/styles/evidence_manager/llm_prompt_datasets.scss
```diff
@@ -298,18 +309,46 @@
...
+  }
+  h4 {
+    @include display-s;
+    margin-top: 16px;
+    margin-bottom: 4px;
+  }
+}
+
+.quill-modal.view-llm-prompt-modal, .quill-modal.view-g-eval-modal {
+  width: 700px;
```
eadams17: maybe do a `max-width: 700px` and `width: 100%`?

eadams17 (approved):
Looking good!

PR #12531, opened 2024-11-08
individual trial page pt 2

emilia-friedberg:
## WHAT
Add the `EvaluationTable` to the `IndividualTrial` page, complete with the ability to filter and create a data subset from this page.

## WHY
This is the last piece of functionality we needed for this page.

## HOW
Have the backend pass up the `LLMExamples` for this table, and update the `DatasetController`.

Note: I decided to split up this API call because getting the `LLMExamples` is slow, and I figured it was better to let the rest of the page load concurrently.

### Screenshots
Screenshot 2024-11-11 at 9 01 31 AM
Screenshot 2024-11-11 at 9 01 17 AM

### Notion Card Links
https://www.notion.so/quill/Gen-AI-CMS-Project-2-Embed-Trial-Runner-in-CMS-61c994a106854512b449ccc82be573c2?pvs=4#135d42e6f941801d805edf526a69e18d

Tested the flows personally, and having Jamie test them on staging.

PR Checklist | Your Answer
------------ | -------------


services/QuillLMS/engines/evidence/app/controllers/evidence/research/gen_ai/trials_controller.rb
```diff
@@ -93,6 +93,10 @@ def show
           end
         end
 
+        def llm_examples
```
emilia-friedberg: This is here due to the error with `LLMExamplesController`: `NameError:
  uninitialized constant Evidence::Research::GenAI::LLMExamplesController`. This is not RESTful, but works.

services/QuillLMS/engines/evidence/app/controllers/evidence/research/gen_ai/llm_examples_controller.rb
```diff
@@ -0,0 +1,15 @@
+# frozen_string_literal: true
+
+module Evidence
+  module Research
+    module GenAI
+      class LLMExamplesController < ApplicationController
```
emilia-friedberg, line 6: Further complication: when I deploy this to staging, the app crashes with this error: `2024-11-11T14:15:41.524077+00:00 app[worker.1]: expected file /app/engines/evidence/app/controllers/evidence/research/gen_ai/llm_examples_controller.rb to define constant Evidence::Research::GenAI::LlmExamplesController, but didn't`. `LLM` is listed in the `initializers/inflections.rb` file, however, and other controllers don't seem to be having this issue.

brendanshean (approved):
This is looking great!  It's very nice to see a lot of that <script> code from the erb files getting cleaned up into proper React components.

services/QuillLMS/client/app/bundles/Staff/components/evidence/llmPromptDatasets/individualTrial/__tests__/individualTrial.test.tsx
```diff
@@ -0,0 +1,89 @@
...
+};
+
+describe('IndividualTrial', () => {
+  it('loads data', async () => {
+    renderWithRouterAndQueryClient();
+
+    await waitFor(() => expect(screen.queryByAltText('Loading spinner')).not.toBeInTheDocument()); // Disappears when data loads
+
+    expect(screen.getByText('Test Activity')).toBeInTheDocument();
+    expect(screen.getByText('Mock EvaluationSection')).toBeInTheDocument(); // Mocked EvaluationSection
```
anathomical: These are both another opportunity to declare a constant and use it in both set-up and evaluation.  As with other items like this that I've flagged, this is totally a take-or-leave comment since these cases are all straightforward enough that I don't think anyone who looks at this later will be confused.

services/QuillLMS/engines/evidence/spec/controllers/evidence/research/gen_ai/llm_examples_controller_spec.rb
```diff
@@ -0,0 +1,33 @@
...
+          allow(controller).to receive(:current_user).and_return(user)
+          allow(user).to receive(:staff?).and_return(true)
+        end
+
+        describe 'GET #index' do
+          it 'returns all llm examples for the trial' do
+            get :index, params: { dataset_id: dataset.id, trial_id: trial.id }, format: :json
+
+            expect(response).to have_http_status(:ok)
+            expect(parsed_response['llm_examples']).to eq(JSON.parse(trial.llm_examples.to_json))
```
anathomical: Can we use `trial.llm_examples.as_json` instead of `JSON.parse(trial.llm.examples.to_json)`?  I suspect that `as_json` adds some additional keys that we wouldn't expect to be in the raw JSON, but that might make this a good candidate for `include` over `eq`.  `expect(A).to include(B)` will pass if A includes at least all values in B, but A is allowed to have additional data beyond that.  If there is any data in B that is not in A, then the test will fail.

This test may actually be specific enough that inclusion isn't detailed enough, in which case I'd say leave it as is.

anathomical (approved):
I still continue to appreciate how we're approaching things here.  Our modern code feels so clean most of the time.  A couple of take-or-leave style items, but this is good to go as-is.

PR #12533, opened 2024-11-08
Evidence Generative Label Check

dandrabik:
## WHAT
Update Evidence Prompts to use the `Activity.ai_type` to choose which AI Type to use: `AutoML`, `GenAI`, or `GenerativeLabel`

Create a `Evidence::Check::GenerativeLabel` check to power our AutoML replacement. Changing the `ai_type` to `GenerativeLabel` turns this on for that activity. 
## WHY
We want to be able to turn on an test our AutoML replacement.
## HOW
Build a response from the label and follow the logic of our discussed algorithm, including using low confidence feedback.

I also included a hash of rag limits as an fast way of getting this up. We can put this in the DB if needed.

You can test the activity [here](https://dan.quill.org/evidence/#/play?uid=221&skipToStep=2)

### Notion Card Links
https://www.notion.so/quill/ea241f31a84c4b79ad56e83e99f7ee93?v=469e2a0c2d8c41539bbe1c8690c4afd7

I'm still doing more QA, but I've been testing on staging extensively and fixing issues as a I find them. I've changed activity 221 (Junk food in school) to `GenerativeLabel` feedback.

Staging has been broken today, but I will confirm on Monday:
It properly does
- Giving feedback
- Storing feedback in feedback_histories
- highlighting
- storing metadata about the request (label_text and closest_distance)

PR Checklist | Your Answer
------------ | -------------


services/QuillLMS/engines/evidence/app/models/evidence/highlight.rb
```diff
@@ -47,6 +47,14 @@ def serializable_hash(options = nil)
       ))
     end
 
+    def feedback_hash
+      {
+        type: highlight_type,
+        text: text,
```
brendanshean: You can use parameter shorthand here: `text:`

services/QuillLMS/engines/evidence/app/services/evidence/gen_ai/label_feedback/response_builder.rb
```diff
@@ -0,0 +1,69 @@
+# frozen_string_literal: true
+
+module Evidence
+  module GenAI
+    module LabelFeedback
+      class ResponseBuilder < ApplicationService
+        OPTIMAL = Evidence::LabeledEntry::COLLAPSED_OPTIMAL_LABEL
+        OPTIMAL1 = 'Optimal_1'
```
brendanshean: I know we are collapsing Optimal_1, Optimal_2,..., Optimal_n to Optimal.  Then it looks like this label is remapped back to Optimal_1 with prompt.active_labels.  Is this Optimal_1 different than the original one?

brendanshean (approved):
Nice work on getting these pieces all connected.

services/QuillLMS/engines/evidence/app/services/evidence/gen_ai/label_feedback/response_builder.rb
dandrabik: I'm not sure I follow the question, but I'm taking it as a sign that this step is confusing. So I re-visited this section and refactored it to be more solid (and hopefully clearer).

The goal of this code is that, for optimal labels, we are getting a output text from the LLM of the form "Optimal", but there isn't a rule with label "Optimal" in our system (since we collapsed them). How do we get that rule? I was translating "Optimal" to "Optimal_1", but on second look that feels a little fragile since maybe "Optimal_1" doesn't exist. Instead I'm now looking for the ["first active optimal automl rule"](https://github.com/empirical-org/Empirical-Core/pull/12533/commits/f7bffd67a3cde112326e1acc3369a73f40cbd41b) using the rule traits and ignoring the label text.

brendanshean: This is much clearer now, specifically, the "first active optimal automl rule" (which might be Optimal_1 in many cases).  Thank you!

PR #12535, opened 2024-11-12
data subset page

emilia-friedberg:
## WHAT
Update the `IndividualDataset` page and related components to work for Data Subset views.

## WHY
We want these components to render Data Subsets as well as parent Datasets.

## HOW
Mostly adding minor conditionals to rendering logic, because the underlying mechanisms are all effectively the same. The trickiest bit was getting `react-query` to handle state reloads when switching back and forth between datasets, because the component itself doesn't change.

### Screenshots
Screenshot 2024-11-12 at 9 35 53 AM

### Notion Card Links
https://www.notion.so/quill/Data-Subset-Trial-Page-13bd42e6f94180948e90cd94eda43ca9?pvs=4

QAed on staging and having Jamie QA as well.

PR Checklist | Your Answer
------------ | -------------


brendanshean (commented):
Nice work on figuring out this nested view.

services/QuillLMS/engines/evidence/app/controllers/evidence/activities_controller.rb
```diff
@@ -69,7 +69,7 @@ def increment_version
     end
 
     def stem_vaults
-      render json: @activity.stem_vaults
+      render json: @activity.stem_vaults.map { |sv| sv.serializable_hash.merge({ datasets: sv.datasets.whole }) }
```
anathomical, line 72: This is probably an over-optimization at this point, so feel free to skip it, but I wonder if it makes sense for the logic in this map to be a named method on the model.  That would make it easy to re-use if this particular data shape is something we use other places:
```
// In the model
def hash_with_whole_dataset = serializable_hash.merge({datasets: datasets.whole})

// In the controller
render json: @activity.stem_vaults.map(&:hash_with_whole_dataset)
```

brendanshean (approved):

anathomical (approved):

PR #12536, opened 2024-11-12
fix renaming classroom save function and add tests

emilia-friedberg:
## WHAT
Fix a bug with renaming classroom save function.

## WHY
We want this to work.

## HOW
Update the logic to use the props that were actually being passed, and add more thorough tests.

### Screenshots

### Notion Card Links
https://www.notion.so/quill/Teacher-can-t-rename-a-class-13cd42e6f941804882dadd4f83e5d89e

Tested that it works now.

PR Checklist | Your Answer
------------ | -------------


brendanshean (approved):
Thank you for fixing this and adding a test as well.

PR #12541, opened 2024-11-13
Pre post data backfill

anathomical:
## WHAT
- Provide code to enqueue data backfill workers to build historical records of `StudentLearningSequence` for old assignments and completions
- Add some unique indexes to the database tables to increase efficiency and accuracy of the recording logic by enabling `create_or_find_by` calls
- Find some edge cases where default scoping was causing some data to be missed and update the logic to ensure that archived data still gets backfilled
## WHY
- Splitting this up into workers allows us to scale workers up over the weekend, or to break the backfill into multiple parts to make sure it runs during downtime only
- This makes our data cleaner and speeds up the processing time for writing it
- This helps ensure that we get more historical data.  Most of it likely won't be relevant since deleted assignments aren't something people will want reports on, but it's worth being completionist
## HOW
- Create a series of workers to query for and enqueue historical `ClassroomUnit` and `ActivitySession` records for recording into `StudentLearningSequences`
- Write a migration to add some new indexes, then drop old redundant indexes
- Make explicit `unscoped` calls and add new `unscoped` sub-class models for use in `joins`

### Notion Card Links
https://www.notion.so/quill/StudentLearningSequence-Backfill-Planning-13ed42e6f9418052beb6cba612f927c7

- I've run this multiple times on Staging from scratch, and then used a series of SQL queries to identify which records we expect to be created by the backfill and flag any missing
- Confirm that "missing" records all fall within clearly-understood parameters where we would expect them to be missing

PR Checklist | Your Answer
------------ | -------------


services/QuillLMS/app/models/activity.rb
```diff
@@ -312,6 +312,10 @@ def is_diagnostic?
     classification&.key == ActivityClassification::DIAGNOSTIC_KEY
   end
 
+  def is_pre_diagnostic?
```
brendanshean, line 315: More a stylistic thing, but I find the `is_` prefix redundant here.  I realize we have existing ones `is_proofreader?`, `is_evidence?`

services/QuillLMS/app/services/student_learning_sequences/handle_assignment.rb
```diff
@@ -2,65 +2,52 @@
...
+    private def activities
+      Activity.joins(unit_activities: { unit_unscoped: :classroom_units })
+        .where(classroom_units: { id: classroom_unit_id })
     end
 
-    private def create_student_learning_sequence
-      # Do not create new sequences unless we're processing a pre diagnostic assignment
-      return unless pre_diagnostic
+    private def backfill_params
+      return {} unless backfill
```
brendanshean: This feels redundant given the `if backfill` earlier.

services/QuillLMS/app/workers/student_learning_sequences/handle_assignment_worker.rb
```diff
@@ -4,8 +4,10 @@ module StudentLearningSequences
   class HandleAssignmentWorker
     include Sidekiq::Worker
 
-    def perform(classroom_unit_id, student_id)
-      HandleAssignment.run(classroom_unit_id, student_id)
+    # rubocop:disable Style/OptionalBooleanParameter
```
brendanshean: I guess you could define a separate worker if you don't want the rubocop disable:
```
class BackfillHandleAssignmentWorker 
  include Sidekiq::Worker
  def perform(classroom_unit_id, student_id)
    HandleAssignment.run(classroom_unit_id, student_id, backfill: true)
  end
end
```
but I don't have strong feelings either way.

services/QuillLMS/app/services/student_learning_sequences/handle_assignment.rb
```diff
@@ -2,65 +2,52 @@
...
 
-    def run
-      return unless student_learning_sequence
-
-      assign_activities_to_sequence
-    end
+    def run = assign_activities_to_sequence
 
     private def assign_activities_to_sequence
       activities.map do |activity|
```
brendanshean: Should this be `.each`?  I doesn't look like `HandleAssignment.run` uses the values.

If it should be `.each`, you can get rid of the `student_learning_sequence_activity` at the end of the loop.

brendanshean (approved):
This is a very impressive PR.  I really like how you've organized Backfill::BaseWorker class and the child workers, nice atomic operations.

services/QuillLMS/app/models/activity.rb
anathomical: I don't have strong feelings on this one, but I'm inclined to maintain the existing nomenclature.  Thoughts?

services/QuillLMS/app/workers/student_learning_sequences/handle_assignment_worker.rb
anathomical: Actually, that's not a bad suggestion.  I'll refactor.

services/QuillLMS/app/services/student_learning_sequences/handle_assignment.rb
anathomical: Good call-out.  That's a leftover from an earlier version of the code.

services/QuillLMS/app/services/student_learning_sequences/handle_assignment.rb
anathomical: Good point.  The `map` call was really helpful when testing and debugging, but it isn't necessary for the task.  I'll update.

services/QuillLMS/app/models/activity.rb
brendanshean: I think maintaining existing nomenclature for this model is probably best for now.  But I think in general, `is_thing?` boolean methods, should just be `thing?`

anathomical: Yeah, I think that the `?` implies a boolean, so that makes sense to me.

dandrabik (approved):
:+1: LGTM, thanks for putting this together

PR #12547, opened 2024-11-14
update student overview query to account for timezones and test

emilia-friedberg:
## WHAT
Update the query that shows on the Student Overview page under Activity Scores to adjust the `completed_at` date for the teacher's timezone.

## WHY
A teacher pointed out that they saw a discrepancy between these dates and what shows on their Activity Summary, which shouldn't be the case.

## HOW
Pass `current_user` to the `student_overview` query and use it to adjust the `completed_at` date that gets sent to the frontend, as we do with other queries.

### Screenshots

### Notion Card Links
https://www.notion.so/quill/Student-s-Activity-Summary-dates-do-not-match-the-student-s-Activity-Score-dates-13ed42e6f941804d8a50e045001891d1?d=13ed42e6f94180ec85ae001cca79f595

Locally tested that timezones are now taken into account on this page.

PR Checklist | Your Answer
------------ | -------------


services/QuillLMS/spec/queries/progress_reports/student_overview_spec.rb
brendanshean, line 1: I appreciate the shout-out to Chicago.

brendanshean (approved):
Thank you for fixing this

eadams17 (approved):
Thanks for fixing this!

anathomical (approved):
Definitely beyond the scope of this PR, but it feels like this is probably something we should standardize an approach for at some point.  I know we do this specific thing in a couple of places, but there's probably some cleaner thing we could be doing more generally.

Not a high priority, but something sitting in the back of my mind now.

services/QuillLMS/spec/queries/progress_reports/student_overview_spec.rb
emilia-friedberg: @brendanshean unfortunately, 'America/Philadelphia' is not a timezone in our system.

brendanshean: Well, at least not _yet_. 🔔

PR #12549, opened 2024-11-14
turn-off-student-learning-sequence-recording

anathomical:
## WHAT
Temporarily disable the StudentLearningSequence recording
## WHY
So that we can run some migrations that require the table to be empty.  We'll turn them back on as soon as the migration completes
## HOW
Early return instead of enqueuing jobs to record the data

- Test locally that these jobs don't get enqueued when activities are assigned or completed

PR Checklist | Your Answer
------------ | -------------
Have you added and/or updated tests? |  Yes, but only to expect code not to run
Have you deployed to Staging? | No, small change


dandrabik (approved):
LGTM

brendanshean (approved):

PR #12550, opened 2024-11-14
Translations/ restrict available languages

eadams17:
## WHAT
add logic to restrict available languages for Connect and Grammar translated activities

## WHY
we want to restrict languages that are still in "alpha mode" (in other words, languages for which we have translations, but the translation texts still need to be reviewed for accuracy)

## HOW
add an environment variable with permitted languages, and update the `showTranslations` check to check for language presence in this array

### Screenshots

### Notion Card Links
https://www.notion.so/quill/Translation-Project-Complete-Translations-of-other-Languages-11dd42e6f94180559387d370d57ea7af?pvs=4

verified locally and on staging that Portuguese translations do not show (after running the translations script for Portuguese on staging)

PR Checklist | Your Answer
------------ | -------------


anathomical (approved):
I can imagine this ENV variable getting pretty clunky when we're using it to enable dozens of different languages, but I think doing this in ENV makes sense for now as it's quick, easy to update, and doesn't require new database stuff.

We may want to revisit this in the future where we might want more complex configuration, but for now this looks good.

eadams17:
> I can imagine this ENV variable getting pretty clunky when we're using it to enable dozens of different languages, but I think doing this in ENV makes sense for now as it's quick, easy to update, and doesn't require new database stuff.
> 
> We may want to revisit this in the future where we might want more complex configuration, but for now this looks good.

totally! I'll just explain a bit more behind my reasoning to go this route for context. 

I was thinking that it would be nice to have the quick and easy Heroku env update each time we finish a language review process, and remove the need to put up a PR to add a new language each time. I added the env variables to all of our staging environments, but I think I will only need to update the prod env and my staging env with each new language since I'll be the one working on the remaining translations. 

Additionally, my plan is to just remove the check for this env variable (as well as the variable) completely once we finish adding the remaining 15 languages as there will be no need to restrict any languages. I think any updates we would do thereafter would be one-off tweaks, so we wouldn't need to restrict the entire set of translation texts

brendanshean (approved):

PR #12562, opened 2024-11-15
fix bug allowing invalid emails to be saved via admin creation

emilia-friedberg:
## WHAT
Fix a bug that allowed school and district admins to create users with invalid emails.

## WHY
We had an issue with a teacher who had two Quill accounts with the same email. Further investigation helped us realize that the second account was created via this method, along with a teacher who was created with a space in their email address. We do not want this to happen.

## HOW
The initial creation call was failing, but because we immediately called `refresh_token!` afterwards, which includes a `save(validate: false)` call, the account was always being saved even if it was invalid. I fixed this loophole so we return after the save attempt if there are errors, and updated the frontend to actually show users these errors.

### Screenshots
Screenshot 2024-11-15 at 1 19 31 PM
Screenshot 2024-11-15 at 1 19 07 PM

### Notion Card Links
https://www.notion.so/quill/Can-archived-classes-that-were-rostered-in-a-teacher-s-account-that-was-deleted-be-restored-13fd42e6f94180ef83d8cd45856dc156?d=13fd42e6f9418051922c001cc64cb6f9

Tested both invalid and valid workflows locally.

PR Checklist | Your Answer
------------ | -------------


anathomical (approved):
Good catch!  I imagine it took some investigation to chase that down.

brendanshean (approved):
Great job tracking this error done.  I could see it being very tricky to find.

PR #12563, opened 2024-11-15
GenAI Labeling Confidence limit backfill

brendanshean:
## WHAT
Backfill some automl responses as Evidence::LabeledEntry records for use in GenAI Labeling RAG queries

## WHY
Filling out the Evidence::LabeledEntry with more data improves accuracy for some activities

## HOW
Previous experiments found optimal confidence_limit for specific prompt_id.  Use those values to get samples to populate Evidence::LabeledEntry

### Screenshots

### Notion Card Links
https://www.notion.so/quill/Clean-LabelEntry-data-on-Production-13dd42e6f941808cb71aca7893eea4cd?pvs=4

I have done the backfill on staging and observed an improvment in accuracy.

PR Checklist | Your Answer
------------ | -------------
Have you added and/or updated tests? | No.  This is a rake task that generates data.


services/QuillLMS/app/models/feedback_history.rb
```diff
@@ -131,6 +131,19 @@ def self.suboptimal_sample(prompt_id:, confidence_limit: 0.90, max_length: 100,
...
   end
 
+  def self.entry_label_sample(prompt_id:, confidence_limit: 0.80, max_length: 100, limit: 2_000)
+    autoML
+      .joins('INNER JOIN comprehension_rules on comprehension_rules.uid = feedback_histories.rule_uid')
+      .joins('INNER JOIN comprehension_labels ON comprehension_rules.id = comprehension_labels.rule_id')
+      .for_prompt(prompt_id)
+      .confidence_greater_than(confidence_limit)
+      .entry_shorter_than(max_length)
+      .order(Arel.sql("CAST(metadata->'api'->'confidence' AS DOUBLE PRECISION) DESC"))
```
brendanshean, line 141: Arel needed here due to:
```
Dangerous query method (method whose arguments are used as raw SQL) called with non-attribute 
argument(s): "CAST(metadata->'api'->'confidence' AS DOUBLE PRECISION) DESC".  This method 
should not be called with user-provided values, such as request parameters or model attributes. 
Known-safe values can be passed by wrapping them in Arel.sql(). 
(ActiveRecord::UnknownAttributeReference)
```

dandrabik: Interesting. Take or leave, you might consider making this a scope, since we might want to reuse this, e.g. `order_by_confidence_desc` or `order_by_confidence("DESC")

dandrabik (approved):
LGTM

services/QuillLMS/app/models/feedback_history.rb
brendanshean: I'll update to a scope if we end up reusing.

PR #12565, opened 2024-11-18
Switch geocoder database to API lookup

cissyxyu:
## WHAT
Remove the IP Address Geocoder folder and replace the IP lookup with abstract API.

## WHY
The geocoder database is taking up too much memory and we can be more efficient by doing an IP lookup with Abstract API.

## HOW
The Geocoder gem allows us to switch out the manual ip lookup with an abstract API lookup -- just tweak the Geocoder configurations and add the API key in the ENV config.

### Screenshots

### Notion Card Links
https://www.notion.so/quill/Needs-details-Convert-Geocoder-to-use-API-service-c516192d8d36471ab205efc6b09cabb2?pvs=4

Test the lookup using the Geocoder object and doing an IP address search in the rails console locally and on staging.

PR Checklist | Your Answer
------------ | -------------
Have you added and/or updated tests? |  NO, tests should still be passing because the lookup results should appear the same.

### DIFF
--- a/services/QuillLMS/app/assets/stylesheets/pages/for_administrators.scss
+++ b/services/QuillLMS/app/assets/stylesheets/pages/for_administrators.scss
@@ -288,8 +288,9 @@
       position: relative;
       width: 100%;
       height: 100%;
-      background-color: $quill-black;
+      background-color: $quill-white;
       max-height: 675px;
+      margin-top: 80px;
     }
   }
 


emilia-friedberg (approved):
The code here looks good to me, but I seem to remember that part of the reason we went with the locally stored db in the first place instead of the API was a rate-limiting issue, so I just want to confirm that the plan we have isn't going to pose any issues on that front.

brendanshean (approved):
Thank you for making this switch.  Nice to see a smaller slug size

anathomical (approved):

cissyxyu:
> The code here looks good to me, but I seem to remember that part of the reason we went with the locally stored db in the first place instead of the API was a rate-limiting issue, so I just want to confirm that the plan we have isn't going to pose any issues on that front.

We're using an api called Abstract this time - https://www.abstractapi.com/api/ip-geolocation-api. On the current plan we have, we can make 5 calls per second, which should be sufficient for our needs. We can also monitor the usage after deploy to confirm that things are under the rate limit.

PR #12575, opened 2024-11-19
Fix a proofreader bug involving incorrect word index.

brendanshean:
## WHAT
Fix a bug on staging involving proofreader's prompt labeling for concept results.

## WHY
An incorrect prompt can potentially be reported when the target word occurs in multiple sentences and the target sentence does not have the first occurrence of the target word.

## HOW
The original implementation of `findSentence` used a `wordIndex` parameter incorrectly.  Namely, it mixed up word indices with character indices.  For example, `indexOfLastWordInSentence += paragraphSentence.length` is tracking character indices, whereas `wordIndex` is tracking the particular index of a word in the current paragraph.

Update the findSentence to iterate word by word through paragraph and check to see if the targetWord is in a particular paragraph sentence but also that the wordIndex is within the bounds of the current paragraphSentence.

### Screenshots

### Notion Card Links
https://www.notion.so/quill/Proofreading-Future-Perfect-Squirrels-13ed42e6f9418075ad88f5306e530f88?pvs=4

I have tested the activity on staging and confirmed that the correct prompt is being reported now.

PR Checklist | Your Answer
------------ | -------------


services/QuillLMS/client/app/bundles/Proofreader/components/proofreaderActivities/container.tsx
brendanshean, line 1: Most of the diff here is just a linter fixing some indentation.  The only real update is the `findSentence` method and the test file.

emilia-friedberg (approved):
Good catch -- that looks like it was tricky to find. Thank you for fixing this.

PR #12578, opened 2024-11-19
Remove weird zombie migrations with no corresponding file

anathomical:
## WHAT
These migration records ended up in `db/structure.sql` because my local database had references to them in `schema_migrations`, but there are no corresponding files in `db/migrations`, so we don't actually need these records since there's no way for anyone else to get them.
## WHY
We want our `db/structure.sql` to reflect our actual migration records, and be unified across all developers.
## HOW
Manually remove the zombie `schema_migrations` records in my local db and run `rails db:migrate` to re-generate `db/structure.sql`

PR Checklist | Your Answer
------------ | -------------
Have you added and/or updated tests? | No, this is just some administrative clean-up
Have you deployed to Staging? | No, not a code change


happythenewsad (approved):
🙏

brendanshean (approved):

anathomical:
After some discussion with @brendanshean , and a look into the production database where I've confirmed that these records do exist in the `schema_migrations` table, I think it's actually better to try to recover the missing migration files rather than remove the records of the migrations since they do appear to have been run on production and are not a weird artifact of local development inconsistencies.

I've written up a task card for that work here: https://www.notion.so/quill/Attempt-to-recover-missing-db-migrate-files-that-have-entries-in-schema_migrations-but-are-not-in-th-144d42e6f9418011920de906ea253453

PR #12582, opened 2024-11-20
llm prompt templates page

emilia-friedberg:
## WHAT
Add the LLM Prompt Templates index page and a modal to view the template and edit the notes.

## WHY
This is the last part of the spec to gain feature parity with the ERB tooling Brendan built.

## HOW
Just build out an index view and the aforementioned modal, along with the backend routing necessary to support this functionality. I moved some modals that are now shared into the `shared/` folder as well.

### Screenshots
Screenshot 2024-11-20 at 2 57 04 PM
Screenshot 2024-11-20 at 2 57 16 PM
Screenshot 2024-11-20 at 2 57 27 PM
Screenshot 2024-11-20 at 2 57 39 PM

### Notion Card Links
https://www.notion.so/quill/LLM-Prompt-Templates-Page-143d42e6f941800f97d1e8545d9d3951?pvs=4

Tested all flows locally, and Jamie is QAing on staging.

PR Checklist | Your Answer
------------ | -------------


services/QuillLMS/engines/evidence/app/controllers/evidence/research/gen_ai/llm_prompt_templates_controller.rb
```diff
@@ -15,9 +15,21 @@ def index
...
-                llm_prompt_templates: @llm_prompt_templates
-              }
+              if include_trial_count_and_formatted_contents
+                render json: {
+                  llm_prompt_templates: @llm_prompt_templates.map do |llm_prompt_template|
+                    colored_contents = ColoredContentsBuilder.run(contents: llm_prompt_template.contents)
+                    serialized_llm_prompt_template = llm_prompt_template.serializable_hash(methods: [:num_trials])
+                    serialized_llm_prompt_template['colored_contents'] = colored_contents
+                    serialized_llm_prompt_template['previewed_contents'] = PreviewedContentsBuilder.run(colored_contents:, llm_prompt_template:)
+                    serialized_llm_prompt_template
```
anathomical: Take or leave this since it's purely stylistic, but I'd do this with with `merge` instead of modifying a temporary variable:
```
llm_prompt_template.serializable_hash(methods: [:num_trials]).merge({
  'colored_contents' => colored_contents,
  'previewed_contents' => PreviewedContentsBuilder.run(colored_contents:, llm_prompt_template:)
})
```
Mostly I prefer this because I it feels awkward to me to create a temporary variable, modify it, and then pass it back at the end of the block.  Sometimes that's unavoidable, but if I can avoid it in a way that's not too crazy, I tend to try.

anathomical (approved):
Looks good to me!  I've got a small stylistic thing, but it's totally take-or-leave.

eadams17 (approved):
LGTM!

services/QuillLMS/engines/evidence/app/controllers/evidence/research/gen_ai/llm_prompt_templates_controller.rb
```diff
@@ -73,13 +85,40 @@ def show
...
+              else
+                render json: { errors: llm_prompt_template.errors, }, status: :unprocessable_entity
+              end
+            end
           end
         end
 
+        def bulk_update
+          ActiveRecord::Base.transaction do
+            params[:research_gen_ai_llm_prompt_templates].each do |llm_prompt_template|
```
brendanshean: Since `llm_prompt_template` reads like a record or object, I think `llm_prompt_template_params` would be more clear for this loop variable.


Also, I would use `find` and `update!` so that we don't have silent failures: `LLMPromptTemplate.find(llm_prompt_template_params[:id]).update!(updatable_params)`

Contrast in lines 92 and 100 the update calls don't fail silently due to the if/else blocks.
(In general we should avoid using `update` without a if/else construction)

brendanshean (approved):
Great work on this!

PR #12584, opened 2024-11-20
Get rid of zombie tables 🧟 

brendanshean:
## WHAT
Add in some missing early migration files for `comments` and `rules_misseds` tables
Add new migrations that drop these tables

## WHY
They're referenced in structure.sql and they are no longer in use.  The corresponding migration files were deleted at some point which makes their existence puzzling.

## HOW
I located four of the migration files, which were all from the comments table, using `git log --all --name-only --pretty=format: | grep '20130510221334'` 

I was unable to located the remaining migration file but I was able to figure out that it was rules_misseds based on some structure.sql diffs through another process.

### Screenshots
Before
Screenshot 2024-11-20 at 12 46 58 PM

After
Screenshot 2024-11-20 at 1 06 58 PM

### Notion Card Links
https://www.notion.so/quill/Attempt-to-recover-missing-db-migrate-files-that-have-entries-in-schema_migrations-but-are-not-in-th-144d42e6f9418011920de906ea253453?pvs=4

I have run these migrations on staging

PR Checklist | Your Answer
------------ | -------------


anathomical (approved):
Thanks for doing this investigation.  And for cleaning up the tables that we don't actually use anymore.  I wonder if there would be value in some day doing a more general audit for unused tables in the production database that don't have model definitions.

brendanshean:
> Thanks for doing this investigation. And for cleaning up the tables that we don't actually use anymore. I wonder if there would be value in some day doing a more general audit for unused tables in the production database that don't have model definitions.

Good idea.   I made a [card](https://www.notion.so/quill/Get-rid-of-potentially-unusued-tables-models-144d42e6f941806b9cfbc0ff68cce9dc?pvs=4).  It looks like there are six potential ones.

happythenewsad (approved):
🧹 🧹 🧹 

You probably already did this during QA, but for the record, `comments` is an empty table

PR #12607, opened 2024-11-27
Add Staggered release recalculate cron job

brendanshean:
## WHAT
~~Add a nightly cron job that recalculates staggered release locks for any activity sessions completed in the past 24 hours~~
Add an hourly cron job that recalculates staggered release locks for any activity sessions completed in the past 1 to 2 hours. 

## WHY
The teacher fix for recalculate staggered release locks is having to be called too often.  The hope is that nightly job will  cover some of the Activity Packs that are incorrectly locked.

~~I'm running this at hour == 22 since there doesn't seem to be anything else scheduled then.  Based on staging tests, I think should take about 10 minutes to complete.~~
I'm running this hourly since I think the load is doable.

## HOW
Query all completed activity sessions and join the classroom units table to get the associated classroom_id.  Then call all unique pairs of user_id, classroom_id on `SaveUserPackSequenceItemsWorker`

### Screenshots

### Notion Card Links
https://www.notion.so/quill/Significant-rise-in-use-of-staggered-release-unlock-fix-145d42e6f9418078a287d881c7b63ace?pvs=4

I tested the job on staging with ~40,000 (user_id, classroom_id) pairs.

PR Checklist | Your Answer
------------ | -------------


services/QuillLMS/spec/models/cron_spec.rb
```diff
@@ -24,22 +24,33 @@
   end
 
   describe '#interval_1_hour' do
+    before { Cron.instance_variable_set(:@t, nil) }
```
brendanshean, line 27: `@t` gets memoized over these tests which messes with the `travel_to` block.  This ensures that it gets reset each test.

services/QuillLMS/app/workers/recalculate_staggered_release_locks_for_recently_completed_activities_worker.rb
```diff
@@ -0,0 +1,32 @@
...
+
+  def perform
+    recently_completed_activity_sessions.each do |activity_session|
+      enqueue_worker(activity_session.user_id, activity_session.classroom_id)
+    end
+  end
+
+  private def enqueue_worker(user_id, classroom_id)
+    SaveUserPackSequenceItemsWorker.perform_async(user_id, classroom_id)
+  rescue StandardError => e
```
dandrabik: This `rescue` only rescues the job creation, which can't really fail. Maybe remove (unless I'm missing something)?

services/QuillLMS/app/models/cron.rb
```diff
@@ -89,6 +91,11 @@ def self.run_at_21_hour_mark
     AdminDiagnosticReports::PerformanceBenchmarkWorker.perform_async
   end
 
+  # 6/7PM depending on daylight savings
```
dandrabik: Is there a reason this needs to run specifically at 6pm, or could we move this to run once a day `def self.interval_1_day`? From the description you are trying to avoid other jobs, but I'm not sure we need to do that.

services/QuillLMS/app/models/cron.rb
```diff
@@ -89,6 +91,11 @@ def self.run_at_21_hour_mark
...
+  def self.run_at_22_hour_mark
```
dandrabik: Relatedly, if this takes 10 minutes for 24 hours of data, we might consider running this more often during the day, e.g. if this takes a minute to run for an hour of data, maybe we should just run it every hour and issues would fix quicker (not sure if the runtime scales that way).

dandrabik (approved):
Overall looks good. I added some notes about when this job is run and a rescue that are worth looking at, but I'll defer to your decision on those.

services/QuillLMS/app/workers/recalculate_staggered_release_locks_for_recently_completed_activities_worker.rb
brendanshean: No, you're right, that only helps with enqueuing jobs.

services/QuillLMS/app/models/cron.rb
brendanshean: I thought about this but was worried about adding unnecessary load to the system during peak hours.  However, it would probably be better from a user perspective since these lock issues could _potentially_  be fixed more quickly.  I will switch to hourly and monitor.

brendanshean: I've updated the query to look at an hourly range, but I've shifted the window from `2.hours...1.hours` rather than `1.hour...Time.current` since I don't want to be calling very recently completed ones twice.

https://github.com/empirical-org/Empirical-Core/pull/12607/files#diff-0624ca7f303f8e052d6a2c91703502f154e5dc027e59d4ccb6cbdacb19a0c5a2R15

services/QuillLMS/app/workers/recalculate_staggered_release_locks_for_recently_completed_activities_worker.rb
```diff
@@ -0,0 +1,18 @@
...
+  def perform
+    recently_completed_activity_sessions.each do |activity_session|
+      SaveUserPackSequenceItemsWorker.perform_async(activity_session.user_id, activity_session.classroom_id)
+    end
+  end
+
+  private def recently_completed_activity_sessions
+    ActivitySession
+      .joins(:classroom_unit)
+      .where(completed_at: 2.hours.ago..1.hour.ago)
```
anathomical: Given that the cron timing is a little fuzzy, should we broaden this window slightly in case we run five minutes off between calls and end up with a tiny gap that's not covered?

anathomical (approved):
Small question about the query window, but the approach seems like a good one to me, and I like this as a quick solution to the issue.

services/QuillLMS/app/workers/recalculate_staggered_release_locks_for_recently_completed_activities_worker.rb
brendanshean: Oh.  That's good point.  Will change this to `125.minutes.ago...55.minutes.ago`

PR #12616, opened 2024-12-02
User newsletter flag to segment

anathomical:
## WHAT
Send data about whether the user is signed up to receive Teacher Newsletters to Segment
## WHY
We use Segment to source user data for Ortto, and Ortto is used to send newsletters (at least I think that's how the chain of logic is set up).  So when we pull data from Segment to Ortto we want to make sure we respect user choices about receiving email.
## HOW
Pass the `user.send_newsletter` value along to Segment when we make `identify` calls.

### Notion Card Links
https://www.notion.so/quill/Add-teacher-newsletter-subscription-to-the-Segment-Identify-call-12ef535045de4828b9dd28fb579ef418

- Deploy to staging
- Log in with a known user
- Confirm that the `/i` call to Segment includes the new attribute `subscribedToTeacherNewsletter`
- Confirm that the value in the new attribute matches the one assigned to the user in the database

PR Checklist | Your Answer
------------ | -------------


brendanshean (approved):
Thank you for fixing this!

happythenewsad (approved):

emilia-friedberg (approved):

PR #12626, opened 2024-12-04
prompt activation frontend

emilia-friedberg:
## WHAT
Build out the frontend and controller endpoints to make it possible to activate a prompt.

## WHY
This is what will allow the curriculum team to pick which LLM Prompt actually gets used to serve feedback.

## HOW
Pretty straightforward React and Rails code! We also changed the styling of some other components to make room for the new columns.

### Screenshots
Screenshot 2024-12-04 at 12 55 18 PM
Screenshot 2024-12-04 at 12 55 13 PM
Screenshot 2024-12-04 at 12 55 04 PM
Screenshot 2024-12-04 at 12 54 53 PM

### Notion Card Links
https://www.notion.so/quill/Activate-A-Winning-Prompt-146d42e6f941806dabc3e98fae6fadf2?pvs=4

Tested locally and on staging; Jamie will also be QAing on staging.

PR Checklist | Your Answer
------------ | -------------


services/QuillLMS/client/app/bundles/Shared/components/shared/checkBox.tsx
```diff
@@ -61,7 +61,7 @@ export const Checkbox = ({ label, mode=LIGHT, onClick, state, selected, id, }: c
   return (
     <div className="checkbox-container">
       {renderCheckbox()}
-      <label className={labelClass} htmlFor={checkboxId}>{label}</label>
+      {label ? <label className={labelClass} htmlFor={checkboxId}>{label}</label> : null}
```
anathomical: This is purely stylistic, and I don't even know if it's better, but thoughts on using
```javascript
{label && <label className={labelClass} htmlFor={checkboxId}>{label}</label>}
```
instead of a ternary?  That may not actually do exactly the same thing, but I feel like I've seen us use that pattern in a couple of places, and I personally find it slightly more readable.  But, again, it's just a style thing, so not a big deal either way.

services/QuillLMS/client/app/bundles/Staff/components/evidence/llmPromptDatasets/individualDataset/__tests__/activateLLMPromptsSection.test.tsx
```diff
@@ -0,0 +1,127 @@
...
+
+    const { asFragment } = render(
+      <QueryClientProvider client={queryClient}>
+        <ActivatedLLMPromptsSection {...sharedProps} />
+      </QueryClientProvider>
+    );
+
+    expect(screen.getByText(/Activated Prompts \(All Time\)/i)).toBeInTheDocument();
+    expect(screen.getByText(/Prompt Name/i)).toBeInTheDocument();
+    expect(screen.getByText(/Dataset 1.0/i)).toBeInTheDocument();
```
anathomical: Are these values hard-coded into the template, or are they based on some of the data defined in `activated_llm_prompts`?  If they're hard-coded, leaving this as-is makes sense, but if they're based on values in the input object, I think it would be clearer to have them referenced by variable name.  Such as
```javascript
expect(screen.getByText(/${activatedPromptsResponse. activated_llm_prompts[0]. trial.llm_prompt.name}/i)).toBeInTheDocument();
```
Or maybe something more readable, but something that makes it clear that the input should impact the output.

services/QuillLMS/engines/evidence/app/models/evidence/prompt.rb
```diff
@@ -84,6 +84,12 @@ def active_llm_prompt
         .first
     end
 
+    def active_activated_llm_prompt
+      activated_llm_prompts
+        .merge(Evidence::ActivatedLLMPrompt.active)
```
anathomical, line 101: I like the use of `merge` here to apply a scope.

services/QuillLMS/engines/evidence/spec/controllers/evidence/activated_llm_prompts_controller_spec.rb
```diff
@@ -0,0 +1,80 @@
...
+      it 'returns a successful response' do
+        get :index, params: { prompt_id: prompt.id }
+
+        expect(response).to be_successful
+        expect(json_response['activated_llm_prompts']).to be_present
+        expect(json_response['activated_llm_prompts'].first['id']).to eq(activated_llm_prompt.id)
+      end
+
+      it 'includes associated trial data' do
+        trial = create(:evidence_research_gen_ai_trial, llm_prompt:)
```
anathomical: This is purely stylistic, so take or leave it.  I prefer to use `let` or `let!` whenever a `create` is called, but you can't call `let` inside an `it` block, so I'd structure this with a `context` wrapper.
```ruby
context 'with associated trial data' do
  let!(:trial) { create(:evidence_research_gen_ai_trial, llm_prompt:) }

  it 'includes associated trial data' do
    ... # same code
  end
end
```

services/QuillLMS/engines/evidence/spec/controllers/evidence/activated_llm_prompts_controller_spec.rb
```diff
@@ -0,0 +1,80 @@
...
+        activated_llm_prompt.update!(llm_prompt:)
```
anathomical: Does this doe anything?  The `create` call on line 24 looks like it already sets `llm_prompt` to the same value as this.

services/QuillLMS/engines/evidence/spec/controllers/evidence/activated_llm_prompts_controller_spec.rb
```diff
@@ -0,0 +1,80 @@
...
+          expect {
+            post :create, params: valid_params
+          }.to change(Evidence::ActivatedLLMPrompt, :count).by(1)
+
+          expect(response).to have_http_status(:ok)
+          expect(json_response['id']).to be_present
+        end
+
+        it 'sets other prompts to inactive if active is true' do
+          existing_prompt = create(:evidence_activated_llm_prompt, llm_prompt:, prompt:, active: true)
```
anathomical: Again, take or leave style thing, but I like `create` calls in `let` which means you'd need a `context` wrapper.

anathomical (approved):
This looks good to me.  I've got a couple of style suggestions, but they're all take-or-leave items, so you could ship this as-is.

services/QuillLMS/engines/evidence/app/controllers/evidence/activated_llm_prompts_controller.rb
```diff
@@ -0,0 +1,29 @@
...
+      activated_llm_prompt = prompt.activated_llm_prompts.build(activated_llm_prompt_params)
+
+      if activated_llm_prompt.save
+        render json: { id: activated_llm_prompt.id }
+      else
+        render json: { error: activated_llm_prompt.errors }, status: 400
+      end
+    end
+
+    def prompt
```
brendanshean: I would make this method private since I don't think we need it exposed outside the controller..  In general, I try to make the only public methods in a controller correspond to specified routes.

brendanshean (approved):
So exciting to see some of these final pieces!  Awesome work

services/QuillLMS/client/app/bundles/Staff/components/evidence/llmPromptDatasets/individualDataset/activatedLLMPromptsSection.tsx
```diff
@@ -0,0 +1,203 @@
...
+      width: '55px',
+      rowSectionClassName: 'allow-wrap',
+      noTooltip: true
+    },
+    {
+      name: 'Prompt Template',
+      attribute: 'llmPromptName',
+      width: '72px',
+      headerClassName: 'center-content',
+      rowSectionClassName: 'center-content allow-wrap',
```
eadams17, line 77: Just an idea... I think at some point down the road, we should probably update Backpack to include these different class instances so we have them all documented somewhere

eadams17 (approved):
This looks great! Thanks for putting this together

PR #12631, opened 2024-12-04
Evidence GenAI OnTopic Detector

dandrabik:
## WHAT
Write a 'OnTopic' detector to be used by Evidence to determine if an entry is anywhere near the subject at hand. Add a test data set and the ability to run experiments in a script.

Note, there will still be iteration on the template and dataset, but I don't see the underlying code changing much.
## WHY
Completely Off Topic entries are leading to some "bonkers" GenAI feedback (e.g., it sometimes "yes, ands" nonsense or feels to write about what the student wrote when it's irrelevant, etc.), when we really just want the student to delete their entry and start again. This piece might also be usable for AutoML activities as a distinct check.
## HOW
Do the following:
1. Create a test data set (Rebekah is working on a more refined one)
2. Create a PromptBuilder
3. Create a Checker service that returns true/false
4. Write a test harness to print overall results and save detailed results to a csv.

### Screenshots
Initial Results with the passage and few shot examples
Note, We are going to refine the off-topic dataset (pulled from "Label_0" data) because many of those "failures" are actually on-topic for the passage, e.g. they discuss items in the passage but they are far away from the conjunction answer. So that number should go up significantly.
Screenshot 2024-12-04 at 3 56 23 PM

### Notion Card Links
https://www.notion.so/quill/Miscellaneous-Entry-Detector-145d42e6f9418008a9e4e8017dbe5844

I've created an initial dataset and a script to run tests. I've run multiple tests and feel this general code approach will work, we just need a refined dataset to test on and some more iteration on the template.

PR Checklist | Your Answer
------------ | -------------
Have you added and/or updated tests? |   'YES'


services/QuillLMS/engines/evidence/app/services/evidence/gen_ai/on_topic/checker.rb
```diff
@@ -0,0 +1,36 @@
...
+        def initialize(prompt:, entry:)
+          @prompt = prompt
+          @entry = entry
+        end
+
+        # default to true for malformed responses, e.g. nil, missing key
+        def run
+          response = llm_response[KEY_ON_TOPIC]
+
+          response.nil? ? true : response
```
anathomical: This is a take-or-leave stylistic thing, but I think the following code does the same thing and is (to me) more readable:
```ruby
llm_response.fetch(KEY_ON_TOPIC, true)
```

services/QuillLMS/engines/evidence/app/services/evidence/gen_ai/on_topic/data_fetcher.rb
```diff
@@ -0,0 +1,57 @@
...
+          @on_topic = on_topic
+          @limit = limit
+          @file = file || DEFAULT_FILE
+        end
+
+        def run
+          data
+            .select { |i| i.type == type }
+            .select { |i| prompt_id.nil? ? true : i.prompt_id == prompt_id }
+            .select { |i| on_topic.nil? ? true : i.on_topic == on_topic }
```
anathomical, line 35: I like the readability of making multiple `select` calls here.  Are we at all concerned about possible performance impacts of iterating over potentially large sets of data more than once?  Like, if `data` is a million records, and half of them are a `type` match, the second `select` will increase the number of processed rows by 50%.

Given that we're building the CSV data for testing I doubt these will be that big, and the readability of this approach is superior, but I wanted to double-check.

services/QuillLMS/engines/evidence/spec/services/evidence/gen_ai/on_topic/checker_spec.rb
```diff
@@ -0,0 +1,102 @@
...
+  describe 'private methods' do
+    it 'uses the correct chat API model' do
+      expect(subject.send(:model)).to eq(Evidence::Gemini::Chat::SMALL_MODEL)
+    end
+
+    context 'system_prompt' do
+      it 'calls PromptBuilder with the correct prompt' do
+        expect(Evidence::GenAI::OnTopic::PromptBuilder)
+          .to receive(:run)
+          .with(prompt: prompt)
```
anathomical: Purely stylistic, but you can use the implicit value here: `.with(prompt:)`

services/QuillLMS/engines/evidence/spec/services/evidence/gen_ai/on_topic/checker_spec.rb
```diff
@@ -0,0 +1,102 @@
...
+      before do
+        allow(Evidence::Gemini::Chat)
+          .to receive(:run)
+          .once
+          .and_return(expected_response)
+      end
+
+      it 'memoizes the LLM response' do
+        2.times { subject.send(:llm_response) }
+      end
```
anathomical, line 77: Nice approach here.  I didn't realize that `allow(...).once` would error if you tried to call it more than once.

services/QuillLMS/engines/evidence/spec/services/evidence/gen_ai/on_topic/data_fetcher_spec.rb
```diff
@@ -0,0 +1,148 @@
...
+    context 'with optional parameters' do
+      let(:fetcher) { described_class.new(type: 'test', prompt_id: 123, on_topic: true, limit: 5, file: 'custom.csv') }
+
+      it 'sets all attributes correctly' do
+        expect(fetcher.type).to eq('test')
+        expect(fetcher.prompt_id).to eq(123)
+        expect(fetcher.on_topic).to be(true)
+        expect(fetcher.limit).to eq(5)
+        expect(fetcher.file).to eq('custom.csv')
+      end
```
anathomical: This is a take-or-leave item since I think my approach here is a lot more verbose, and may not be worth it, but I'm always nervous when there's a hard-coded value in setup like a `let` that's then hard-coded again in an `expect`.  I think it's more clear that the input and output are connected when it's explicit:
```ruby
let(:type) { 'test' }
...
let(:fetcher) { described_class.new(type:, ...) }

it ...
  expect(fetcher.type).to eq(type)
  ...
end
```

services/QuillLMS/engines/evidence/spec/services/evidence/gen_ai/on_topic/data_fetcher_spec.rb
```diff
@@ -0,0 +1,148 @@
...
+    end
+
+    context 'filtering by type' do
+      let(:fetcher) { described_class.new(type: 'test') }
+
+      it 'returns only test type entries' do
+        results = fetcher.run
+        expect(results.size).to eq(3)
+        expect(results.map(&:type)).to all(eq('test'))
+      end
```
anathomical: Same as my above note about hard-coding a constant in two different places.

services/QuillLMS/engines/evidence/spec/services/evidence/gen_ai/on_topic/data_fetcher_spec.rb
```diff
@@ -0,0 +1,148 @@
...
+      it 'creates an Item with correct attributes' do
+        item = fetcher.send(:item_from_row, csv_row)
+
+        expect(item).to be_a(described_class::Item)
+        expect(item.prompt_id).to eq(42)
+        expect(item.entry).to eq('Test Entry')
+        expect(item.on_topic).to be(true)
+        expect(item.type).to eq('test')
+      end
+    end
```
anathomical, line 144: Same take-or-leave note about hard-coding constants in two places.

anathomical (approved):
This looks good to me.  As someone who doesn't read all the GenAI code, it's nice to see the patterns and extensibility that we've established.  I like what we've got going here.  I've got some take-or-leave style notes, but this could go out as-is as far as I can tell.

services/QuillLMS/engines/evidence/app/services/evidence/gen_ai/on_topic/checker.rb
dandrabik: Yeah, that's better. Thanks!

services/QuillLMS/engines/evidence/app/services/evidence/gen_ai/on_topic/data_fetcher.rb
dandrabik: Fair concerns if this was a large set, but this is for offline testing in which the dataset is a max 1,200 records. Scaling isn't a concern here.

services/QuillLMS/engines/evidence/spec/services/evidence/gen_ai/on_topic/data_fetcher_spec.rb
dandrabik: Fair, I'll clean these up a bit.

brendanshean (approved):
This looks great.  The code was very easy to follow.

PR #12638, opened 2024-12-09
Evidence CMS general updates

emilia-friedberg:
## WHAT
Updates to the Evidence CMS to hide AutoML columns when the `ai_type` is not `auto_ml`, and fix a bug where filtering rules by conjunction didn't work.

## WHY
These are parts of cards 9 and 10 that can be completed before the new rules themselves are in place.

## HOW
Mostly just React updates. I did take the opportunity to add more extensive tests to the components I was changing, and also set the timezone for all jest tests so I stop encountering mismatched snapshots between my local env and CircleCI (because we have historically gotten around this by mocking the time in individual tests, it only updated a small handful of snapshots).

### Screenshots

### Notion Card Links
https://www.notion.so/quill/GenAI-Rules-153d42e6f94180679b48f774351ac63e?pvs=4#157d42e6f941808d8e56da0069d8d430

Tested on staging, and Jamie will as well

PR Checklist | Your Answer
------------ | -------------


services/QuillLMS/client/package.json
```diff
@@ -181,7 +181,7 @@
     "jscs": "jscs --verbose .",
     "jscs:fix": "jscs --verbose -x . --silent",
     "test": "npm run build:test && rspec",
-    "jest": "node --expose-gc ./node_modules/.bin/jest --maxWorkers=1 --logHeapUsage",
+    "jest": "TZ=utc node --expose-gc ./node_modules/.bin/jest --maxWorkers=1 --logHeapUsage",
```
anathomical, line 184: Good call on this update.

services/QuillLMS/client/app/bundles/Staff/components/evidence/configureRules/rules.tsx
```diff
@@ -59,11 +59,12 @@ const Rules = ({ activityId, history, prompt }: RulesProps) => {
...
       );
-      setSortedRules(multiSortRules);
+      const filteredRules = prompt ? multiSortRules.filter(rule => rule.prompt_ids.includes(prompt.id)) : multiSortRules
+      setSortedAndFilteredRules(filteredRules);
     }
-  }, [rulesData]);
+  }, [rulesData, prompt]);
 
-  const formattedRows = sortedRules && sortedRules.map((rule: RuleInterface, i: number) => {
+  const formattedRows = sortedAndFilteredRules && sortedAndFilteredRules.map((rule: RuleInterface, i: number) => {
```
anathomical: I see that you're just modifying the old code pattern, but I think we could clean this up with a safe-accessor.
```javascript
const formattedRows = sortedAndFilteredRules?.map((rule: RuleInterface, i: number) => {
```

services/QuillLMS/client/app/bundles/Staff/components/evidence/configureRules/rulesIndexRouter.tsx
```diff
@@ -64,13 +64,13 @@ const RulesIndexRouter = ({ history, match }) => {
       <Switch>
         <Redirect exact from='/activities/:activityId/rules-index' to='/activities/:activityId/rules-index/all' />
         {/* eslint-disable-next-line react/jsx-no-bind */}
-        <Route path='/activities/:activityId/rules-index/all' render={() => <Rules activityId={activityId} history={history} prompt={getPromptForComponent(activityData, ALL)} />} />
+        <Route path='/activities/:activityId/rules-index/all' render={() => <Rules activityId={activityId} history={history} />} />
         {/* eslint-disable-next-line react/jsx-no-bind */}
-        <Route path='/activities/:activityId/rules-index/because' render={() => <Rules activityId={activityId} history={history} prompt={getPromptForComponent(activityData, BECAUSE)} />} />
+        <Route path='/activities/:activityId/rules-index/because' render={() => <Rules activityId={activityId} history={history} prompt={getPromptArrayForComponent(activityData, BECAUSE)[0]} />} />
```
anathomical: I don't actually have a better suggestion here, but I will say that it was not obvious to me why we'd want to reference by `[0]` on first read, so I can imagine being confused again next time I look at this.  Still, it seems reasonable enough to me given the limited scope here and the lack of an obvious (to me) alternative that doesn't feel too heavy (like a new helper function that includes the `[0]` internally, which feels awkward, too).

anathomical (approved):
This looks good.  I like a lot of the stream-lining you've done here, and with one exception, it was all super-intuitive to read through and understand.

services/QuillLMS/client/app/bundles/Staff/components/evidence/configureRules/rulesIndexRouter.tsx
brendanshean: I would recommend adding a helper method `getPromptForComponent` back into promptHelpers.  Seeing the `[0]` in the code begs the question, what are the other elements and do I need to know about them.  The helper method can explain why you're pulling the first element.

brendanshean (approved):
I really appreciate the testing cleanup here.

eadams17 (approved):

PR #12647, opened 2024-12-10
UI updates/ Launch Lessons page

eadams17:
## WHAT
design overhaul for the Launch Lessons page

## WHY
this is a very old page and was due for a UI makeover

## HOW
a lot of React and CSS refactoring

### Screenshots
Screen Shot 2024-12-10 at 6 55 47 PM

### Notion Card Links
https://www.notion.so/quill/Update-UI-Launch-Lessons-ba7388f4d749439ab8c1ce22c09a5dc2

checked on staging that lessons data render as expected and various states are represented as expected

PR Checklist | Your Answer
------------ | -------------


services/QuillLMS/app/assets/stylesheets/pages/my_lessons.scss
```diff
@@ -2,58 +2,30 @@
...
+      }
+      p {
+        margin: 7px 0px;
+        @include text-s;
+      }
+      span {
+        font-weight: 600;
+      }
+      a {
+        color: #027360;
```
emilia-friedberg: Maybe we could substitute this for a color variable?

services/QuillLMS/app/assets/stylesheets/pages/my_lessons.scss
```diff
@@ -71,6 +43,125 @@
...
+    h2 {
+      font-size: 20px;
+      font-weight: 700;
+      line-height: normal;
+    }
+
+    p {
+      margin-top: 8px;
+      font-size: 14px;
+      font-weight: 400;
```
emilia-friedberg: I think this could be replaced with `@include text-s`.

services/QuillLMS/client/app/bundles/Shared/components/shared/dataTableChip.tsx
```diff
@@ -19,6 +20,14 @@ export const DataTableChip = ({ color, icon, label, link }: DataTableChipProps)
       </a>
     )
   }
+  if (onClick) {
+    return (
+      <button className={`data-table-chip ${color} focus-on-light`} onClick={onClick}>
```
emilia-friedberg: Probably need a button type declaration here for the linter, right?

services/QuillLMS/client/app/bundles/Shared/components/shared/dataTableChip.tsx
```diff
@@ -19,6 +20,14 @@ export const DataTableChip = ({ color, icon, label, link }: DataTableChipProps)
...
+        {icon && <img alt={icon.alt} src={icon.src} />}
+        <p>{label}</p>
```
emilia-friedberg: I think a `span` is the more appropriate semantic element here -- I've been getting a lot of console errors about nesting block elements inside inline elements, which I think is technically not allowed under the HTML5 specification. There are some cases where it's tricky to avoid (like nesting elements in tables), but I think we should try where we can.

services/QuillLMS/client/app/bundles/Teacher/components/assignment_flow/manage_units/__tests__/unit.test.tsx
```diff
@@ -0,0 +1,79 @@
...
+  num_students_assigned: 20,
+  unitName: 'Sample Unit',
+  unitId: 'unit1',
+};
+
+describe('Unit Component', () => {
+  it('renders the unit name and assigned information', () => {
+    render(<Unit data={mockData} />);
+
+    expect(screen.getByText('Sample Unit')).toBeInTheDocument();
```
emilia-friedberg, line 46: This is something @anathomical often calls out, but I think that where something is a reference to a variable (as opposed to static text on the page), we should try to refer to it that way instead of just as a string, so it's clear what it should be testing.

services/QuillLMS/client/app/bundles/Teacher/components/assignment_flow/manage_units/unit.tsx
```diff
@@ -0,0 +1,229 @@
...
+        <p>Assigned to</p>
+        <span className="bolded-value">{`${studentCountData} ${Pluralize('Student', studentCountData)}`}</span>
+        <p>in</p>
+        <span className="bolded-value">{`${classroomsArray.length} ${Pluralize('class', classroomsArray.length)} (${classroomsArray.join(', ')})`}</span>
+      </div>
+    )
+  };
+
+  function lockedInfo () {
+    const firstCa = classroomActivities.values().next().value;
```
emilia-friedberg: I think we should just call this variable `firstClassroomActivity`.

services/QuillLMS/client/app/bundles/Teacher/components/assignment_flow/manage_units/unit.tsx
```diff
@@ -0,0 +1,229 @@
...
+    setShowModal(!showModal)
+  }
+
+  function getClassroomActivities() {
+    if (Object.keys(classroomActivities)) {
+      return classroomActivities
+    }
+    if (Object.keys(classroom_activities)) {
+      return classroom_activities
+    }
```
emilia-friedberg, line 108: Oy vey. My guess is you tried, but is it worth trying to track down why these are both sometimes present? Ideally, we'd only have one.

services/QuillLMS/client/app/bundles/Teacher/components/assignment_flow/manage_units/unit.tsx
```diff
@@ -0,0 +1,229 @@
...
+
+  function getActivityId(lessonActivity) {
+    const { activityId, activityUid, activity } = lessonActivity
+    return activityId || activityUid || activity?.uid
+  }
+
+  function getClassroomUnitId(lessonActivity) {
+    const { cuId, classroom_unit_id } = lessonActivity
+    return cuId || classroom_unit_id
+  }
```
emilia-friedberg, line 119: Same here. If it's just too complicated to deal with for this project that's fine, but might be worth noting in the PR description.

services/QuillLMS/client/app/bundles/Teacher/components/assignment_flow/manage_units/unit.tsx
```diff
@@ -0,0 +1,229 @@
...
+    }
+    if (started) {
+      return <a className={sharedButtonClassName} href={`${process.env.DEFAULT_URL}/teachers/classroom_units/${classroomUnitId}/launch_lesson/${activityId}`}>Resume Lesson</a>;
+    }
+    return <a className={sharedButtonClassName} href={`${process.env.DEFAULT_URL}/teachers/classroom_units/${classroomUnitId}/launch_lesson/${activityId}`}>Launch Lesson</a>;
+  }
+
+  function renderCustomizeCell(id) {
+    return(
+      <a className="customize-link focus-on-light" href={`/customize/${id}`} rel="noopener noreferrer" target="_blank">
```
emilia-friedberg: We should add an `aria-label` to these tags, because it won't be clear to a screenreader user what clicking that icon will do. Then, I'd test for the presence of an element with that aria label in the specs, instead of the alt text, which is less relevant for functionality.

services/QuillLMS/client/app/bundles/Teacher/components/assignment_flow/manage_units/unit.tsx
```diff
@@ -0,0 +1,229 @@
...
+        <img alt="extra small pencil icon" src={`${EXTRA_SMALL_ICON_BASE_SRC}/pencil.svg`} />
+      </a>
+    )
+  }
+
+  function renderDownloadCell(id) {
+    return(
+      <a className="download-link focus-on-light" href={`/activities/${id}/supporting_info`} rel="noopener noreferrer" target="_blank">
```
emilia-friedberg: See note above about aria labels.

services/QuillLMS/client/app/bundles/Teacher/components/assignment_flow/manage_units/unit.tsx
```diff
@@ -0,0 +1,229 @@
...
+        <img alt="extra small arrow down icon" src={`${EXTRA_SMALL_ICON_BASE_SRC}/arrow-down.svg`} />
+      </a>
+    )
+  }
+
+  function renderClassroomActivities() {
+    const lessonsIcon = <img alt="lessons icon" src={`${EXTRA_SMALL_ICON_BASE_SRC}/tool-lessons.svg`} />
+    const rows = [];
+    const activities = getClassroomActivities()
+    activities.forEach((lessonActivity) => {
```
emilia-friedberg, line 192: You could clean this up a little by doing `const rows = getClassroomActivities().map(lessonActivity => {` instead of pushing to the empty `rows` array.

emilia-friedberg (approved):
The CSS here looks great and the React clean-up and tests are much appreciated. I left some comments in places where I think we could take it a little further or make small changes to get us in line with best practices, but no need to review again. This page really needed a facelift and looks great now!

services/QuillLMS/app/assets/stylesheets/pages/my_lessons.scss
eadams17: good catch! with the refactoring, I was in a bit of autopilot for some parts of copy and pasting 😅

services/QuillLMS/client/app/bundles/Shared/components/shared/dataTableChip.tsx
eadams17: ok! I'll make that change. Thank you for the context

services/QuillLMS/client/app/bundles/Teacher/components/assignment_flow/manage_units/__tests__/unit.test.tsx
eadams17: so something like this? 

```
const unitNameSpan = screen.getByText('Sample Unit')
expect(screen.getByText(unitNameSpan)).toBeInTheDocument();
```

services/QuillLMS/client/app/bundles/Teacher/components/assignment_flow/manage_units/unit.tsx
eadams17: I think it has something to do with the DB values for this data sometimes having different shapes (it's lessons so I didn't want to dive too deep into that hole)

services/QuillLMS/client/app/bundles/Teacher/components/assignment_flow/manage_units/unit.tsx
eadams17: I'll add, thanks for the context!

services/QuillLMS/client/app/bundles/Teacher/components/assignment_flow/manage_units/unit.tsx
eadams17: so this data actually comes through as a `Map`, so we have to use forEach to iterate through it and then push that relevant data into the array (I added a comment in the code for better context)

PR #12650, opened 2024-12-11
Add highlight groups

brendanshean:
## WHAT
1. Add `Evidence::GenAI::HighlightGroup` model
1. Add `Evidence::GenAI::Highlight` model
1. Add `optimal_feedback` attribute to `Evidence::Prompt`

## WHY
1. A container is needed to group highlights for a specific type of secondary feedback
1. The existing Highlight model is connected to rules and feedback records, whereas the use case for GenAI highlights are connected more directly to the `Evidence::Prompt`.  Rather than use polymorphism or some kind of double duty with the existing model, I'm creating a new model.  Eventually, we can deprecate `Evidence::Highlight`.
1. This attribute is needed for GenAI integration.

## HOW
1.  Write migration and add model and test files
1.  Write migration file and add model and test files
1.  Write migration file

NB:  Schema annotations for `Evidence::GenAI::{Highlight, HighlightGroup}` require the evidence/gen_ai.rb `self.table_prefix_name` method to work properly.

### Screenshots

### Notion Card Links
https://www.notion.so/quill/Production-Evidence-GenAI-Check-144d42e6f94180268bebc8989c25c7a1?pvs=4

I ran migration on staging and then created an object in the rails console.

PR Checklist | Your Answer
------------ | -------------


services/QuillLMS/db/structure.sql
```diff
@@ -1612,12 +1612,13 @@ ALTER SEQUENCE public.comprehension_feedbacks_id_seq OWNED BY public.comprehensi
 
 CREATE TABLE public.comprehension_highlights (
     id integer NOT NULL,
-    feedback_id integer NOT NULL,
+    feedback_id integer,
```
dandrabik: Hmm, so you are reusing the current `Evidence::Highlight` class for both, and some will have a `feedback_id` and some will have a `highlight_group_id`. If you go this route, you'd have to add conditional logic (`feedback` or `highlight_group`) to [most methods in the highlight class](https://github.com/empirical-org/Empirical-Core/blob/develop/services/QuillLMS/engines/evidence/app/models/evidence/highlight.rb#L20-L110). So as part of this code, we'd need to make the highlight class methods still work in both setups (it'll be easy to write some failing tests for these).

Some paths you might consider:
1) Write one (or two) conditional methods that are used by other methods to point to the relevant object (feedback or rule or whatever)
2) Making the relationship polymorphic
3) A separate class for GenAI Highlights and copy the relevant code

dandrabik (changes requested):
Thanks for pushing this forward. So I left a comment with more details, but the general ask I have is to write tests for the methods in the Highlight class and make them work for both types of highlight associations. I suggested a few different paths that you could do, and will defer to you. Happy to discuss.

services/QuillLMS/engines/evidence/app/models/evidence/gen_ai.rb
brendanshean, line 1: Note this is needed to get the annotations working properly.

brendanshean, line 1: This is needed to get annotations working for `Evidence::GenAI::{Highlight, HighlightGroup}`

brendanshean, line 1: This is needed to get annotations working for Evidence::GenAI::{Highlight, HighlightGroup}

brendanshean, line 1: schema annotations for `Evidence::GenAI::{Highlight, HighlightGroup}` need this file to work properly.

brendanshean, line 1: schema annotations for Evidence::GenAI::{Highlight, HighlightGroup} need this file to work properly.

services/QuillLMS/db/migrate/20241211155124_create_evidence_gen_ai_highlights.evidence.rb
```diff
@@ -0,0 +1,17 @@
+# frozen_string_literal: true
+
+# This migration comes from evidence (originally 20241211154437)
+class CreateEvidenceGenAIHighlights < ActiveRecord::Migration[7.1]
+  def change
+    create_table :evidence_gen_ai_highlights do |t|
+      t.string :text, null: false
+      t.string :highlight_type, null: false
+      t.integer :starting_index
```
dandrabik: For these two columns, I wasn't sure about this, but I just confirmed that 100% of the data in the existing highlights table has highlight_type: 'passage' and starting_index: 0. We could probably drop these two columns for now and hard-code those values (and adjust later if necessary).

Just as an FYI, the other highlight_types: 'grammar' and 'spelling' are generated dynamically, i.e. which word to highlight, so they aren't in the DB. They are also the only ones to use `starting_index` and they use it because you could misspell the same word twice in your entry, but there was a feature request to only highlight the first misspelling (so as to not give the student multiple things to fix).

Screenshot 2024-12-12 at 1 55 45 PM

services/QuillLMS/engines/evidence/spec/models/evidence/gen_ai/highlight_spec.rb
```diff
@@ -0,0 +1,85 @@
...
+      it { expect(build(factory)).to be_valid }
+
+      it { should validate_presence_of(:text) }
+      it { should validate_length_of(:text).is_at_least(described_class::MIN_TEXT_LENGTH) }
+      it { should validate_length_of(:text).is_at_most(described_class::MAX_TEXT_LENGTH) }
+      it { should validate_presence_of(:highlight_type) }
+      it { should validate_inclusion_of(:highlight_type).in_array(described_class::TYPES) }
+      it { should validate_numericality_of(:starting_index).only_integer.is_greater_than_or_equal_to(0) }
+
+      describe '#valid_in_all_targets' do
```
dandrabik, line 30: Nice tests here!

dandrabik (approved):
So, I think you can drop two of the DB columns for now, but otherwise this looks good! Thanks for moving this along!

PR #12662, opened 2024-12-13
Small Fixes for Slow Tests

dandrabik:
## WHAT
Look for `create_list(:)` in test and make those lists a little shorter.
## WHY
Just general good testing hygiene. Testing the creation of 10 records vs. 2 records usually doesn't add more coverage.
## HOW
Search for `create_list(:`, reduce number if it's large, fix test if needed.

Ran tests where I adjusted them

PR Checklist | Your Answer
------------ | -------------


brendanshean (approved):
Good idea on using smaller numbers

happythenewsad (approved):

services/QuillLMS/spec/controllers/api/v1/classroom_units_controller_spec.rb
```diff
@@ -5,7 +5,8 @@
 describe Api::V1::ClassroomUnitsController, type: :controller do
   let(:other_teacher) { create(:teacher) }
   let(:classroom) { create(:classroom) }
-  let(:students) { create_list(:student, 5) }
+  let(:student_count) { 2 }
+  let(:students) { create_list(:student, student_count) }
```
anathomical, line 9: I also appreciate the extraction of the count into its own variable since it's used in comparisons below.

anathomical (approved):
I bet some of these don't even need lists for their tests, but I like just reducing counts as a quick win rather than doing nothing until someone can read through them all.

PR #12666, opened 2024-12-16
Integrate Evidence::GenAI::Check

brendanshean:
## WHAT
* Use the LLM prompt text (template and content) activated by the research tool
* Use highlights that are connected to the Evidence::Prompt
* Use optimal feedback connected to the Evidence::Prompt
* Update the thor task backfill for GenAI rules
* Add suborder attribute
* ~~Add in fixtures for `comprehension_rules`~~
* Add optimal and suboptimal rules for each prompt with their own unique uid.

## WHY
* We'd like to transition from the static prompts and use prompt performing well in trials.
* Highlights are no longer connected via feedback and labels
* Since we're no longer using labels, we need a canonical optimal feedback
* Curriculum has requested non-universal rules associated by conjunction for each prompt
* Curriculum has requested a specific order for the rules index
* ~~Since these rules have hardcoded uids that the logic expects, I've added them as fixtures so that they are loaded.  Still debating whether to make these [global](https://github.com/empirical-org/Empirical-Core/pull/12666/commits/88147f53737e9563a668f11070cbefc5b305e21c) vs [targeted](https://github.com/empirical-org/Empirical-Core/pull/12666/commits/8c3d2539169ff5e4829512d749858cb6a5c1934a)~~
* This was requested by PG

## HOW
* Update the PrimaryFeeback::PromptBuilder to use the ActiveLLMPrompt for a given prompt
* Pull highlights via HighlightGroups and use that text when providing SecondaryFeedback in the PromptBuilder
* Use `prompt.optimal_feedback` instead of `prompt.optimal_label_feedback`
* Iterate over each prompt and create a rule that matches the conjunction and optimal/suboptimal status.
* Assign the suborder based on the hard-coded `suborder_mapping`
* ~~Save `comprehension_rules.yml` in spec/fixtures for both engine and main app, then add `fixtures :comprehension_rules` at the top of tests that use it.~~
* Update the callback on prompt to create unique rules.

### Screenshots

### Notion Card Links
](https://www.notion.so/quill/Production-Evidence-GenAI-Check-144d42e6f94180268bebc8989c25c7a1?pvs=4)

I tested this on staging by playing through a test activity and eliciting optimal and suboptimal feedback including secondary feedback.

PR Checklist | Your Answer
------------ | -------------


services/QuillLMS/engines/evidence/app/models/evidence/prompt.rb
```diff
@@ -28,13 +28,15 @@ class Prompt < ApplicationRecord
 
     MIN_TEXT_LENGTH = 10
     MAX_TEXT_LENGTH = 255
-    CONJUNCTIONS = %w[because but so]
+    CONJUNCTIONS = %w[because but so].freeze
     DEFAULT_MAX_ATTEMPTS = 5
     MIN_MAX_ATTEMPTS = 3
-    MAX_MAX_ATTEMPTS = 6
```
emilia-friedberg, line 34: Why is this changing?

services/QuillLMS/lib/tasks/gen_ai_tasks.thor
```diff
@@ -422,27 +422,32 @@ class GenAITasks < Thor
...
           rule_type: Evidence::Rule::TYPE_GEN_AI,
-          optimal:,
+          optimal: is_optimal,
           state: 'active',
-          name: "GenAI universal #{conjunction} - #{optimal ? 'optimal' : 'suboptimal'}"
+          suborder: suborder_mapping[conjunction][is_optimal],
+          name: "GenAI #{conjunction} - #{is_optimal ? 'optimal' : 'suboptimal'}"
         )
+
+        Prompt.where(conjunction:).each { |prompt| rule.prompt_rules.create(prompt:) }
```
emilia-friedberg: Do we want to scope this to just prompts associated with a `GenAI` activity?

emilia-friedberg (approved):
Looks great! A couple of minor questions, but I think this is ready to go. I appreciate the thorough testing and cleanup of old tests, and it's so cool to see how all the pieces fit together!

services/QuillLMS/lib/tasks/gen_ai_tasks.thor
brendanshean: Yes, that's a good idea.

services/QuillLMS/engines/evidence/app/models/evidence/prompt.rb
brendanshean: That must be a typo.  Good catch.

services/QuillLMS/engines/evidence/app/services/evidence/gen_ai/primary_feedback/prompt_builder.rb
```diff
@@ -4,30 +4,12 @@ module Evidence
...
-
-        private def template_variables
-          {
-            passage:,
-            plagiarism_text:,
-            stem:,
-            optimal_examples:,
-            suboptimal_examples:
-          }
+        def initialize(prompt:, history: [])
```
dandrabik, line 9: Since we're not using templates and overwriting `initialize` and `run`, it might be better to inherit from `ApplicationService` directly to not pull in unused cruft.

services/QuillLMS/engines/evidence/app/models/evidence/gen_ai.rb
```diff
@@ -4,6 +4,24 @@ module Evidence
   module GenAI
     HistoryItem = Data.define(:user, :assistant)
 
+    CONCEPT_MAPPING = {
+      Evidence::Activity::BECAUSE_CONJUNCTION => 'qkjnIjFfXdTuKO7FgPzsIg',
+      Evidence::Activity::BUT_CONJUNCTION => 'KwspxuelfGZQCq7yX6ThPQ',
+      Evidence::Activity::SO_CONJUNCTION => 'IBdOFpAWi42LgfXvcz0scQ'
+    }
+
+    RULES_OPTIMAL = {
```
dandrabik: So, this is small thing, but when PG requested that these rules not be universal, I think (and I could be wrong) that he wanted a different rule per prompt that each pointed to the same 3 concepts rather than 6 total rules that are re-used. We probably should just clarify this. Personally I think the way you have it is cleaner, but I don't have strong enough feelings about if he has strong feelings about making a new set of rules for each prompt.

dandrabik (approved):
Nice work on this. I left one small refactor note and noted something that you should get clarification on (I wouldn't need to re-review, either direction is straight-forward).

services/QuillLMS/engines/evidence/app/models/evidence/gen_ai.rb
brendanshean: I confirmed with Peter that we want a different rule per prompt

services/QuillLMS/engines/evidence/app/models/evidence/prompt.rb
```diff
@@ -124,6 +132,30 @@ def optimal_label_feedback
         &.text
     end
 
+    def gen_ai_rule_attributes(is_optimal:)
```
dandrabik: Take or leave, you might consider extracting the generation of the array of 6 attribute hashes to a service (this code and the constants), e.g. `gen_ai_rule_attributes = GenAI::RuleAttributesGenerator.run(prompt)`.

dandrabik (approved):
Looks good, thanks for cleaning this up! One small refactor suggestion that you can take or leave.

services/QuillLMS/engines/evidence/app/models/evidence/prompt.rb
brendanshean: Good idea.

PR #12683, opened 2025-01-06
remove ip location worker files

emilia-friedberg:
## WHAT
Remove the call to the `IpLocationWorker` and the file itself.

## WHY
Peter Gault alerted me that we may need to change our billing information to `PointPin`, the API we were using in this worker. I did some investigation, and it seems like we never actually use the `IpLocation` model that gets created here for anything -- we do use the user's ip to determine their timezone in other parts of the codebase, but that comes out of the `ip_address` stored on the user record. Deleting this worker removes our reliance on an external API and saves us the small amount of money we were spending on its license. 

## HOW
Just delete the files and clean up the tests.

### Screenshots

### Notion Card Links
https://www.notion.so/quill/Not-Time-Sensitive-PiinPoint-Integration-170d42e6f94180709eb3ef2383d51dec?pvs=4

I deleted these files, then tested the new account flow on staging to confirm that the timezone still populates correctly. It does.

PR Checklist | Your Answer
------------ | -------------


happythenewsad (approved):
🧹 🧹 🧹

anathomical (approved):

PR #12691, opened 2025-01-08
Evidence Label Feedback - More explicit error handling

dandrabik:
## WHAT
OpenAI had some recent outages, and we hit a [500 error in an embeddings](https://quillorg-5s.sentry.io/issues/6205214818/?alert_rule_id=7252857&alert_type=issue&environment=production&notification_uuid=a3433890-1671-48d7-8fe1-1e0784df4f00&project=11238&referrer=slack) call. I've adjusted the code so we record the source of the error (optimal, no data). 
## WHY
Given our error handling in the `Check.run` code, this would have been rescued down the line (and stored as a generic 'something errored, move the student along'), but this code makes that logic more explicit and traceable for debugging purposes.
## HOW
Return `nil` if the API errors or timeouts and record the source of the optimal in the feedback_history metadata.

I've done some testing in the console and will test a little bit on staging. It involves replicating an external error, so I'm relying a bit of new automated tests here.

PR Checklist | Your Answer
------------ | -------------


brendanshean (approved):
Thank you for adding this much needed error handling.

services/QuillLMS/engines/evidence/spec/lib/evidence/open_ai/embedding_fetcher_spec.rb
```diff
@@ -44,10 +44,17 @@ module OpenAI
         end
       end
 
-      context 'error handling' do
+      context 'error handling for timeout' do
         it 'returns an empty array when encountering a timeout' do
```
anathomical: It probably doesn't matter, but is it worth updating the `it` text of this test and the new one copied below to refer to `nil` instead of `empty array`?

anathomical (approved):
Minor comment on `it` text, but this looks like a good approach for clarifying error chains to make debugging clearer.

happythenewsad (approved):

PR #12692, opened 2025-01-09
Add highlight_type to the Evidence::GenAI::Highlight table

emilia-friedberg:
## WHAT
Add `highlight_type` to the `Evidence::GenAI::Highlight` table.

## WHY
We want to be able to set this from the CMS, as we do for other highlights.

## HOW
Add a migration to add this column, then update the model file to use `highlight_type` the way the standard `Evidence::Highlight` does.

### Screenshots

### Notion Card Links
https://www.notion.so/quill/Gen-AI-CMS-Project-2-Embed-Trial-Runner-in-CMS-61c994a106854512b449ccc82be573c2?pvs=4#159d42e6f94180bba278fb6637f8ef07

No user-facing change right now, so I just ran the tests.

PR Checklist | Your Answer
------------ | -------------


anathomical (approved):

dandrabik:
One note on this change. This `highlight_type` for GenAI will always be `passage` and could be hardcoded (I believe that's why it's not there). If adding the column makes cross-compatibility with different components easier, that's fine, just wanted to flag we don't actually need its value for this purpose.

emilia-friedberg:
> One note on this change. This `highlight_type` for GenAI will always be `passage` and could be hardcoded (I believe that's why it's not there). If adding the column makes cross-compatibility with different components easier, that's fine, just wanted to flag we don't actually need its value for this purpose.

@dandrabik thank you for flagging that. Jamie's spec explicitly calls for the same options for this highlight as with other highlights, so I'll double check with her.

emilia-friedberg:
@dandrabik Jamie responded:

> Great question. Right now there's a drop down for all three in the semantic labels, but in the curriculum guides and in asking around with the curriculum team  we never change it to anything other than passage. I'm honestly not sure what we'd use the other two drop downs for. So in essence, that is accurate.
>
> I left it as is just for consistency in porting over, but I think you could move forward with only passage if that's easier or makes most sense. Passage is the only one we really need.

So I'll close this PR. Thank you for flagging.

PR #12693, opened 2025-01-09
improve search in activity library

emilia-friedberg:
## WHAT
Update the function we use to search the Activity Library to allow access to nested data.

## WHY
The curriculum team flagged in the OKR meeting that we were missing out on potentially valuable search terms, including the various tags that show beneath the activity. This was because our string conversion didn't work for nested items, so our "topics" weren't being included in the search.

## HOW
Just update this function to convert the hash to JSON rather than stringifying the `value` of each field directly.

### Screenshots
Screenshot 2025-01-09 at 12 20 18 PM

### Notion Card Links

Tested locally.

PR Checklist | Your Answer
------------ | -------------


eadams17 (approved):

anathomical (commented):
I'm not sure I understand why this results in any different behavior.  It believe that it does since you tested it, but I don't understand it.  After all, this is the difference between searching `"value 1 value 2 value 3"` and searching `'["value 1", "value 2", "value 3"]'`.  A search for `value 1` should successfully return both versions.  Is there a case that I'm not thinking of?

emilia-friedberg:
@anathomical it has to do with how `join()` handles nested hashes. You can see the behavior in the console with something like:

```
hash = { item: { nested: 'blah1', nested2: 'blah2' } }
Object.values(hash).join(' ')
-> '[object Object'] // obviously not what we want
JSON.stringify(Object.values(hash))
-> '[{"nested":"blah1","nested2":"blah2"}]' // much better
```

anathomical (commented):
I mentioned this in Slack, but your explanation clears things up for me, and I put my stamp of approval on this PR.

PR #12694, opened 2025-01-09
fill-missing-recommendation-sequences

anathomical:
## WHAT
Add code to identify records that should have recorded `StudentLearningSequence` rows, but have not for some reason.
## WHY
There's some sort of race condition that I haven't been able to figure out that causes 1 in about 300 records to just not record `StudentLearningSequence` rows on their first attempt.  The jobs complete, they just don't create the records.  Re-running the jobs with the same params has succeeded in creating desired rows in all known cases, so this just automates that process.
## HOW
Once an hour look back at the past two hours for records that should have created `StudentLearningSequence` rows but didn't and re-process those records.  Use a two hour window despite the one-hour interval to cover cases where a job may run with a gap of slightly more than 60 minutes due to Heroku fuzziness around cron timing.

### Notion Card Links
https://www.notion.so/quill/Placeholder-Address-issues-with-missing-assignment-data-150d42e6f94180e3b0bac573af60cb20?pvs=4

- Deploy to staging
- Enqueue job
- Confirm that it runs in less than a couple of minutes

PR Checklist | Your Answer
------------ | -------------


brendanshean (approved):
Nice use of inheritance here.  Very easy to follow code.

services/QuillLMS/spec/models/cron_spec.rb
```diff
@@ -44,6 +44,13 @@
       Cron.interval_1_hour
     end
 
+    it 'enqueues StudentLearningSequences::FillMissingWorker' do
```
happythenewsad, line 47: Slightly off-topic, but this whole file is essentially testing an external library's API.

services/QuillLMS/app/services/student_learning_sequences/fill_missing/post_diagnostics.rb
```diff
@@ -0,0 +1,29 @@
...
+              JOIN units ON classroom_units.unit_id = units.id
+              JOIN unit_activities ON classroom_units.unit_id = unit_activities.unit_id
+              LEFT OUTER JOIN student_learning_sequences
+                  ON classroom_units.id = student_learning_sequences.initial_classroom_unit_id
+                      AND student_learning_sequences.user_id = assigned
+              LEFT OUTER JOIN student_learning_sequence_activities
+                  ON student_learning_sequences.id = student_learning_sequence_activities.student_learning_sequence_id
+                      AND student_learning_sequences.initial_classroom_unit_id = student_learning_sequence_activities.classroom_unit_id
+                      AND student_learning_sequences.initial_activity_id = student_learning_sequence_activities.activity_id
+              WHERE unit_activities.activity_id IN (1774, 1814, 1818, 1664, 1669, 1680, 2538, 2540, 2542, 2551, 2557, 2564)
```
happythenewsad: Let's make the activity list a constant (we recently finished cleaning up the older references in the codebase last year)

happythenewsad (commented):
The underlying bug looks very difficult to diagnose, so I understand the motivation for this band aid. But I hope we can find the root cause eventually, as it doesn't feel good to leave the root cause dormant forever, while also adding this extra logic we have to maintain.

services/QuillLMS/spec/models/cron_spec.rb
anathomical: It's not even really doing that because `expect` means that the calls are never made.  What this file actually does, I think, is backstop us from accidentally deleting a registered job.  You're right that it's not actually testing anything.  It's a lot more like an `assert` call that will error if the `perform_async` calls aren't being made for some reason.

services/QuillLMS/app/services/student_learning_sequences/fill_missing/post_diagnostics.rb
anathomical: Good call.  I'll update.

happythenewsad (approved):

PR #12702, opened 2025-01-10
update the user cms search to allow the option to see all records

emilia-friedberg:
## WHAT
Update the CMS User Search with an option to see all records, and fix broken pagination.

## WHY
We had a broken system for pagination here, where we were never returning more than 30 results.

## HOW
I didn't want to always search for all records, because we might accidentally make really expensive searches that way, so I put a setting under advanced options to fetch all records. Also added tests, cleaned up file.

### Screenshots

### Notion Card Links
https://www.notion.so/quill/CMS-User-Search-Maxing-out-at-30-Records-153d42e6f94180f09ce4d93fb48d60d8?pvs=4

Tested locally.

PR Checklist | Your Answer
------------ | -------------


services/QuillLMS/spec/controllers/cms/users_controller_spec.rb
```diff
@@ -17,174 +17,82 @@
   describe '#index' do
     before { allow(RawSqlRunner).to receive(:execute) { ['results'] } }
 
-    it 'should assign the search query, the results, and number of spaces' do
+    it 'assigns the default record limit in the search query' do
       get :index
-      expect(assigns(:user_search_query)).to eq({ sort: 'last_sign_in', sort_direction: 'desc' })
+      expect(assigns(:user_search_query)).to include(record_limit: 30.0)
```
brendanshean: `30.0` here is a magic number.  I would add a 
`let(:record_limit) { described_class::USERS_PER_PAGE }` and then reference in this expectation.

services/QuillLMS/spec/controllers/cms/users_controller_spec.rb
```diff
@@ -17,174 +17,82 @@
...
       end
+    end
+  end
 
-      it 'should return results for partial email non-exact search' do
-        post :search, params: { user_email: '@testerson.com' }, as: :json
-
-        json = JSON.parse(response.body)
+  describe 'pagination-related tests' do
+    let!(:users) { create_list(:user, 35) }
```
brendanshean: 35 here is a magic number. I would add a
```
let(:record_limit) { described_class::USERS_PER_PAGE }
let(:num_new_users) { record_limit + 5 }
let!(:users) { create_list(:user, num_new_users) }
```

services/QuillLMS/spec/controllers/cms/users_controller_spec.rb
```diff
@@ -17,174 +17,82 @@
...
 
-        expect(json['userSearchQueryResults'].count).to be 1
-        expect(json['userSearchQueryResults'].first['id']).to eq(search_user.id)
-      end
+    it 'returns paginated results based on the default limit' do
+      get :search, params: { page: 1 }
+      json = JSON.parse(response.body)
+      expect(json['userSearchQueryResults'].size).to eq(30) # Default USERS_PER_PAGE
```
brendanshean: You can use the `record_limit` defined earlier.

services/QuillLMS/spec/controllers/cms/users_controller_spec.rb
```diff
@@ -17,174 +17,82 @@
...
-            'class_code' => class_code
-          }
-        }
-      )
-      expect(ChangeLog.last.action).to eq(ChangeLog::USER_ACTIONS[:search])
-      expect(ChangeLog.last.explanation).to include('class_code')
+    it 'returns all users if record_limit is set to "all"' do
+      get :search, params: { page: 1, record_limit: 'all' }
+      json = JSON.parse(response.body)
+      expect(json['userSearchQueryResults'].size).to eq(36) # users created here plus the one searching
```
brendanshean: `expect(json['userSearchQueryResults'].size).to eq(num_new_users + 1)`   # plus the one searching

brendanshean (approved):
Thank you for fixing the pagination so quickly

eadams17 (approved):
Thanks for fixing this and all of the refactoring!

services/QuillLMS/app/controllers/cms/users_controller.rb
```diff
@@ -271,9 +272,10 @@ def complete_sales_stage
       WHERE classrooms.code = #{sanitized_class_code}))"
   end
 
-  protected def pagination_query_string
-    page = [user_query_params[:page].to_i - 1, 0].max
-    "LIMIT #{USERS_PER_PAGE} OFFSET #{USERS_PER_PAGE * page}"
+  protected def limit_query_string
+    return '' if user_query_params[:record_limit].to_s.downcase == 'all'
+
+    "LIMIT #{user_query_params[:record_limit] || USERS_PER_PAGE}"
```
anathomical, line 278: It looks like this new query segment doesn't let you fetch anything other than the first page of results, but there's still a `page` param that's passable.  Probably doesn't really matter, but might be confusing if someone comes in later trying to figure out why fetching "page 2" just provides the data from page 1.

anathomical (approved):
A note about pagination, but this is probably fine as-is considering the limited scope.

PR #12703, opened 2025-01-13
append numbers to class codes

emilia-friedberg:
## WHAT
Start appending numbers to class codes.

## WHY
I think we ran out of combinations, and this is the quickest solution.

## HOW
Append a random number between 1 and 9.

### Notion Card Links
https://www.notion.so/quill/Teachers-cannot-create-new-classes-using-Google-Classroom-or-by-manually-creating-17ad42e6f94180f9884ecf1e483d9a81?pvs=4

Confirmed locally that we are now putting numbers on the ends of our class codes.

PR Checklist | Your Answer
------------ | -------------
Have you added and/or updated tests? | NO - still running


happythenewsad (approved):

anathomical (approved):
I wonder if we should just move to three word class codes, but numbers seem like a good quick fix to get things under control now.

PR #12705, opened 2025-01-13
NameGenerator - flip file order as stop gap

dandrabik:
## WHAT
This is an option that keeps the original code structure, but gives us more class codes (500K)
## WHY
To keep the original structure while we decide what the long-term structure is (which may be numbers).
## HOW
Flip order, write test to confirm it generates a good amount of new names.

Confirmed that it generates many new names. Ran the code in console.

PR Checklist | Your Answer
------------ | -------------
Have you added and/or updated tests? | Yes, added a benchmark test


emilia-friedberg (approved):
Good idea.

PR #12729, opened 2025-01-23
diagnostic insights email verification

emilia-friedberg:
## WHAT
Add email verification support for teachers in order to use Diagnostic Insights, and build out the basic page structure for Diagnostic Insights itself.

## WHY
This was an easy-to-isolate step of this project.

## HOW
Build out all of the requisite routing, etc for Diagnostic Insights, then update how we handle email verification on the backend to allow teachers to request verification from this page. I also added a rake task to backfill verification for third-party auth users and add it to new ones as part of the existing flow -- see discussion [here](https://www.notion.so/quill/RFC-for-Third-Party-Email-Verification-189d42e6f941805f8c4dd7860eab74f0?pvs=4).

### Screenshots
Screenshot 2025-01-28 at 3 40 22 PM

### Notion Card Links
https://www.notion.so/quill/Email-Verification-for-Diagnostic-Insights-182d42e6f94180009e20e5713bf3c968

Tested flows on staging, as well as older email verification flows to confirm they still work as expected. Peter Sharkey QAed.

PR Checklist | Your Answer
------------ | -------------


services/QuillLMS/config/routes.rb
```diff
@@ -313,6 +313,8 @@
     get 'progress_reports/landing_page' => 'progress_reports#landing_page'
     get 'progress_reports/activities_scores_by_classroom' => 'progress_reports#activities_scores_by_classroom'
     get 'progress_reports/real_time' => 'progress_reports#real_time'
+    get 'progress_reports/diagnostic_insights/*path' => 'progress_reports#diagnostic_insights'
```
brendanshean: I think you can combine these both using an optional route parameter:
`get 'progress_reports/diagnostic_insights(/*path)' => 'progress_reports#diagnostic_insights'`

services/QuillLMS/lib/tasks/users.rake
```diff
@@ -54,6 +54,19 @@ namespace :users do
...
   end
 
+  task mark_google_clever_teachers_verified: :environment do
+    User.left_outer_joins(:user_email_verification)
+      .where.not(clever_id: nil)
+      .or(User.where.not(google_id: nil)) # the position of `or` matters a lot when you have multiple `where` clauses
+      .where(role: User::TEACHER)
+      .where(user_email_verification: { id: nil })
+      .each do |user|
+      verification_method = UserEmailVerification::GOOGLE_VERIFICATION if user.google_id
```
brendanshean: I would indent 64-67 for readability.

brendanshean (commented):
Great work on this.  The screenshots look awesome. I really appreciate the thorough testing especially in teachers/progress_reports_controller_spec.rb

brendanshean (approved):

services/QuillLMS/app/controllers/teachers/progress_reports_controller.rb
```diff
@@ -152,4 +158,12 @@ def building_ai_knowledge_unit_templates
   private def recreate_staff_demo
     Demo::ReportDemoCreator.create_demo("hello+#{staff_demo_name}@quill.org")
   end
+
+  private def needs_email_verification_for_diagnostic_insights
+    # we don't require email verification for admins who were created prior to email verification, so they can access unless verification is set to pending
+    return current_user.email_verification_pending? if current_user.admin?
```
anathomical: Since this is a "teacher"-facing feature, even for admins, could we simplify this by requiring email verification for them, too, here?  I think the desire to avoid burdening admins with verifying if they were grandfathered in was to reduce friction for the admin reports which have a relatively small pool of potential users.  That logic may not apply here, and it'd be nice not to have to special case admins all over the place.

If we do want to maintain this distinction, I wonder if we can find a clean way to fold it into the `User` model.  Maybe modify `email_verified?` to return `true` for older admin users even if that's kind of a lie?

services/QuillLMS/app/controllers/verify_emails_controller.rb
```diff
@@ -58,4 +59,14 @@ def verify_by_token
   private def token_verification_params
     params.permit(:token)
   end
+
+  private def handle_verification_email_send(email_redirect_url = nil)
+    if current_user.user_email_verification
+      current_user.user_email_verification.set_new_token
+    else
+      current_user.require_email_verification
+    end
```
anathomical, line 68: There's an alternative syntax here that I'm not sure is actually more readable, so take or leave this based on how you think it reads, but you could do:
```ruby
current_user.user_email_verification&.set_new_token || current_user.require_email_verification
```
In Ruby, the `||` operator will only evaluate the right side if the left side is false-y.  In this case, you'd expect that if the safe-accessor for `user_email_verification` returned `nil` because that model doesn't exist.  If `set_new_token` triggers and returns `true` (because it makes an `update` call), then the right side of the `||` never gets evaluated.

services/QuillLMS/app/services/complete_account_creation.rb
```diff
@@ -23,6 +24,11 @@ def perform_admin_actions
     user_email_verification.send_email
   end
 
+  def verify_email_for_third_party_auth
+    return user.verify_email(UserEmailVerification::GOOGLE_VERIFICATION) if user.google_id
+    return user.verify_email(UserEmailVerification::CLEVER_VERIFICATION) if user.clever_id
+  end
```
anathomical: I think that this method probably belongs on the `User` model.  And it may be worth adding a call to it to the `require_email_verification` method.  I think you'd end up with something like this to avoid recursion:
```ruby
def verify_email_for_third_party_auth
  return verify_email(UserEmailVerification::GOOGLE_VERIFICATION) if google_id
  return verify_email(UserEmailVerification::CLEVER_VERIFICATION) if clever_id
end

def require_email_verification
  create_user_email_verification unless user_email_verification
  verify_email_for_third_party_auth
end

def verify_email(verification_method, verification_token = nil)
  require_email_verification unless user_email_verification

  user_email_verification.verify(verification_method, verification_token)
end
```
This allows you to just call `user.require_email_verification if user.teacher?` in the `call` method above.

Also, definitely beyond the scope of this PR, but we should probably figure out if we also want to do third-party auto-verification with Canvas users, if we actually have any of those.

services/QuillLMS/client/app/bundles/Teacher/components/accounts/new/verify_email.tsx
```diff
@@ -41,7 +43,11 @@ const VerifyEmail = ({ user, location, }) => {
   function verifyToken() {
     requestPut(`${process.env.DEFAULT_URL}/verify_emails/verify_by_token`, { token, },
       () => {
-        window.location.href = user.role === ADMIN ? '/sign-up/select-sub-role' : '/profile'
+        if (redirect && isValidRedirect(redirect)) {
+          window.location.href = redirect
+        } else {
+          window.location.href = user.role === ADMIN ? '/sign-up/select-sub-role' : '/profile'
+        }
```
anathomical: I think this might be a little cleaner to understand with an `else if`:
```javascript
if (redirect && isValidRedirect(redirect)) {
  window.location.href = redirect
} else if user.role === ADMIN {
  window.location.href = '/sign-up/select-sub-role'
} else {
  window.location.href = '/profile'
}
```

services/QuillLMS/spec/controllers/teachers/progress_reports_controller_spec.rb
```diff
@@ -199,6 +199,76 @@
...
+        let!(:email_verification) { create(:user_email_verification, user: teacher, verified_at: Date.yesterday) }
+
+        it 'should assign needs_email_verification to false' do
+          get :diagnostic_insights
+          expect(assigns(:needs_email_verification)).not_to be
+        end
+      end
+    end
+
+    context 'when the user is an admin' do
```
anathomical: If we do get to clean up the grandfathering in admins for this feature, we get to remove this context block, too, which feels kind of like a win.

anathomical (approved):
I have a few syntax and style notes, but nothing that should prevent merging this code if you don't want to take any of them.  Overall I like your approach here, and was particularly impressed by the query you managed to put together for the rake task.  That one looked thorny.

services/QuillLMS/app/controllers/teachers/progress_reports_controller.rb
emilia-friedberg: Good question. I think this is a product question so I messaged Peter Sharkey to ask how he feels.

services/QuillLMS/app/controllers/verify_emails_controller.rb
emilia-friedberg: This is a good callout. I think I prefer the explicit if/else in this case, but thank you for pointing out the alternate option here.

services/QuillLMS/app/services/complete_account_creation.rb
emilia-friedberg: Yeah, I was wondering about Canvas too. I'm unclear about how that authorization flow works, because we don't explicitly give teachers the option to sign up that way in the sign up flow, and we don't store a `canvas_id` on the user model. @brendanshean, do you have thoughts here?

eadams17 (approved):
LGTM! Thanks for all of your work on this

services/QuillLMS/app/controllers/teachers/progress_reports_controller.rb
emilia-friedberg: He agreed! Cleaning this up. Thank you for the call-out!

services/QuillLMS/app/services/complete_account_creation.rb
emilia-friedberg: @anathomical as far as this goes:

> This allows you to just call user.require_email_verification if user.teacher? in the call method above.

We actually [don't want to require email verification at this stage](https://www.notion.so/Diagnostic-Insights-Project-Specification-135d42e6f94180398609cc33bf52684f?d=145d42e6f9418070bedc001c147bcb7e&pvs=4#13ed42e6f94180b6bca5e916c72a6540) -- we're automatically granting it to third-party users so we don't have to do some weird logic in different places down the line, per the RFC, but we don't want to send emails to users who are signing up within our system. I think that, given that we don't get that benefit here, I'm going to move this method to the `user` model as suggested but leave the logic of `require_email_verification` alone because, as you point out, some of the logic starts to get circular and I think more difficult to read.

brendanshean: That's a good question.  canvas_id is not stored on user.rb since they can actually have multiple canvas_accounts.  You could use this method in user.rb to set the appropriate third-party auth:

```
  def provider
    return GOOGLE_PROVIDER if google_id.present?
    return CLEVER_PROVIDER if clever_id.present?
    return CANVAS_PROVIDER if canvas_accounts.present?
  end
  ```

emilia-friedberg: Thanks, @brendanshean! I'm gonna check in with Peter Sharkey to confirm that we should be auth-ing this way for Canvas accounts, and if he agrees I'll make another card to address this.

emilia-friedberg: @brendanshean just to confirm, Canvas users log into Quill using Canvas?

brendanshean: It's a much more complicated flow, but yes.  https://github.com/empirical-org/Empirical-Core/blob/develop/services/QuillLMS/app/controllers/canvas_integration/lti_controller.rb#L8

PR #12748, opened 2025-01-29
diagnostic-insights-history-by-classroom

anathomical:
## WHAT
The first API endpoint, built on top of a data service, to retrieve data for the "History and Recommendations" by classroom pages in Diagnostic Insights.

This includes a service pattern as well as a pattern for snapshotting data from staging that we can use to compare against when making updates to the data calculation and parsing so that we can insure that data shape and contents don't drift without our express knowledge.

Caveat: this current code is known to have performance issues, with some requests taking over 10 seconds.  We want to get it out in this state to validate data shape and approach, and work on performance as our next step.
## WHY
We want to establish a good pattern to follow regarding service design and testing frameworks moving forward, and provide a backend that works so that early front-end prototyping can begin.
## HOW
- Create a Service that can take a `classroom_id`, find all students assigned to the classroom, and retrieve their `StudentLearningSequence` history.
- Use Serializers to put this history into a useful shape for the front-end to parse
- Create a controller that validates that the current user is a teacher for the specified classroom
- Create a thor task that builds a snapshot of stable data from the service to validate data shape and maintain consistency moving forward

### Notion Card Links
https://www.notion.so/quill/Build-the-first-DiagnosticInsights-API-HistoryByClassroom-18ad42e6f941806abaa3d4e494c6b377

- Extensive runs of the service with different classroom_id values
- Most of the testing logic is in specs, or in the thor task which generates a JSON blob of data, by-eye analysis of that JSON data is where I've done a bunch of validation of output against real-world examples from our staging database

PR Checklist | Your Answer
------------ | -------------


services/QuillLMS/lib/snapshots/diagnostic_insights/history_by_classroom.json
anathomical, line 1: This file contains a massive JSON snapshot of sample data from the service.  It does not need to be reviewed, but if you're curious about the actual data shape, you're welcome to take a look.

services/QuillLMS/spec/controllers/diagnostic_insights_controller_spec.rb
```diff
@@ -0,0 +1,86 @@
...
+
+      it do
+        subject
+
+        expect(response).to have_http_status(:forbidden)
+      end
+    end
+  end
+
+  describe '#history_by_classroom' do
```
brendanshean, line 35: You could refactor the duplication of testing into its own method:

```
describe '#history_by_classroom' do
  def blah
    expect(DiagnosticInsights::HistoryByClassroom)
      .to receive(:run)
      .with(classroom_id, ell:, start_date:, end_date:)

    subject
  end

  let(:this_year_start) { School.school_year_start(DateTime.current) }
  let(:last_year_start) { School.school_year_start(DateTime.current - 1.year) }
  let(:end_date) { nil }

  context 'ell=true' do
    let(:params) { { classroom_id:, ell: } }
    let(:ell) { false }
    let(:start_date) { this_year_start }

    it { blah }
  end

  context 'timeframes' do
    let(:params) { { classroom_id:, timeframe: } }
    let(:ell) { false }

    context 'all-time' do
      let(:timeframe) { described_class::TIMEFRAME_ALL_TIME }
      let(:start_date) { last_year_start }

      it { blah }
    end

    context 'last-year' do
      let(:timeframe) { described_class::TIMEFRAME_LAST_YEAR }
      let(:start_date) { last_year_start }

      it { blah }
    end

    context 'this-year' do
      let(:timeframe) { described_class::TIMEFRAME_THIS_YEAR }
      let(:start_date) { this_year_start }

      it { blah }
    end
  end
end
```

services/QuillLMS/spec/services/diagnostic_insights/history_by_classroom_spec.rb
```diff
@@ -0,0 +1,114 @@
...
+          it { expect(subject.keys).to eq([user.id, user2.id]) }
+        end
+
+        context 'order by student_learning_sequence.created_at' do
+          let(:new_created_at) { student_learning_sequence.created_at + 1.day }
+          let(:pre_unit2) { create(:unit, activities: [pre_diagnostic]) }
+          let(:pre_classroom_unit2) { create(:classroom_unit, classroom:, unit: pre_unit2, assigned_student_ids:) }
+          let(:student_learning_sequence2) { create(:student_learning_sequence, user:, initial_activity_id: pre_diagnostic.id, initial_classroom_unit: pre_classroom_unit2, created_at: new_created_at) }
+
+          it { expect(subject[user.id].values.first.map { |i| i[:id] }).to eq([student_learning_sequence.id, student_learning_sequence2.id]) }
```
brendanshean: You could use `pluck(:id)` for  `map { |i| i[:id] }`

services/QuillLMS/spec/services/diagnostic_insights/history_by_classroom_spec.rb
```diff
@@ -0,0 +1,114 @@
...
+
+          context 'other order' do
+            let(:new_created_at) { student_learning_sequence.created_at - 1.day }
+
+            it { expect(subject[user.id].values.first.map { |i| i[:id] }).to eq([student_learning_sequence2.id, student_learning_sequence.id]) }
```
brendanshean: You could use `pluck(:id)` for  `map { |i| i[:id] }`

services/QuillLMS/spec/services/diagnostic_insights/history_by_classroom_spec.rb
```diff
@@ -0,0 +1,114 @@
...
+          end
+
+          context 'order by activity completed_at' do
+            let(:new_created_at) { student_learning_sequence.created_at }
+            let(:new_completed_at) { completed_at + 1.day }
+
+            it { expect(subject[user.id].values.first.map { |i| i[:id] }).to eq([student_learning_sequence.id, student_learning_sequence2.id]) }
```
brendanshean: You could use `pluck(:id)` for  `map { |i| i[:id] }`

services/QuillLMS/spec/services/diagnostic_insights/history_by_classroom_spec.rb
```diff
@@ -0,0 +1,114 @@
...
+
+            context 'other order' do
+              let(:new_completed_at) { completed_at - 1.day }
+
+              it { expect(subject[user.id].values.first.map { |i| i[:id] }).to eq([student_learning_sequence2.id, student_learning_sequence.id]) }
```
brendanshean: You could use `pluck(:id)` for  `map { |i| i[:id] }`

services/QuillLMS/spec/models/user_spec.rb
```diff
@@ -682,6 +682,24 @@
...
+
+    let(:sort_earlier_name) { 'Alice Banana' }
+    let(:sort_later_name) { 'Alice Zombie' }
+
+    let!(:sort_later) { create(:user, name: sort_later_name) }
+    let!(:sort_earlier) { create(:user, name: sort_earlier_name) }
+
+    it { expect(subject).to eq([sort_earlier, sort_later]) }
+
+    context 'with a "middle" name: our current logic specifies that anything after the first name is the sort target' do
```
brendanshean, line 696: For completeness, maybe add flip side:

```
context do
  let(:sort_later_name) { 'Alice Aardvark Zombie' }

  it { expect(subject).to eq([sort_later, sort_earlier]) }
end
```

services/QuillLMS/app/services/diagnostic_insights/history_by_classroom.rb
```diff
@@ -0,0 +1,79 @@
...
+        .where(user: { students_classrooms: { classrooms: { id: classroom_id } } })
+        .where(initial_activity_id: pre_diagnostic_ids)
+        .where(student_learning_sequence_activities: { completed_at: start_date..end_date })
+        .merge(User.ordered_by_last_name('user'))
+        .order(:created_at)
+        .order('student_learning_sequence_activities.completed_at')
+    end
+
+    private def fetch_pre_diagnostic_ids
+      unless ell
```
brendanshean: These seem to be pulling from Activity.  You could introduce a couple new constants in activity.rb:
```
DIAGNOSTIC_IDS = STARTER_PRE_DIAGNOSTIC_IDS + 
  INTERMEDIATE_PRE_DIAGNOSTIC_IDS + 
  ADVANCED_PRE_DIAGNOSTIC_IDS

ELL_DIAGNOSTIC_IDS = ELL_STARTER_PRE_DIAGNOSTIC_IDS +
  ELL_INTERMEDIATE_PRE_DIAGNOSTIC_IDS + 
  ELL_ADVANCED_PRE_DIAGNOSTIC_IDS
```

Then the method becomes:

`private def fetch_pre_diagnostic_ids = ell ? Activity::ELL_PRE_DIAGNOSTIC_IDS : Activity::PRE_DIAGNOSTIC_IDS`

services/QuillLMS/app/services/diagnostic_insights/history_by_classroom.rb
```diff
@@ -0,0 +1,79 @@
...
+    def transformed_data
+      grouped_by_user_id = group_by_user_id(serialized_data)
+      grouped_by_diagnostic_type = group_by_diagnostic_type(grouped_by_user_id)
+      attach_name_key(grouped_by_diagnostic_type)
+    end
+
+    def serialized_data = query.map { |row| StudentLearningSequenceSerializer.new(row).as_json }
+
+    # TODO: rename
+    def query
```
brendanshean: Maybe `def student_learning_sequences` for the rename?

services/QuillLMS/app/services/diagnostic_insights/history_by_classroom.rb
```diff
@@ -0,0 +1,79 @@
...
+    ADVANCED_DIAGNOSTIC = 'advanced'
+
+    def initialize(classroom_id, ell: false, start_date: nil, end_date: nil)
+      @classroom_id = classroom_id
+      @ell = ell
+      @start_date = start_date || DEFAULT_START_DATE
+      @end_date = end_date || Float::INFINITY
+    end
+
+    def run
```
brendanshean, line 21: You could get rid of the passthrough method `transformed_data` and introduce some [then](https://til.hashrocket.com/posts/f4agttd8si-chaining-then-in-ruby-26) calls for method chaining:

```
def run
  serialized_data
    .then(&method(:group_by_user_id))
    .then(&method(:group_by_diagnostic_type))
    .then(&method(:attach_name_key))
end
```

services/QuillLMS/app/models/user.rb
```diff
@@ -268,6 +268,9 @@ class User < ApplicationRecord
   scope :teacher, -> { where(role: [ADMIN, TEACHER]) }
   scope :student, -> { where(role: STUDENT) }
   scope :admin, -> { where(role: ADMIN) }
+  # Take a variable with default value for cases where we want to apply this scope in a `joins` call for a `belongs_to` relationship which automatically aliases the table name to `user` instead of `users` (see the `DiagnosticInsights::HistoryByClassroom` service)
+  # Then wrap the table_name in quotes because the common alias "user" is a protected term in PostgreSQL and requires double-quotes to be safe to use
+  scope :ordered_by_last_name, ->(table_name = 'users') { order(Arel.sql("SUBSTRING(TRIM(\"#{table_name}\".name), E'[\\\\s]([^\\\\s]+)')")) }
```
brendanshean: 1). Since we're passing in an argument for table_name, you might want to add a validation for that parameter.
2). I would add a comment about this not always being last_name, but instead the first word after a space in name.
3). You might want to try `split_part` instead of regex since I think it's more [efficient](https://dba.stackexchange.com/a/232392), and a little easier to read.

```
scope :ordered_by_last_name, lambda { |table_name = 'users'|
  raise ArgumentError, 'Invalid table name' unless table_name.match?(/\A[a-zA-Z_][a-zA-Z0-9_]*\z/)

  order(Arel.sql("SPLIT_PART(TRIM(\"#{table_name}\".name), ' ', 2)"))
}
```

If you wanted to order by word after the last space you could use a negative index.
  `    order(Arel.sql("SPLIT_PART(TRIM(\"#{table_name}\".name), ' ', -1)"))`

services/QuillLMS/app/serializers/diagnostic_insights/student_learning_sequence_serializer.rb
```diff
@@ -0,0 +1,15 @@
+# frozen_string_literal: true
+
+module DiagnosticInsights
+  class StudentLearningSequenceSerializer < ActiveModel::Serializer
+    attributes :id, :created_at, :initial_activity_id, :name, :user_id
+
+    has_many :student_learning_sequence_activities do
+      serialized_activities = object.student_learning_sequence_activities.map { |activity| StudentLearningSequenceActivitySerializer.new(activity).as_json }
```
brendanshean: Take or leave, I find the method chaining easier to parse:
```
object
  .student_learning_sequence_activities
  .map { |activity| StudentLearningSequenceActivitySerializer.new(activity).as_json }
  .group_by(&:activity_type)
  .symbolize_keys
```

brendanshean (approved):
Nice work on this.  I left some minor refactoring comments throughout, but nothing that needs re-review.  I like that you've broken this up into setting pieces first and then focusing on performance in a follow-up PR.

I think there's a lingering question about the last_name thing but I think it's out of scope and not something we can solve quickly.

services/QuillLMS/app/models/activity.rb
```diff
@@ -128,6 +128,14 @@ class Activity < ApplicationRecord
   OLD_ELL_STARTER_DIAGNOSTIC_ACTIVITY_ID = 1161
   OLD_ELL_INTERMEDIATE_DIAGNOSTIC_ACTIVITY_ID = 1568
   OLD_ELL_ADVANCED_DIAGNOSTIC_ACTIVITY_ID = 1590
+
+  STARTER_PRE_DIAGNOSTIC_IDS = [STARTER_DIAGNOSTIC_ACTIVITY_ID, OLD_STARTER_DIAGNOSTIC_ACTIVITY_ID]
```
emilia-friedberg, line 132: Not in scope for this, but given that it sounds like we may be switching diagnostics again this summer, we'll probably want to change the name of the `OLD` ones to reflect their specific time scope.

services/QuillLMS/app/models/activity_session_unscoped.rb
```diff
@@ -0,0 +1,38 @@
...
+# Indexes
+#
+#  new_activity_sessions_activity_id_idx        (activity_id)
+#  new_activity_sessions_classroom_unit_id_idx  (classroom_unit_id)
+#  new_activity_sessions_completed_at_idx       (completed_at)
+#  new_activity_sessions_uid_idx                (uid) UNIQUE
+#  new_activity_sessions_uid_key                (uid) UNIQUE
+#  new_activity_sessions_user_id_idx            (user_id)
+#
+class ActivitySessionUnscoped < ActivitySession
```
emilia-friedberg, line 36: It wouldn't have occurred to me to set this up this way, and I like it!

services/QuillLMS/app/serializers/diagnostic_insights/student_learning_sequence_activity_serializer.rb
```diff
@@ -0,0 +1,44 @@
...
+    PRE_RECOMMENDATION = 'pre_recommendation'
+    POST_RECOMMENDATION = 'post_recommendation'
+
+    attributes :activity_id, :activity_type, :completed_at, :created_at, :activity_session_id, :proficient
+
+    def proficient
+      return nil unless pre_diagnostic? || post_diagnostic?
+
+      object.activity_session_unscoped&.concept_results&.all?(&:correct)
+    end
```
emilia-friedberg, line 28: Just confirming that we want to define proficiency as having all concept results correct? That seems like a high bar.

services/QuillLMS/app/serializers/diagnostic_insights/student_learning_sequence_serializer.rb
```diff
@@ -0,0 +1,15 @@
...
+      serialized_activities.group_by { |activity| activity[:activity_type] }.symbolize_keys
+    end
+
+    def name = object.user.name
```
emilia-friedberg: Take or leave but I kind of want this to be `user_name`? When I first read it I was like, oh, learning sequences have names?

emilia-friedberg (approved):
This looks good to me overall! It's a little hard for me to parse the overall data shape (I did look at that massive file!), which makes sense because it's so big, but maybe in a test or elsewhere we could document it so that you can look in one place to see what each of the variable key values refer to?

The other concern I have is around proficiency, as noted above -- is getting every question correct the same as having every concept correct? Maybe it is and I'm misremembering, since how we grade diagnostics have gone through a few changes in the last few years. Either way, I don't necessarily think it should be a blocker, since this shape is good enough for me to get going on the frontend.

happythenewsad (approved):
LGTM

services/QuillLMS/app/models/user.rb
```diff
@@ -268,6 +268,9 @@ class User < ApplicationRecord
   scope :teacher, -> { where(role: [ADMIN, TEACHER]) }
   scope :student, -> { where(role: STUDENT) }
   scope :admin, -> { where(role: ADMIN) }
+  # Take a variable with default value for cases where we want to apply this scope in a `joins` call for a `belongs_to` relationship which automatically aliases the table name to `user` instead of `users` (see the `DiagnosticInsights::HistoryByClassroom` service)
```
dandrabik: It's somewhat odd to need to pass the model's table name into a scope. The table name is a constant, couldn't that just be explicit in the `order`, e.g. `{ order('users.name DESC')`?

services/QuillLMS/app/serializers/diagnostic_insights/student_learning_sequence_activity_serializer.rb
anathomical: Good question.  Peter S flagged this during the spec-writing process, but we're deferring to Curriculum on this one.  We may change our minds later about what counts as "proficient", but right now it's getting every question correct.

services/QuillLMS/app/serializers/diagnostic_insights/student_learning_sequence_activity_serializer.rb
```diff
@@ -0,0 +1,44 @@
...
+    POST_DIAGNOSTIC = 'post_diagnostic'
+    PRE_RECOMMENDATION = 'pre_recommendation'
+    POST_RECOMMENDATION = 'post_recommendation'
+
+    attributes :activity_id, :activity_type, :completed_at, :created_at, :activity_session_id, :proficient
+
+    def proficient
+      return nil unless pre_diagnostic? || post_diagnostic?
+
+      object.activity_session_unscoped&.concept_results&.all?(&:correct)
```
dandrabik, line 27: Note, this is a faster version of this. Rails has some finder methods that translate to pure SQL (see screen for difference).
```ruby
concept_results.where(correct: false).none?
```
Screenshot 2025-01-29 at 4 54 48 PM

services/QuillLMS/app/serializers/diagnostic_insights/student_learning_sequence_activity_serializer.rb
```diff
@@ -0,0 +1,44 @@
...
+    end
+
+    def activity_type
+      return PRE_DIAGNOSTIC if PRE_DIAGNOSTIC_IDS.include?(object.activity_id)
```
dandrabik, line 31: Curious, do you need `object` in all these methods, or could you just call `activity_id` since it's an attribute?

services/QuillLMS/app/services/diagnostic_insights/history_by_classroom.rb
```diff
@@ -0,0 +1,79 @@
...
+          Activity::INTERMEDIATE_PRE_DIAGNOSTIC_IDS +
+          Activity::ADVANCED_PRE_DIAGNOSTIC_IDS)
+      end
+
+      (Activity::ELL_STARTER_PRE_DIAGNOSTIC_IDS +
+        Activity::ELL_INTERMEDIATE_PRE_DIAGNOSTIC_IDS +
+        Activity::ELL_ADVANCED_PRE_DIAGNOSTIC_IDS)
+    end
+
+    private def diagnostic_type(activity_id)
```
dandrabik, line 48: This method is hard to read, consider breaking the id logic off, e.g.
```ruby
STARTER_PRE_DIAGNOSTIC_IDS = Activity::STARTER_PRE_DIAGNOSTIC_IDS + Activity::ELL_STARTER_PRE_DIAGNOSTIC_IDS
...
return STARTER_DIAGNOSTIC if STARTER_PRE_DIAGNOSTIC_IDS.include?(activity_id)

services/QuillLMS/lib/snapshots/diagnostic_insights/history_by_classroom.json
dandrabik: Are we considered at all at the size of this payload? 10k JSON lines?

dandrabik (approved):
I've requested some small refactors, but overall looks good. 

One thing I would request is to run this endpoint locally and tail the log to see all the SQL it generates to see if there are some optimizations to make.

Also, that's a pretty large JSON payload (this example is 10K lines). Are some going to be larger than is? What would the largest on look like?

services/QuillLMS/app/serializers/diagnostic_insights/student_learning_sequence_activity_serializer.rb
emilia-friedberg: @anathomical but is having all concept results correct the same as having every question correct? Is it not possible to have like an incorrect concept result for spelling but still get the question's concept and therefore the question itself right?

PR #12752, opened 2025-01-30
classlink - sync all district data

happythenewsad:
## WHAT
- specify the final ActiveRecord models and relationships required for Classlink
- implements minimum viable Classlink rostering server auto-discovery and idempotent data syncing 
- added a global mode to AppSetting that doesn't depend upon `current_user`

## NOT INCLUDED
- resource deletion propagation
- SSO models (e.g. `AuthCredentials`), although SSO controller are route are included and functional
- complete set of postgres table indices
- performance optimization
- rate limiting (haven't had an issue so far)
- AcademicSession filtering (we probably want to limit syncs to the current academic session eventually) 

## WHY
- structure Classlink code such that development can continue behind an AppSetting upon merging, without affecting existing users 

## HOW
- Some tables are STI, some are not. I followed the convention set by Canvas and Clever on this. 
- added a ProviderClassroom polymorphic reference. This pattern should be used for future 3rd party rostering integrations
- various migrations 
- added Fetcher and Persister classes for required Classlink API endpoints 

### Screenshots

### Notion Card Links
https://www.notion.so/quill/Product-Board-30f97e4eb01246dbb8264253fb385073

1. On development (staging afterwards), run 
`importer = ClasslinkIntegration::DistrictImporter.new(district_oneroster_id: 'JUzdk05jMjM%3D', district_bearer_token: 'ebe3cd0c-300a-4b3d-ac94-75a770e66f1e', district_id: 22553).run `

Then cross-check several of:
- users
- enrollments
- schools
- classes 

against the raw API data retrievable from our Classlink postman request library

Then repeat and ensure sync is idempotent but detecting no changes in ActiveRecord counts.

2. As a google teacher, click the `Import classrooms` button and observe no errors
3. Assign an activity

This is the programmatic portion of the QA process:

```ruby
district = District.find_by_name "ClassLink Test District"

# verify counts
ClasslinkAccount.count => 1261
ClasslinkClassroom.count => 186
district.schools.count => ???

# verify users
User.where(email: "billy.flores@cert3.classlink.k12.nj.us").includes(:classlink_accounts).first.tap do |s1|
  raise 'fail' unless s1.role == 'student' && s1.classlink_accounts.first.external_id == "18125"
end

User.where(email: "blas.morales@cert3.classlink.k12.nj.us").includes(:classlink_accounts).first.tap do |s1|
  raise 'fail' unless s1.role == 'student' && s1.classlink_accounts.first.external_id == "29117"
end

User.where(email: "taylor.helms@cert3.classlink.k12.nj.us").includes(:classlink_accounts).first.tap do |s1|
  raise 'fail' unless s1.role == 'teacher' && s1.classlink_accounts.first.external_id == "28620"
end

# verify classrooms
ClasslinkClassroom.find_by(external_id: "31763").tap do |cc|
  raise 'fail' unless cc.classroom.name == "ELA 6(6A ELA)"
end

ClasslinkClassroom.find_by(external_id: "D1_A22_103422").tap do |cc|
  raise 'fail' unless cc.classroom.name == "LA/Eng 8 G/H"
end

# verify enrollments
user_id = ClasslinkAccount.find_by_external_id("19214").user.id
classroom_id = ClasslinkClassroom.find_by_external_id("32522").classroom.id
raise 'fail' unless StudentsClassrooms.find_by(student_id: user_id, classroom_id:)

user_id = ClasslinkAccount.find_by_external_id("44839").user.id
classroom_id = ClasslinkClassroom.find_by_external_id("32073").classroom.id
raise 'fail' unless ClassroomsTeacher.find_by(user_id:, classroom_id:)
```
PR Checklist | Your Answer
------------ | -------------


services/QuillLMS/app/models/provider_classroom.rb
```diff
@@ -4,18 +4,21 @@
...
-#  external_id        :string           not null
+#  id                     :bigint           not null, primary key
+#  provider_instance_type :string
+#  type                   :string           not null
+#  created_at             :datetime         not null
+#  updated_at             :datetime         not null
+#  canvas_instance_id     :bigint
+#  classroom_id           :bigint           not null
+#  external_id            :string           not null
+#  provider_instance_id   :bigint
```
happythenewsad, line 15: Adding a polymorphic reference here, in preparation for possible future integrations

services/QuillLMS/app/models/classlink_instance.rb
```diff
@@ -0,0 +1,33 @@
...
+#
+#  index_classlink_instances_on_district_id  (district_id)
+#
+# Foreign Keys
+#
+#  fk_rails_...  (district_id => districts.id)
+#
+class ClasslinkInstance < ApplicationRecord
+  belongs_to :district
+  belongs_to :classlink_instance
```
brendanshean: Is this belongs_to a typo?

services/QuillLMS/spec/factories/classlink_integration/classlink_accounts.rb
```diff
@@ -0,0 +1,9 @@
+# frozen_string_literal: true
+
+FactoryBot.define do
+  factory :classlink_account do
+    user
+    external_id { 'external-999' }
```
brendanshean: I would use Faker:
    `external_id { Faker::Number.number.to_s }`
or sequence:
`sequence(:external_id) { |n| "external-#{n}" }`

services/QuillLMS/spec/services/classlink_integration/enrollment_persister_spec.rb
```diff
@@ -0,0 +1,156 @@
...
+    }
+  end
+
+  let(:classlink_instance) { create(:classlink_instance) }
+
+  describe '#run' do
+    context 'no user or classroom found' do
+      it 'should do nothing and return early' do
+        expect do
+          described_class.new(
```
brendanshean: You can use `described.class.run` instead of `described_class.new(...).run` since this is an `ApplicationService`

services/QuillLMS/spec/services/classlink_integration/enrollment_persister_spec.rb
```diff
@@ -0,0 +1,156 @@
...
+            classlink_instance_id: classlink_instance.id
+          ).includes(:user)
+
+          existing_provider_classrooms = ClasslinkClassroom.where(
+            provider_instance_id: classlink_instance.id
+          ).includes(:classroom)
+          # TODO: refactor to use StudentClassroomsAssociator
+          # expect(StudentClassroomAssociator).to receive(:run).once
+          expect do
+            described_class.new(
```
brendanshean: You can use described.class.run instead of described_class.new(...).run since this is an ApplicationService

services/QuillLMS/spec/services/classlink_integration/enrollment_persister_spec.rb
```diff
@@ -0,0 +1,156 @@
...
+          classroom = create(:classroom)
+          classlink_classroom = create(
+            :classlink_classroom,
+            classroom:,
+            external_id: incoming_teacher_enrollment[:classroom_sourced_id],
+            provider_instance: classlink_instance
+          )
+
+          expect do
+            described_class.new(
```
brendanshean: You can use described.class.run instead of described_class.new(...).run since this is an ApplicationService

services/QuillLMS/app/services/classlink_integration/districts_importer.rb
```diff
@@ -0,0 +1,58 @@
...
+          ErrorNotifier.report(
+            "Invalid payload for district: #{district_hash['oneroster_application_id']}",
+            'ClasslinkIntegration::DistrictImporter#run'
+          )
+          next
+        end
+
+        district_id = find_or_create_district_and_instance(district_hash)
+
+        DistrictImporter.new(
```
brendanshean: You can use DistrictImporter.run instead of DistrictImporter.new(...).run since this is an ApplicationService

services/QuillLMS/spec/services/classlink_integration/user_persister_spec.rb
```diff
@@ -0,0 +1,55 @@
...
+          .to not_change(ClasslinkAccount, :count)
+          .and not_change(User, :count)
+
+        existing_user.reload
+        expect(existing_user.email).to eq('foo@bar.com')
+      end
+    end
+
+    context 'user does not exist' do
+      existing_provider_users = []
```
brendanshean: A `let(:existing_provider_users) { [] }` would be more conventional here.

services/QuillLMS/spec/services/classlink_integration/user_persister_spec.rb
```diff
@@ -0,0 +1,55 @@
...
+      password: 'Booth',
+      role: 'student',
+      signed_up_with_google: false,
+      username: 'johnbooth',
+      sourced_id: 'external-123'
+    }
+  end
+
+  describe '#run' do
+    let(:classlink_instance) { create(:classlink_instance) }
```
brendanshean, line 27: Since the setup for persister is the same parameters, you could refactor out a `subject { described_class.new(incoming_user:, existing_provider_users:, classlink_instance_id: classlink_instance.id) }`

Then your two tests become:

`expect { subject }`

This communicates that you're testing the same structure of `subject` in each test.

services/QuillLMS/spec/services/classlink_integration/enrollment_persister_spec.rb
```diff
@@ -0,0 +1,156 @@
...
+      let(:classroom2) { create(:classroom) }
+      let!(:classlink_classroom2) do
+        create(
+          :classlink_classroom,
+          classroom: classroom2,
+          provider_instance: classlink_instance
+        )
+      end
+
+      context 'role: teacher' do
```
brendanshean: This context block should be outside of `#run` since it's not calling `run` but `create_classroom_teacher`

services/QuillLMS/spec/services/classlink_integration/enrollment_persister_spec.rb
```diff
@@ -0,0 +1,156 @@
...
+          end
+            .to change(StudentsClassrooms, :count).by(1)
+            .and change(ClasslinkClassroomUser, :count).by 1
+        end
+      end
+    end
+
+    context 'association does not exist' do
+      context 'role: teacher' do
+        let(:existing_provider_users) do
```
brendanshean: Good opportunity for one-liner:
`let(:existing_provider_users) { [create(:teacher)] }`

services/QuillLMS/spec/services/classlink_integration/classroom_persister_spec.rb
```diff
@@ -0,0 +1,83 @@
...
+        expect { persister.run }.to not_change(ClasslinkClassroom, :count)
+
+        existing_classroom.reload
+        expect(existing_classroom.name).to eq('Math 101')
+        expect(existing_classroom.visible).to be true
+      end
+    end
+
+    context 'classroom does not exist' do
+      existing_provider_classrooms = []
```
brendanshean: Using a `let` block for variable assignment outside of test block is more conventional for RSpec.

services/QuillLMS/spec/services/classlink_integration/classroom_persister_spec.rb
```diff
@@ -0,0 +1,83 @@
...
+  let(:incoming_classroom) do
+    {
+      sourced_id: 'external-123',
+      name: 'Math 101',
+      visible: true
+    }
+  end
+
+  describe '#run' do
+    let(:classlink_instance) { create(:classlink_instance) }
```
brendanshean, line 23: Since the setup for persister is the same parameters, you could refactor out a 
`subject { described_class.new(incoming_classroom:, existing_provider_classrooms:, classlink_instance_id: classlink_instance.id)`

Then your two tests become:

`expect { subject }`

This communicates that you're testing the same structure of subject in each test.

services/QuillLMS/app/services/classlink_integration/schools_fetcher.rb
```diff
@@ -0,0 +1,15 @@
+# frozen_string_literal: true
+
+module ClasslinkIntegration
+  class SchoolsFetcher < BaseFetcher
+    def resource_query_string = "schools?filter=type='school'"
```
brendanshean: I would extract a constant for this query string.

services/QuillLMS/app/services/classlink_integration/enrollments_fetcher.rb
```diff
@@ -0,0 +1,16 @@
+# frozen_string_literal: true
+
+module ClasslinkIntegration
+  class EnrollmentsFetcher < BaseFetcher
+    def resource_query_string = "enrollments?filter=status='active'"
```
brendanshean: I would extract a constant for this query string.

services/QuillLMS/app/services/classlink_integration/enrollment_persister.rb
```diff
@@ -0,0 +1,65 @@
...
+    OWNER = ClassroomsTeacher::ROLE_TYPES[:owner]
+
+    def initialize(incoming_enrollment:, existing_provider_users:, existing_provider_classrooms:)
+      @incoming_enrollment = incoming_enrollment
+      @existing_provider_users = existing_provider_users
+      @existing_provider_classrooms = existing_provider_classrooms
+    end
+
+    def user_and_classroom
+      quill_user = existing_provider_users.find do |ec|
```
brendanshean: For clarity, I think `ec` should be `eu`

services/QuillLMS/app/services/classlink_integration/enrollment_persister.rb
```diff
@@ -0,0 +1,65 @@
...
+        ec.external_id == incoming_enrollment[:user_sourced_id]
+      end&.user
+      quill_classroom = existing_provider_classrooms.find do |ec|
+        ec.external_id == incoming_enrollment[:classroom_sourced_id]
+      end&.classroom
+      { quill_user:, quill_classroom: }
+    end
+
+    def run
+      user_and_classroom => { quill_user:, quill_classroom: }
```
brendanshean: I think pattern matching here is kind of confusing.  

Instead, you could introduce two separate methods:

```
def quill_user
   @quill_user ||= 
      existing_provider_users
        .find {  |eu|. ec.external_id == incoming_enrollment[:user_sourced_id] }
        &.user
end
```

```
def quill_classroom
  @quill_classroom ||=
      existing_provider_classrooms
        .find { |ec| ec.external_id == incoming_enrollment[:classroom_sourced_id] }
        &.classroom
end
```

then you're the beginning of `run` becomes simplified:

```
def run
  return unless quill_classroom && quill_user
```

services/QuillLMS/app/services/classlink_integration/classrooms_fetcher.rb
```diff
@@ -0,0 +1,16 @@
+# frozen_string_literal: true
+
+module ClasslinkIntegration
+  class ClassroomsFetcher < BaseFetcher
+    def resource_query_string = "classes?filter=status='active'"
```
brendanshean: I would extract constants for the query_string and the selector_override

services/QuillLMS/app/controllers/auth/classlink_controller.rb
```diff
@@ -0,0 +1,41 @@
+# frozen_string_literal: true
+
+require 'net/http'
+require 'uri'
+
+module Auth
+  class ClasslinkController < ApplicationController
+    around_action :force_writer_db_role, only: [:classlink]
+    CLIENT_ID = 'c17370377633256aee86a52043b96d2913e27b36df24'
```
brendanshean: Let's extract `CLIENT_ID`, `OAUTH_TOKEN_ENDPOINT` and `client_secret` into a new file `config/initializers/auth/classlink` per the pattern with existing providers `auth/{canvas, clever, google}`

That way these are loaded once.

services/QuillLMS/spec/services/classlink_integration/classroom_persister_spec.rb
```diff
@@ -0,0 +1,83 @@
...
+
+    context 'when the classroom already exists' do
+      let(:existing_classroom) { create(:classroom, name: 'Old Math Class', visible: false) }
+      let(:classlink_classroom) do
+        create(:classlink_classroom,
+          external_id: 'external-123',
+          classroom: existing_classroom,
+          provider_instance: classlink_instance,
+          type: ClasslinkClassroom)
```
brendanshean: I don't think you need the type here as it should be inferred from the factory.

services/QuillLMS/spec/services/classlink_integration/classroom_persister_spec.rb
```diff
@@ -0,0 +1,83 @@
...
+    end
+
+    context 'with multiple existing classrooms' do
+      let(:matching_classroom) { create(:classroom, name: 'Matching Class') }
+      let(:matching_classlink_classroom) do
+        create(:classlink_classroom,
+          external_id: 'external-123',
+          classroom: matching_classroom,
+          provider_instance: classlink_instance,
+          type: ClasslinkClassroom)
```
brendanshean: I don't think you need the type here as it should be inferred from the factory.

services/QuillLMS/spec/services/classlink_integration/classroom_persister_spec.rb
```diff
@@ -0,0 +1,83 @@
...
+      end
+
+      let(:non_matching_classroom) { create(:classroom, name: 'Non-matching Class') }
+      let(:non_matching_classlink_classroom) do
+        create(:classlink_classroom,
+          external_id: 'external-456',
+          classroom: non_matching_classroom,
+          provider_instance: classlink_instance,
+          type: ClasslinkClassroom)
```
brendanshean: I don't think you need the type here as it should be inferred from the factory.

services/QuillLMS/app/services/classlink_integration/classroom_persister.rb
```diff
@@ -0,0 +1,42 @@
...
+      begin
+        if existing_classroom
+          existing_classroom.classroom.update!(
+            visible: incoming_classroom[:visible],
+            name: incoming_classroom[:name]
+          )
+        else
+          Classroom.create!(name: incoming_classroom[:name], visible: incoming_classroom[:visible]) do |c|
+            ClasslinkClassroom.create!(
+              type: ClasslinkClassroom,
```
brendanshean: I don't think you need the type here as it should be inferred from the model.

brendanshean (approved):
This well-structured first piece of Classlink integration.  Specifically, I liked:
1. clean separation between fetchers and persisters
2. feature toggling for rollout
3. ErrorNotifier for tracking potential issues when these objects are created/updated.

I've left some refactoring comments that I'd like to be addressed, but nothing that needs a re-review.

Great work on this!

services/QuillLMS/app/models/classlink_instance.rb
happythenewsad: Yes; removed

services/QuillLMS/app/controllers/auth/classlink_controller.rb
```diff
@@ -0,0 +1,37 @@
...
+
+      uri = URI.parse(::Classlink::OAUTH_TOKEN_ENDPOINT)
+      form_data = {
+        client_id: ::Classlink::CLIENT_ID,
+        client_secret: ::Classlink::CLIENT_SECRET,
+        code:
+      }
+
+      response = Net::HTTP.post_form(uri, form_data)
+      parsed = JSON.parse response.body
```
dandrabik, line 21: You might consider extracting this out into a service, e.g. `ClasslinkDataFetcher.run(params['code'])` or something like that so you can handle exceptions and edge cases and make the code easily testable.

services/QuillLMS/spec/services/classlink_integration/classroom_persister_spec.rb
happythenewsad: In this case, the persister does not have the same params for each test. This is missable due to the use of the implicit `key:` syntax

services/QuillLMS/app/models/app_setting.rb
```diff
@@ -32,9 +32,11 @@ class AppSetting < ApplicationRecord
 
   scope :enabled, -> { where(enabled: true) }
 
-  def self.enabled?(name:, user:)
+  def self.enabled?(name:, user: nil)
```
dandrabik: Hmm, I wonder if it would be better to have a separate method for logged_out logic, like `logged_out_enabled?` or something. I am a little worried about defaulting to `enabled=true` if a user isn't passed in, e.g. if there is a bug upstream when a user should be passed in, but it passes in `nil`.

services/QuillLMS/app/models/classlink_classroom_user.rb
```diff
@@ -0,0 +1,30 @@
...
+#
+# Indexes
+#
+#  index_provider_type_and_classroom_id_and_user_id  (type,classroom_external_id,user_external_id) UNIQUE
+#
+
+# This class should be called ClasslinkClassroomsStudent instead of ClasslinkClassroomUser.
+# But sticking to existing convention.
+class ClasslinkClassroomUser < ProviderClassroomUser
+  def classlink_classroom_id
```
dandrabik: Take or leave, consider one-liner methods here.

services/QuillLMS/app/models/user.rb
```diff
@@ -910,18 +914,25 @@ def provider
...
       .find_by(canvas_instance:)
       &.user_external_id
+
+    return canvas_external_id if canvas_external_id
+
+    classlink_external_id = classlink_accounts.first&.user_external_id
+
+    return classlink_external_id if classlink_external_id
+
+    nil
```
dandrabik, line 954: Take or leave, since you might want the extra clarity, but this is equivalent to these 3 lines:
```ruby
classlink_accounts.first&.user_external_id

File: services/QuillLMS/app/models/user.rb [hunk 28]
```diff
@@ -992,7 +1003,7 @@ def segment_admin_report_subscriptions
   end
 
   private def requires_password?
-    return false if clever_id || google_id || canvas_accounts.present?
+    return false if clever_id || google_id || canvas_accounts.present? || classlink_accounts.present?
```
dandrabik: Take or leave, consider extracting `clever_id || google_id || canvas_accounts.present? || classlink_accounts.present?` into a method, e.g.
```ruby
return false if external_ids_present?

File: services/QuillLMS/spec/services/classlink_integration/user_persister_spec.rb (hunk 9 above)
Comment (happythenewsad) [reply services/QuillLMS/spec/services/classlink_integration/user_persister_spec.rb:27]: Similar to above, the structure of `persister` is subtly different between the tests

File: services/QuillLMS/app/services/classlink_integration/base_fetcher.rb [hunk 29]
```diff
@@ -0,0 +1,69 @@
...
+      items = 1.upto(total_pages - 1).reduce(initial_page[api_response_selector]) do |memo, page_num|
+        memo.concat(fetch_page(offset: page_num * DEFAULT_API_PAGE_SIZE)[api_response_selector])
+      end
+
+      items
+        .uniq { |hash| hash['sourcedId'] }
+        .compact
+    end
+
+    def fetch_page(offset: 0, limit: DEFAULT_API_PAGE_SIZE)
```
dandrabik, line 55: Consider breaking off the HTTP request code into its own service(s) and pull in the paging logic, e.g.
```ruby
ClasslinkAPIRequest.run(district_oneroster_id:, page:, resource:, query_options:)

File: services/QuillLMS/app/services/classlink_integration/base_fetcher.rb [hunk 30]
```diff
@@ -0,0 +1,69 @@
...
+    def resource_query_string = raise NotImplementedError
+
+    # sometimes, Classlink API responses are nested under a key that is not the resource name
+    # for example, the 'schools' resource is nested under the 'orgs' key
+    def api_response_selector_override = nil
+
+    # according to OneRoster spec, section 4.1, sourcedId is a GUID within a System
+    # in other words, it's the most globally unique identifier available
+    def fetch_all
+      resource = resource_query_string.split('?').first
```
dandrabik: Maybe make this two variables in the child class, `resource` and `query_string`

services/QuillLMS/app/services/classlink_integration/base_fetcher.rb
```diff
@@ -0,0 +1,69 @@
...
+      api_response_selector = api_response_selector_override || resource
+
+      initial_page = fetch_page
+      total_item_count = initial_page.headers['x-total-count'].to_i
+      total_pages = (total_item_count / DEFAULT_API_PAGE_SIZE.to_f).ceil
```
dandrabik, line 44: It adds a request, but you might consider doing a separate request / service that uses a limit of one in its request  to just calculate the number of pages (via the header info). That might clean up some of the code here so you could do something like:
```ruby
page_count = ClasslinkAPIPageCounter.run(district_oneroster_id:, resource:, query_string:)

results = (1..page_count).map {|page_number| fetch_page(page_number)[selector] }.flatten

File: services/QuillLMS/spec/services/classlink_integration/user_persister_spec.rb (hunk 9 above)
Comment (brendanshean) [reply services/QuillLMS/spec/services/classlink_integration/user_persister_spec.rb:27]: I'm not sure I follow.  They look the same to me:

        persister = described_class.new(incoming_user:, existing_provider_users:, classlink_instance_id: classlink_instance.id)
        persister = described_class.new(incoming_user:, existing_provider_users:, classlink_instance_id: classlink_instance.id)

File: services/QuillLMS/app/services/classlink_integration/classroom_persister.rb [hunk 32]
```diff
@@ -0,0 +1,41 @@
...
+    attr_reader :incoming_classroom, :existing_provider_classrooms, :classlink_instance_id
+
+    def initialize(incoming_classroom:, existing_provider_classrooms:, classlink_instance_id:)
+      @incoming_classroom = incoming_classroom
+      @existing_provider_classrooms = existing_provider_classrooms
+      @classlink_instance_id = classlink_instance_id
+    end
+
+    def run
+      existing_classroom = existing_provider_classrooms.find do |ec|
```
dandrabik, line 14: Consider making extracting this as a private method to simplify this long method.
```ruby
private def existing_classrooms
  @existing_classrooms || = existing_provider_classrooms.find do |ec|
    ec.external_id == incoming_classroom[:sourced_id]
  end
end

File: services/QuillLMS/app/services/classlink_integration/classroom_persister.rb [hunk 33]
```diff
@@ -0,0 +1,41 @@
...
+        ec.external_id == incoming_classroom[:sourced_id]
+      end
+
+      begin
```
dandrabik: Consider removing the begin/end and making the whole method rescured.

services/QuillLMS/app/services/classlink_integration/district_importer.rb
```diff
@@ -0,0 +1,80 @@
...
+      # there's a 1:1 relationship between a District and a ClasslinkInstance
+      schools.each do |incoming_school|
+        School.find_or_create_by!(name: incoming_school[:name], district_id:)
+      end
+
+      classlink_instance_id = ClasslinkInstance.find_by(oneroster_application_id: district_oneroster_id).id
+
+      existing_provider_classrooms = ClasslinkClassroom.where(
+        provider_instance_id: classlink_instance_id
+      ).includes(:classroom)
```
dandrabik: Consider making this a private method that is memoized.

services/QuillLMS/app/services/classlink_integration/district_importer.rb
```diff
@@ -0,0 +1,80 @@
...
+    def persist_records
+      return unless AppSetting.enabled?(name: Constants::APP_SETTING)
+
+      # Unlike Canvas, we don't need the ***InstanceSchool join table, since
+      # there's a 1:1 relationship between a District and a ClasslinkInstance
+      schools.each do |incoming_school|
+        School.find_or_create_by!(name: incoming_school[:name], district_id:)
+      end
+
+      classlink_instance_id = ClasslinkInstance.find_by(oneroster_application_id: district_oneroster_id).id
```
dandrabik: Consider making this a private method that is memoized.

services/QuillLMS/app/services/classlink_integration/district_importer.rb
```diff
@@ -0,0 +1,80 @@
...
+        provider_instance_id: classlink_instance_id
+      ).includes(:classroom)
+
+      classrooms.each do |incoming_classroom|
+        ClassroomPersister.run(incoming_classroom:, existing_provider_classrooms:, classlink_instance_id:)
+      end
+
+      existing_provider_users = ClasslinkAccount.where(
+        classlink_instance_id:
+      ).includes(:user)
```
dandrabik: Consider making this a private method that is memoized.

services/QuillLMS/app/services/classlink_integration/user_persister.rb
```diff
@@ -0,0 +1,43 @@
...
+
+    def initialize(incoming_user:, existing_provider_users:, classlink_instance_id:)
+      @incoming_user = incoming_user
+      @existing_provider_users = existing_provider_users
+      @classlink_instance_id = classlink_instance_id
+    end
+
+    def run
+      existing_user = existing_provider_users.find { |ec| ec.external_id == incoming_user[:sourced_id] }
+      creatable_user_hash = UserPersister.creatable_user(incoming_user)
```
dandrabik: Consider making this two private methods that are memoized.

dandrabik (changes requested):
I think overall this is headed in the right direction. Thanks for your work on this. Since I commented a decent amount of refactor suggestions, I'd like you to take a refactor pass, then I'll review again. You don't need to take all of my suggestions, but I think there are opportunities to make this code a little cleaner, so it's worth a pass.

services/QuillLMS/app/models/user.rb
happythenewsad: I anticipate that the current pattern, while more verbose, will be long term more successful, since

- we want to clearly signal that a `nil` return value is expected in some cases 
- if we add another 3rd party provider, we just continue the existing pattern

services/QuillLMS/app/models/user.rb
happythenewsad: found a way to make it even more succinct -> recycle the `provider` method here

services/QuillLMS/app/controllers/auth/classlink_controller.rb
happythenewsad: Agreed; I'm going to nudge this controller into its final form later this week, when I associate the SSO flow with the LMS persistence layer. Right now, the controller is a stub.

services/QuillLMS/app/models/classroom.rb
```diff
@@ -64,6 +65,7 @@ class Classroom < ApplicationRecord
 
   accepts_nested_attributes_for :canvas_classroom, :classrooms_teachers
 
+  delegate :classroom_external_id, to: :canvas_classroom, allow_nil: true
   delegate :classroom_external_id, to: :canvas_classroom, allow_nil: true
```
anathomical: Is this a duplicate line?

services/QuillLMS/app/models/classroom.rb
```diff
@@ -185,6 +187,8 @@ def classroom_type_for_segment
       'Clever'
     elsif canvas_classroom
       'Canvas'
+    elsif classlink_classroom
+      'Classlink'
```
anathomical: At some point I think it'd be cleaner to move from an if/else stack to a hash, but it may be premature to do so right now.

anathomical: I recognize that you're extending an existing pattern, so feel free to take or leave this.  Are the values other than 'Manual' here identical to the ones returned by the `provider` method below?  Would it make sense to combine these into one method?

services/QuillLMS/app/models/user.rb
```diff
@@ -1010,7 +1021,7 @@ def segment_admin_report_subscriptions
   end
 
   private def requires_password?
-    return false if clever_id || google_id || canvas_accounts.present?
+    return false if provider
```
anathomical, line 1025: I like this simplification.

services/QuillLMS/app/services/classlink_integration/district_importer.rb
```diff
@@ -0,0 +1,86 @@
...
+
+    def validate_users!
+      # TODO: implement
+    end
+
+    def run
+      @schools = SchoolsFetcher.run(district_oneroster_id:, district_bearer_token:, district_id:)
+      @classrooms = ClassroomsFetcher.run(district_oneroster_id:, district_bearer_token:, district_id:)
+      @users = UsersFetcher.run(district_oneroster_id:, district_bearer_token:, district_id:)
+      @enrollments = EnrollmentsFetcher.run(district_oneroster_id:, district_bearer_token:, district_id:)
```
anathomical: This is purely stylistic, so take or leave this, but I think we've been moving more toward a pattern where these could be written as methods:
```ruby
def schools = @schools ||= SchoolsFetcher.run(district_oneroster_id:, district_bearer_token:, district_ids:)
```

services/QuillLMS/app/services/classlink_integration/district_importer.rb
```diff
@@ -0,0 +1,86 @@
...
+      users.each do |incoming_user|
+        UserPersister.run(incoming_user:, existing_provider_users:, classlink_instance_id:)
+      end
+
+      enrollments.each do |incoming_enrollment|
+        EnrollmentPersister.run(incoming_enrollment:, existing_provider_users:, existing_provider_classrooms:)
+      end
+    end
+
+    private def classlink_instance_id = ClasslinkInstance.find_by(oneroster_application_id: district_oneroster_id).id
```
anathomical: Probably worth memoizing this so that we don't have to make another trip to the database every time this is referenced:
```ruby
private def classlink_instance_id = @classlink_instanced_id ||= ClasslinkInstance.find_by(oneroster_application_id: district_oneroster_id).id
```

services/QuillLMS/app/services/classlink_integration/districts_importer.rb
```diff
@@ -0,0 +1,58 @@
...
+          district_bearer_token: district_hash['bearer'],
+          district_id:
+        )
+      end
+    end
+
+    def find_or_create_district_and_instance(district_hash)
+      ActiveRecord::Base.transaction do
+        instance = ClasslinkInstance.find_or_create_by(oneroster_application_id: district_hash['oneroster_application_id']) do |inst|
+          district = District.create!(name: district_hash['tenant_name'])
```
anathomical, line 33: Should this be a `find_or_create_by!` call for cases where there's already a District with the same name, or would we prefer to create an additional one rather than risk "merging" two different districts that just happen to have the same names even if they're not the same?

services/QuillLMS/app/services/classlink_integration/enrollment_persister.rb
```diff
@@ -0,0 +1,63 @@
...
+        # We cannot use StudentClassroomAssociator unless we process all teachers before all students
+        # ::StudentClassroomAssociator.run(quill_user, quill_classroom)
+        ActiveRecord::Base.transaction do
+          StudentsClassrooms.unscoped.find_or_create_by!(student: quill_user, classroom: quill_classroom)
+
+          ClasslinkClassroomUser.find_or_create_by!(
+            classroom_external_id: incoming_enrollment[:classroom_sourced_id],
+            user_external_id: incoming_enrollment[:user_sourced_id]
+          )
+        end
```
anathomical, line 46: Is it worth putting this in a method `create_classroom_student` or something to match the pattern of `create_classroom_teacher`?

services/QuillLMS/app/services/classlink_integration/user_persister.rb
```diff
@@ -0,0 +1,44 @@
...
+    end
+
+    def run
+      creatable_user_hash = UserPersister.creatable_user(incoming_user)
+
+      if existing_user
+        existing_user.user.update!(**creatable_user_hash)
+      else
+        ActiveRecord::Base.transaction do
+          User.transaction do
```
anathomical: Do we need two layers of `transaction` declarations here?

services/QuillLMS/app/services/classlink_integration/user_persister.rb
```diff
@@ -0,0 +1,44 @@
...
+            user = User.create!(**creatable_user_hash)
```
anathomical: I think there may be an edge case here where we'll get validation failures if a non-classlink user account already exists in the Quill database for the email address we get from Classlink.  In that case, `existing_user` won't find a result (since there's no matching `external_id`), so `create!` will try to make a new user with a duplicate email.

services/QuillLMS/config/initializers/auth/classlink.rb
```diff
@@ -0,0 +1,9 @@
+# frozen_string_literal: true
+
+module Auth
+  module Classlink
+    CLIENT_ID = 'c17370377633256aee86a52043b96d2913e27b36df24'
```
anathomical: No strong feelings on this, but should we make this an ENV variable, too?  If we had a staging Classlink environment, would we have the same `CLIENT_ID` or would it change?

services/QuillLMS/db/migrate/20250306185139_create_classlink_accounts.rb
```diff
@@ -0,0 +1,12 @@
+# frozen_string_literal: true
+
+class CreateClasslinkAccounts < ActiveRecord::Migration[7.1]
+  def change
+    create_table :classlink_accounts do |t|
+      t.string :external_id, null: false
+      t.references :user, foreign_key: true, null: false
+      t.references :classlink_instance, foreign_key: true, null: false
```
anathomical, line 8: Not an objection, but we've tended to be cautious about declaring `foreign_key` values due to the additional overhead it imposes on database operations.  Are we getting enough value out of the extra validation here to be worth the computational overhead?

services/QuillLMS/db/migrate/20250307233440_create_classlink_instances.rb
```diff
@@ -0,0 +1,16 @@
...
+
+class CreateClasslinkInstances < ActiveRecord::Migration[7.1]
+  def change
+    create_table :classlink_instances do |t|
+      t.integer :application_id, null: false
+      t.boolean :enabled, null: false
+      t.string :oneroster_application_id, null: false
+      t.string :name, null: false
+      t.string :tenant_name, null: false
+      t.references :district, foreign_key: true, null: false
```
anathomical: Again with the foreign key, though I'll point out that `districts` should be a small enough table that any costs imposed here are very small.

services/QuillLMS/spec/models/app_setting_spec.rb
```diff
@@ -22,6 +22,25 @@
 RSpec.describe AppSetting, type: :model do
   let(:user) { create(:user) }
 
+  describe '#logged_out_enabled?' do
+    let!(:app_setting) do
+      create(:app_setting, name: 'egosum', enabled: true, percent_active: 100)
```
anathomical: The repeated use of `'egosum'` in this spec suggests to me that it's probably worth extracting into `let(:name) { 'egosum' }` to avoid risks around typo-ing a repeated constant.

That would also let you use: `subject { AppSetting.logged_out_enabled?(name:) }` in all of the tests.

services/QuillLMS/spec/models/app_setting_spec.rb
```diff
@@ -22,6 +22,25 @@
...
+    end
+
+    context 'AppSetting does not exist' do
+      it { expect(AppSetting.logged_out_enabled?(name: 'nemo')).to be false }
```
anathomical, line 31: If you use my above suggestion, you can use `let(:name) { 'nemo' }` with `subject` in this test.

services/QuillLMS/spec/services/classlink_integration/user_persister_spec.rb
```diff
@@ -0,0 +1,60 @@
...
+          classlink_instance:)
+      end
+      let!(:existing_provider_users) { [classlink_account] }
+
+      it 'updates the attributes of the existing user' do
+        expect { subject.run }
+          .to not_change(ClasslinkAccount, :count)
+          .and not_change(User, :count)
+        existing_user.reload
+        expect(existing_user.email).to eq('foo@bar.com')
```
anathomical: I think you can actually chain this expectation onto the `change` chain:
```ruby
expect { subject.run }
  .to not_change(ClasslinkAccount, :count)
  .and not_change(User, :count)
  .and change(existing_user.reload, :email).to eq('foo@bar.com')
```

services/QuillLMS/spec/services/classlink_integration/schools_fetcher_spec.rb
```diff
@@ -0,0 +1,69 @@
...
+  describe '#transform_api_response' do
+    it 'should return a hash with the district_id and name' do
+      sample_input = {
+        'sourcedId' => '12',
+        'status' => 'active',
+        'name' => 'Hawkins HS'
+      }
+
+      expect(subject.transform_api_response(sample_input)).to eq({
+        district_id: 1,
```
anathomical: This test does not clearly communicate whether we expect this value to be equal to the value of `district_oneroster_id` or `district_id`.  I presume the latter, but could imagine the former.  I think you could clarify this ambiguity by using `let` to name whichever one you want to confirm the `to eq` value of.
```ruby
# assuming it's district_id we care about
subject { described_class.new(district_oneroster_id: 1, district_bearer_token: 1, district_id:) }

let(:district_id) { 1 }

expect(...).to eq({
  district_id:,
  ...
})
```

services/QuillLMS/spec/services/classlink_integration/classroom_persister_spec.rb
```diff
@@ -0,0 +1,85 @@
...
+      end
+      let(:existing_provider_classrooms) { [classlink_classroom] }
+
+      it 'updates the attributes of the existing classroom' do
+        subject
+        expect { subject.run }.to not_change(ClasslinkClassroom, :count)
+
+        existing_classroom.reload
+        expect(existing_classroom.name).to eq('Math 101')
+        expect(existing_classroom.visible).to be true
```
anathomical, line 41: Take or leave, but I think you can make this all part of the `to change` chain:
```ruby
expect { subject.run }.to not_change(ClasslinkClassroom, :count)
  .and change(existing_classroom.reload, :name).to(incoming_classroom[:name]) # note that I also prefer referencing the saved value rather than typing the constant a second time
  .and change(existing_classroom.reload, :visible).to(true)
```

services/QuillLMS/spec/services/classlink_integration/classroom_persister_spec.rb
```diff
@@ -0,0 +1,85 @@
...
+      let(:existing_provider_classrooms) { [matching_classlink_classroom, non_matching_classlink_classroom] }
+
+      it 'only updates the matching classroom' do
+        subject.run
+        matching_classroom.reload
+        non_matching_classroom.reload
+
+        expect(matching_classroom.visible).to be true
+
+        expect(non_matching_classroom).not_to receive(:update!)
```
anathomical: Does this expectation trigger as expected when it is declared after the `subject.run` call?  I've always thought that `expect (not) to receive` had to be declared before the code that might call it, but sometimes `expect` matchers work differently than I expect.

anathomical (approved):
Some minor notes, most of them take-or-leave, or even just points of clarification.  Overall, I like the approach you've taken here, and appreciate how complex this project must be.

services/QuillLMS/app/models/classroom.rb
happythenewsad: Thanks for noticing that. Neither of these methods execute, since `classroom_external_id` is overridden lower in the file. I removed these lines

services/QuillLMS/app/models/classroom.rb
happythenewsad: Yeah, just made this a bit DRYer

services/QuillLMS/app/services/classlink_integration/districts_importer.rb
happythenewsad: It's fine for multiple districts to share the same name, from Classlink's perspective, but I'm adding the bang anyway

services/QuillLMS/app/services/classlink_integration/user_persister.rb
happythenewsad: no; removing

services/QuillLMS/app/services/classlink_integration/user_persister.rb
happythenewsad: Yes, handling this case is part of the spec, but not part of this PR. I should have added an interim clarifying comment, sorry.

https://www.notion.so/quill/ClassLink-Project-160d42e6f941806ea889ca975a685a7a?pvs=4#1bbd42e6f941803bb9fafd8a8d081dd0

services/QuillLMS/db/migrate/20250306185139_create_classlink_accounts.rb
happythenewsad: Kind of a judgment call. I'm inclined to use the extra validation for now + do the rails convention, and optimize if it ever becomes a problem.

Two arguments in support of this:
- the database access pattern for this table will be: read often, create once, and update infrequently
- Classlink will have low usage to start, and we don't know if it will become super popular

services/QuillLMS/db/migrate/20250307233440_create_classlink_instances.rb
happythenewsad: Adding to eng discussion

services/QuillLMS/app/services/classlink_integration/base_fetcher.rb
happythenewsad: For now, I want to minimize requests, since Classlink mentions that it rate limits APIs, but does not advertise exact rate limits: https://help.classlink.com/s/article/rs-data-best-practices#complete

services/QuillLMS/app/services/classlink_integration/enrollment_persister.rb
happythenewsad: Yes, except that I'm replacing this code block with an Associator - see comment

services/QuillLMS/spec/services/classlink_integration/classroom_persister_spec.rb
happythenewsad: It's an interesting pattern. I changed the previous spec to use it, per your suggestion, but it seems harder to parse here, when the lines are longer.

dandrabik (approved):
Looks good. Thanks for making the changes and talking this through. 🚢

PR #12769, opened 2025-02-04
remove-adgr-benchmark-cron

anathomical:
## WHAT
Remove Admin Diagnostic performance benchmark code, both removing scheduled cron calls to trigger the code, and the code itself since it's not called anywhere else

NOTE: Once merged, we'll want to clean up the `SLACK_ADGR_PERFORMANCE_BENCHMARKS_EMAIL` ENV variable from production, as well as deleting the email-reception integration to the `#adgr-performance-benchmarks` Slack channel.  We likely want to retain that channel to keep access to the historical data if we ever want it.
## WHY
We're no longer analyzing the benchmark results, so we should disable this code and clean it up rather than leave it around to clutter up our codebase
## HOW
- Remove job calls in the cron model
- Delete the worker and mailer files along with their associated specs

- Deploy to staging, confirm that the Admin Diagnostic reports still run as expected
- These cron jobs only run in prod, so we can only test that this is actually working once it's deployed, but none of this code results in user-facing changes

PR Checklist | Your Answer
------------ | -------------


brendanshean (approved):
Thank you for cleaning up this code.

happythenewsad (approved):
🧹 🧹 🧹

PR #12778, opened 2025-02-06
Heroku pipelines - promote CLI

happythenewsad:
## WHAT
adds `promote` option to `deploy.sh` 

## WHY
In the previous eng meeting, we resolved to try this workflow (aka `cowboy+`) 

## HOW
See code annotations. 

### Screenshots

### Notion Card Links
https://www.notion.so/quill/Heroku-pipelines-LMS-implementation-71f771ab827f46bda1457d3bf183ac7a

- ran script in normal mode, observed successful staging build 
- ran script in promote mode (dry run), observed correct notifications

PR Checklist | Your Answer
------------ | -------------


services/QuillLMS/deploy.sh
```diff
@@ -71,6 +68,41 @@ case $1 in
...
+
+for arg in "$@"; do
+  case $arg in
+    --osx-notification)
+      osx_notification=1
+      ;;
+  esac
+  case "$arg" in
+    -p|--promote)
+      promote
```
happythenewsad, line 100: The `--promote` flag intercepts the normal flow, then exits, so the normal deploy flow is untouched.

services/QuillLMS/deploy.sh
```diff
@@ -71,6 +68,41 @@ case $1 in
     exit 1
 esac
 
+promote() {
+  git fetch origin
+  local latest_prod_release_sha=`heroku releases -a empirical-grammar | grep Deploy | cut -d ' ' -f 4 | head -n 1`
+  local promotion_candidate_sha=`heroku releases -a $HEROKU_APP | grep Deploy | cut -d ' ' -f 4 | head -n 1`
+
+  # if promotion candidate is not a commit descendent of the current production slug, abort
+  git merge-base --is-ancestor $latest_prod_release_sha $promotion_candidate_sha
```
happythenewsad, line 77: I iterated on the safety mechanism a couple times, first with timestamps, then with git tags. Those other mechanisms are viable, but I believe that directly checking for ancestry is the cleanest and most precise method. 

We are able to use this method now because our prod branch, now called `develop`, is fast-forward (no merges, only squashes). But even in the case where a conventional merge to `develop` happens (someone misclicks the `Merge` button and does not specify `squash`, this method fails gracefully: the promotion will not happen, and the deployer will be nudged to promote a different slug, or deploy conventionally.

brendanshean: Do we want to disable merge commits in the UI flow?

brendanshean (approved):
This looks very clean.  I was unaware of the `-is-ancestor`, good find on that.

services/QuillLMS/deploy.sh
happythenewsad: We could; there's a conceptual elegance to that that is appealing. I'm ultra-cautious of adding checks that could prevent us from getting new code to production in an emergency scenario or in some weird edge case I haven't anticipated, so I opted not to do that here. To me, the minor annoyance of getting my promotion rejected by the CLI is acceptable.

anathomical (approved):

PR #12779, opened 2025-02-06
diagnostic insights history by class

emilia-friedberg:
## WHAT
Build out the frontend for the Diagnostic Insights: History By Class page.

## WHY
This is the first full page of this project!

## HOW
Just React and CSS.

### Screenshots
Screenshot 2025-02-06 at 11 02 02 AM

### Notion Card Links
https://www.notion.so/quill/Diagnostic-Insights-History-by-Class-190d42e6f941806caf67d7b5b1c0104b?pvs=4

Tested locally and on staging, Peter Sharkey will QA on staging as well.

PR Checklist | Your Answer
------------ | -------------


services/QuillLMS/client/app/bundles/Teacher/components/progress_reports/diagnostic_insights/history.tsx
```diff
@@ -0,0 +1,201 @@
...
+        icon = pre_diagnostic[0].proficient || post_diagnostic[0].proficient ? filledBlueDiamondWithCheckMarkImg : filledBlueDiamondImg
+      } else {
+        icon = pre_diagnostic[0].proficient ? outlinedBlueDiamondWithCheckMarkImg : outlinedBlueDiamondImg
+      }
+
+      const tooltipText = []
+
+      // Baseline Pre Completed
+      if (pre_diagnostic?.length) {
+        const preDiagnosticCompletionDate = moment(pre_diagnostic[pre_diagnostic.length - 1].completed_at).format("MM/DD/YY");
```
happythenewsad: We could put this format param in a constant somewhere, since it's being used many places

services/QuillLMS/client/app/bundles/Teacher/components/progress_reports/diagnostic_insights/history.tsx
```diff
@@ -0,0 +1,201 @@
...
+          <p>
+            <strong>Baseline Pre Practice:</strong> {pre_recommendation.length} activities (
+            {firstPracticeDate} - {lastPracticeDate})
+          </p>
+        );
+      }
+
+      // Growth Post Completed
+      if (post_diagnostic?.length) {
+        const postDiagnosticCompletionDate = moment(post_diagnostic[post_diagnostic.length - 1].completed_at).format("MM/DD/YY")
```
happythenewsad: `moment` is in [maintenance mode til EOL](https://momentjs.com/docs/#/-project-status/). 

I think we should not add momentjs code to any codebases, and instead use the current builtins or the [shim ](https://github.com/tc39/proposal-temporal)for the upcoming native timezone-friendly builtin ("Temporal") . Happy to debate this though.

services/QuillLMS/client/app/bundles/Teacher/components/progress_reports/diagnostic_insights/sidebar.tsx
```diff
@@ -16,9 +16,22 @@ const Sidebar = ({
   selectedClassroomId,
   setSelectedClassroom
 }) => {
+  const navigate = useNavigate();
+
+  React.useEffect(() => {
```
happythenewsad, line 21: Style/opinion thing: I find it surprising that a component called `Sidebar` has the ability to mutate the URL upon rendering, absent a user interaction

services/QuillLMS/client/app/bundles/Teacher/components/progress_reports/diagnostic_insights/__tests__/history.test.tsx
```diff
@@ -0,0 +1,183 @@
...
+
+  const mockStudentHistories = [
+    {
+      user_id: 1,
+      user_name: 'Alice Johnson',
+      student_learning_sequences: {
+        starter: [],
+        intermediate: [],
+        advanced: [],
+      },
```
anathomical: I don't think this will matter for your purposes, but my currently open PR to add students with no sequences to the dataset provides this data as `student_learning_sequences: {}` without the `starter` and other keys if they're empty.

services/QuillLMS/client/app/bundles/Teacher/components/progress_reports/diagnostic_insights/__tests__/history.test.tsx
```diff
@@ -0,0 +1,183 @@
...
+
+    const dataTable = screen.getByRole('table');
+
+    await waitFor(() => {
+      expect(within(dataTable).getByText(studentWithDiagnostics.user_name)).toBeInTheDocument();
+      expect(within(dataTable).getByAltText(FILLED_BLUE_DIAMOND_ALT_TEXT)).toBeInTheDocument();
+      expect(within(dataTable).getByAltText(FILLED_BLUE_DIAMOND_WITH_CHECKMARK_ALT_TEXT)).toBeInTheDocument();
+      expect(within(dataTable).getByAltText(OUTLINED_BLUE_DIAMOND_ALT_TEXT)).toBeInTheDocument();
+      expect(within(dataTable).getByAltText(OUTLINED_BLUE_DIAMOND_WITH_CHECKMARK_ALT_TEXT)).toBeInTheDocument();
+    });
```
anathomical, line 121: I'm not sure it's worth doing more than this, but I will note that this particular test doesn't confirm which data states correspond to which icons being in the document, just that if the data has all conditions, all icons appear.  It might be worth doing this as a series of tests with different input data, but that could be more work than it's worth, so take that or leave it.

services/QuillLMS/client/app/bundles/Teacher/components/progress_reports/diagnostic_insights/history_and_recommendations.tsx
```diff
@@ -2,23 +2,38 @@ import * as React from 'react';
...
+  },
+  {
+    label: 'Last school year',
+    value: 'last-year'
+  },
+  {
+    label: 'This school year',
+    value: 'this-year'
+  }
+]
```
anathomical: Thoughts on me building this array on the back-end and fetching it from a new API endpoint to insure we don't have to update both front-end and back-end when we add a new option, or is that overkill since this works just fine?

services/QuillLMS/client/app/bundles/Teacher/components/progress_reports/diagnostic_insights/interfaces.ts
```diff
@@ -1,32 +1,46 @@
...
+  code: string;
+  created_at: string;
+  updated_at: string;
+  clever_id: number|null;
+  grade: string;
+  grade_level: string|null;
+  visible: boolean;
+  google_classroom_id: number|null;
+  synced_name: string;
+}
```
anathomical, line 64: This feels like it might be useful to have defined in some shared path since it's broader than just this project, but that may not be worth worrying about at this time.

services/QuillLMS/client/app/bundles/Teacher/components/progress_reports/diagnostic_insights/history.tsx
```diff
@@ -0,0 +1,201 @@
...
+  const renderDiagnosticSquares = (studentLearningSequences: StudentLearningSequenceInterface[]) => {
+    if (!studentLearningSequences) { return null }
+
+    return studentLearningSequences.map(studentLearningSequence => {
+      const { student_learning_sequence_activities, } = studentLearningSequence
+      const { pre_diagnostic, post_diagnostic, pre_recommendation, post_recommendation, } = student_learning_sequence_activities
+
+      // sometimes there is a studentLearningSequence when a student has completed an activity that belongs to the recommendation sequence,
+      // even if they didn't complete either of the diagnostics, in which case we don't want to render anything
+      if (!pre_diagnostic) { return null }
```
anathomical, line 71: This is probably something we should suppress on the back-end.  I'll make a card to do that and then remove this check once that's done, but putting it in the front-end for now makes sense to me.

services/QuillLMS/client/app/bundles/Teacher/components/progress_reports/diagnostic_insights/history.tsx
anathomical: We might also decide to pass these around as seconds-since-epoch integers if that will make conversions easier on the front-end.

services/QuillLMS/client/app/bundles/Teacher/components/progress_reports/diagnostic_insights/history.tsx
```diff
@@ -0,0 +1,201 @@
...
+        />
+      )
+    }).filter(Boolean)
+  }
+
+  const rows = studentHistories.map(studentHistory => {
+    const { user_id, user_name, student_learning_sequences, } = studentHistory
+    const { starter, intermediate, advanced, } = student_learning_sequences
+    const nameAsArray = user_name.split(' ')
+    const nameForSort = `${nameAsArray[nameAsArray.length - 1]} ${nameAsArray[0]}`
```
anathomical, line 140: We're doing explicit sorting on the backend.  It makes sense to also allow sorting on the front-end so that it's easy to reverse order in the UI, but it makes me wonder if we're getting value out of sorting on the back-end too.  I don't see a pressing reason to stop, but figured I'd ask since it feels a little bit like we're doing it twice.

anathomical (approved):
This looks great!  So glad you got it turned around so quickly, and I'm excited to see this get into some early testing.

services/QuillLMS/client/app/bundles/Teacher/components/progress_reports/diagnostic_insights/history.tsx
anathomical: I also agree that adding more `moment` to our codebase should be avoided if it's not onerous to do so.

eadams17 (approved):

services/QuillLMS/client/app/bundles/Teacher/components/progress_reports/diagnostic_insights/history.tsx
emilia-friedberg: I think it's still useful to have them sorted on the backend, because there are other cases where we'll use the student names in order (like in dropdowns for selection) where it would be an additional step to have to format them on the frontend.

services/QuillLMS/client/app/bundles/Teacher/components/progress_reports/diagnostic_insights/history_and_recommendations.tsx
emilia-friedberg: I think it might be more work than necessary, but another option is to put it in some kind of helper file and then just pass it up as a prop to the main component here. If you want to get it into a helper that's accessible from the view, I'm happy to make that change.

services/QuillLMS/client/app/bundles/Teacher/components/progress_reports/diagnostic_insights/__tests__/history.test.tsx
emilia-friedberg: Yeah, you're right. I added a snapshot as kind of an in-between here (that should ensure the positioning is correct, and if it updates unexpectedly we'll know something is awry).

services/QuillLMS/client/app/bundles/Teacher/components/progress_reports/diagnostic_insights/interfaces.ts
emilia-friedberg: Yeah, I think it might be worthwhile, but we don't really have standard interfaces app-wide right now so I'd need to do some thinking about how to organize that.

services/QuillLMS/client/app/bundles/Teacher/components/progress_reports/diagnostic_insights/sidebar.tsx
emilia-friedberg: I'm not sure I understand -- would it be clearer if I renamed this component `Navigation`, since that's effectively what it is (a sidebar with links)?

services/QuillLMS/client/app/bundles/Teacher/components/progress_reports/diagnostic_insights/history.tsx
emilia-friedberg: Good call-out. I'm replacing it with a built-in and standardizing it as a function since it's less succinct than the moment formatting.

happythenewsad (approved):
Thanks for shedding momentjs!

services/QuillLMS/client/app/bundles/Teacher/components/progress_reports/diagnostic_insights/sidebar.tsx
happythenewsad: I don't think renaming the component would make the behavior less surprising. It's just something I haven't encountered before

PR #12781, opened 2025-02-06
diagnostic-insights-include-students-without-completions

anathomical:
## WHAT
Fix two issues, one identified by Emilia, and a second one that came up during my additional testing:
- Students who had not completed any diagnostics were not showing up in the data, but we want them to be included so that they can be shown as having no completions
- Students who have been removed from a classroom were continuing to show up in the report, but we don't want a teacher to keep having to see data for a student that they've removed from their classroom
## WHY
We want this API to produce all of the data that the front-end needs in a single call so that it doesn't have to cobble together things like which students to show with no data
## HOW
- Use `User` as the query root instead of `StudentLearningSequence` to include users even if they have no `StudentLearningSequence` records (this also lets us remove `group_by(&:user_id)`)
- Filter out "out of date range" `StudentLearningSequenceActivity` records at the Serializer, relying on `includes` to let us avoid making additional database queries (this feels awkward, but none of the other approaches I tried worked without getting incredibly hard to follow with tons of custom `Arel` code)
- - Note that I'm totally open to doing this another way given the surprise I imagine this code would cause me if someone else wrote it, so if you've got suggestions, I'm all ears
- Filter out `StudentLearningSequence` records for a User if there aren't in-date-range `StudentLearningSequenceActivity` records associated with it so that the API stays easy-to-read in "no data" scenarios

- Once again, update the snapshot and ensure that the changes to the JSON payload correspond with expected changes
- - Examine new users added to payloads to confirm that they are in the class, but never took the diagnostic in question
- - Examine users removed from the payloads to confirm that they have `StudentsClassrooms` records tying them to the classroom, but with `visible: false`

PR Checklist | Your Answer
------------ | -------------


services/QuillLMS/app/serializers/diagnostic_insights/student_learning_sequence_serializer.rb
```diff
@@ -2,16 +2,31 @@
 
 module DiagnosticInsights
   class StudentLearningSequenceSerializer < ActiveModel::Serializer
-    attributes :id, :created_at, :initial_activity_id, :user_name, :user_id
+    attr_reader :start_date, :end_date
 
-    has_many :student_learning_sequence_activities do
+    attributes :id, :created_at, :initial_activity_id, :user_name, :user_id, :student_learning_sequence_activities
+
+    def initialize(object, start_date = nil, end_date = nil)
```
brendanshean: I think the AMS pattern for [custom options](https://github.com/rails-api/active_model_serializers/tree/0-10-stable?tab=readme-ov-file#high-level-behavior) is via an options hash:

ActiveModelSerializers::SerializableResource.new(resource, options ={})

So in this case you'd probably want:     `def initialize(object, start_date: nil, end_date: nil)`

brendanshean (approved):
Nice simplification of grouping and filtering of student learning sequence queries.  I also appreciate the thorough testing.

emilia-friedberg (approved):
Thanks for that quick fix, and for finding the other issue!

happythenewsad (approved):

dandrabik (approved):
:+1:

PR #12790, opened 2025-02-12
Extract some MaterializedView code to mixins.

dandrabik:
## WHAT
Small Refactor to the MaterializedView code to use Mixins instead of inheritance.

I also updated the SQL snapshots which were a little out of date.
## WHY
I'm experimenting with using generated tables from queries instead of MaterializedViews, but the current structure was making it a little hard to work with and it was very duplicative. I extracted this refactor out of the feature branch to keep it isolated. 

Make the change easy, then make the easy change.
## HOW
Use concerns instead of inheritance and group features into discrete concerns. This is the major change, the rest is code to make this work
Before:
```ruby
DiagnosticAggregateViewQuery < ::QuillBigQuery::MaterializedViewQuery
```
After:
```ruby
DiagnosticAggregateViewQuery < ::QuillBigQuery::Query
    include ::QuillBigQuery::Concerns::ViewQueryFallback
    include ::QuillBigQuery::Concerns::MaterializedViewQueryable
```

### Notion Card Links
This is part of the BigQuery Billing investigation.

- I deployed to staging and ensured that the admin diagnostic report and snapshot report ran properly.
- I ran the local thor task for admin_diagnostics and snapshots. There were some SQL changes, but that seems to be because the snapshots were out of date. All of the SQL changes are because of previous code changes, plus a test user seems to have a new school district.

PR Checklist | Your Answer
------------ | -------------


services/QuillLMS/spec/queries/quill_big_query/concerns/view_query_fallback_spec.rb
```diff
@@ -2,34 +2,54 @@
...
+
+      define_method(:fallback_views) do
+        [@materialized_view]
+      end
+
+      define_method(:runner) do
+        @runner
+      end
+
+      def post_query_transform(result) = result
```
anathomical, line 33: Should we use `define_method` here to be consistent with the rest of the methods in this test class?

services/QuillLMS/lib/query_examples/admin_diagnostics/diagnostic-students-view.sql
```diff
@@ -1,19 +1,19 @@
...
-           BI Engine Mode Used:     FULL_INPUT
-             BI Engine Code:          
-             BI Engine Message:       
+           Data Processed By Query: 2.49 GB
+           Bytes Billed For Query:  2.49 GB
+           Total Query Time:        2509 ms
+           Total Slot Time:         15991 ms
+           BI Engine Mode Used:     BI_ENGINE_DISABLED
+             BI Engine Code:          INPUT_TOO_LARGE
+             BI Engine Message:       Cannot broadcast table lms.active_user_names_view: number of rows 9636555 > supported limit of 8000000.
```
anathomical, line 8: This falling out of BI Engine is concerning.  I wonder if we've had data size growth pushing `active_user_names_view` out of BI Engine, or if this is just one of those cases where sometimes queries get this `INPUT_TOO_LARGE` error, but usually don't, and that this would be in BI Engine if it were re-run.

anathomical (approved):
I concur with your analysis that these snapshots are just out of date.  They look like they should be correct in their new form.  One thing I did note going through is that the "Data Processed By Query" value for most of these went up by 50~100%.  That seems like a pretty significant growth in data, which may be a contributor to the growing query costs.  Especially if things that were fitting in BI Engine before are starting to fall out without notifying us.

services/QuillLMS/app/queries/quill_big_query/concerns/view_query_fallback.rb
```diff
@@ -0,0 +1,48 @@
...
+        ::Google::Cloud::InvalidArgumentError
+      ]
+
+      def fallback_views = raise NotImplementedError
+      def run_query = run_query_with_fallback
+
+      def run_query_with_fallback
+        query_runner(query)
+      rescue *BROKEN_VIEW_ERRORS => e
+        ErrorNotifier.report(BrokenViewError.new(e.message))
```
brendanshean, line 21: Is there any more context that would be useful to include here,  such as view key or query?  (It might be in e.message but I can't really tell).

services/QuillLMS/app/queries/admin_diagnostic_reports/diagnostic_aggregate_view_query.rb
```diff
@@ -26,6 +29,7 @@ def initialize(timeframe_start:, timeframe_end:, school_ids:, aggregation:, grad
     end
 
     def materialized_views = [filter_view, performance_view]
+    def fallback_views = materialized_views
```
brendanshean, line 32: This method feels like a passthrough.  Maybe remove?

services/QuillLMS/app/queries/quill_big_query/concerns/configurable_view.rb
brendanshean, line 1: I like this declaration of views.  Very clear.

brendanshean (approved):
I think the switch from inheritance to composition has improved the clarity and structure of the code.  This should make future changes around queries easier.  Nice work

services/QuillLMS/app/queries/admin_diagnostic_reports/diagnostic_aggregate_view_query.rb
dandrabik: This is probably harder to see without the upcoming code, but `fallback_views` is used by `ViewQueryFallback`. I'm decoupling that code from `materialized_views` so that there can be multiple types , e.g. the upcoming `table_from_query_views`. A class can assign any combination of `ConfigurableViews` here.

services/QuillLMS/app/queries/quill_big_query/concerns/view_query_fallback.rb
dandrabik: The message logs the full query. There might be additional info we could like like query class. Let me think on that. Long term, I'm debating moving away from this query fallback anyway if we fully move off of MaterializedViews (we'd adjust the tables to only refresh once a day overnight, so there'd only be about 30 seconds in the middle of the night where the query might error.

services/QuillLMS/lib/query_examples/admin_diagnostics/diagnostic-students-view.sql
dandrabik: Yeah, it's unclear what is happening. The query has been edited also, so that could have caused it too.

services/QuillLMS/spec/queries/quill_big_query/concerns/view_query_fallback_spec.rb
dandrabik: So the `define_method` is only needed for methods that need data from initialize. I've edited the other static method to use the normal `def` setup.

PR #12792, opened 2025-02-14
Updated AminDiagnostic snapshots.

dandrabik:
## WHAT
Re-run the AdminDiagnostic snapshots on `develop` to get latest numbers
## WHY
I have a PR with query changes so I want an accurate baseline to compare to.
## HOW
`bundle exec thor queries:snapshot_diagnostic_queries`

No user facing features.

PR Checklist | Your Answer
------------ | -------------
Have you added and/or updated tests? |  No, these are sorta a regression test though.
Self-Review: Have you done an initial self-review of the code below on Github? | Yes, all generated files.


anathomical (approved):
This data should, in theory, be pretty static.  I suspect that the changes in numbers are likely coming from things like `students_classrooms` changing, and that data not being "snapshot-able".  I wonder if it's a signal that we should rethink some of our data models to support better "historical" data snapshots.  That might involve additional denormalization.

Anyway, that's well beyond the scope of this PR, and this one looks fine to me.

dandrabik:
> This data should, in theory, be pretty static. I suspect that the changes in numbers are likely coming from things like `students_classrooms` changing, and that data not being "snapshot-able". I wonder if it's a signal that we should rethink some of our data models to support better "historical" data snapshots. That might involve additional denormalization.
> 
> Anyway, that's well beyond the scope of this PR, and this one looks fine to me.

Yeah, ok, thanks for looking through this. I was seeing changes in these numbers in a branch, so I wanted to isolate them so we know what is caused by code changes, and what is caused by data shifting over time.

PR #12793, opened 2025-02-14
Remove horizontal scrollbar in ManageActivities UI

brendanshean:
## WHAT
Remove the horizontal scrollbar from "My Open Activity Packs" and "My Closed Activity Pack" pages in the 'Manage Activities' tab.

## WHY
The content fits already so there should not be a horizontal scrollbar.

## HOW
`overflow: scroll;` forces scrollbars for [both axes](https://developer.mozilla.org/en-US/docs/Web/CSS/overflow#scroll), even if the content fits perfectly. 
To fix, we provide settings for each axis:
`overflow-x: hidden;` hides the horizontal scrollbar
`overflow-y: auto;`, shows the vertical scrollbar when needed.

### Screenshots

### Notion Card Links
https://www.notion.so/quill/Unneeded-scroll-element-on-the-Manage-Activities-My-Open-Activity-Packs-My-Closed-Activity-pages-19ad42e6f9418028a338f23d84eaedfd?pvs=4

I have verified on staging that the horizontal scroll bar is no longer present for these components.

PR Checklist | Your Answer
------------ | -------------
Have you added and/or updated tests? |  NO.


emilia-friedberg (changes requested):
This solution works for normal screen sizes, but I think we do want to allow scrolling at smaller viewpoints, so I would take a different approach here. Using dev tools, I added a rule: `.activity-pack * { border: 1px solid black; }`. That shows this:

Screenshot 2025-02-18 at 8 42 04 AM

Which makes it clear that it's the tooltips from the buttons in the top right corner that are exceeding the set width and causing the scroll. The easiest fix here is to add a rule like:

`.activity-pack .quill-tooltip-wrapper { max-width: 250px }`

Which is fairly minimal, but allows us to remove the scroll at standard screen widths without compromising smaller viewports.

brendanshean:
> This solution works for normal screen sizes, but I think we do want to allow scrolling at smaller viewpoints, so I would take a different approach here. Using dev tools, I added a rule: `.activity-pack * { border: 1px solid black; }`. That shows this:
> 
> Screenshot 2025-02-18 at 8 42 04 AM
> Which makes it clear that it's the tooltips from the buttons in the top right corner that are exceeding the set width and causing the scroll. The easiest fix here is to add a rule like:
> 
> `.activity-pack .quill-tooltip-wrapper { max-width: 250px }`
> 
> Which is fairly minimal, but allows us to remove the scroll at standard screen widths without compromising smaller viewports.

Thank you for the suggestion as well as the explanation with bounding boxes.

emilia-friedberg (approved):

eadams17 (approved):

PR #12794, opened 2025-02-15
Fix activity versions bug

brendanshean:
## WHAT
Fix a bug involving end date of activity versions.

## WHY
Activity versions sometimes skip a number, (e.g. version 3, version 5 but there is no version 4) and this causes an error with calculating the `next_change_log`

## HOW
Update the `next_change_log` method to look for the next highest number, rather than just +1 the current value.

Utilize postgres `new_value::integer` [syntax](https://neon.tech/postgresql/postgresql-tutorial/postgresql-cast#postgresql-cast-operator)

Also, in attempting to refactor the RSpec tests to use conventional patterns, I uncovered an error with `assert` method (basically it doesn't work the way we expect it to).   I'd like to leave a note here for now and get this fix up first and deal with that after some discussion.

### Screenshots

### Notion Card Links
https://www.notion.so/quill/Dogs-Sense-of-Smell-Activity-Stats-version-filtering-isn-t-using-the-correct-end-date-19ad42e6f941801bb933fb51ed423891?pvs=4

I checked that the Activity Versions are calculated properly for a few activities.

PR Checklist | Your Answer
------------ | -------------


anathomical (approved):
Interesting find re: `assert`

PR #12796, opened 2025-02-18
BigQuery - TableFromQueryView

dandrabik:
## WHAT
BigQuery - Use a generated `Table` instead of a `MaterializedView` for the largest and most data-intensive/expensive view (if this works well, we'll likely replace the rest of the `MaterializedViews`).

## WHY
This should give us better performance and cost less because:
1. Generated `Table`s don't refresh themselves (this is a big source of data usage)
2. We can put more columns in a `Table`  than a MaterializedView, to make queries more efficient. We can also cluster and partition (more experiments on that to follow).
3. We can likely reduce the size of our BI Engine cache since underlying tables don't also need to be in cache for a table

The only downside is that the data will refresh less (Materialized Views refresh in the background) so the data will be slightly more stale (we will likely end up only refreshing once per day), but given our cost cutting goals, this is a good trade-off.
## HOW
Introduce a new `GeneratedView` called `TableFromQueryView`

Replace the `pre_post_diagnostic_skill_group_performance_view` `MaterializedView` with a  `TableFromQueryView`:

| |Before|After|
|--|--|--|
|Type|MaterializedView(s)|TableFromQueryView|
|Views|pre_post_diagnostic_skill_group_performance_view, school_classroom_teachers_view| pre_post_diagnostic_skill_group_performance_table|
|Clustering?| Not allowed | Yes|
|Partitioning?| Not allowed | Available, but not used yet|

To make these both `MaterializedViews` and `TableFromQueryViews`, I've abstracted some of the mixins and renamed them `GeneratedViews...`

### Notion Card Links
https://www.notion.so/quill/BigQuery-use-tables-instead-of-mat-views-198d42e6f94180c0a47ff4b1d12ff74d

Running the queries scripts locally to determine that the queries produce comparable results (and better performance). Deployed to staging and clicked through Diagnostic reports to ensure no errors.

PR Checklist | Your Answer
------------ | -------------


services/QuillLMS/app/controllers/quill_big_query/generated_views_controller.rb
```diff
@@ -1,7 +1,7 @@
 # frozen_string_literal: true
 
 module QuillBigQuery
-  class MaterializedViewsController < ApplicationController
+  class GeneratedViewsController < ApplicationController
```
happythenewsad, line 4: The old nomenclature described an implementation detail (mat views), which led to a lot of code diffs here that are only name changes. 

+1 for moving to a naming scheme that describes the function, rather than the implementation

services/QuillLMS/app/queries/quill_big_query/concerns/generated_view_query_fallback.rb
```diff
@@ -0,0 +1,47 @@
...
+      BROKEN_VIEW_ERRORS = [
+        ::Google::Cloud::NotFoundError,
+        ::Google::Cloud::InvalidArgumentError
+      ]
+
+      def run_query = run_query_with_fallback
+
+      def run_query_with_fallback
+        query_runner(query)
+      rescue *BROKEN_VIEW_ERRORS => e
```
happythenewsad: Are these the only probable emitted errors at this path in the execution? If not, consider a more liberal rescue matcher.

services/QuillLMS/app/queries/quill_big_query/concerns/generated_view_query_fallback.rb
```diff
@@ -0,0 +1,47 @@
...
+        ErrorNotifier.report(BrokenViewError.new(e.message))
+        query_runner(query_fallback)
+      end
+
+      private def query_runner(query)
+        post_query_transform(runner.execute(query))
```
happythenewsad: `post_query_transform` has previously been guaranteed to exist via inheritance from `query.rb`. Including this concern in a non-descendent of `Query` would cause an error.

happythenewsad (approved):
I'm excited about the data savings, per the `.sql` comments

anathomical (approved):

services/QuillLMS/app/queries/quill_big_query/concerns/generated_view_query_fallback.rb
anathomical: When we did our original testing for fallback logic, these were the only two errors that were emitted in cases where the referenced Materialized Views were unavailable.

services/QuillLMS/app/queries/quill_big_query/concerns/generated_view_query_fallback.rb
dandrabik: That's a good point, let me think about this. I originally had a `def post_query_transform = (raise NotImplementedError) but it was not ideal since it forced me to move the includes further up in the hierarchy than I wanted to avoid overwriting a base class method. This will be a loud error if the class doesn't have it, so I'm not too worried, but let me think on this.

PR #12797, opened 2025-02-18
print git log

happythenewsad:
## WHAT
When deploying, print the git log between the last deployed sha and the head of `develop`

## WHY
This replaces the visual diff that the `develop -> production` Github merge gave us 

### Screenshots

### Notion Card Links
eng meeting discussion

PR Checklist | Your Answer
------------ | -------------
Have you added and/or updated tests? |  n/a - no code change


services/QuillLMS/deploy.sh
```diff
@@ -117,6 +117,12 @@ then
         # Documented here: https://github.com/empirical-org/test_repo/blob/destination_branch/test_file.txt
         git fetch origin develop
         git fetch origin $DEPLOY_GIT_BRANCH
+
+        latest_prod_release_sha=`heroku releases -a empirical-grammar | grep Deploy | cut -d ' ' -f 4 | head -n 1`
```
happythenewsad, line 121: Note that when the previous deploy was a promotion, the sha of the promoted slug will differ from the sha of the merged PR (because of squash/rebase). Thus, `git log` will print the most recent commit as if it is not in production. I added the `--stat` flag to make this more obvious. Is this something we want?

Alternatives include:
- don't print anything / close this PR
- print the diff (we lose the commit messages this way)

brendanshean: This feels a little confusing.  I would be fine with either alternative you list (print nothing or just print diff). Sorry for the trouble.

PR #12799, opened 2025-02-19
activity-session-skill-performance

anathomical:
## WHAT
- Add service to calculate "skill performance" on diagnostic and recommendation activities
- Add a method to `ActivitySession` to utilize this new service to calculate `skill_performance` and store it in the `data` JSONB blob
- Add a worker to queue `skill_performance` calculation for ActivitySessions, and set it up so that this worker runs after all `ConceptResult` records for a newly-completed `ActivitySession` are persisted
- Update the `DiagnosticInsights::StudentLearningSequenceActivitySerializer` to use `ActivitySession.data['skill_performance']` for determining `proficiency` values, and also include skill performance in the returned payload
- - Include logic to mark diagnostic play-throughs as `proficient` as long as a student has demonstrated proficiency on at least a minimum percentage of skills in the diagnostic instead of requiring all skills to be proficient
- Add a utility Worker to backfill `skill_performance` data for `ActivitySession`s related to a specific teacher (so that we can target backfills of data for specific teachers for testing purposes)

NOTE: Beyond the scope of this PR, but the next thing I intend to work on making this calculation more efficient so that we can store these `skill_performance` values in the database for all historical `ActivitySession`s.  Currently this is not efficient enough to do that, processing on the order of 25/second.
## WHY
- For reporting we care about a level of reporting that's more granular than "how many questions you got right" on a diagnostic, but less granular than individual `concept_results`.
- We need to be able to provide more detailed reporting for Diagnostic Insights about which skills a student did and did not demonstrate proficiency for on a diagnostic as well as how they performed on any practice activities
- After discussion with Peter G and Peter S, and some analysis of the numbers, we found that the percentage of students who are fully proficient on every skill on the diagnostics is so low that requiring that level of performance to consider a student "proficient" and recommend their teacher assign the next diagnostic that almost no one would get that recommendation.  We're going to try tweaking the requirement down slightly to open this up to more students and teachers.  This threshold may be adjusted later on based on feedback and further research.
## HOW
- The service understands how to calculate which questions in an activity are linked to which skills, and whether or not the student got the answer correct or not for that question under three conditions:
- - Pre-2024 diagnostics (which have `diagnostic_question_optimal_concepts` related to them which tells us to ignore certain `concept_results` for the purposes of proficiency`
- - "Modern" diagnostics (which do not have `diagnostic_question_optimal_concepts` which makes scoring more straightforward)
- - Non-diagnostic recommended activities (which get their "Skill" values from a different table, and have a more complex scoring logic since they allow multiple attempts)
- Add a call to calculate the new data in the `on_complete` of the `BatchSaveActivitySessionConceptResultsWorker` so that it calculates for newly completed `ActivitySession`s
- For each skill, determine if every question testing that skill was answered correctly (what we consider "proficient" in the skill), then calculate how many skills are proficient, and if a high enough percentage of skills are proficient, mark the entire diagnostic as "proficient"

### Notion Card Links
https://www.notion.so/quill/Diagnostic-Insights-performance-improvements-18ad42e6f94180919741cb38b27db5d4?pvs=4

- Run selective backfill for test teachers
- Run the thor snapshot generator to confirm that the payloads change in the ways we expect
- Deploy to staging
- Complete a new diagnostic playthrough and confirm that `skill_performance` is saved to the `ActivitySession.data` column
- Log in as a test teacher
- Viewed Diagnostic Insights
- Confirmed that API payloads contained new keys for `skill_performance`
- - Manually examined specific student performance in the database, using `ConceptResult` records to manually calculate expected `skill_performance` and confirmed that they matched

PR Checklist | Your Answer
------------ | -------------


services/QuillLMS/app/models/concept_result.rb
```diff
@@ -104,6 +104,18 @@ def self.parse_extra_metadata(metadata)
       .reject { |_, v| v.nil? || (v.is_a?(Enumerable) && v.empty?) }
   end
 
+  def question_uid = question_uid_from_extra_metadata || question_uid_from_activity
+  def question_uid_from_extra_metadata = extra_metadata&.fetch('question_uid', nil)
+  def question = @question ||= Question.find_by(uid: question_uid)
+
+  # Apparently some ConceptResult records don't have question_uid in their extra_metadata, so we may need to do this complicated lookup
```
emilia-friedberg, line 111: This is true, but I don't think it should be true for any of the questions for diagnostics completed in the last year and a half (because we started tracking question_uid in [summer 2023](https://github.com/empirical-org/Empirical-Core/pull/10755)).

services/QuillLMS/client/app/bundles/Teacher/components/progress_reports/diagnostic_insights/interfaces.ts
```diff
@@ -32,8 +32,16 @@ export interface StudentLearningSequenceActivityInterface {
...
   proficient: boolean,
-  score: number,
-  timespent: number
+  score: float,
+  timespent: number,
+  skill_performance: SkillPerformance[]
+}
+
+export interface SkillPerformance {
+  skill_id: integer,
```
emilia-friedberg: NBD, but these actually should be `number`, not `integer`.

emilia-friedberg (approved):
I really like this approach. It'll be nice to have the results of this calculus actually stored in the database (and easily re-runnable if we change the scoring logic).

services/QuillLMS/app/services/diagnostic_insights/activity_session_skill_performance.rb
```diff
@@ -0,0 +1,116 @@
...
+      return unless diagnostic? || recommendation?
+
+      skill_performance
+    end
+
+    private def skill_performance
+      # NOTE: Because we've never consolidated these, "Skills" for Diagnostic activities are derived from the "SkillGroup" model,
+      # while "Skills" for recommended activities are derived from the "Recommendation" model.
+      # There is no direct database mapping linking these together, just the intentional coincidence of names
+      questions_correct
```
happythenewsad, line 23: It's safer to move the 

`return unless diagnostic? || recommendation?` guard

into this method, instead of `run`. If the implementation of `run` changes in the future, `skill_performance` could error, since `questions_correct` can return nil without the 

`return unless diagnostic? || recommendation?`  guard

happythenewsad (changes requested):
Nice work; will approve after re-sequencing a guard clause

services/QuillLMS/app/serializers/diagnostic_insights/student_learning_sequence_activity_serializer.rb
```diff
@@ -19,15 +19,19 @@ def self.recommendations(diagnostic_ids)
...
+    SKILLS_MASTERED_PROFICIENCY_THRESHOLD = 0.80
+
+    attributes :activity_id, :activity_type, :completed_at, :created_at, :activity_session_id, :proficient, :percentage, :skill_performance, :timespent
 
     def percentage = object.activity_session_unscoped&.percentage
     def timespent = object.activity_session_unscoped&.timespent
 
     def proficient
-      return nil unless pre_diagnostic? || post_diagnostic?
+      return unless pre_diagnostic? || post_diagnostic?
```
dandrabik: You might consider avoiding combining an `unless` with an `or` here. Consider making a variable `def diagnostic? = pre_diagnostic? || post_diagnostic?` so this could be `return unless diagnostic?`

services/QuillLMS/app/serializers/diagnostic_insights/student_learning_sequence_activity_serializer.rb
```diff
@@ -19,15 +19,19 @@ def self.recommendations(diagnostic_ids)
...
+      return unless skill_performance
 
-      object.activity_session_unscoped&.concept_results&.all?(&:correct)
+      skills_proficient.to_f / skills_total >= SKILLS_MASTERED_PROFICIENCY_THRESHOLD
```
dandrabik: Consider making a variable `def skills_proficiency = skills_proficient.to_f / skills_total`

services/QuillLMS/app/services/diagnostic_insights/activity_session_skill_performance.rb
```diff
@@ -0,0 +1,116 @@
...
+module DiagnosticInsights
+  class ActivitySessionSkillPerformance < ApplicationService
+    attr_reader :activity_session
+
+    def initialize(activity_session)
+      @activity_session = activity_session
+    end
+
+    def run
+      return unless concept_results.present?
```
dandrabik: consider using `return if concept_results.blank?`

services/QuillLMS/app/services/diagnostic_insights/activity_session_skill_performance.rb
```diff
@@ -0,0 +1,116 @@
...
+    private def activity = activity_session&.activity
+    private def activity_id = activity&.id
+    private def activity_session_id = activity_session&.id
+    private def classroom_unit = activity_session&.classroom_unit
+    private def concept_results = activity_session&.concept_results
+    private def diagnostic? = Activity::ALL_PRE_TEST_DIAGNOSTIC_IDS.include?(activity_id) || Activity::ALL_POST_TEST_DIAGNOSTIC_IDS.include?(activity_id)
+    private def question_uids = concept_results.map(&:question_uid)
+    private def recommendation? = classroom_unit && recommendation
+
+    private def skill_groups
```
dandrabik, line 74: I think it would be worth tailing the develop log while running this job since you're not going through active_record associations (you're doing each model query individually) to see the SQL generated. There may be a more optimized way to do this (it's hard to see just with the code).

dandrabik (approved):
Overall looks good. Some notes on potential optimizations.

happythenewsad (approved):

PR #12806, opened 2025-02-21
diagnostic insights recommendations by class

emilia-friedberg:
## WHAT
Build out the page for Diagnostic Insights: Recommendations.

## WHY
This is the second full page in this project!

## HOW
Lots of CSS and React, refactoring of the code I already had to share relevant data.

### Screenshots
Screenshot 2025-02-21 at 2 59 21 PM
Screenshot 2025-02-21 at 2 49 56 PM

### Notion Card Links
https://www.notion.so/quill/Diagnostic-Insights-Recommendations-by-Class-19fd42e6f941802586afe1bf2ba932e2?pvs=4

Tested with a bunch of different teachers and classrooms to make sure the page works in various states.

PR Checklist | Your Answer
------------ | -------------


services/QuillLMS/app/assets/stylesheets/pages/progress_reports/diagnostic_insights.scss
```diff
@@ -194,16 +194,110 @@
...
         border-radius: 8px;
         border: 1px solid $quill-grey-15;
         background: $quill-grey-1;
-        padding: 24px;
-        display: flex;
         margin-bottom: 32px;
+      }
+
+      .data-table {
+        overflow: visible;
```
happythenewsad, line 210: Is this to avoid scrollbars, when content is slightly too big for the container?

services/QuillLMS/client/app/bundles/Teacher/components/progress_reports/diagnostic_insights/history_and_recommendations.tsx
```diff
@@ -93,9 +147,33 @@ const HistoryAndRecommendations = ({ classroom, needsEmailVerification, }) => {
           />
         </div>
       </div>
-      {tabContent}
+
+      {needsEmailVerification ? (
```
happythenewsad: Suggestion: the return statement of this function is 65 lines. Abstracting 152-174 into a function would make it easier to read. 

Abstracting the header into a function would require less params though.

happythenewsad (approved):
💪

services/QuillLMS/app/assets/stylesheets/pages/progress_reports/diagnostic_insights.scss
emilia-friedberg: Yes, specifically with tooltips that we want to extend past the edge of the table (in this design, these are frequent).

services/QuillLMS/app/controllers/teachers/progress_reports/diagnostic_reports_controller.rb
```diff
@@ -188,6 +188,14 @@ def assign_post_test
     render json: {}
   end
 
+  def assign_diagnostics
+    params[:diagnostics].each do |diagnostic|
+      Units::AssignmentHelpers.assign_unit_to_one_class(params[:classroom_id], diagnostic[:unit_template_id], diagnostic[:student_ids], false)
+    end
+
+    render json: {}
+  end
```
anathomical, line 197: I'm somehow surprised that we didn't already have an endpoint for this, but it totally makes sense to put it in this controller.

services/QuillLMS/client/app/bundles/Teacher/components/progress_reports/diagnostic_insights/interfaces.ts
```diff
@@ -1,18 +1,20 @@
+import { STARTER, INTERMEDIATE, ADVANCED, } from './shared'
```
anathomical, line 1: Love moving these to constants.  I wish we had a clean way to share constants from front-end to back-end so we could guarantee syncing them across both sides of the code, but that's well beyond the scope of this PR.

services/QuillLMS/client/app/bundles/Teacher/components/progress_reports/diagnostic_insights/interfaces.ts
```diff
@@ -28,16 +30,16 @@ export interface StudentLearningSequenceActivityInterface {
...
 export interface DiagnosticsAssignedInterface {
-  starter?: StudentDiagnosticAssignmentInterface[],
-  intermediate?: StudentDiagnosticAssignmentInterface[],
-  advanced?: StudentDiagnosticAssignmentInterface[],
+  [STARTER]?: StudentDiagnosticAssignmentInterface[],
+  [INTERMEDIATE]?: StudentDiagnosticAssignmentInterface[],
+  [ADVANCED]?: StudentDiagnosticAssignmentInterface[],
   ell_starter?: StudentDiagnosticAssignmentInterface[],
   ell_intermediate?: StudentDiagnosticAssignmentInterface[],
   ell_advanced?: StudentDiagnosticAssignmentInterface[]
```
anathomical: Actually, now that I look at the code, I don't think that these three values are ever returned by the API.  When `ELL` is checked, we still use `starter`, `intermediate`, and `advanced` in the API response.  This is on me for including these values in the interfaces file.

services/QuillLMS/client/app/bundles/Teacher/components/progress_reports/diagnostic_insights/__tests__/history.test.tsx
```diff
@@ -8,148 +8,105 @@ import History, {
...
       </MemoryRouter>
     );
 
     await waitFor(() => {
-      expect(screen.getByText('Student')).toBeInTheDocument();
-      expect(screen.getByText('Starter Diagnostic')).toBeInTheDocument();
-      expect(screen.getByText('Intermediate Diagnostic')).toBeInTheDocument();
-      expect(screen.getByText('Advanced Diagnostic')).toBeInTheDocument();
+      expect(screen.getByText('Alice Johnson')).toBeInTheDocument();
+      expect(screen.getByText('Bob Smith')).toBeInTheDocument();
```
anathomical: This is stylistic, so take or leave it, but I kind of like dynamically pulling this data out of the mock payload.  I think this would work:
```javascript
mockStudentHistories.map((history) => history.user_name).each((user_name) => {
  expect(screen.getByText(user_name).toBeInTheDocument();
})
```

services/QuillLMS/client/app/bundles/Teacher/components/progress_reports/diagnostic_insights/__tests__/history.test.tsx
```diff
@@ -162,19 +119,16 @@ describe('History Component', () => {
...
-
     const { asFragment } = render(
       <MemoryRouter>
-        <History classroom={mockClassroom} timeframe="2024-Q1" useELLDiagnostics={false} />
+        <History studentHistories={mockStudentHistories} />
       </MemoryRouter>
     );
 
-    await waitFor(() => expect(screen.getByText(mockStudentHistories[0].user_name)).toBeInTheDocument());
+    await waitFor(() => expect(screen.getByText('Alice Johnson')).toBeInTheDocument());
```
anathomical: Does this duplicate the testing done in the `it('renders student history table when studentHistories is provided')` test?  If it does, I don't think that's necessarily bad, I'm just making sure I understand whether this is the same setup as that `it` block or not.

eadams17 (approved):
LGTM!

services/QuillLMS/client/app/bundles/Teacher/components/progress_reports/diagnostic_insights/__tests__/recommendations.test.tsx
```diff
@@ -0,0 +1,145 @@
...
+    expect(checkboxes[0]).toBeChecked();
+
+    await userEvent.click(checkboxes[0]);
+    expect(checkboxes[0]).not.toBeChecked();
+
+    await userEvent.click(checkboxes[0]);
+    expect(checkboxes[0]).toBeChecked();
+  });
+
+  it('assigns the correct students', async () => {
```
anathomical, line 120: I really dig that this is a test that we can write so cleanly.

services/QuillLMS/client/app/bundles/Teacher/components/progress_reports/diagnostic_insights/__tests__/shared.test.tsx
```diff
@@ -0,0 +1,90 @@
...
+});
+
+describe('generateRecommendations', () => {
+  const classroomId = 1;
+
+  it('recommends Starter for students with no proficiency', () => {
+    const studentHistories = [createStudentHistory(1, false, false, false)];
+    const recommendations = generateRecommendations(false, studentHistories, classroomId);
+
+    expect(recommendations[STARTER].student_ids).toContain(1);
```
anathomical: It feels a little awkward to be hard-coding this integer ID  in both `createStudentHistory` call and the `toContain` call, but I'm not sure it's worth making this verbose by splitting it out into a new `const` for every one of these combinations, so I don't know that I actually think we need to change anything here.  Unless it makes sense to use the same ID for every case and just define the `const` for it next to `classroomId`?  Do what you think makes sense here, even if it's just leaving things alone.

services/QuillLMS/client/app/bundles/Teacher/components/progress_reports/diagnostic_insights/history_and_recommendations.tsx
```diff
@@ -1,63 +1,140 @@
...
 
-  function handleTabChange(e) {
-    setActiveTab(e.currentTarget.value)
-  }
+  React.useEffect(() => {
+    if (!classroom) { return }
+
+    setStudentHistories(null)
+
+    getStudentHistories()
```
anathomical: The first thing we do during `getStudentHistories` is call `setStudentHistories(null)`, so I'm not sure we also need to set it in this effect.

services/QuillLMS/client/app/bundles/Teacher/components/progress_reports/diagnostic_insights/recommendations.tsx
```diff
@@ -0,0 +1,228 @@
...
+    noTooltip: true
+  },
+]
+
+interface StudentDiagnosticCellProps {
+  studentHistory: StudentHistoryInterface;
+  diagnosticLevel: typeof STARTER | typeof INTERMEDIATE | typeof ADVANCED;
+  recommendations: LeveledAssignmentsInterface;
+  selectedAssignments: LeveledAssignmentsInterface;
+  setSelectedAssignments: React.Dispatch<React.SetStateAction<LeveledAssignmentsInterface>>;
```
anathomical, line 54: I don't know if I've noticed us enforcing `setState` input interfaces before, but I love it.

services/QuillLMS/client/app/bundles/Teacher/components/progress_reports/diagnostic_insights/shared.tsx
```diff
@@ -0,0 +1,93 @@
...
+
+import { EXTRA_SMALL_ICON_BASE_SRC, } from '../../../../Shared/index'
+import {
+  STARTER_DIAGNOSTIC_UNIT_TEMPLATE_ID,
+  INTERMEDIATE_DIAGNOSTIC_UNIT_TEMPLATE_ID,
+  ADVANCED_DIAGNOSTIC_UNIT_TEMPLATE_ID,
+  ELL_STARTER_DIAGNOSTIC_UNIT_TEMPLATE_ID,
+  ELL_INTERMEDIATE_DIAGNOSTIC_UNIT_TEMPLATE_ID,
+  ELL_ADVANCED_DIAGNOSTIC_UNIT_TEMPLATE_ID,
+} from '../../assignment_flow/assignmentFlowConstants'
```
anathomical, line 13: Another case where I wish we had a clean way to pass these constant values from the back-end, especially if we end up in a world where we are updating Diagnostics every Summer it'd be nice not to have to update these constants, too.  But, again, well beyond the scope of this PR.

services/QuillLMS/client/app/bundles/Teacher/components/progress_reports/diagnostic_insights/shared.tsx
```diff
@@ -0,0 +1,93 @@
...
+    [ADVANCED]: ADVANCED_DIAGNOSTIC_UNIT_TEMPLATE_ID
+  }
+}
+
+export function proficientDiagnostic(studentLearningSequences: StudentLearningSequenceInterface[]) {
+  return studentLearningSequences?.find(studentLearningSequence => {
+    const { student_learning_sequence_activities, } = studentLearningSequence
+    const { pre_diagnostic, post_diagnostic, } = student_learning_sequence_activities
+
+    return pre_diagnostic?.find(preDiagnosticActivity => preDiagnosticActivity.proficient) || post_diagnostic?.find(postDiagnosticActivity => postDiagnosticActivity.proficient)
```
anathomical, line 49: I was surprised this worked.  I didn't realize that `.find()` returns `undefined` if no matches are found instead of `[]` (which is what happens in Ruby).  No need to change anything, just flagging that today I learned...

services/QuillLMS/client/app/bundles/Teacher/components/progress_reports/diagnostic_insights/shared.tsx
```diff
@@ -0,0 +1,93 @@
...
+    }
+
+    if (!proficientDiagnostic(intermediate)) {
+      studentIdsForIntermediate.push(user_id)
+      return
+    }
+
+    if (!proficientDiagnostic(advanced)) {
+      studentIdsForAdvanced.push(user_id)
+    }
```
anathomical, line 76: I like the pattern you've used here with early returns.  It makes the logic very clean.

services/QuillLMS/spec/controllers/teachers/progress_reports/diagnostic_reports_controller_spec.rb
```diff
@@ -470,4 +470,49 @@
...
+      expect(response).to be_successful
+
+      unit1 = Unit.find_by(user_id: teacher.id, unit_template_id: activity_pack1.id)
+      classroom_unit1 = ClassroomUnit.find_by(unit_id: unit1.id, classroom_id: classroom.id)
+      unit2 = Unit.find_by(user_id: teacher.id, unit_template_id: activity_pack2.id)
+      classroom_unit2 = ClassroomUnit.find_by(unit_id: unit2.id, classroom_id: classroom.id)
+
+      expect(unit1).to be
+      expect(classroom_unit1).to be
+      expect(classroom_unit1.assigned_student_ids).to eq([student1.id])
```
anathomical: It feels like it's worth using `contain_exactly` here instead of `eq` to match the pattern we use below.

services/QuillLMS/spec/controllers/teachers/progress_reports/diagnostic_reports_controller_spec.rb
```diff
@@ -470,4 +470,49 @@
...
+      expect(unit2).to be
+      expect(classroom_unit2).to be
+      expect(classroom_unit2.assigned_student_ids).to eq([student1.id])
```
anathomical: Again, I like using `contains_exactly` here.

services/QuillLMS/spec/controllers/teachers/progress_reports/diagnostic_reports_controller_spec.rb
```diff
@@ -470,4 +470,49 @@
...
+
+    it 'creates a new pack and assigns it if there is no existing pack' do
+      expect do
+        get :assign_diagnostics, params: { classroom_id: classroom.id, diagnostics: [{ unit_template_id: activity_pack1.id, student_ids: [student1.id] }, { unit_template_id: activity_pack2.id, student_ids: [student1.id] }] }
+      end.to change(ClassroomUnit, :count).by(2)
+
+      expect(response).to be_successful
+
+      unit1 = Unit.find_by(user_id: teacher.id, unit_template_id: activity_pack1.id)
+      classroom_unit1 = ClassroomUnit.find_by(unit_id: unit1.id, classroom_id: classroom.id)
```
anathomical: Take or leave this, as it's just stylistic, but Rails does this magic thing where you can use an object or its ids in these kind of methods.  So `find_by(classroom_id: classroom.id)` works exactly the same as `find_by(classroom: classroom)`.  This means that because `classroom` is defined, you can use implicit values to simplify this call slightly: `find_by(classroom:, unit: unit1)` which is slightly easier to read.

anathomical (approved):
I've flagged some take-or-leave style items, but basically I really like how this is structured and appreciate the thorough test coverage.

services/QuillLMS/client/app/bundles/Teacher/components/progress_reports/diagnostic_insights/__tests__/history.test.tsx
emilia-friedberg: It does, you're right. It's just to confirm that the data is loaded before taking a snapshot, but I'll rewrite to make it clearer.

services/QuillLMS/client/app/bundles/Teacher/components/progress_reports/diagnostic_insights/__tests__/shared.test.tsx
emilia-friedberg: Good suggestion; I've refactored accordingly.

PR #12808, opened 2025-02-24
BugFix for AdminUsageSnapshotEmailWorker reports without school_ids

dandrabik:
## WHAT
Check for a failing condition early in `AdminUsageSnapshotEmailWorker` instead of failing and re-running. Also, limit retries to prevent over billing.
## WHY
If the query in has no `school_id`s, it will fail for being an invalid SQL.Trying to get a clearer picture when BigQuery Fallback queries happen, but there are a lot of errors in our logs from failing jobs when there are no `school_ids`.
## HOW
Early Return.
### Screenshots
[Sentry Error](https://quillorg-5s.sentry.io/issues/6315555767/events/0dac1ef94bbb4dce9c5638483299d05b/?environment=production&project=11238&query=is%3Aunresolved%20issue.priority%3A%5Bhigh%2C%20medium%5D&referrer=previous-event&statsPeriod=24h&stream_index=4) Example

Screenshot 2025-02-24 at 11 24 30 AM

The additional logic is very small, so I did limited testing. I would appreciate a second set of eyes to determine the product logic is valid.

PR Checklist | Your Answer
------------ | -------------


anathomical (approved):

happythenewsad (approved):
Consider adding an inline comment, explaining that PDFs generated from queries with zero schools will likely not be missed by the recipient admin. A year from now, someone seeing this early return could be surprised at its existence.

dandrabik:
> Consider adding an inline comment, explaining that PDFs generated from queries with zero schools will likely not be missed by the recipient admin. A year from now, someone seeing this early return could be surprised at its existence.

Good idea. I've added a comment about a potential feature. Note this code doesn't actually change the current behavior (those emails aren't currently sent because the SQL fails with syntax error), this PR just makes it less resource intensive (hitting BigQuery) and noisy (raising a syntax error) in our system.

PR #12809, opened 2025-02-24
diagnostic-insights-show-data-with-no-assignments

anathomical:
## WHAT
Filter student_learning_sequences in Rails instead of SQL
## WHY
The combination of JOINs for both non-ELL and ELL diagnostics in the SQL combined with filtering via WHERE for only non-ELL OR ELL diagnostics meant that if there were some non-ELL assignments, but no ELL diagnostics, specific User records were being removed from the results for ELL assignments.  This could be solved in SQL by using an additional ON condition in the LEFT JOIN in SQL, but Rails does not provide a clean way of applying such a condition, so it's easiest to move this logic to Rails.
## HOW
- Remove a `.where` clause from the SQL that fetches data
- Add an additional Ruby `.select` directive to the code that determines which StudentLearningSequence records to include in the returned payload

- Deploy to staging
- Using an account provided by Emilia, test on Staging to ensure that payloads are coming back for ELL in cases where non-ELL data exists instead of being buried.

PR Checklist | Your Answer
------------ | -------------


emilia-friedberg (approved):
Thank you for figuring this out!

PR #12811, opened 2025-02-24
Run some cron jobs only in Production

dandrabik:
## WHAT
We have many internal `staging` environments, but not all of our third-party integrations have a proper staging, e.g. BigQuery. So staging cron jobs essentially hit production in those instances. Add some logic to `Cron.rb` to have some tasks ONLY run on production and not on staging (or local).
## WHY
This is causing extra billing in BigQuery and potentially could potentially cause data issues in other third-parties like Vitally and Segment.
## HOW
Find jobs in `Cron.rb` that hit third-parties (or that generally shouldn't be run on staging), add `if production?` check.

Refactor tests to accommodate
### Screenshots
Screenshot 2025-02-24 at 2 57 04 PM

I ran the code locally, and seems to work properly. It's slightly hard to test this type of code, but I will check the jobs that ran tomorrow.

PR Checklist | Your Answer
------------ | -------------


services/QuillLMS/app/models/cron.rb
```diff
@@ -17,23 +17,27 @@ def self.interval_1_hour
...
+  def self.interval_1_day
+    run_nondaily_tasks
 
     # pass yesterday's date for stats email queries and labels
     date = Time.current.getlocal('-05:00').yesterday
 
     # internal analytics and security
-    DailyStatsEmailJob.perform_async(date)
-    QuillStaffAccountsChangedWorker.perform_async
+    DailyStatsEmailJob.perform_async(date) if production?
```
dandrabik, line 39: I'm definitely open to which jobs it's better to run on both Prod and Staging. I mainly targetted third-party calls, but I also turned off these internal ones since the staging ones aren't useful. Open to feedback on that.

anathomical (approved):
I think the way you've structured the tests makes a lot of sense.  I also think that your instinct about which jobs to disable in staging is pretty solid: anything that hits external APIs is a good candidate except for potential special circumstances.  We might also disable most things that send emails, since I don't think there's high value in that from staging.  So this potentially adds the following candidates for production-only:
```ruby
TeacherNotifications::EnqueueUsersForRollupEmailWorker # there are hourly, daily, and weekly instances of this worker call (each taking a different arg), it processes a bunch of teacher configurations based on production data but blackholes the resultant emails, can be tested by manually triggering jobs when desired
RenewExpiringRecurringSubscriptionsWorker # while we do have a staging instance of stripe, I'm not sure that there's value in creating invoices and fake charges in it
AlertSoonToExpireSubscriptionsWorker # most of these will be users with data drawn from prod, again with their emails being blackholed
PopulateAnnualVitallyWorker # you could add the same guard clause here, but the worker in question already has code in it to bail early if the environment isn't production-ish (it actually checks for a SYNC_TO_VITALLY ENV variable which is not set in non-prod environments)
```

dandrabik:
> I think the way you've structured the tests makes a lot of sense. I also think that your instinct about which jobs to disable in staging is pretty solid: anything that hits external APIs is a good candidate except for potential special circumstances. We might also disable most things that send emails, since I don't think there's high value in that from staging. So this potentially adds the following candidates for production-only:
> 
> ```ruby
> TeacherNotifications::EnqueueUsersForRollupEmailWorker # there are hourly, daily, and weekly instances of this worker call (each taking a different arg), it processes a bunch of teacher configurations based on production data but blackholes the resultant emails, can be tested by manually triggering jobs when desired
> RenewExpiringRecurringSubscriptionsWorker # while we do have a staging instance of stripe, I'm not sure that there's value in creating invoices and fake charges in it
> AlertSoonToExpireSubscriptionsWorker # most of these will be users with data drawn from prod, again with their emails being blackholed
> PopulateAnnualVitallyWorker # you could add the same guard clause here, but the worker in question already has code in it to bail early if the environment isn't production-ish (it actually checks for a SYNC_TO_VITALLY ENV variable which is not set in non-prod environments)
> ```

These are good ones to look at. I've added the `PopulateAnnualVitallyWorker` to the prod-only list. I don't have a good feel if we should get rid of the Stripe one and I'm on the fence a bit about the emails ones, so I'm going to deploy as is (since the BigQuery ones affect billing) and put a note to discuss the others at the next engineering meeting, e.g. do we get value by having some jobs run on staging. Thanks for the review!

PR #12818, opened 2025-03-04
standardize-last-name-sorting

anathomical:
## WHAT
Update `User.last_name` logic to match contemporary rules we use for identifying "last name" for users: the "last name" is everything after the "first name".  This bundles all variants of "middle names" into "last name", which is a known issue, but we'd rather be consistent about it.
## WHY
The immediate trigger for this code change is a report that students are being sorted "by last name" in different places on two specific pages.  This is caused by inconsistent implementation of "last name" across our code.

`last_name` should be everything after `first_name` instead of just the string after the final space in the name.  This ensures that there are no parts of `name` that aren't included in one of `first_name` or `last_name`.  It also ensures that Ruby-side sorting (i.e. `students.sort_by(&:last_name)`) produces the same results as our contemporary SQL based sorting (i.e. `order(Arel.sql("SPLIT_PART(TRIM(\"#{table_name}\".name), ' ', 2)"))`).  Also, the old way of doing things was having the side effect of deleting parts of names when `first_name=` was called sometimes even though the specs weren't written in a way that caught that edge case.
## HOW
- Modify the `split` call used in `User.last_name` to only split names into two parts so that there aren't middle parts that are excluded from `first_name` and `last_name`
- Update specs to ensure that we test for cases where `user.first_name=` would remove name parts in the old code so that edge case is covered moving forward

### Notion Card Links
https://www.notion.so/quill/Class-roster-in-the-Manage-Classes-tab-is-in-a-different-order-than-the-Activity-Summary-1abd42e6f94180598568c3385f1ce182

- Create a classroom for student sorting
- Add two students: "Alice Beaty" and "Alice Ann Carlisle".
- Ensure that the sorting order places "Alice Ann Carlisle" above "Alice Beaty" in both the Classroom Management page and in the Activity Summary page.

PR Checklist | Your Answer
------------ | -------------


anathomical:
After a discussion with Support, I'm electing to close this PR rather than try to figure out all the edge cases.  I've added discussion of this issue to the next engineering meeting.

PR #12821, opened 2025-03-05
update how we handle confirmation for erb forms

emilia-friedberg:
## WHAT
Install `turbo` package and use that for erb files that need a window.confirm message.

## WHY
I noticed while testing something unrelated that the "Clear User Data" button on the user editor no longer had a pop-up confirmation message before submitting the action (guess how I found that out). This turns out to be because Rails 7 no longer has this functionality built-in, and we now [need to use `Turbo`](https://guides.rubyonrails.org/working_with_javascript_in_rails.html#confirmations) to handle it. Since we definitely don't want staff members inadvertently clearing user data, I decided it was worth fixing immediately.

## HOW
I installed Turbo and imported it in `applications.js`, our only entryfile that ERB pages use (with one exception, see below). Then I updated all of the places we were using `confirm:` to use `turbo_confirm` instead and verified that they work as expected.

Exception: hiding guidelines within Brendan's ERB version of the trial runner tool still uses the old syntax. However, there is no Javascript entrypoint to this part of the app, and since the plan is to deprecate it anyway, I didn't think it was worth setting up a Javascript import just to fix this.

### Screenshots

### Notion Card Links

Tested all the files I touched to confirm that the confirmation (haha) is working as expected.

PR Checklist | Your Answer
------------ | -------------
Have you added and/or updated tests? | N/A - no tests for eb
Have you deployed to Staging? |  NO - tiny change, tested locally


brendanshean (approved):
Thank you for taking the initiative to fix this.  It's definitely surprising behavior that could prove costly.

eadams17 (approved):
Great catch on this! Definitely would have caused a lot of headaches-- thank you 🙏

happythenewsad (changes requested):
I think this merits further discussion. Turbo adds 16k LoC to our packages, and 1.1MB. 

We only need it in a few places, so I advocate for a packageless alternative:

```
<script>
  document.getElementById('confirmButton').addEventListener('click', function() {
    const result = confirm('Are you sure you want to proceed?');
  });
</script>
```

Update: given the time-sensitivity of the issue, consider simply removing the CMS buttons temporarily, before adding the terminal fix. Thanks for addressing the reversion caused by the Rails 7 upgrade

emilia-friedberg:
> I think this merits further discussion. Turbo adds 16k LoC to our packages, and 1.1MB.
> 
> We only need it in a few places, so I advocate for a packageless alternative:
> 
> ```
> <script>
>   document.getElementById('confirmButton').addEventListener('click', function() {
>     const result = confirm('Are you sure you want to proceed?');
>   });
> </script>
> ```
> 
> Update: given the time-sensitivity of the issue, consider simply removing the CMS buttons temporarily, before adding the terminal fix. Thanks for addressing the reversion caused by the Rails 7 upgrade

@happythenewsad I've updated these files in accordance with your suggestion.

happythenewsad (approved):
Thanks for keeping our app lean!

anathomical (approved):
Nice job figuring this out.

PR #12822, opened 2025-03-06
use teacher verification properties in internal tools

emilia-friedberg:
## WHAT
Allow staff members to edit teacher verification from the User Manager and put the basic logic structure in place in the models to set `verifier_id` correctly when the `verification_status` changes.

## WHY
This allows staff members to edit teacher verification and puts the framework in place that we'll need to support this flow through teacher and admin accounts.

## HOW
- add to the User Manager forms and the controller actions to handle `teacher_verification_status`
- rename "Admin approval status" to "Admin verification status" on the frontend
- add a new `Current` model to make what is effectively `current_user` accessible in the model file so we can set `verifier` correctly
- add a bunch of logic in teacher info to make sure we are always setting the verifier_id when verification_status changes, unless it is changed to Pending or nil, in which case `verifier_id` should be nil
- reset teacher verification status and verifier id when the `schools_users` record is updated or destroyed.

### Screenshots
Screenshot 2025-03-06 at 11 03 27 AM

### Notion Card Links
https://www.notion.so/quill/Diagnostic-Insights-Handling-Verification-Attributes-Internally-1aed42e6f94180d1b085f2081087f9fa?pvs=4

Tested locally.

PR Checklist | Your Answer
------------ | -------------


services/QuillLMS/spec/models/current_spec.rb
```diff
@@ -0,0 +1,23 @@
+# frozen_string_literal: true
+
+require 'rails_helper'
+
+RSpec.describe Current, type: :model do
+  after { Current.reset }
```
brendanshean: Why is the `Current.reset` needed here?

services/QuillLMS/spec/models/teacher_info_spec.rb
```diff
@@ -143,4 +144,112 @@
       end
     end
   end
+
+  describe 'verification_status, verifier_id, and assign_verifier callback' do
+    context 'when verification_status changes and Current.user is present' do
+      let(:verifier) { create(:user) }
+
+      before do
```
brendanshean: Purely aesthetic, but to keep consistent with one liner `after { Current.reset }`

`before { Current.user = verifier }`

services/QuillLMS/spec/models/current_spec.rb
emilia-friedberg: Good question! I don't think it's necessary so I removed it.

services/QuillLMS/spec/models/teacher_info_spec.rb
```diff
@@ -143,4 +144,112 @@
...
+        Current.user = verifier
+      end
+
+      after { Current.reset }
+
+      it 'leaves verifier_id unchanged if verification_status remains the same' do
+        original_verifier_id = 99
+        teacher_info = create(:teacher_info, verification_status: TeacherInfo::APPROVED, verifier_id: original_verifier_id)
```
brendanshean: There's some boilerplate with teacher_info that can be cleaned up:
Add a trait to the factory:
`trait(:verification_completed) { verification_status: TeacherInfo::APPROVED }`

then this:
`teacher_info = create(:teacher_info, verification_status: TeacherInfo::APPROVED, verifier_id: original_verifier_id)`

becomes
`teacher_info = create(:teacher_info, :verification_completed, verifier_id: original_verifier_id)`

services/QuillLMS/spec/models/teacher_info_spec.rb
```diff
@@ -143,4 +144,112 @@
...
+        teacher_info.update!(notification_email_frequency: TeacherInfo::NOTIFICATION_EMAIL_FREQUENCIES.first)
+        expect(teacher_info.verifier_id).to eq(original_verifier_id)
+      end
+    end
+
+    context 'when verification_status changes to pending' do
+      it 'sets the verifier_id to nil' do
+        original_verifier_id = 99
+        teacher_info = create(:teacher_info, verification_status: TeacherInfo::APPROVED, verifier_id: original_verifier_id)
```
brendanshean: `teacher_info = create(:teacher_info, :verification_completed, verifier_id: original_verifier_id)`

services/QuillLMS/spec/models/teacher_info_spec.rb
```diff
@@ -143,4 +144,112 @@
...
+        teacher_info.update!(verification_status: TeacherInfo::PENDING)
+        expect(teacher_info.verifier_id).to be_nil
+      end
+    end
+
+    context 'when verification_status changes to nil' do
+      it 'sets the verifier_id to nil' do
+        original_verifier_id = 99
+        teacher_info = create(:teacher_info, verification_status: TeacherInfo::APPROVED, verifier_id: original_verifier_id)
```
brendanshean: `teacher_info = create(:teacher_info, :verification_completed, verifier_id: original_verifier_id)`

services/QuillLMS/spec/models/teacher_info_spec.rb
```diff
@@ -143,4 +144,112 @@
...
+        expect(teacher_info.verifier_id).to be_nil
+      end
+    end
+
+    context 'when verification_status changes and a verifier_id is explicitly provided' do
+      let(:provided_verifier) { create(:user) }
+      let(:new_verifier) { create(:user) }
+      let(:current_user) { create(:user) }
+
+      before do
```
brendanshean: Purely aesthetic, but to keep consistent with one liner `after { Current.reset }`

`before { Current.user = verifier }`

services/QuillLMS/spec/models/teacher_info_spec.rb
```diff
@@ -143,4 +144,112 @@
...
+      let(:verifier) { create(:user) }
+
+      before do
+        Current.user = verifier
+      end
+
+      after { Current.reset }
+
+      it 'assigns verifier_id from Current.user when no verifier_id is explicitly provided' do
+        teacher_info = create(:teacher_info)
```
brendanshean: I would extract the `teacher_info = create(:teacher_info)` into a `let(:teacher_info) { create(:teacher_info) }`

Think of your tests like 
1. setup
2. run method[s]
3. check expectation

Refactoring common setup outside of the test block makes it quicker to identify what changes across different test.

services/QuillLMS/spec/models/teacher_info_spec.rb
```diff
@@ -143,4 +144,112 @@
...
+        teacher_info.update!(verification_status: TeacherInfo::APPROVED)
+        expect(teacher_info.verifier_id).to eq(verifier.id)
+      end
+
+      it 'does not assign a verifier_id if the verification_status is PENDING' do
+        teacher_info = create(:teacher_info)
```
brendanshean: see previous comment on line 159 about refactoring this line out.

services/QuillLMS/spec/models/teacher_info_spec.rb
```diff
@@ -143,4 +144,112 @@
...
+      let(:current_user) { create(:user) }
+
+      before do
+        Current.user = current_user
+      end
+
+      after { Current.reset }
+
+      it 'uses the explicitly provided verifier_id when it is not the same as the previous value' do
+        teacher_info = create(:teacher_info, verification_status: TeacherInfo::APPROVED, verifier_id: provided_verifier.id)
```
brendanshean: see previous comment on line 159 about refactoring this line out.

services/QuillLMS/spec/models/teacher_info_spec.rb
```diff
@@ -143,4 +144,112 @@
...
+        teacher_info.update!(verification_status: TeacherInfo::DENIED, verifier_id: new_verifier.id)
+        expect(teacher_info.verifier_id).to eq(new_verifier.id)
+      end
+
+      it 'uses the explicitly provided verifier_id when it is the same as the previous value' do
+        teacher_info = create(:teacher_info, verification_status: TeacherInfo::APPROVED, verifier_id: provided_verifier.id)
```
brendanshean: see previous comment on line 159 about refactoring this line out.

services/QuillLMS/spec/models/teacher_info_spec.rb
```diff
@@ -143,4 +144,112 @@
...
+          teacher_info.update!(verification_status: nil)
+        }.not_to raise_error
+        expect(teacher_info.reload.verification_status).to be_nil
+      end
+    end
+
+    context 'when verification_status does not change' do
+      let(:verifier) { create(:user) }
+
+      before do
```
brendanshean: Purely aesthetic, but to keep consistent with one liner `after { Current.reset }`

`before { Current.user = verifier }`

services/QuillLMS/spec/models/user_spec.rb
brendanshean, line 1: I feel like the `it` block descriptions in lines 1856, 1862, 1874, 1879 are kind of redundant and don't add much value.  I suggest rewriting them to one liner without loss of clarity.

For example:

```
      it 'returns the verifier_id from the teacher info' do
        expect(user.teacher_verifier_id).to eq(teacher_info.verifier_id)
      end
```
     
` it { expect(user.teacher_verifier_id).to eq(teacher_info.verifier_id) } `

services/QuillLMS/spec/models/teacher_info_spec.rb
```diff
@@ -143,4 +144,112 @@
...
+      it 'uses the explicitly provided verifier_id when it is the same as the previous value' do
+        teacher_info = create(:teacher_info, verification_status: TeacherInfo::APPROVED, verifier_id: provided_verifier.id)
+        teacher_info.update!(verification_status: TeacherInfo::DENIED, verifier_id: provided_verifier.id)
+        expect(teacher_info.verifier_id).to eq(provided_verifier.id)
+      end
+    end
+
+    context 'when verification_status changes and neither an explicit verifier_id nor Current.user is provided' do
+      it 'does not save the update and adds an error on verifier_id' do
+        teacher_info = create(:teacher_info)
```
brendanshean: see previous comment on line 159 about refactoring this line out.

services/QuillLMS/spec/models/teacher_info_spec.rb
```diff
@@ -143,4 +144,112 @@
...
+        expect {
+          teacher_info.update!(verification_status: TeacherInfo::DENIED)
+        }.to raise_error(ActiveRecord::RecordInvalid, /Verifier must be provided when verification_status changes/)
+      end
+
+      it 'saves the update and does not add an error when the verification status is pending' do
+        teacher_info = create(:teacher_info)
```
brendanshean: see previous comment on line 159 about refactoring this line out.

services/QuillLMS/spec/models/teacher_info_spec.rb
```diff
@@ -143,4 +144,112 @@
...
+        expect {
+          teacher_info.update!(verification_status: TeacherInfo::PENDING)
+        }.not_to raise_error
+        expect(teacher_info.reload.verification_status).to eq(TeacherInfo::PENDING)
+      end
+
+      it 'saves the update and does not add an error when the verification status is nil' do
+        teacher_info = create(:teacher_info)
```
brendanshean: see previous comment on line 159 about refactoring this line out.

brendanshean:
I'm adding a note to discuss the `Current` pattern in engineering meeting.

brendanshean (commented):
I've left some minor refactoring comments in the specs.

services/QuillLMS/app/views/cms/users/_form.html.erb
```diff
@@ -39,14 +39,19 @@
         <%= f.select :admin_sub_role, [nil].concat(AdminInfo::SUB_ROLES) %>
       </div>
       <div class='cms-form-row'>
-        <%= f.label :admin_approval_status %>
+        <%= f.label 'Admin verification status' %>
```
anathomical: Any reason for the change from symbol to string here?  I doubt it matters, but I was curious.

services/QuillLMS/app/views/cms/users/_form.html.erb
```diff
@@ -39,14 +39,19 @@
...
         <%= f.text_field :admin_approval_status, :readonly => true %>
       </div>
+    <% elsif @user.teacher? %>
       <div class='cms-form-row'>
-        <%= f.label :email_verification_status %>
-        <%= f.select :email_verification_status, [nil, UserEmailVerification::PENDING, UserEmailVerification::VERIFIED] %>
+        <%= f.label :teacher_verification_status %>
+        <%= f.select :teacher_verification_status, [nil].concat(TeacherInfo::VERIFICATION_STATUSES) %>
```
anathomical: It doesn't matter for this case, but I've had unexpected behavior with `concat`.  Specifically, `concat` modifies the array that it's called on.  In this case, that's a newly instantiated array `[nil]` every time, but in the past when I've done things like `a.concat(b)`, where `a` is kept in memory, subsequent calls re-appended `b` to the newly-combined value.  Eventually you end up with something that looks like `a + b + b + b + b + b + ...`

All of this is a long-winded way of making a take-or-leave comment that I've generally come to prefer `a + b` for array combinations, or in your case here: `[nil] + TeacherInfo::VERIFICATION_STATUSES`.

services/QuillLMS/spec/models/current_spec.rb
```diff
@@ -0,0 +1,23 @@
+# frozen_string_literal: true
```
anathomical: I don't think we need to remove this file, but as far as I can tell what we're really testing here is stuff that should be covered in `ActiveSupport::CurrentAttributes`.  None of the code in your model definition has any implications for this spec (like `reset` isn't even in your code).  So, in theory, this behavior should be tested by the underlying parent class.

That said, this spec file serves as a useful bit of documentation.  For instance, I didn't know that `reset` was a call that could be made, so seeing it here was helpful.

services/QuillLMS/spec/models/schools_users_spec.rb
```diff
@@ -33,6 +35,37 @@
...
+        schools_user.destroy
+        teacher_info.reload
+        expect(teacher_info.verification_status).to be_nil
+        expect(teacher_info.verifier_id).to be_nil
+      end
+    end
+
+    context 'when user does not have teacher_info' do
+      let(:user_without_teacher_info) { create(:user) }
+      let(:schools_user) { create(:schools_users, user: user_without_teacher_info, school:) }
```
anathomical: Minor potential efficiency: calling `create(:schools_users)` with no additional arguments will automatically generate a `create(:teacher)` and a `create(:school)` which should get you a user without teacher_info.  So I think you can simplify your `let` calls to just `let(:schools_user) { create(:schools_users) }`

services/QuillLMS/spec/models/teacher_info_spec.rb
```diff
@@ -143,4 +144,112 @@
...
+      before do
+        Current.user = verifier
+      end
+
+      after { Current.reset }
+
+      it 'assigns verifier_id from Current.user when no verifier_id is explicitly provided' do
+        teacher_info = create(:teacher_info)
+        teacher_info.update!(verification_status: TeacherInfo::APPROVED)
+        expect(teacher_info.verifier_id).to eq(verifier.id)
```
anathomical: This is pretty stylistic, but I've come to appreciate minimizing the contents of an `it` block by using `subject` to define the trigger code.  Definitely a take-or-leave comment, but I'd probably write this as (building on @brendanshean's comment):
```ruby
subject { teacher_info.update!(verification_status: updated_verification_status) }

let(:teacher_info) { create(:teacher_info) }
let(:updated_verification_status) { TeacherInfo::APPROVED }

it { expect { subject }.to change { teacher_info.reload.verifier_id }.from(nil).to(verifier.id) }
```

services/QuillLMS/spec/models/teacher_info_spec.rb
```diff
@@ -143,4 +144,112 @@
...
+      end
+
+      it 'does not assign a verifier_id if the verification_status is PENDING' do
+        teacher_info = create(:teacher_info)
+        teacher_info.update!(verification_status: TeacherInfo::PENDING)
+        expect(teacher_info.verifier_id).to be_nil
```
anathomical: Building on my previous comment, this could be:
```ruby
context 'setting verification_status to PENDING' do
  let(:updated_verification_status) { TeacherInfo::PENDING }

  it do
    subject
    expect(teacher_info.verifier_id).to be_nil
  end
end
```

services/QuillLMS/spec/models/teacher_info_spec.rb
```diff
@@ -143,4 +144,112 @@
...
+      before do
+        Current.user = current_user
+      end
+
+      after { Current.reset }
+
+      it 'uses the explicitly provided verifier_id when it is not the same as the previous value' do
+        teacher_info = create(:teacher_info, verification_status: TeacherInfo::APPROVED, verifier_id: provided_verifier.id)
+        teacher_info.update!(verification_status: TeacherInfo::DENIED, verifier_id: new_verifier.id)
+        expect(teacher_info.verifier_id).to eq(new_verifier.id)
```
anathomical, line 176: Another approach here:
```ruby
subject { teacher_info.update!(verification_status: updated_verification_status, verifier_id:) }

let(:updated_verification_status) { TeacherInfo::DENIED }
let(:verifier_id) { new_verifier.id }

it do
  subject
  expect(teacher_info.verifier_id).to eq(new_verifier.id)
end
```

services/QuillLMS/spec/models/teacher_info_spec.rb
```diff
@@ -143,4 +144,112 @@
...
+      end
+
+      it 'uses the explicitly provided verifier_id when it is the same as the previous value' do
+        teacher_info = create(:teacher_info, verification_status: TeacherInfo::APPROVED, verifier_id: provided_verifier.id)
+        teacher_info.update!(verification_status: TeacherInfo::DENIED, verifier_id: provided_verifier.id)
+        expect(teacher_info.verifier_id).to eq(provided_verifier.id)
```
anathomical, line 181: Following my previously-discussed pattern, this could then become:
```ruby
context 'when explicitly setting the verifier to the Current.user' do
  let(:verifier_id) { provided_verifier.id }

  it do
    subject
    expect(teacher_info.verifier_id).to eq(provided_verifier.id)
  end
end
```

anathomical (approved):
I've got a few style comments around specs, but the fundamental approach you're using here looks good.  I appreciate the thorough use of callbacks to ensure that the state of verification stays updated through various changes to other models.

happythenewsad (commented):
Holding off on reviewing until the team discussion re: Current

services/QuillLMS/app/views/cms/users/_form.html.erb
emilia-friedberg: No, good question. I thought I needed it because the actual attribute is "approval_status", not "verification_status" here, but Peter Sharkey requested the copy change for consistency, but actually using the symbol works just fine.

happythenewsad (approved):
LGTM

PR #12823, opened 2025-03-06
do-not-list-units-from-archived-classrooms

anathomical:
## WHAT
Update the logic that populates the actiivty packs dropdown in the Data Export report
## WHY
To hide units that are associated with classrooms the teacher was involved with that have been archived since those classrooms are no longer relevant to the teacher
## HOW
Update the query used to generate this list so that it joins `units` to `classrooms` to ensure that we only include the former if the latter is `visible: true`.

### Screenshots
OLD
image
NEW
image

### Notion Card Links
https://www.notion.so/quill/Can-activity-packs-that-have-been-deleted-be-removed-from-the-All-activity-packs-dropdown-menu-on-th-1a8d42e6f94180cea64bc310acaa006c

- In staging confirm that the "old" 2023 activities don't show up for this user anymore
- (There is data in production mentioned in this Notion card that isn't in staging yet, so...) Run the generated query in metabase against the production database to confirm that it produces the intended list of activity packs

PR Checklist | Your Answer
------------ | -------------


emilia-friedberg (approved):
This query looks good to me, but I'm confused about why the join only to `classrooms.visible = TRUE` isn't already doing what we want here.

anathomical:
> This query looks good to me, but I'm confused about why the join only to `classrooms.visible = TRUE` isn't already doing what we want here.

It took me some time to figure this out, too, but it has to do with the fact that the `JOIN units` doesn't actually join on `classrooms`.  So what happens is you get this massive join where every Classroom has every Unit "assigned" to it, not just the ones that are in that Classroom.  Then we filter out the non-visible Classrooms, but the Units from those classrooms are just linked to classrooms that are visible.

PR #12825, opened 2025-03-07
include-visible-false-activity-sessions

anathomical:
## WHAT
Use unscoped when finding activity_sessions
## WHY
So that the worker doesn't crash if the activity_session is marked visible: false.  We still want to fill this data for those records.
## HOW
Add `.unscoped` to the model find query

- Tested in local db with an activity session marked visible false gets processed without error

PR Checklist | Your Answer
------------ | -------------


happythenewsad (approved):

dandrabik (approved):
Thanks for cleaning this up.

PR #12826, opened 2025-03-07
Add LLM Workflow backend

brendanshean:
## WHAT

Add the following models to our `Evidence::Research::GenAI` namespace:
`Workflow` 
`WorkflowStep`
`StepConnection`
`WorkflowExecution`
`StepExecution`

## WHY
The Trial runner currently only allows for a simple input-output prompt interaction with an LLM.  

The following three models are a way of persisting a DAG in the database.
`Workflow` - pointer to the whole graph
`WorkflowStep` - represents a node in the graph
`StepConnection` - represents a directed edge in the graph

```mermaid
flowchart TD
    classDef titleClass fill:none,stroke:none,font-size:24px,font-weight:bold;
    title["Workflow"]:::titleClass
    
    workflow_step1 -->|step_connection1| workflow_step2
    workflow_step1 -->|step_connection2| workflow_step3
```

`WorkflowExecution` instantiates a `Workflow` for each `TestExample` in a `Trial`.  
`StepExecution` stores intermediate values in the graph which aids in results investigation.

This abstraction gives a framework for arbitrary computation graphs (e.g. two-step prompts)

## HOW
For a trial, users currently select a `LLMPromptTemplate` with which to run over a set of `TestExample` the plan here is to give the option to instead select a `Workflow` where a chain of steps is specified.

`Workflow` has a `template_id` field that is nil for templates.  All others are instantiated for use in a `Trial`.  (I debated introducing a separate class `WorkflowTemplate` to model both these objects but opted not to as it would require corresponding WorkflowStepTemplate classes along with STI which felt heavy.

`WorkflowStep` uses STI;  for this PR, the only type is `PromptChainStep` but the list will expand to `ParallelizationStep` and `RoutingStep`. (see [post](https://www.anthropic.com/research/building-effective-agents) for more details).

When a user selects a `Workflow` for a `Trial`, a duplicate is made from the `Workflow` template and linked with the trial.  This derived `Workflow` is then used a script to be run over all test examples in the Trial which is facilitated by `WorkflowExecution` and `StepExecution`.

`WorkflowExecution` starts by calling `run` on any WorkflowStep nodes that have no StepConnections coming in (e.g. workflow_step_1 in the diagram above).

It then iterates through the remaining nodes in the graph, passing the output from one node to the next nodes as input.  When all nodes are complete, final input is stored in `WorkflowExecution`.

### Screenshots

### Notion Card Links
https://www.notion.so/quill/Breaking-up-Primary-Feedback-into-2-prompts-Optimal-Suboptimal-Suboptimal-Feedback-144d42e6f941801e8443f9ed22b8ba33?pvs=4

I've tested this locally and on staging.

PR Checklist | Your Answer
------------ | -------------


services/QuillLMS/engines/evidence/app/models/evidence/research/gen_ai/step_connection.rb
```diff
@@ -0,0 +1,44 @@
...
+        end
+
+        private def steps_belong_to_same_workflow
+          return unless source_step && target_step
+          return if source_step.workflow_id == target_step.workflow_id
+
+          errors.add(:base, 'Source and target steps must belong to the same workflow')
+        end
+
+        private def no_circular_references
```
dandrabik, line 35: Probably worth expanding this logic for multi-step circular references (either here or somewhere else), e.g. 1->2->1

dandrabik (approved):
This looks really good, thanks for putting this together. One small note about checking for multi-step circular references, but that's also not a blocker and can be done later.

Looking forward to experimenting with this!

services/QuillLMS/engines/evidence/app/models/evidence/research/gen_ai/step_connection.rb
brendanshean: When building this out, I first [spiked](https://github.com/empirical-org/Empirical-Core/commit/12efb792f8f019bab7fd8815368eefb1071ff915) out a proof-of-concept to accommodate all four types outlined in https://www.anthropic.com/research/building-effective-agents.

The EvaluatorOptimizer node will have two components, an evaluator and an optimizer, as well as the evaluation condition which determines if the node should be called on itself again.  This permits a looping operation within the node, but the overall graph structure remains acyclic, which makes various graph operations (e.g. traversal) easier.

PR #12835, opened 2025-03-12
Evidence - OnTopic Detector

dandrabik:
## WHAT
A revised version of the OnTopic Detector focused on 'gibberish', 'frustration', and 'random off topic' responses.

Note, this is just the service and isn't turned on in the Evidence Feedback yet (that PR will come next).

## WHY
Two main reasons:
- The feedback for these is very different that the general GenAI feedback (we use 'clear your answer and start again' feedback). So by isolating these OffTopic responses, we can make the GenAI prompt simpler
- Responding to random off-topic entries often leads the GenAi feedback down bad paths, so we'd like to isolate them and use static feedback
Screenshot 2025-03-12 at 4 57 07 PM

## HOW
Start with a data set across many prompts with a few on-topic response and a few off-topic responses. Iterate on the algorithm to find one that works. Also use a test set to pull student data from a single passage and run 500 examples on that. Confirm reasonable results.

Landed on: 
- Check for a matching 7-character+, non-generic word between the entry and passage
- Otherwise send to an LLM with a simple prompt that asks if this is on topic for the passage. Ask it to account for spelling mistakes. 
### Screenshots
Screenshot 2025-03-12 at 5 01 07 PM

### Notion Card Links
https://www.notion.so/quill/OnTopic-Detector-V2-19fd42e6f9418077a736e0ea75b3702a

Running the script many times on data sets until the answers improved.

PR Checklist | Your Answer
------------ | -------------
Have you added and/or updated tests? |  Yes, Note, there's some issue with the code that is causing local tests to pass, but fail on Circle. Still debugging. Something to do with my TypeCastData not type-casting on some tests (or something).
Have you deployed to Staging? | No, there's no live code here yet.


happythenewsad (approved):

brendanshean (approved):
Nice work on this.  I really like the introduction of the `TypeCastData` class and then the update to the `Result` class

PR #12837, opened 2025-03-13
Diagnostic insights skill performance api

anathomical:
## WHAT
- Create a new service that
- - Finds all students in a listed classroom
- - Finds all diagnostic performance for each of those students
- - Identifies which students do not have performance (due to not completing the diagnostic)
- - Shapes that data into an agreed-upon data shape for the front-end to parse
- Add a controller endpoint to enable access to the new service from the front-end
- Update thor tasks so that we can record snapshots of the new API shape to validate the stability of the data
- - Save an initial version of the snapshot
## WHY
In order to display data for the Performance Insights section of the Diagnostic Insights reports, we need to be able to provided data broken out by skill with each student's best performance on each skill from the timeframe displayed.
## HOW
- Make one database query for the classroom using `includes` to handle joins
- Filter and aggregate data in Ruby to achieve the intended payload shape
- Add a controller endpoint to mediate between the front-end and the new service re-using timeframe logic from existing Diagnostic Insights API endpoints

### Notion Card Links
https://www.notion.so/quill/Build-Diagnostic-Insights-Skill-Performance-API-1abd42e6f94180a5af73e392cfefe93f

Using a test account (classroom id: 1803821)
- Run the roll-up for just 2025
- Compare service payload generation against Teacher Diagnostic Report (by restricting to just 2025, the reports should match)
- Run the roll-up without the time restriction
- Confirm that students are rated the same OR BETTER allowing for potential better performances on earlier assignments

PR Checklist | Your Answer
------------ | -------------


services/QuillLMS/app/services/diagnostic_insights/skill_performance_by_classroom.rb
```diff
@@ -0,0 +1,180 @@
+# frozen_string_literal: true
```
anathomical, line 1: I'm especially interested in feedback about how this code is organized.  There's a lot of filtering and data-shape manipulation that's chained together in this service.  More than I'm comfortable with, to be completely honest, but a cleaner approach wasn't obvious to me.  Still, I welcome any thoughts folks have on how to make this more readable and easier to understand.

services/QuillLMS/app/services/diagnostic_insights/skill_performance_by_classroom.rb
```diff
@@ -0,0 +1,180 @@
...
+    def initialize(classroom_id, diagnostic_type, start_date: nil, end_date: nil)
+      @classroom_id = classroom_id
+      @diagnostic_type = validate_diagnostic_type(diagnostic_type)
+      @start_date = start_date
+      @end_date = end_date
+    end
+
+    def run = results
+
+    private def results = clean_buckets
```
dandrabik: You should remove this passthrough method.

services/QuillLMS/app/services/diagnostic_insights/skill_performance_by_classroom.rb
```diff
@@ -0,0 +1,180 @@
...
+    attr_reader :classroom_id, :diagnostic_type, :start_date, :end_date
+
+    def initialize(classroom_id, diagnostic_type, start_date: nil, end_date: nil)
+      @classroom_id = classroom_id
+      @diagnostic_type = validate_diagnostic_type(diagnostic_type)
+      @start_date = start_date
+      @end_date = end_date
+    end
+
+    def run = results
```
happythenewsad: `def run = clean_buckets`

`private def results = clean_buckets` is currently redundant

services/QuillLMS/app/services/diagnostic_insights/skill_performance_by_classroom.rb
happythenewsad: I think your spidey sense is correct. Here's a sketch of a refactor (I haven't tested). It would get rid of a lot of the `map` calls. As a bonus, pure lambdas are less buggy and the test setup is lighter weight.

```ruby 
  ADD_SKILL_METADATA = ->(skill_performances) {
    first_performance = skill_performances.values.flatten.first
    skill_performances.merge({
      'skill_id' => first_performance['skill_id'],
      'skill_name' => first_performance['skill_name']
    })
  }

  ADD_NO_DATA_STUDENTS = ->(skill_performances) {
    user_ids_with_data = skill_performances.values.flatten.map { |performance| performance['user_id'] }
    no_data_students = students
      .reject { |student| user_ids_with_data.include?(student.id) }
      .map { |student| { 'user_id' => student.id, 'user_name' => student.name } }
    
    skill_performances.merge({ NO_DATA => no_data_students })
  }

  GROUP_BY_PROFICIENCY = ->(skill_performances) {
    skill_performances.group_by { |skill_performance| determine_proficiency_bucket(skill_performance) }
  }

private def clean_buckets
  # Single map operation with composed transformations
  performance_grouped_by_skills
    .map do |skills|
      skills
        .then(GROUP_BY_PROFIENCY)
        .then(ADD_NO_DATA_STUDENTS)
        .then(ADD_SKILL_METADATA)
        # et cetera
    end
end


```

dandrabik: Yeah, I think some refactor would be nice here, but I'm not sure the best route. Here are some potential ideas that you might consider (in addition to PK's suggestion). As the code stands now, I think it will be slightly hard to test and debug.

1. Break into multiple services. Maybe one to fetch the all the data columns need, and one to group, decorate, and manipulate the raw data. Or some other breakdown of services that have clear inputs and outputs.
2. Change the data shape to make the calculation easier, e.g. flatten it. This could also be part of the multiple services breakdown. Example: 
```json
{
      "user_id": 11403905,
       "completed_at": "2024-11-22 15:58:32 UTC",
       "question_count": 7,
       "question_total": 10,
       "skill_id" : 225,
       "diagnostic_id" : 762
  }
```
3. Consider using presenter objects if they are helpful, e.g. would this be cleaner if there was a `SkillPerformance` class
```ruby
class SkillPerformance

  def initialize(hash, user_id, ...)
  end
  
# Bunch of helper methods
  def proficiency_bucket
    return 'full_proficiency' if questions_correct == questions_total
    return 'no_proficiency' if questions_correct.zero?
    'partial_proficiency'
  end
end
```
I'm having a hard time recommending anything specific here, so just trying to present some general options.

dandrabik (commented):
I just left some general comments here, but I think this is worth another look and have some refactoring. That being said, we do want to keep this moving, so I would timebox a refactor to like half a day.

services/QuillLMS/app/services/diagnostic_insights/skill_performance_by_classroom.rb
brendanshean: I spoke with Thomas about this question earlier today and one of the solutions involved the `then` block.  While this approach feels the cleanest of the ones we looked at, I was hesitant to recommend it since it doesn't seem to be idiomatic Rails; there's very little mention or documentation or use of `then` online.  Makes me wonder if there's some downside/wrong with it.

```
def run
  fetch_students
    .then { |students| build_skill_entries(students) }
    .then { |entries| group_by_skill(entries) }
    .then { |skills| group_by_proficiency(skills) }
    .then { |skills| attach_no_data_students(skills) }
    .then { |skills| shape_final_output(skills) }
end
```

anathomical: I combined some of @brendanshean's and some of @happythenewsad's feedback and decided to try this with a `then` call chain.  I took a little bit of time to try options around some of Dan's suggestions, but they felt similarly complex to keep in mind at once.  Open to further changes if anyone thinks they're worth pursuing, but I'm mindful of trying to keep this somewhat time-boxed.

happythenewsad (approved):

brendanshean (approved):
Nice work on pulling together this complex transformation.

services/QuillLMS/app/services/diagnostic_insights/skill_performance_by_classroom.rb
```diff
@@ -0,0 +1,191 @@
...
+      students
+        .then { |payload| find_all_skill_performances(payload) }
+        .then { |payload| group_performances_by_skill_for_student(payload) }
+        .then { |payload| find_best_performance_per_skill(payload) }
+        .then { |payload| attach_user_data_to_skill_payloads(payload) }
+        .then { |payload| group_by_skills_for_classroom(payload) }
+        .then { |payload| group_performance_by_proficiency(payload) }
+        .then { |payload| add_no_data_students(payload) }
+        .then { |payload| attach_skill_information_to_buckets(payload) }
+        .then { |payload| remove_skill_information_from_performance(payload) }
```
emilia-friedberg: I really like how this reads, personally.

emilia-friedberg (approved):

services/QuillLMS/app/services/diagnostic_insights/skill_performance_by_classroom.rb
```diff
@@ -0,0 +1,191 @@
...
+    def initialize(classroom_id, diagnostic_type, start_date: nil, end_date: nil)
+      @classroom_id = classroom_id
+      @diagnostic_type = validate_diagnostic_type(diagnostic_type)
+      @start_date = start_date
+      @end_date = end_date
+    end
+
+    def run
+      students
+        .then { |payload| find_all_skill_performances(payload) }
```
dandrabik: This `.then` syntax is very readable, I like it. One nit, is that the params all being `payload` makes it seem like all the inputs and outputs are the same shape, when they are not. Consider using more specific names for the params.

services/QuillLMS/app/services/diagnostic_insights/skill_performance_by_classroom.rb
```diff
@@ -0,0 +1,191 @@
...
+        .then { |payload| find_best_performance_per_skill(payload) }
+        .then { |payload| attach_user_data_to_skill_payloads(payload) }
+        .then { |payload| group_by_skills_for_classroom(payload) }
+        .then { |payload| group_performance_by_proficiency(payload) }
+        .then { |payload| add_no_data_students(payload) }
+        .then { |payload| attach_skill_information_to_buckets(payload) }
+        .then { |payload| remove_skill_information_from_performance(payload) }
+    end
+
+    private def remove_skill_information_from_performance(payload)
```
dandrabik: I think I would use more specific param names here and in the following methods (that I believe you had before). `payload` is very vague and makes this slightly harder to understand. As a reminder, the name of the param in the method doesn't need to match the method var name passed in, so even if you call it via `remove_skill_information_from_performance(payload)` you can have it defined as `def remove_skill_information_from_performance(skill_performances_array)`

dandrabik (approved):
Small naming suggestions that you can take or leave. Nothing to hold this up from shipping.

PR #12843, opened 2025-03-18
diagnostic insights teacher verification flow

emilia-friedberg:
## WHAT
Build out the backend for teacher verification.

## WHY
This flow is the primary way that admins will verify teachers, allowing them access to Diagnostic Insights reports.

## HOW
Add all the routing, workers, emails etc to make it possible for admins to verify teachers, plus some small erb templates to render when they've taken action on verification.

### Screenshots
Page admins see when they click the "Approve request" email link
Screenshot 2025-03-19 at 11 23 07 AM

Page admins see when they click the "Deny request" email link
Screenshot 2025-03-19 at 11 24 42 AM

Page admins see when they click either "Approve" or "Deny" in the email but the user has already had their verification processed
Screenshot 2025-03-19 at 11 23 21 AM

### Notion Card Links
https://www.notion.so/quill/Diagnostic-Insights-Teacher-Verification-Flow-1bbd42e6f9418014bbbfe018aba03880?pvs=4

Tested locally, will have Peter Sharkey QA on staging once the last piece of this is ready.

PR Checklist | Your Answer
------------ | -------------


services/QuillLMS/app/controllers/admins_controller.rb
```diff
@@ -132,6 +144,39 @@ def deny_admin_request
...
+  end
+
+  def deny_teacher_verifications
+    teacher_ids = params[:teacher_ids]
+
+    teacher_ids.each { |teacher_id| handle_deny_teacher_verification_request(teacher_id) }
+
+    reset_admin_users_cache
+    render json: { message: teacher_ids.length > 1 ? t('admin.deny_teacher_verifications') : t('admin.deny_teacher_verification') }, status: 200
+  end
```
emilia-friedberg: These endpoints aren't being called anywhere right now, but they will be in the third and final PR.

services/QuillLMS/app/controllers/admins_controller.rb
```diff
@@ -28,6 +34,12 @@ class AdminsController < ApplicationController
       sign_in_account_settings
     ]
 
+  before_action :redirect_if_teacher_verification_request_already_processed,
+    only: %w[
+      approve_teacher_verification
+      deny_teacher_verification
+    ]
```
anathomical: This isn't an issue that you're introducing in this PR, but I notice that we have a mix of `only: %w[some_string]` and `only: [:some_symbol]` in this controller, and we should probably standardize.  I believe that, generally speaking, the symbol form is preferred in Ruby.

services/QuillLMS/app/controllers/admins_controller.rb
```diff
@@ -132,6 +144,39 @@ def deny_admin_request
...
+  def deny_teacher_verification
+    handle_deny_teacher_verification_request(@teacher.id)
+
+    reset_admin_users_cache
+  end
+
+  def teacher_verification_request_already_processed; end
+
+  def approve_teacher_verifications
+    teacher_ids = params[:teacher_ids]
```
anathomical: Because `teacher_ids` is used in more than one action, I'd probably put it in a one-line method:
```ruby
private def teacher_ids = params[:teacher_ids]
```
That saves you from having to explicitly define it in every relevant action.

services/QuillLMS/app/controllers/admins_controller.rb
```diff
@@ -132,6 +144,39 @@ def deny_admin_request
...
+
+    teacher_ids.each { |teacher_id| handle_approve_teacher_verification_request(teacher_id) }
+
+    reset_admin_users_cache
+
+    render json: { message: teacher_ids.length > 1 ? t('admin.approve_teacher_verifications') : t('admin.approve_teacher_verification') }, status: 200
```
anathomical: This is a take-or-leave thing, but I'd probably also extract a bunch of the values here into one-line methods:
```ruby
private def multiple_teacher_ids? = teacher_ids.length > 1
private def singular_teacher_approval_message = 'admin.approve_teacher_verification'
private def multiple_teacher_approval_message = 'admin.approve_teacher_verifications'
private def teacher_approval_message = multiple_teacher_ids? ? multiple_teacher_approval_message : singular_teacher_approval_message
```
Then you can just `render json: { message: teacher_approval_message }, status: 200`.  It doesn't really change the complexity, but it shifts it around a bit.  Like I said, take or leave this based on which one you find easier to keep in your head.

services/QuillLMS/app/controllers/admins_controller.rb
anathomical: Same take-or-leave notes as I had in the `approve_teacher_verifications` action.

services/QuillLMS/app/controllers/admins_controller.rb
```diff
@@ -245,4 +290,24 @@ def create_and_link_accounts
...
+    TeacherVerificationApprovedAnalyticsWorker.perform_async(teacher_id)
+    TeacherVerificationApprovedEmailWorker.perform_async(teacher_id)
+  end
+
+  private def handle_deny_teacher_verification_request(teacher_id)
+    teacher = User.find_by(id: teacher_id)
+    teacher.update!(current_teacher_verifier: current_user, teacher_verification_status: TeacherInfo::DENIED)
+    TeacherVerificationDeniedAnalyticsWorker.perform_async(teacher_id)
+    TeacherVerificationDeniedEmailWorker.perform_async(teacher_id)
+  end
```
anathomical, line 302: Another take-or-leave thing, but I'd be tempted to move this logic into the `User` model or the `Teacher` concern.  So you'd have something like:
```ruby
class User
  def approve_teacher_verification(verifier_id)
    update(current_user_verifier_id: verifier_id, teacher_verification_status: TeacherInfo::APPROVED)
    TeacherVerificationApprovedAnalyticsWorker.perform_async(teacher_id)
    TeacherVerificationApprovedEmailWorker.perform_async(teacher_id)
  end
end
```
Then in the controller you can just call:
```ruby
teacher.approve_teacher_verification
```
I don't think this actually does much in this particular PR, but if we ever add additional paths to approving a teacher, it makes sure that the analytics and email stuff happens without having to duplicate that logic in every place you can approve someone.

services/QuillLMS/app/controllers/teacher_infos_controller.rb
```diff
@@ -28,6 +28,20 @@ def update
     }, status: 200
   end
 
+  def request_teacher_verification
+    admin_ids = request_teacher_verification_params[:admin_ids]
+    message = request_teacher_verification_params[:message]
```
anathomical, line 33: Beyond the scope of this PR, but it's worth flagging that by not persisting `message` anywhere we make it impossible for a different admin to view the list of all pending requests with their messages.  We don't need to do that right now, but it's worth having on our radar as we consider expanding permissions management.

services/QuillLMS/spec/controllers/admins_controller_spec.rb
```diff
@@ -251,6 +251,141 @@
     end
   end
 
+  describe '#approve_teacher_verification' do
+    context 'when teacher verification request is not processed' do
+      it 'updates the teacher record with approved status and sets the current teacher verifier' do
+        post :approve_teacher_verification, params: { id: teacher.id }
+        teacher.reload
+        expect(teacher.teacher_verifier_id).to eq(admin.id)
+        expect(teacher.teacher_verification_status).to eq(TeacherInfo::APPROVED)
```
anathomical: Take or leave, but when I see `reload` I often think about whether a `expect to change` call would make sense.  I think you could rewrite this with one, though it may not be better:
```ruby
expect { post :approve_teacher_verification, params: { id: teacher.id } }
  .to change(teacher, :teacher_verifier_id).from(nil).to(admin.id)
  .and change(teacher, :teacher_verification_status).from(nil).to(TeacherInfo::APPROVED)
```

services/QuillLMS/spec/controllers/admins_controller_spec.rb
```diff
@@ -251,6 +251,141 @@
     end
   end
 
+  describe '#approve_teacher_verification' do
+    context 'when teacher verification request is not processed' do
+      it 'updates the teacher record with approved status and sets the current teacher verifier' do
+        post :approve_teacher_verification, params: { id: teacher.id }
```
anathomical: I'd probably make a `subject` definition for this call: `subject { post :approve_teacher_verification, params: { id: teacher.id } }`.  That lets you just call `subject` any time you're duplicating this whole `post` call.

services/QuillLMS/spec/controllers/admins_controller_spec.rb
```diff
@@ -251,6 +251,141 @@
...
+      it 'redirects to the teacher verification request already processed page' do
+        post :approve_teacher_verification, params: { id: teacher.id }
+        expect(response).to redirect_to("/users/#{teacher.id}/teacher_verification_request_already_processed")
+      end
+
+      it 'does not update the teacher record' do
+        original_status = teacher.teacher_verification_status
+        post :approve_teacher_verification, params: { id: teacher.id }
+        teacher.reload
+        expect(teacher.teacher_verification_status).to eq(original_status)
```
anathomical: This is, to me, an even clearer use-case for `expect to not_change`:
```ruby
expect { subject }.to not_change(teacher, :teacher_verification_status)
```

services/QuillLMS/spec/controllers/admins_controller_spec.rb
```diff
@@ -251,6 +251,141 @@
...
+    it 'updates each teacher record to approved and sets the current teacher verifier' do
+      post :approve_teacher_verifications, params: { teacher_ids: [teacher.id, teacher2.id] }
+
+      teacher.reload
+      teacher2.reload
+
+      expect(teacher.teacher_verifier_id).to eq(admin.id)
+      expect(teacher.teacher_verification_status).to eq(TeacherInfo::APPROVED)
+      expect(teacher2.teacher_verifier_id).to eq(admin.id)
+      expect(teacher2.teacher_verification_status).to eq(TeacherInfo::APPROVED)
```
anathomical: Another place I'd use `expect to change`
```ruby
expect { subject }
  .to change(teacher, :teacher_verifier_id).to(admin.id)
  .and change(teacher, :teacher_verification_status).to(TeacherInfo::APPROVED)
  .and change(teacher2, :teacher_verifier_id).to(admin.id)
  .and change(teacher2, :teacer_verification_status).to (TeacherInfo::APPROVED)
```

services/QuillLMS/spec/controllers/admins_controller_spec.rb
```diff
@@ -251,6 +251,141 @@
...
+      expect {
+        post :approve_teacher_verifications, params: { teacher_ids: [teacher.id, teacher2.id] }
+      }.to change(TeacherVerificationApprovedAnalyticsWorker.jobs, :size).by(2)
+        .and change(TeacherVerificationApprovedEmailWorker.jobs, :size).by(2)
+    end
+
+    it 'returns a pluralized success message' do
+      post :approve_teacher_verifications, params: { teacher_ids: [teacher.id, teacher2.id] }
+      response_message = JSON.parse(response.body)['message']
+      expect(response_message).to eq(I18n.t('admin.approve_teacher_verifications'))
```
anathomical, line 358: Take or leave this, but having the constant string defined here makes me wonder if the controller should actually define these as constants directly, so that you could references those constants in the spec:
```ruby
class AdminsController
  APPROVE_TEACHER_VERIFICATIONS_MESSAGE = 'admin.approve_teacher_verifications'
end
```
```
expect(response_message).to eq(described_class::APPROVE_TEACHER_VERIFICATIONS_MESSAGE)
```

services/QuillLMS/spec/controllers/admins_controller_spec.rb
```diff
@@ -251,6 +251,141 @@
...
+    it 'updates each teacher record to denied and sets the current teacher verifier' do
+      post :deny_teacher_verifications, params: { teacher_ids: [teacher.id, teacher2.id] }
+
+      teacher.reload
+      teacher2.reload
+
+      expect(teacher.teacher_verifier_id).to eq(admin.id)
+      expect(teacher.teacher_verification_status).to eq(TeacherInfo::DENIED)
+      expect(teacher2.teacher_verifier_id).to eq(admin.id)
+      expect(teacher2.teacher_verification_status).to eq(TeacherInfo::DENIED)
```
anathomical: Take or leave again, but as with the multi-approval process, I'd personally do these with `expect to change`

services/QuillLMS/app/controllers/teacher_infos_controller.rb
```diff
@@ -28,6 +28,20 @@ def update
...
+
+    @teacher_info.update!(verification_status: TeacherInfo::PENDING)
```
anathomical, line 35: Take or leave, but I'd be tempted to put this `update` logic on the model rather than the controller.
```ruby
class User
  def request_teacher_verification
    update!(verification_status: TeacherInfo::PENDING)
  end
end
```
That will make it easier to add additional pathways into verification workflows in the future.

services/QuillLMS/spec/models/schools_users_spec.rb
```diff
@@ -36,13 +36,15 @@
...
     context 'when user has teacher_info' do
       let(:teacher_info) { create(:teacher_info, :verification_approved, verifier_id: 123) }
       let(:user_with_teacher_info) { create(:user, teacher_info:) }
-      let(:schools_user) { create(:schools_users, user: user_with_teacher_info, school:) }
+      let!(:schools_user) { create(:schools_users, user: user_with_teacher_info, school:) }
 
-      it 'resets teacher_info verification status and verifier_id after save' do
-        schools_user.save
+      it 'resets teacher_info verification status and verifier_id after school is changed' do
+        schools_user.update(school: other_school)
```
anathomical, line 47: The relevant part of this test isn't part of this PR, so definitely take or leave this, but again I'd do this with
```ruby
expect { schools_user.update(school: other_school) }
  .to change(teacher_info, :verification_status).to(nil)
  .and change(teacher_info, :verifier_id).to(nil)
```

services/QuillLMS/spec/mailers/user_mailer_spec.rb
```diff
@@ -169,6 +169,47 @@
...
 
+  describe 'teacher_requested_verification_email' do
+    let(:teacher) { create(:teacher) }
+    let!(:admin) { create(:admin) }
+    let(:message) { 'Some message' }
+    let(:school) { create(:school) }
+    let!(:schools_user) { create(:schools_users, user: teacher, school:) }
+    let(:mail) { described_class.teacher_requested_verification_email(teacher, admin, message) }
+
+    before { teacher.reload }
```
anathomical, line 180: I'm totally willing to believe that this is a necessary line, but I can't imagine what `teacher.reload` would do in a `before` block since there haven't been any changes yet when this is triggered.  So are the tests actually failing without this?  (In an ideal world, I'd say figure out why and fix it, but it's not a big enough deal to be worth you spending a couple of hours on, I don't think.)

anathomical (approved):
This looks good.  A couple of suggestions about moving logic from controllers into modles/concerns, but it's all take-or-leave stylistic stuff.  Thanks for putting together some pretty complex use-cases.

eadams17 (approved):

services/QuillLMS/app/controllers/admins_controller.rb
emilia-friedberg: Yeah, this is a good call-out. We actually do have another way to verify teachers right now (through the staff panel), but we don't want the same email and analytics to fire there, so I think I'm going to leave this in place right now because by the time we're moving `admin_approve_teacher_verification` into the model/concern it starts to feel a little too specific to me. I appreciate the note, though, and I'll keep that in mind moving forward.

services/QuillLMS/spec/controllers/admins_controller_spec.rb
emilia-friedberg: For some reason, this rewrite, and the others as well, are failing, but reloading works to get the result I expect. I like your syntax better here, but I don't want to spend a ton of time troubleshooting these tests so I'm keeping the `subject` change but stick with the original syntax. I will keep that in mind going forwards, though.

services/QuillLMS/spec/mailers/user_mailer_spec.rb
emilia-friedberg: It seems like it's actually forcing `teacher.school` to work correctly -- for some reason, even though `schools_user` is instantiated with a `let!`, teacher.school returns undefined otherwise.

PR #12848, opened 2025-03-20
fix bug with individual student diagnostic responses

emilia-friedberg:
## WHAT
Fix the #individual_student_diagnostic_responses so it both retrieves and caches the correct activity session when there are multiple diagnostic activities with the same classroom unit.

## WHY
A teacher who had assigned multiple diagnostic activities in one activity pack was encountering this issue.

## HOW
Update both the cache and the activity finder method to care about `activity_id`.

### Screenshots

### Notion Card Links
https://www.notion.so/quill/A-teacher-sees-a-student-s-Advanced-Diagnostic-Baseline-Pre-data-when-they-click-on-the-Intermedia-1bcd42e6f9418055a2b9ffb188293c55?pvs=4

Tested locally to confirm the fix works.

PR Checklist | Your Answer
------------ | -------------
Have you deployed to Staging? | Tested locally with staging data


brendanshean (approved):
Thank you for fixing this.  I appreciate the testing!

happythenewsad (approved):

services/QuillLMS/spec/controllers/teachers/progress_reports/diagnostic_reports_controller_spec.rb
```diff
@@ -515,4 +515,114 @@
...
+          activity: activity1,
+          classroom_unit:,
+          state: 'finished',
+          user: student)
+      end
+
+      before do
+        # Stub out methods that format results so we can predict the JSON structure.
+        allow(controller).to receive(:format_concept_results) do |session, _results_relation|
+          session == activity_session1 ? ['session1_concept'] : ['unknown']
```
anathomical: Is this ternary necessary?  It looks like there are only tests that go through one branch.

services/QuillLMS/spec/controllers/teachers/progress_reports/diagnostic_reports_controller_spec.rb
```diff
@@ -515,4 +515,114 @@
...
+      end
+
+      before do
+        allow(controller).to receive(:format_concept_results) do |session, _results_relation|
+          if session.id == activity_session1.id
+            ['concept1']
+          elsif session.id == activity_session2.id
+            ['concept2']
+          else
+            ['unknown']
```
anathomical: Similar to the above, do we need this `else`?  It's fine to leave these in if we think we might add more tests later that need them.

anathomical (approved):
Nice job chasing down the issue here.

services/QuillLMS/spec/controllers/teachers/progress_reports/diagnostic_reports_controller_spec.rb
emilia-friedberg: It just helps confirm that the right session is getting passed to `format_concept_results`, but maybe just returning null otherwise is cleaner (and will mirror the logic before, where I think you're right that we don't specifically need `unknown`.

PR #12853, opened 2025-03-21
sls-handle-completion-performance-enhancement

anathomical:
## WHAT
User `eager_load` instead of `includes` when fetching data.  While `includes` uses `OUTER JOIN` in the initial query, ActiveRecord won't populate relevant models with `includes` and makes follow-up queries for data that's contained in that initial join.  Using `eager_load` results in the same query, but avoids follow-up queries.
## WHY
We need to fetch data from multiple tables for this service to work, and we were doing this via a separate query for each table.  That's inefficient which matters extra for a job we A) run regularly.
## HOW
Replace `includes` with `eager_load` directive to initial `ActivitySession` query so that the related models are fetched up front.

- Trigger calls to the service in console
- Watch the SQL calls triggered by the call that pop up in the development console
- Confirm that we go from 5 separate `SELECT`s to two (one for precursor data, and one for the target `StudentLearningSequenceActivity` to update)

PR Checklist | Your Answer
------------ | -------------
Have you added and/or updated tests? | No.  I don't know that we have a clean method for measuring SQL round-trips in specs


dandrabik (approved):
:+1: Looks good.

happythenewsad (approved):

PR #12861, opened 2025-03-26
fix repeatable value for diagnostic activities

emilia-friedberg:
## WHAT
Default `repeatable` to `false` for Diagnostic activities, and prevent save if it's set to `true`.

## WHY
In the course of testing something unrelated, I noticed that `repeatable` was set to `true` for a whole bunch of Diagnostic activities when it shouldn't be, allowing students to replay these activities on the frontend. I manually fixed those records, but we want to prevent this from happening again.

## HOW
Edit the writer method and add a `before_save` callback.

### Screenshots

### Notion Card Links

Tested locally that diagnostic activities instantiate correctly and are not saveable with `repeatable: true`.

PR Checklist | Your Answer
------------ | -------------


services/QuillLMS/spec/models/activity_spec.rb
```diff
@@ -877,4 +877,38 @@
       it { expect(subject).to eq(activity.data) }
     end
   end
+
+  describe 'diagnostic repeatable behavior' do
+    context 'when activity is diagnostic' do
+      let(:diagnostic_classification) { create(:diagnostic) }
+
+      it 'defaults repeatable to false on new record' do
+        diagnostic_activity = build(:activity, classification: diagnostic_classification)
```
anathomical: I'd move this `build` into a `let` since it's used in every `it` block in this `context`:
```ruby
let(:diagnostic_classification) { create(:diagnostic) }
let(:diagnostic_activity) { build(:activity, classification: diagnostic_classification) }
```

services/QuillLMS/app/models/activity.rb
```diff
@@ -153,6 +154,11 @@ class Activity < ApplicationRecord
 
   DEFAULT_MAX_GRADE_LEVEL = 12
 
+  def classification=(value)
+    super(value)
+    self.repeatable = false if value == ActivityClassification.diagnostic
+  end
```
anathomical, line 160: I didn't see a spec for this behavior, and it feels like it might be worth making one:
```
let(:diagnostic_classification) { create(:diagnostic) }
let(:repeatable) { true }
let(:activity) { create(:activity, repeatable:) }

it do
  expect { activity.classification = diagnostic_classification }
    .to change(activity.reload, :repeatable)
    .from(repeatable)
    .to(false)
end
```

anathomical (approved):
Thanks for finding and fixing this.  I think this is a clean approach to the fix.  A couple of take-or-leave notes on specs, but this could be merged as-is.

services/QuillLMS/spec/models/activity_spec.rb
```diff
@@ -877,4 +877,38 @@
...
+        expect(diagnostic_activity.repeatable).not_to be
```
brendanshean: Since repeatable is a boolean attribute, you can use the helper:

`expect(diagnostic_activity).not_to be_repeatable` which I think reads more fluidly.

brendanshean (approved):
Thank you for cleaning this up!

happythenewsad (approved):

PR #12863, opened 2025-03-27
improve NameGenerator

happythenewsad:
## WHAT / HOW
1. after two generation attempts, gracefully fallback to higher entropy name generation and generate alert
2. use singleton module methods over `extend self`
3. remove private method only used for benchmark testing
4. maintain name generation logic out of Classroom, in ClassCodeGenerator 
5. rename NameGenerator to ClassCodeGenerator
6. memoize expensive File IO

## WHY
1. to avoid endless recursion causing a stack overflow
2. for clarity. We could force supporting methods to be private via `private_class_method`, but stubability in RSpec would be lost
3. cleanup
4. separation of concerns
5. so that the generator won't accidentally be used to generate non-Classroom names
6. makes retries less computationally expensive

### Screenshots

### Notion Card Links

1. Create a classroom via the `Create a classroom` modal; observe no errors
2. Run the generator a few times from Heroku CLI:
Screenshot 2025-03-27 at 3 49 30 PM

PR Checklist | Your Answer
------------ | -------------


services/QuillLMS/app/lib/classroom_name_generator.rb
```diff
@@ -0,0 +1,27 @@
...
+
+module ClassroomNameGenerator
+  class MaxRetriesExceededError < StandardError; end
+
+  MAX_RETRY_COUNT = 2
+
+  def self.generate_unique_code
+    MAX_RETRY_COUNT.times do
+      code = generate
+      return code unless Classroom.unscoped.exists?(code:)
```
happythenewsad, line 11: slightly faster than `find_by_x`

services/QuillLMS/app/lib/classroom_name_generator.rb
```diff
@@ -0,0 +1,31 @@
...
+    generate_fallback_and_report_error
+  end
+
+  def self.generate_fallback_and_report_error
+    ErrorNotifier.report(MaxRetriesExceededError.new)
+
+    generate_fallback
+  end
+
+  def self.noun1 = @noun1 ||= File.readlines(Rails.root.join('db/noun_1.db'))
```
happythenewsad, line 22: over-optimization? maybe

services/QuillLMS/app/lib/class_code_generator.rb
```diff
@@ -0,0 +1,31 @@
...
+    generate_fallback
+  end
+
+  def self.noun1 = @noun1 ||= File.readlines(Rails.root.join('db/noun_1.db'))
+
+  def self.noun2 = @noun2 ||= File.readlines(Rails.root.join('db/noun_2.db'))
+
+  def self.generate = "#{sample(noun2)}-#{sample(noun1)}"
+
+  def self.generate_fallback = "#{generate}-#{rand(999)}"
```
anathomical, line 32: We may not need to worry about this, but it is at least theoretically possible for this to cause a collision if `rand(999)` returns the same thing for an already-extant class code.

I think we can probably just allow that to bubble up into some other error in the unlikely case that it happens rather than making this process more complicated.

anathomical (approved):
I like breaking this into a Service, and the clearer naming used here.

emilia-friedberg (approved):
LGTM!

services/QuillLMS/app/lib/class_code_generator.rb
happythenewsad: Your observation is correct; this is an implementation of what [we resolved during the post-mortem](https://www.notion.so/quill/Post-Mortem-Jan-13-2025-class-code-bug-17ad42e6f941802a8b9acd23905d350a). The intention is to gracefully alert developers when double-retries start to spike. At that time we will ship a product-facing change in the class code generation logic. We want to punt on the product-facing change for a while (couple years?).

dandrabik (changes requested):
If I'm reading this correctly, this doesn't change the core algorithm for generation (2 words from the same lists). With the same word list and a cap of 2 retries, the fallback `ErrorNotifier` would start firing almost immediately if deployed.

See the screenshot, a small test on the `staging` DB shows it only takes a few tries to get one over 2 retries. I don't think we want to deploy a solution that will immediately alert us that we need to update it. If we want to get this out to eliminate the infinite loop. We could make the limit much higher, say 100. Or we can edit the underlying algorithm.

Screenshot 2025-03-28 at 4 24 22 PM

happythenewsad:
> If I'm reading this correctly, this doesn't change the core algorithm for generation

Yes, PR is sticking to the resolution from the [post-mortem](https://www.notion.so/quill/Post-Mortem-Jan-13-2025-class-code-bug-17ad42e6f941802a8b9acd23905d350a)

> I don't think we want to deploy a solution that will immediately alert us that we need to update it

low threshold pro: we can detect the saturation of classcode names earlier
low threshold con: we get a steady drip of alerts

I parameterized this number with an environment variable and defaulted it to 10 for now. Hopefully this is a healthy compromise.

dandrabik (approved):
Yeah, sounds good, you can ship. I still think we want to make this higher to avoid spurious alerts, happy we can do that without a deploy.

PR #12864, opened 2025-03-27
slsa-with-multiple-playthroughs

anathomical:
## WHAT
Handle cases where a student somehow has multiple plays of a diagnostic by taking the most recent best performance from the student
## WHY
This should be very rare, but when it happens we want to handle it the same way we do on the Teacher Diagnostic Report: show the student's best performance.

One example of how this could happen is if a student has completed the pre and the post, the teacher may manually create a new Activity Pack from the Activity Library for the Post diagnostic and assign it.  In this case, it will be attached to the existing `StudentLearningSequence` that already has a pre and a post as a way of connecting the new post to a pre.
## HOW
Add some code to the serializer to filter for these values in the rare cases where the data is in this shape.

### Notion Card Links
https://www.notion.so/quill/Diagnostic-Insights-QA-March-10-2025-1b2d42e6f9418005b38cca6dc474422c?pvs=4#1bcd42e6f9418064aeddccfb30656f60

- Testing with staging teacher with id `19180770`
- There is a known student with multiple post-diagnostics as part of the same `StudentLearningSequence` (id: )
- Confirm that this student is marked "proficient" on the history page matching their best post-diagnostic performance
- Examine the student's detailed history page to confirm that the listed post-diagnostic in their sequence is their best performance

PR Checklist | Your Answer
------------ | -------------


services/QuillLMS/lib/snapshots/diagnostic_insights/history_by_classroom.json
anathomical, line 1: You can ignore this file.  These changes are based on a backfill that I'm testing in another branch, but I wanted to make sure we kept these snapshots up-to-date here.

emilia-friedberg (approved):
LGTM

brendanshean (approved):

PR #12867, opened 2025-03-28
backfill-fix-student-learning-sequence-created-at

anathomical:
## WHAT
Backfill fix student learning sequence created at
## WHY
We realized upon reflection that our original backfill logic used a non-optimal date when creating `StudentLearningSequence` and `StudentLearningSequenceActivity` records.  This code corrects that.
## HOW
- Update the backfill worker to use the new, better choice for source date in case we ever need to run backfill jobs in the future
- Write a job that can updates to the old records in batches

NOTE: Not captured in this PR, but we'll be paging these up to the point where we executed the original backfill in mid-November, not all possible records.  It'll still be the majority, but we won't be updating all of them.  The current plan is to update `StudentLearningSequence` records up to id `12_779_297` and `StudentLearningSequenceActivity` records up to id `356_097_408`

- Deployed to staging
- Enqueued jobs in `migration` queue
- Run through migrations
- Confirm that dates are updated

PR Checklist | Your Answer
------------ | -------------
Have you added and/or updated tests? | No tests on data backfilling


services/QuillLMS/app/workers/student_learning_sequences/backfill/fix_slsa_created_at_worker.rb
brendanshean, line 1: Had to do a double take comparing this worker name and the other one.  I think for one off, this is fine.

brendanshean (approved):
Thanks for cleaning up this data.

PR #12876, opened 2025-04-02
Feedback Moderation

brendanshean:
## WHAT
  A feedback moderation system that evaluates and improves LLM-generated feedback in the Evidence engine. The system identifies problematic feedback using a series of rule-based checkers and generates improved versions.
 
## WHY
The `FeedbackEvaluator` takes some LLM text and then runs some rules to check for feedback quality errors.  This PR leverages that tool and uses the error codes to direct a future LLM call to revise the original feedback.

## HOW
Add several new models (see [diagram](https://github.com/empirical-org/Empirical-Core/blob/5010d52d48aec22cbefe75afeafe629a1ade03b2/services/QuillLMS/engines/evidence/app/models/evidence/research/readme.md#6-feedback-moderation-system) for overall structure)

This system identifies and improves problematic LLM feedback through the following process:
  1. Detection: The `FeedbackEvaluator` service analyzes feedback and creates `FeedbackEvaluation` records with an `errored_checks` array listing all detected issues.
  2. Rules Creation: Each error code in `errored_checks` corresponds to a `FeedbackRule` that defines:
    - What the problem is (title)
    - Why it's a problem (goal)
    - How to fix it (rule)
  3. Revision Strategy: `FeedbackRevision` records group `FeedbackRules` to address multiple issues that frequently occur together:
    - Each revision has a title matching the combined error codes
    - Contains one or more `FeedbackRule`s through a join table `FeedbackRevisionRule`
    - Includes example pairs showing original and improved feedback
  4. Testing Improvements: The trial system tests how well a revision approach works:
    - `FeedbackRevisionTrial` finds `FeedbackEvaluation` records with matching error codes
    - Uses the `FeedbackRevision`'s rules and examples as a prompt for the `LLM`
    - Measures accuracy by checking if revised feedback resolves the original issues

  This creates a systematic approach to continuously improve AI feedback quality by identifying patterns of issues, defining clear rules for improvement, and measuring the effectiveness of revision strategies.

### Screenshots

### Notion Card Links
https://www.notion.so/quill/Reference-Guide-Rules-for-Feedback-Moderator-1a4d42e6f94180248b20c3e2a7e24447?pvs=4

PR Checklist | Your Answer
------------ | -------------


services/QuillLMS/engines/evidence/app/views/evidence/research/gen_ai/feedback_revision_trials/new.html.erb
```diff
@@ -0,0 +1,203 @@
...
+  <div class="actions">
+    <%= form.submit "Run Trial" %>
+  </div>
+<% end %>
+
+<div class="actions">
+  <%= link_to "Back to Feedback Revision", research_gen_ai_feedback_revision_path(@feedback_revision) %>
+</div>
+
+<style>
```
dandrabik, line 62: I'm on the fence on whether it would be better to extract these to css files. I get this is a playground for testing, but it also might get a little unwieldy/inconsistent to have individual styles for each page. Ok to leave for now though.

services/QuillLMS/engines/evidence/app/views/evidence/research/gen_ai/feedback_revision_trials/new.html.erb
```diff
@@ -0,0 +1,203 @@
...
+    background-color: #e0e0e0;
+  }
+
+  #copy-previous-prompt:disabled {
+    opacity: 0.6;
+    cursor: not-allowed;
+  }
+</style>
+
+<script>
```
dandrabik, line 164: Take or leave, you might consider a js.erb partial.

services/QuillLMS/engines/evidence/app/workers/evidence/research/gen_ai/build_feedback_revision_trial_llm_example_worker.rb
```diff
@@ -0,0 +1,44 @@
...
+    module GenAI
+      class BuildFeedbackRevisionTrialLLMExampleWorker
+        include Evidence.sidekiq_module
+
+        OPTIMAL = HasAssignedStatus::OPTIMAL
+        SUBOPTIMAL = HasAssignedStatus::SUBOPTIMAL
+
+        sidekiq_options queue: 'default'
+
+        attr_reader :trial, :revision, :raw_text, :llm_feedback, :feedback, :feedback_evaluation, :optimal
```
dandrabik: I don't think you need instance variables (drop the `@`) or attr_readers in this file, it's all one method.

services/QuillLMS/engines/evidence/app/controllers/evidence/research/gen_ai/feedback_rules_controller.rb
```diff
@@ -0,0 +1,49 @@
...
+        end
+
+        private def feedback_rule_params
+          params
+            .require(:research_gen_ai_feedback_rule)
+            .permit(:error_category, :error_key, :rule, :goal, :title)
+        end
+
+        private def build_checker_error_keys
+          {}.tap do |checker_error_keys|
```
dandrabik:
```ruby
Evidence::GenAI::Evaluation::FeedbackEvaluator::CHECKERS.to_h {|checker| [checker.error_category, checker.error_keys.map(&:to_s)]}

File: services/QuillLMS/engines/evidence/app/models/evidence/research/gen_ai/feedback_revision_trial.rb [hunk 5]
```diff
@@ -0,0 +1,73 @@
...
+
+        def error_codes = feedback_revision.title
+
+        def llm_examples = feedback_revision_trial_llm_examples.includes(:llm_example)
+
+        def set_trial_duration = update!(trial_duration: (Time.current - created_at).to_f)
+        def set_status = completed!
+        def set_results = set_accuracy
+
+        def example_fixed?(example)
```
dandrabik, line 62: I think this evaluation metric is fine for now (did it clean up the error(s) we asked it to, even if it didn't solve all errors OR it created other errors?), but when we get to trying to clean all errors, we'll probably want `result.errors.empty?`, since if we ask it to clean errors and it creates another, we shouldn't consider that a fix.

services/QuillLMS/engines/evidence/app/helpers/evidence/research/gen_ai/feedback_revision_trials_helper.rb
```diff
@@ -0,0 +1,59 @@
...
+      module FeedbackRevisionTrialsHelper
+        ADDITION_CLASS = 'addition'
+        DELETION_CLASS = 'deletion'
+        UNCHANGED_CLASS = 'unchanged'
+
+        def generate_diff_html(original, revised)
+          original = original.to_s
+          revised = revised.to_s
+
+          return '' if original.empty? && revised.empty?
```
dandrabik: Note, if you use `.blank?` here, you also catch multispace empty strings.

services/QuillLMS/engines/evidence/app/models/evidence/research/gen_ai/feedback_revision.rb
```diff
@@ -0,0 +1,38 @@
...
+        has_many :feedback_revision_examples, dependent: :destroy
+        has_many :feedback_revision_trials, dependent: :destroy
+
+        validates :title, presence: true, uniqueness: true
+
+        def self.generate_title_from_feedback_rules(feedback_rules) = feedback_rules.map(&:error_code).sort.join(', ')
+
+        def to_s = title
+
+        def prompt_text
```
dandrabik: Take or leave, consider making the static parts a template that you format with vars, e.g.
```ruby
TEMPLATE = <<-FEEDBACK
            <feedback_revision>
              <feedback_rules>%<rule_list>s</feedback_rules>
              <feedback_revision_examples>%<example_list></feedback_revision_examples>
            </feedback_revision>
          FEEDBACK

def prompt_text = format(TEMPLATE, rule_list:, example_list:)
def rule_list =
def example_list =

File: services/QuillLMS/engines/evidence/app/models/evidence/research/gen_ai/feedback_revision_example.rb [hunk 8]
```diff
@@ -0,0 +1,34 @@
...
+module Evidence
+  module Research
+    module GenAI
+      class FeedbackRevisionExample < ApplicationRecord
+        belongs_to :feedback_revision
+
+        validates :before_text, presence: true
+        validates :after_text, presence: true
+
+        def prompt_text
```
dandrabik: Take or leave, same here.

services/QuillLMS/engines/evidence/app/models/evidence/research/gen_ai/feedback_revision_trial.rb
```diff
@@ -0,0 +1,73 @@
...
+        has_many :feedback_evaluations, through: :feedback_revision_trial_llm_examples
+
+        validates :prompt, presence: true
+        validates :temperature, presence: true
+
+        store_accessor :results, :accuracy
+
+        def run = FeedbackRevisionTrialRunner.run(self)
+
+        def update_results!(updates)
```
dandrabik: Just noting that this has a similar race condition to [this](https://github.com/empirical-org/Empirical-Core/blob/develop/services/QuillLMS/engines/evidence/app/models/evidence/research/gen_ai/trial.rb#L113-L125). If multiple jobs will write to this, you may need to update this to that form.

services/QuillLMS/engines/evidence/app/models/evidence/research/gen_ai/feedback_rule.rb
```diff
@@ -0,0 +1,78 @@
...
+#  title          :text             not null
+#  created_at     :datetime         not null
+#  updated_at     :datetime         not null
+#
+# Indexes
+#
+#  index_feedback_rules_on_error_category_and_error_key  (error_category,error_key)
+#
+module Evidence
+  module Research
```
dandrabik, line 21: This might be worth keeping as is for simplicity sake for now, but some of this code, like this model, would also be used in the production path and activity settings. This might not belong in the `Research` module.

services/QuillLMS/engines/evidence/app/services/evidence/gen_ai/evaluation/feedback_evaluator.rb
```diff
@@ -8,7 +8,7 @@ class FeedbackEvaluator < ApplicationService
 
         VERSION = 2
 
-        CHECKS = [
+        CHECKERS = [
```
dandrabik, line 11: Good rename.

services/QuillLMS/engines/evidence/app/services/evidence/gen_ai/evaluation/feedback_evaluator.rb
```diff
@@ -27,7 +27,7 @@ def initialize(feedback:, entry:)
         def run
           errors = []
 
-          CHECKS.each do |check|
+          CHECKERS.each do |check|
```
dandrabik: maybe use `checker` as the local var here.

services/QuillLMS/engines/evidence/app/views/evidence/research/gen_ai/feedback_revisions/new.html.erb
```diff
@@ -0,0 +1,113 @@
...
+    function updateSubmitButton() {
+      const checkedBoxes = document.querySelectorAll('input[type="checkbox"]:checked');
+      submitButton.disabled = checkedBoxes.length === 0;
+    }
+
+    checkboxes.forEach(checkbox => {
+      checkbox.addEventListener('change', function() {
+        const card = this.closest('.card');
+        if (this.checked) {
+          card.style.borderColor = '#007bff';
```
dandrabik, line 93: Consider using a class and add/remove it.

services/QuillLMS/engines/evidence/app/views/evidence/research/gen_ai/feedback_rules/show.html.erb
```diff
@@ -0,0 +1,225 @@
...
+    border-color: #0366d6;
+  }
+
+  .toggle-btn:hover:not(.active) {
+    background-color: #e6e6e6;
+  }
+</style>
+
+<script>
+  function showBeforeAfterView() {
```
dandrabik, line 182: I believe this exact code is also in another file above.

dandrabik (approved):
Overall looks good. I left a bunch of comments that are worth looking at, but I believe most are take or leave.

I will note, because this is our beta/scrap/playground, I'm a little more ok with this be rough draft style code. We'll see how that goes, and may want to clean some of this up at some point.

services/QuillLMS/engines/evidence/app/workers/evidence/research/gen_ai/build_feedback_revision_trial_llm_example_worker.rb
brendanshean: Good point.

services/QuillLMS/engines/evidence/app/helpers/evidence/research/gen_ai/feedback_revision_trials_helper.rb
brendanshean: Good call

services/QuillLMS/engines/evidence/app/models/evidence/research/gen_ai/feedback_revision.rb
brendanshean: Great idea.  This is more performant too.

services/QuillLMS/engines/evidence/app/models/evidence/research/gen_ai/feedback_revision_trial.rb
brendanshean: I agree.  I debated adding another column for "new errors" since I feel that should be reported differently than if the original error is not fixed.  But, I wanted to cut scope since the "new error" feels like something we'd have to address with another step.

services/QuillLMS/engines/evidence/app/models/evidence/research/gen_ai/feedback_rule.rb
brendanshean: This is a good point and definitely something I've been wrestling with, specifically with some of the other models in 'Research'.  Let's discuss offline.

services/QuillLMS/engines/evidence/app/models/evidence/research/gen_ai/feedback_revision_trial.rb
brendanshean: Great point.  I forgot about this race condition.

PR #12884, opened 2025-04-03
dropdown input accessibility

emilia-friedberg:
## WHAT
Update `DropdownInput` to:

- make the container a `div` instead of a button, because screenreaders can't consistently find nested interactive elements
- ensure the dropdown input consistently points to the label (which has an `aria-label` if it's blank)
- accord with the behavior described by the `react-select` `aria-context`, which selects the focused option when the user tabs away and opens the menu with the down arrow keypress

Also updates `ItemDropdown` to handle the behavior correctly.

I tried to locate all external-user-facing dropdowns and make sure they had either a `label` or an `ariaLabel`, but there's a backup label just in case I missed some. We should be consistent about that going forwards.

Note: in testing this I discovered that our dropdowns that render custom components are not always accessible -- going to fix that in a follow-up PR.

## WHY
We want our DropdownInputs to be accessible on all screenreaders.

## HOW
See "What".

### Screenshots

### Notion Card Links
https://www.notion.so/quill/Teacher-facing-accessibility-audit-1b9d42e6f9418021acdcdc39cb0ced56?pvs=4

Tested all the dropdowns I could find to make sure they worked.

PR Checklist | Your Answer
------------ | -------------


eadams17 (approved):
love seeing this cleanup!

PR #12886, opened 2025-04-04
handle-skill-performance-backfill-error

anathomical:
## WHAT
Ignore questions we can't process when calculating skill_performance
## WHY
We need these jobs not to crash.  Additionally, while the data in these `ConceptResult` records is not fully processable, we still want to ensure that we compile as accurate a view of the student's skill performance as possible, even if there's some data that can't be used in that calculation.
## HOW
If a `ConceptResult` doesn't have a `question_uid` that we can tie it to (via a couple of different methods), then we exclude it from the calculations for skill performance.

- Using an `ActivitySession` id that we know fails in production, ensure that it processes with the new code in staging

PR Checklist | Your Answer
------------ | -------------


happythenewsad (approved):

dandrabik (approved):
LGTM.

PR #12887, opened 2025-04-04
skill-performance-backfill-worker

anathomical:
## WHAT
Add worker to handle paginated backfill management
## WHY
We need to work through ~100M backfill jobs, which is too many to all fit on the Sidekiq queue comfortably (we can fit ~2M at a time without issue).  This worker is intended to queue workers in acceptably sized blocks, then call itself once they've all been processed so that the queue keeps working until all intended jobs are completed.
## HOW
Enqueue groups of backfill jobs in specified blocks with code to re-enqueue this job for the next page at the end.

- Deploy to staging
- Run for smaller page size (10,000)
- Confirm jobs complete
- Confirm that self enqueue logic works
- Confirm that number of enqueued jobs doesn't grow significantly beyond page size

PR Checklist | Your Answer
------------ | -------------


services/QuillLMS/app/workers/backfill_skill_performance_worker.rb
```diff
@@ -0,0 +1,923 @@
...
+      .order(:completed_at).offset((page - 1) * page_size)
+      .limit(page_size)
+      .pluck(:id)
+  end
+
+  # This is:
+  # - All 12 pre diagnostics (old and new, non-ELL and ELL)
+  # - All 12 post diagnostics
+  # - Every activity connected to a Recommendation for those diagnostics
+  private def activity_ids
```
brendanshean, line 47: Normally, I would argue placing this in yaml file or somewhere else, but since this is backfill that will probably get removed shortly thereafter, probably fine as-is.

services/QuillLMS/app/workers/backfill_skill_performance_worker.rb
```diff
@@ -0,0 +1,923 @@
...
+        'args' => [id]
+      )
+    end
+
+    return if (page * page_size) >= MAX_RECORDS
+
+    # This should add a job to the end of the migration queue to enqueue
+    # the next page once all jobs enqueued by this worker are completed
+    # so that we don't have to babysit the process
+    BackfillSkillPerformanceWorker.perform_async(page + 1, page_size)
```
brendanshean, line 31: The recursive pattern involving sidekiq feels odd since unexpected failures will retry the batching worker, which I don't think we want.  In that case, I would use sidekiq batch and a callback with a recursive call to the next page.

```
def perform(page, page_size)
    batch = Sidekiq::Batch.new
    batch.on(:complete, 'BatchCallback#on_complete', { 'page' => page, 'page_size' => page_size })
    batch.jobs do
      activity_session_ids(page, page_size).each do |id|
        SaveActivitySessionSkillPerformanceWorker.perform_async(id)
      end
    end
  end


class BatchCallback
  def on_complete(status, options)
    page = options['page']
    page_size = options['page_size']
    if (page * page_size) < BackfillSkillPerformanceWorker::MAX_RECORDS
      BackfillSkillPerformanceWorker.perform_async(page + 1, page_size)
    end
  end
end
```

However, batch mode might be adding unnecessary complexity and ultimately require more monitoring if scattered failures happen in individual jobs and prevent the next page from being enqueued.  I think non batch mode might be faster in this case.

brendanshean (approved):
I think this is an interesting approach.  I will be curious to hear about how it manages the load.

dandrabik (approved):
I think this is fine enough for a temporary backfill.

PR #12891, opened 2025-04-07
Classlink role coercion, validation

happythenewsad:
## WHAT
1. ignores password property of incoming classlink users
2. coerces 'administrator' role to Quill's 'admin' role
3. reports error when UserPersister validations fail

## WHY
1. Not necessary, now that SSO is implemented 
2. 'administrator' is not a supported role in Quill
3. We want to gracefully fail when encountering invalid Classlink resources, so that the encompassing data sync can continue

### Screenshots

### Notion Card Links
https://www.notion.so/quill/Product-Board-30f97e4eb01246dbb8264253fb385073?p=1ced42e6f941803b9e5cf4bf689a847f&pm=s

(todo)
1. Run sync
2. Observe no unexpected NewRelic errors
3. Check that ClasslinkUser counts have not meaningfully changed

PR Checklist | Your Answer
------------ | -------------


services/QuillLMS/app/services/classlink_integration/user_persister.rb
```diff
@@ -12,6 +14,7 @@ def initialize(incoming_user:, existing_provider_users:, classlink_instance_id:)
 
     def run
       creatable_user_hash = UserPersister.creatable_user(incoming_user)
+      raise ValidationError unless self.class.valid?(creatable_user_hash)
```
brendanshean, line 19: The `valid?` code feels like something that should be asked of user.rb:

`def self.safe_role?(role) = SAFE_ROLES.include?(role)`

Then your you can get rid of the `self.valid?` and make the check directly:

`raise ValidationError unless User.safe_role?(creatable_user_hash[:role])`

services/QuillLMS/app/services/classlink_integration/users_fetcher.rb
```diff
@@ -5,19 +5,27 @@ class UsersFetcher < BaseFetcher
...
+    }.freeze
+
     def transform_api_response(user_hash)
       {
         account_type: 'classlink',
         email: user_hash['email'],
         name: "#{user_hash['givenName']} #{user_hash['middleName']} #{user_hash['familyName']}".squish,
-        password: generate_password(user_hash['familyName'], user_hash['givenName']), # TODO: remove after SSO linkage
-        role: user_hash['role'],
+        password: nil,
```
brendanshean, line 19: Do you need we still need the `generate_password` method?

services/QuillLMS/app/services/classlink_integration/users_fetcher.rb
```diff
@@ -5,19 +5,27 @@ class UsersFetcher < BaseFetcher
...
+        role: self.class.coerce_role(user_hash['role']),
```
brendanshean: This feel like a passthrough method.  What about using ROLES directly here:

        role: ROLES[user_hash['role'])

brendanshean (approved):
Just a couple of very minor changes requested, otherwise looks good.

services/QuillLMS/app/services/classlink_integration/user_persister.rb
happythenewsad: [offline discussion]

services/QuillLMS/app/services/classlink_integration/user_persister.rb
```diff
@@ -31,6 +36,12 @@ def run
       ErrorNotifier.report(e, { message: msg })
     end
 
+    def self.valid?(creatable_user_hash)
+      return false unless CLASSLINK_ROLES.include?(creatable_user_hash[:role])
+
+      true
+    end
```
anathomical, line 43: A purely stylistic take-or-leave comment, but I think this could be one bit of logic:
```ruby
def self.valid?(creatable_user_hash)
  CLASSLINK_ROLES.include?(create_user_hash[:role])
end
```
If you anticipate the need to add additional validation conditions, then I think leaving this as-is makes sense, but I figured I'd flag it.

anathomical (approved):

services/QuillLMS/app/services/classlink_integration/user_persister.rb
happythenewsad: Thanks for pointing that out - I'm planning to add other validations, so I mentally stubbed out the pattern in advance. I didn't tell you that though :-)

```
return false unless <clause>
return false unless <clause>
...
true
```

PR #12900, opened 2025-04-10
diagnostic-insights-benchmark

anathomical:
## WHAT
Add a benchmarking thor task for Diagnostic Insights
## WHY
We want to have a reliable way to understand how performant we can expect core Diagnostic Insights queries to be.  This will let us make more informed decisions about how much additional time and energy we want to spend on performance improvements.
## HOW
We maintain a list of 46 teachers that we think provide representative data for testing Diagnostic Insights with.  This script runs the key API call (`HistoryByClassroom`) for every classroom that every one of these teachers has, then does some aggregation math to understand performance.

- Run script, confirm no errors

PR Checklist | Your Answer
------------ | -------------
Have you added and/or updated tests? | No tests on thor scripts
Have you deployed to Staging? | No, there is no application code in this PR


services/QuillLMS/lib/benchmarks/diagnostic_insights/history_by_classroom.json
```diff
@@ -0,0 +1,275 @@
+{
+  "samples": 265,
+  "min_runtime": "84ms",
+  "mean_runtime": "1742ms",
+  "median_runtime": "702ms",
+  "ninety_fifth": "6475ms",
+  "max_runtime": "23925ms",
+  "data": {
```
anathomical, line 8: I don't know that recording the runtime for each individual classroom is particularly useful, but the data was already collected, so I figured I'd include it. That's all that's in the rest of this file, so doesn't need review.

happythenewsad (approved):

brendanshean (approved):

dandrabik (approved):
This will be super handy, good idea!

PR #12901, opened 2025-04-10
Evidence - Add llm.api_type and use same code for Trials and Check::GenAI

dandrabik:
## WHAT
This code add several features:
1. Add `llm.api_type` to allow use of `Completion` and `Chat` APIs
2. Make the interface of `Chat` and `Completion` APIs to be compatible
3. Update `Check::GenAI` to use the same path as TrialRunner
4. Allow `Chat` and `Completion` APIs to be used in Trials
5. For `Check::GenAI`, use the `API` and `model version` of the activated trial.
## WHY
We want the Trials and the student experience (`Check::GenAI`) to be the same for consistent results. We also don't know if one API type is better for our use case than the other, so we want both available for experiments.
## HOW
Add `llm.api_type` field and do the steps above.

During this, I also refactored `Check::GenAI` to use service classes (`PrimaryFeedback::ResponseBuilder` and `SecondaryFeedback::ResponseBuilder`) to make that class more manageable.

### Notion Card Links
https://www.notion.so/quill/Bug-Different-optimal-suboptimal-in-use-vs-TrialRunner-1c0d42e6f94180f6ae7fd82f99b7e6a0

https://www.notion.so/quill/Always-use-the-same-model-in-trial-runner-and-student-experience-1ced42e6f9418096817bf5b7a4b6ef8e

https://www.notion.so/quill/Make-Chat-and-Completion-APIs-interoperable-1d1d42e6f9418058840cc57e4841692c

I added a thor script to run sessions locally. I've deployed to staging and tested that feedback and trials work properly with a variety of models and api_types.

PR Checklist | Your Answer
------------ | -------------


services/QuillLMS/engines/evidence/app/models/evidence/research/gen_ai/llm.rb
```diff
@@ -43,13 +50,29 @@ def self.g_eval = find_by(vendor: GOOGLE, version: GEMINI_1_5_FLASH_LATEST)
...
+        def completion(prompt, temperature = DEFAULT_TEMPERATURE) = completion_client.run(prompt:, model:, temperature:)
 
         def google? = vendor == GOOGLE
         def open_ai? = vendor == OPEN_AI
 
+        def api_client = "Evidence::#{api_vendor_module}::#{api_class}".constantize
+
+        private def api_class = api_type.camelize
+        private def api_vendor_module = VENDOR_API_MAP.fetch(vendor) { raise UnsupportedVendorError }
+
```
dandrabik, line 62: This is the main method feedback runs through now for both the TrialRunner and Check::GenAI

services/QuillLMS/engines/evidence/lib/evidence/check/base.rb
```diff
@@ -24,7 +26,7 @@ def self.run(entry, prompt, previous_feedback, session_uid = nil)
         context = { entry:, prompt_id: prompt&.id, prompt_text: prompt&.text }
         Evidence.error_notifier.report(e, context)
```
dandrabik, line 28: This was an existing bug that I'm fixing. This is a class method, so assigning to @error wasn't assigning it to the object.

services/QuillLMS/engines/evidence/app/models/evidence/research/gen_ai/llm.rb
```diff
@@ -18,7 +19,7 @@ module GenAI
       class LLM < ApplicationRecord
         class UnsupportedVendorError < StandardError; end
 
-        DEFAULT_TEMPERATURE = 1.0
+        DEFAULT_TEMPERATURE = 0.0
```
brendanshean, line 22: Good catch here!  I have it set 0 in the UI, but that should be the same in model.

services/QuillLMS/engines/evidence/app/models/evidence/research/gen_ai/llm.rb
```diff
@@ -43,13 +50,29 @@ def self.g_eval = find_by(vendor: GOOGLE, version: GEMINI_1_5_FLASH_LATEST)
 
         def completion_client = VENDOR_COMPLETION_MAP.fetch(vendor) { raise UnsupportedVendorError }
 
-        def completion(prompt, temperature = DEFAULT_TEMPERATURE) = completion_client.run(prompt:, llm: self, temperature:)
+        def completion(prompt, temperature = DEFAULT_TEMPERATURE) = completion_client.run(prompt:, model:, temperature:)
 
         def google? = vendor == GOOGLE
         def open_ai? = vendor == OPEN_AI
 
+        def api_client = "Evidence::#{api_vendor_module}::#{api_class}".constantize
```
brendanshean: I think you could use `const_get` here:

`def api_client = Evidence.const_get(api_vendor_module).const_get(api_class)` which gets rid of the string interpolation and is probably a little faster than using `constantize`.

services/QuillLMS/engines/evidence/lib/evidence/open_ai/chat.rb
```diff
@@ -58,6 +52,8 @@ def cleaned_results
       def endpoint = ENDPOINT
       def response_format = RESPONSE_FORMAT
 
+      def self.chat_api? = false
```
brendanshean: Should this be true?

services/QuillLMS/engines/evidence/lib/evidence/open_ai/completion.rb
brendanshean, line 1: I like the refactoring of llm -> model.  Only pass in what you need.

services/QuillLMS/lib/tasks/evidence_session.thor
```diff
@@ -0,0 +1,34 @@
+# frozen_string_literal: true
+
+require_relative '../../config/environment'
+
+class EvidenceSession < Thor
+  # bundle exec thor evidence_session:run 506 'because' --entries 'they have public celebrations.' 'they have celebrations in public.'
+  desc 'run_session activity_id --entries', 'Run an feedback evalution trial'
```
brendanshean: typo with "Run an", should be "Run a"

brendanshean (approved):
Thanks so much for this much needed cleanup/standardization of the code.  I also really like the breaking out of response builder for primary/secondary feedback.

services/QuillLMS/engines/evidence/lib/evidence/open_ai/chat.rb
dandrabik: Oh, good catch.

services/QuillLMS/engines/evidence/app/models/evidence/research/gen_ai/llm.rb
dandrabik: Nice, I like that. Will update.

PR #12903, opened 2025-04-14
Classlink associator

happythenewsad:
## WHAT
- Use StudentClassroomAssociator in the Classlink Data Sync process

## WHY
- The Associator is the main way LMS persists user-classroom associations. It contains validations that were not present before this PR. 

## HOW
1. Call StudentClassroomAssociator in the EnrollmentPersister
2. Manage the order of persistence operations to allow for (1)

### Screenshots

### Notion Card Links
https://www.notion.so/quill/Product-Board-30f97e4eb01246dbb8264253fb385073?p=1c0d42e6f941800e8fa3c15b575fb7c1&pm=s

Run DistrictsImporter on staging; observe no unexpected errors

PR Checklist | Your Answer
------------ | -------------


services/QuillLMS/app/services/classlink_integration/district_importer.rb
```diff
@@ -62,7 +73,17 @@ def persist_records
         UserPersister.run(incoming_user:, existing_provider_users:, classlink_instance_id:)
       end
 
-      enrollments.each do |incoming_enrollment|
+      # these arrays must be refreshed, since the preceeding persisters may have created new records
```
happythenewsad: We could use the `users` and `classrooms` memoized instance variables instead of making another DB call here. 

I favor the current implementation because:
1. `users` is an array of hashes that have not yet gone through the `User` active record validations. We don't want to throw errors on new Enrollments that reference a user that doesn't exist due to failed validation.
2. Using `users` would make the the `EnrollmentPersister` parameter shape subtly different from the other persisters, which may be confusing.
3. `DistrictImporter` is a nightly sync, so the small performance hit of re-querying the DB seems acceptable.

brendanshean (approved):
This feels like a big piece of the sync puzzle.  Nice work!

services/QuillLMS/app/services/classlink_integration/district_importer.rb
```diff
@@ -40,6 +40,14 @@ def enrollments
...
 
+    def teacher_enrollments
+      @teacher_enrollments ||= enrollments.select { |enrollment| enrollment[:role] == 'teacher' }
+    end
+
+    def student_enrollments
+      @student_enrollments ||= enrollments.select { |enrollment| enrollment[:role] == 'student' }
+    end
+
     def run
```
dandrabik, line 13: Consider moving `run` method directly below the `initialize` method, and making the rest of the methods `private`. This makes the interface of the service object clear, i.e. only `run` and `new` should ever be called. This is especially important in a class like this in which order of method calls matters (you only want them run in the order in the `run` method.

dandrabik (approved):
Small comment about the structure of a service object for readability, but otherwise, looks good.

anathomical (approved):

PR #12912, opened 2025-04-17
Fix student reflection rendering

brendanshean:
## WHAT
Fix a bug involving the rendering of student reflection page.

## WHY
The `but` and `so` conjunction responses aren't rendering properly for some activities.

## HOW
The current setup assumes the prompt ids ordering mirrors because, but, so but this is not always the case.  Instead look for the first conjunction occurrence and label it that way.

### Screenshots

### Notion Card Links
https://www.notion.so/quill/Bug-Reflection-Page-Glitch-1d1d42e6f94180cb9808ccc55209acdb?pvs=4

I played through the offending activity on staging at it is now rendering properly.

PR Checklist | Your Answer
------------ | -------------


anathomical (approved):

services/QuillLMS/client/app/bundles/Evidence/components/activitySlides/postActivitySlide.tsx
```diff
@@ -14,6 +14,45 @@ interface PostActivitySlideProps {
...
+
+  for (const key of responsesKeys) {
+    const responseArray = responses[key];
+    if (!responseArray || !responseArray.length) continue;
+
+    const lastResponse = responseArray[responseArray.length - 1];
+    if (!lastResponse || !lastResponse.entry) continue;
+
+    const entry = lastResponse.entry;
+    const targetConjunctionRegex = new RegExp(`\\s${conjunction}\\s`);
```
emilia-friedberg: I'd pull this out of the loop because it's the same in every iteration.

services/QuillLMS/client/app/bundles/Evidence/components/activitySlides/postActivitySlide.tsx
```diff
@@ -14,6 +14,45 @@ interface PostActivitySlideProps {
   responses: any
 }
 
+// Helper function that can be exported for testing
+export function getResponsesForConjunction(responses: any, conjunction: string) {
+  const responsesKeys = Object.keys(responses);
+
+  for (const key of responsesKeys) {
```
emilia-friedberg: I prefer to use builtin iteration methods rather than `for` loops in general -- since you're not actually using the `key` here except to extract its values, you could turn this whole thing into:

`return Object.values(responses).find(responseArray => ...)`

Your `continue`s will have to become `return false`s, but I think otherwise the logic is the same.

services/QuillLMS/client/app/bundles/Evidence/components/activitySlides/postActivitySlide.tsx
```diff
@@ -14,6 +14,45 @@ interface PostActivitySlideProps {
...
+      const otherConjunctionRegex = new RegExp(`\\s${otherConjunction}\\s`);
+      const otherMatch = entry.match(otherConjunctionRegex);
+
+      if (otherMatch && otherMatch.index < conjunctionPosition) {
+        hasOtherConjunctionBefore = true;
+        break;
+      }
+    }
+
+    if (!hasOtherConjunctionBefore) { return responseArray; }
```
emilia-friedberg: These other conjunction regexes are also going to be the same in every iteration, so you could do something like:

```
const otherConjunctionRegexes = conjunctions.filter(conj => conj !== conjunction).map(otherConjunction => new RegExp(`\\s${otherConjunction}\\s`))
```

at the top of this function definition, and then inside your iterative loop after `conjunctionPosition` is defined, you can just do:

```
const hasOtherConjunctionBefore = otherConjunctionRegexes.find(otherConjunctionRegex => {
  const otherMatch = entry.match(otherConjunctionRegex)
  return otherMatch && otherMatch.index < conjunctionPosition
}
```

emilia-friedberg (approved):
Thank you for catching this and implementing this fix! I have some suggestions for cleaning up the new `getResponsesForConjunction` function, mostly just using built-in array methods instead of `for` loops, but your thorough testing should it make it easy to confirm that the behavior is consistent.

services/QuillLMS/client/app/bundles/Evidence/components/activitySlides/postActivitySlide.tsx
brendanshean: Much cleaner

services/QuillLMS/client/app/bundles/Evidence/components/activitySlides/postActivitySlide.tsx
brendanshean: Thanks.  Good idea on pulling these out of the loop.

PR #12915, opened 2025-04-21
diagnostic-insights-history-summary

anathomical:
## WHAT
- Extract practice data into a separate API instead of rolling it up with diagnostic info in Diagnostic Insights History pages
- Add the new service to our benchmarking thor task
## WHY
- Recommendation/practice data is an order of magnitude more plentiful than diagnostic play-throughs, but is a secondary concern when displaying data to teachers (the History page only shows practice data in pop-ups)
- We also don't need the same level of data for practice activities as we do for diagnostics, so pulling this into its own API that pulls less data speeds up the page on multiple fronts
## HOW
- Filter out non-diagnostic activities from the `HistoryByClassroom` service
- Create a new `PracticeByClassroom` service which fetches only Practice/Recommendation activities and provides a stripped-down set of aggregation limited to only what is needed to display data in the front-end
- Update the front-end to fetch the new `PracticeByClassroom` API endpoint and pass the data to relevant sub-components
- Update the logic that's used to display practice data to parse the new API shape and combine diagnostic and practice data where appropriate

### Notion Card Links
https://www.notion.so/quill/Diagnostic-Insights-QA-March-10-2025-1b2d42e6f9418005b38cca6dc474422c?pvs=4#1d5d42e6f9418096940cffab48c93939

- Deploy to staging
- Test, in conjunction with Peter S, that a set of known users "feel" fast enough
- Supplement this with analysis of "timing" in the API Network calls, confirming that the "main" API goes from ~10 seconds to ~3 seconds with our known test cases
- Also confirm that the new Practice API returns in a similar timeframe to ensure that the practice data isn't coming back slower in the new regime
- Confirm that pages still display the data they do in current `develop`, ensuring that the UI isn't interpreting the old and new APIs differently

PR Checklist | Your Answer
------------ | -------------


services/QuillLMS/lib/snapshots/diagnostic_insights/history_by_classroom.json
anathomical, line 1: There are significant changes to this file as we've stopped including Practice Recommendations in this API payload (they've been aggregated and moved to the `practice_by_classrooms`).  No need to do a deep-dive on this one, and the shape should remain stable moving forward.

emilia-friedberg (approved):
LGTM! Thanks for tackling this work from both the back- and front-end.

happythenewsad (approved):

PR #12916, opened 2025-04-21
Classlink data sync improvements

happythenewsad:
## WHAT / HOW
1. Throw custom error in DistrictsImporter for invalid districts, rather than string
2. Generate users' usernames in UserPersister instead of UsersFetcher 
3. Filter out irrelevant user roles in UsersFetcher
4. Use the rails' `validation_context` feature to allow Classlink users to bypass our standard password validation on the `User` model.

## WHY
1. ErrorNotifier requires a parameter of type Error, not string
2. It's redundant to generate usernames at fetch time for existing users.
3. It's good practice to move validation as early in a transaction as possible, so that you don't have to check for invalid data later on. This also saves a bit of network bandwdith from the Classlink API.
4. I use this pattern to get around a catch-22 in our business logic: `User.create` requires a password unless the user has a `provider`. `provider` is computed via existence of `ClasslinkAccount`, which foreign-key-references `User`. We could effect the same behavior by using `new_user.build_classlink_account`, but that requires changing `User has_many :classlink_accounts` to `User has_one :classlink_account`. Also doable, but it breaks the existing pattern of `User has_many :canvas_accounts`, etc.

### Screenshots

### Notion Card Links
https://www.notion.so/quill/Product-Board-30f97e4eb01246dbb8264253fb385073?p=1d8d42e6f94180cb8638f3009830ab63&pm=s

1. Run the main DistrictsImporter.run sync on staging; observe no errors

2. Integration test on staging, for the new Users filter:

```ruby
module ClasslinkIntegration 
  class TempFetcher < BaseFetcher 
    RESOURCE = 'users' 
    QUERY_STRING = "?filter=role='student' OR role='teacher' OR role='administrator'&sort=sourcedId&orderBy=asc'"

    def transform_api_response(user_hash) 
      {role: user_hash['role'], email: user_hash['email']}
    end
  end 
end

# These are test tokens, not prod tokens :-)
users = ClasslinkIntegration::TempFetcher.run(district_oneroster_id: 'JUzdk05jMjM%3D', 
district_bearer_token: 'ebe3cd0c-300a-4b3d-ac84-75a770e66f1e',district_id: 22527)
users.map {|x| x[:role]}.uniq
=> ["student", "teacher", "administrator", "parent", "guardian"] # PREVIOUSLY
=> ["student", "teacher", "administrator"] # WITH PR
```

PR Checklist | Your Answer
------------ | -------------


services/QuillLMS/app/models/user.rb
```diff
@@ -234,7 +234,8 @@ class User < ApplicationRecord
 
   validates :password,
     presence: { if: :requires_password? },
-    length: { maximum: CHAR_FIELD_MAX_LENGTH }
+    length: { maximum: CHAR_FIELD_MAX_LENGTH },
+    unless: -> { validation_context == :classlink_user }
```
anathomical: I'm not sure this thought is an improvement, but should we consider making `:classlink_user` a constant of some sort with the same thought as usual: it helps avoid typo-ing the value when used in a call?

anathomical (approved):

services/QuillLMS/app/services/classlink_integration/user_persister.rb
```diff
@@ -47,7 +51,11 @@ def self.valid?(creatable_user_hash)
     end
 
     def self.creatable_user(user_hash)
```
brendanshean: Both of these methods feel like one liners:

```
def self.creatable_user(user_hash) = user_hash.except(:sourced_id, :given_name, :family_name)
def self.updatable_user(user_hash) = creatable_user(user_hash).except(:username)
```

services/QuillLMS/app/models/user.rb
brendanshean: Haven't seen `validation_context` before  😎

brendanshean: I think it be instructive to add a test in user_spec.rb that shows this `validation_context` in action.

brendanshean (approved):
Looks good to me.  Nice solution to the password validation problem.

PR #12910, opened 2025-04-17
Fix corrupted active activity sessions

brendanshean:
## WHAT
Fix a bug involving ActiveActivitySession#data

## WHY
`Activity` model has an association to `Question` model through `uid`.  Questions are periodically removed from Activities, however the ActiveActivitySession data still contains stale references to these uids.  This causes session.js to trigger an error at:
```const denormalizedQuestion = JSON.parse(JSON.stringify(allQuestions[questionUid].question))```
 where it's using a questionUid from the data that no longer is attached to the Activity.

## HOW
~~When active_activity_session#show is called, make a call to clean_data where various keys: `questionSet`, `unansweredQuestions`, `currentQuestion` and `answeredQuestions` are inspected for these stale uids.  If they exist, we remove them.~~

Add an `after_update` callback to Activity that checks to see if the question uids in `data` have changed.  If they have, run `ActiveActivitySessionDataCleaner` on any existing `ActiveActivitySession` records that have been started and point to that activity (via the associated `ActivitySession`)

`ActiveActiveSessionDataCleaner`:  Inspect `questionSet`, `unansweredQuestions`, `currentQuestion` and `answeredQuestions` keys from the ActiveActivitySession#data.  If they exist, then remove them.

### Screenshots
(If applicable. Also, please censor any sensitive data)

### Notion Card Links
https://www.notion.so/quill/Student-sees-a-page-with-a-loading-circle-when-they-click-the-Resume-button-for-the-activity-Subord-1d7d42e6f94180748096f58ba559a633?pvs=4
https://www.notion.so/quill/A-student-sees-a-page-with-a-loading-circle-when-they-click-the-Resume-button-for-an-activity-1d7d42e6f941803a8cf7c734a14cbe08?pvs=4

### What have you done to QA this feature?
I tried the two failures from the support tickets on staging and they work now.

PR Checklist | Your Answer
------------ | -------------
Have you added and/or updated tests? |  YES
Have you deployed to Staging? |  YES
Self-Review: Have you done an initial self-review of the code below on Github? | YES

PR #12918, opened 2025-04-21
Use separate job for cleaning individual active_activity_sessions

brendanshean:
## WHAT
A previous [PR](https://github.com/empirical-org/Empirical-Core/pull/12910) added a worker that cleaned `ActiveActivitySession` records when an activity's questions are updated.  This PR moves each cleaning of an `ActiveActivitySession` to an individual worker.

## WHY
Currently the `CleanActiveActivitySessionsWorker` is cleaning all records which could potentially be tens of thousands of them.

## HOW
Create new worker that does the actual cleaning and provision it in the parent job.

### Screenshots

### Notion Card Links

Tested on staging.

PR Checklist | Your Answer
------------ | -------------


services/QuillLMS/spec/workers/clean_active_activity_session_worker_spec.rb
```diff
@@ -0,0 +1,35 @@
...
+
+require 'rails_helper'
+
+describe CleanActiveActivitySessionWorker, type: :worker do
+  subject { described_class.new }
+
+  let(:question) { create(:question) }
+  let(:activity) { create(:activity, data: { 'questions' => [{ 'key' => question.uid }] }) }
+  let(:activity_session) { create(:activity_session, :started, activity:) }
+  let(:active_activity_session) { create(:active_activity_session, activity_session:) }
```
anathomical: It's not wrong to set up all this data, but given the actual tests in this file, could we just get away with creating `active_activity_session` and skip the other models we're establishhing?

anathomical (approved):
Good call-out to split this into a separate worker to manage workload size.  One minor comment on pre-conditions in one of the test files, but this looks good to me.

services/QuillLMS/spec/workers/clean_active_activity_session_worker_spec.rb
brendanshean: Good point.  I will remove those.

PR #12919, opened 2025-04-22
qa updates for diagnostic insights: teacher verification, plus one for the recommendation cards

emilia-friedberg:
## WHAT
QA updates for the new Account Management page, plus one for the recommendation cards when an activity session has been archived, including:

- showing a tooltip and disabling the button when a student's diagnostic report would normally be viewable, but has been archived
- adding custom sort (via numbers before the string) to the account management table
- fixing alignment of text in bordered dropdowns (site-wide change)
- making sure tooltips are right-justified on initial load if they'll extend past the screen (site-wide change, just noticed while QAing this)
- automatically logging admins in when they follow the link to approve or deny a teacher verification request from their email
- fixing the role display when new users sign up as admins
- prevent multiple clicks on "Verify email" button

## WHY
These are all things Peter Sharkey found in the process of QAing.

## HOW
Mostly CSS/React changes, but I did update the backend to pass up data we need to confirm that an activity session's report is inaccessible, as well as adding the logic to sign in admins from the email links.

### Screenshots

### Notion Card Links
https://www.notion.so/quill/Diagnostic-Insights-Teacher-approval-process-improvements-1d6d42e6f94180f4b2e3d936d16235bd?pvs=4

Tested locally, having Peter Sharkey QA on staging

PR Checklist | Your Answer
------------ | -------------


services/QuillLMS/app/services/diagnostic_insights/diagnostics_assigned_by_classroom.rb
```diff
@@ -34,7 +34,9 @@ def run = grouped_results
           activity_id: slsa.activity_id,
           completed_at: slsa.completed_at,
           created_at: slsa.created_at,
-          user_id: slsa.student_learning_sequence.user_id
+          user_id: slsa.student_learning_sequence.user_id,
+          activity_session_id: slsa.activity_session_id,
+          student_learning_sequence_id: slsa.student_learning_sequence.id
```
anathomical, line 39: This is super-minor, and given how this data is pulled, probably doesn't actually matter, but there is a difference between `slsa.student_learning_sequence.id` and `slsa.student_learning_sequence_id`.  The first loads a `StudentLearningSequence` model and looks at the `id` column from the underlying table.  If that model hasn't already been loaded into memory, that will result in a database call.  The second loads the same value from the `StudentLearningSequenceActivity` table directly which lets you avoid any potential additional database loads.

In the actual case we have here, the `student_learning_sequence` data has been called with `eager_load`, so it's already been fetched from the database, which means it doesn't matter for practice purposes, but I wanted to call out the distinction anyway.

eadams17 (approved):

services/QuillLMS/client/app/bundles/Teacher/components/progress_reports/diagnostic_insights/history_and_recommendations/index.tsx
```diff
@@ -74,14 +74,14 @@ const HistoryAndRecommendations = ({ teacherClassroomIds, classroom, needsEmailV
   }, [location.pathname, navigate]);
 
   React.useEffect(() => {
-    if (byClass || selectedStudentId) { return }
+    if (byClass || selectedStudentId || !studentHistories) { return }
 
     const match = location.pathname.match(/by_student\/(\d+)$/);
     const studentId = match ? Number(match[1]) : null;
 
     if (studentId && studentHistories?.find(s => s.user_id === studentId)) {
```
anathomical: Not necessary, but given that we are now guaranteeing that `studentHistories` exists in the early return above, we could remove the `?.` safe accessor from this line as well as the one below.

services/QuillLMS/client/app/bundles/Teacher/components/progress_reports/diagnostic_insights/sidebar.tsx
```diff
@@ -36,9 +36,9 @@ const Sidebar = ({
   const activeClassName = ({ isActive }) => (isActive ? 'active focus-on-light' : 'focus-on-light');
 
   return (
-    <section className={`sidebar-container ${showMobileSidebar ? 'mobile-open' : 'mobile-hidden'}`} data-testid="sidebar" >
+    <aside className={`sidebar-container ${showMobileSidebar ? 'mobile-open' : 'mobile-hidden'}`} data-testid="sidebar" >
```
anathomical, line 39: I didn't know about the `aside` tag.  Neat!

services/QuillLMS/spec/controllers/admins_controller_spec.rb
```diff
@@ -284,6 +284,22 @@
...
       end
     end
+
+    context 'when token param is passed' do
+      let(:new_admin) { create(:admin) }
+
+      before do
+        new_admin.refresh_token!
+        # Spy on the sign_in method
+        allow(controller).to receive(:sign_in).and_call_original
```
anathomical: Given that you have an explicit `excpect` call below, I don't think you need this here.  You would need to move the `.and_call_original` from here to the end of line 298, though.  I could be wrong, in which case, just leave this as-is.

services/QuillLMS/spec/controllers/admins_controller_spec.rb
```diff
@@ -319,6 +335,22 @@
...
```
anathomical: Same note as above re moving `and_call_original` in the `expect` call below.

services/QuillLMS/spec/workers/teacher_requested_verification_email_worker_spec.rb
```diff
@@ -14,5 +14,19 @@
...
+
+      subject.perform(admin.id, teacher.id, message)
+    end
+
+    it 'should call refresh_token! on the admin if they do not already have a token' do
+      admin.refresh_token!
+
+      expect_any_instance_of(User).not_to receive(:refresh_token!)
+      subject.perform(admin.id, teacher.id, message)
+    end
```
anathomical: The linter may already be complaining about this, but `expect_any_instance_of` is generally frowned upon.  The "more approved" approach is something like this:
```ruby
before do
  allow(User).to receive(:find_by_id).and_return(admin, teacher)
end

it 'should call refresh_token! on the admin and enqueue a token expiration job if they do not already have a token' do
  expect(admin).to receive(:refresh_token!)
  expect(ExpirePasswordTokenWorker).to receive(:perform_in).with(30.days, admin.id)

  subject.perform(admin.id, teacher.id, message)
end
```
That `before` block should work for both `it` blocks.

anathomical (approved):
A couple of small take-or-leave suggestions, but overall this looks good to me.

services/QuillLMS/spec/workers/teacher_requested_verification_email_worker_spec.rb
emilia-friedberg: This is good to know, thanks! I didn't like the `expect_any_instance_of` here but was having a hard time figuring out a way around it.

PR #12920, opened 2025-04-22
Fix stripe subscription bug with past invoice subscription

brendanshean:
## WHAT
Fix a bug where users are hitting an error page when attempting to renew a subscription.

## WHY
A group of users had premium subscriptions that were signed up for by another individual that doesn't exist in our system but was processed via invoice.

Those same users are now attempting to renew their subscription but using their own email

## HOW
Our current code populates the customer_email via `subscriptionStatus.customer_email` which might be nil if the subscription was initiated via an invoice.

When renewing we do want to try the `subscriptionStatus.customer_email` but if it doesn't exist, the user's actual email should be allowable.  (The reason   `subscriptionStatus.customer_email`  takes precedence is some users initially sign up for the subscription via a different email)

### Screenshots

### Notion Card Links
https://www.notion.so/quill/Four-teachers-see-Something-went-wrong-message-when-attempting-to-purchase-Quill-Teacher-Premium-1ddd42e6f94180659dfbebb6bbe5a043?pvs=4

I tested this out on staging and the user can now proceed with the flow to the Stripe payment page.

PR Checklist | Your Answer
------------ | -------------


services/QuillLMS/app/models/user.rb
brendanshean, line 1: Adding an extra check here.  Having this initially would have made diagnoising the bug much easier.

anathomical (approved):
Nice find.

PR #12923, opened 2025-04-23
districts allow list

happythenewsad:
## WHAT
Add an allowlist to the District Sync

## WHY
Our Classlink contact has been adding test servers into our prod environment. I did not anticipate this at the start of the project. Now; I want an extra layer of safety, so we don't auto-ingest unexpected data into prod.

## HOW
Use an env var allowlist 

### Screenshots

### Notion Card Links
none

run the sync script on staging with an empty env var

PR Checklist | Your Answer
------------ | -------------


anathomical (approved):
Controlling this via ENV variable seems reasonable.  Are we concerned about this potentially getting very large, or is it just one ID for production Classlink so that this will mostly just be one value?

happythenewsad:
> Controlling this via ENV variable seems reasonable. Are we concerned about this potentially getting very large, or is it just one ID for production Classlink so that this will mostly just be one value?

Forgot to respond to this: Yes, it will grow, one per district. @anathomical

PR #12924, opened 2025-04-23
expand-diagnostic-insights-benchmarks

anathomical:
## WHAT
- Fix a `where` order issue that was causing an inefficient query construction
- Fix an issue that arises in rare cases of weird data discrepancy
- Expand diagnostic insights benchmarks
## WHY
- The mis-ordering of conditions was resulting in an overly broad WHERE condition which was taking longer to run than necessary
- I'm not sure how this happens, but apparently sometimes there are cases where a `StudentLearningSequenceActivity` refers to a "completed" `ActivitySession` that is not actually completed, and we want to ignore those records for reporting
- We want to understand how fast our APIs run on a standard set of classroom IDs to track performance changes over time
## HOW
- Move the `or` condition to be adjacent to the `where` clause we want to combine it with directly, and move the condition that we want to apply as a broad `and` condition to the end of the `where` chain
- Use `select` to remove `ActivitySession` records that aren't actually completed
- Refactor the thor tasks used for benchmarking to cover all of our Diagnostic Insights data services

- Run snapshot generation to confirm that things run as quickly as expected
- Also confirm that none of our test classrooms crash on any of the services

PR Checklist | Your Answer
------------ | -------------


happythenewsad (approved):

dandrabik (approved):
thanks for putting this together. These baseline numbers will be helpful.

PR #12926, opened 2025-04-24
Classlink integration test importer; update env var schema

happythenewsad:
## WHAT
- creates temporary Classlink Users importer 
- the district allow list should support comma separated strings 

## WHY
- Classlink prod integration requires importing test data into production. The temp importer maximally limits the imported data, so that set down is easier after testing has completed. 
- we want to support an array of allowed districts, since long term we expect dozens of Classlink districts

### Screenshots

### Notion Card Links
none

Run the TempUsersImporter script on staging. Observed only Users are created

PR Checklist | Your Answer
------------ | -------------


anathomical (approved):
My understanding is that we'll be removing most/all of this code once testing is done.  If so, this all looks good to me.  If we were going to keep this around long-term, I'd be concerned about the possibility of the ENV variable growing to include hundreds or thousands of values over time, but it sounds like that's not a risk here.

brendanshean (approved):

PR #12930, opened 2025-04-24
Feedback moderation integration

brendanshean:
## WHAT
* Introduces feedback‑moderation as a first‑class step in the GenAI check pipeline.
* New domain objects: `FeedbackRevision`, `FeedbackRevisionRule`, plus factories, specs and migrations for unique titles and rule keys.
* Adds revised_feedback prompt/response builders and swaps `ResponseBuilder` to consume a revised_response rather than the raw primary response. 
* Adds `FeedbackRevisionTrialRunner` to bulk‑rewrite feedback and surfaces the revised text alongside originals in the trials UI. 

## WHY
* The primary LLM occasionally returns feedback that is off‑rubric or unsafe. Moderating every answer by hand doesn’t scale.
* Codifying moderation logic as rules + examples makes revisions reproducible, searchable and easy to tune, while the “active” flag lets us retire rules without data loss.
* Unique DB constraints prevent duplicate rules/revisions and guard against race conditions in async batch creation. 

## HOW
1.  Error detection – `Evidence::Check::GenAI` still flags error codes on the primary response.
2.  Revision lookup/build – `FeedbackRevision.find_or_create_by_error_codes` either fetches an existing revision or composes a new one from the relevant active FeedbackRules, then persists it inside a transaction. 
3.  Prompt assembly – `RevisedFeedback::PromptBuilder` injects the rule + example XML into the revision template and returns nil when no revision is needed, short‑circuiting the extra API call.
4.  Second pass to the LLM – `RevisedFeedback::ResponseBuilder` calls LLM only when the primary response is non‑optimal, producing a revised_response object.
5.  Final packaging – `ResponseBuilder` now builds the API payload, highlight array, and optimal flag from the revised response, falling back to secondary feedback when present. 
6. Persistence & visibility – migrations add active:boolean to rules and unique multi‑column indexes; trials UI shows both raw and moderated feedback so reviewers can audit deltas. 

### Screenshots

### Notion Card Links
https://www.notion.so/quill/Add-the-Feedback-Moderator-to-Production-1ced42e6f94180cd9b6bd62dc936987c?pvs=4

I have tested this on staging for the Trial runner.

PR Checklist | Your Answer
------------ | -------------


services/QuillLMS/engines/evidence/app/models/evidence/research/gen_ai/feedback_revision.rb
```diff
@@ -9,14 +9,22 @@
...
+#
 module Evidence
   module Research
     module GenAI
       class FeedbackRevision < ApplicationRecord
         TEMPLATE = <<-FEEDBACK
           <feedback_revision>
-            <feedback_rules>%<feedback_rules_text>s</feedback_rules_text>
-            <feedback_revision_examples>%<feedback_revision_examples_text>s</feedback_revision_examples>
+            <feedback_rules>
```
dandrabik, line 22: A bit of an aside, but if we are getting good results from this `xml` style prompt, I'd be interested to try that for other prompts/uses.

services/QuillLMS/engines/evidence/app/services/evidence/gen_ai/revised_feedback/response_builder.rb
```diff
@@ -0,0 +1,38 @@
+# frozen_string_literal: true
+
+module Evidence
+  module GenAI
+    module RevisedFeedback
+      class ResponseBuilder < Evidence::ApplicationService
+        KEY_FEEDBACK = 'feedback'
+        KEY_ENTRY = 'optimal'
```
dandrabik: If you end of changing the other one, let's change this one to match.

services/QuillLMS/engines/evidence/app/services/evidence/gen_ai/revised_feedback/response_builder.rb
```diff
@@ -0,0 +1,38 @@
...
+          return primary_response if primary_response.optimal
+          return primary_response if llm_prompt.nil?
+
+          Response.new(optimal:, feedback:, llm_id:)
+        end
+
+        private def feedback = response[KEY_FEEDBACK]
+        private def optimal = response[KEY_ENTRY]
+
+        private def response = @response ||= JSON.parse(response_json)
```
dandrabik, line 29: This can probably be a separate PR (or fix down the road), but since we aren't doing the structured output, we probably should integrate `LLMFeedbackResolver` (or some version of it. I need to add it to some of my paths as well, siince I'm also still doing some regular JSON parsing.

dandrabik (approved):
This looks good, thanks for pushing this forward! 🚢

services/QuillLMS/engines/evidence/app/services/evidence/gen_ai/revised_feedback/response_builder.rb
brendanshean: I forgot about this.  Thank you for the reminder.

PR #12936, opened 2025-04-28
assign-diagnostics-without-assign-on-join

anathomical:
## WHAT
- Respect explicitly set `assign_on_join: false` when creating or updating `ClassroomUnit` records even in cases where every member of the classroom is assigned to the `ClassroomUnit`
- Tweak the Diagnostic Insights Recommendations tab so that it doesn't update diagnostic assignment/open state unless at least one student is being assigned a recommendation
## WHY
- If a teacher assigns all current students in their classroom the recommendation of a single diagnostic, and a new student joins the classroom, they should not be automatically assigned to the diagnostic but should require an explicit assignment
- When the Recommendation assignment workflow is used, we should avoid surprising a teacher by potentially re-opening previously-closed diagnostics of a different level than the one being assigned to
## HOW
- Add some logic to the `ClassroomUnit` logic that bypasses the `assign_on_join` logic in `ClassroomUnit.check_for_assign_on_join_and_update_students_array_if_true` if `assign_on_join` is explicitly set in a `create` or `update` call
- Filter the payload submitted to the assignment API to only include diagnostics with students assigned

### Screenshots

### Notion Card Links

PR Checklist | Your Answer
------------ | -------------


services/QuillLMS/spec/models/classroom_unit_spec.rb
```diff
@@ -142,6 +142,40 @@
           expect(other_classroom_unit.reload.assign_on_join).not_to be(true)
         end
       end
+
+      context 'when assign_on_join is explicitly set to false' do
+        it 'respsects the explicitly set value in create' do
```
brendanshean: typo: `respsects` on line 147, 153, 159, 165

services/QuillLMS/client/app/bundles/Teacher/components/progress_reports/diagnostic_insights/history_and_recommendations/recommendations.tsx
```diff
@@ -139,9 +139,11 @@ const Recommendations = ({ classroom, getPreviouslyAssignedDiagnosticDataForClas
   function assignDiagnostics() {
     setAssigning(true)
 
+    const selectedAssignmentsWithStudents = Object.values(selectedAssignments).filter(diagnostic => diagnostic.student_ids.length > 0)
```
emilia-friedberg, line 142: Good catch.

emilia-friedberg (approved):
LGTM! Thanks for this fix.

services/QuillLMS/spec/models/classroom_unit_spec.rb
```diff
@@ -142,6 +142,40 @@
           expect(other_classroom_unit.reload.assign_on_join).not_to be(true)
         end
       end
+
+      context 'when assign_on_join is explicitly set to false' do
```
brendanshean, line 154: I think you pull out:
```
let(:assigned_student_ids) { [student.id, student2.id] }
let(:assign_on_join) { false }
```
which will make clear what's really varying test to test.

brendanshean (commented):
What about instead of attr_accessor and overrides of create/update, leverage dirty-tracking and replace 
`before_save :check_for_assign_on_join_and_update_students_array_if_true` with two callbacks:

```
before_validation :default_assign_on_join, on: %i[create update]
before_save       :ensure_assigned_student_ids_if_true

private

def default_assign_on_join
  return unless will_save_change_to_assign_on_join? == false

  self.assign_on_join = classroom.assign_students_on_join
end

def ensure_assigned_student_ids_if_true
  return unless assign_on_join
  self.assigned_student_ids = StudentsClassrooms.where(classroom_id:).pluck(:student_id)
end
```

brendanshean (approved):
Nice work on figuring out the assign_attributes approach.  Much cleaner.

services/QuillLMS/spec/models/classroom_unit_spec.rb
anathomical: This is actually probably a good call, but would involve rewriting most of the tests in the containing context block to use `expect.to change` patterns instead of the current `expect(pre_condition), update, expect(post_condition)`.

This may be worth doing, but I wanted to avoid touching working specs here.  Thoughts on whether it's worth it to overhaul the section to be more in line with our more state-of-the-art spec approach?

brendanshean: You're right.  My mistake.

PR #12940, opened 2025-04-29
Diagnostic insights practice api

anathomical:
## WHAT
- Create a new `SkillPracticeByClassroom` service to generate data for the Practice Insights section of the Diagnostic Insights tool
- Add a controller endpoint that allows the front-end to access this data using the same pattern as we use for `SkillPerformanceByClassroom`
- Move shared constant declarations across the various services into a shared lib
- Add a thor task to benchmark the new service to get a baseline of expected runtime
## WHY
- We have another new data shape that is optimal for displaying the data this report requires
- We want this data to be easily accessible from the front-end
- This is work to DRY up the code, avoiding defining the same constants in multiple files when they have the same values
- We're getting better at keeping track of how our big data queries are at returning data in reasonable timeframes, and this is a key tool in understanding how we're doing
## HOW
- Following a pattern similar to `SkillPerformanceByClassroom`, create a new service to fetch and shape data
- Use existing patterns for the new endpoint
- Create a new `app/lib/diagnostic_insights/constants.rb` file to keep our shared constants for this module in
- Add a new thor task modeled on the one we used for `SkillPerformanceByClassroom`

### Notion Card Links
https://www.notion.so/quill/Diagnostic-Insights-Practice-Insights-API-1d2d42e6f9418064b3c9ffc1968f6ff7

- Use the Rails console to generate an API response for a known classroom (id: 1670362)
- Look at the Teacher Diagnostic Report for that classroom and confirm that activity counts reported there match counts from new service
- - Looking at student (16962122), confirm that they have been assigned all Pre recommendations and some of the Post recommendations on the Teacher Diagnostic Report (41 total activities)
- - Note that the new service reports the student count for specific skills as higher than the total number of assigned recommendations
- - Dig into the data and confirm that the discrepancy is accounted for by replays of practice activities as indicated in the database

PR Checklist | Your Answer
------------ | -------------


services/QuillLMS/app/services/diagnostic_insights/diagnostics_assigned_by_classroom.rb
```diff
@@ -4,20 +4,13 @@ module DiagnosticInsights
...
     attr_reader :classroom_id
 
-    STARTER_DIAGNOSTIC = :starter
-    INTERMEDIATE_DIAGNOSTIC = :intermediate
-    ADVANCED_DIAGNOSTIC = :advanced
-    ELL_STARTER_DIAGNOSTIC = :ell_starter
-    ELL_INTERMEDIATE_DIAGNOSTIC = :ell_intermediate
-    ELL_ADVANCED_DIAGNOSTIC = :ell_advanced
-
     DIAGNOSTIC_TYPE_MAPPING = {
```
anathomical, line 7: While this has the same name as the value that was moved to `Constants`, the version in this service has different values (Pre diagnostic IDs only, not Pre + Post)

emilia-friedberg (approved):
LGTM!

services/QuillLMS/app/lib/diagnostic_insights/constants.rb
brendanshean, line 1: A pattern I've used in engines/evidence/app/models/evidence/research/gen_ai.rb is to use the top level file as a store of constants.  This removes the need of having to preface `Constants::`

For this move the file to:
services/QuillLMS/app/lib/diagnostic_insights.rb

and have:

```
module DiagnosticInsights
  DEFAULT_START_DATE = School.school_year_start(DateTime.new(2023, 8, 1))

  STARTER_DIAGNOSTIC = 'starter'
  ...
end
```

services/QuillLMS/app/services/diagnostic_insights/skill_practice_by_classroom.rb
```diff
@@ -0,0 +1,68 @@
...
+
+module DiagnosticInsights
+  class SkillPracticeByClassroom < ApplicationService
+    class UnknownDiagnosticType < StandardError; end
+
+    attr_reader :classroom_id, :diagnostic_type, :start_date, :end_date
+
+    def initialize(classroom_id, diagnostic_type, start_date: nil, end_date: nil)
+      @classroom_id = classroom_id
+      @diagnostic_type = validate_diagnostic_type(diagnostic_type)
```
brendanshean: For service objects, my current preference is to keep `initialize` lean and only the assigning of values.  I put validations in `run` so that overall control flow is specified in one place. Claude and ChatGPT were split on this question so there might not be a correct answer here, but I think it's important to keep this pattern consistent.

```
def run
  raise UnknownDiagnosticType if unknown_diagnostic_type?
  
  skills_practice_by_classroom
end

private def skills_practice_by_classroom
  students.map do |student|
    {
      user_id: student.id,
      user_name: student.name,
      skills: format_skills_for_student(student)
    }
  end
end
```

services/QuillLMS/app/controllers/diagnostic_insights_controller.rb
```diff
@@ -43,6 +47,7 @@ def complete_history
   private def fetch_history_by_classroom = DiagnosticInsights::HistoryByClassroom.run(classroom_id, ell:, start_date:, end_date:)
   private def fetch_practice_by_classroom = DiagnosticInsights::PracticeByClassroom.run(classroom_id, start_date:, end_date:)
   private def fetch_skill_performance_by_classroom = DiagnosticInsights::SkillPerformanceByClassroom.run(classroom_id, diagnostic_type, start_date:, end_date:)
+  private def fetch_skill_practice_by_classroom = DiagnosticInsights::SkillPracticeByClassroom.run(classroom_id, diagnostic_type, start_date:, end_date:)
 
   private def authorize_request
     return if current_user.classrooms_i_teach.pluck(:id).include?(classroom_id)
```
brendanshean: I think `exists?` might be more performant here:
`return if current_user.classrooms_i_teach.exists?(id: classroom_id)`

services/QuillLMS/app/services/diagnostic_insights/skill_practice_by_classroom.rb
```diff
@@ -0,0 +1,68 @@
...
+      end
+    end
+
+    private def format_skills_for_student(student)
+      student_activity_sessions = activity_sessions_by_student[student.id]
+
+      return [] unless student_activity_sessions
+
+      student_activity_sessions
+        .group_by { |activity_session| activity_session.skill_performance.first['skill_name'] }
```
brendanshean, line 39: Do we need safe navigation here?
`activity_session.skill_performance.first&.fetch('skill_name')`

brendanshean (approved):
I left a question about placement of validation, but otherwise looks good.

services/QuillLMS/app/controllers/diagnostic_insights_controller.rb
anathomical: Your suggested change doesn't work because `classrooms_i_teach` is no longer a query object, but a collection (so `exists?` can't be used with it), but you are right that there's a more efficient approach, which I've implemented: `return if ClassroomsTeacher.exists?(user: current_user, classroom_id:)`

brendanshean: Sorry, you're right.  I forgot `classrooms_i_teach` is that `find_by_sql` type.

services/QuillLMS/app/services/diagnostic_insights/skill_performance_by_classroom.rb
```diff
@@ -29,12 +13,14 @@ class UnknownDiagnosticType < StandardError; end
 
     def initialize(classroom_id, diagnostic_type, start_date: nil, end_date: nil)
       @classroom_id = classroom_id
-      @diagnostic_type = validate_diagnostic_type(diagnostic_type)
-      @start_date = start_date
-      @end_date = end_date
+      @diagnostic_type = diagnostic_type
+      @start_date = start_date || DEFAULT_START_DATE
```
dandrabik: If there's not special logic, you can move these to params, e.g.
```ruby
def initialize(classroom_id, diagnostic_type, start_date: DEFAULT_START_DATE, end_date: Float::INFINITY)

File: services/QuillLMS/app/services/diagnostic_insights/skill_practice_by_classroom.rb [hunk 6]
```diff
@@ -0,0 +1,74 @@
+# frozen_string_literal: true
+
+module DiagnosticInsights
+  class SkillPracticeByClassroom < ApplicationService
+    class UnknownDiagnosticType < StandardError; end
+
+    attr_reader :classroom_id, :diagnostic_type, :start_date, :end_date
+
+    def initialize(classroom_id, diagnostic_type, start_date: nil, end_date: nil)
```
dandrabik: same comment as above.

services/QuillLMS/app/services/diagnostic_insights/skill_practice_by_classroom.rb
```diff
@@ -0,0 +1,74 @@
...
+        last_practiced: activity_sessions.map(&:completed_at).max,
+        percentages: activity_sessions.map(&:percentage)
+      }
+    end
+
+    private def students
+      @students ||= User
+        .joins(students_classrooms: :classroom_unscoped)
+        .where(students_classrooms: { classrooms: { id: classroom_id } })
+        .ordered_by_last_name
```
dandrabik, line 59: If you are ultimately putting these in a hash, do they need to be ordered?

services/QuillLMS/app/services/diagnostic_insights/skill_practice_by_classroom.rb
```diff
@@ -0,0 +1,74 @@
...
+        timespent: activity_sessions.map(&:timespent).sum,
+        last_practiced: activity_sessions.map(&:completed_at).max,
+        percentages: activity_sessions.map(&:percentage)
+      }
+    end
+
+    private def students
+      @students ||= User
+        .joins(students_classrooms: :classroom_unscoped)
+        .where(students_classrooms: { classrooms: { id: classroom_id } })
```
dandrabik: You can save a join here.
```suggestion
        .joins(:students_classrooms)
        .where(students_classrooms: { classroom_id: } )
```

dandrabik (approved):
Overall looks good, small changes suggested that you can take or leave.

services/QuillLMS/app/services/diagnostic_insights/skill_practice_by_classroom.rb
anathomical: We're actually returning an array of hashes, with one hash per student.

PR #12941, opened 2025-04-30
Classlink email merge

happythenewsad:
## WHAT
- if an incoming Classlink classroom matches the NCES id of an existing Quill class, don't create an additional Quill class
- if an incoming Classlink user matches the email of an existing Quill user, don't create an additional Quill user 
- refactor User Persister 

## WHY
- project spec (we assume existing users want to keep their Quill accounts and classrooms)
- same 
- it was getting hard to read 

## HOW
- Update SchoolsFetcher to include NCES id 
- update UserPersister with new conditional 
- make `run` a 3-way conditional and push other logic down into subroutines. 

### Screenshots

### Notion Card Links
https://www.notion.so/quill/Product-Board-30f97e4eb01246dbb8264253fb385073?p=1d0d42e6f9418023bf5ffcc73d732365&pm=s

(todo)
1. Run district sync
2. observe no changes to sample teacher's (Argimiro Alta) classroom and student relationships, via the Manage Classrooms component
(before)
Screenshot 2025-04-30 at 2 11 10 PM

(after)
Screenshot 2025-04-30 at 2 13 47 PM

3. For [schools,users] 
a) drop 1 record, instantiate new record with email/NCES known to exist in Classlink server
b) run district sync, and 
c) observe that no new records are added (The user's ClasslinkAccount should be added, though)

PR Checklist | Your Answer
------------ | -------------


services/QuillLMS/app/services/classlink_integration/schools_fetcher.rb
```diff
@@ -3,15 +3,18 @@
...
     def transform_api_response(school_hash)
       {
         district_id:,
-        name: school_hash['name']
+        name: school_hash['name'],
+        nces_id: self.class.coerce_nces_id(school_hash['identifier'])
       }
     end
+
+    def self.coerce_nces_id(maybe_nces_id) = maybe_nces_id.to_s =~ /\A\d{12}\z/ ? maybe_nces_id : nil
```
happythenewsad, line 18: It's possible for Districts to use this field for non-NCES values, so validation is warranted

services/QuillLMS/app/services/classlink_integration/school_persister.rb
```diff
@@ -0,0 +1,24 @@
...
+    def run
+      if nces_id
+        School.find_or_create_by!(nces_id:) do |school|
+          school.name = name
+          school.district_id = district_id
+        end
+      else
+        School.find_or_create_by!(name:, district_id:)
+      end
+    end
```
anathomical, line 22: There may be plans to extend this later, in which case my suggestion won't make sense, but I prefer early returns over `if ... else` blocks.  Take or leave this, but I'd probably write it as:
```ruby
def run
  return School.find_or_create_by!(name:, district_id:) unless nces_id

  School.find_or_create_by!(nces_id:) do |school|
     school.name = name
     school.district_id = district_id
   end
end
```

services/QuillLMS/app/services/classlink_integration/schools_fetcher.rb
anathomical, line 18: Another thing worth noting here is that we've run into issues in the past with some databases left-padding `0`s in NCES IDs when appropriate, and some not such that `000000000001` being recorded as `1`.  Hard to know if that'll happen in this database, but worth being aware of if we're going to validate data like this looking for a specific number of characters.

services/QuillLMS/spec/services/classlink_integration/school_persister_spec.rb
```diff
@@ -0,0 +1,65 @@
...
+        expect {
+          subject.run
+        }.not_to change(School, :count)
+
+        expect(school.reload).to have_attributes(
+          name:,
+          district_id: district.id,
+          nces_id:
+        )
+      end
```
anathomical, line 38: Is it also worth having a test case for the "update `name` and `district_id` if `nces_id` match is found` logic?

services/QuillLMS/spec/services/classlink_integration/user_persister_spec.rb
```diff
@@ -48,6 +48,17 @@
...
+    context 'when non classlink user already exists' do
+      let(:existing_provider_users) { [] }
+      let!(:existing_non_classlink_user) { create(:user, role: 'student', email: incoming_user[:email]) }
+
+      it 'creates new Account record' do
+        expect { subject.run }
+          .to change(ClasslinkAccount, :count).by(1)
+          .and not_change(User, :count)
+      end
+    end
```
anathomical: Again, is it worth writing a test case for update behavior when a match is found by email, but (for instance) name in Classlink does not match name in LMS database?

services/QuillLMS/app/services/classlink_integration/user_persister.rb
```diff
@@ -17,27 +17,51 @@ def initialize(incoming_user:, existing_provider_users:, classlink_instance_id:)
...
-            user:
-          )
-        end
+        create_new_user_with_classlink
       end
     rescue => e
-      existing_or_new = existing_user ? 'existing' : 'new'
-      msg = "Failed to persist #{existing_or_new} classlink user with sourcedId=#{incoming_user[:sourced_id]}"
-      ErrorNotifier.report(e, { message: msg })
+      handle_error(e)
```
anathomical, line 28: Again, this is purely stylistic so take or leave it, but I prefer early return patterns over `if ... else` blocks
```ruby
def run
  return update_existing_classlink_user if existing_classlink_user
  return link_existing_user_to_classroom if existing_non_classlink_user

  create_new_user_with_classlink
rescue => e
  ...
end
```

anathomical (approved):
This looks good to me.  A couple of style notes, but I think this could easily go out as-is.

services/QuillLMS/app/services/classlink_integration/schools_fetcher.rb
happythenewsad: Thanks for the heads up. The left-padding risk should be mitigated by the fact that we're doing regex matching, not integer matching here.

If they _send_ us "1" but intend "000000000001", we'll need to take action.

services/QuillLMS/app/services/classlink_integration/school_persister.rb
happythenewsad: noted

services/QuillLMS/spec/services/classlink_integration/school_persister_spec.rb
happythenewsad: Those fields are not intended to be updated (the block only runs on create)

services/QuillLMS/spec/services/classlink_integration/user_persister_spec.rb
happythenewsad: Yes, here it makes sense to express that an incoming new name should be updated. I've added that to the spec.

services/QuillLMS/app/services/classlink_integration/user_persister.rb
happythenewsad: 👀

brendanshean (approved):

PR #12943, opened 2025-04-30
more diagnostic insights qa

emilia-friedberg:
## WHAT
- swap out landing page image
- change date formatting in recommendation cards
- add a Segment event for recommended diagnostic assignment
- remove a period

## WHY
These are all requested QA changes.

## HOW
Mostly tiny HTML changes, except adding the Segment event, which just required writing and calling a worker and then testing it.

### Screenshots

### Notion Card Links
https://www.notion.so/quill/Diagnostic-Insights-QA-Ongoing-1b2d42e6f9418005b38cca6dc474422c?pvs=4

Tested locally, having Peter Sharkey test on staging

PR Checklist | Your Answer
------------ | -------------


services/QuillLMS/app/controllers/teachers/progress_reports/diagnostic_reports_controller.rb
```diff
@@ -192,6 +192,8 @@ def assign_post_test
   def assign_diagnostics
     params[:diagnostics].each do |diagnostic|
       Units::AssignmentHelpers.assign_unit_to_one_class(params[:classroom_id], diagnostic[:unit_template_id], diagnostic[:student_ids], false)
+      diagnostic_pack = UnitTemplate.find_by_id(diagnostic[:unit_template_id])
```
anathomical: This is a SUPER minor performance tweak that, given the size of the data involved, almost certainly won't matter, but since we only actually care about the `name` of the `UnitTemplate`, we can marginally reduce the size of the data payload that we fetch from the database by using `select`
```ruby
diagnostic_pack = UnitTemplate.select(:name).find_by_id(diagnostic[:unit_template_id])
```
This will create a `UnitTemplate` object that only has a `name` attribute.  Trying to call `diagnostic_pack.id` will error in this case because we didn't fetch that data.  Again, in this case, it's only one row, so we're not really saving much by fetching only the name, but I figured I'd mention it since there are use-cases where this can matter.

anathomical (approved):
A take-or-leave style note, but this looks good to me.

eadams17 (approved):

PR #12947, opened 2025-05-01
handle-skill-performance-with-missing-question-uid

anathomical:
## WHAT
Safely handle cases where ConceptResult.question_uid is nil
## WHY
This is quite rare for Diagnostic ConceptResults, maybe on the order of a couple of thousand in the whole system, but we still should handle these cases so that we get some reporting data for related activity sessions
## HOW
Exclude records where no `question_uid` was extractable (and thus no Skill can be identified) when calculating Skill Performance

- Run a number of known erroring Activity Session ids through the updated code and confirm that it works

PR Checklist | Your Answer
------------ | -------------


emilia-friedberg (approved):

happythenewsad (approved):
Style thing: I prefer `filter`, since `select` to me implies selecting one thing, not many 🤷‍♂️

PR #12953, opened 2025-05-01
fix-slsa-for-replayed-diagnostics

anathomical:
## WHAT
Add some workers to clean up ~250k records
## WHY
We had a bug that allowed students to replay the diagnostics earlier this year.  Most of the time they would launch a replay but not complete it, but initiating the second (or more) playthrough has resulted in ambiguity in StudentLearningSequenceActivity data where an incomplete replays sometimes squash complete playthroughs in reporting.  This should find all such instances and fix them.
## HOW
- Find any `ActivitySession` records where a student has replayed a diagnostic
- Find out which playthrough is the earliest completion
- Assign that playthrough to the `StudentLearningSequence`

### Notion Card Links
https://www.notion.so/quill/Diagnostic-Insights-QA-Ongoing-1b2d42e6f9418005b38cca6dc474422c?pvs=4#1e6d42e6f9418083964dd12b1fe18059

- Run on staging (approximately 60 minutes to enqueue all jobs)
- Look at known cases where this has happened and confirm that the data is now fixed

PR Checklist | Your Answer
------------ | -------------


services/QuillLMS/app/workers/student_learning_sequences/fix_replay_scoring_worker.rb
```diff
@@ -0,0 +1,27 @@
...
+    end
+
+    private def replayed_diagnostics
+      RawSqlRunner.execute(
+        <<-SQL
+          SELECT user_id, classroom_unit_id, activity_id
+            FROM activity_sessions
+            WHERE activity_id IN (2537, 2539, 2541, 2550, 2555, 2563, 1663, 1668, 1678, 1161, 1568, 1590, 1664, 2538, 1669, 2540, 1680, 2542, 1774, 2551, 1814, 2557, 1818, 2564)
+            GROUP BY user_id, classroom_unit_id, activity_id
+            HAVING COUNT(*) > 1
```
brendanshean, line 22: What about adding this condition on the group by:
`HAVING SUM( (completed_at IS NOT NULL)::int ) > 0`
That way we're restricting the job queue to only candidates with at least one completed sessions.

services/QuillLMS/app/workers/student_learning_sequences/find_and_set_completed_playthrough_worker.rb
```diff
@@ -0,0 +1,22 @@
+# frozen_string_literal: true
+
+module StudentLearningSequences
+  class FindAndSetCompletedPlaythroughWorker
+    include Sidekiq::Worker
+
+    sidekiq_options queue: SidekiqQueue::MIGRATION
+
+    def perform(user_id, classroom_unit_id, activity_id)
+      activity_sessions = ActivitySession
```
brendanshean: I think you could keep the entire computation in postgres:

```
first_completion_id  = ActivitySession
  .where(user_id:, classroom_unit_id:, activity_id:)
  .where.not(completed_at: nil)
  .order(:completed_at)
  .limit(1)
  .pluck(:id)
  .first
```

brendanshean (approved):
Thanks for cleaning these up!

PR #12957, opened 2025-05-02
Evidence: Add proper misc_threshold and rag_limit to retries.

dandrabik:
## WHAT
For retries, ensure that rag_limit and misc_threshold is set.
## WHY
misc_threshold is new, so wasn't in the code. Added a fallback for each since we always want this set.
## HOW
Add fallbacks to trialConfig

Ran a trial locally and seems to work.

PR Checklist | Your Answer
------------ | -------------
Have you deployed to Staging? |  No small change
Self-Review: Have you done an initial self-review of the code below on Github? |  Yes


services/QuillLMS/engines/evidence/app/views/evidence/research/gen_ai/trials/new.html.erb
brendanshean, line 1: I like the pattern of trial and then bumping up to dataset if it's not there.

brendanshean (approved):

PR #12958, opened 2025-05-05
Fix trial suites bug with relevant_text and classification

brendanshean:
## WHAT
Fixes a regression that crashed classification `LLMTrialSuitesCreator` when `relevant_text_id` was nil. The PR makes `RelevantText` optional during prompt creation

## WHY
Classification prompts never supply snippet‑level context, so forcing relevant_text_id blows up the trial runner. Making the association optional allows the trials to proceed as expected.
(The individual classification trials don't have this issue because `TrialRetryConfigLoader` is not used)
 

## HOW
Builder validation – dropped validates :relevant_text_id, presence: true. 
Prompt factory – wrap RelevantText.find and LLMPromptRelevantText.create! in if relevant_text_id.present? guards. 
Associations/specs – mark belongs_to :llm_prompt_template as optional, add missing has_one :llm, through: :trial, and expand spec coverage for nil‑context cases. 

### Screenshots

### Notion Card Links

I have tested some classification models on trial suites and they are now working properly.

PR Checklist | Your Answer
------------ | -------------


services/QuillLMS/engines/evidence/spec/models/evidence/research/gen_ai/llm_prompt_spec.rb
```diff
@@ -14,7 +14,7 @@
 #  suboptimal_guidelines_count :integer          not null
 #  created_at                  :datetime         not null
 #  updated_at                  :datetime         not null
-#  llm_prompt_template_id      :integer          not null
+#  llm_prompt_template_id      :integer
```
brendanshean, line 17: This is cleanup from an older commit.

services/QuillLMS/engines/evidence/app/models/evidence/research/gen_ai/llm_prompt.rb
```diff
@@ -64,7 +64,8 @@ def text = prompt
         def self.create_from_template!(dataset_id:, guideline_ids:, llm_prompt_template_id:, prompt_example_ids:, relevant_text_id:)
           guidelines = Guideline.where(id: guideline_ids)
           prompt_examples = PromptExample.where(id: prompt_example_ids)
-          relevant_text = RelevantText.find(relevant_text_id)
+          relevant_text = RelevantText.find(relevant_text_id) if relevant_text_id.present?
```
anathomical: Take or leave, since this is purely stylistic, but I tend to prefer using `.find_by(id: relevant_text_id)` over a guard clause as that will return `nil` if there is no match (as would be the case if `relevant_text_id` is `nil`).  Though that would open up the possibility of a present `relevant_text_id` not finding a match and failing to error, so your current approach may be safer.

services/QuillLMS/engines/evidence/spec/models/evidence/research/gen_ai/llm_prompt_spec.rb
```diff
@@ -42,12 +42,60 @@ module GenAI
...
+
+          before { allow(LLMPromptBuilder).to receive(:run).and_return('Generated prompt') }
+
+          context 'with relevant_text_id' do
+            let(:relevant_text) { create(:evidence_research_gen_ai_relevant_text, stem_vault:) }
+            let(:relevant_text_id) { relevant_text.id }
+
+            before { create(:evidence_research_gen_ai_dataset_relevant_text, dataset:, relevant_text:) }
+
+            it { expect(subject).to be_persisted }
```
anathomical, line 84: Today I discovered that RSpec will use metaprogramming to predicate-match `be_X` expectations.  Neat!

anathomical (approved):

services/QuillLMS/engines/evidence/spec/models/evidence/research/gen_ai/llm_prompt_spec.rb
brendanshean: RSpec is a really cool demonstration of metaprogramming, also check out other constructs like `let`, `subject`, and `before`.  If you want to go down a rabbit hole, look through the internals of FactoryBot.

Now that you mention, it feels like most effective ruby DSLs end up leveraging metaprogramming.

services/QuillLMS/engines/evidence/app/models/evidence/research/gen_ai/llm_prompt.rb
brendanshean: I actually like your suggestion better here as I don't actually want this to error.

PR #12959, opened 2025-05-05
diagnostic-insights-diagnostic-id-over-type

anathomical:
## WHAT
Use `diagnostic_id` as the param for performance and practice APIs
## WHY
We don't want to mix different versions of the diagnostics together when building reports about skills because the skills measured on different versions may differ from each other and result in confusing experiences for teachers.
## HOW
- Take `diagnostic_id` instead of `diagnostic_type` in the Performance and Practice services
- In the services, use `diagnostic_id` to find any existing `follow_up_activity_id` and use both pre and post diagnostic activity ids when building data
- Update controller to pass through the new value

### Notion Card Links
https://www.notion.so/quill/Diagnostic-Insights-QA-Ongoing-1b2d42e6f9418005b38cca6dc474422c?pvs=4#1e6d42e6f94180a09e02cc8e93de7162

- There's a set of front-end changes that need to be made to fully test this, but...
- Deployed to staging
- Manually hit the API endpoints with the new shape to confirm that they retrieve expected data and do not error

PR Checklist | Your Answer
------------ | -------------


emilia-friedberg (approved):
LGTM! Thanks for getting that up so fast.

PR #12960, opened 2025-05-05
Feedback Moderation Misc Part 1

brendanshean:
## WHAT
* Moderation toggles & status badges – Adds `active` switch to new/edit/show/index views of `FeedbackRule`, exposing the existing active boolean so rules can be enabled/disabled from the UI. 
* Restricts backend processing – `RunFeedbackEvaluationWorker` is updated to skip rules that are not active, preventing unnecessary evaluations. 
* Trial explorer UX polish – In trials#show the “#” and “match?” columns are merged, columns are resizable‑narrower, and table headers are now clickable for server‑side sorting. 
 
## WHY
* Fast rule moderation – exposing `active` so curriculum can flip a switch on rules instead of editing YAML or redeploying.
* Cost – The worker previously evaluated all rules, burning tokens on inactive/bad ones. Skipping inactive saves LLM calls and keeps Sidekiq queues lean.
* Trial running productivity – The trials table is the main debugging surface. Sortable headers and condensed columns cut scrolling time and make “problem first” triage simpler.  Badges & slimmer columns bring the feedback‑moderation screens in line with the rest of the Evidence admin UI.

## HOW
* Views & controller – Added form‑switch checkbox for :active; permitted param in `FeedbackRulesController`. Badge helper renders status everywhere.
* Worker guard – Early‑return unless feedback_rule.active?
* Sortable trials table – trials_controller#show now accepts sort/direction params and applies an order. ERB adds <%= link_to "Examples", sort_path(...) %> helpers; CSS tightens g-eval and match widths.

### Screenshots

### Notion Card Links
https://www.notion.so/quill/Add-the-Feedback-Moderator-to-Production-1ced42e6f94180cd9b6bd62dc936987c?pvs=4

I have tested these features on staging

PR Checklist | Your Answer
------------ | -------------


services/QuillLMS/engines/evidence/app/controllers/evidence/research/gen_ai/trials_controller.rb
```diff
@@ -79,34 +79,69 @@ def update
...
+          respond_to do |format|
+            format.html
+            format.json { render json: trial_json_response }
+          end
+        end
+
+        private def setup_trial_data
           @trial = trial
-          @histogram = @trial.api_call_times.map(&:round).tally if @trial.api_call_times.present?
+          @histogram = extract_histogram
```
anathomical: Take or leave this, since it's stylistic, but can we replace these `setup` methods with one-liner private methods?  For instance, `@trial` calls `trial` which already sets `@trial`.  This suggests to me that we should just be accessing `trial` everywhere in this class.  Similarly, we could have `private def histogram = @histogram ||= extract_histogram` and just reference `histogram`.  I've come to like that pattern quite a bit.  It reads cleaner to me these days, and it also allows for implicit hash construction:
```ruby
{
  trial:,
  histogram:,
}
```

services/QuillLMS/engines/evidence/app/helpers/evidence/research/gen_ai/trials_helper.rb
```diff
@@ -4,6 +4,12 @@ module Evidence
   module Research
     module GenAI
       module TrialsHelper
+        include ApplicationHelper
+
+        GENERATIVE_SORTABLE_COLUMNS = %w[match student_response curriculum_feedback llm_feedback revised_feedback feedback_errors revised_feedback_errors]
+        SORTABLE_COLUMNS = %w[match student_response curriculum_feedback llm_feedback revised_feedback feedback_errors revised_feedback_errors g_eval]
```
anathomical: These are almost identical to each other.  Is it worth using one to expand the other?
```ruby
GENERATIVE_SORTABLE_COLUMNS = %w[match student_response curriculum_feedback llm_feedback revised_feedback feedback_errors revised_feedback_errors]
SORTABLE_COLUMNS = GENERATIVE_SORTABLE_COLUMNS + %w[g_eval]
```

services/QuillLMS/engines/evidence/app/helpers/evidence/research/gen_ai/trials_helper.rb
```diff
@@ -18,7 +24,142 @@ def trial_duration(trial)
...
 
           duration = trial.trial_duration.round(2)
-          "#{duration} s (#{(duration / @trial.test_examples_count)&.round(2)}s / example)"
+          "#{duration} s (#{(duration / trial.test_examples_count)&.round(2)}s / example)"
+        end
+
+        def sort_examples(llm_examples:, example_groups:, sort_column:, trial:, g_evals:, g_eval_scores:, sort_direction: 'asc')
+          return llm_examples unless sort_column.in?(SORTABLE_COLUMNS)
+          return llm_examples if needs_generative_check? && !trial.generative?
+          return llm_examples if sort_column == 'g_eval' && !g_evals.present?
```
anathomical, line 44: This is stylistic, so take or leave, but given that all three conditions return the same result, would it make sense to write this as one complex condition or a helper method?  I can imagine one big condition being pretty hard to parse, and a helper method would need access to args passed to the method, so it's not obvious that either approach is superior to what you have here.

services/QuillLMS/engines/evidence/app/helpers/evidence/research/gen_ai/trials_helper.rb
```diff
@@ -18,7 +24,142 @@ def trial_duration(trial)
...
+
+          @sorting_groups = example_groups if example_groups
+
+          send("sort_by_#{sort_column}", llm_examples)
```
anathomical: Again, this is stylistic.  I'm always a little leery of `send` calls, especially with dynamic values.  I'd be inclined to set this up with a hash:
```ruby
SORT_BY_MAP = {
  "student_response" => sort_by_student_response,
}

SORT_BY_MAP[sort_column](llm_examples)
``` 
If you take this approach, you could also choose to make some of the values in the hash lambdas in cases where the logic is a one-liner, but I don't know if a mix of lambda and non-lambda references is actually more readable.

services/QuillLMS/engines/evidence/app/helpers/evidence/research/gen_ai/trials_helper.rb
```diff
@@ -18,7 +24,142 @@ def trial_duration(trial)
...
+          without_errors = llm_examples.reject { |e| e.feedback_errors.present? }
+
+          sorted_with_errors = apply_text_sort(with_errors) { |e| e.feedback_errors.to_s.downcase }
+
+          sorted_with_errors + without_errors
+        end
+
+        def sort_by_curriculum_feedback(llm_examples)
+          apply_text_sort(llm_examples) { |e| e.curriculum_proposed_feedback.to_s.downcase }
+        end
```
anathomical: This pattern is super-common throughout the file.  While I'm generally hesitant to engage in too much meta-programming (as indicated by my above flag of `send`), so feel free to ignore this thought, but I wonder if this isn't a reasonable use-case for it:
```ruby
def simple_text_sort(collection, attribute)
  apply_text_sort(llm_examples) { |e| e.send(attribute).to_s.downcase }
end
```
Which allows for: `def sort_by_curriculum_feedback(llm_examples) = simple_text_sort(llm_examples, :curriculum_proposed_feedback)` and similar patterns in all the other places that use the `e.<VALUE>.to_s.downcase` pattern.

anathomical (approved):
A number of style suggestions, but this seems like a reasonable set of updates to me.

services/QuillLMS/engines/evidence/app/helpers/evidence/research/gen_ai/trials_helper.rb
brendanshean: Good idea.  I actually don't like `send` either but was focused more on fixing the cyclotomic complexity linter error.

brendanshean: Actually, I'm making some updates in a another branch so I will make this requested in change in that branch.

services/QuillLMS/engines/evidence/app/helpers/evidence/research/gen_ai/trials_helper.rb
brendanshean: Oh.  Good point.

services/QuillLMS/engines/evidence/app/helpers/evidence/research/gen_ai/trials_helper.rb
brendanshean: I'm actually fine with `send` on a model attribute.  I think I find send with method calls a little more opaque.

services/QuillLMS/engines/evidence/app/controllers/evidence/research/gen_ai/trials_controller.rb
brendanshean: Yeah, this is a good call out.  I prefer this pattern too.  I was again focusing more on refactoring to get cyclomatic complexity and forgot to do another refactor.

services/QuillLMS/engines/evidence/app/helpers/evidence/research/gen_ai/trials_helper.rb
brendanshean: I debated this one too.  I'm not crazy about either approach but if we end up needing more conditions, I'll probably refactor then.

PR #12964, opened 2025-05-07
Feat/diagnostic insights practice insights

emilia-friedberg:
## WHAT
Add Diagnostic Insights Practice Insights pages.

## WHY
These are the last pieces of core functionality for Diagnostic Insights!

## HOW
React and CSS. Also, in response to a QA item Peter Sharkey identified, I've introduced a new hook to allow data table column widths to scroll down to a given width.

### Screenshots
Screenshot 2025-05-07 at 11 58 42 AM
Screenshot 2025-05-07 at 11 58 56 AM

### Notion Card Links
https://www.notion.so/quill/Diagnostic-Insights-Practice-Insights-1e3d42e6f94180389bcde8d9d7ee5423?pvs=4

Tested locally, having Peter Sharkey QA on staging.

PR Checklist | Your Answer
------------ | -------------


services/QuillLMS/client/app/bundles/Teacher/components/progress_reports/diagnostic_insights/performance_insights/index.tsx
```diff
@@ -18,12 +19,7 @@ import {
 const fileDocumentImgSrc = `${EXTRA_SMALL_ICON_BASE_SRC}/file-document.svg`;
 
 interface SkillPerformancesByDiagnosticTypeInterface {
-  starter?: SkillPerformanceInterface[];
-  intermediate?: SkillPerformanceInterface[];
-  advanced?: SkillPerformanceInterface[];
-  ell_starter?: SkillPerformanceInterface[];
-  ell_intermediate?: SkillPerformanceInterface[];
-  ell_advanced?: SkillPerformanceInterface[];
+  [key:number]: SkillPerformanceInterface[];
```
anathomical, line 22: Whoops.  Thanks for catching this.

services/QuillLMS/client/app/bundles/Shared/hooks/useScaledTableColumnWidths.tsx
```diff
@@ -0,0 +1,75 @@
...
+
+function areArraysEqual(arr1: string[], arr2: string[]): boolean {
+  return arr1.length === arr2.length && arr1.every((val, i) => val === arr2[i]);
+}
+
+export default function useScaledTableColumnWidths({
+  columnWidths,
+  containerRef,
+  maxTableWidth = 1200,
+  minTableWidth = 963,
```
anathomical: Is it worth defining these as potentially shared constants somewhere?

services/QuillLMS/client/app/bundles/Shared/hooks/useStickyTableHeader.tsx
```diff
@@ -4,93 +4,107 @@ interface UseStickyTableHeaderOptions {
   wrapperRef: React.RefObject<HTMLDivElement>;
   tableRef: React.RefObject<HTMLTableElement>;
   enabled: boolean;
+  dependencies?: any[];
```
anathomical: This is a pure style question, but does it make sense to make this an optional param with `?` when we're declaring a default below on line 14?  I doubt it matters much to go either way, but I'm just wondering what the "right" way to do this is, and it might be the case that what you have here is already right.

anathomical (approved):
Looks good.  I like the approach you've taken with flexible row widths, and the thorough testing you've put in place around that is awesome.

services/QuillLMS/client/app/bundles/Shared/hooks/useScaledTableColumnWidths.tsx
emilia-friedberg: Good callout -- I don't think these should be defined defaults here, because they're pretty specific to the diagnostic insights pages (where we have a limited container width due to the sidebar). I'll put them as consts in that shared directory, though, so if the style changes we only have to update it in one place.

services/QuillLMS/client/app/bundles/Shared/hooks/useStickyTableHeader.tsx
emilia-friedberg: Good catch. It shouldn't be optional, both because there's a default value and also because you can't spread a null value, like we're doing below. I'll fix that.

eadams17 (approved):
Very excited to see these last parts go out!

PR #12979, opened 2025-05-14
backfill-missing-slsa-completions

anathomical:
## WHAT
Backfill missing `StudentLearningSequenceActivity` completions
## WHY
We want this data to show up in Diagnostic Insights and future `StudentLearningSequence` based reports
## HOW
- Use BigQuery to find `ActivitySession` records that should be associated with a completed `StudentLearningSequenceActivity` that aren't so associated
- Enqueue those `ActivitySession` records for `HandleCompletion` runs so that they will be properly associated
- Add an early return in the service to handle cases where BigQuery has records of an `ActivitySession` that has been deleted in the LMS

### Notion Card Links
https://www.notion.so/quill/Diagnostic-Insights-QA-Ongoing-1b2d42e6f9418005b38cca6dc474422c?pvs=4#1edd42e6f941807190e3f265403b6b8f

- Ran the entire workflow through staging with 4 `migration` workers, and ensured that all jobs completed
- Once this has run on production and that data syncs to BigQuery, will use the same BigQuery query to confirm that these records are no longer "missing"

PR Checklist | Your Answer
------------ | -------------


dandrabik (approved):
Looks good to me, thanks for chatting this through.

PR #12988, opened 2025-05-16
diagnostic insights qa

emilia-friedberg:
## WHAT
Fix several outstanding Diagnostic Insights QA issues, including:

- timespent values on Practice Insights not reflecting the sum of the relevant sessions when you look at Activity Summary
- persist student selection when the classroom hasn't changed, even if a user navigates to a different tab in the Diagnostic Insights and then back by way of By Class
- fix bug where the recommendation card on the Individual Student page was showing that the diagnostic was not completed when there are any assignments that are incomplete
-  fix bug where changing the timeframe disabled the View Results button on the aforementioned recommendation card

## WHY
These are all QA items we want to fix before launch.

## HOW
- the timespent issue was a rounding problem, where we were always rounding down on the frontend to the nearest minute in both the Activity Summary and these reports, but because these reports are using the sum of many activity sessions' times, the extra seconds aggregate into multiple extra minutes, resulting in discrepancies. I updated the backend to pass back an array of timespent values, rather than a sum, and then wrote a new frontend function to calculate the total with each individual value rounded down the same way we do in the normal `getTimeSpent` function
- the student persistence issue was caused by a callback on `[classroom]` that reset the `selectedStudentId` value to null when it was called -- since this is called on initial load, it was always resetting it, so I just updated it to only reset if the student does not exist in that classroom's `studentLearningSequences` data
- the diagnostic not completed issue was caused by us using the first `previouslyAssignedDiagnostic` record for that student, which was sometimes incomplete even when the student had another, completed, record. I've updated that code to first check for a record with a `completed_at` value, before defaulting to the first if none exist.
- the timeframe changing issue was caused by a relevant `studentLearningSequence` not being present when it didn't exist in that timeframe, so `activity_session_visible` was always falsey. I updated the backend here to pass a `visible` attribute on the `previouslyAssignedDiagnostic` record itself, which is agnostic to timeframe, and then updated the frontend to check that instead.

### Screenshots

### Notion Card Links
https://www.notion.so/quill/Diagnostic-Insights-QA-Ongoing-1b2d42e6f9418005b38cca6dc474422c?pvs=4

Tested all of these changes locally, Peter Sharkey will confirm on production

PR Checklist | Your Answer
------------ | -------------


happythenewsad (approved):
LGTM

services/QuillLMS/app/services/diagnostic_insights/diagnostics_assigned_by_classroom.rb
```diff
@@ -29,7 +29,8 @@ def run = grouped_results
           created_at: slsa.created_at,
           user_id: slsa.student_learning_sequence.user_id,
           activity_session_id: slsa.activity_session_id,
-          student_learning_sequence_id: slsa.student_learning_sequence.id
+          student_learning_sequence_id: slsa.student_learning_sequence.id,
+          visible: slsa.activity_session&.visible
```
anathomical: I think that here, and in the `eager_load` below, you probably want to use `activity_session_unscoped` instead of `activity_session` so that we include records with `visible: false`.

services/QuillLMS/client/app/bundles/Teacher/helpers/studentReports.tsx
```diff
@@ -1,6 +1,50 @@
...
+    } else {
+      const hours = Math.floor(seconds / 3600);
+      const minutes = Math.floor((seconds % 3600) / 60);
+      totalMinutes += hours * 60 + minutes; // e.g. 7320 → 2 hrs 2 mins
+    }
+  })
+
+  if (totalMinutes < 1) return '<1 min';
+  if (totalMinutes === 1) return '1 min';
+  if (totalMinutes < 60) return `${totalMinutes} mins`;
```
anathomical: Should we be using `pluralize` for this and the `hoursText` below?

services/QuillLMS/client/app/bundles/Teacher/helpers/studentReports.tsx
```diff
@@ -1,6 +1,50 @@
...
+    } else if (seconds < 3600) {
+      totalMinutes += Math.floor((seconds % 3600) / 60); // e.g. 299 → 4 mins
+    } else if (seconds < 3660) {
+      totalMinutes += 60; // "1 hr"
+    } else {
+      const hours = Math.floor(seconds / 3600);
+      const minutes = Math.floor((seconds % 3600) / 60);
+      totalMinutes += hours * 60 + minutes; // e.g. 7320 → 2 hrs 2 mins
+    }
+  })
```
anathomical: This is stylistic, so take or leave, but this stack of if/else statements is kind of hard to follow.  I think I'd do a map into a reduce:
```javascript
const roundedMinutesArray = secondsArray.map(seconds => Math.floor(seconds / 60))
const totalRoundedMinutes = roundedMinutesArray.reduce((seconds, sum) => sum + seconds, 0)
const minutes = totalRoundedMinutes % 60
const hours = Math.floor(totalRoundedMinutes / 60)
```

anathomical (approved):
Looks good.  A stylistic suggestion on how we do the rounding and stuff, but you could keep the approach as-is, I think.

services/QuillLMS/app/services/diagnostic_insights/diagnostics_assigned_by_classroom.rb
emilia-friedberg: @anathomical I updated this, but I'm not sure it does really matter -- we're only using the activity session for its `visible` value, and if the activity session isn't visible, this will just return `nil` anyway, right? I guess maybe it's better to have an actual boolean than a falsy value.

services/QuillLMS/client/app/bundles/Teacher/helpers/studentReports.tsx
emilia-friedberg: Much cleaner, thanks!

PR #12994, opened 2025-05-19
sls-practice-assignments-disambiguation

anathomical:
## WHAT
Identify and re-assign another batch of cases where a `StudentLearningSequenceActivity` was assigned to the wrong `StudentLearningSequence`
## WHY
We want practice and post diagnostic data to attach to the correct playthrough of the pre diagnostic in cases where there have ben multiple playthroughs in the same classroom.
## HOW
- This version identifies any `StudentLearningSequenceActivities` which have a `completed_at` date that is earlier than the `completed_at` date for the pre diagnostic associated with the same `StudentLearningSequence`.  In cases where these exist (~260,425 in staging), we check other `StudentLearningSequences` for the same activity and classroom, and if one exists with a valid (earlier) pre diagnostic completion, reassign the `StudentLearningSequenceActivity`.  Note that this is about 50% of cases from what I can tell with database queries.
- Note that I've refactored the enqueue job to take a "max pages" argument instead of hard-coding it.  This is a better pattern, I think, and also allows for easier parallelization if that seems useful.  I've run it in staging as `StudentLearningSequences::Backfill::EnqueueMisassignedSlsaWorker.perform_async(1, 100_000, 3)`

### Notion Card Links
https://www.notion.so/quill/Practice-assigned-to-the-wrong-diagnostic-1f4d42e6f94180fb93d1f6a6a3c7e00c?pvs=23

- Deploy workers to staging
- Run them
- Confirm that known test case from original QA
- Identify "remaining" cases to understand what these look like:
- - Deleted `classroom_units`
- - Archived teachers
- - Unclear data discrepancies (see note "Example of potentially unresolvable case" in Notion card linked above)

PR Checklist | Your Answer
------------ | -------------


happythenewsad:
@anathomical  Presumably we're moving from BQ to PG for staleness reasons? Any other reason?

anathomical:
> @anathomical Presumably we're moving from BQ to PG for staleness reasons? Any other reason?

That is the only reason.  That and confirmation that the PostgreSQL queries are performant enough for the scale we're operating at.

happythenewsad (approved):

services/QuillLMS/app/workers/student_learning_sequences/backfill/enqueue_misassigned_slsa_worker.rb
```diff
@@ -7,45 +7,46 @@ class EnqueueMisassignedSlsaWorker
...
-            WHERE initial_classroom_units.created_at > classroom_units.created_at
+                AND slsa.completed_at IS NOT NULL
+                AND sls.initial_activity_id <> slsa.activity_id
+            JOIN student_learning_sequence_activities AS initial_slsa
+              ON sls.id = initial_slsa.student_learning_sequence_id
+                AND initial_slsa.completed_at IS NOT NULL
+                AND sls.initial_activity_id = initial_slsa.activity_id
+            WHERE initial_slsa.completed_at > slsa.completed_at
+              AND slsa.completed_at IS NOT NULL
+              AND slsa.completed_at >= '2023-08-01'
```
brendanshean, line 49: What's the purpose of the '2023-08-01' here?

anathomical: This is the farthest back that Diagnostic Insights reports will go, so we're avoiding digging into older edge cases from the messier data that exists in the database further back.

dandrabik (approved):
:+1: Looks good to me, thanks for moving these forward.

brendanshean (approved):

PR #13000, opened 2025-05-20
make sure classroom unit callback fires synchronously for demo creation script

emilia-friedberg:
## WHAT
Update the demo account creation script to run the same callback as `ClassroomUnit`, but run it synchronously.

## WHY
Peter Sharkey noticed that the demo account wasn't always having practice activities populate correctly. This turned out to be because they weren't getting `StudentLearningSequenceActivities` saved for them, because sometimes when the worker to populate that fired, the original `StudentLearningSequence` hadn't yet been created. That only happens when a pre-diagnostic is assigned, so it's important that these workers fire in order, which wasn't the case with the async jobs. In normal practice, it's not really possible for a teacher to assign a diagnostic pack and then assign a recommendation quickly enough for that to be an issue, but because the demo account script generates so much data as once we need a workaround here.

## HOW
Just update the `create_classroom_unit` function to 

### Screenshots

### Notion Card Links
https://www.notion.so/quill/Diagnostic-Insights-QA-Ongoing-1b2d42e6f9418005b38cca6dc474422c?pvs=4#1f9d42e6f941804d80f9ee48deba5f33

Ran the script multiple times locally to confirm that data populates as expected.

PR Checklist | Your Answer
------------ | -------------


anathomical (approved):
Good find.

PR #13002, opened 2025-05-21
put back the delay on calling handlecompletionworker in the demo creator script

emilia-friedberg:
## WHAT
Put back in place the delay on calling `HandleCompletionWorker` 

## WHY
This is a weird issue I don't totally understand, but calling HandleCompletionWorker without the delay results in the `HandleCompletion` module not being able to find the newly-created `ActivitySession`, so the `StudentLearningSequenceActivity` records never get populated with completion data.  I had originally had a delay in place here because I noticed when originally updating this script that the data wasn't getting populated, but I thought it was an issue with the `ClassroomUnit` callback not having run yet, so I had removed the delay when I updated the `HandleAssignmentWorker` to be called synchronously.

## HOW
Just add back the delay I'd removed, which seems to fix the issue. The other solution would be to retry the worker if no activity session is found, but since I think we're only seeing this bug in demo account creation scripts (activity sessions are created when students start an activity normally, so they'll have been around for at least a couple of minutes before they're completed and this worker runs in the normal flow), it makes more sense to me to patch it here. Please let me know if you disagree.

### Screenshots

### Notion Card Links
https://www.notion.so/quill/Diagnostic-Insights-QA-Ongoing-1b2d42e6f9418005b38cca6dc474422c?pvs=4#1f9d42e6f941804d80f9ee48deba5f33

Tested locally and on staging.

PR Checklist | Your Answer
------------ | -------------


brendanshean (approved):
I think this is probably fine for this type of ad hoc job.

I try to avoid using `perform_in` for normal flow since there can be redis performance issues.

PR #13004, opened 2025-05-21
Evidence - Script to update mislabeled rag data.

dandrabik:
## WHAT
Re-labels a tiny amount of `LabeledEntry` records when the latest test data has a different label.
## WHY
We have some identical and near identical data in our rag examples (LabeledEntry) that doesn't match the label in the latest test sets. These might have been mislabeled or pulled from AutoML that mislabeled them.
## HOW
Loop through and check for matching conditions. Print out to save for auditing label

Note, for now we are only running this on prompts that haven't been migrated to avoid having to re-run experiments.
I've debated making this an exact match search or using 0.01, but spot checking, 0.02 seems to be a valid limit. Open to thoughts on this.
### Screenshots
```
Edited 2 records
#<data Thor::Sandbox::LabeledEntries::Edited prompt_id="255", labeled_entry_id=149217, entry="loss of creatures will be less.", curriculum_label="Label_3", labeled_entry_text="loss of creatures will be less.", labeled_entry_label="Label_0">
#<data Thor::Sandbox::LabeledEntries::Edited prompt_id="255", labeled_entry_id=267941, entry="farmers can clone healthy animals in order to grow the population of animals like cows, pigs, and sheep helping improve livestock.", curriculum_label="Optimal_1", labeled_entry_text="farmers can clone healthy animals in order to grow the population of animals like cows, pigs, and sheep.", labeled_entry_label="Label_1">
```

Ran this on staging for some activities and did a spot check on the updates, which seems reasonable so far.

PR Checklist | Your Answer
------------ | -------------
Have you added and/or updated tests? | No, non app change.
Have you deployed to Staging? | No, running this locally on staging DB


brendanshean (approved):
Thank you for cleaning these up.

Without getting deep into the data, I don't think I can say 0.02 or 0.01, and I think you're best able to make that call.

PR #13010, opened 2025-05-22
slsa-for-replays

anathomical:
## WHAT
- Migration to update unique indexes on `student_learning_sequence_activities` to allow for cases where a student replays a practice activity
- Update `HandleAssignment` logic to account for the changes to the index
- Update `HandleCompletion` service to properly set data when processing a replay completion
- - A since-fixed issue allowed students to improperly replay Diagnostics themselves, so in those cases we want to make sure the canonical completion is the first completion
- Add a job to identify all replays and process them for backfill processing
- Update `FixSlsaSlsAssignmentWorker` to properly respond to backfill processing in cases where the new replay logic has already created "correct" data and we no longer need to move one
## WHY
- The existing index only allows for one play-through of a given User + ClassroomUnit + Activity combination, but replays will have valid duplicates, so we should add ActivitySession to that combination for unique indexing
- We `HandleAssignment` to have the same behavior after this change
- We need logic so that `HandleCompletion` can deal with edge cases that we were handling via the index before
- We want to re-process only data for replays to keep this backfill relatively quick
- After this fix is run, some diagnostic play-throughs will have fixed completed_at dates (in cases where diagnostics got replayed), so we'll want to re-run `FixSlsaSlaAssignmentWorker` for records that will now be able to be assigned to the correct `StudentLearningSequence`
## HOW
- Drop the old index and add a new index with an additional column
- We were using `create_or_find_by` relying on an index violation to trigger `find_by` behavior, but with the change to the index we can't do that so we need `find_or_create_by`
- Update the service to handle a number of specific edge-cases discovered during this process, most notably an update to handle cases where a diagnostic was replayed and we want to ensure we only record the earliest playthrough from an assignment
- Query for `ActivitySession` records where there are multiple records for the same activity assignment
- Update the logic to handle a couple of newly-discovered edge cases from running this on staging

### Notion Card Links
https://www.notion.so/quill/Practice-assigned-to-the-wrong-diagnostic-1f4d42e6f94180fb93d1f6a6a3c7e00c
https://www.notion.so/quill/Practice-count-mismatch-1f4d42e6f941803c9683d2af12731233

- Deploy to staging
- Run all backfill code
- Confirm fix for 80% of identified cases
- - Identify an additional edge case that needs data clean-up which I will handle in a follow-on PR

PR Checklist | Your Answer
------------ | -------------


dandrabik (approved):
:+1: Thanks for moving this along!

PR #13018, opened 2025-05-23
report-earliest-diagnostic-replay

anathomical:
## WHAT
- Fix snapshot generation for Skill Performance API
- Add snapshot generation for Skill Practice API
- Fix a bug in History API that was finding most recent instead of first-completed diagnostic playthroughs in edge cases where a student replayed the diagnostic
- Update History Practice API to include practice records for units that have been marked `visible: false` by the teacher who assigned them
- Update Skill Practice API to pull practice data from `StudentLearningSequence` records instead of just `ActivitySession` to handle edge cases where a student completed an activity that was marked as recommended before but is not any longer
- Update `HandleCompletion` service to properly record completion records in edge cases where a student replays the diagnostic
- - We prefer the oldest play-through that is marked `visible: true`, or if none are marked `visible: true` we simply prefer the oldest
## WHY
- Snapshots help a lot with maintaining confidence that we're not introducing regressions
- The "most recent" logic is closer to what a teacher sees on the Teacher Diagnostic report, but we want Diagnostic Insights to show first-exposure
- We don't want actions by a teacher to hide data that other teachers might be able to access and find useful in the Diagnostic Insights report
- We want the report to be a valid historical snapshot of what happened at the moment, not one that can change as other data in the system changes
- This is a rare edge case, but one that can arise.  Because we want a teacher to be able to access the detailed report of a play-through if one is available, it is important that we use data from a `visible: true` play-through if there is one
## HOW
- Update thor task to pass activity_id instead of activity_type
- Mirror the logic used in the Skill Performance snapshot generation code, but applied to Skill Practice
- Change the sort logic we use to find the appropriate diagnostic play-through when there are multiples
- Join our practice counting through `student_learning_sequence_activities` instead of through `recommendations`
- Add additional logic to `HandleCompletion` when processing an `ActivitySession` for a diagnostic play-through

### Notion Card Links
- https://www.notion.so/quill/View-results-improperly-disabled-deleted-or-unassigned-1f4d42e6f941808996c3d0dd38a018f1
- https://www.notion.so/quill/Undercounting-Practice-when-activities-are-no-longer-recommended-1fcd42e6f94180c99f89d8ab493a7095
- https://www.notion.so/quill/Practice-count-mismatch-1f4d42e6f941803c9683d2af12731233
- https://www.notion.so/quill/Data-request-for-teachers-with-specific-edge-cases-in-their-data-1f4d42e6f941804eabf5d2c7a739675d

- Deployed to staging
- Confirmed code changes resolved known issues
- Did additional database queries to identify additional examples of edge cases, confirmed they were resolved by new code
- - Detailed: https://www.notion.so/quill/Practice-should-include-hidden-activity-sessions-1f9d42e6f94180558b13e3f72abdcf9d?pvs=25
- Confirmed with Peter S via call that the issues that he had observed were resolved

PR Checklist | Your Answer
------------ | -------------


services/QuillLMS/lib/tasks/diagnostic_insights.thor
```diff
@@ -130,7 +135,7 @@ class DiagnosticInsights < Thor
       results = classroom_ids.to_h do |diagnostic_type, classroom_id|
         data = runtime_and_results do
           ::DiagnosticInsights::SkillPerformanceByClassroom.run(classroom_id,
-            diagnostic_type,
+            2541,
```
dandrabik: Is this a purposeful hardcoded var here. If so, you might consider making this a constant whose name makes that clear.

services/QuillLMS/lib/tasks/diagnostic_insights.thor
```diff
@@ -146,9 +151,25 @@ class DiagnosticInsights < Thor
         end
         ["#{diagnostic_type} (#{data[0]}ms)", data[1]]
       end
+    end
+
+    private def save_skill_practice_by_classroom_snapshot(classroom_ids)
+      results = classroom_ids.to_h do |diagnostic_type, classroom_id|
+        data = runtime_and_results do
+          ::DiagnosticInsights::SkillPracticeByClassroom.run(classroom_id,
+            2541,
```
dandrabik: Same comment here.

dandrabik (approved):
Overall looks good. I left some comments about whether a variable was intentionally hard coded or not.

services/QuillLMS/app/services/student_learning_sequences/handle_completion.rb
```diff
@@ -30,16 +30,39 @@ def run
...
+      # In cases where a student replays a diagnostic (a very rare edge case)
+      # and only one of the play-throughs is visible: true (even rarer)
+      # we want to ensure we report on the visible: true play-through.
+      # Otherwise we want the earliest play-through
+      existing_completed_assignment.nil? ||
+      new_session_is_more_visible? ||
+      (new_session_completed_earlier? && old_and_new_have_same_visibility?)
+    end
+
+    private def new_session_is_more_visible?
```
happythenewsad: The `more_visible` phrasing is confusing. How about the more clunky but literal `new_session_visible_and_completed_session_invisible?`

happythenewsad (approved):
Approving, with a naming suggestion

PR #13019, opened 2025-05-27
pre-cache-recommendation-count

anathomical:
## WHAT
Pre cache recommendation count when a teacher lands on their dashboard
## WHY
This data can take up to 10 seconds to load for teachers with lots of student history data, but we don't want that data to pop in for them when they go to the Diagnostic Assignment page.  By triggering a the calculation on the dashboard, where a teacher is redirected to when they log in or when they just go to `quill.org`, it should be in cache for most users by the time they get to a page where it is needed.
## HOW
- Create a Sidekiq worker to calculate the data for a given user and cache it
- Add a `before_action` for the `dashboard` action to trigger the worker
- Update the code in the API that fetches this data to use the `.new.perform` on the newly-created worker instead of duplicating code
- - Note that we ultimately want to replace this with a Pusher listener rather than a synchronous API call in order to match what we do for other long-running API calls, but I wanted to consult with Emilia before pursuing that, and this code should accomplish what we want (and is an improvement over what we had) so seemed worth deploying now

### Notion Card Links
https://www.notion.so/quill/Recommendation-count-optimization-1f4d42e6f94180dfbb2befb2a4f70be6

- Deployed to staging
- Cleared cache
- Monitored Sidekiq when loading `/dashboard` to confirm the job was getting enqueued
- Navigated to a page that calls the load API and confirmed that a new job is not enqueued as the data is fetched from cache
- Confirmed that "normal" amounts of time navigating to the Diagnostic Assignment page (where the API is called) provide enough time for the worker to complete and the data to load instantly

PR Checklist | Your Answer
------------ | -------------


services/QuillLMS/app/controllers/diagnostic_insights_controller.rb
```diff
@@ -30,15 +30,8 @@ def diagnostics_assigned
   end
 
   def complete_history
-    # TODO: make this cache smarter or eradicate the need for it by speeding up the query
     complete_history_by_classroom_data = current_user.all_classrooms_cache(key: 'diagnostic_insights.complete_history_with_classroom_name') do
```
dandrabik: This should use the constant `RECOMMENDATION_COUNT_CACHE_KEY` defined below

services/QuillLMS/app/workers/diagnostic_insights/cache_recommendation_counts_worker.rb
```diff
@@ -0,0 +1,30 @@
...
+
+    sidekiq_options queue: SidekiqQueue::CRITICAL
+
+    PUSHER_EVENT = 'diagnostic-insights-recommendation-count-cached'
+
+    def perform(id)
+      user = User.find_by(id:)
+      return unless user
+
+      result = user.all_classrooms_cache(key: RECOMMENDATION_COUNT_CACHE_KEY) do
```
dandrabik, line 15: So this cache will expire all classrooms if any of a teacher's classrooms changes. You could cache on the classroom level so it would only expire the cache of updated classrooms, i.e. move the caching inside the loop. Depending how often these change and how long these queries are, that might be more efficient (I should note, there's some chance it would be slower or the same, if the queries are fast).

services/QuillLMS/app/workers/diagnostic_insights/cache_recommendation_counts_worker.rb
```diff
@@ -0,0 +1,30 @@
...
+            classroom_name: classroom.name,
+            non_ell: HistoryByClassroom.run(classroom.id, ell: false),
+            ell: HistoryByClassroom.run(classroom.id, ell: true)
+          }
+        end
+      end
+
+      SendPusherMessageWorker.perform_async(id, PUSHER_EVENT)
+
+      result
```
dandrabik, line 27: Seems like this isn't really needed here. Is this for testing, since it wouldn't be used by the application? Maybe this should test that the cache is set?

dandrabik (approved):
Overall looks good. For the sake of pushing things forward, you can ignore my performance and testing comments if they take time. We can always revisit later.

services/QuillLMS/app/workers/diagnostic_insights/cache_recommendation_counts_worker.rb
anathomical: I believe that ultimately this won't be necessary, but because we're making a synchronous call to `perform` in the controller to handle cases where the cache is empty (instead of setting the front-end up to handle Pusher messages yet, even though that should be something we do soon), the return value of this job actually does matter.

PR #13043, opened 2025-06-09
refactor out react-select-search

emilia-friedberg:
## WHAT
Replace all instances of react-select-search with our own `DropdownInput` component, via a new shared component called `QuestionSearch`.

## WHY
The curriculum team had several tweaks they wanted to make to the function of existing instances of react-select-search. Rather than modify the behavior of that library, I just replaced it with our own dropdown component, and fixed the other outstanding issues in a centralized new component.

## HOW
See "What".

### Screenshots

### Notion Card Links
https://www.notion.so/quill/Investigate-the-Search-within-the-Diagnostic-flow-1c1d42e6f94180c7b49af9f2e536f1b6

Tested locally, having Rachel QA on staging.

PR Checklist | Your Answer
------------ | -------------


happythenewsad (approved):
Nice cleanup 

I notice that we have both
```
app/bundles/Shared/components/internalTools
app/bundles/Shared/components/shared
```
That both contain shared components. 

I'm unclear on what the distinction is. If there is no distinction, we could merge these directories as part of your larger front end project.

services/QuillLMS/client/app/bundles/Shared/components/shared/dropdownInput.tsx
```diff
@@ -112,21 +113,20 @@ export class DropdownInput extends React.Component<DropdownInputProps, DropdownI
   }
 
   handleInputInteraction = (e) => {
-    const { disabled, isSearchable, onClick, value, } = this.props
-    const { active, menuIsOpen, computedOptions, } = this.state
+    const { disabled, onClick, } = this.props
+    const { active, menuIsOpen, } = this.state
```
anathomical, line 117: Love cleaning this up.

services/QuillLMS/client/app/bundles/Shared/components/internalTools/questionSearch.tsx
```diff
@@ -0,0 +1,37 @@
...
+  questionType?: string;
+}
+
+export const QuestionSearch = ({ questionType, options, onChange, }: QuestionSearchProps) => {
+  let formattedOptions
+
+  if (questionType === 'titleCards') {
+    formattedOptions = options.map((opt: { key: string, title: string }) => {
+      return { label: stringNormalize(opt.title), value: opt.key }
+    });
```
anathomical, line 22: This just reminds me of the super-hacky way that we manage Title Card data in our activities.  Fixing that is well beyond the scope of this PR, obviously, but one day it would be nice to make that data make sense.

services/QuillLMS/client/app/bundles/Shared/components/internalTools/questionSearch.tsx
```diff
@@ -0,0 +1,37 @@
...
+  } else {
+    formattedOptions = options.map((opt: { key: string, prompt: string } )=> {
+      return { label: opt.prompt ? stringNormalize(opt.prompt.replace(/(<([^>]+)>)/ig, '').replace(/&nbsp;/ig, ' ').replace(/&#x27;/g, '\u0027')) : 'No name', value: opt.key, };
```
anathomical: Take or leave this since it's purely stylistic, but I wonder if the chain of `replace` calls here is worth breaking into its own method.  I don't know that this particular set of replacements is worth putting in a shared method, but even if it was just a method in this file I would find this a bit more readable.

services/QuillLMS/client/app/bundles/Shared/components/internalTools/questionSearch.tsx
```diff
@@ -0,0 +1,37 @@
+import * as React from 'react'
+import { stringNormalize, } from 'quill-string-normalizer'
+
+import { DropdownInput, } from '../../index'
+
+interface QuestionSearchProps {
+  options: any[];
+  onChange: ({ value }) => void;
+  questionType?: string;
```
anathomical, line 9: Would it make sense to just use `titleCards?: boolean` here?  It looks like we don't actually use the value other than to see if it's actually `titleCards`.  Though it may make more sense to do it as you have here since it allows you to just pass through the `questionType` from the actual question data instead of examining it for the `titleCard` value earlier in the process.

anathomical (approved):

emilia-friedberg:
> Nice cleanup
> 
> I notice that we have both
> 
> ```
> app/bundles/Shared/components/internalTools
> app/bundles/Shared/components/shared
> ```
> 
> That both contain shared components.
> 
> I'm unclear on what the distinction is. If there is no distinction, we could merge these directories as part of your larger front end project.

Yeah, good call-out! `shared` is where we put Backpack components that could be used across all other bundles, but it's also become kind of a catch-all for things that are a little more specific than that -- like the `teacherPreviewMenu`, which is shared across some but not all of the tools and never in the `Teacher` or `Student` bundles. `internalTools` is a folder Eric created more recently as a place to put some of the duplicated components we're extracting from the internal tools of the pre-Evidence frontend tools, but I agree that in general these folders could be better organized. I'll take another look at that once I get back into this work.

PR #13046, opened 2025-06-10
fix caching bug for lesson recommendations

emilia-friedberg:
## WHAT
Make sure the cache for lesson recommendations takes the activity id into account.

## WHY
We were seeing a bug where the lesson recommendations endpoint was always returning the same recommendations, regardless of which activity the recommendations were supposed to be generated for. This fixes that.

## HOW
Just make sure the `activity_id` gets passed as part of the `groups` attribute to the caching logic, so it becomes part of the key for the cache.

Note: I'm not entirely sure we really need to be cacheing this endpoint. We don't cache the individual recommendations, and it seems pretty fast without the cache wrapper (as tested to confirm the bug was, in fact, a cacheing issue). Git Blame indicates it may have just been introduced as a proof of concept for the original implementation for the UserCacheable concern. Let me know if you think I should just remove it.

### Screenshots

### Notion Card Links
https://www.notion.so/quill/Updated-Diagnostics-Issues-20dd42e6f94180b0a673ddd73cb8e544?source=copy_link

Tested locally.

PR Checklist | Your Answer
------------ | -------------


brendanshean (approved):
Nice work on figuring this bug out.

As far as whether to keep or remove, I would probably remove it.  We can always add it back in later.

anathomical (approved):
I agree with Brendan that we could just remove this and add it back later, but I think that should just be a separate PR so that we get this fix merged.   That way if we do decide to roll back to using caching there, we do so with the fixed code not the old code.

brendanshean:
> I agree with Brendan that we could just remove this and add it back later, but I think that should just be a separate PR so that we get this fix merged. That way if we do decide to roll back to using caching there, we do so with the fixed code not the old code.

Good idea on the separate commit.

emilia-friedberg:
@ana

> > I agree with Brendan that we could just remove this and add it back later, but I think that should just be a separate PR so that we get this fix merged. That way if we do decide to roll back to using caching there, we do so with the fixed code not the old code.
> 
> Good idea on the separate commit.

Great points. Since I can't actually deploy this yet due to the Heroku situation, I'll hold off for now, but I'll write up another PR once this is merged to just remove the caching layer entirely.

happythenewsad (approved):

PR #13050, opened 2025-06-11
make sure to clear unit template caches for all flags on activity save

emilia-friedberg:
## WHAT
Add logic to:

- clear unit templates caches when activities are updated
- make sure caches for ALL user testing flags are cleared when unit template caches are cleared

## WHY
Jamie reported that changing the name of an activity in an activity pack was not changing the name displayed on the Featured Activity Packs page. I found two problems that may have been contributing to this issue:

1. We were not clearing unit template caches when an activity in the template was changed, and
2. We were only clearing caches for five of the possible flags that we use to retrieve these caches.

## HOW
See "What"

### Screenshots

### Notion Card Links
https://www.notion.so/quill/Copy-Bug-on-Activity-Packs-with-name-changed-20fd42e6f9418021b111cfdd2d78ce68?d=693d42e6f94182aba92b03e46754f9ae

Confirmed locally that updating an activity name now clears the cache and it is reset on load with the correct data.

PR Checklist | Your Answer
------------ | -------------


services/QuillLMS/app/models/unit_template.rb
```diff
@@ -138,20 +138,20 @@ def self.user_scope(user_flag)
 
   def get_cached_serialized_unit_template(flag = nil)
     cache_expiration_time = 600
-    cached = $redis.get("unit_template_id:#{id}_serialized")
+    cached = Rails.cache.read("unit_template_id:#{id}_serialized")
```
anathomical, line 141: Love this cleanup.

services/QuillLMS/app/models/unit_template.rb
```diff
@@ -190,20 +190,16 @@ def self.previously_assigned_activity_data(activity_ids = [], current_user = nil
...
+
+  private def wrap_and_clear_unit_template_cache
+    delete_relevant_caches!
     yield
-    $redis.del("unit_template_id:#{id}_serialized")
-    %w[private_ production_ gamma_ beta_ alpha_].each do |flag|
-      $redis.del("#{flag}unit_templates")
-    end
+    delete_relevant_caches!
   end
```
anathomical, line 204: Are we deleted on both sides of the `yield`?  Is that necessary?  I mean, cache clearing should be pretty fast, but it still feels like it might be inefficient.

services/QuillLMS/spec/models/unit_template_spec.rb
```diff
@@ -323,37 +323,34 @@
...
     before do
-      $redis.redis.flushdb
-      $redis.multi {
-        $redis.set('beta_unit_templates', 'a')
-        $redis.set('production_unit_templates', 'a')
-        $redis.set('gamma_unit_templates', 'a')
-        $redis.set('alpha_unit_templates', 'a')
-      }
+      Rails.cache.clear
+      Flags::FLAGS.each do |flag|
```
anathomical, line 327: I love this update since I bet this used to clear all the flags, but then we added new ones and didn't update this code.

anathomical (approved):

services/QuillLMS/app/models/unit_template.rb
emilia-friedberg: Yeah, I didn't want to change the existing behavior here, but it does seem strange to me too.

PR #13053, opened 2025-06-12
Zeitwerk cleanup

brendanshean:
## WHAT
Remove almost all zeitwerk configurations
* I'm leaving automl since the spelling is inconsistent across files (automl, auto_ml, autoML, Automl) and we can eventually just remove the code completely.
* I also left Graphiql and graphql as those should be refactored into something else at some point.

## WHY
These configurations are more trouble than they are worth since quite a bit of rails magic fails to account for them creating extra work each time (e.g. various rails generators).  This technical debt will only grow as we add more domain model elements that use these acronyms.   Also, Claude (and I wager other LLMs) do well with framework conventions, and I've noticed zeitwerk poses problems for them, even if you add explicit notes in a CLAUDE.md file.
Screenshot 2025-06-12 at 1 29 47 PM

## HOW
Rename 
1.  various forms of  `LLM` -> `Llm`
1. `GenAI` -> `GenAi`
1. `EthicalAIHelper` -> `EthicalAiHelper`
1. `SSORequest` -> `SsoRequest`
1. `StaffCSVUploader` -> `StaffCsvUploader`
1. `HTMLTagRemover` -> `HtmlTagRemover`
1. `MalformedJSONRemover` -> `MalformedJsonRemover`

Update zeitwerk files.

### Screenshots

1.  GenAI flows were confirmed to work
2. 'ConceptReplacementCMSWorker', 'ConceptReplacementLMSWorker',
    These are callable, but don't work, since they reference nonexistent class `OldConceptResult`.  Presumably this code isn't called anymore. 
3. 'ELLStarterDiagnosticEmailJob' - this works, got email    
4.  SSORequest - was able to log into LearnWorlds
5. Remaining ones were all checked manually:
    'EthicalAIHelper'
    'HTMLTagRemover'
    'PusherCSVExportCompleted'
    'ReportDemoAPCreator'
    'StaffCSVUploader'

PR Checklist | Your Answer
------------ | -------------


services/QuillLMS/engines/evidence/app/models/evidence/research/gen_ai/comparison.rb
```diff
@@ -21,9 +21,6 @@ class Comparison < ApplicationRecord
 
         validates :dataset_id, presence: true
 
-        scope :llms, -> { where(independent_variable: LLM).order(id: :desc) }
```
brendanshean, line 24: Dead code

services/QuillLMS/engines/evidence/lib/evidence/check.rb
```diff
@@ -11,8 +11,11 @@ class NoMatchedFeedbackTypesError < StandardError; end
       Plagiarism
     ]
 
-    AI_CHECK_MAPPING = Evidence::Activity::AI_TYPES
-      .index_with { |type| "Evidence::Check::#{type}".constantize }
+    AI_CHECK_MAPPING = {
+      Evidence::Activity::AUTO_ML => Evidence::Check::AutoML,
+      Evidence::Activity::GEN_AI => Evidence::Check::GenAi,
```
brendanshean, line 16: `GEN_AI = 'GenAI'` which is then used in `{ |type| "Evidence::Check::#{type}".constantize }` but this would break since the constant's value is now `GenAi`.  However, we can't change to ``GEN_AI = 'GenAi' since that constant is used in many places in the frontend, hence the explicit mapping declaration.

happythenewsad:
Thanks for addressing this tech debt @brendanshean . Given the app-wide scope of the changes, I'd like to see some QA tests beyond the scope of the LLM work. I will review in parallel.

happythenewsad (approved):

brendanshean:
> Thanks for addressing this tech debt @brendanshean . Given the app-wide scope of the changes, I'd like to see some QA tests beyond the scope of the LLM work. I will review in parallel.

Good idea.  I will check the other ones too.

anathomical (approved):

PR #13058, opened 2025-06-12
Add reasoning attribute to Llm model.

brendanshean:
## WHAT
Add a boolean field `reasoning` to the `Evidence::Research::GenAi::Llm` model.

## WHY
Adding this model-level distinction enables customized timeout behavior for reasoning models, which typically require extended processing time.

## HOW
Write a migration file and update model and controller code.

### Screenshots

### Notion Card Links

I have tested this on staging.

PR Checklist | Your Answer
------------ | -------------


happythenewsad (approved):

services/QuillLMS/db/migrate/20250612210306_add_reasoning_to_llms.evidence.rb
```diff
@@ -0,0 +1,16 @@
...
+# This migration comes from evidence (originally 20250612205600)
+class AddReasoningToLlms < ActiveRecord::Migration[7.1]
+  def change
+    add_column :evidence_research_gen_ai_llms, :reasoning, :boolean
+
+    reversible do |dir|
+      dir.up do
+        Evidence::Research::GenAi::Llm.update_all(reasoning: false)
+      end
+    end
```
anathomical, line 12: Hadn't seen this pattern.  Is this better than just setting a default value of `false` on the column?

anathomical (approved):
A question about setting historical values vs a database default, but I think either approach would be fine.

services/QuillLMS/db/migrate/20250612210306_add_reasoning_to_llms.evidence.rb
brendanshean: That's a good question.  Lately, I've been experimenting with this pattern (minus the `reversible` block) in migrations that add a column  (e.g. [here](https://github.com/empirical-org/Empirical-Core/blob/bs-llm-reasoning/services/QuillLMS/engines/evidence/db/migrate/20240625122344_add_various_fields_to_trial_and_llm_prompt.rb#L17)).  I've noticed that being deliberate when setting values to ends up paying dividends later when auditing values of records.  For example, If I have a record has an attribute equal to `false` is that because it's a default or because I explicitly set it to `false`.

brendanshean: Also in the case of LLMs, it's not clear what the default should be here as more models add in reasoning.  The reasoning attribute here is really a proxy for subtype and the question then becomes, what is the default subtype.

PR #13061, opened 2025-06-13
new-diagnostics

anathomical:
WARNING: Do not merge until we're ready to launch the new diagnostics.  Intended to be June 25th.

## WHAT
Update all the constants that point to diagnostics to use the new diagnostics

Note that we're only updating two diagnostics at this time: the non-ELL Starter and the non-ELL Intermediate.
## WHY
We want teachers to get access to the new diagnostics so that they get better skill differentiation and practice recommendations
## HOW
- Update constants in the `Activity` model (and add new constants to keep track of the old values)
- Change the `uid` referenced on the Home page and the Diagnostic Tool page so that users preview the new Starter if they're not logged in
- Update shared constants used in admin reports that determine which Diagnostics to allow filtering by based on selected report timeframe (note: the `two-years-ago` timeframe does not actually exist, but it is something we discussed implementing so I left the bones in for it to follow up on later)
- Update front-end constants for the new Diagnostics while adding new values to store the old values (similar to what we did in `Activity` model)
- Add the newly-retired Diagnostics to the Diagnostic Insights drop-down selection list for reports that require selection of a specific diagnostic
- Update the BigQuery materialized view for the Admin Diagnostic Growth Report so that it pulls in data from the new diagnostics
- Update front-end snapshots that include the new constants

### Notion Card Links
https://www.notion.so/quill/New-Diagnostic-Switchover-Code-Priority-1-20ad42e6f941809f9c26dc5c1cde3355

- Use the `/assign/diagnostic` workflow
- Confirm that it assigns the new diagnostic
- Complete the diagnostic with a test student
- Confirm that the Teacher Diagnostic Report shows all the new skills]
- Use the "Assign Recommendations" interface to assign the Post diagnostic, confirm that it assigns the new post diagnostic
- Confirm that Diagnostic Insights reports shows data for the new diagnostic by default
- Confirm that Diagnostic Insights recommendation assignment assigns the new diagnostics
- Confirm that Skill Performance and Skill Practice show the new skills for the diagnostic
- Find a teacher already approved for Diagnostic Insights, confirm that selecting the "2024" versions of the diagnostics in the drop-downs for Performance and Practice sections finds the appropriate data (and that the new unqualified versions have no data)

NOTE: We'll need to get the BigQuery changes into production to test the Admin Diagnostic Growth Report code, but since no one will have taken the new diagnostics, we should have time to do that testing.

PR Checklist | Your Answer
------------ | -------------


services/QuillLMS/client/app/bundles/PremiumHub/shared.tsx
```diff
@@ -17,7 +17,8 @@ export const STUDENT = 'student'
 export const FETCH_ACTION = 'fetch'
 export const RESET_ACTION = 'reset'
 
-export const OLD_DIAGNOSTIC_TIMEFRAME = 'last-school-year'
+export const PRE_2024_DIAGNOSTIC_TIMEFRAME = 'two-years-ago'
```
anathomical, line 20: `two-years-ago` is not actually a defined timeframe, but it is one that we discussed adding over this Summer when we originally launched Admin Diagnostic Growth Report last year, so I went ahead and scaffolded it here.  Note that that also means that the `oldDiagnosticTypeDropdownOptions` is never actually returned in production at the moment, so there could be arguments for removing it until we implement the "two years ago" timeframe, but I thought it'd be safer to leave it in place.

brendanshean (approved):
Nice to see some more updates to Diagnostics!

services/QuillLMS/client/app/bundles/PremiumHub/shared.tsx
```diff
@@ -67,14 +80,15 @@ export const restrictedPage = (
...
+  if (timeframe === PRE_2024_DIAGNOSTIC_TIMEFRAME) return oldDiagnosticTypeDropdownOptions;
+  if (timeframe === PRE_2025_DIAGNOSTIC_TIMEFRAME) return pre2025DiagnosticTypeDropdownOptions;
 
   return diagnosticTypeDropdownOptions
 }
 
 export function selectionsEqual(firstSelection, secondSelection) {
   return (
-    firstSelection == secondSelection || // less strict comparison so that undefined and null are treated as equal
+    firstSelection === secondSelection || // less strict comparison so that undefined and null are treated as equal
```
emilia-friedberg: We should remove this comment if we no longer need to preserve the `undefined/null` equality.

services/QuillLMS/client/app/bundles/Teacher/components/assignment_flow/assignmentFlowConstants.tsx
```diff
@@ -37,22 +37,27 @@ export const PRE_AP_WRITINGS_SKILLS_2_UNIT_TEMPLATE_ID = 195
...
 export const ELL_STARTER_DIAGNOSTIC_ACTIVITY_ID = 2550
 export const ELL_INTERMEDIATE_DIAGNOSTIC_ACTIVITY_ID = 2555
 export const ELL_ADVANCED_DIAGNOSTIC_ACTIVITY_ID = 2563
 
+export const OLD_STARTER_DIAGNOSTIC_2024_ACTIVITY_ID = 2537
+export const OLD_INTERMEDIATE_DIAGNOSTIC_2024_ACTIVITY_ID = 2539
+export const OLD_STARTER_DIAGNOSTIC_2024_POST_ACTIVITY_ID = 2538
+export const OLD_INTERMEDIATE_DIAGNOSTIC_2024_POST_ACTIVITY_ID = 2540
+
 export const OLD_STARTER_DIAGNOSTIC_ACTIVITY_ID = 1663
```
emilia-friedberg: The plan is to come back through and change this (and the backend equivalent) in the near future, right? Otherwise I think it might just be worth changing to include 2023 in these titles as well.

emilia-friedberg (approved):
Thank you for doing this!

services/QuillLMS/client/app/bundles/Teacher/components/assignment_flow/assignmentFlowConstants.tsx
anathomical: It is the plan to remove this, but I think updating the name here until we do makes sense anyway.  I'll do that.

PR #13063, opened 2025-06-16
address reflow issue in admin reports and partially in my diagnostics

emilia-friedberg:
## WHAT
Address some reflow issues around the site, primarily in Admin Reports and to a less-complete extent in My Diagnostics.

## WHY
We were out of compliance with this WCAG standard: https://www.w3.org/WAI/WCAG21/Understanding/reflow.html. Basically, this means that at a viewport of 320px, only elements that require a two-dimensional layout should scroll -- so tables themselves can scroll, but the page around them should not. This PR fixes that for the admin reports that were out of compliance, and improves compliance in My Diagnostics (full compliance here would be a larger undertaking, and since the design roadmap includes redoing these pages, we decided not to address them until we implement that redesign).

## HOW
Just CSS! Also, I noticed while testing that Diagnostic Insights wasn't linked to on the mobile nav, so I fixed that.

### Screenshots
Screenshot 2025-06-16 at 12 46 53 PM

### Notion Card Links
https://www.notion.so/quill/Handle-reflow-issue-for-admin-reports-1d1d42e6f94180cbb6e9d30811665813?source=copy_link

Tested on variety of screen widths.

PR Checklist | Your Answer
------------ | -------------
Have you added and/or updated tests? | Manually tested -- CSS


services/QuillLMS/app/assets/stylesheets/pages/progress_reports/diagnostic_recommendations.scss
```diff
@@ -88,9 +88,7 @@
       margin-right: 24px;
     }
     .selection-buttons {
-      .quill-button {
-        margin-right: 8px;
-      }
+      gap: 8px;
```
anathomical, line 91: I didn't know about `gap`.  That's a useful one!

anathomical (approved):

services/QuillLMS/app/assets/stylesheets/pages/progress_reports/diagnostic_recommendations.scss
emilia-friedberg: @anathomical soooooo useful. It works for both flex and grid layouts.

PR #13069, opened 2025-06-20
Add active storage configuration to Dataset model

brendanshean:
## WHAT
1. Configure [Active Storage](https://guides.rubyonrails.org/v7.1/active_storage_overview.html) for attachments
1. Add file attribute to Dataset Model using active storage

## WHY
1.  Active storage has a tighter integration with Rails as it is built-in.  Also, it allows for multiple services and has a cleaner architecture than Carrierwave (e.g. no more custom uploader classes)
1.  Currently, CSV files are imported to create datasets, but they are not persisted.

## HOW
1.  Active Storage setup
    - Added require 'active_storage/engine' to config/application.rb
    - Installed and ran Active Storage migrations
    - Created storage.yml with local and S3 configurations
    - Configured environments (development, test, production) to use appropriate storage services
1.  Dataset configuration
    - Replaced attr_accessor :file with has_one_attached :file
    - Modified DatasetImporter calls to use StringIO with downloaded file content
    - Added file download functionality to show view

### Screenshots

### Notion Card Links

I have confirmed that the new attachment works by importing a dataset, and then successfully downloading the copy on the datasets#show page.

PR Checklist | Your Answer
------------ | -------------


services/QuillLMS/engines/evidence/app/controllers/evidence/research/gen_ai/datasets_controller.rb
```diff
@@ -63,13 +63,35 @@ def show
           end
         end
 
+        def download_file
+          dataset = Dataset.find(params[:id])
+          if dataset.file.attached?
```
dandrabik, line 68: Consider an early return, e.g.
```ruby
if !dataset.file.attached?
  redirect_to dataset, alert: 'No file attached to this dataset'
  return
end

send_data dataset.file.download,
              filename: dataset.file.filename.to_s,
              type: dataset.file.content_type,
              disposition: 'attachment'

File: services/QuillLMS/engines/evidence/app/controllers/evidence/research/gen_ai/datasets_controller.rb [hunk 2]
```diff
@@ -63,13 +63,35 @@ def show
...
 
         private def create_dataset_from_file_for_html_request
-          @dataset = stem_vault.datasets.new(task_type: Dataset::GENERATIVE)
+          @dataset = stem_vault.datasets.new(task_type: Dataset::GENERATIVE, file:)
 
           if @dataset.save
-            DatasetImporter.run(dataset: @dataset, file:)
+            if @dataset.file.attached?
+              errors_message = DatasetValidator.run(file: @dataset.string_io_file)
+              if errors_message.present?
```
dandrabik: Take or leave (since I think this would be cleaner but not 100% sure), you could use [validations contexts](https://boringrails.com/tips/activerecord-validation-context) here, e.g. 

```ruby
#controller
dataset.valid?(:file_upload)

# model
validate :valid_file_content, on: :file_upload

**Review by dandrabik — APPROVED** [review]
Nice work on this! I haven't used active_storage before, the DSL looks pretty nice! Small take or leave clean up notes, but nothing blocking.

File: services/QuillLMS/engines/evidence/app/controllers/evidence/research/gen_ai/datasets_controller.rb (hunk 1 above)
Comment (brendanshean) [reply services/QuillLMS/engines/evidence/app/controllers/evidence/research/gen_ai/datasets_controller.rb:68]: Normally, I prefer guard clauses, but I find the traditional if/else to be easier to read in this case, especially with the negated if.

File: services/QuillLMS/engines/evidence/app/controllers/evidence/research/gen_ai/datasets_controller.rb (hunk 2 above)
Comment (brendanshean) [reply services/QuillLMS/engines/evidence/app/controllers/evidence/research/gen_ai/datasets_controller.rb]: Much cleaner.  Thanks for this suggestion.  I didn't know about the context passing

File: services/QuillLMS/config/storage.yml [hunk 3]
```diff
@@ -0,0 +1,14 @@
...
+local:
+  service: Disk
+  root: <%= Rails.root.join("storage") %>
+
+amazon:
+  service: S3
+  access_key_id: <%= ENV['AWS_ACCESS_KEY_ID'] %>
+  secret_access_key: <%= ENV['AWS_SECRET_ACCESS_KEY'] %>
+  region: <%= ENV['AWS_REGION'] || 'us-east-1' %>
+  bucket: <%= ENV['AWS_S3_BUCKET'] || 'quill-lms-uploads' %>
```
happythenewsad: Since we use other S3 buckets in LMS, let's make this name more specific. `AWS_S3_UPLOADS_BUCKET`, for example

services/QuillLMS/db/migrate/20250620190548_create_active_storage_tables.active_storage.rb
```diff
@@ -0,0 +1,58 @@
...
+    create_table :active_storage_variant_records, id: primary_key_type do |t|
+      t.belongs_to :blob, null: false, index: false, type: foreign_key_type
+      t.string :variation_digest, null: false
+
+      t.index [:blob_id, :variation_digest], name: :index_active_storage_variant_records_uniqueness, unique: true
+      t.foreign_key :active_storage_blobs, column: :blob_id
+    end
+  end
+
+  private def primary_and_foreign_key_types
```
happythenewsad, line 51: Note to me: this is a Rails-generated method

services/QuillLMS/engines/evidence/spec/dummy/config/storage.yml
```diff
@@ -0,0 +1,7 @@
+test:
+  service: Disk
+  root: <%= Rails.root.join("tmp/storage") %>
+
+local:
+  service: Disk
+  root: <%= Rails.root.join("storage") %>
```
happythenewsad: No new line

happythenewsad (approved):
Cool; I hope this results in more modular code in the parent app in the future. 

No need to re-request review after the ENV VAR rename

services/QuillLMS/db/migrate/20250620190548_create_active_storage_tables.active_storage.rb
```diff
@@ -0,0 +1,58 @@
...
+# This migration comes from active_storage (originally 20170806125915)
+class CreateActiveStorageTables < ActiveRecord::Migration[7.0]
+  def change
+    # Use Active Record's configured type for primary and foreign keys
+    primary_key_type, foreign_key_type = primary_and_foreign_key_types
+
+    create_table :active_storage_blobs, id: primary_key_type do |t|
+      t.string   :key,          null: false
+      t.string   :filename,     null: false
+      t.string   :content_type
```
happythenewsad, line 12: I see no harm and only benefit in requiring making content_type a required field. We should always know the content_type at upload/download time

brendanshean: That's an interesting observation.  Initially, I handled content_type validation via a gem, [active_storage_validations](https://github.com/empirical-org/Empirical-Core/pull/13069/commits/78278c119bf534f131a50331576dd811f099151a), but opted not to include it just to keep the PR lighter and this particular feature is for internal tooling and only CSV's.  That flow already has data validation so I think we're good there.  

As far as the tool in general, some other points:

1. The creation APIs expect it can be nil.
`create_before_direct_upload!` (the path used for direct-uploads from the browser) explicitly accepts content_type: nil and persists the blob before the client uploads the file [api.rubyonrails.org](https://api.rubyonrails.org/v7.1/classes/ActiveStorage/Blob.html).  Likewise, `create_and_upload!` saves the blob before the file is inspected, so the column can be empty for a short window.

1. Identification is asynchronous.
Rails runs a background “identify” step that sets the MIME once the file is available; requiring NOT NULL would block that first insert and explode your direct-upload flow.

1. Rails already has a graceful fallback.
If both client input and automatic sniffing fail, Active Storage falls back to application/octet-stream—but that happens after the row exists [guides.rubyonrails.org](https://guides.rubyonrails.org/v7.0/active_storage_overview.html).

1. Edge cases matter.
You can deliberately skip sniffing with `identify: false`; in that case you may want content_type to stay blank until a later pass.  Finally, old data or third-party imports often lack MIME info. A hard DB constraint turns a harmless omission into a fatal error.

brendanshean: I think if we want to monitor content_type in a future PR, I'd recommend adding that gem back in.

happythenewsad: Got it, thanks for the context

services/QuillLMS/config/storage.yml
brendanshean: That's a good point.  Updated

PR #13071, opened 2025-06-24
classlink - final sign up / login flows

happythenewsad:
## WHAT
1. Key ClasslinkInstances off of tenant_id, not oneroster_application_id
2. relax various DB constraints
3. Allow any user (Quill account, no Quill account) x (Classlink acct, no Classlink acct) to login/signup

## WHY
1. tenant_id is the only unique shared resource key across the rostering flow and the new get-user-data-during-sso flow
2. because we need to create stub Districts and ClasslinkInstances in the new SSO flow, and that API does not have access to metadata like oneroster_application_id 
3. part of v1 spec

## HOW
1.  straightforward - see code
2.  migration
3. Update Classlink controller, but keep complexity to the Services as much as possible. Re-use rostering logic as much as possible, to minimize divergent business logic.

### Screenshots

### Notion Card Links

### Teachers
- [x] test normal (rostered Teacher) CL login (Julia West)
- [x] test new CL user signup (delete tqateam-0006@qateam.demo User and ClasslinkAccount from staging), then click `Sign up with ClassLink`
- [x] logout and sign in again 
- [x] run Rostering sync, and confirm that no duplicate assets were created for this user (idempotency)
- [x] confirmed CL user signup creates Instance and District, idempotently
- [x] ensure state is being passed properly upon button click (to distinguish logins from signups)

### Students
- [x] existing student should be able to sign in via the Login flow
- [x] existing student should be able to sign in via the Sign Up flow 

PR Checklist | Your Answer
------------ | -------------


services/QuillLMS/app/models/classlink_instance.rb
```diff
@@ -29,4 +30,6 @@ class ClasslinkInstance < ApplicationRecord
   has_many :users, through: :classlink_accounts
 
   has_many :provider_classrooms, as: :provider_instance, dependent: :destroy
+
+  validates :tenant_id, uniqueness: true, presence: true
```
happythenewsad, line 34: Keeping the validation at the AR level for now, since ClasslinkInstances already exist in the wild. There are < 6, so I plan to manually add tenant_ids to the existing records after deploy.

services/QuillLMS/app/services/classlink_integration/sso_user_importer.rb
```diff
@@ -0,0 +1,66 @@
...
+      @access_token = access_token
+      @classlink_instance = classlink_instance
+    end
+
+    def run
+      account = ClasslinkAccount.find_by(classlink_instance:, external_id: user_sourced_id)
+
+      return account if account
+
+      instance = DistrictsImporter.find_or_create_district_and_instance(
```
happythenewsad, line 21: A Classlink User being related to a `ClasslinkInstance` via a `ClasslinkAccount` is a heavy assumption in the LMS. Thus, we create stub `ClasslinkInstances` and `Districts` for new "orphan" ClassLink users, in a way that is compatible for future Rostering syncs, should they ever occur.

services/QuillLMS/app/services/classlink_integration/sso_user_importer.rb
```diff
@@ -0,0 +1,66 @@
...
+        {
+          'name' => "#{tenant_id} District",
+          'tenant_id' => tenant_id,
+          'tenant_name' => "#{tenant_id} District",
+
+        }
+      )
+
+      incoming_user = UsersFetcher.new(
```
happythenewsad, line 30: Lines 21-36 are the same flow an incoming user takes during the Rostering sync. UserPersister is idempotent

emilia-friedberg (approved):
LGTM!

services/QuillLMS/app/controllers/auth/classlink_controller.rb
```diff
@@ -3,33 +3,48 @@
...
+
     around_action :force_writer_db_role, only: [:classlink]
-    before_action :set_code, :fail_conflicting_auth_user_action, only: [:classlink]
+    before_action :set_code, :set_auth_credential, :fail_conflicting_auth_user_action, only: [:classlink]
+
+    attr_reader :code, :state, :auth_credential
 
-    attr_reader :code
+    NIL_AUTH_CREDENTIAL_ERROR_MSG = <<~MSG
+      A Quill account with your ClassLink account does not yet exist.
```
dandrabik: Should this be something like `"a Quill account with the email of your Classlink account does not exist.  Please sign up for a Quill account with that email"`? Basically something to mention it needs to share the email (or whatever needs to match to make this work).

services/QuillLMS/app/services/classlink_integration/auth_credential_saver.rb
```diff
@@ -2,18 +2,24 @@
...
     JWT_PASSWORD = nil
     JWT_SHOULD_VALIDATE = false
-    attr_reader :code
 
-    def initialize(code:)
+    attr_reader :code, :state
+
+    def initialize(code:, state:)
       @code = code
+      @state = state
```
dandrabik: You might consider a different var name than `state` since `login` isn't really a state. Maybe `action` or something else?

services/QuillLMS/app/services/classlink_integration/auth_credential_saver.rb
```diff
@@ -2,18 +2,24 @@
...
     end
 
     def run
       fetch_credentials
         .then { |creds| AuthCredentialSaver.format_credentials(creds) }
```
dandrabik: Is there a reason that `format_credentials` is a class method? Would a instance method like `generate_auth_credential` be a better fit?

services/QuillLMS/app/services/classlink_integration/auth_credential_saver.rb
```diff
@@ -27,8 +33,12 @@ def fetch_credentials
       Net::HTTP.post_form(uri, form_data)
     end
 
-    def self.generate_auth_credential(access_token:, user_sourced_id:)
-      account = ClasslinkAccount.find_by(external_id: user_sourced_id)
+    def generate_auth_credential(access_token:, user_sourced_id:, tenant_id:)
```
dandrabik: Consider making the non-run methods of your service object `private` to make it clear what should be called directly. You can always test private methods with `.send()`

services/QuillLMS/app/services/classlink_integration/auth_credential_saver.rb
```diff
@@ -39,12 +49,17 @@ def self.generate_auth_credential(access_token:, user_sourced_id:)
       )
     end
 
+    def new_user_attempting_login?(classlink_instance)
+      state == 'login' && classlink_instance.nil?
+    end
+
     def self.format_credentials(credentials)
       parsed_body = JSON.parse(credentials.body)
```
dandrabik: Take or leave, there's a way to refactor this class in which you memoize the API call and/or the JSON.parse and have more methods that can be called instead of passing around the vars and doing multiple things in each of the methods, e.g.

```ruby

def credentials
  @credentials ||= JSON.parse(fetch_credentials.body)
end
def access_token = credentials['access_token']
def id_token = credentials['id_token']

def jwt_payload = JWT.decode(id_token, JWT_PASSWORD, JWT_SHOULD_VALIDATE)
def tenant_id = ...

...

File: services/QuillLMS/app/services/classlink_integration/client.rb [hunk 9]
```diff
@@ -2,11 +2,14 @@
 
 module ClasslinkIntegration
   class Client
-    PROXY_URL = 'https://oneroster-proxy.classlink.io'
+    ENDPOINTS = {
```
dandrabik, line 5: Consider making this class a service object to match the other classes.

services/QuillLMS/app/services/classlink_integration/districts_importer.rb
```diff
@@ -27,30 +30,34 @@ def run
       end
     end
 
-    def find_or_create_district_and_instance(district_hash)
+    def valid_district?(district_response)
+      in_allowlist?(district_response['tenant_id']) &&
```
dandrabik: You are using these string keys for logic in multiple places, so they should be constants.

services/QuillLMS/app/services/classlink_integration/sso_user_importer.rb
```diff
@@ -0,0 +1,66 @@
...
+        district_oneroster_id: nil,
+        district_bearer_token: nil,
+        district_id: nil
+      ).transform_api_response(fetch)
```
dandrabik, line 34: You could consider making this a service:

```ruby

UserFetcher.run(
  district_oneroster_id: nil,
  district_bearer_token: nil,
  district_id: nil,
  api_response:
)

File: services/QuillLMS/app/services/classlink_integration/auth_credential_saver.rb [hunk 12]
```diff
@@ -27,8 +33,12 @@ def fetch_credentials
       Net::HTTP.post_form(uri, form_data)
```
dandrabik: Also, is there a reason we use `Net::HTTP` here and `HTTParty` in other places seems like we should use one (I believe we use `HTTParty` in most of our new code.)

dandrabik (changes requested):
Overall this looks like the right direction. I added enough refactor suggestions, that I think this is worth another pass. Happy to talk through these, so let me know if you want to chat.

services/QuillLMS/app/controllers/auth/classlink_controller.rb
nattnattdai: I think the goal of the error message is to encourage the user to sign up for a Quill account with ClassLink. That said, I suggest the message below!
"A Quill account linked with your ClassLink account does not exist. Please sign up for a Quill account with ClassLink [here](https://www.quill.org/account/new)."

happythenewsad: Thanks, Nattalie; using this updated copy

services/QuillLMS/app/services/classlink_integration/auth_credential_saver.rb
happythenewsad: The reason is that `format_credentials` doesn't read/write instance state. Using a class method is the most Ruby-native way to communicate that a function doesn't modify state. As of this PR, it's sibling method `generate_auth_credential` reads from state.

happythenewsad: (I opted to refactor, per your suggestion below, so this specific thread is not longer applicable)

services/QuillLMS/app/services/classlink_integration/sso_user_importer.rb
happythenewsad: `UsersFetcher` is already an ApplicationService. I am side calling one of it's methods here, in order to maximize shared code.

services/QuillLMS/app/services/classlink_integration/auth_credential_saver.rb
happythenewsad: Converted to HTTParty (HTTParty is better, since it auto-parses JSON when `Content-Type: application.json`, etc

services/QuillLMS/app/services/classlink_integration/client.rb
happythenewsad: The ApplicationService pattern is not the best fit for a RESTful client - see latest commit

dandrabik (approved):
Looks good, thanks for taking a refactor pass on this.

PR #13078, opened 2025-06-26
add accessibility settings table to the database

emilia-friedberg:
## WHAT
Add a new accessibility_settings table to the database.

## WHY
We need this in place to store `snackbar_duration`, and likely for other settings in the future.

## HOW
Just run the migration and add basic model/factory/spec files with annotations.

### Screenshots

### Notion Card Links

Ran the migrations and tests and confirmed that everything looks good.

PR Checklist | Your Answer
------------ | -------------


services/QuillLMS/app/models/accessibility_setting.rb
```diff
@@ -0,0 +1,17 @@
...
+# Table name: accessibility_settings
+#
+#  id                :bigint           not null, primary key
+#  snackbar_duration :integer
+#  created_at        :datetime         not null
+#  updated_at        :datetime         not null
+#  user_id           :integer          not null
+#
+class AccessibilitySetting < ApplicationRecord
+  belongs_to :user
```
brendanshean, line 18: What about a corresponding `has_one` in user.rb?

services/QuillLMS/app/models/accessibility_setting.rb
```diff
@@ -0,0 +1,17 @@
...
+
+  validates :user, presence: true
```
brendanshean: Some potential validations:

validates :snackbar_duration, presence: true, numericality: { greater_than: 0 }
validates :user_id, uniqueness: true. (<- this could also be db constraint)

services/QuillLMS/app/models/accessibility_setting.rb
brendanshean, line 1: I think there's a question of whether to use [store_accessor](https://api.rubyonrails.org/classes/ActiveRecord/Store.html) on the User model which means you don't have to do migrations in the future once that settings attribute is created (e.g. [trial.rb](https://github.com/empirical-org/Empirical-Core/blob/develop/services/QuillLMS/engines/evidence/app/models/evidence/research/gen_ai/trial.rb#L73))

I like to use that pattern if I suspect there will be a lot of attributes discovered in the future that don't require too much validation.  

Maybe a compromise would be adding a store_accessor on `AccessibilitySetting`

`store_accessor :config, :snackbar_duration`
 which would keep our User.rb from becoming even more heavy.

brendanshean (commented):
I left some comments, but nothing to hold up approval.

services/QuillLMS/app/models/accessibility_setting.rb
emilia-friedberg: I'm disinclined to put a `presence: true` validation on `snackbar_duration`, because I think we'll have to remove it in order to extend this table, and my plan is still to just default to the current timeout on the frontend if there's no value set, so it's not strictly necessary to have a value stored here. It is currently the case, however, that it doesn't really make sense to have this record if `snackbar_duration` doesn't exist, so let me know if you think it would be better to add it here and we can just remove it down the line if necessary.

services/QuillLMS/app/models/accessibility_setting.rb
emilia-friedberg: Good thought. I'm going with your compromise solution because I really do want to avoid storing additional stuff on `User`.

happythenewsad (approved):

services/QuillLMS/app/models/accessibility_setting.rb
brendanshean: That makes sense.  I would leave it out then.

brendanshean (approved):

anathomical (approved):
I think that using a new table and combining it with `store_accessor` is a great approach.  I like both avoiding new columns on the `users` table, and also letting us add new values without having to do future migrations.

PR #13080, opened 2025-06-26
fix sortable list handling

emilia-friedberg:
## WHAT
While working on replacing our datetime pickers, I noticed that using keyboard navigation within sortable list elements (ie, when the sortable element is a whole table row that includes buttons) wasn't working. This fixes that.

## WHY
All elements on a page need to be accessible by keyboard.

## HOW
Swap out `PointerSensor` and `KeyboardSensor` for built-ins that have element type awareness, so we can decide whether or not to activate the drag event based on the element type it's activated by.

### Screenshots

### Notion Card Links

Tested locally.

PR Checklist | Your Answer
------------ | -------------


brendanshean (approved):
Nice troubleshooting on this.

services/QuillLMS/client/app/bundles/Shared/components/shared/sortableList.tsx
```diff
@@ -29,246 +29,216 @@ const dropAnimationConfig: DropAnimation = {
...
+    return false;
+  }
+  const activeElement = document.activeElement;
+  if (
+    activeElement &&
+    ([INPUT, SELECT, TEXTAREA, BUTTON].includes(activeElement.tagName) ||
+      activeElement.getAttribute('contenteditable') === 'true')
+  ) {
+    return false;
+  }
```
anathomical, line 89: I don't know if this is actually any better, but another approach here to simplify your conditional would be to split the presence of `activeElement` out into its own early return:
```javascript
const activeElement = document.activeElement;

if (!activeElement) return true;

if ([INPUT, SELECT, TEXTAREA, BUTTON].includes(activeElement.tagName) || activeElement.getAttribute('contenteditable') === 'true') return false;

return true;
```

anathomical (approved):
Nice job figuring out what was going on here.  I've got a minor stylistic idea, but don't know if it's actually better than what you already have, so take it or leave it.

happythenewsad (approved):

PR #13084, opened 2025-06-30
update (almost) all instances of snackbar to use snackbar monitor hook

emilia-friedberg:
## WHAT
Update almost every place in the codebase that we use the Snackbar component to determine its visibility based on the `useSnackbarMonitor` hook, with a few exceptions (see HOW).

## WHY
We want to allow users to change the length of the snackbar timeout from the default. In order to do that, I'm consolidating all Snackbar monitoring code to the existing `useSnackbarMonitor` hook, and will make a follow-up PR to add logic to that hook to retrieve the setting if it exists and handle its own timing.

## HOW
Mostly just replacing existing timeout logic with a call to the hook. In some instances, I had to rewrite class-based components to functional ones in order to be able to use this hook. In those cases, I also rewrote or added tests to use react-testing-library in order to confirm parity between the class-based component and the functional one. Most of these files are old, since we haven't written new class-based ones in several years, and in some instances I found dead codepaths that I have removed.

There are a few outlying exceptions that use the `Snackbar` component but are still not using the `useSnackbarMonitor` hook:

These three files use the Snackbar component with no timeout (ie, it stays on the page in the current version), so I'm not updating the behavior:

```
./client/app/bundles/Evidence/components/turkerView/completed.tsx
./client/app/bundles/Staff/components/evidence/gatherResponses/turkSessions.tsx
./client/app/bundles/Teacher/components/accounts/login/login_form.jsx
```

And there is a final one:

`./client/app/bundles/Teacher/components/assignment_flow/create_unit/stage2/review_activities.jsx`

which has a big changeset in another open PR, so I'm implementing this change there instead to avoid merge issues.

### Screenshots

### Notion Card Links

Tested all changed components to confirm both snackbar behavior and broader behavior where more extensive changes were made.

PR Checklist | Your Answer
------------ | -------------


services/QuillLMS/client/app/bundles/Teacher/components/assignment_flow/manage_units/activities_unit.jsx
```diff
@@ -141,10 +128,6 @@ export default class ActivitiesUnit extends React.Component {
     return firstCa.ownedByCurrentUser;
   }
 
-  saveSortOrder = () => {
```
emilia-friedberg, line 144: This is never actually getting called in this component, so I deleted this and all of the relevant Snackbar code because that was the only place it was used.

anathomical (approved):

services/QuillLMS/client/app/bundles/Teacher/components/subscriptions/current_subscription.tsx
```diff
@@ -237,35 +251,34 @@ export default class CurrentSubscription extends React.Component {
...
+  const handleUpdateRecurringClick = () => {
     const { stripe_subscription_id, recurring, id, } = subscriptionStatus
 
     const newRecurringValue = !recurring
     const snackbarText = `Automatic renewal turned ${newRecurringValue ? 'on' : 'off'}.` // "recurring" is the old value, so the description is the inverse
 
-    const callback = this.setState({ showAutomaticRenewalModal: false, }, () => {
-      this.showSnackbar(snackbarText)
-    })
+    const callback = () => {
```
happythenewsad: `updateSubscriptionSucessCallback` is a more descriptive name

happythenewsad (approved):
Nice work!

PR #13087, opened 2025-07-01
diagnostic-insights-data-integrity-verification

anathomical:
## WHAT
Add a set of services that verify that the data integrity for the Diagnostic Insight report is valid for any given user and compiles any validation errors into a report for investigation.  Run this report every time someone loads the Diagnostic Insights page and send an email if data validation fails.
## WHY
We want to be confident in our data, that there are no bugs or edge cases that we haven't caught, and that if there are we find them even if a user doesn't notice them.  We'd rather find and fix these if they exist before a user has an issue.
## HOW
- Create a set of services to do the three tests we know are important for data integrity
- Create a wrapper service that runs all of those services for a given classroom
- Create a wrapper service that runs the validation for all classrooms for a specific user by email
- Create a thor task to allow arbitrary users to be tested on the console
- Create a worker that runs the by-email service
- - Include logic to debounce so that this worker doesn't run the calculation more than once every 24 hours
- - Have the worker send an email if there is a validation failure so that we can investigate
- Hook the worker up to the first API controller for Diagnostic Insights so that it triggers for every user who loads the report

### Notion Card Links
https://www.notion.so/quill/Diagnostic-Insights-QA-Ongoing-1b2d42e6f9418005b38cca6dc474422c?d=1dfd42e6f94180138b7d001c49cf2a70#1dfd42e6f9418077b2f8de852ce004d8
https://www.notion.so/quill/Write-script-to-automate-QA-process-1f4d42e6f94180a7a863f10e896983d4?d=1dfd42e6f94180138b7d001c49cf2a70&pvs=25#1dfd42e6f9418077b2f8de852ce004d8

- Deploy to staging
- Load Diagnostic Insights with a user (id `13749037`) that has known data issues and confirm email is received
- Load Diagnostic Insights again with the user and confirm that a second email is not generated
- Clear cache
- Load a third time and confirm that a new email is sent

PR Checklist | Your Answer
------------ | -------------


services/QuillLMS/app/mailers/diagnostic_insights/data_integrity_notification_mailer.rb
```diff
@@ -0,0 +1,14 @@
...
+
+module DiagnosticInsights
+  class DataIntegrityNotificationMailer < ::UserMailer
+    def invalid_data(data)
+      @data = data
+
+      mail to:, subject:
+    end
+
+    private def to = ENV.fetch('DIAGNOSTIC_INSIGHTS_REPORTING_EMAIL_ADDRESS')
```
brendanshean: I would store this as a constant, maybe in an initializer, that way it only needs to be fetched once.

services/QuillLMS/app/services/diagnostic_insights/data_integrity/verify_by_classroom.rb
```diff
@@ -0,0 +1,29 @@
...
+
+      def initialize(classroom_id)
+        @classroom_id = classroom_id
+      end
+
+      def run
+        return unless classroom
+
+        CLASSROOM_TESTS.to_h do |test|
+          [test.to_s.split('::')[-1], test.run(classroom_id)]
```
brendanshean: I think `demodulize` here is more explicit about intent:
```
CLASSROOM_TESTS.to_h do |test|
  [test.name.demodulize, test.run(classroom_id)]
end
```

services/QuillLMS/app/controllers/diagnostic_insights_controller.rb
```diff
@@ -10,6 +10,8 @@ class DiagnosticInsightsController < ApplicationController
   before_action :authorize_request, only: [:history_by_classroom, :skill_performance_by_classroom, :skill_practice_by_classroom, :diagnostics_assigned]
 
   def history_by_classroom
+    DiagnosticInsights::VerifyDataIntegrityWorker.perform_async(current_user.email)
```
dandrabik, line 13: Maybe condition this by `if current_user.email`

services/QuillLMS/app/mailers/diagnostic_insights/data_integrity_notification_mailer.rb
dandrabik: Note, I can set up a google group email address for this, or you set up an email address for a  Slack channel. Either way, let's send it to a group rather than an individual.

services/QuillLMS/app/services/diagnostic_insights/data_integrity/practice_counts_match.rb
```diff
@@ -0,0 +1,61 @@
...
+      private def ell_history = @ell_history ||= HistoryByClassroom.run(classroom_id, ell: true)
+      private def practice_by_classroom = @practice_by_classroom ||= PracticeByClassroom.run(classroom_id)
+
+      private def skill_practice_by_classroom
+        @skill_practice_by_classroom ||= Hash.new do |hash, diagnostic_id|
+          hash[diagnostic_id] = SkillPracticeByClassroom.run(classroom_id, diagnostic_id)
+        end
+      end
+
+      private def practice_counts_match_for_classroom(history)
```
dandrabik, line 29: Maybe I'm missing something, but this doesn't return a boolean, but a hash of diagnostic info. How is this verifying the counts match, i.e. how would this return false?

services/QuillLMS/app/services/diagnostic_insights/data_integrity/verify_by_classroom.rb
dandrabik: This first past can be `test.to_s.demodulize` or `test.name.demodulize`

services/QuillLMS/spec/services/diagnostic_insights/data_integrity/verify_by_classroom_spec.rb
```diff
@@ -0,0 +1,22 @@
...
+
+module DiagnosticInsights
+  module DataIntegrity
+    describe VerifyByClassroom do
+      subject { described_class.run(classroom.id) }
+
+      let(:user) { create(:teacher_with_one_classroom) }
+      let(:classroom) { user.classrooms_i_teach.first }
+
+      it do
```
brendanshean: I think this test file could be improved with some more tests that show the final output, specifically the return type of a successful interaction and maybe an unsuccessful one.

brendanshean (commented):
I feel like the implementation and organization of code here is pretty solid, but the test coverage is either missing files or relying exclusively on stubs.  Let's discuss offline to shore up that part of the PR.  

The nature of the computation here is quiet complex and I think more thorough testing will help facilitate refactoring and/or debugging down the road if needed.

services/QuillLMS/app/services/diagnostic_insights/data_integrity/verify_by_email.rb
```diff
@@ -0,0 +1,41 @@
...
+        compact_all_true(users.to_h do |user|
+          [user.email, compact_all_true(verify_user_data_integrity(user))]
+        end)
+      end
+
+      private def users = @users ||= User.where(email: email_addresses)
+
+      # Recursively collapses hashes in which all values are `true` into the simple value `true`.
+      # This allows us to report back cases where data doesn't match, but to avoid having to parse cases where data matches
+      private def compact_all_true(data)
```
happythenewsad, line 52: Unclear if the complexity of this compaction algorithm is worth the ease of using JSON.pretty_generate later on. Personally I'd prefer complexity at the decoration step, since testing the data before decoration is simpler and thus less error prone.

services/QuillLMS/app/services/diagnostic_insights/data_integrity/verify_by_email.rb
```diff
@@ -0,0 +1,41 @@
...
+        recursed_data = data.transform_values do |value|
```
dandrabik: This is a take or leave, but you might consider making this code more explicit using `Data` objects, e.g. `Result = Data.define(:user_email, :all_true, :data)` and just filter out the ones where `all_true` is true (their data can be nil)

services/QuillLMS/app/workers/diagnostic_insights/verify_data_integrity_worker.rb
```diff
@@ -0,0 +1,27 @@
...
+    sidekiq_options queue: SidekiqQueue::LOW
+
+    DEBOUNCE_CACHE_KEY = 'diagnostic_insights_verify_data_integrity:'
+
+    def perform(user_email)
+      user = User.find_by(email: user_email)
+      return unless user
+
+      cache_key = "#{DEBOUNCE_CACHE_KEY}#{user_email}"
+      return if Rails.cache.fetch(cache_key)
```
dandrabik: I think this should be refactored to:

```ruby
Rails.cache.fetch(cache_key) do
  data_valid = DataIntegrity::VerifyByEmail.run(user_email)

  if !data_valid
    DataIntegrityNotificationMailer.invalid_data(data).deliver_now!
  end
  return data_valid
end
```
I don't think your current code caches for valid data

services/QuillLMS/app/services/diagnostic_insights/data_integrity/verify_by_email.rb
dandrabik: Are we worried about this recursive function eating up memory? Maybe there's a way to do this without iteration, like filtered out and not calculating the data for matches?

dandrabik (changes requested):
Nice work moving this forward. I think there are a few things to check in this code before shipping that I'm requesting changes. Happy to chat through these items tomorrow.

services/QuillLMS/app/mailers/diagnostic_insights/data_integrity_notification_mailer.rb
anathomical: I think a Slack alert to `#diagnostic-insights` makes sense, so I'll set that up for now.

services/QuillLMS/app/controllers/diagnostic_insights_controller.rb
anathomical: I could do that, but the `before_action :authorize_request` should return a 403 if there is no `current_user`.  It probably wouldn't hurt to have an additional check, but it seemed unnecessary to me.  Let me know if you still think it's worth doing.

services/QuillLMS/app/services/diagnostic_insights/data_integrity/practice_counts_match.rb
anathomical: Good catch.  I've fixed this.

services/QuillLMS/app/workers/diagnostic_insights/verify_data_integrity_worker.rb
anathomical: I think this is slightly misleading.  We're not actually trying to cache the data, just using the presence of a cache key to debounce the job so we don't run it more than once a day.

dandrabik (approved):
We discussed some changes to make on Zoom, but you can ship once those are done without re-review. Thanks for closing out this project!

brendanshean (approved):
Thanks for adding in some more testing!

PR #13092, opened 2025-07-02
revert datetime reversion and fix package issues

emilia-friedberg:
## WHAT
This PR reverts the reversion of the new datetime picker and associated changes from develop and fixes the issues that caused the need to revert, as well as adding tests for the base components in each app so that we can catch similar issues before they go out in the future.

Note for review: the first commit here is just the reversion of the reversion, so I'd recommend only reviewing the later commits, which actually fix the issue: https://github.com/empirical-org/Empirical-Core/pull/13092/files/3582cf135e375ed0ce1f70d23112f4b20785ae54..739700e7e98892bf0df23c5966e0d9f8fb62864f

## WHY
We want this code to be live, but we don't want anything else to be broken as a result.

## HOW
When I was making all of the babel package changes in the reverted PR, I'd deleted the package locks to get the changes to apply and the tests to run properly, which inadvertently caused our versions of react-router related libraries to bump, resulting in the issue where the apps weren't loading as a result of package errors. I've fixed this by locking the package versions and applying "overrides" for the relevant packages to make sure their versions are stable. I also added snapshot tests to the base components for each of the Apps so that they'll fail if a similar issue arises in the future.

### Screenshots

### Notion Card Links

Testing on staging that each of the apps loads correctly.

PR Checklist | Your Answer
------------ | -------------


brendanshean (approved):
Nice work on quickly figuring out the issue.

happythenewsad (approved):
🚢

anathomical (approved):

PR #13093, opened 2025-07-02
allow users to change snackbar timeout

emilia-friedberg:
## WHAT
Allow users to change the duration for which the snackbar element persists on the screen from their Account Settings page.

## WHY
This puts us into compliance with https://www.w3.org/TR/WCAG22/#timing-adjustable.

## HOW
Add controller actions (and model modifications) to get and update the setting, then use those actions on the frontend to retrieve the setting within the `useSnackbarMonitor` hook and set it from the Teacher and Student account settings pages.

### Screenshots
Screenshot 2025-07-02 at 3 42 24 PM

### Notion Card Links

Tested locally to confirm that it is possible to set and change the setting, and that the new setting is applied both on the page where it is set and sitewide.

PR Checklist | Your Answer
------------ | -------------


services/QuillLMS/spec/controllers/accessibility_settings_controller_spec.rb
```diff
@@ -0,0 +1,80 @@
...
+  let(:user) { create(:user) }
+
+  before { allow(controller).to receive(:current_user) { user } }
+
+  describe 'GET #show' do
+    context 'when the user has no AccessibilitySetting yet' do
+      it 'returns null snackbar_duration' do
+        get :show, format: :json
+        expect(response).to have_http_status(:ok)
+        body = JSON.parse(response.body)
```
anathomical: This line comes up often enough that I think you would benefit from defining this in a `let` at the top of the file right under `let(:user)`.  `let(:body) { JSON.parse(response.body) }` will not actually call the parse block until the first reference in an `it` block, so this should work seemlessly and you won't need to define `body` in every `it` block.

anathomical (approved):

services/QuillLMS/app/models/accessibility_setting.rb
```diff
@@ -21,4 +21,12 @@ class AccessibilitySetting < ApplicationRecord
...
 
   store_accessor :config, :snackbar_duration
+
+  def snackbar_duration
+    super().present? ? super().to_i : nil
+  end
+
+  def snackbar_duration=(value)
+    super(value.to_i) if value.present?
+  end
```
happythenewsad, line 29: What do you think about this alternative syntax? The benefits I see are:
1. less code (leverage the store_accessor getters and setters that are already provided)
2. more self documenting (we can see the expected default values of json properties at a glance)

```ruby
class AccessibilitySetting < ApplicationRecord
  belongs_to :user

  validates :user_id, presence: true, uniqueness: true
  validates :snackbar_duration, numericality: { greater_than: 0 }

  store_accessor :config, :snackbar_duration

  after_initialize :set_defaults, if: :new_record?

  private def set_defaults
    self.snackbar_duration ||= 1
    # initialize future defaults here.....
  end
end
```

services/QuillLMS/client/app/bundles/Teacher/containers/StudentAccount.tsx
```diff
@@ -44,19 +47,21 @@ const StudentAccount: React.FC<StudentAccountProps> = ({
...
 }) => {
   const [activeSection, setActiveSection] = React.useState<string | null>(null);
   const [email, setEmail] = React.useState<string>(initialEmail);
   const [errors, setErrors] = React.useState<Record<string, string>>({});
   const [firstName, setFirstName] = React.useState<string>(() => name.split(' ')[0]);
   const [lastName, setLastName] = React.useState<string>(() => name.split(' ').slice(1).join(' '));
+  const [snackbarDuration, setSnackbarDuration] = React.useState<number>(initialSnackbarDuration || defaultSnackbarTimeout);
   const [showSnackbar, setShowSnackbar] = React.useState<boolean>(false);
   const [snackbarCopy, setSnackbarCopy] = React.useState<string>('');
   const [timesSubmitted, setTimesSubmitted] = React.useState<number>(0);
```
happythenewsad, line 60: spec question, not code question: why are we tracking this value?

services/QuillLMS/app/models/accessibility_setting.rb
emilia-friedberg: Hey @happythenewsad -- getter and setter methods actually aren't setting defaults (we don't need them, because we already have a default on the frontend, so this value is only getting set if a user actually goes and changes the setting on the my account page). They're just making sure the value is always set and returned as an integer, because otherwise the `store_accessor` stores them as a string. I couldn't figure out a better way to do that, but I am definitely open to suggestions if you have another idea there.

services/QuillLMS/client/app/bundles/Teacher/containers/StudentAccount.tsx
emilia-friedberg: @happythenewsad thanks for asking! We need to allow users to adjust the timeout duration in order to be in compliance with [this WCAG standard](https://www.w3.org/TR/WCAG22/#timing-adjustable). Implementing this meets the success criterion because "the user is allowed to adjust the time limit before encountering it over a wide range that is at least ten times the length of the default setting." Does that make sense?

happythenewsad: Sorry, I meant to reference line 60 - `timesSubmitted`

services/QuillLMS/app/models/accessibility_setting.rb
happythenewsad: Here's a working example without the defaults. 



```ruby
class AccessibilitySetting < ApplicationRecord
  belongs_to :user

  validates :user_id, presence: true, uniqueness: true
  validate :snackbar_duration_must_be_positive_integer

  store_accessor :config, :snackbar_duration

  private

  def snackbar_duration_must_be_positive_integer
    return if snackbar_duration.is_a?(Integer) && snackbar_duration > 0

    errors.add(:snackbar_duration, 'must be a positive integer')
  end
end
```

services/QuillLMS/client/app/bundles/Teacher/containers/StudentAccount.tsx
happythenewsad: EF mentioned this is for better validations on the front end ✅

happythenewsad (approved):
Thanks for updating 🚢

PR #13094, opened 2025-07-02
Only run feedback moderator when triggered rules exists

brendanshean:
## WHAT
Prevent FeedbackModeration from running on a particular `LlmExample` if a feedback error was triggered but no corresponding feedback_rule exists.

## WHY
If there is no feedback_rule in the database, then the Feedback Moderator doesn't have explicit instructions on how to revise the response.

## HOW
Add a guard clause that checks for feedback rules based on the `LlmExample` error codes.

### Screenshots

### Notion Card Links
https://www.notion.so/quill/Evidence-feedback-moderator-not-working-as-intended-223d42e6f941800193a9fe9052fc2f10?source=copy_link

I ran a trial that triggered a feedback moderation error.
I created an active feedback_rule corresponding to that error.
I then retried that same trial and noted that `RevisedLlmExampleBuilder` was run
I made that same feedback_rule inactive and retried the trial, and then confirmed `RevisedLlmExampleBuilder` was not run.

PR Checklist | Your Answer
------------ | -------------


dandrabik (approved):
Good optimization!

emilia-friedberg (approved):

PR #13098, opened 2025-07-07
add new milestone, along with controller and segment tracking for it

emilia-friedberg:
## WHAT
Add a new milestone for `See Onboarding Slides V1', along with a controller action that stores a `user_milestone` when a user sees the slides and Segment tracking so we can view data based on whether a user saw these slides or not.

## WHY
In order to track data about the effectiveness of the onboarding slides.

## HOW
Add a new milestone type to the model file, and a controller action to create a `UserMilestone` join record. Then add tracking via params to the Segment identify call for teachers.

### Screenshots

### Notion Card Links
https://www.notion.so/quill/Onboarding-Slides-Eng-UX-spec-21fd42e6f94180b78072d165a5941de4?source=copy_link#229d42e6f9418066a987ced897911267

Nothing user-facing here, so I just ran the tests.

PR Checklist | Your Answer
------------ | -------------


services/QuillLMS/app/models/milestone.rb
```diff
@@ -28,6 +28,7 @@ class Milestone < ApplicationRecord
     dismiss_grade_level_warning: 'Dismiss Grade Level Warning',
     dismiss_school_selection_reminder: 'Dismiss School Selection Reminder',
     dismiss_teacher_info_modal: 'Dismiss Teacher Info Modal',
-    dismiss_unassign_warning_modal: 'Dismiss Unassign Warning Modal'
+    dismiss_unassign_warning_modal: 'Dismiss Unassign Warning Modal',
```
happythenewsad, line 33: Should `Milestone.name` be unique?

services/QuillLMS/app/models/segment_integration/user.rb
```diff
@@ -73,6 +74,16 @@ def school_params
       end
     end
 
+    def onboarding_group_params
+      return {} if admin? || !teacher?
+
+      onboarding_slides_milestone = Milestone.find_by_name(Milestone::TYPES[:see_onboarding_slides_v1])
+
+      {
+        onboarding_group: user_milestones.find_by(milestone: onboarding_slides_milestone) ? 'Slides V1' : 'Default'
```
happythenewsad: 'Default' should be a constant somewhere. 

Unless there's some segment-specific business logic that requires a bespoke string literal, `'Slides V1'` should be managed by Milestone::Types. Perhaps you can leverage `Milestone::TYPES[:see_onboarding_slides_v1]` directly

emilia-friedberg: @happythenewsad 'Slides V1' was just the requested text here from the spec, which is distinct from the name of the milestone because the milestone itself is an event that determines which group they should be assigned to. I'll make that and DEFAULT a constant at the top of this file, but I don't think Milestone should be concerned about them directly -- let me know if you disagree.

services/QuillLMS/app/models/milestone.rb
emilia-friedberg: Yes, I think so! I'll add a validation to the model.

happythenewsad (approved):
LGTM; thanks for adding the validator!

services/QuillLMS/spec/models/milestone_spec.rb
```diff
@@ -16,9 +16,16 @@
 require 'rails_helper'
 
 describe Milestone, type: :model do
+  before do
```
brendanshean: Let's use `subject { build(:milestone) }` instead of the before block.  This will prevent an unnecessary creation of milestone records in the test.

services/QuillLMS/app/models/segment_integration/user.rb
```diff
@@ -73,6 +78,16 @@ def school_params
       end
     end
 
+    def onboarding_group_params
+      return {} if admin? || !teacher?
+
+      onboarding_slides_milestone = Milestone.find_by_name(Milestone::TYPES[:see_onboarding_slides_v1])
+
+      {
+        onboarding_group: user_milestones.find_by(milestone: onboarding_slides_milestone) ? SLIDES_V1 : DEFAULT
```
brendanshean: `exists?` is more appropriate (and faster) here since the record isn't being used:

user_milestones.exists?(milestone: onboarding_slides_milestone)

services/QuillLMS/app/models/segment_integration/user.rb
```diff
@@ -73,6 +78,16 @@ def school_params
       end
     end
 
+    def onboarding_group_params
+      return {} if admin? || !teacher?
```
brendanshean, line 82: Are there other roles than teacher that we're looking for?

Otherwise, maybe: `return {} unless teacher?`

brendanshean (approved):
Some minor comments, but nothing requiring re-review.

services/QuillLMS/app/models/segment_integration/user.rb
emilia-friedberg: @brendanshean the problem here is that technically, all admins are teachers!

brendanshean: That's right.  My mistake.

anathomical (approved):

PR #13103, opened 2025-07-08
introductory onboarding slides

emilia-friedberg:
## WHAT
Build out the introductory onboarding slides.

## WHY
This is the first part of the Onboarding Tour project. Follow-up PRs will:

- create a user milestone when users visit this page
- add the additional slide sequences as they're ready
- replace the "Select" button at the end of the introductory tour with "Coming Soon" for sequences that are not yet finished
- hook this up to the end of the sign-up flow
- give a user free premium credit when they complete the tour, and redirect them to the dashboard 

## HOW
Lots of HTML and CSS!

### Screenshots
Screenshot 2025-07-08 at 3 12 57 PM
Screenshot 2025-07-08 at 3 13 07 PM
Screenshot 2025-07-08 at 3 13 15 PM
Screenshot 2025-07-08 at 3 13 27 PM
Screenshot 2025-07-08 at 3 13 42 PM

### Notion Card Links
https://www.notion.so/quill/Onboarding-Slides-Eng-UX-spec-21fd42e6f94180b78072d165a5941de4?source=copy_link

Tested locally, having Jack QA on staging

PR Checklist | Your Answer
------------ | -------------


services/QuillLMS/client/app/bundles/Teacher/components/accounts/new/onboarding_tour/slide_pagination_button.tsx
```diff
@@ -0,0 +1,21 @@
...
+const SlidePaginationButton = ({ index, slideIndex, setSlideIndex, }) => {
+  function handleClick() { setSlideIndex(index) }
+
+  const isCurrentSlide = slideIndex === index
+
+  return (
+    <li>
+      <button
+        aria-current={isCurrentSlide ? 'step' : undefined}
+        aria-label={isCurrentSlide ? `Current slide: ${index + 1}` : `Go to slide ${index + 1}`}
```
anathomical, line 12: This is super-minor, but it looks like the two text options here are slightly different: the one on the left includes a ":", while the one on the right doesn't.  Given that one is informational and the other is navigational, that may be intentional, but it felt worth flagging in case we want to use the same pattern for both.

anathomical (approved):
I like the approach you've taken here, and especially love how thorough the testing is on the components.  I could imagine this being abstracted out into a more general carousel component at some point in the future if we use this design pattern more, but it definitely feels like it would be overkill to commit to that right now.

happythenewsad (approved):

services/QuillLMS/client/app/bundles/Teacher/components/accounts/new/onboarding_tour/slide_pagination_button.tsx
emilia-friedberg: Thank you for pointing it out! I just wanted both to read naturally, which I think the current labels do (and they won't voice the colon on a screenreader anyway), but I appreciate you calling this out.

PR #13104, opened 2025-07-09
partition-concept-results-table

anathomical:
## WHAT
Add a migration to create the partition tables
## WHY
So that we can migrate `concept_results` into a partitioned table
## HOW
- Create the partitioned data structure
- Create individual partitions covering dates from the earliest `concept_results` records until 10 years in the future (we'll need some process to add more partitions after that point once we get that far out)

### Notion Card Links

- Run the entire process on staging, including this migration

PR Checklist | Your Answer
------------ | -------------


services/QuillLMS/db/migrate/20250701190404_create_partitioned_concept_results_table.rb
```diff
@@ -0,0 +1,68 @@
...
+  end
+
+  def create_partitions
+    date = START_PARTITION_DATE.dup
+
+    while date < LAST_PARTITION_DATE
+      execute(create_partition_sql(date))
+      date += 6.months
+    end
+  end
```
anathomical, line 56: This will create partitions up through the end of June 2036, giving us a decade.  If we try to insert new records without partitions after this point, the database will choke, rejecting the inserts because there's no valid partition.  Having this happen on July 1, 2036 puts it in the Summer.  Hopefully we don't forget to extend these partitions if necessary, but if we do this should be the safest time for it to happen.

happythenewsad: These terms are acceptable.

happythenewsad (approved):

brendanshean (approved):
It's exciting to see these new partitions

dandrabik (approved):
Looks good, thanks for moving this along!

PR #13106, opened 2025-07-09
partition-concept-results-rename-tables

anathomical:
NOTE: This will need to be deployed and migrated while the application is in maintenance mode

## WHAT
- Swap out the old unpartitioned version of `concept_results` with the new partitioned version
- Update `ConceptResults` model to provide `activity_session_completed_at` when inserting new records because the value is now required to identify which partition to insert the record into
- Update any specs that create `ConceptResults` to be tied to an `ActivitySession` with a set `completed_at` value so that the insertion is valid
- Update the Demo YML file to include the new data structure
## WHY
- We want to actually start using the new partitioned data now that we have it
## HOW
- Rename the old table to `concept_results_old`
- Rename the new partitioned table to `concept_results`
- Add a `before_save` callback to `ConceptResult` to ensure that `activity_session_completed_at` is set
- Run `Demo::SessionData.new.generate_files` while connected to staging in a post-migration state

### Notion Card Links

- Run migration in staging
- Confirm that reports load (comparing to load times in the existing regime)
- Play through an activity, confirm that new `concept_results` are written to the database
- Confirm that the new completions show up in reporting data

PR Checklist | Your Answer
------------ | -------------


services/QuillLMS/spec/controllers/api/v1/activity_sessions_controller_spec.rb
```diff
@@ -28,7 +28,7 @@
     context 'user_activity_classification counts increment when they should' do
       it 'should count_for if the state of the session is "finished"' do
         expect(UserActivityClassification).to receive(:count_for).with(user, activity_classification)
-        put :update, params: { id: activity_session.uid, state: 'finished' }, as: :json
+        put :update, params: { id: activity_session.uid, state: 'finished', completed_at: Time.current }, as: :json
```
brendanshean: Why is Time.current used here and DateTime.current in the others?

services/QuillLMS/spec/controllers/api/v1/activity_sessions_controller_spec.rb
```diff
@@ -151,7 +151,7 @@ def subject
       it 'the active activity session gets deleted' do
         active_activity_session = create(:active_activity_session, uid: activity_session.uid)
 
-        put :update, params: { id: activity_session.uid, state: 'finished' }, as: :json
+        put :update, params: { id: activity_session.uid, state: 'finished', completed_at: Time.current }, as: :json
```
brendanshean: Why is Time.current used here and DateTime.current in the others?

brendanshean (approved):

services/QuillLMS/spec/controllers/api/v1/activity_sessions_controller_spec.rb
anathomical: Great question.  I looked over the file, and the existing code uses a mix of the two.  But that's silly, so I've pushed an update to use `DateTime` on throughout the file to match what we do in most other spec files.

dandrabik (approved):
:+1:

PR #13107, opened 2025-07-09
partition-concept-results-drop-old-table

anathomical:
## WHAT
Add a migration to drop the old un-partitioned concept_results table
## WHY
We don't want to store an old version of this data after we've recording this data in the partitioned table
## HOW
Add a migration to drop the old table

### Notion Card Links

- Ran in staging
- Confirmed that there were no errors
- Confirmed reports still showed expected data
- Confirmed that completion of activities successfully recorded data to the new table and showed up on reports

PR Checklist | Your Answer
------------ | -------------


happythenewsad (approved):

brendanshean (approved):

dandrabik (approved):
Note, as discussed, let's hold onto this for a day just in case. Also, let's make a manual DB backup once this table is no longer used so that we have a final version.

PR #13108, opened 2025-07-09
Evidence - Allow optional use of structured output for LLM API calls

dandrabik:
## WHAT
Add the optional ability to provide an output schema for LLM API calls which is compatible with both Gemini and OpenAI, e.g. pass in ` schema: {optimal: 'boolean', feedback: 'string'}` and you should get structured output in that format.

Note, this isn't turned on for anything in production yet, but testing `Gemini 2.5 Flash Lite` (the newest model) I am getting a lot of `JSON` parsing errors, so I want to test this option for some highlighting code. Also note, for primary feedback, the code worked, but weirdly `Gemini::Chat` API  performs worse on random trials on `suboptimals` for some reason. I don't know why, so turning that off until we can investigate. 

## WHY
This is the recommended and most foolproof way to ensure we get back the correct JSON schema for our API calls. 

## HOW
Pass in an optional variable called schema (a hash of `{ key: type }` into any of the 4 LLM apis, e.g.
```ruby
Evidence::Gemini::Chat.run(
  system_prompt:, 
  entry:, 
  schema: {optimal: 'boolean', feedback: 'string'}
)
```

I deployed this to my staging env, and ran experiments with each of the 4 APIs for primary feedback and saw no errors. I did a bunch on anecdotal testing on in the console to confirm this is sending correct info and receiving valid responses. I then removed that code so the option exists, but it isn't 'ON' for any production path.

PR Checklist | Your Answer
------------ | -------------
Have you deployed to Staging? |YES 


services/QuillLMS/engines/evidence/app/models/evidence/research/gen_ai/llm_prompt.rb
```diff
@@ -92,10 +92,11 @@ def prompt_with_rag_label_examples_and_student_response(entry:, prompt_id:, labe
           "#{prompt}\n\n#{rag}\n\n{student_response: #{entry}}\nProvide feedback in the following JSON format: #{CLASSIFICATION_SCHEMA}"
         end
 
-        def prompt_with_format(student_response: nil)
+        def prompt_with_addendums(student_response: nil, format_instructions: true)
           student_response_section = student_response.nil? ? nil : "{student_response: #{student_response}}"
+          format_instructions = format_instructions ? FEEDBACK_FORMAT_INSTRUCTION : nil
```
anathomical: Using the same name as the method param is a little confusing here, especially since it may look like we're doing a pass-through if someone missed this line and just saw `def` and the final line.

anathomical (approved):
A minor comment on naming, but this looks good.

services/QuillLMS/engines/evidence/app/models/evidence/research/gen_ai/llm_prompt.rb
brendanshean: I agree.  I was surprised rubocop doesn't flag this, but I guess we don't have the cop turned on: https://www.rubydoc.info/gems/rubocop/RuboCop/Cop/Lint/ShadowedArgument

brendanshean (approved):
Thanks for adding in this configuration.

services/QuillLMS/engines/evidence/app/models/evidence/research/gen_ai/llm_prompt.rb
dandrabik: Oh, good catch, I'll fix this.

PR #13109, opened 2025-07-09
Assign practice activity bug

brendanshean:
## WHAT
Fix a bug where some users are unable to assign recommended activity packs through.

## WHY
If a user has an existing unit with the same name but from a different unit template (e.g. a recently renamed diagnostic), the flow will fail with a generic error message, preventing the user from creating new activity packs.

## HOW
Check to see if the user has an existing units with the same name.  If so, append the current date to the unit name to get a new name that should be unique (there's an edge case where a unit with the date appended also exists, but we'll ignore that for simplicity).

### Screenshots

### Notion Card Links
https://www.notion.so/quill/Teacher-sees-an-error-message-when-they-attempt-to-assign-the-practice-recommendation-activity-packs-22ad42e6f94180fa9865c307a2e08220?source=copy_link

I have checked this on staging.

PR Checklist | Your Answer
------------ | -------------


services/QuillLMS/app/services/independent_practice_packs_assigner.rb
```diff
@@ -68,10 +68,19 @@ def run
...
       selections
         .map { |selection| UnitTemplate.find(selection['id']) }
-        .each { |ut| Unit.unscoped.find_or_create_by!(unit_template: ut, name: ut.name, user:) }
+        .each { |unit_template| find_or_create_activity_pack(unit_template) }
     end
   end
 
+  private def find_or_create_activity_pack(unit_template)
+    name = unit_template.name
+    return if Unit.unscoped.exists?(unit_template:, name:, user:)
```
anathomical, line 77: If there's an existing record, will this actually result in `nil` being returned instead of the existing record?

anathomical (approved):
I think that the general approach here makes sense, but I've flagged a potential issue in the case where there's an existing record to be fetched and nothing needs to be created.

services/QuillLMS/app/services/independent_practice_packs_assigner.rb
brendanshean: The return type of `find_or_create_activity_pack` doesn't matter here since we're using `each` here instead of `map`.

brendanshean: But, the naming is confusing since it's different than the regular find_or_create so I will give it a better name.

anathomical: In that case, should we consider renaming the method `create_if_missing` or something like that to make it clear what the intent of the method is?

brendanshean: jinx

PR #13112, opened 2025-07-10
Evidence - Store original feedback when we alter it.

dandrabik:
## WHAT
Store the original feedback in `feedback_histories` when we either 1) revise the primary feedback or 2) replace it with secondary feedback.
## WHY
This will make analysis of the steps after the primary feedback easier, i.e. revisions, repeated feedback, and secondary feedback. Right now it's difficult to analyze the performance of the repeated feedback check in particular since we aren't storing the data that the `RepeatedFeedbackChecker` was passed in.
## HOW
Populate an `original_feedback` in the `Response` objects of `PrimaryFeedback` and `RevisedFeedback`. That's enough info to determine if we should store it in `feedback_history.metadata.api.original_feedback`

I played a few activities on staging and verified that it stored the data in the appropriate place in the feedback_history table, but only when the primary feedback was revised. See screenshot

Screenshot 2025-07-10 at 4 30 17 PM

PR Checklist | Your Answer
------------ | -------------


brendanshean (approved):
This seems like very useful information to store.

PR #13116, opened 2025-07-11
Fix assign practice activity bug

brendanshean:
## WHAT
Fix a bug where some users are unable to assign recommended activity packs through.  This is a second attempt from a previous an earlier PR: https://github.com/empirical-org/Empirical-Core/pull/13109

## WHY
The current code appends a date to the unit if there's a name collision, but doesn't check to see that that new name is also unique.

## HOW
Add code that loops through until it finds a unique name, adding a numbered suffix for duplicate dates. Also, to prevent an infinite loop, stop at 100 and then just add a timestamp from then on.

### Screenshots

### Notion Card Links
https://www.notion.so/quill/Teacher-sees-an-error-message-when-they-attempt-to-assign-the-practice-recommendation-activity-packs-22dd42e6f94180eb93cac768b550f515?source=copy_link

I tested this fix on the user that was experiencing the issue and it worked.

PR Checklist | Your Answer
------------ | -------------
Have you deployed to Staging? | No.


anathomical (approved):
This seems reasonable to me.  Also, this looks somewhat similar to the code in `app/lib/class_code_generator.rb` to handle cases where we run out of unique class-code combinations.  I don't think it's worth consolidating the logic now, but it's worth keeping in mind in case we end up needing to use something like this in a third place.

PR #13121, opened 2025-07-14
activity analysis index design updates

emilia-friedberg:
## WHAT
Update the design of the Activity Analysis index page. In the process of doing this, I realized that there was a lot of old and disorganized code in this page, so I effectively just rebuilt its components and deleted some old and unused files in the process. I also made the following broader changes for accessibility/usability reasons:

- when rendering a DropdownInput with custom HTML options (like we use for options that include images), pass an optional `labelToShow` prop to the option, so that screenreaders don't hear the full HTML as a label.
- update our `useScaledTableColumnWidths` React hook to accept an optional `additionalWidthRestrictions`, which defaults to 0, so that we can account for borders/padding/additional restrictions on the containing component that limit the space the table can actually take up.

## WHY
We're updating all of our reports pages to make sure we're using up-to-date design language and frontend best practices.

## HOW
See "What"

### Screenshots
Screenshot 2025-07-14 at 4 35 39 PM

### Notion Card Links
https://www.notion.so/quill/Reports-Design-Updates-Activity-Analysis-Index-229d42e6f941803b8180c8329edab74a?source=copy_link

Tested locally, having Jack QA on staging.

PR Checklist | Your Answer
------------ | -------------


brendanshean (approved):
Thanks for cleanup along with the design update!

services/QuillLMS/client/app/bundles/Shared/hooks/__tests__/useScaledTableColumnWidths.test.tsx
```diff
@@ -103,4 +103,51 @@ describe('useScaledTableColumnWidths', () => {
...
+        columnWidths: [
+          { width: 300 },
+          { width: 300 },
+          { width: 300 },
+          { width: 300 },
+        ],
+        containerRef: ref,
+        maxTableWidth: 1200,
+        minTableWidth: 963,
+      })
```
anathomical: The difference between this big object and the one in the test below is pretty subtle.  I wonder if there's a way to define the base object at a higher level, and then add the `additionalWidthRestrictions` key by itself inside this case so it's obvious what the difference between the two cases is.

anathomical (approved):

services/QuillLMS/client/app/bundles/Teacher/components/progress_reports/activity_analysis/activity_pack.tsx
```diff
@@ -0,0 +1,201 @@
...
+    <div className="assigned-to">
+      <span>
+        Assigned to {classrooms.length} {Pluralize('class', classrooms.length)}:
+      </span>
+      <ul>
+        {classrooms.map((c, i) => (
+          <li key={i}>
+            <b>{c.name}</b>
+            <span>
+              ({c.assignedStudentCount}/{c.totalStudentCount}{' '}
```
happythenewsad, line 98: This will print `Infinity` if totalStudentCount is zero.

happythenewsad (approved):
nice and clean

services/QuillLMS/client/app/bundles/Teacher/components/progress_reports/activity_analysis/activity_pack.tsx
emilia-friedberg: @happythenewsad It actually isn't doing the calculation here, it's just rendering the counts as a string with the `/` in between (it would be doing the calculation if it were `{c.assignedStudentCount/c.totalStudentCount}`. So it would render `0/0`, which is admittedly weird, but that's not a state the data should be in anyway.

PR #13124, opened 2025-07-15
refactor out hash routing for reports

emilia-friedberg:
## WHAT
Replace the `HashRouter` wrapper for the diagnostic_reports links with a `BrowserRouter`. I've maintained support for old links by adding a redirect for hash routes, but also updated everywhere I could find within our codebase to remove the hash.

## WHY
While updating the styling for the Activity Analysis and sub-pages, I was testing tab navigation, and I noticed that our "Skip to" links available for keyboard navigators actually break these pages, because they change the path using a hashlink and these pages have hash routing. Rather than apply custom code not to show these links on these pages, I think it's better to replace the `HashRouter` with a `BrowserRouter`, since there's no good reason for the former in the first place.

## HOW
See "What".

Clicked around various links to and between these pages and confirmed that everything works as expected.

PR Checklist | Your Answer
------------ | -------------


happythenewsad (approved):
Nice cleanup! Consider adding a slack reminder in a year or so, to remove the backwards compatibility code

services/QuillLMS/client/app/bundles/Teacher/containers/DiagnosticReports.jsx
```diff
@@ -1,22 +1,51 @@
...
+const DiagnosticReports = ({ show_lessons_banner }) => {
+  React.useEffect(() => {
+    const { pathname, hash } = window.location;
+
+    if (hash.startsWith('#/')) {
+      const cleanHash = hash.slice(1); // Remove the '#' character
+      const newUrl = `${pathname.replace(/\/$/, '')}${cleanHash}`;
+      window.history.replaceState(null, '', newUrl);
+    }
+  }, []);
```
anathomical: Is this effect just duplicating the effects of `HashRedirect`, or are there cases where one would catch an old-style URL but the other wouldn't?  If they're catching different edge cases, that seems reasonable, but if they're just doing the same thing it'd probably be nice to consolidate to a single option.

anathomical (approved):
I think that this is a smart choice.  Removing hash routing makes a ton of sense to solve this problem, and makes our links cleaner, to boot.  I like the call to transition here.

services/QuillLMS/client/app/bundles/Teacher/containers/DiagnosticReports.jsx
emilia-friedberg: Whoops! This was an older version I meant to delete, thank you for flagging.

PR #13127, opened 2025-07-16
skip onboarding slides if user joined premium school

emilia-friedberg:
## WHAT
Skip the onboarding slides if the user joined a premium school.

## WHY
From the Notion card:

1. The Partner Success and Professional Learning teams will use the onboarding content in their trainings. Forcing these teachers through the slides adds friction to their trainings. Erika: “We will instruct them to click through the 14 slides quickly”
2. Currently the flow is forced. There is no option to exit. Because of that, every teacher is rewarded with one month of premium. That creates confusion for teachers who are already at Premium schools. Therefore, the **onboarding slide experience should be limited to just free teachers.**

## HOW
Pass the user's current subscription into the router component (which will exist if they have joined a premium school), and use its presence to determine whether to pass the user to the onboarding tour or directly to the dashboard.

Note for @happythenewsad -- we'll need to integrate this change into the Posthog PR, since it touches the same logic. Let me know if you need a hand. 

### Screenshots

### Notion Card Links
https://www.notion.so/quill/Onboarding-Slides-Add-logic-to-remove-Premium-teachers-from-flow-231d42e6f941808cbb35dc986795e954?source=copy_link

Tested locally and confirmed that users who join a premium school do not go to the onboarding slides.

PR Checklist | Your Answer
------------ | -------------


services/QuillLMS/client/app/bundles/Teacher/components/accounts/new/shared.ts
```diff
@@ -0,0 +1,3 @@
+export function redirectTo(path: string) {
```
anathomical: I appreciate the decision to generalize the `redirectTo` function into a shared space.  I wonder if it's general enough that it would make sense to put it in `client/app/bundles/Shared` in `lib` or `utils` or something since I could imagine it being used in other component paths, too.

anathomical (approved):
I really like the new `add_teacher_info` tests.  Especially the backfill for existing behavior on top of writing conditions for the new code.

happythenewsad (approved):

PR #13128, opened 2025-07-16
Evidence - Update Repeated Feedback checks with new Gemini model and deterministic checks.

dandrabik:
## WHAT
Add testing harness and update Repeated Feedback algorithm to improve its performance using a new prompt and some algorithmic checks.

There is also some secondary updates:
- Updated to Gemini 2.0 flash lite
- Created a new repeated_feedback test set
- Added better JSON parsing
- Logged repeated_feeback metadata in feedback_history

## WHY

We want to improve on the performance of this algorithm based on feedback from our user testing. Note this only improves test performance slightly (89.5% -> 90.5%), but I am shipping it now because there are a bunch of other benefits:

- it gets us off an about-to-be-deprecated API
- it should make the call much faster for in a lot of cases (and not require an LLM call)
- it makes it less error prone by fixing common JSON errors
- it logs more so we can have better analysis. 

Note, I'm somewhat hitting a ceiling at 90%, but I believe some of the test set is mislabeled (I had to regenerate some of the data since we don't log original feedback), so I suspect the accuracy number is higher in reality.

## HOW
The main work was creating a test set with the Curriculum Team and then a testing harness in a thor script. Ran lot of tests and iterated on the algorithm and prompt. 
 
### Screenshots
Screenshot 2025-07-16 at 2 03 08 PM

### Notion Card Links
https://www.notion.so/quill/Repeated-Feedback-Experiments-22ad42e6f94180149d89d81316b02565

I've run 100s of experiments to test the algorithm. I've played through a few activities on staging and verified that it is flagging repeated when expected and when not expected.

PR Checklist | Your Answer
------------ | -------------


services/QuillLMS/lib/tasks/gen_ai_repeated_feedback.thor
```diff
@@ -0,0 +1,198 @@
...
+
+  Result = Data.define(:test_data, :marked_as_repeat, :correct, :reason)
+
+  # bundle exec thor gen_ai_repeated_feedback:baseline_test 40
+  desc 'baseline_test limit', 'Test a number or entries from the test.csv file'
+  def baseline_test(limit = 500, threads = 10)
+    test_set = all_test_data.first(limit.to_i)
+
+    # Parallel version
+    results = Parallel.map(test_set, in_processes: threads) do |data|
```
brendanshean, line 34: Cool use of `Parallel`

services/QuillLMS/engines/evidence/app/services/evidence/gen_ai/repeated_feedback/prompt_builder.rb
```diff
@@ -3,14 +3,39 @@
...
   module GenAi
     module RepeatedFeedback
-      class PromptBuilder < Evidence::GenAi::PromptBuilder
-        TEMPLATE_FOLDER = 'app/services/evidence/gen_ai/repeated_feedback/prompts/'
-        DEFAULT_TEMPLATE = '2024_08_28_gemini2.md'
+      class PromptBuilder < ApplicationService
+        include Evidence::MarkdownFormatter
 
-        private def template_variables = { feedback_history: }
+        # Copied from prompts/2025_07_14_gemini-flash-2.0-2.md
```
brendanshean, line 9: I'm curious why you switched from TEMPLATE_FOLDER and DEFAULT_TEMPLATE pattern here to hard coding within class.

brendanshean (approved):
Nice work on this.  While a tiny bump in accuracy, I think there's a a lot of other good stuff in the PR.

services/QuillLMS/engines/evidence/app/services/evidence/gen_ai/repeated_feedback/prompt_builder.rb
dandrabik: I had two main reasons (these are somewhat minor, and I may revisit in the future):
1. As the prompt is in a final-ish state, it feels like a small optimization to not have to open a file and read the contents only to get a few lines every time this is called in production.
2. I wanted to move this off of the class inheritance since, it doesn't use a lot of the parent.

PR #13131, opened 2025-07-17
Classlink email notifications

happythenewsad:
## WHAT
- send Classlink account creation email notifications 
## WHY
- part of project spec 

## HOW
- Schedule a Sidekiq worker in ClasslinkAccount's after_commit lifecycle hook 

### Screenshots

### Notion Card Links
https://www.notion.so/quill/post-V1-ClassLink-automated-emails-21cd42e6f9418075bcb3cea877510b65

(staging)
- [x] Manually create a ClasslinkAccount
- [x] Observe in the Sidekiq dash that the correct job is scheduled for 8am next day
- [x] trigger the mailer directly, observe the properly formatted email in inbox

PR Checklist | Your Answer
------------ | -------------


dandrabik (approved):

services/QuillLMS/app/models/classlink_account.rb
```diff
@@ -24,4 +24,16 @@
...
+
+  after_commit :send_classlink_account_created_email, on: :create
+
+  private def send_classlink_account_created_email
+    est_time = Time.current.in_time_zone('Eastern Time (US & Canada)')
+    next_8am_est = if est_time.hour < 8
+                     est_time.change(hour: 8, min: 0, sec: 0)
+                   else
+                     est_time.tomorrow.change(hour: 8, min: 0, sec: 0)
+                   end
```
anathomical, line 36: I imagine that this timing is just part of the spec, but are we concerned about this not coming early enough in a day?  If a teacher tries to log in before 8, they may be confused.  I imagine this would only be an issue if they had an existing email-based (or Google or whatever) account that's had its provider updated to ClassLink, and suddenly they can't log in and don't know why.  Not necessarily an issue, but I figured it was worth flagging to make sure we'd considered it.

anathomical (approved):
I like the use of the after_commit hook here.

services/QuillLMS/app/models/classlink_account.rb
happythenewsad: That's a legitimate concern. There was a product discussion and this was Partnerships' specific preference.

PR #13135, opened 2025-07-21
default to old diagnostic if no data for newer one in performance and practice insights

emilia-friedberg:
## WHAT
Add logic in the Performance Insights and Practice Insights index files to request data for the 2024 Starter Diagnostic if the user loads the page and has no data for the 2025 version (the state for most teachers to start the year). If there is data for 2024, make that the default diagnostic instead.

## WHY
Peter Sharkey requested this change so that most teachers starting the year don't see an empty state on this page.

## HOW
Track whether the diagnostic has been changed from the default at the top-level index. If it hasn't, when we request data and it's blank, we make another request for last year's data. If that's also blank, we don't do anything; if it isn't, we change the default to show that data instead.

### Screenshots

### Notion Card Links
https://www.notion.so/quill/Diagnostic-Insights-Change-default-selection-logic-for-Performance-Insights-and-Practice-Insights-t-231d42e6f94180f1badac0eaa8a4f40f?source=copy_link

Tested locally, having Peter Sharkey QA on staging.

PR Checklist | Your Answer
------------ | -------------


happythenewsad (approved):

services/QuillLMS/client/app/bundles/Teacher/components/progress_reports/diagnostic_insights/practice_insights/index.tsx
```diff
@@ -108,28 +109,45 @@ const PracticeInsights = ({ openMobileSidebar, classroom, selectedDiagnostic, se
...
+      if (classroomIdRef.current !== requestedClassroomId) { return }
+
+      if (callback) { callback(body)}
+
+      // using ref here guarantees other diagnostic levels don't get overwritten, regardless of when the requests fire and return
+      setStudentPracticesByDiagnosticType({...studentPracticesByDiagnosticRef.current, [selectedDiagnosticId]: body});
     });
   }
 
+  function getDataFor2024DiagnosticIfNoDataForThisYear(students) {
```
brendanshean: Naming a method with a year 2024 in it feels like a code smell.  If this is a one-off that we won't have to do again, probably fine to leave as-is.  However, if this is a pattern we will use in the future, you could instead make method generic and then set the default diagnostic via a parameter and constant.

brendanshean (approved):

services/QuillLMS/client/app/bundles/Teacher/components/progress_reports/diagnostic_insights/shared.tsx
```diff
@@ -67,6 +67,11 @@ export const timeframeOptions = [
   },
 ];
 
+export const OLD_2024_STARTER_DIAGNOSTIC_OPTION = {
```
anathomical: Building on Brendan's comment about function naming, it might also make sense to give this constant a more generic name so that we can just update it here somehow next year if we do another update.

anathomical (approved):
This feels like a reasonable approach.  I especially appreciate the thorough test coverage to make sure that we don't accidentally break this at some point in the future.  I think Brendan's call-out that we might have to do this again next year is a reasonable point even if I'm not sure how likely that is for the Starter (our default).

One product thing worth considering: next year, after we've had the "new" diagnotic in place for 12 months, would we still want to do this fallback?  We'd basically be moving teachers to two-years-ago.  If that's the desired behavior, I wonder if we actually want to go back farther if the data is available.  (Like, if there's no 2024 data, go back to 2023, etc.)  That would suggest that the logic might be "Keep going back on Starters until you find one with data".

Not sure if we want that, and this may be beyond the scope of this PR and maybe we should just merge this now and consider the other broader issue as a potential follow-on.

emilia-friedberg:
@brendanshean @anathomical I've implemented your naming suggestions here --  I think it's likely that we won't want to keep this functionality next school year unless we introduce another version of the starter diagnostic, which Peter Sharkey notes in his Notion card. I've added a Notion reminder for us next July 1st to check in about whether or not we should remove or update this logic.

PR #13138, opened 2025-07-21
automatically verify classlink and clever teachers

emilia-friedberg:
## WHAT
Automatically verify Clever Secure Sync and ClassLink teachers so they can use Diagnostic Insights.

## WHY
We want to remove friction for these teachers, since their integrations are enough to identify them as legitimately belonging to the school.

## HOW
Add logic to the `classlink_controller` and Clever `district_teacher_integration` files to create the verification record, then fire events to Segment. Also slightly modify `teacher_infos` to allow for a `nil` reviewer_id if the associated user is a Clever or ClassLink user, and pass the data up to the frontend so we can display it on admin dashboards.

### Screenshots

### Notion Card Links
https://www.notion.so/quill/Diagnostic-Insights-Automatic-Teacher-Verification-for-Clever-Secure-Sync-and-ClassLink-Teachers-22cd42e6f94180348083dc449a51ff33#22cd42e6f94180c8b1c4eb54f773147c

Nattalie and I tested on staging that this works as expected for both ClassLink and Clever Secure Sync users.

PR Checklist | Your Answer
------------ | -------------


services/QuillLMS/app/controllers/auth/classlink_controller.rb
```diff
@@ -46,5 +49,12 @@ def classlink
     private def user
       auth_credential&.user
     end
+
+    private def verify_teacher_if_necessary
+      return if user.teacher_verification_status == TeacherInfo::APPROVED || user.admin_approval_status == AdminInfo::APPROVED
```
happythenewsad: I think we want to keep teacher auto-approval and admin auto-approval in a single code location. Here's the admin auto-approval code: https://github.com/empirical-org/Empirical-Core/blob/develop/services/QuillLMS/app/services/classlink_integration/instructor_linker.rb#L32

Note that the admin approval code happens at Rostering time. It may make sense to move admin- and teacher- auto approvals to sign-in time, given the product ask.

happythenewsad (commented):
This code may no longer be necessary with this new change, although it updates slightly different properties: https://github.com/empirical-org/Empirical-Core/blob/develop/services/QuillLMS/app/services/classlink_integration/instructor_linker.rb#L32

Let's discuss

services/QuillLMS/app/controllers/auth/classlink_controller.rb
emilia-friedberg: Ah, I didn't know that existed. I agree they should be in the same place. I'm going to check in with Nattalie and confirm that it's okay if this happens at rostering instead of sign-in.

services/QuillLMS/spec/controllers/auth/classlink_controller_spec.rb
```diff
@@ -22,8 +22,8 @@ def setup_conflicting_user_response(user)
     subject { get :classlink, params: { code: } }
 
     before do
-      allow(ClasslinkIntegration::AuthCredentialSaver).to receive(:run)
-        .and_return(auth_credential)
+      allow(ClasslinkIntegration::AuthCredentialSaver).to receive(:run).and_return(auth_credential)
+      allow(controller).to receive(:sign_in)
```
happythenewsad: This change shouldn't be necessary since there were only whitespace edits to `ClasslinkController`

happythenewsad (approved):

services/QuillLMS/app/services/classlink_integration/instructor_linker.rb
```diff
@@ -39,6 +42,13 @@ def run
       )
     end
 
+    private def auto_verify_teacher
+      return if persisted_user.teacher_verification_status == TeacherInfo::APPROVED
```
brendanshean: Clever has a disjunct "teacher.admin_approval_status == AdminInfo::APPROVED".  Does this not apply to Classlink?  Maybe a question for @happythenewsad

services/QuillLMS/app/services/clever_integration/district_teacher_integration.rb
```diff
@@ -43,5 +44,12 @@ def run
         user: teacher
       )
     end
+
+    private def auto_verify_teacher
+      return if !SchoolsUsers.exists?(user: teacher) || teacher.teacher_verification_status == TeacherInfo::APPROVED || teacher.admin_approval_status == AdminInfo::APPROVED
```
brendanshean: I feel like these methods are violating law of demeter since they're reaching into teacher which then reaches into teacher_info.  They're also long to read and I think could benefit from a helper:

teacher.admin_info_approved?  or  teacher.teacher_verified? or whatever makes sense.

brendanshean (approved):
Nice work on figuring out this verification!
Some minor comments, but nothing holding up re-review.

services/QuillLMS/app/services/classlink_integration/instructor_linker.rb
emilia-friedberg: @brendanshean In the ClassLink flow, we are already checking to see if the user is an admin before running this branching logic, and the admin path is going to give them an approved `AdminInfo` record instead. Here, the existing logic doesn't handle admins (on the Quill side), so I just didn't want to create extra records if it's not necessary. Similarly, this code won't run here if the user isn't associated with a school, but that's not true for the Clever path, so I'm checking that there as well.

PR #13134, opened 2025-07-21
Dataset Editor v1.1 - Part 1 & 2

brendanshean:
## WHAT
The Dataset Annotation Tool has five main sections, the first three allow for configuration of the LLM Prompt.
1) LLM Config - specify llm, prompt template, and temperature
2) Guidelines - create, show, hide optimal/suboptimal guidelines
3) Prompt Examples - show/hide prompt examples
4) Clusters - organizing the different test and prompt examples; apply the same feedback for these groupings
5) Annotations table - add/delete examples, change prompt to test and vice versa, edit feedback, edit optimal/suboptimal, edit cluster assignment

Finally, there's a button 'Run Trial' that takes the current configuration and allows you to run trials within the page.

## WHY
This editor is attempt to provide tooling to speed up the cold start process.

## HOW
Build out the necessary components.
1) llmConfig

2) guidelines
  guidelinesSection.tsx
  guidelineTable.tsx
  guidelineRow.tsx

3) promptExamples
  promptExamplesSection.tsx
  promptExamplesList.tsx
  promptExampleItem.tsx

4) clusters
  clustersSection.tsx
  clusterTable.tsx
  clusterRow.tsx

5) annotations
  annotationRow.tsx
  annotationsTable.tsx
  tableFilters.tsx

Most of the components involve keeping track of edits to persisted objects as well as requests for new objects.  Part 2 will allow you to save these edits and re-run new trials with the updated configuration.

### Screenshots
Screenshot 2025-07-21 at 7 46 06 PM

### Notion Card Links
https://www.notion.so/quill/V1-1-Dataset-Editor-225d42e6f9418030b850e5c8dc264e8d?source=copy_link

### What have you done to QA this feature?
I have tested the table out and its various features locally.  Part 2 will implement 'save progress' at which point I will test more thoroughly on staging.

PR Checklist | Your Answer
------------ | -------------
Have you added and/or updated tests? |  YES
Have you deployed to Staging? |  YES
Self-Review: Have you done an initial self-review of the code below on Github? | YES

PR #13144, opened 2025-07-23
Dataset Editor v1.1 - Part 2

brendanshean:
## WHAT
* Frontend: Continue [Dataset Editor v1.1](https://github.com/empirical-org/Empirical-Core/pull/13134) rollout
  1. Add hooks for keeping track of state for 'Save progress' functionality
  1. Change annotation table to examples table  
  1. AnnotationTool renamed DatasetEditor language;   
* Backend: Introduces the dataset editor service layer 
  1. change detection/saving, 
  1. major/minor version ops, 
  1. copy/update helpers

## WHY
* Frontend: 
  1. State needs to be tracked so major / minor version can be determined
  1. Examples is more accurate as table is composed of TestExamples and PromptExamples
  1. Dataset Editor is the name in the spec
* Backend 
  1. Changes need to be calculated before persistence 
  1. Clean major vs. minor version semantics will help with tracking
  1. Copy is used in major version, update is used in minor version

## HOW
* Frontend
  1. Hooks
    a. useDatasetEditor: main orchestrator
    a. useLocalChanges: keep track of various states in the DatasetEditor, to calculate isDirty
    a. useTrialManager: handles running of trials
  1. Rename a bunch of things
  1. Rename a bunch of things
* Backend
  i-iii. Add various Service Object to determine changes and then persist them via major version (copiers) or minor version (updaters).  Service objects receive JSON in the form:

```
export interface LocalChanges {
  guidelines: {
    created: Guideline[];
    updated: Record<string, Partial<Guideline>>;
  };
  clusters: {
    created: Cluster[];
    updated: Record<string, Partial<Cluster>>;
  };
  testExamples: {
    created: Example[];
    updated: Record<string, Partial<Example>>;
    deleted: string[];
  };
  promptExamples: {
    created: Example[];
    updated: Record<string, Partial<Example>>;
    deleted: string[];
  };
}
```

ServiceObject call chain is roughly:

```
ChangesSaver
  ChangesDetector
    MajorVersionDatasetCreator
      ClustersCopier
      ExamplesCopier
      DatasetRelevantTextsCopier
      GuidelinesCopier
    MinorVersionDatasetUpdater
      ClustersUpdater
      GuidelinesUpdater
      PromptExampleUpdater
```

### Screenshots

### Notion Card Links
https://www.notion.so/quill/V1-1-Dataset-Editor-225d42e6f9418030b850e5c8dc264e8d?source=copy_link

I have tested this locally and will be doing more testing on staging.

PR Checklist | Your Answer
------------ | -------------


services/QuillLMS/client/app/bundles/Shared/styles/snackbar.scss
```diff
@@ -7,7 +7,7 @@
   text-align: center;
   padding: 13px 16px;
   position: fixed;
-  z-index: 1;
+  z-index: 9999;
```
brendanshean, line 10: The snackbar was being hidden behind the Clusters Table.  Not sure if there's a better way to handle this.

services/QuillLMS/client/app/bundles/Staff/components/evidence/datasetEditor/components/examples/__tests__/exampleFilters.test.ts
```diff
@@ -0,0 +1,230 @@
...
+    clusters: [UNASSIGNED, '1', '2', '3'],
+    clusterNames: [UNASSIGNED, 'Introduction', 'Body Paragraph', 'Cluster 3'],
+    curriculumOptimal: [OPTIMAL, SUBOPTIMAL],
+    usage: [USAGE_TEST, USAGE_PROMPT],
+    responseSearch: '',
+    feedbackSearch: '',
+    ...overrides
+  });
+
+  describe('cluster filtering', () => {
```
dandrabik, line 47: Nice tests!

services/QuillLMS/engines/evidence/app/controllers/evidence/research/gen_ai/datasets_controller.rb
```diff
@@ -84,15 +84,33 @@ def annotation_tool_data
...
+            if result[:new_dataset_id]
+              render json: { success: true, new_dataset_id: result[:new_dataset_id], version_type: result[:version_type] }
+            else
+              render json: { success: true }
+            end
+          else
+            render json: { success: false, error: result[:error] }, status: :unprocessable_entity
+          end
+        rescue StandardError => e
+          Rails.logger.error "Error saving changes: #{e.message}"
```
dandrabik: Probably should don't need this in prod. Maybe use `ErrorNotifier` if you need it.

services/QuillLMS/engines/evidence/app/services/evidence/research/gen_ai/dataset_editor/changes_detector.rb
```diff
@@ -0,0 +1,47 @@
...
+          attr_reader :dataset, :changes, :klass, :version_type
+
+          def initialize(dataset:, changes:, klass:, version_type:)
+            @dataset = dataset
+            @changes = changes
+            @klass = klass
+            @version_type = version_type
+          end
+
+          def run = !!deleted_changes? || !!created_changes? || !!updated_changes?
```
dandrabik, line 17: Maybe move the !! only to where it's needed (explicit booleans are returned in a lot of places) to make this more readable, i.e. `deleted_changes? || created_changes? || updated_changes?`

dandrabik (approved):
Nice work on this. I don't see anything glaring here that needs re-working, so seems good to ship (I'm skimming a lot of the frontend code). 

I think one think we should keep an eye on is the size of tables as they grow from initial use. I think it should be fine, but if saving ends up being relatively rapid, those tables could theoretically grow very fast (probably better to just observe it to start).

brendanshean:
> Nice work on this. I don't see anything glaring here that needs re-working, so seems good to ship (I'm skimming a lot of the frontend code).
> 
> I think one think we should keep an eye on is the size of tables as they grow from initial use. I think it should be fine, but if saving ends up being relatively rapid, those tables could theoretically grow very fast (probably better to just observe it to start).

Originally, I had the minor versions creating new datasets with associated records, which would have been even more records.  Then I noticed that the spec called for all trials within a minor version to be visible which suggested not creating new datasets.

services/QuillLMS/client/app/bundles/Shared/styles/snackbar.scss
emilia-friedberg: This seems fine to me. I don't think we'd ever want our snackbar below a different element.

services/QuillLMS/client/app/bundles/Staff/components/evidence/datasetEditor/components/examples/exampleRow.tsx
```diff
@@ -32,40 +44,31 @@ const renderClusterOptions = (clusters: Cluster[], showNames: boolean = false) =
...
+          {trialResult.llm_feedback}
         </div>
-      ) : trialResult.status === 'error' ? (
-        <span className="trial-error">Error</span>
+      ) : trialResult.status === TRIAL_STATUS_ERROR ? (
+        <span className="trial-error">{TRIAL_ERROR}</span>
       ) : (
-        <span className="trial-pending">Pending</span>
+        <span className="trial-pending">{TRIAL_PENDING}</span>
       )}
```
emilia-friedberg, line 65: I don't love the nested ternary thing, but this one is pretty readable.

emilia-friedberg (approved):
Mostly skimmed the frontend stuff here -- I like the approach of moving a lot of the logic out into shared hooks. I still find some of the updating functions difficult to read, which could potentially be addressed with less-nested state, but I think just having all the related logic in one, separate file helps with that (kind of like the action file pattern in Redux). Overall LGTM!

PR #13148, opened 2025-07-26
Classlink School rostering improvements

happythenewsad:
## WHAT
- simplifies Classlink Rostering Sync code given a new hard assumption: `School` records must exist in LMS, along with premium subscriptions, before sync. Incoming API schools not meeting this criterion will be ignored. 
- don't allow the Sync to update LMS `School.name` 
- remove circular dependency where Schools would not get `ClasslinkInstanceSchools` unless `ClasslinkInstanceSchool` records already existed for them 

## WHY
- we want to remove unneeded complexity from this flow
- Names and NCES id status will be managed by Partnerships for Classlink syncs 
- bug  

### Screenshots

### Notion Card Links

- [x] Create a test school ("ClassLink HS") and district. Attach a premium subscription 
- [x] Ensure the district name matches the one in the Classlink portal ("Dev: ClassLink Test District")
- [x] Run the sync
- [x] check that the School now has a ClasslinkInstanceSchool
- [x] Run the sync again 
- [x] ensure no unexpected errors, incl. in Sentry
- [x] confirm that all teachers associated with the school in the Partner Portal have been synced to LMS

PR Checklist | Your Answer
------------ | -------------


services/QuillLMS/app/services/classlink_integration/district_importer.rb
```diff
@@ -6,14 +6,14 @@ class DistrictImporter < ApplicationService
...
-      :premium_school_external_ids
+      :district
 
     def initialize(district_oneroster_id:, district_bearer_token:, district_id:, tenant_id:)
       @district_oneroster_id = district_oneroster_id
       @district_bearer_token = district_bearer_token
       @district_id = district_id
       @tenant_id = tenant_id
-      @premium_school_external_ids = fetch_premium_school_external_ids
+      @district = District.find(district_id)
```
brendanshean: I think it's a good practice to keep `initialize` as just setting raw input values.  Adding a database call here introduces side effects.

It looks like district is only used in `premium_school_ids` method.  If so, a potential rewrite with a subscriptions scope

```
private def premium_school_ids
    @premium_school_ids ||= 
      School
        .where(district_id:)
        .joins(:subscriptions)
        .not_expired
        .pluck(:id) 
end

File: services/QuillLMS/app/services/classlink_integration/school_persister.rb [hunk 2]
```diff
@@ -2,22 +2,26 @@
...
       @name = name
       @district_id = district_id
       @classlink_instance_id = classlink_instance_id
       @nces_id = nces_id
       @sourced_id = sourced_id
+      @premium_school_ids = premium_school_ids
     end
 
     def run
+      found_school = school
```
anathomical: Is this indirection necessary?  Should we just reference `school` in the `run` method, or possibly rename the `school` method to `found_school`?

services/QuillLMS/app/services/classlink_integration/school_persister.rb
```diff
@@ -27,8 +31,8 @@ def run
     end
 
     private def school
-      school = nces_id ? School.find_by(nces_id:, district_id:) : School.find_by(name:, district_id:)
-      school.tap { |s| s.update(name:) }
+      found_school = (nces_id && School.find_by(nces_id:, district_id:)) || School.find_by(name:, district_id:)
```
anathomical: This feels slightly hard to parse.  I wonder if it's worth breaking this into a pair of sub-methods:
```ruby
private def school_by_nces_id
  return unless nces_id

  School.find_by(nces_id: district_id:)
end

private def school_by_name = School.find_by(name:, district_id:)

private def school
  found_school = school_by_nces_id || school_by_name
  found_school if premium_school_ids.include?(found_school&.id)
end
```

services/QuillLMS/spec/services/classlink_integration/school_persister_spec.rb
brendanshean, line 1: I feel like this test file could use some refactoring.  Instead of defining let(:persister) or let(:premium_persister) in each block, define only one let(:persister) at the top of the file, then in each context block specify the variations via let variables:

```
let(:persister) {... }

context do
   let(:school_name) { ... }
   let(:nces_id) {  ... }
    ...
end

context do 
  let(:nces_id) { ... }
   ...
end
```

This allows someone to look through the tests quickly and determine what is changing between the tests.  Also, if the interface for persister changes in the future, you only have to update it in one place.

anathomical (approved):
I think the tweaked approach is clean.  Allowing for the edge-case of non-NCES schools makes sense, and is probably something we need to consider in future code as well.

brendanshean (approved):
I left some minor `refactoring` comments, but nothing that requires a re-review.  Otherwise, looks good to go.

PR #13151, opened 2025-07-29
sidekiq-ify User's Geocode API call

happythenewsad:
## WHAT
- sidekiq-ify User's Geocode API call
## WHY
- Classlink's bulk User creations are hitting our Geocode API's rate limit. 
- it's best practice to make network calls asynchronous anyway, so might as well scope this solution to all Users, not just Classlink-created Users 

### Screenshots

### Notion Card Links
none

- create a user
- observe the worker perform 
- reload the user and observe the new time zone

PR Checklist | Your Answer
------------ | -------------


services/QuillLMS/app/workers/geocoder_timezone_worker.rb
```diff
@@ -0,0 +1,22 @@
+# frozen_string_literal: true
+
+class GeocoderTimezoneWorker
+  include Sidekiq::Worker
+
+  def perform(user_id, ip_address)
+    user = User.find_by_id(user_id)
+    return unless user && ip_address.present?
```
anathomical, line 11: Technically we should be persisting `ip_address` on the User model, so should be able to fetch it if it wasn't passed, but I think that passing its value at time of enqueue seems reasonable for weird edge cases where it might be updated while the job is on the queue.  I think that in such cases we want to use the value at time of enqueue and not whatever is "current"?

anathomical (approved):
Looks good.  A note on `ip_address` persistence, but I think this approach probably makes sense as-is.

services/QuillLMS/app/workers/geocoder_timezone_worker.rb
```diff
@@ -0,0 +1,22 @@
...
+    return unless time_zone
+
+    user.update!(time_zone:)
+  end
+
+  private def fetch_timezone_from_ip(ip_address)
+    geocoder_results = Geocoder.search(ip_address.to_s)
+    geocoder_results.first&.data&.dig('timezone', 'name')
+  rescue Geocoder::ConfigurationError => e
+    nil
```
dandrabik: We probably want an ErrorNotifier call here so this doesn't fail in silence, e.g. something happens with our account.

services/QuillLMS/app/workers/geocoder_timezone_worker.rb
```diff
@@ -0,0 +1,22 @@
+# frozen_string_literal: true
+
+class GeocoderTimezoneWorker
+  include Sidekiq::Worker
```
dandrabik, line 4: We might consider putting this in the 'low' queue so it naturally throttles with typically one worker. Not sure if there is an impact on user experience though (if so, maybe leave in 'default' queue).

dandrabik (approved):
Overall looks good! I left two small comments for consideration.

services/QuillLMS/app/workers/geocoder_timezone_worker.rb
happythenewsad: 👀

PR #13162, opened 2025-08-04
teacher email verification - extend to canvas and classlink

happythenewsad:
## WHAT
- add email verification flow to canvas and classlink auth controllers, to match google and clever

## WHY
- product ask

### Screenshots

### Notion Card Links
https://www.notion.so/quill/P1-Add-Classlink-and-Canvas-as-a-third-party-auth-type-for-email-verification-237d42e6f94180ecbcbad69f41ccc8b8

1. as a classlink teacher with no verification record, sign in 
2. ensure that `user.email_verified?` returns true
3. Repeat again, to check idempotency
4. Run the rake task on staging
5. spot check 1 classlink user and 1 canvas user; confirm `email_verified? == true`

PR Checklist | Your Answer
------------ | -------------


services/QuillLMS/app/controllers/auth/canvas_controller.rb
```diff
@@ -9,6 +9,7 @@ def canvas
       sign_in(user)
 
       redirect_to profile_path
+      user.verify_email(UserEmailVerification::CANVAS_VERIFICATION) if user&.teacher?
```
emilia-friedberg, line 12: For the other auth paths, we perform this action here: https://github.com/empirical-org/Empirical-Core/blob/develop/services/QuillLMS/app/services/complete_account_creation.rb/#L11-L12. I wonder if it makes sense to have this and the ClassLink controller call `CompleteAccountCreation`, and update the `verify_email_for_third_party_auth` to handle Canvas and ClassLink users?

happythenewsad: Yeah, I saw that too. That worker does other things that are not mentioned in the notion card, like create AdminInfo. I'll ask Peter Sharkey

happythenewsad: Emilia and I discussed the pros and cons of a couple different approaches offline. We aligned on the current approach. Here is some supporting reasoning:

1. The `verify_email_for_third_party_auth` is a pass-through method, so there's little value-add in adapting it to new providers, if they're already calling `verify_email`
2. `CompleteAccountCreation` is not idempotent, so adding it to Classlink and Controller flows requires checking extra state (whether the User record was created just now, or not)
3. `CompleteAccountCreation` does create AdminInfos, so this approach extends the current spec. These could be added later if they are ever requested

emilia-friedberg (approved):

PR #13169, opened 2025-08-06
tweak-student-learning-sequence-assignment

anathomical:
## WHAT
- Adjust `FindStudentLearningSequence` logic to prefer `StudentLearningSequence` records with completed pre diagnostics
- Update the `MovePracticeWorker` logic to the `FindStudentLearningSequence` service for consistency

NOTE: Once this is in place, I'll use a Rails console to enqueue ~200k `MovePracticeWorker` jobs to the `migration` queue.  This took a couple of hours to process on staging, so I plan to just shepherd it once the deploy completes
## WHY
- If a student has been assigned the Pre Diagnostic twice in the same classroom this helps us ensure that any practice gets attached to a `StudentLearningSequence` with a completed Pre Diagnostic instead of being attached to simply the "last assignment".  This handles cases where a student is assigned and completes the pre, gets assigned a second instance of the pre, and then gets assigned practice.
- This simplifies the code and ensures that we aren't duplicating look-up logic
## HOW
- Add an additional `order` call to `FindStudentLearningSequence` to find the most recently-completed pre diagnostic.  In most cases this will return the same thing it always has, but it handles this edge case
- Remove custom code for finding target SLS in the worker and replace it with a call to `FindStudentLearningSequence`

### Notion Card Links
https://www.notion.so/quill/Investigate-Possible-Data-Discrepancy-for-Diagnostic-Insights-report-233d42e6f941801faf69d8d868e99273?source=copy_link

- Run backfill on staging (~200k records)
- Spot check reported teachers with issues from #diagnostic-insights reports and confirm that most of them seem fixed

PR Checklist | Your Answer
------------ | -------------


emilia-friedberg (approved):
LGTM! Thanks for figuring out this issue!

dandrabik (approved):
LGTM. 🚢

PR #13170, opened 2025-08-06
don't approve verification requests if teacher no longer has a school

emilia-friedberg:
## WHAT
Don't approve verification requests if the teacher no longer has a school.

## WHY
We were getting a Sentry error from the worker that sends the emails telling a teacher they've been approved when `teacher.school` is undefined. Since we shouldn't be verifying them at all if they don't have a school, I'm moving the fix further up the change of logic so we don't perform the verification itself or run those workers in the event that they have no school.

## HOW
Just add a check for `teacher.school` before running any of this logic.

### Screenshots

### Notion Card Links
https://www.notion.so/quill/TeacherVerificationApproved-EmailWorker-unhandled-exception-247d42e6f9418094975cc20c0016ee42

Just ran tests.

PR Checklist | Your Answer
------------ | -------------


services/QuillLMS/app/controllers/admins_controller.rb
```diff
@@ -305,7 +305,7 @@ def create_and_link_accounts
 
   private def handle_approve_teacher_verification_request(teacher_id)
     teacher = User.find_by(id: teacher_id)
-    return if teacher.teacher_verification_status == TeacherInfo::APPROVED || teacher.admin_approval_status == AdminInfo::APPROVED
+    return if !teacher.school || teacher.teacher_verified? || teacher.admin_verified?
```
anathomical: This is purely stylistic, but I think I'd almost make this new condition a new early return call since it's kind of unrelated logic.  So I might do:
```ruby
return if teacher.teacher_verified? || teacher.admin_verified?
return unless teacher.school.present?
```
Or `return if teacher.school.nil?`

anathomical (approved):
Minor style comment, but I very much like your decision to move this check up to the approval step.  That makes the most sense.

happythenewsad (approved):
❤️

PR #13174, opened 2025-08-06
allow unlinked teachers to connect to clever and other auth methods

emilia-friedberg:
## WHAT
Update the `fail_conflicting_auth_user` method not to fail if a user attempting to log in through an SSO is not linked to a different SSO (ie, currently signs in using a password).

## WHY
This code is breaking our Clever setup, which previously linked existing teacher accounts to Clever through the sign-in flow. Currently, Clever sign-in attempts that aren't new users are just failing. Instead, I think we should just allow the linking here, unless there is another SSO configured. This is consistent with how other websites manage auth, in my experience. 

## HOW
Update the code to allow the sign-in flow to continue if the user's attempted provider matches their existing provider, or if their existing provider is :password.

### Screenshots

### Notion Card Links
https://www.notion.so/quill/Unable-to-connect-Quill-account-with-Clever-sync-247d42e6f9418069814dd9bcb9f199bd?source=copy_link

Ran the tests, QAing on staging now.

PR Checklist | Your Answer
------------ | -------------


anathomical (commented):
Am I reading this right that this change would make it possible for a user to log in via password in cases where they are linked to Google?  This could allow someone to log into a teacher's account without having access to (say) their Google account, but has figured out their old Quill password, right?  That might be acceptable since it's probably better than teachers not being able to auth with Clever to have their accounts linked at all, but I want to make sure I understand the implication of this change.

emilia-friedberg:
> Am I reading this right that this change would make it possible for a user to log in via password in cases where they are linked to Google? This could allow someone to log into a teacher's account without having access to (say) their Google account, but has figured out their old Quill password, right? That might be acceptable since it's probably better than teachers not being able to auth with Clever to have their accounts linked at all, but I want to make sure I understand the implication of this change.

@anathomical no -- in that instance, line 105 would evaluate like this:

`return if [:password, :password].include?(User::GOOGLE_PROVIDER) => false`

So it'll still fail. I added a test to that effect, so it's clearly documented for the future.

happythenewsad (approved):
LGTM

PR #13176, opened 2025-08-07
migrate existing CL users

happythenewsad:
## WHAT
- during a rostering sync, if an incoming API user belongs to another ClasslinkInstance, fold them into the current ClasslinkInstance

## WHY
- Because of the hard requirement that `User has_one ClasslinkAccount has_one Classlinkinstance`, we create "stub" ClasslinkInstances when unrostered users sign in with Classlink for the first time (otherwise, SSO would be impossible). An artifact of this is that the sequence of 

A. `user` SSO's via Classlink into LMS
B. `user`'s canonical district starts a rostering sync sometime later 

will have the effect of keeping `user` in their own orphaned ClasslinkInstance, without access to their canonical district's resources. 

## HOW
- During rostering sync, if an incoming user matches an existing LMS Classlink user, migrate that user to the correct ClasslinkInstance

### Screenshots

### Notion Card Links
https://www.notion.so/quill/Product-Board-30f97e4eb01246dbb8264253fb385073?p=23ed42e6f94180dcbb99dea8e889bd32&pm=s

1. Update an existing CL user's ClasslinkAccount to point to any other instance
2. run the rostering sync
3. observe that the user's ClasslinkAccount points to the new ClasslinkInstance
4. impersonate the user; check that dashboard looks normal

PR Checklist | Your Answer
------------ | -------------


services/QuillLMS/app/services/classlink_integration/user_persister.rb
```diff
@@ -42,6 +44,11 @@ def link_existing_user_to_classlink
       existing_user
     end
 
+    def migrate_to_new_instance(user)
```
happythenewsad: emails are globally unique across Classlink tenants, so there is no risk of a user migrating back and forth between tenants

services/QuillLMS/app/services/classlink_integration/user_persister.rb
```diff
@@ -30,6 +30,8 @@ def run
     end
 
     def link_existing_user_to_classlink
+      return migrate_to_new_instance(existing_user) if existing_user.classlink_user?
```
dandrabik: Take or leave, since `existing_user` is a memoized method, you might consider not passing it as a param to just pass it back, e.g.
```ruby
if existing_user.classlink_user?
  migrate_to_new_instance
  return existing_user
end

dandrabik (approved):
Looks reasonable. Small take or leave comment.

PR #13177, opened 2025-08-07
fix-incomplete-slsa-records

anathomical:
## WHAT
- Add new validation to prevent `StudentLearningSequenceActivity` records from being linked to incomplete `ActivitySession` records
- Create a `rake` task to clean up existing records and output a stream of changes made to record for posterity (for potential reversal if it's ever needed)
## WHY
- `StudentLearningSequenceActivity` records are only intended to link to `ActivitySession` records upon the completion of the session, never before
- We have some records that have gotten into a bad state, and should fix them, but want a record of the changes just in case we ever need to go back
## HOW
- Add new conditional validations so that `activity_session_id` must be present when `completed_at` is present, and vice versa
- Add a rake task using BigQuery to generate the list of candidate items (because it's faster than PostgreSQL on these tables), then confirm each item actually is in an invalid state on Postgres, and if so clean up that item and output the `id` and newly-removed `activity_session_id` values that we altered the row from

- Run locally against the staging database

PR Checklist | Your Answer
------------ | -------------
Have you deployed to Staging? | Not deployed, but run agains the db


services/QuillLMS/lib/tasks/student_learning_sequence_activities.rake
```diff
@@ -0,0 +1,36 @@
+# frozen_string_literal: true
+
+namespace :student_learning_sequence_activities do
+  # Pipe output into a local CSV
+  # heroku run rake student_learning_sequence_activities:disconnect_unfinished_activity_sesions > slsa_disconnect_activity_sessions.csv
+  desc 'Stop pointing StudentLearningSequenceActivity records at incomplete ActivitySessions if they are imporoperly connected'
+  task disconnect_unfinished_activity_sesions: :environment do
```
happythenewsad: method name should be `*sessions`

happythenewsad (commented):
I like the validations you've added. I'm holding off on approving because I want to first get your thoughts on the usefulness of further investigating the root cause, before applying the bandaid the rake task provides.

As part of that conversation: We create SLSA's in two places: handle_completion.rb and handle_assignment.rb. 
I wonder if it's worthwhile to hook in ErrorNotifier in these paths, injecting diagnostic metadata, so that 
we can better understand why some SLSA's are referencing incomplete ActivitySessions.

services/QuillLMS/lib/tasks/student_learning_sequence_activities.rake
anathomical: Good catch.

services/QuillLMS/lib/tasks/student_learning_sequence_activities.rake
```diff
@@ -0,0 +1,36 @@
...
+        JOIN lms.activity_sessions AS act_s
+          ON slsa.activity_session_id = act_s.id
+        WHERE slsa.activity_session_id IS NOT NULL
+          AND slsa.completed_at IS NULL
+          AND act_s.completed_at IS NULL
+    SQL
+    results = QuillBigQuery::Runner.execute(sql)
+
+    puts 'slsa_id,activity_session_id'
+    StudentLearningSequenceActivity.transaction do
```
dandrabik, line 20: I'm surprised you would need Bigquery here instead of just a `.find_each` which would just find batches of 1000 for you.

dandrabik (commented):
I just have a general question that I'll ping you about.

dandrabik (approved):
LGTM after a discussion.

PR #13179, opened 2025-08-08
replace stripped punctuation with spaces rather than just stripping

emilia-friedberg:
## WHAT
Fix a bug that sometimes arises in the mismatch detector between plagiarism texts and their passage, stemming from html tags being directly adjacent to punctuation.

## WHY
Matt reported this bug.

## HOW
Replace the punctuation with spaces, rather than stripping it entirely, because in the previous version we were sometimes winding up with words that no longer had a space between them, which didn't match the passage.

### Screenshots

### Notion Card Links
https://www.notion.so/quill/CMS-bug-with-Plagiarism-rules-249d42e6f94180cea72efeddde35dfb7?source=copy_link

Tested that the plagiarism text was no longer falsely reported not to match the passage, and that the plagiarism detector still works within the passage.

PR Checklist | Your Answer
------------ | -------------


anathomical (approved):

services/QuillLMS/engines/evidence/app/helpers/text_formatter.rb
```diff
@@ -4,6 +4,11 @@ module TextFormatter
   require 'cgi'
   require 'nokogiri'
 
+  def replace_punctuation_with_spaces(str)
```
happythenewsad: This method can be private, since it is only used within this module

happythenewsad (changes requested):
Looks good, but please add spec coverage

happythenewsad (approved):

PR #13181, opened 2025-08-08
confirm-practice-assignments

anathomical:
## WHAT
- Make a small tweak to a clean-up script so that it handles an additional edge case around pre diagnostic completion
- Add a new scope to the `StudentLearningSequence` model.  Not used in this PR, but likely to be useful in the near future.
## WHY
- We want to run this clean-up again and fix this even-rarer than the original case edge case
## HOW
- Update the worker with an additional `JOIN` condition and slightly tweaked `ORDER` params
- Refactor the worker code to reduce cylcomatic complexity while minimizing actual changes

- Deployed to staging
- Run workers on staging
- Confirm that diagnostic query to identify cases of this error reports fixed instances

PR Checklist | Your Answer
------------ | -------------


services/QuillLMS/app/workers/student_learning_sequences/backfill/fix_slsa_sls_assignment_worker.rb
```diff
@@ -47,7 +66,7 @@ def perform(id)
         @intended_sls ||= StudentLearningSequence
           .joins(:initial_classroom_unit)
           .joins(join_sql)
-          .order(created_at: :desc)
+          .order('student_learning_sequence_activities.completed_at DESC')
```
happythenewsad, line 69: Am I correct in surmising that this line is intended to prevent future mal-formed SLSAs, but doesn't interact with the worker changes in this PR?

happythenewsad (commented):
Consider adding specs to the worker as well, both to mitigate regressions and to document the intent of this patch. Thanks!

The patterns in this file make worker rspec testing almost as light as standard Ruby class testing: https://github.com/empirical-org/Empirical-Core/blob/develop/services/QuillLMS/spec/tasks/users_rake_spec.rb

services/QuillLMS/app/workers/student_learning_sequences/backfill/fix_slsa_sls_assignment_worker.rb
anathomical: For this project I've used the `Backfill` namespace to indicate workers that are effectively data clean-up scripts.  They're intended to be run entirely manually, and for the most part intended to be run once.  This was originally written to clean-up a data issue we found during testing, but once the code got into production we discovered that there was a rare edge case that it didn't handle properly.  Thus I've made updates to this script with the intent to run it once, let it clean up our data, and then never again.

I don't know if was a good judgment call, but I basically decided back when we were writing a bunch of these `Backfill` workers that writing specs for one-off scripts of this type wasn't worth it, but maybe that's not correct. Open to thoughts on this front.

happythenewsad: I understand that the team convention has been not to test small one-off scripts. However, in my view, this PR calls for tests because it:

- has fairly complex logic, to the extent that it would difficult to reverse engineer the full intent of the patch in the future, without tests acting as documentation
- destroys/mutates data, and is thus high impact
- even if we `git rm` the script, git history with the script + specs could be helpful with (1)

dandrabik (approved):
Thanks for cleaning this up. Adding a spec as a gut check is a good idea, so I agree with PK on that.

happythenewsad (approved):
Thanks for adding the tests

PR #13183, opened 2025-08-11
count-recommendation-completions-accurately

anathomical:
## WHAT
- Update `PracticeByClassroom` so that it applies `distinct` to `StudentLearningSequenceActivities` lookup
- Update `PracticeByClassroom` to use explicit pre and post diagnostic IDs instead of inferring based on `follow_up_activity_id`
- Update `SkillPracticeByClassroom` defaults so that they apply even if `nil` is explicitly passed in
- Update `SkillPracticeByClassroom` to confirm that the Pre Diagnostic associated with a recommendation was completed inside the specified timeframe before counting the recommendation as practice
## WHY
- There exist some `Recommendation` records that point to archived Diagnostics and `UnitTemplate` records that are used for live diagnostics (see activity 413).  These resulted in rows doubling on the join in certain data edge cases.
- We were assuming that Recommendations were only connected to activities that we cared about, but there are some activities tied to Recommendations that we don't care about, and being explicit lets us exclude them from our queries.
- If you passed `nil` explicitly to `start_date` or `end_date` the code would error because the default would be overwritten in an unsupported manner.  Now if `nil` is passed in explicitly it will be replaced by the default.
- There are some rare cases where a student completed the Pre Diagnostic outside of our Diagnostic Insights timeframe (a little over two years), but then completed a recommendation within the timeframe.  One example had a student finishing the diagnostic and recommendations two years apart.  In these cases we were counting the practice correctly, but on the Student History view we don't show data if the Pre Diagnostic wasn't completed within the timeframe, so the practice was only showing up on the Practice by Student page, appearing as a potential discrepancy.
## HOW
- Add `.distinct` to the data call
- Replace the `pre_diagnostic_practice_data` and `post_diagnostic_practice_data` conditions
- Use `nil` as the default in the method signature and then use `value || DEFAULT` when setting the class variable in `initialize`
- Add a `JOIN` in the query to find the completion date for the pre diagnostic associated with any recommendation completion and include a date check on it

- Wrote specs based on known cases
- While data on staging and production has drifted with start-of-year, I've fudged around in the console to run these reports for affected students and confirm that they're in proper state with the new structure
- Deploy to staging and test which accounts do have the same data, or find teachers who have some of the problematic students to see the report from their perspective (classroom `1799211`)

PR Checklist | Your Answer
------------ | -------------


dandrabik (approved):
:+1: LGTM

PR #13193, opened 2025-08-13
skills report pt 1: routing and sidebar

emilia-friedberg:
## WHAT
Get routing and bare-minimum rendering in place for the Skills Report (just the sidebar).

## WHY
This is the first step of this project! Follow-up PR will have the first report page, then all the rest.

## HOW
Boilerplate routing stuff, plus the index component and sidebar. Also, I moved the `MobileSidebarButtonContainer` to the Shared bundle because I'm going to be using it in the next PR.

### Screenshots
Screenshot 2025-08-13 at 2 14 13 PM

### Notion Card Links

Just locally rendering, not much to QA yet.

PR Checklist | Your Answer
------------ | -------------


services/QuillLMS/client/app/bundles/Teacher/components/progress_reports/skills_report/index.tsx
```diff
@@ -0,0 +1,109 @@
...
+const openSidebarIconSrc = `${SMALL_ICON_BASE_SRC}/sidebar-open.svg`
+const closeSidebarIconSrc = `${SMALL_ICON_BASE_SRC}/sidebar-close.svg`
+
+interface DiagnosticInsightsProps {
+  classrooms: ClassroomInterface[];
+  allSkillCategories: SkillCategoryInterface[];
+  allSkillSubCategories: SkillCategoryInterface[];
+}
+
+const DiagnosticInsights = ({ classrooms, allSkillCategories, allSkillSubCategories, }: DiagnosticInsightsProps) => {
```
emilia-friedberg: Note: this is wrong! I've got it fixed in the follow-up PR.

services/QuillLMS/client/app/bundles/Teacher/components/progress_reports/skills_report/index.tsx
```diff
@@ -0,0 +1,109 @@
...
+        selectedClassroomId={classroom?.id}
+        setSelectedClassroom={setClassroom}
+        showMobileSidebar={showMobileSidebar}
+        showSidebarButton={renderShowSidebarButton()}
+      />
+    )
+  }
+
+  return (
+    <div className="diagnostic-insights-container">
```
emilia-friedberg: Likewise here.

services/QuillLMS/client/app/bundles/Teacher/components/progress_reports/skills_report/interfaces.ts
```diff
@@ -0,0 +1,26 @@
+export interface ClassroomInterface {
```
happythenewsad, line 1: Since we don't enforce TypeScript, interfaces like these can drift over time and become counterfactual.

happythenewsad (approved):

services/QuillLMS/app/controllers/teachers/progress_reports_controller.rb
```diff
@@ -58,6 +54,16 @@ def diagnostic_insights
     render 'diagnostic_insights'
   end
 
+  def skills_report
+    @classrooms = classrooms
+
+    # TODO: update these two once we have more clearly scoped concepts to pull from
+    @all_skill_categories = Concept.where(parent_id: nil).order(:name)
+    @all_skill_sub_categories = Concept.where(parent_id: @all_skill_categories.ids).order(:name)
```
anathomical: We have a new helper pattern that we can use here, I believe: `Concept.level2` and `Concept.level1` should do this without having to specify ids directly.

services/QuillLMS/spec/controllers/teachers/progress_reports_controller_spec.rb
```diff
@@ -281,6 +281,33 @@
...
+
+    it 'should assign all_skill_categories to level 2 concepts' do
+      get :skills_report
+      expect(assigns(:all_skill_categories)).to eq [level_two_concept]
+    end
+
+    it 'should assign all_skill_categories to level 1 concepts' do
+      get :skills_report
+      expect(assigns(:all_skill_sub_categories)).to eq [level_one_concept]
+    end
```
anathomical: I don't know that this is actually a better pattern, but all of these `assigns` tests could be wrapped in a `context` with the `get` call in a `before` block to reduce repitition.
```ruby
context 'assigns checks' do
  before { get :skills_report }

  it { expect(assigns(:classrooms)).to eq [visible_classrooms] }
  ... etc.
end
```

services/QuillLMS/client/app/bundles/Teacher/components/progress_reports/skills_report/interfaces.ts
anathomical: This is a good point.  Probably worth discussing at an Engineering Meeting.  I'll put it on the agenda.  That said, I think that it's better to have these than not as they offer coordination value between front and back end engineers.

anathomical (approved):

services/QuillLMS/app/controllers/teachers/progress_reports_controller.rb
emilia-friedberg: Yeah, this branch doesn't have those merged in yet. I will try to remember to change it, but we'll have to update it later when actually scope those concepts anyway.

eadams17 (approved):
Looks good to me!

PR #13195, opened 2025-08-14
make sure last name sorting is consistent in student reports

emilia-friedberg:
## WHAT
Make sure last name sorting is consistent in student reports.

## WHY
A teacher specifically reported this as an issue, and in general we want this to be consistent. Note that I really think the ideal solution here would be to actually extract first name and last name as different database columns, so we can respect whatever sorting the user set (the downside to this way of sorting is that two-part first names, like "Mary Jo", are now going to be treated as two-part last names). However, this is just a support ticket, and I think it's more important to be consistent both in our code and in the UI.

## HOW
Go through the various reports and fix everywhere we were not using the "everything after the first name" version of last name definition to do so -- mostly My Diagnostics and the premium reports.

### Screenshots

### Notion Card Links
https://www.notion.so/quill/Two-part-last-names-sorting-in-Reports-24dd42e6f941803e85ffcf7301f14f7c?d=a0dd42e6f94182fa957103281af31af8

Confirmed locally that last names now show up in a consistent order.

PR Checklist | Your Answer
------------ | -------------
Have you deployed to Staging? | NO -  tiny change


services/QuillLMS/client/app/modules/__tests__/sortingMethods.test.js
brendanshean, line 1: Nice testing!

brendanshean (approved):
I really like the standardizing of sorting with `ordered_by_last_name`, rather than custom logic each time.

anathomical (approved):
Looks good to me.  Thanks for cleaning this up in so many places.

PR #13196, opened 2025-08-14
handle-additional-di-cleanup-cases

anathomical:
## WHAT
After the last update and run of this script, our alerts revealed another ~10k records that the code didn't handle.  I've streamlined the logic and specs to cover existing cases and the newly revealed one
## WHY
We want teachers to get accurate data when looking at Diagnostic Insights reports which are built on these records
## HOW
Streamline the logic that determines when to remove "duplicate" SLSA records so that the target records for this worker will be pointed at the appropriate StudentLearningSequences

- Deploy to staging
- Run on staging
- Confirm that known cases of this edge case are resolved

PR Checklist | Your Answer
------------ | -------------


services/QuillLMS/spec/workers/student_learning_sequences/backfill/fix_slsa_sls_assignment_worker_spec.rb
```diff
@@ -65,78 +66,81 @@ module Backfill
           let!(:intended_pre_diagnostic_slsa) { nil }
 
           it 'returns early without updating the student_learning_sequence_id' do
-            expect { worker.perform(target_slsa.id) }
-              .not_to change { target_slsa.reload.student_learning_sequence_id }
+            expect { subject }
+              .not_to change(target_slsa, :reload)
```
brendanshean: I don't think this is testing what you intend here.   In before/after, `reload` should return the same object.  I think you still want to check the attribute `student_learning_sequence_id` for an update.

services/QuillLMS/spec/workers/student_learning_sequences/backfill/fix_slsa_sls_assignment_worker_spec.rb
```diff
@@ -65,78 +66,81 @@ module Backfill
...
         end
 
         context 'when student_learning_sequence_id is already correct' do
           let(:assigned_student_learning_sequence) { intended_student_learning_sequence }
 
           it 'returns early without making changes' do
-            expect { worker.perform(target_slsa.id) }
-              .not_to change { target_slsa.reload.student_learning_sequence_id }
+            expect { subject }
+              .not_to change(target_slsa, :reload)
```
brendanshean: I don't think this is testing what you intend here.   In before/after, `reload` should return the same object.  I think you still want to check the attribute `student_learning_sequence_id` for an update.

services/QuillLMS/spec/models/student_learning_sequence_spec.rb
```diff
@@ -59,4 +59,43 @@
...
     end
   end
+
+  context '#initial_student_learning_sequence_activity' do
+    subject { object.initial_student_learning_sequence_activity }
+
+    let(:object) { create(:student_learning_sequence) }
+
+    context 'there is no associated SLSA' do
+      it 'returns nil safely' do
```
brendanshean: I think the example description here isn't adding much value.  I would go for the one-liner:  
`it { is_expected.to be_nil }`

services/QuillLMS/spec/models/student_learning_sequence_spec.rb
```diff
@@ -59,4 +59,43 @@
...
+          student_learning_sequence: object,
+          activity:,
+          classroom_unit:)
+      end
+
+      context 'that is not for the initial activity' do
+        let(:activity) { create(:activity) }
+        let(:classroom_unit) { create(:classroom_unit) }
+
+        it 'returns nil safely' do
```
brendanshean: I think the example description here isn't adding much value.  I would go for the one-liner:  
`it { is_expected.to be_nil }`

services/QuillLMS/spec/models/student_learning_sequence_spec.rb
```diff
@@ -59,4 +59,43 @@
...
+          expect(subject).to be_nil
+        end
+      end
+
+      context 'that is for the initial activity' do
+        let(:activity) { object.initial_activity }
+        let(:classroom_unit) { object.initial_classroom_unit }
+
+        it 'returns nil safely' do
```
brendanshean: My copy-paste error with example description.  In any event, I don't it adds much value and I would use a one-liner: `it { is_expected.to eq(slsa }`

services/QuillLMS/app/workers/student_learning_sequences/backfill/fix_slsa_sls_assignment_worker.rb
brendanshean, line 1: I like the organization of this worker.

services/QuillLMS/spec/workers/student_learning_sequences/backfill/fix_slsa_sls_assignment_worker_spec.rb
anathomical: Good catch on this.  I was expecting it to us `eq` for the comparison, which I think should compare the values in the object, not just its identity, but you are right that that's not what is happening with this syntax, so I've updated it to `change { target_slsa.reload.student_learning_sequence_id }` as you suggested.

brendanshean (approved):

dandrabik (approved):
LGTM.

PR #13197, opened 2025-08-14
fix publish dates not being set properly

emilia-friedberg:
## WHAT
Fix how publish and due dates are being passed back from the frontend so that they are naive dates (ie, have no timezone) and our UTC adjustments work correctly.

## WHY
While teachers were seeing the correct date on their interfaces, the backend was actually storing them with additional offsets due to the fact that they were being passed *with* timezones that may or may not match the teacher's set timezone. This led to students not seeing assignments that should have appeared on their dashboards for whatever the additional offset was.

## HOW
Pass back a naive datetime string, rather than one that included timezone data. Also, clean up a bunch of unnecessary moment-ing, which was confusing things further.

### Screenshots

### Notion Card Links
https://www.notion.so/quill/Delay-in-Activity-Assignment-Times-and-Due-Date-Times-Between-Teacher-and-Student-Accounts-24fd42e6f941800ca704d97f5d2a107e?v=77f294a5b0df466e98633244de7ebfef&source=copy_link

Tested locally that assigning activities with publish dates and then changing them worked as expected in terms of when the students saw them on their dashboards.

PR Checklist | Your Answer
------------ | -------------


anathomical (approved):
Timezones are hard!

PR #13201, opened 2025-08-15
fix classroom unit archiving behavior

emilia-friedberg:
## WHAT
Fix bug where classrooms getting archived did not necessarily archive their classroom units, which cascaded into a bug where the units associated with those classroom units did not get archived, which led to weird ghost units with "visible" and "open" set to true but that did not show up anywhere in the teacher interface.

## WHY
Teachers are seeing a bug that is at least partially due to this issue, where assigning this year's Diagnostic packs results in packs with last year's Diagnostic in it as well. This is because they'd previously assigned that pack, then archived associated classes, and then when they go to assign the new one we have some logic that looks for units with the same name. Because those packs weren't archived and had the same name as this year's pack (`Starter Diagnostic (Baseline Pre)`), it would update them with the new activity rather than just creating a new one. I am also investigating whether we can just remove that logic or not, but getting the archiving logic working as expected would at least partially solve it going forwards and should be in place anyway.

## HOW
Change from `visible_changed?`, which presumably worked in some previous version of Rails we were using but does not now, to `saved_change_to_visible?`, and update tests.

Ideally, after merging this, we'd run:

```
archived_classrooms_with_visible_classroom_units = Classroom.unscoped
  .where(visible: false)
  .where(
    ClassroomUnit
      .where("classroom_units.classroom_id = classrooms.id")
      .where(visible: true)
      .arel.exists
  )

archived_classrooms_with_visible_classroom_units.each do |archived_classroom|
  archived_classroom.hide_all_classroom_units
end
```

### Screenshots

### Notion Card Links

Manually tested that archiving a classroom now archives all of its associated classroom units, which in turn archives the units if those were the only classroom units associated with it.

PR Checklist | Your Answer
------------ | -------------


brendanshean (approved):
Nice work on figuring this one out.  I think the attribute_changed? functionality was deprecated in Rails 6 so I'm confused why we didn't see this earlier.  We probably want to check for other cases of `*_changed?` and update those as well.

emilia-friedberg:
@brendanshean I think we probably had the issue earlier, but the ghost packs weren't causing notable issues til now.

services/QuillLMS/spec/models/classroom_spec.rb
```diff
@@ -200,16 +200,56 @@ def invalid_classroom_attributes
...
+    context 'as after_commit callback' do
+      let(:unit) { create(:unit, user: classroom.owner) }
+      let!(:classroom_unit) { create(:classroom_unit, unit:, classroom:, visible: true) }
+
+      it 'should automatically hide classroom units when classroom becomes invisible' do
+        expect(classroom_unit.visible).to be
+
+        classroom.update!(visible: false)
+
+        expect(classroom_unit.reload.visible).not_to be
```
anathomical: I would probably make this a `expect to change` test:
```ruby
expect { classroom.update!(visible: false) }
  .to change { classroom_unit.reload.visible }
  .from(true)
  .to(false)
```

services/QuillLMS/spec/models/classroom_spec.rb
```diff
@@ -200,16 +200,56 @@ def invalid_classroom_attributes
...
+      end
+
+      it 'should not hide classroom units when classroom remains visible' do
+        expect(classroom_unit.visible).to be
+
+        classroom.update!(name: 'Updated Name') # Change something other than visible
+
+        expect(classroom_unit.reload.visible).to be
```
anathomical: Similar to my above comment, but with `expect { ... }.to not_change { classroom_unit.reload.visible }`

services/QuillLMS/spec/models/classroom_spec.rb
```diff
@@ -200,16 +200,56 @@ def invalid_classroom_attributes
...
+      end
+
+      it 'should not hide classroom units when classroom becomes visible' do
+        classroom.update!(visible: false) # First make it invisible
+        expect(classroom_unit.reload.visible).not_to be # Verify it was hidden
+
+        classroom_unit.update!(visible: true) # Reset to visible
+        classroom.update!(visible: true) # Now make classroom visible again
+
+        expect(classroom_unit.reload.visible).to be # Should remain visible
```
anathomical: I think this could be
```ruby
expect { classroom.update!(visible: false) }
  .to change { classroom_unit.reload.visible }
  .to(false)

expect {
  classroom_unit.update!(visible: true)
  classroom.update!(visible: true)
}.to change { classroom_unit.reload.visible }
  .to(true)
```

anathomical (approved):
Good catch on this.  I find failing callbacks to be very hard to debug.

PR #13202, opened 2025-08-15
stop updating units with the same name on create

emilia-friedberg:
## WHAT
Remove code that finds units with the same name on the create unit action, and updates them rather than creating a new unit.

## WHY
Teachers are seeing a bug that is at least partially due to this issue, where assigning this year's Diagnostic packs results in packs with last year's Diagnostic in it as well. This is because they'd previously assigned that pack, then archived associated classes, and then when they go to assign the new one, it would find last year's unit and update that instead of creating a new one. There is an additional issue here, which is that those packs largely should have already been archived (and if they had been, this never would have been possible, due to the logic [here](https://github.com/empirical-org/Empirical-Core/develop/services/QuillLMS/app/controllers/teachers/units_controller.rb/#L22-L26)), and that's addressed in [#](https://github.com/empirical-org/Empirical-Core/pull/13201).

This code is very old, and should be handled by not letting teachers assign packs with the same name, as in the code linked above. If there remains some edge case that gets around this once the new code is out, I think it's better to have code that results in two packs with the same name, rather than a pack that gets updated unexpectedly.

## HOW
Remove code that was doing weird stuff we don't want anymore.

### Screenshots

### Notion Card Links

Manually tested that assigning a pack with the same name as a weird ghost pack (due to the situation above) does not update the weird ghost pack, but creates a new one with the same name.

PR Checklist | Your Answer
------------ | -------------


brendanshean (approved):
Thank you for cleaning this code up.  Definitely feels better not seeing Updater call in a create action.

services/QuillLMS/app/controllers/teachers/units_controller.rb
```diff
@@ -8,19 +8,14 @@ class Teachers::UnitsController < ApplicationController
   before_action :authorize!
 
   def create
-    units_with_same_name = current_user.units_with_same_name(params[:unit][:name])
```
anathomical, line 11: This is such a crazy line of code when I think about it.  I'm glad you've killed it.

anathomical (approved):

PR #13206, opened 2025-08-18
worker to archive classroom units associated with archived classrooms

emilia-friedberg:
## WHAT
Add a worker to archive classroom units associated with archived classrooms, and clean up the logic in the hide_all_classroom_units function to avoid creating a massive number of workers.

## WHY
We had a bug for a long time that was making it so that classroom units did not automatically get archived when their classrooms did. This caused a bug in our system where we had "ghost units" floating around that were not visible in the teacher interface, but caused unexpected update behavior on activity pack assignment. The logic itself has been fixed, but now we need to clean up the data for 300,000 classrooms, necessitating a worker so we can these in parallel.

Re: the hide_all_classroom_units function: I reworked this logic a little bit so we create fewer `SaveUserPackSequencesWorkers`. In normal function, we actually shouldn't need to create any of these, because there's a callback on the `classroom` that will call `save_user_pack_sequences` on all of the `classroom_units` when the visibility of the classroom changes. However, since it is possible to run this function in isolation (as we are doing in this worker), I didn't want to remove it entirely. That said, the current implementation was creating a worker call for each student, for each classroom unit they were assigned to (already redundant, because the worker itself only cares about the classroom id and student id), and then doing the same thing all over again for each unit that got updated. This was creating a massive number of jobs. It's still creating a lot of jobs, but many fewer! However, if we think we should not be updating this functionality here, I'm open to discussing.

## HOW
Find all archived classrooms with at least one visible `classroom_unit` and run the `hide_all_classroom_units` function on it.

### Screenshots

### Notion Card Links

Ran it on staging.

PR Checklist | Your Answer
------------ | -------------


services/QuillLMS/app/workers/archive_classroom_units_worker.rb
```diff
@@ -0,0 +1,23 @@
+# frozen_string_literal: true
+
+class ArchiveClassroomUnitsWorker
+  include Sidekiq::Worker
+  sidekiq_options queue: SidekiqQueue::HIGH_MEMORY
+
+  def perform(classroom_ids)
```
anathomical: One risk about the way this is structured is that if any id from the batch errors, the entire job will end up back in the re-try queue, and the ones that succeeded will be processed again.  That shouldn't break any data, but will spend a bunch of resources on the items that have already been fixed.  An alternative approach is to make this one take a single `classroom_id` and to not batch them at the worker.  This will end up enqueuing 50 times more jobs (given a batch size of 50), but will make individual failures only retry themselves.

services/QuillLMS/spec/workers/archive_classroom_units_worker_spec.rb
```diff
@@ -0,0 +1,93 @@
...
+  before do
+    allow(Rails.logger).to receive(:error)
+  end
+
+  context 'when all classrooms exist and process successfully' do
+    it 'calls hide_all_classroom_units on each classroom' do
+      allow(Classroom).to receive(:unscoped).and_return(Classroom)
+      allow(Classroom).to receive(:find).with(classroom1.id).and_return(classroom1)
+      allow(Classroom).to receive(:find).with(classroom2.id).and_return(classroom2)
+      allow(Classroom).to receive(:find).with(classroom3.id).and_return(classroom3)
```
anathomical: I think you can put these in your `before` block on line 13 and not have to declare them in every `it`.

services/QuillLMS/spec/workers/archive_classroom_units_worker_spec.rb
```diff
@@ -0,0 +1,93 @@
...
+    end
+  end
+
+  context 'when a classroom does not exist' do
+    let(:classroom_ids) { [classroom1.id, 999999, classroom2.id] }
+
+    it 'logs error for missing classroom and raises exception with details' do
+      allow(Classroom).to receive(:unscoped).and_return(Classroom)
+      allow(Classroom).to receive(:find).with(classroom1.id).and_return(classroom1)
+      allow(Classroom).to receive(:find).with(999999).and_raise(ActiveRecord::RecordNotFound)
```
anathomical: It is possible that this is necessary, but I suspect that if you don't declare this and just let the code run normally, it'll fail to find the specified id and raise the error without your override.  That should allow the other `allows` here to be inherited from the higher-up `before` block if you take that suggestion.

anathomical (approved):
Some relatively minor organizational suggestions, but given that this is intended to be a one-off and only run manually where we can babysit it, I think it could go out as-written.

services/QuillLMS/app/workers/archive_classroom_units_worker.rb
```diff
@@ -0,0 +1,23 @@
...
+      classroom.hide_all_classroom_units
+    rescue => e
+      Rails.logger.error "Failed to archive classroom units for classroom #{classroom_id}: #{e.message}"
+      Rails.logger.error e.backtrace.join("\n")
+      failed_classrooms << { id: classroom_id, error: e.message }
+    end
+
+    return unless failed_classrooms.any?
+
+    raise "Failed to process #{failed_classrooms.size} classrooms: #{failed_classrooms.map { |f| "#{f[:id]} (#{f[:error]})" }.join(', ')}"
```
happythenewsad: To address Thomas's concern a different way, you could alternatively recursively call the worker here wit the failed classroom ids, instead of `raise`

services/QuillLMS/lib/tasks/archive_classroom_units.rake
```diff
@@ -0,0 +1,42 @@
...
+
+    puts "Enqueueing #{batches.size} jobs..."
+
+    batches.each_with_index do |batch_ids, index|
+      ArchiveClassroomUnitsWorker.perform_async(batch_ids)
+      puts "Enqueued batch #{index + 1}/#{batches.size} with #{batch_ids.size} classrooms"
+    end
+
+    puts 'All jobs enqueued! Monitor progress in Rails logs or Sidekiq web UI at /sidekiq'
+    puts 'Usage: bundle exec rake archived_classrooms:archive_units[200] to set batch size to 200'
```
happythenewsad: it's better practice to put usage messages at the start of a script (can therefore return early)

happythenewsad (approved):

emilia-friedberg:
@anathomical @happythenewsad re-requesting review here because I changed the logic in `hide_all_classroom_units` itself to reduce redundant worker calls -- @brendanshean, also curious about your thoughts there since you're the most familiar with the `SaveUserPackSequences` logic -- do we really need to be making these calls here at all, if the classroom is archived?

brendanshean (approved):
I originally put the  `classroom_units.each(&:save_user_pack_sequence_items)` and   `units.each(&:save_user_pack_sequence_items)` since the `update_all` calls weren't triggering the necessary callbacks.   Great work on cleaning up those redundancies!

services/QuillLMS/app/workers/archive_classroom_units_worker.rb
```diff
@@ -0,0 +1,27 @@
+# frozen_string_literal: true
+
+class ArchiveClassroomUnitsWorker
+  include Sidekiq::Worker
+  sidekiq_options queue: SidekiqQueue::HIGH_MEMORY
+
+  def perform(classroom_ids, retry_attempt)
```
happythenewsad: style note: `retry_attempt` scans as an integer variable. `retry_attempt?` would more accurately describe the intent

services/QuillLMS/app/workers/archive_classroom_units_worker.rb
```diff
@@ -0,0 +1,27 @@
...
+
+    classroom_ids.each do |classroom_id|
+      classroom = Classroom.unscoped.find(classroom_id)
+      classroom.hide_all_classroom_units
+    rescue => e
+      raise e if retry_attempt
+
+      Rails.logger.error "Failed to archive classroom units for classroom #{classroom_id}: #{e.message}"
+      Rails.logger.error e.backtrace.join("\n")
+      failed_classrooms << { id: classroom_id, error: e.message }
```
happythenewsad: The code isn't doing anything with `error:` here, so we could use an integer array instead of a hash array, for simplicity

happythenewsad (approved):
Couple optional things; no need to re-review

anathomical (approved):

PR #13208, opened 2025-08-19
ignore-assign-on-join-case

anathomical:
## WHAT
- Update verification code to not alarm on known edge case
- Add a little utility scope to StudentLearningSequenceActivity
## WHY
There's a known data situation caused by a since-fixed bug that allowed students to erroneously complete practice before they completed the associated the diagnostic.  We've decided that it makes more sense to show this data in Diagnostic Insights because the students really did do the practice, and accept that the timeline may look off if a teacher delves into it in-depth.

Because this is an accepted edge case, we shouldn't send an alarm when this happens so that we can reduce noise around alerts.
## HOW
Add an additional condition to the condition chain so that even if practice does happen before a diagnostic, if that practice is associated with an `assign_on_join` `ClassroomUnit`, we won't alarm.

### Notion Card Links
https://www.notion.so/quill/Diagnostic-Insights-Update-data-alarm-to-ignore-assign_on_join-cases-254d42e6f941805097d4f3a9c0f632b9?source=copy_link

- Feed a bunch of known classrooms with known students in this state through the updated service and confirm that it no longer marks them as in error

PR Checklist | Your Answer
------------ | -------------


dandrabik (approved):
LGTM. Thanks for moving this along!

PR #13209, opened 2025-08-20
revisions to existing slides

emilia-friedberg:
## WHAT
Revisions to existing slide paths, including:

- removing the slide container and making the navigation stick to the bottom of the page
- copy updates
- making some slides that were previously unique to one path or another shared
- image updates

## WHY
We made design changes to these slides after their last iteration. These designs are final. There will be one more PR forthcoming with the High School path, and then eventually Social Studies, but that's still in design.

## HOW
CSS/React.

### Screenshots

### Notion Card Links

Tested both existing slide paths locally at full width and down to 325px.

PR Checklist | Your Answer
------------ | -------------
Have you deployed to Staging? | Not yet- will deploy after next PR


services/QuillLMS/client/app/bundles/Teacher/components/accounts/new/onboarding_tour/shared_slides/our_aim_slide.tsx
```diff
@@ -0,0 +1,20 @@
+import * as React from 'react'
+
+import { sharedSlidesBaseSrc } from '../constants'
+
+const OurAimSlide = ({ slideGroupClassName, }) => (
+  <section className={`slide ${slideGroupClassName} our-aim-slide`} key="our-aim-slide">
+    <h1>Our aim is to support you, not replace you</h1>
+    <div className="flex-wrapper">
+      <div>
+        <p className="quill-is">“<b>Quill is one of the best tools</b> that I, as an ELA teacher, have found that helps my students <b>become better writers</b>.”</p>
```
anathomical: I don't know that this is necessarily the wrong approach, but using a `className` that is linked to the content of the section feels a little bit risky.  Potential future changes that this approach could make more complicated:
- We change one of the quotes and the `className` no longer refers to text in the quote
- We re-order the quotes without changing any content, but we'd need to update the CSS rather than just changing which element has which `className`
- We add additional quotes to the list.  This actually may not be worth trying to fix more generally since some of the sizing rules probably depend on total count of quotes, but if there was a way to make this more generic that'd be cool.

anathomical (approved):
Style comment about `className`s, but overall this looks good to me.

eadams17 (approved):

services/QuillLMS/client/app/bundles/Teacher/components/accounts/new/onboarding_tour/shared_slides/our_aim_slide.tsx
emilia-friedberg: Yeah, I thought about this too, but using classNames like 'first-quote' or :nth selectors pose other issues for future theoretical changes. I agree in general that classNames should not depend on content, but I feel okay about it in this very specific instance.

PR #13214, opened 2025-08-20
Classlink: fan out workers for rostering sync

happythenewsad:
## WHAT
- have DistrictsImporter call a Sidekiq job for each district 
- refactor specs to make them DRYer, with less overstubbing
- reduces the parameter complexity of DistrictImporter, to `tenant_id` only

## WHY
- so that individual jobs don't take too much memory, and so that district-level sync failures do not propagate to successive districts 
- for legibility; for more code coverage
- easier to manually fire off syncs by district (which is useful when onboarding new districts). Also encapsulates bearer token authentication within individual workers, which is a better separation of concerns

### Screenshots

### Notion Card Links
https://www.notion.so/quill/Product-Board-30f97e4eb01246dbb8264253fb385073?p=255d42e6f9418055b5dfd20e3e6c2c3b&pm=s

- [x] SSO with ClassLink
- [x] run sync; observe multiple queued jobs that complete without error

PR Checklist | Your Answer
------------ | -------------


services/QuillLMS/app/services/classlink_integration/district_importer.rb
```diff
@@ -61,40 +65,57 @@ def run
...
 
-    private def teacher_enrollments
-      @teacher_enrollments ||= enrollments.select { |enrollment| enrollment[:role] == 'teacher' }
-    end
-
-    private def student_enrollments
-      @student_enrollments ||= enrollments.select { |enrollment| enrollment[:role] == 'student' }
-    end
+    private def fetch_district_data
+      proxy_response = Client.get(
```
happythenewsad, line 115: Although this API is already called in DistrictsImporter, I prefer to call it again here so that we don't persist bearer secrets in sidekiq job args.

There is no API support for requesting a single `application` resource by tenant id

Docs: https://apisandbox.classlink.com/reference/districts

services/QuillLMS/app/services/classlink_integration/district_importer.rb
```diff
@@ -61,40 +67,57 @@ def run
...
+        request: 'applications',
+        authorization: ENV['CLASSLINK_ROSTER_PROXY_BEARER_SECRET']
+      )
+      district_hash = proxy_response['applications'].find { |app| app[TENANT_ID].to_i == tenant_id }
```
dandrabik: Note, you might want a `&.to_i` here or it will raise with a vaguer error than the one below

services/QuillLMS/app/services/classlink_integration/districts_importer.rb
```diff
@@ -4,70 +4,38 @@ module ClasslinkIntegration
...
-    TENANT_NAME = 'tenant_name'
 
     def run
       proxy_response = Client.get(
         request: 'applications',
         authorization: ENV['CLASSLINK_ROSTER_PROXY_BEARER_SECRET']
       )
 
-      proxy_response['applications'].each do |district_hash|
+      valid_application_list(proxy_response).each_with_index do |district_hash, idx|
```
dandrabik: If you are doing the call to the API in the DistrictImporter, I believe you could just loop through `allowed_districts` here, e.g.

```ruby
allowed_district_tenant_ids.each.with_index do |district_tenant_id, index|
  perform_in_seconds = (index * SMEAR_RATE_IN_SECONDS).seconds
  
  DistrictImporterWorker.perform_in(perform_in_seconds, district_tenant_id)
end
```
I believe this current code is just checking everything twice, once in the parent job, and once in the child, and I'm not sure there is a value add to that (unless I am missing something).

services/QuillLMS/app/workers/classlink_integration/district_importer_worker.rb
```diff
@@ -0,0 +1,16 @@
...
+  class DistrictImporterWorker
+    include Sidekiq::Worker
+    sidekiq_options queue: SidekiqQueue::HIGH_MEMORY
+
+    class RunError < StandardError; end
+
+    def perform(tenant_id)
+      DistrictImporter.run(tenant_id:)
+    rescue => e
+      ErrorNotifier.report(RunError.new, { message: "#{e.message} #{e.backtrace}" })
```
dandrabik, line 13: So this `rescue` will never let the job retry. Maybe that's ok for now, but it might make imports fail semi-silently.

Let's make sure this error goes to a Slack channel or some place very visible while this `rescue` is in place.

dandrabik (changes requested):
So this overall looks. I requested a few changes that I'd like to see implemented (so flagging as 'Request Change')
1. Removing the API call from the parent job (or let's discuss if there's a reason to leave it)
2. (non-code): Making sure that district failures send a very visible alert to a Slack channel that is monitored.

services/QuillLMS/app/services/classlink_integration/districts_importer.rb
happythenewsad: Yes, this is a legacy pattern no longer needed (and will eventually get replaced again, once the partnerships tool is live)

services/QuillLMS/spec/services/classlink_integration/districts_importer_spec.rb
```diff
@@ -4,136 +4,65 @@
...
-    end
-
-    it 'processes each valid district' do
-      instance_double = double('ClasslinkInstance', district_id: 1, tenant_id: 456)
-      expect(importer).to receive(:valid_district?).twice.and_return(true)
-      expect(described_class).to receive(:find_or_create_district_and_instance).twice.and_return(instance_double)
-      expect(ClasslinkIntegration::DistrictImporter).to receive(:run).twice
-      importer.run
+      allow(ClasslinkIntegration::Client).to receive(:get).and_return(valid_response)
+      allow(ENV).to receive(:fetch).and_return('456,999')
```
anathomical: I don't think this is strictly necessary, but I think this `allow` call is more communicative of expectations if it has a `with` call attached so that we can understand which key this is intended to impact.  Alternatively, this might be a good use-case for `stub_const('ENV', { key => '456,999' })`

Given that this value changes in a number of different contexts, you might even generalize this slightly:
```ruby
let(:env_tenant_ids) { '456,999' }

before do
  stub_const('ENV', { key => env_tenant_ids })
end
```
Which would allow you to just use `let(:env_tenant_ids) { '123' }` or whatever to change this out rather than duplicating the stub logic.

services/QuillLMS/app/services/classlink_integration/district_importer.rb
```diff
@@ -2,19 +2,25 @@
...
+
+    class RunError < StandardError; end
+
+    APPLICATION_ID = 'application_id'
+    BEARER = 'bearer'
+    ENABLED = 'enabled'
+    NAME = 'name'
+    ONEROSTER_APPLICATION_ID = 'oneroster_application_id'
+    TENANT_ID = 'tenant_id'
+    TENANT_NAME = 'tenant_name'
```
anathomical, line 17: I super-appreciate these being constantized.

services/QuillLMS/app/workers/classlink_integration/district_importer_worker.rb
```diff
@@ -11,6 +11,7 @@ def perform(tenant_id)
       DistrictImporter.run(tenant_id:)
     rescue => e
       ErrorNotifier.report(RunError.new, { message: "#{e.message} #{e.backtrace}" })
+      raise e
```
happythenewsad, line 14: This allows us to apply a specific rule set to the Sentry -> Slack pipeline

anathomical (approved):
I like the more atomic organization: having workers that take fewer arguments does feel cleaner to me.  A couple of style-focused comments, but I think this could go out as-is.

services/QuillLMS/app/services/classlink_integration/district_importer.rb
```diff
@@ -61,40 +67,57 @@ def run
...
-      @teacher_enrollments ||= enrollments.select { |enrollment| enrollment[:role] == 'teacher' }
-    end
-
-    private def student_enrollments
-      @student_enrollments ||= enrollments.select { |enrollment| enrollment[:role] == 'student' }
-    end
+    private def fetch_district_data
+      proxy_response = Client.get(
+        request: 'applications',
+        authorization: ENV['CLASSLINK_ROSTER_PROXY_BEARER_SECRET']
```
dandrabik: Small comment. You should consider assigning this ENV to a local constant. This makes it easier in tests to `stub_const` locally, and not have to stub a global list of parameters such as `ENV`, e.g.
```ruby
# In this class
BEARER_SECRET = ENV['CLASSLINK_ROSTER_PROXY_BEARER_SECRET']

# in a test
stub_const('ClasslinkIntegration::DistrictImporter::BEARER_SECRET', '1234')

services/QuillLMS/spec/services/classlink_integration/districts_importer_spec.rb
dandrabik: I added a way above to make this a little more scoped. We should avoid stubbing the global `ENV` in tests

dandrabik (approved):
Overall, this looks good. Thanks for making the changes. I left one note that I'd like to see implemented (but don't need to re-review) to avoid stubbing the global 'ENV' in a test.

PR #13216, opened 2025-08-21
update-merge-student-accounts

anathomical:
## WHAT
Update teacher fix to also move student_learning_sequences
## WHY
This code already transfers assigned `user_id` values for `activity_sessions`.  Failing to also transfer ownership of `student_learning_sequences` can result in cases where a `StudentLearningSequence` is owned by one user account and associated with `ActivitySessions` owned by a different user account.  We don't want that.
## HOW
Add a new method to the `Student` concern that is called a part of `merge_student_account`.  It has two logic paths:
1. Attempt to change the `user_id` for any `StudentLearningSequence` that is associated with the merge source to match the merge target
2. If that fails because the merge target already has a `StudentLearningSequence` that matches (this can happen if both user accounts are in the same classroom prior to merger), transfer the associated `StudentLearningSequenceActivity` records from the source `StudentLearningSequence` to the one owned by the target

### Notion Card Links
https://www.notion.so/quill/Update-Student-merge_student_account-to-find-and-re-assign-StudentLearningSequences-254d42e6f9418060a3dcc764f77d6fa9?source=copy_link

- On staging create a new three test students in the same classroom
- Assign the pre diagnostic to students A and B, but not to C
- Use the teacher-fix to merge student B into C
- Confirm that B no longer has StudentLearningSequences assigned
- Use the teacher-fix to merge student A into student C
- Confirm that A no longer has StudentLearningSequences assigned
- Confirm that the StudentLearningSequenceActivity that was attached to the student A data has been cleaned up

PR Checklist | Your Answer
------------ | -------------


dandrabik (approved):
LGTM! Thanks for closing this out.

PR #13218, opened 2025-08-21
preload assets for onboarding tour and final revisions

emilia-friedberg:
## WHAT
Add some logic that preloads the assets for the onboarding tour, so users don't see a lag, and make final copy changes.

## WHY
We were noticing that sometimes the images took a noticeable amount of time to render, which isn't ideal here.

## HOW
Make requests for all assets when the index slide loads, so they are cached by the time the user actually gets to the slide. Consolidate all image imports to one place for consistency, and also to act as an indicator for future changes that we should be putting them in this centralized place.

### Screenshots

### Notion Card Links
This came out of QA with Jack and Katie, but I decided to make a separate PR.

Tested locally, having Jack and Katie QA on staging.

PR Checklist | Your Answer
------------ | -------------


anathomical (approved):
This seems like a solid approach.  Probably not worth pursuing now, but I could see us wanting a more generic version of this pre-loader logic to allow us to pre-load assets for other projects when that becomes necessary, and this general pattern does feel like it can be expanded easily when we need to do that.

happythenewsad (approved):
This looks like a good solution for flows where we can anticipate a user loading lots of path-dependent assets, with high probability. 

Question: what is your motivation for putting the Preloader in bundles/Teacher rather than bundles/Shared?

emilia-friedberg:
> This looks like a good solution for flows where we can anticipate a user loading lots of path-dependent assets, with high probability.
> 
> Question: what is your motivation for putting the Preloader in bundles/Teacher rather than bundles/Shared?

@happythenewsad since this functionality is not currently being shared, and I don't anticipate its near-term utilization in other places, it lives in the folder where it's used. It could be hoisted to Shared, but I think that's potentially a premature optimization, and would be easy to make in the future if we need it.

PR #13219, opened 2025-08-21
add-merge-students-change-log

anathomical:
## WHAT
Add a ChangeLog save when executing the merge_student_accounts fix
## WHY
We've discovered at least one case where we need to backfill some data based on the existence of these mergers, so we should start tracking when they happen in case this comes up again
## HOW
Record a `ChangeLog` record when this particular Teacher Fix is executed successfully

### Notion Card Links
https://www.notion.so/quill/Add-code-to-record-ChangeLog-records-for-merge_student_accounts-TeacherFix-254d42e6f941800e8060e85c5d3c5de7?source=copy_link

- Merge students on staging
- Confirm that there is a new ChangeLog tied to those accounts

PR Checklist | Your Answer
------------ | -------------


brendanshean (approved):
Seems like a good thing to log.

PR #13222, opened 2025-08-21
Evidence: Add Tertiary Feedback Layer

dandrabik:
## WHAT
In Evidence, we currently use an LLM to rewrite `feedback` if it is flagged as a repeat (we make it shorter and add a highlight). We refer to that as `secondary_feedback`. This PR adds a process to re-write `secondary_feedback` if a student triggers `secondary_feedback` twice in a row. We are calling this `tertiary_feedback`. 

## WHY
In our session analysis, we found sessions in which the student was getting the same `secondary_feedback` over and over, because that LLM translation from feedback to secondary_feedback has identical or near-identical results. 
## HOW
`tertriary_feedback` uses the `secondary_feedback` as a base and varies its opening sentences from an approved list and uses an LLM to rephrase the core question posed in the `secondary_feedback`. It also re-uses its highlight.

To make sense of the code, start with `TertiaryFeedback::Creator`, then see how it plugs into `SecondaryFeedback::ResponseBuilder` and then `GenAi::ResponseBuilder`

### Screenshots
Two sessions showing this:
### Session one 
**(these previous would have likely given identical feedback twice in a row)**
---
Screenshot 2025-08-21 at 5 26 01 PM
Screenshot 2025-08-21 at 5 26 18 PM
---
### Session 2
Screenshot 2025-08-21 at 5 29 07 PM
Screenshot 2025-08-21 at 5 29 18 PM

### Notion Card Links
https://www.notion.so/quill/Secondary-Feedback-prompt-review-242d42e6f941806fbba1d5af460084fb

PR Checklist | Your Answer
------------ | -------------


services/QuillLMS/lib/tasks/gen_ai_secondary_feedback.thor
```diff
@@ -7,16 +7,17 @@ class GenAiSecondaryFeedback < Thor
 
   Result = Data.define(:prompt_id, :stem, :entry, :feedback, :secondary_original, :secondary_generated, :similarity, :highlight_match, :highlight_key, :highlight_original, :highlight_generated, :response_json)
 
-  # bundle exec thor gen_ai_secondary_feedback:feedback_test 'description' 2
+  API = Evidence::Gemini::Chat
+  MODEL = API::DEFAULT_MODEL
```
anathomical, line 11: I like moving these to constants.

services/QuillLMS/engines/evidence/spec/services/evidence/gen_ai/tertiary_feedback/creator_spec.rb
```diff
@@ -0,0 +1,209 @@
...
+          .to receive(:run)
+          .with(prompt:, question_sentence: 'What caused the war?')
+          .and_return('REPHRASE_PROMPT')
+
+        expect(described_class::API)
+          .to receive(:run)
+          .with(prompt: 'REPHRASE_PROMPT')
+          .and_return(%({"rephrase":"Rephrased question goes here."}))
+
+        expect(subject).to include('Rephrased question goes here.')
```
anathomical: I tend to think repeating static values in set-up and expectations is a good signal that it may be worth extracting into a variable so that the static declaration only has to happen once.

anathomical (approved):
A minor style suggestion, but feel free to take-or-leave it.

services/QuillLMS/engines/evidence/spec/services/evidence/gen_ai/tertiary_feedback/creator_spec.rb
dandrabik: Good point. I've done another pass on this spec file to replace the string expectations with `let` variables.

services/QuillLMS/engines/evidence/app/services/evidence/character_distance_checker.rb
```diff
@@ -8,6 +8,7 @@ class CharacterDistanceChecker < ApplicationService
 
     DEFAULT_THRESHOLD = 3
 
+    # return true if any of the text_array items are within threshold characters of text
```
brendanshean: I don't think this comment is necessary as the `run` method in this service object feels pretty self-explanatory.

services/QuillLMS/engines/evidence/app/services/evidence/gen_ai/secondary_feedback/prompt_builder.rb
```diff
@@ -20,14 +20,23 @@ class PromptBuilder < Evidence::GenAi::PromptBuilder
         private def default_template = DEFAULT_TEMPLATE
         private def template_folder = TEMPLATE_FOLDER
 
-        private def primary_secondary_examples = markdown_table_rows(feedback_data_tuples)
-        private def feedback_data_tuples = feedback_data.map { |f| [f.primary, f.secondary] }
+        private def primary_secondary_examples = markdown_table_rows(feedback_example_tuples)
+        private def feedback_example_tuples = feedback_examples.map { |f| [f.primary, f.secondary] }
```
brendanshean, line 24: I like `pluck` here since it gets rid of the block and temp variables:
`private def feedback_example_tuples = feedback_examples.pluck(:primary, :secondary)`

brendanshean (approved):
This is a nice solution for tertiary feedback.

services/QuillLMS/engines/evidence/app/services/evidence/gen_ai/secondary_feedback/prompt_builder.rb
dandrabik: This isn't an ActiveRecord finder, so I don't believe that pluck is available here.

brendanshean: Pluck is for `Enumerable` too:

https://guides.rubyonrails.org/active_support_core_extensions.html#pluck

PR #13223, opened 2025-08-22
Classlink sync guardrails; quality of life updates

happythenewsad:
## WHAT
1. Remove allowlist ENV var and accompanying logic.
2. Make sidekiq smear rate an env var 
3. ClasslinkInstances should be default-disabled upon creation (and not runnable if disabled)

## WHY
1. We're pivoting to ClasslinkInstance.enabled instead. This will be more easily managable by partnerships staffers in a CRUD view later on. 
2. so we can experiment with different smear rates without a code deploy
3. So that enabling a given district's sync is up to the discretion of Partnerships 

## HOW
Note: this must be executed before deployment:
```ruby 
ClasslinkInstance.all.each{|x| x.update!(enabled: false)}
```

### Screenshots

### Notion Card Links
https://www.notion.so/quill/add-safeguard-for-district-matching-248d42e6f94180da8d85e2d4f465dd35

- [x] create a stub instance & district by signing in as a first-time, pre-rostering, Classlink SSO user
- [x] ensure that the created instance is disabled 
- [x] point the instance to the existing District
- [x] run sync. Ensure only enabled instances enqueue sync jobs

PR Checklist | Your Answer
------------ | -------------


services/QuillLMS/app/services/classlink_integration/districts_importer.rb
```diff
@@ -21,10 +19,6 @@ def run
       raise e
     end
 
-    private def allowed_districts
-      @allowed_districts ||= ALLOWLIST
-        .split(',')
-        .map(&:to_i)
-    end
+    private def allowed_classlink_instances = ClasslinkInstance.enabled
```
happythenewsad, line 22: I get that this method is just an alias, but it signals that only enabled instances are allowed to be processed.

anathomical: I am not opposed to these types of alias when they provide additional context, as you describe here.

services/QuillLMS/spec/services/classlink_integration/district_importer_spec.rb
```diff
@@ -188,6 +188,16 @@ def mock_api_response(data, count = 0)
       end
     end
 
+    context 'when classlink instance is not enabled' do
+      before do
+        classlink_instance.update!(enabled: false)
+      end
```
anathomical, line 194: This is purely stylistic, so take or leave it, but I personally find `let` slightly more clear than `before ... update`.  Again, this is fine as-is, but I'd write this as
```
let(:enabled) { true }
let!(:classlink_instance) { create(:classlink_instance, tenant_id:, district_id:, enabled:) }

...
context 'when classlink instance is not enabled' do
  let(:enabled) { false }
  it
    ...
  end
end
```
For whatever reason, my brain parses this more quickly as an "initial condition", while `before ... update` is initially parsed in my brain as a change in condition where the change my be part of the test.

services/QuillLMS/spec/services/classlink_integration/districts_importer_spec.rb
```diff
@@ -4,27 +4,21 @@
...
-  end
 
   describe '#run' do
     before do
-      stub_const('ClasslinkIntegration::DistrictsImporter::ALLOWLIST', '456,999')
+      enabled_instances = [
+        create(:classlink_instance, tenant_id: 456, enabled: true),
+        create(:classlink_instance, tenant_id: 99, enabled: true)
+      ]
+      allow(ClasslinkInstance).to receive(:enabled).and_return(enabled_instances)
```
anathomical: Do we actually need to `allow(ClasslinkInstance)` here?  We actually are creating the database records with `create`, and we're not `expect`ing the instances to do anything in particular, so won't this run successfully without the `allow`?

anathomical (approved):
A couple of style comments, but this could go out as-is.  And I'll reiterate here: I think that aliasing the `ClasslinkInstance.enabled` call makes perfect since in this case, so I like that you used it.

services/QuillLMS/app/services/classlink_integration/constants.rb
```diff
@@ -3,6 +3,6 @@
 module ClasslinkIntegration
   module Constants
     APP_SETTING = 'classlink_integration_enabled'
-    DISTRICT_TENANT_ID_ALLOWLIST = 'CLASSLINK_DISTRICT_TENANT_ID_ALLOWLIST'
+    SMEAR_RATE_IN_SECONDS = 'CLASSLINK_SMEAR_RATE_IN_SECONDS'
```
emilia-friedberg, line 6: Just out of curiosity, what's a smear rate? Google turns up information about cancer screenings.

emilia-friedberg (approved):

services/QuillLMS/app/services/classlink_integration/constants.rb
happythenewsad: @emilia-friedberg   it's jargon for "programatically scheduling async jobs in the future to avoid performance bottlenecks" 

See also: https://github.com/empirical-org/Empirical-Core/blob/850a23bd160f89f5a924f3fccd0293c3ae5909ec/services/QuillLMS/app/services/learn_worlds_integration/sync_orchestrator.rb#L5

If it were up to me I would have picked a different word

services/QuillLMS/spec/services/classlink_integration/districts_importer_spec.rb
happythenewsad: Correct; I removed this line

PR #13228, opened 2025-08-26
routing updates for world history pages

emilia-friedberg:
## WHAT
Address some routing issues we missed in original QA for the new World History page -- mostly more sophisticated logic where we determine where to route a user based on content they've assigned.

## WHY
Jamie noticed this in prod.

## HOW
Introduce logic in various places to determine which page a user should be routed to.

### Screenshots

### Notion Card Links
https://www.notion.so/quill/World-History-1750-CE-Present-Course-Page-d65706c3be2343bda7c98518d109c8e2?source=copy_link#25ad42e6f94180909152f775bef8e049

Tested locally -- a lot of relevant data doesn't exist on staging, so Jamie will double check on prod.

PR Checklist | Your Answer
------------ | -------------


happythenewsad (approved):
Let's ship this in order to fix the user facing issues, but later, there's opportunity here to handle state in data structures (hashes), rather than if/elses.

PR #13229, opened 2025-08-26
fix-scoring-edge-case

anathomical:
## WHAT
Fix fill-in-the-blank scoring calculation logic
## WHY
We want the scoring to reflect the spec, and contrary to the original implementation, FITB scoring has a different rate of score fall-off than default question types, so needs to use a tweaked lookup table
## HOW
- Add a correct lookup table for fill in the blank scoring
- Tweak scoring logic to use the new table when appropriate
- Add an extra condition to check for cue count when determining whether to use special scoring since once a fill in the blank question has six or more cues, we use the default scoring

### Notion Card Links
https://www.notion.so/quill/Skills-Report-2-0-190d42e6f941801bbab8e37e97e68147?source=copy_link#1bad42e6f94180768cc7f6ac3018b74c

- Updated specs to include edge cases
- Look at report for known test user to confirm that their score is being correctly calculated

PR Checklist | Your Answer
------------ | -------------


happythenewsad (approved):

services/QuillLMS/app/services/skills_report/question_response_scorer.rb
```diff
@@ -33,6 +40,12 @@ def run
     private def optimal = response.optimal
 
     private def default_scoring = optimal ? DEFAULT_MAPPING[attempts] : 0
-    private def fitb_scoring = optimal && attempts < cue_count ? DEFAULT_MAPPING[attempts] : 0
+    private def fitb_scoring = optimal && attempts < cue_count ? FITB_MAPPING[attempts] : 0
```
dandrabik: Just a note, that both fitb_scoring and default_scoring have the potential to be `nil` with weird/bad data (if the key is not in the mapping). Consider defensive programming with a fetch with a default, e.g. `FITB_MAPPING.fetch(attempts, 0)` to ensure the score is a number.

dandrabik (approved):
One note about guarding against bad data, but LGTM! Thanks for figuring this out.

PR #13230, opened 2025-08-26
skills report: performance breakdown: by student

emilia-friedberg:
## WHAT
Build out the Skills Report: Performance Breakdown: By Student page.

## WHY
This is the first complete page of the Skills Report! Note: actual assignment logic is not implemented yet.

## HOW
I copied the backend endpoints to a new controller, since they were located in the CMS and those requests automatically fail for non-staff requests. I also added an "id" attribute to those responses since we need it to track selected students. On the frontend, just a bunch of React and CSS.

### Screenshots
Screenshot 2025-08-26 at 3 08 45 PM

### Notion Card Links
https://www.notion.so/quill/Skills-Report-Performance-Breakdown-By-Student-24ed42e6f94180bdb6f0f7422f417767?source=copy_link

Tested locally, had Jack QA on staging -- Peter Sharkey is out this week but I'll have him check next week.

PR Checklist | Your Answer
------------ | -------------


services/QuillLMS/app/controllers/skills_reports_controller.rb
```diff
@@ -0,0 +1,37 @@
+# frozen_string_literal: true
+
+class SkillsReportsController < ApplicationController
+  before_action :authorize_request, only: [:performance_breakdown_by_student]
+
+  def performance_breakdown_by_student
+    data = SkillsReport::ClassroomScoresByUser.run(
+      classroom_id:,
+      concept_id: use_sub_concept_id? ? params[:sub_concept_id].to_i : params[:concept_id].to_i,
```
happythenewsad: how about 
```ruby
before_action :set_ids 

private def set_ids 
  @concept_id = params[:concept_id]
  @sub_concept_id = params[:sub_concept_id]
end 
```

This would clean up the six references of `params[:[sub_]concept_id]`, making the code easier to parse

you could also do some validation checking in the before_action, returning a 4xx early, if, for example, concept_id is nil

services/QuillLMS/app/controllers/skills_reports_controller.rb
```diff
@@ -0,0 +1,37 @@
...
+    )
+
+    render json: { data: }
+  end
+
+  private def use_sub_concept_id?
+    params[:sub_concept_id].present? && Concept.exists?(id: params[:sub_concept_id].to_i, parent_id: params[:concept_id].to_i)
+  end
+
+  private def classroom_id = params.permit(:classroom_id)[:classroom_id].to_i
```
happythenewsad, line 38: unguarded access to classroom_id here

services/QuillLMS/client/app/bundles/Teacher/components/progress_reports/skills_report/performance_breakdown/by_student.tsx
```diff
@@ -0,0 +1,460 @@
...
+  });
+
+  useStickyTableHeader({
+    wrapperRef,
+    tableRef,
+    enabled: true,
+    dependencies: [data],
+  });
+
+  const formatTime = (seconds: number) => {
```
happythenewsad, line 159: how about renaming it to `secondsToStringHoursAndMinutes`? It's more easily searchable for future features that may want to recycle this logic

services/QuillLMS/client/app/bundles/Teacher/components/progress_reports/skills_report/performance_breakdown/index.tsx
```diff
@@ -151,62 +152,38 @@ const PerformanceBreakdown = ({ practiceScope, setPracticeScope, classroom, open
...
+    <main className="performance-breakdown">
       <header>
         <h1>
           Performance Breakdown
           {byStudent ? <span>/ By Student</span> : <span>/ By Skill</span>}
           {selectedSkillCategory ? <span>/ {selectedSkillCategory.name}</span> : null}
         </h1>
         <div className="links">
-          <a className="quill-button outlined small grey focus-on-light" href='' rel="noopener noreferrer" target="_blank">
+          <a className="quill-button outlined small grey focus-on-light" href='https://support.quill.org/en/articles/12062553-how-do-i-read-the-skills-report-quill-premium' rel="noopener noreferrer" target="_blank">
```
happythenewsad, line 186: We're already importing constants from `./shared`; this could be an imported constant as well

services/QuillLMS/app/controllers/skills_reports_controller.rb
anathomical: You could even take this a step further by also creating a method to handle the ternary logic for you:
```ruby
private def concept_id = use_sub_concept_id? ? @sub_concept_id : @concept_id
```

services/QuillLMS/app/controllers/skills_reports_controller.rb
```diff
@@ -0,0 +1,37 @@
+# frozen_string_literal: true
+
+class SkillsReportsController < ApplicationController
+  before_action :authorize_request, only: [:performance_breakdown_by_student]
+
+  def performance_breakdown_by_student
+    data = SkillsReport::ClassroomScoresByUser.run(
+      classroom_id:,
+      concept_id: use_sub_concept_id? ? params[:sub_concept_id].to_i : params[:concept_id].to_i,
+      all_history: ActiveModel::Type::Boolean.new.cast(params[:all_history])
```
anathomical: This is purely stylistic, so take or leave this, but I'd probably extract this into a one-line method, too:
```ruby
private def all_history = ActiveModel::Type::Boolean.new.cast(params[:all_history])
```
That allows you to just make your method calls with `run(..., all_history:)` in both places rather than having the casting logic repeated.

services/QuillLMS/client/app/bundles/Teacher/components/progress_reports/skills_report/performance_breakdown/__tests__/by_student.test.tsx
```diff
@@ -0,0 +1,282 @@
...
+    expect(calledUrl).toContain('all_history=true');
+  });
+
+  it('renders student data in table', async () => {
+    renderComponent();
+
+    await waitFor(() => {
+      expect(screen.getByText('Alice Smith')).toBeInTheDocument();
+      expect(screen.getByText('Bob Johnson')).toBeInTheDocument();
+      expect(screen.getByText('Charlie Brown')).toBeInTheDocument();
```
anathomical: I think that it may not be worth the extra work in this case given that we only test this data in the one place, but I always get worried when I type a string literal in two different places worried that I might not get the same value in both places, and so like to do it with a shared variable like 'student1Name' or something that's used both here and in `mockStudentData`.

services/QuillLMS/client/app/bundles/Teacher/components/progress_reports/skills_report/performance_breakdown/__tests__/by_student.test.tsx
```diff
@@ -0,0 +1,282 @@
+import * as React from 'react';
```
anathomical, line 1: I just wanted to flag that I feel like this is a super-useful test that does a great job of communicating how we expect to interpret API data, which I love.

services/QuillLMS/client/app/bundles/Teacher/components/progress_reports/skills_report/performance_breakdown/by_student.tsx
anathomical: We actually have a pretty similar piece of logic in `client/app/bundles/Teacher/helpers/studentReports.tsx` in a function called `getTimeSpent`.  It's not exactly the same (it uses 'hrs' instead of 'h', and 'mins' instead of 'm`, for instance), and it's just the one other place right now, so it may not be worth abstracting, but I wonder if it's worth trying to reuse or share that code somehow.

services/QuillLMS/spec/controllers/skills_reports_controller_spec.rb
```diff
@@ -0,0 +1,125 @@
...
+
+      context 'with valid concept_id only' do
+        it 'calls SkillsReport::ClassroomScoresByUser with concept_id' do
+          expect(SkillsReport::ClassroomScoresByUser).to receive(:run)
+            .with(classroom_id: classroom.id, concept_id: concept.id, all_history: false)
+            .and_return([])
+
+          get :performance_breakdown_by_student, params: valid_params
+          expect(response).to have_http_status(:ok)
+          expect(JSON.parse(response.body)['data']).to eq([])
```
anathomical: The `JSON.prase(response.body)` call is used in a couple of places, so I'd be tempted to just do `let(:json_response) { JSON.parse(response.body) }` and then do your tests around that value: `expect(json_response['data']).to eq([])`

anathomical (approved):
Some take or leave style stuff, but overall this looks really good.

PR #13231, opened 2025-08-26
FeedbackHistory performance fix.

dandrabik:
## WHAT
Fix some queries to FeedbackHistory to include `prompt_type='Evidence::Prompt'` to make sure it uses the index. (we should get rid of `prompt_type` long-term)
## WHY
If you only join on `prompt_id` it doesn't use the `(prompt_type, prompt_id)` index and is very slow.
## HOW
Add a `left_join_prompts` scope and use `for_prompt` scope where appropriate.
### Screenshots
Before for sample query:

Screenshot 2025-08-26 at 12 42 51 PM

After for sample query:
Screenshot 2025-08-26 at 12 42 57 PM
### Notion Card Links
https://www.notion.so/quill/Student-sessions-taking-a-long-time-to-load-25ad42e6f941804aa662d6791b8b3607

Confirmed the page that wasn't working on prod loaded quickly on staging.

PR Checklist | Your Answer
------------ | -------------
Have you added and/or updated tests? |  No, optimization, no functionality change


happythenewsad (approved):
Nice refactor.

It would be good to also QA on a query that has a non-null result set

dandrabik:
> Nice refactor.
> 
> It would be good to also QA on a query that has a non-null result set

Agreed. That screenshot was just confirming an improvement in timing and the query changing properly. I've done further QA to see results in the console and on staging.

PR #13232, opened 2025-08-26
fix-skills-report-timespent

anathomical:
## WHAT
- Tweak logic that calculates `Activity.question_count` to handle Diagnostic activities with `titleCards` properly
- Replace per-student queries for timespent with a generalized roll-up at the main data query so `timespent` can be part of the `Response` that comes from the database
- Update the template code that displays timespent to not show hours or minutes with decimals
## WHY
- This logic currently counts all entries in `data['questions']`, but some of those entries are for `TitleCard` records, not `Questions`, but we only want to count actual questions
- The old approach was highly non-performant as it made new database queries for every row of aggregation which could add seconds to processing time, but the actual calculation is not that complicated so doing it in the initial query is more performant
- We want this information to be legible
## HOW
- Modify this code to not just count the entries in `data['questions']`, but to actually count the number of those entries that correspond to actual `Question` records
- Add a `SELECT` value that calculates the timespent for each question Response by dividing `activity_sessions.timespent` by `activities.question_count` assuming that each question takes an equal proportion of the total activity play-through time
- Tweak the math and rounding of before displaying the numbers

### Notion Card Links

- On staging confirm that all users who have practice have timespent
- Confirm that no one who has no practice has any timespent
- Confirm that queries run as quickly as they did before this change

PR Checklist | Your Answer
------------ | -------------


services/QuillLMS/app/services/skills_report/user_responses_processor.rb
```diff
@@ -21,11 +21,16 @@ def run
...
     private def non_empty_responses = @non_empty_responses ||= responses.filter { |r| r.activity_session_id.present? && r.question_uid.present? }
-    private def unique_answered = @unique_answered ||= non_empty_responses.uniq { |r| r.question_uid }.count
-    private def total_answered = @total_answered ||= non_empty_responses.count
+    private def questions_to_score = non_empty_responses.first(EmaTrendCalculator::MAX_RECORDS_NEEDED)
+    private def replay_rate = (replays_answered / [total_answered.to_f, 1].max) * 100
     private def replays_answered = total_answered - unique_answered
-    private def latest_response = @latest_response ||= non_empty_responses.first
+    private def scores = @scores ||= questions_to_score.map { |response| QuestionResponseScorer.run(response:, question: questions_hash[response.question_uid].first) }
+    private def total_answered = @total_answered ||= non_empty_responses.count
+    private def unique_answered = @unique_answered ||= non_empty_responses.uniq { |r| r.question_uid }.count
```
anathomical, line 33: This is a re-organization of one-line method definitions, none of this code is actually changed.

services/QuillLMS/spec/factories/activities.rb
```diff
@@ -52,7 +52,6 @@
     classification          { create(:classification) }
     activity_categories     { create_pair(:activity_category) }
     raw_score               { create(:raw_score) }
-    question_count          { 10 }
```
anathomical, line 55: Removing this default value so that the underlying calculation in `QuestionCounter` will be used without being overwritten by default.

dandrabik (approved):
Thanks for fixing this up!

PR #13234, opened 2025-08-27
Classroom index speedup

dandrabik:
## WHAT
This is a performance refactor pass on `/teachers/classrooms/` index action. (I also had refactored it a bit because it hit the complexity lint check).
## WHY
According to New Relic, this is our "Apdex Most Dissatisfying" page. That's basically means this page is visited someone frequently and a large percentage of users have a slow page load.
## HOW
There are a few items that helped:
1. Only load visible classrooms on the the 'active' page and only non-visible classrooms on the 'archived' page
2. Improve eager loading to add a few relations and make sure it is used
3. Take a remove an ActivitySession.count query from within loop and do one large query per classroom and make a lookup hash.

Note, I tried to keep the functionality the same, except for only loading visible on 'active' and non-visible on 'archived'. There are probably some more refactors we could do here.

### Screenshots
Screenshot 2025-08-27 at 3 39 44 PM
Screenshot 2025-08-27 at 3 40 00 PM
Screenshot 2025-08-27 at 3 44 57 PM
Screenshot 2025-08-27 at 3 46 12 PM

I've compared the basic numbers/values for two users and a few classrooms on my local machine and on staging and they all seem to match.

PR Checklist | Your Answer
------------ | -------------
Have you added and/or updated tests? |  Not really, just a performance pass


services/QuillLMS/app/models/activity_session.rb
```diff
@@ -105,6 +105,7 @@ class StudentNotAssignedActivityError < StandardError; end
   # FIXME: do we need the below? if we omit it, may make things faster
   default_scope -> { joins(:activity) }
 
+  scope :finished, -> { where(state: STATE_FINISHED) }
   scope :completed,  -> { where.not(completed_at: nil) }
```
anathomical, line 112: I don't know that this is wrong, but `finished` and `completed` scopes should be identical.  If they aren't that's some sort of bug.  So it may make sense to settle on one to be the "canonical" scope to use rather than having some places in the code use one and other places use the other.

anathomical (approved):
Good catch with the eager loading.  And sorting student names in Ruby makes sense here.  I have one comment about the addition of the `finished` scope, but I think this is good to go out as-is if we want it to.

services/QuillLMS/app/controllers/teachers/classrooms_controller.rb
```diff
@@ -213,26 +230,31 @@ def transfer_ownership
...
-        student.merge(
-          number_of_completed_activities: ActivitySession.where(state: ActivitySession::STATE_FINISHED, user_id: student['id'], classroom_unit_id: classroom_unit_ids).count || 0,
-          provider: User.find(student['id']).provider
-        )
-      end
-    end
+    # do sorting in ruby so it uses the eager loaded students
+    students = classroom
+      .students
+      .sort_by { |s| s.name.split[1] || s.name }
```
emilia-friedberg: Just flagging that it's possible we'd get some unexpected behavior here with two-part last names -- probably not a big deal in practice.

emilia-friedberg (approved):
LGTM! Thanks for cleaning this up.

services/QuillLMS/app/models/activity_session.rb
dandrabik: Yeah, I agree. I was basically just trying to not change the functionality of the page, but agreed, we should be using only one. I'm going to add a comment and a TODO, but consider that change out of scope here.

services/QuillLMS/app/controllers/teachers/classrooms_controller.rb
dandrabik: Agreed. This is just a ruby rewrite of what was already there. I've added a comment to make that clear and added better handling of nils and blank spaces.

https://github.com/empirical-org/Empirical-Core/blob/develop/services/QuillLMS/app/models/user.rb#L284-L286

PR #13236, opened 2025-08-27
skills-report-performance-enhancement

anathomical:
## WHAT
- Move query aggregation into the sub-query where there are fewer values needed in the GROUP BY
- Move the `activity_session_completed_at` date range condition into the sub-query
- Actually use the `concept_results_for_students` sub-query instead of `concept_results` directly
## WHY
- The biggest performance hit we were facing was being caused by the `GROUP BY` statement with so many columns included.  The SQL query planner was first doing an `ORDER BY` on all of those columns, repeatedly running across the full set of results (~50k rows in our one of our test cases).  By doing the `GROUP BY` earlier in the process we eliminate extra rows (we really only need about 20% of those 50k rows in our final calculations) and simplify the GROUP BY by grouping on fewer columns
- Since the sub-query is the spot where we reference the `concept_results` table directly, it's the part of the query that actually benefits from being able to limit which partitions it looks at
- We were building the `concept_results_for_students` data, but then not using it.  Before the aggregation move, this was not causing a significant difference in performance (we still should have either used it or removed it), but moving the aggregation requires using the sub-query
## HOW
- Do aggregation on the `concept_results` we care about BEFORE joining all the ancillary tables (which should be one-to-one with the `concept_results` we care about, so shouldn't cause duplicate rows)
- Move the condition out of the main query into the sub-query
- JOIN to the sub-query table instead of the raw underlying table

### Notion Card Links

- Test old code and new code side-by-side locally, and then in staging
- Confirm that
- - The results returned for our test users are the same across both versions
- - The response times are below 5 seconds for the new code even in the most heavily loaded scenario (all concepts with all_history checked)

PR Checklist | Your Answer
------------ | -------------


dandrabik (approved):
LGTM, thanks for speeding this up!

PR #13237, opened 2025-08-28
skills report: performance breakdown: by skill

emilia-friedberg:
## WHAT
Build out the Skill Report: Performance Breakdown: By Skill page, with minor adjustments to the backend to get the `last_practiced` date for the individual student view and include students in the passed `classrooms` array.

## WHY
This is the second complete page for the new Skills Report!

## HOW
Pull out logic that's shared between this and the `by_student` page into the Shared file, then the aforementioned backend adjustments and React and CSS to get all functionality on this page working correctly.

### Screenshots
Full class, no skill category selected
Screenshot 2025-08-28 at 12 24 34 PM

Full class, skill category selected
Screenshot 2025-08-28 at 12 24 48 PM

Individual student, no skill category selected
Screenshot 2025-08-28 at 12 25 07 PM

Individual student, skill category selected
Screenshot 2025-08-28 at 12 25 18 PM

### Notion Card Links
https://www.notion.so/quill/Skills-Report-Performance-Breakdown-By-Skill-25cd42e6f941806f9d55e09374aeb689?source=copy_link

Tested all functionality on various screenwidths, Jack is QAing on staging.

PR Checklist | Your Answer
------------ | -------------


services/QuillLMS/client/app/bundles/Teacher/components/progress_reports/skills_report/performance_breakdown/by_skill.tsx
```diff
@@ -0,0 +1,379 @@
...
+  const tableRef = React.useRef<HTMLTableElement>(null)
+  const wrapperRef = React.useRef<HTMLDivElement>(null)
+  const containerRef = React.useRef<HTMLDivElement>(null)
+
+  React.useEffect(() => {
+    const queryHash = {
+      concept_id: selectedSkillCategory?.id,
+      classroom_id: classroom.id,
+      all_history: practiceScope === ALL_PRACTICE,
+      include_ell: includeELL,
```
emilia-friedberg, line 66: This isn't actually being used on the backend right now, but Thomas and I agreed to use this key so we won't have to make frontend changes when it is.

services/QuillLMS/app/controllers/teachers/progress_reports_controller.rb
```diff
@@ -55,7 +55,7 @@ def diagnostic_insights
   end
 
   def skills_report
-    @classrooms = classrooms
+    @classrooms = classrooms.map { |classroom| classroom.with_students }
```
happythenewsad: this is an n+1 query 

better to use classrooms.includes(students) and do an `attributes.merge` once loaded

services/QuillLMS/client/app/bundles/Teacher/components/progress_reports/skills_report/performance_breakdown/by_skill.tsx
```diff
@@ -0,0 +1,379 @@
...
+    wrapperRef,
+    tableRef,
+    enabled: true,
+    dependencies: [data],
+  });
+
+
+  const formatStudentCount = (numberOfStudents: number, totalClassroomStudents: number) => {
+    if (numberOfStudents === 0 || numberOfStudents == null) return '—';
+    const percentage = Math.round((100.0 * numberOfStudents) / totalClassroomStudents);
```
happythenewsad, line 116: should guard against zero division here

services/QuillLMS/client/app/bundles/Teacher/components/progress_reports/skills_report/performance_breakdown/by_skill.tsx
```diff
@@ -0,0 +1,379 @@
...
+      width: scaledWidths[0],
+      ariaLabel: selectedSkillCategory ? 'Skill Sub-Category' : 'Skill Category',
+      name: selectedSkillCategory ? 'Skill Sub-Category' : 'Skill Category',
+      attribute: 'name',
+      sortAttribute: 'nameForSort',
+      isSortable: true,
+      noTooltip: true
+    },
+    {
+      width: scaledWidths[1],
```
happythenewsad, line 142: moving to a semantic api like `scaledWidths['med-small']` or `scaledWidths['140px']` would avoid magical integers and improve maintainability here

services/QuillLMS/app/controllers/teachers/progress_reports_controller.rb
anathomical: You might actually be able to get Rails to make this performant by just doing `classrooms.includes(:students).map { |classroom| classroom.with_students }` as that SHOULD make Rails pull all the data in a single query and then during the map process fetch everything from memory.  If that doesn't work, `eager_load` in place of `includes` might do the trick.

services/QuillLMS/app/services/skills_report/skill_responses_processor.rb
```diff
@@ -40,6 +40,15 @@ def no_data_result
     private def avg_time_spent = avg_of(:timespent)
     private def number_of_students = student_skill_results.length
 
+    private def last_practiced
+      return nil unless student_skill_results.present?
+
+      last_practiced_dates = student_skill_results.map(&:last_practiced).compact
+      return nil if last_practiced_dates.empty?
```
anathomical: I don't think you actually need this guard clause because in Ruby `[].max` is `nil` already.  So I think you can just return `last_practiced_dates&.max`.

services/QuillLMS/client/app/bundles/Teacher/components/progress_reports/skills_report/performance_breakdown/by_skill.tsx
anathomical: This is tricky because `scaledWidths` is a dynamic calculated value based on static `columnWidths`, which are also not named.  But maybe they should be, and then the same names could be used in `scaledWidths`?

anathomical (approved):

services/QuillLMS/client/app/bundles/Teacher/components/progress_reports/skills_report/performance_breakdown/by_skill.tsx
emilia-friedberg: Yeah, that's fair. I think we can pass a name per column into `columnWidths`, update the hook to return `scaledWidths` with their name, and then use that here instead. Since we use that hook in several places, though, I think it would be better to break out that logic update into its own card and update everywhere we're using it. I'll make a card to do that.

services/QuillLMS/client/app/bundles/Teacher/components/progress_reports/skills_report/performance_breakdown/by_skill.tsx
emilia-friedberg: Added a guard clause above.

services/QuillLMS/client/app/bundles/Teacher/components/progress_reports/skills_report/performance_breakdown/by_skill.tsx
anathomical: Yeah, I agree that making this a separate task makes sense given how many places are now using scaledWidths.

happythenewsad (approved):

eadams17 (approved):
this looks great! thanks for all of your hard work on this

PR #13239, opened 2025-08-28
add-skills-report-concept-flags

anathomical:
## WHAT
- Add new columns to `Concept` model to be used to indicate how Concepts should be displayed in the new Skills Report
- Update the Staff Concept management UX to allow staff to set or unset these values on L2 and L1 concepts

NOTE: Code to have the backend services actually respect these flags is pending an initial population of which ones are supposed to be "on" to test with
## WHY
- Some skills (such as "spelling") are not deemed appropriate or useful for the new Skills Report, so we want a way to be able to flag which Concepts we want to surface in this report
- Additionally, some skills are considered ELL-focused, and thus not relevant to non-ELL teachers, so we want a way to be able to simplify the reports they see by suppressing lower-relevant Concepts unless they explicitly want to view them
- Decisions about which Concepts belong in the report, and in the ELL sub-category, belong to Curriculum, so we want to ensure that they can modify these values at-will
## HOW
- Run a migration to add two new columns to the `concepts` table
- Update our Graphql definitions for the `Concept` type to include the new columns (since we use Graphql in the Staff management workflows for Concepts)
- Add new `ChangeLog` types so that we can record when changes are made to these values on a `Concept`
- Update the Staff front-end for `Concept` management so that these values can be viewed and modified for L2 and L1 concepts

NOTE: We don't do this for L0 concepts because those can't be directly filtered for on the new Skills Report

### Screenshots
image

### Notion Card Links
https://www.notion.so/quill/Skills-Report-2-0-Add-new-Skills-Report-Settings-to-the-Concepts-Manager-for-L2-and-L1-Concepts-22cd42e6f94180c88a88d44ef49956ff

- Manually set "show in skills report" flag on a concept in the database
- confirm that the appropriate checkbox is checked for the concept in question
- toggle checkbox off and save
- enter ChangeLog message
- confirm that value is turned off in database and that new ChangeLog is created
- toggle both "show in report" and "ell" flag
- Confirm that two different ChangeLog explanations are required
- Confirm that both values are turned on in database and that two new ChangeLogs are created

PR Checklist | Your Answer
------------ | -------------


happythenewsad (approved):

emilia-friedberg (approved):
LGTM! Nice work here!

PR #13245, opened 2025-09-02
show-only-flagged-concepts

anathomical:
## WHAT
Update Skills Report to only include concepts flagged for it

Note that this was accompanied by a manually setting all Level 2 and Level 1 concepts to have this flag set to `true` for now.  Curriculum will go through and disable the ones they want at their leisure.
## WHY
We want the report to respect this flag when Curriculum toggles a Concept on or off for Skills Reports
## HOW
- Add a scope definition to check for the flag
- Add the scope to the controller that populates the drop-downs on the front-end

### Notion Card Links
https://www.notion.so/quill/Skills-Report-2-0-Add-new-Skills-Report-Settings-to-the-Concepts-Manager-for-L2-and-L1-Concepts-22cd42e6f94180c88a88d44ef49956ff

- On staging toggle a skill to `show_in_skills_report: false` and confirm that it no longer appears in drop-downs
- Toggle the flag back to `true` and confirm that the skill appears again

PR Checklist | Your Answer
------------ | -------------


brendanshean (approved):

services/QuillLMS/app/models/concept.rb
```diff
@@ -31,6 +31,7 @@ class Concept < ApplicationRecord
   scope :visible, -> { where(visible: true) }
   scope :level2, -> { where(parent_id: nil).visible }
   scope :level1, -> { where(parent: Concept.level2.all).visible }
+  scope :in_skills_report, -> { where(show_in_skills_report: true).visible }
```
emilia-friedberg: I'm just noticing that we are automatically applying the visible scope to these three, and wondering if it might be clearer to stack the scopes when necessary (ie `.level2.in_skills_report.visible`) so each scope is only doing exactly what's in the name.

emilia-friedberg (approved):
One thought, but you can take or leave it. Overall looks good!

happythenewsad (approved):

PR #13253, opened 2025-09-04
skills-report-include-proofreader-counts

anathomical:
## WHAT
Handle cases for Proofreader questions which don't have question_uids
## WHY
`concept_results` for Proofreader questions do not have `extra_metadata.question_uid` values because there are no underlying `Question` records for Proofreader data.  Instead the questions are just hard-coded directly into the passage for the `Activity`
## HOW
- In cases where there is a `concept_result`, but that there is no `question_uid` value, assume that the question is a Proofreader question.  In these cases:
- - Build a placeholder `question_uid` so that these can be included in our data aggregation
- - Add a fallback to the 'passage' `ActivityClassification` key when there is no actual `Question` entry
- Use a safe accessor when doing score calculations because we don't need the Question object to score Proofreader activities

### Notion Card Links

- Confirm that known proofreader skill completions show up in the report rollups for students

PR Checklist | Your Answer
------------ | -------------


services/QuillLMS/spec/services/skills_report/question_responses_for_classroom_by_skill_spec.rb
```diff
@@ -337,6 +337,27 @@ module SkillsReport
             expect(skill_names).to eq(skill_names.sort)
           end
         end
+
+        context 'when question_uid is not set (which happens with proofreader activities)' do
```
emilia-friedberg: Tiny suggestion: should we say something like `(which is expected for proofreader activities)` here? I think that might be helpful documentation if we're looking at this in the future and trying to figure out if this logic is accounting for a bug or anticipated behavior.

emilia-friedberg (approved):

services/QuillLMS/app/services/skills_report/question_responses_for_classroom.rb
```diff
@@ -21,10 +21,12 @@ class QuestionResponsesForClassroom < ApplicationService
             sc.student_id AS user_id,
             cr.activity_session_id,
             cr.question_number,
-            cr.extra_metadata->>'question_uid' as question_uid,
+            COALESCE(cr.extra_metadata->>'question_uid',#{' '}
```
dandrabik, line 24: Out of scope for this PR, but does this skills report include Evidence Activities? Do we need to do something for those (they don't have question_uids).

dandrabik (approved):
Small out-of-scope question about Evidence scores, but LGTM. Thanks for fixing these issues.

services/QuillLMS/app/services/skills_report/question_responses_for_classroom.rb
anathomical: Good call-out.  This code update will include Evidence questions, but only if their associated concepts are flagged for the report, which I think they are not intended to be currently.  It will also treat those questions as "proofreader" questions, but since `question_type` is only used to determine if we need to do special Fill in the Blank scoring, it shouldn't matter if we ever do decide to include those Concepts in the report.

PR #13255, opened 2025-09-04
Classcode Generator - Add new names and restructure code

dandrabik:
## WHAT
Increase the size of the name list for class codes, and update the logic to avoid a loop and prioritize shorter codes.
## WHY
We were running out of names and using the fallback with numbers. This buys us years of time and makes the code slightly more performant for the average and worst case scenaries.
## HOW
I used AI to generate nouns in multiple different ways, then used AI and manual checking to clean up the file. It still has some verbs and non-nouns, but I did a bunch of testing, and I'm not finding offensive codes. I did several passes to remove potentially offensive words and I think I got the lions share.
Note, the biggest gain was adding plurals, since there is a limit to nouns at elementary school level.

Also, I changed the algorithm to make 30 (or some number) codes, filter which ones are still available, then pick the shortest one. This makes the best case scenario slightly slower something like (30ms to 40ms), but it has two upsides
1. Decreases the worst case scenario from 900ms+ to 40ms
2. Allows us to prioritize shorter codes

### Notion Card Links
https://www.notion.so/quill/Blocked-NameGenerator-Increase-size-17dd42e6f9418001a9a7eec8f2e727ad

PR Checklist | Your Answer
------------ | -------------
Have you deployed to Staging? | No, pretty isolated service


anathomical (approved):
I think that the approach of pre-generating all of our name candidates up-front and then removing duplicates makes a lot of sense.  As you point out, it lets us avoid worst-case scenarios where we have to check individual candidates dozens of times.

emilia-friedberg (approved):
The logic updates look good to me. I didn't read the word lists closely in their entirety, but I think we've removed body part words before just because those are more liable to turn into something that, if not offensive, might be unfortunate. It might be worth doing another pass to remove those, but since teachers can regenerate their class code if they don't like it, it might not really matter.

happythenewsad (approved):

dandrabik:
> The logic updates look good to me. I didn't read the word lists closely in their entirety, but I think we've removed body part words before just because those are more liable to turn into something that, if not offensive, might be unfortunate. It might be worth doing another pass to remove those, but since teachers can regenerate their class code if they don't like it, it might not really matter.

Good advice. I did another pass generating batches of 30 and looking for weird combos, then edited the list. Then I did another pass on the shorter words. In total, I removed another 217 words. I generated a few hundred codes without hitting a weird one, so feeling like this is in a good place (I'm sure I missed some, but (hopefully) it should be a low percentage).

PR #13263, opened 2025-09-05
skills-report-ell-flag

anathomical:
## WHAT
- Update controller to extract `include_ell` flag and pass it through to services
- Update services to pass `ell` param through to the underlying queries
- Update underlying queries to exclude Concepts flagged with `ell = true` unless the `ell` param passed through us `true`
## WHY
We want to exclude ELL-only Concepts from the Skills Report unless a teacher explicitly requests them
## HOW
- Extract param that was already being set by the front-end, and pass it through to services
- Pass through `ell` param from outer services to data-fetching services
- Update the code that figures out what `concepts` to include in the report to exclude ones flagged `ell = true` unless the query service gets `ell: true` in its invocation

### Notion Card Links

- Load by Skill view of the Skills Report
- Select "all skills"
- Confirm all skills show up
- Toggle one of the skills to `ell: true` in the database
- Refresh and confirm that the toggled skill does not appear in the report
- Check the "include ELL skills" checkbox, wait for report to reload
- Confirm that the toggled skill is included

PR Checklist | Your Answer
------------ | -------------


emilia-friedberg (approved):
This looks good to me -- I assume the version that checks that the concept has `not_ell` or whatever we're going to call that flag is going to come after a migration PR?

anathomical:
> This looks good to me -- I assume the version that checks that the concept has `not_ell` or whatever we're going to call that flag is going to come after a migration PR?

Yes, that's correct.  I'd basically finished this code before we discussed the change, so I figured I'd get this out the door as-is.  I expect all the changes here to accommodate the updates will be back-end.  We'll still just pass a single boolean `include_ell` (or we might rename it?) to the back-end, and let it cobble things together from there.

PR #13264, opened 2025-09-08
skills-report-updated-ell-flags

anathomical:
## WHAT
Replace the old logic for flagging Concepts for the Skills Report:
- From a two-tiered set of flags, one for whether a Concept can appear in the report at all, and one for whether a Concept only shows up in the report if an ELL box is checked
- To a multi-flag system in which there is one flag for whether a Concept appears in the report if the ELL box is not checked, and for whether it appears when the box is checked

While doing that, I've also updated the styling for the CMS Concepts manager to bring it in line with the spec.
## WHY
Curriculum has indicated that the behavior they desire is not for ELL to be a strict sub-set of Concepts shown in the report, but that they want two overlapping sets for non-ELL and ELL views (some skills appear in both views, but each view excludes at least some skills that the other view shows)
## HOW
- Migrate the current `show_in_skills_report` column to be `non_ell`
- Tweak CMS interface for managing these flags to display the new labels as requested by Curriculum
- Update `Concept` model scopes to more conveniently access this data
- Update query services to restrict which Concepts show up in reports based on the `ell` flag passed in

### Screenshots
image

### Notion Card Links
https://www.notion.so/quill/Skills-Report-2-0-Add-new-Skills-Report-Settings-to-the-Concepts-Manager-for-L2-and-L1-Concepts-22cd42e6f94180c88a88d44ef49956ff?source=copy_link#265d42e6f94180ae9bb6efc79770fe6e

- Flag skill as `non_ell` only
- Confirm that it shows up in the "by Skills" report by default
- Confirm that it is no longer listed when "show ELL skills" is checked
- Flag skill as `ell` only
- Confirm that it shows up when "show ELL skills" is checked, but not when it is unchecked
- Flag skill as `'non_ell` and `ell` at the same time
- Confirm that it shows up whether or not "show ELL skills" is checked or unchecked

PR Checklist | Your Answer
------------ | -------------


services/QuillLMS/app/controllers/teachers/progress_reports_controller.rb
```diff
@@ -57,8 +57,12 @@ def diagnostic_insights
   def skills_report
     @classrooms = classrooms.includes(:students).map { |classroom| classroom.with_students }
 
-    @all_skill_categories = Concept.level2.in_skills_report.visible.order(:name)
-    @all_skill_sub_categories = Concept.level1.in_skills_report.visible.order(:name)
+    # NOTE: the inclusion of an `or` in this call chain makes order important.
+    # The `OR` will be parenthetically combined with everything earlier in the chain,
+    # so this construction gives us "WHERE (non_ell = true OR ell = true) AND <other scopes>"
+    concepts_in_report = Concept.non_ell.or(Concept.ell)
```
dandrabik: You might consider making this a scope on it's own since it is complicated, e.g.
```ruby
scope :skills_report -> { non_ell.or(Concept.ell) }

File: services/QuillLMS/app/services/skills_report/classroom_scores_by_skill.rb [hunk 2]
```diff
@@ -28,27 +26,20 @@ def response_hash
...
     private def placeholder_for_unplayed_concepts
-      return unplayed_level1_placeholders if concept_id.present?
-
-      unplayed_level2_placeholders
-    end
-
-    private def unplayed_level2_placeholders
-      ALL_L2_CONCEPTS.to_h do |concept|
+      # This will naturally fetch L2s if concept_id is nil, and appropriate L1s if it isn't
+      concepts.where(parent_id: concept_id).to_h do |concept|
```
dandrabik: take or leave, consider chaining these on new lines for readability and using pluck,e.g.
```ruby
concepts
  .where(parent_id: concept_id)
  .pluck(:name)
  .to_h { |name| [name, []] }

File: services/QuillLMS/app/services/skills_report/classroom_scores_by_skill.rb [hunk 3]
```diff
@@ -28,27 +26,20 @@ def response_hash
...
         [concept.name, []]
       end
     end
 
-    private def unplayed_level1_placeholders
-      Concept.in_skills_report.visible.where(parent_id: concept_id).select(:id, :name).to_h do |concept|
-        [concept.name, []]
-      end
+    private def concepts
```
dandrabik: consider two methods for readability, e..g
```ruby
private def concepts = base_concepts.visible.select(:id, :name)
private def base_concepts = ell ? Concept.ell : Concept.non_ell

File: services/QuillLMS/app/models/concept.rb [hunk 4]
```diff
@@ -31,7 +31,8 @@ class Concept < ApplicationRecord
   scope :visible, -> { where(visible: true) }
   scope :level2, -> { where(parent_id: nil) }
   scope :level1, -> { where(parent: Concept.level2.all) }
-  scope :in_skills_report, -> { where(show_in_skills_report: true) }
+  scope :ell, -> { where(ell: true) }
+  scope :non_ell, -> { where(non_ell: true) }
```
happythenewsad: If `non_ell` describes something other than `!ell`, it should be named differently to avoid confusion

services/QuillLMS/app/services/skills_report/question_responses_for_classroom_by_skill.rb
```diff
@@ -85,7 +83,14 @@ def initialize(classroom_id:, concept_id:, user_id: nil, all_history: false, ell
     def run = raw_results.map { |h| Result.new(**h) }
 
     private def raw_results = RawSqlRunner.execute(sql)
-    private def concept_ids = concept_id.present? ? [concept_id] : L2_CONCEPT_ID_WHITELIST
+    private def concept_ids = @concept_ids ||= fetch_concept_ids
+
+    private def fetch_concept_ids
```
dandrabik, line 89: consider the two methods here also, e.g. isolated the branching in a simple method so the more complicated method gets simpler.

```ruby
private def fetch_concept_ids
  return [concept_id] if concept_id.present?
    
  base_concepts.level2.pluck(:id)
end

private def base_concepts = ell ? Concept.ell : Concept.non_ell

services/QuillLMS/app/services/skills_report/classroom_scores_by_skill.rb
dandrabik: You also might consider getting rid of the `concepts` method for `base_concepts.visible` here

dandrabik (approved):
LGTM, thanks for setting this all up. Some very small refactor suggestions that you can take or leave.

emilia-friedberg (approved):
LGTM!

happythenewsad (approved):
👍

PR #13266, opened 2025-09-09
Upgrades/ replace react-html-parser

eadams17:
## WHAT
replace `react-html-parser` package with `html-react-parser`

## WHY
`html-react-parser` is no longer being maintained

## HOW
refactor code to handle new configuration

note: the new package triggers a heavier re-render/re-parsing load and was causing issues with text editor focus, so I had to refactor some of the StudentViewContainer state handling logic to be more conservative with re-renders 

### Screenshots

### Notion Card Links
https://www.notion.so/quill/Convert-react-html-parse-to-html-react-parser-127d42e6f94180a9a96ee85abda6349c

- tested on staging that Evidence activity plays as expected and time tracking logic performs the same
- tested on staging that ELL activity translations work as expected
- tested on staging that Evidence Seed Data generation works as expected

PR Checklist | Your Answer
------------ | -------------


services/QuillLMS/client/app/bundles/Evidence/components/studentView/container.tsx
```diff
@@ -631,18 +732,18 @@ export const StudentViewContainer = ({ dispatch, session, isTurk, location, acti
     return ''
   }
 
-  function transformMarkTags(node) {
+  const transformMarkTags = React.useCallback((node: any) => {
```
emilia-friedberg, line 735: Since this is already a really big file that's only getting bigger with these changes, and it looks like this function is only used inside of ReadPassageContainer (and I think is also the only place we were previously using ReactHTMLParser), I'm wondering if it might work to pull this whole function down to a lower-level component, and maybe see if we can get away without as many updates here?

emilia-friedberg (commented):
Most of this looks good to me, but I'm concerned about the large changeset in that container file -- having to monitor re-renders so closely feels very brittle to me. I made a suggestion there to see if we can move the function in question a level lower, to avoid re-rendering that top-level component so much. If that doesn't help, I think it may be worth another pass through the LLM to ask it to only change directly relevant code -- I think at least some of that diff is cleanup that isn't actually changing functionality here, and it'd be easier to review/diagnose any unexpected behavior down the line if the diff is smaller.

services/QuillLMS/client/app/bundles/Evidence/components/studentView/container.tsx
eadams17: This text editor losing focus is tricky, I think it's related to it being contenteditable... I extracted the `transformMarkTags` logic in the `ReadPassageContainer`, but unfortunately I'm not sure there's a way around not introducing a big set of changes. You can see what those changes look like in [this draft PR](https://github.com/empirical-org/Empirical-Core/pull/13267). Let me know which path you think is preferable! There's some cleanup I'd do in that draft PR if you think that route is better

PR #13274, opened 2025-09-10
Fix issue with classlink classrooms with no owner

brendanshean:
## WHAT
Fix a bug involving classlink classrooms being imported with no owner.

## WHY
There was previous code in enrollment_persister:
```
def create_teacher_linkages
    user.classrooms_teachers.find_or_create_by!(classroom:) do |ct|
      ct.role = classroom.classrooms_teachers.exists?(role: OWNER) ? COTEACHER : OWNER
    end
  end
end
```

The issue here is that the block is only called on `create!`
```
# File activerecord/lib/active_record/relation.rb, line 238
def find_or_create_by!(attributes, &block)
  find_by(attributes) || create_or_find_by!(attributes, &block)
end
```

that was refactored in the last [day](https://github.com/empirical-org/Empirical-Core/commit/f76b084da224bab2b0aebfd14f1de49ca772a43f), but previous classrooms were created with this bug.

## HOW
Identify classlink classrooms with only coteachers and promote the first coteacher as the owner.

### Screenshots

### Notion Card Links
https://www.notion.so/quill/When-two-District-Premium-teachers-click-on-a-class-they-co-teach-on-the-Manage-Classes-page-in-thei-26ad42e6f941802fa04be7ea7e596a55?source=copy_link

I've done a dry run and located 
PR Checklist | Your Answer
------------ | -------------
Have you added and/or updated tests? |  No. Rake task
Have you deployed to Staging? | No.  Ran dry-run on follower database.


anathomical (approved):
This approach looks like it will accomplish the stated goal, so looks good to me.

PR #13284, opened 2025-09-12
skills-report-practice-cleanup

anathomical:
## WHAT
- Update CSV with L1 Practice data
- Add caching to the Practice Recommendations controller action
- Tweak calculation of `ConceptPracticeRecommendation` readability to be more efficient
## WHY
- Curriculum renamed some L1 Concepts in the database and the names need to match
- This data is very stable, so no need to hit the database all the time
- The way we were doing this was hitting a call in `unit_template` that re-fetched data from the database, meaning that we couldn't take advantage of pre-fetching data with `includes`
## HOW
- Update the associated Google spreadsheet and populate the CSV from it
- Wrap the data generation call in `Rails.cache.fetch`
- Write some logic that has some overlap with `UnitTemplate.readability` that avoids re-fetching data from the database

### Notion Card Links
https://www.notion.so/quill/Skills-Report-2-0-Assign-Targeted-Practice-Workflow-24dd42e6f94180ecb134feeb9169eef9#250d42e6f941803081e7cdd833fea5bf

- Confirm that the CSV successfully feeds into the Production thor script
- Run the code to generate our Recommendations for the controller and confirm that
- - It runs much faster than before
- - It caches to run even faster on subsequent requests

PR Checklist | Your Answer
------------ | -------------
Have you deployed to Staging? | No - should be tested in conjunction with other code


emilia-friedberg (approved):
Looks great. Thanks for making those changes so quickly.

PR #13286, opened 2025-09-12
Feat/ allow admins bulk archive classes

eadams17:
## WHAT
allow admins to bulk archive classes

## WHY
we want admins to be able to bulk archive various teachers' classes

## HOW
add new action to Account Management page and API to bulk archive various teachers' classes

### Screenshots
Screenshot 2025-09-12 at 2 31 28 PM

### Notion Card Links
https://www.notion.so/quill/Allow-admins-to-bulk-archive-classes-262d42e6f94180abb255f0ed29acd44d

tested on staging that multiple teachers' classes can be archived by admins filtered by date

PR Checklist | Your Answer
------------ | -------------


emilia-friedberg (approved):
It looks like we have some failing specs and lints, but in general, the code looks good to me.

On a more general note, though, I wonder if it makes sense to have Jack take a look at this. You implemented the design as spec-ed, but it feels inconsistent with other parts of the site. For instance, the bolded text above the calendar isn't really a pattern we've used elsewhere, and for the calendar itself, since we have other modals with a calendar view, I wonder if it would be better to reuse an element that looks more like that (from react-dates) rather than from MUI. Obviously these are more product management decisions and not engineering ones, but since Nattalie is newer to PMing, it might be worth talking to her about syncing with design before implementation for this kind of thing -- Jack could just offer input instead of necessarily having to do the design itself. That said, maybe there was a conversation about not adding more to his plate that I'm just unaware of.

services/QuillLMS/app/controllers/admins_controller.rb
```diff
@@ -331,18 +339,34 @@ def create_and_link_accounts
...
+    classroom_ids = teacher&.classrooms_i_own&.pluck(:id) || []
+
+    Classroom.where(id: classroom_ids).each do |classroom|
+      next if classroom.owner.id != teacher.id
+      next if classroom.created_at > filter_date
+
+      classroom.visible = false
+      # we want to skip validations here because otherwise they can prevent archiving the classroom, when the classroom was created before the validation was added
+      classroom.save(validate: false)
+    end
```
anathomical: I think you could make this slightly more efficient by doing your conditions on the actual query:
```ruby
Classroom.joins(:classrooms_teachers)
  .where(id: classroom_ids)
  .where(classrooms_teachers: { user_id: teacher.id, role: 'owner' })
  .where('created_at > ?', filter_date)
  .update_all(visible: false)
```
Note that `update_all` should bypass validation.

anathomical (approved):
A small note on a different approach to archiving the classrooms that reads a bit cleaner to me, but it's ultimately a style issue for a function that won't get called that often, so take it or leave it.

services/QuillLMS/app/controllers/admins_controller.rb
eadams17: ok yeah, this looks cleaner-- I'll update, thank you!

eadams17:
> It looks like we have some failing specs and lints, but in general, the code looks good to me.
> 
> On a more general note, though, I wonder if it makes sense to have Jack take a look at this. You implemented the design as spec-ed, but it feels inconsistent with other parts of the site. For instance, the bolded text above the calendar isn't really a pattern we've used elsewhere, and for the calendar itself, since we have other modals with a calendar view, I wonder if it would be better to reuse an element that looks more like that (from react-dates) rather than from MUI. Obviously these are more product management decisions and not engineering ones, but since Nattalie is newer to PMing, it might be worth talking to her about syncing with design before implementation for this kind of thing -- Jack could just offer input instead of necessarily having to do the design itself. That said, maybe there was a conversation about not adding more to his plate that I'm just unaware of.

Thanks for calling this out! To close the loop on this, Jack was fine with using this Calendar component for the modal. He just suggested some tweaks to the text styling and margins

PR #13290, opened 2025-09-16
skills report locked views

emilia-friedberg:
## WHAT
Add the locked view to the main page of the Skills Report if the user does not have premium, and a banner to the header once they're in it if they're not verified (they can also only see the Class view in this state).

## WHY
We want to restrict access to this report.

## HOW
Passing up requisite data from backend and using it here, lots of React and CSS. Refactored shared code from Diagnostic Insights to use in both instances.

### Screenshots

### Notion Card Links
https://trello.com/c/GJei1wzO/82-implement-locked-page-and-approval-process

Tested locally, Jack and Peter QAing on staging.

PR Checklist | Your Answer
------------ | -------------


services/QuillLMS/app/controllers/teachers/progress_reports_controller.rb
```diff
@@ -61,6 +61,14 @@ def skills_report
     @all_skill_categories = Concept.in_skills_report.level2.visible.order(:name)
     @all_skill_sub_categories = Concept.in_skills_report.level1.visible.order(:name)
 
+    @needs_email_verification = needs_email_verification_for_diagnostic_insights
+    @school = current_user.school
+    @role = current_user.role
+    @admin_verification_status = current_user.admin_approval_status
+    @teacher_verification_status = current_user.teacher_verification_status
+    @school_admins = @school&.admins
+    @premium_state = current_user.premium_state
```
anathomical: Probably not worth doing now, but I wonder if this bundle of attributes is going to be set on more controller actions moving forward and maybe we should start thinking about a way to make this a single command.  I think we only do this in two places right now: here and Diagnostic Insights, but I could imagine this being something we want in more places.

services/QuillLMS/app/views/teachers/progress_reports/skills_report.html.erb
```diff
@@ -1,5 +1,13 @@
...
-  allSkillSubCategories: @all_skill_sub_categories
+  allSkillSubCategories: @all_skill_sub_categories,
+  needsEmailVerification: @needs_email_verification,
+  school: @school,
+  role: @role,
+  adminVerificationStatus: @admin_verification_status,
+  teacherVerificationStatus: @teacher_verification_status,
+  schoolAdmins: @school_admins,
+  alternativeSchoolNames: School::ALTERNATIVE_SCHOOL_NAMES,
+  premiumState: @premium_state
```
anathomical, line 12: Same comment as the one above: I suspect we'll be doing this more in the future, so while it's not worth doing anything right now, but it's worth considering how we'd generalize passing a bunch of this stuff forward.  At least the stuff that really is going to be broadly used.

anathomical (approved):
Nice!  I like the way you've extracted out the verification and no-classrooms states and only load `routes` if we're in "normal" state.  It makes it really clear what "normal" looks like, and what special cases there are.

services/QuillLMS/client/app/bundles/Teacher/components/progress_reports/verification/verification_banner.tsx
```diff
@@ -0,0 +1,194 @@
...
+      href: '/teachers/premium_hub',
+      text: 'Go to Premium Hub'
+    }
+
+    switch(adminVerificationStatus) {
+      case PENDING:
+        heading = 'We\'re reviewing your request'
+        copy = 'Your verification request is pending approval. Once approved, you will be able to use this feature. If you need help in the meantime, contact us.'
+        button = {
+          href: 'mailto:hello@quill.org',
```
happythenewsad: This should be a constant

happythenewsad (approved):

services/QuillLMS/client/app/bundles/Teacher/components/progress_reports/verification/verification_banner.tsx
```diff
@@ -0,0 +1,194 @@
...
+      }
+    } else if (teacherVerificationStatus === PENDING || submittedVerificationRequest) {
+      heading = 'Your request has been sent'
+      copy = 'Your verification request is pending review. Once verified, you will receive an email letting you know that you have access to this feature. If you need help in the meantime, contact an admin.'
+      button = {
+        href: '/admin_access',
+        text: 'Contact an admin'
+      }
+    } else if (teacherVerificationStatus === DENIED) {
+      heading = 'Your admin couldn\'t verify you'
```
eadams17: is this backslash intentional?

eadams17 (approved):
LGTM!

services/QuillLMS/app/controllers/teachers/progress_reports_controller.rb
emilia-friedberg: Yeah, I'll go ahead and add a shared setter for these for now.

services/QuillLMS/client/app/bundles/Teacher/components/progress_reports/verification/verification_banner.tsx
emilia-friedberg: It was just escaping the apostrophe, but I think it's better to just wrap the string in double quotes, so I updated it.

PR #13304, opened 2025-09-24
skills-report-pre-cache-api-responses

anathomical:
## WHAT
- Refactor `classroom_scores_by_` both `skill` and `user` to use new `Compose` services
- Add a new `for_cache` flag to the `question_responses_for_classroom` and `_by_skill` run parameter `for_cache`
- Add individual workers for pre-caching both "by Student" and "by Skill" report data
- Add a fan-out worker to enqueue individual workers for each classroom of a specific user
- Enqueue the fan-out worker for a teacher when they load their dashboard if they are a premium user
## WHY
- We want to separate final payload processing from localized data fetches, ensuring that `Compose` services never fetch from database, so you can call it and know there won't be db-related performance costs
- The new `for_cache` flag is a "maximalist" version of the data.  It allows for a single data fetch of all facets of the appropriate data, intending to slice it up in Ruby
- Have workers that fetch all possible report data, and then slice it into its various facets in memory to construct and cache payloads with instead of fetching each facet individually from the database
- Data must be fetched on a per-classroom basis, so having the cache workers take the classroom_id as an argument makes sense, but we want a straightforward way of caching report data for all of a given teacher's classrooms
- We want to actually pre-cache reporting data for most teachers so that when they reach the reports, performance is super-snappy
## HOW
- Create new services to `Compose` data, and insert it into the `ClassroomScore`s services
- Add the flag to `initialize`, tweak the query to remove many normal restrictions when the flag is passed, and add some additional values to the `Response` objects to allow for in-Ruby filtering in lieu of in-Postgres filtering
- Create workers that take `user_id` and `classroom_id`, fetch the data, slice it up, and use `classroom_cache` to cache this data in the same keys the controllers fetch from
- Add a worker that uses `classrooms_i_teach` to fan out for all classrooms of a given user
- Add a `before_action` to enqueue the fan-out worker for a teacher if they are a premium user

### Notion Card Links

- Manually clear cache
- Load teacher's dashboard
- Navigate to Skills Report
- Confirm that selection load times are < 1 second
- Confirm that new database calls are not being made for the report
- Use a script to confirm that the value being cached matches the value that is generated by the narrower database-based data generation

PR Checklist | Your Answer
------------ | -------------


services/QuillLMS/app/workers/skills_report/cache_users_api_data_worker.rb
```diff
@@ -0,0 +1,69 @@
...
+    private def all_rows = @all_rows ||= QuestionResponsesForClassroom.run(classroom_id:, concept_id: nil, for_cache: true)
+    private def concept_ids = concepts.map(&:id)
+    private def concepts = @concepts ||= Concept.in_skills_report.visible.select(:id, :name, :parent_id, :ell, :gen_ed)
+    private def question_uids = all_rows.map(&:question_uid).uniq
+    private def questions_hash = @questions_hash ||= Question.where(uid: question_uids).group_by(&:uid)
+    private def user = @user ||= User.find_by(id: user_id)
+    private def user_hash = @user_hash ||= User.where(id: user_ids).pluck(:id, :name).to_h
+    private def user_ids = all_rows.map(&:user_id).uniq
+
+    private def filter_classroom_history(rows, classroom_id) = rows.filter { |row| row.classroom_id == classroom_id }
```
happythenewsad, line 60: This method and several others are duplicates of those in `cache_skills_api_data_worker`

services/QuillLMS/app/services/skills_report/question_responses_for_classroom_by_skill.rb
```diff
@@ -51,12 +51,20 @@ class QuestionResponsesForClassroomBySkill < ApplicationService
...
             ROUND(act_s.timespent::numeric / act.question_count, 0)::int AS timespent,
-            act_s.activity_id
+            act_s.activity_id,
+            (c_grandparent.gen_ed AND c_parent.gen_ed) AS gen_ed,
+            (c_grandparent.ell AND c_parent.ell) AS ell,
+            cu.classroom_id
         FROM students_classrooms sc
             JOIN activity_sessions act_s ON sc.student_id = act_s.user_id
-                #{'AND act_s.classroom_unit_id IN (SELECT id FROM classroom_unit_ids)' unless all_history}
+                #{'AND act_s.classroom_unit_id IN (SELECT id FROM classroom_unit_ids)' unless all_history || for_cache}
```
happythenewsad: Optional style thing: I find SQL is more readable and maintainable when Ruby logic is maximally extracted. How about:
```ruby
def join_classroom_unit_ids 
  return if all_history || for_cache
  'AND act_s.classroom_unit_id IN (SELECT id FROM classroom_unit_ids)'
end
```

happythenewsad (approved):
There's an opportunity to DRY up some methods, but overall this is very clearly organized . Nice... composition

services/QuillLMS/app/controllers/teachers/classroom_manager_controller.rb
```diff
@@ -391,7 +391,9 @@ def set_diagnostic_variables
     @alternative_schools_name_map = School::ALTERNATIVE_SCHOOLS_DISPLAY_NAME_MAP
   end
 
-  private def cache_diagnostic_insights_recommendation_count
+  private def pre_cache_report_data
     DiagnosticInsights::CacheRecommendationCountsWorker.perform_async(current_user.id)
+    # Only pre-cache Skills Report if the user is premium since it's a premium-only report
```
dandrabik: Consider an early `return unless current_user.subscription` here. That might make the comment less needed.

services/QuillLMS/app/services/skills_report/classroom_scores_by_skill.rb
```diff
@@ -13,45 +13,21 @@ def initialize(classroom_id:, concept_id: nil, user_id: nil, all_history: false,
...
 
     def run
-      placeholder_for_unplayed_concepts
-        .merge(response_hash_with_ids)
-        .map do |skill_name, skill_data|
-          user_responses = skill_data[:responses]
-          id = skill_data[:id]
-          SkillResponsesProcessor.run(user_responses:, name: skill_name, id:, questions_hash:, total_classroom_students:)
-        end
+      return ComposeScoresBySkill.run(concepts:, question_responses:, questions_hash:, total_classroom_students:)
```
dandrabik: You can remove the `return`

services/QuillLMS/app/services/skills_report/compose_scores_by_user.rb
```diff
@@ -0,0 +1,21 @@
...
+
+    def initialize(question_responses:, user_hash:, questions_hash:)
+      @question_responses = question_responses
+      @questions_hash = questions_hash
+      @user_hash = user_hash
+    end
+
+    def run
+      response_hash
+        .to_a
```
dandrabik: You don't need this `to_a`. `hash.map { |user_id, response| ... }`

services/QuillLMS/app/services/skills_report/question_responses_for_classroom.rb
```diff
@@ -53,27 +62,39 @@ class QuestionResponsesForClassroom < ApplicationService
...
     end
 
     def run = raw_results.map { |h| Result.new(**h) }
 
     private def raw_results = RawSqlRunner.execute(sql)
 
-    private def sql = ActiveRecord::Base.sanitize_sql([query_sql, { classroom_id:, concept_id: }])
+    private def sql = ActiveRecord::Base.sanitize_sql([query_sql, { classroom_id:, concept_ids: }])
+
+    private def concept_ids
```
dandrabik, line 110: Take or leave, consider making this its own method for readability:

```ruby
private def all_skills_report_concept_ids = Concept.in_skills_report.visible.pluck(:id)

...
return all_skills_report_concept_ids if for_cache

[concept_id]

dandrabik (approved):
Overall this looks really good. Well structured and easy to read. Small take or leave notes, but looks great! :ship:

services/QuillLMS/app/workers/skills_report/cache_users_api_data_worker.rb
anathomical: I think that this is a good call-out, but will address DRY issues in a separate PR to keep this simple.

PR #13313, opened 2025-09-29
skills report: add segment event; update tooltip copy

emilia-friedberg:
## WHAT
Add a segment event, worker, and endpoint to track teacher assigning targeted practice event. Also change the copy in two tooltips.

## WHY
Peter Sharkey requested this so we can keep track of how teachers are using these reports.

## HOW
Update the frontend to pass the requisite data, then make a call to a new controller endpoint that calls a background wroker that calls a Segment track event.

Note: I'm not sure if we use this pattern elsewhere (a controller endpoint that only calls an analytics worker). Another option I considered was to pass optional params to the `assign_single_pack` endpoint, but that's used in multiple places that won't be calling this worker, so it seemed a little messy. Happy to revisit if people feel strongly.

### Screenshots

### Notion Card Links
https://www.notion.so/quill/Skills-Report-2-0-Add-Teacher-assigned-targeted-practice-Segment-event-279d42e6f9418012a637e66cd5e60d08

Tested locally that the Segment events get called with the correct data.

PR Checklist | Your Answer
------------ | -------------


services/QuillLMS/app/workers/track_targeted_practice_assignment_worker.rb
```diff
@@ -0,0 +1,11 @@
+# frozen_string_literal: true
+
+class TrackTargetedPracticeAssignmentWorker
+  include Sidekiq::Worker
+  sidekiq_options queue: SidekiqQueue::LOW, retry: false
+
+  def perform(teacher_id, activity_pack_name, report, section)
+    analytics = Analytics::SegmentAnalytics.new
```
brendanshean: I would add a guard clause at the top of this:  `return unless teacher_id` and then get rid of the conditional on line 9.

brendanshean (approved):
Nice work on this.

eadams17 (approved):

PR #13317, opened 2025-09-30
skills-report-fix-skills-cms

anathomical:
## WHAT
- Remove `id` from `Types::ConceptPracticeRecommendationType`
- Update `conceptId` in `Types::ConceptPracticeRecommendationInput` to be `Int` instead of `String`
- Remove `toString()` casts on the front-end
- Strip `__typename` key from `conceptPracticeRecommendations` before submitting edit updates
## WHY
- This is an unnecessary key for our use-case since we handle updates via a different logic based on other keys
- It comes out of the database as an Int, and we want to match the value between `Type` and `Input` versions of the Graphql definitions
- Since we've changed the Graphql definition to Int, we should be consistent about what we submit
- This value comes from the fetch automatically, but we don't need it in the edit, so we need to strip it for submitting the same values back when practice recommendations aren't changed
## HOW
- Update Type definitions for Graphql
- Remove unneeded casts
- Add logic to strip unnecessary key before saving records

- Confirm that updating just checkboxes (checking/unchecking) works when keeping practice the same

PR Checklist | Your Answer
------------ | -------------


emilia-friedberg (approved):
Good job figuring this out!

eadams17 (approved):
👏

PR #13326, opened 2025-10-02
skills-report-refactor

anathomical:
## WHAT
- Refactor 'app/services/skills_report` Student and Skill data fetch services to have a shared parent
- Refactor `app/workers/skills_report` Student and Skill pre-cache workers to have a shared parent
- Add an early return for Student data controller caching when called without a concept_id
- Refactor SkillsReportController to avoid using `to_i` on incoming params.
- Add an early return on the SkillsReportController `practice_assignments` in cases where there are no `classroom_ids` to pull data for
## WHY
- There is a lot of shared code between the two services, so we should consolidate that code to highlight what's actually different between them
- Same as above: consolidate the shared code to highlight what's actually different
- The front-end employs a pre-caching strategy where it loads Student data when the teacher looks at Skill data, and vice versa so that if the teacher switches from one mode to the other it is almost instantaneous, but the Skill data allows for concept_id of `NULL`, while that is meaningless for Student data, so there's no point in pre-caching the junk data
- This lets us avoid cases where un-passed params (`nil`) get converted to `0` (`nil.to_i == 0`) preventing us from using `.present?` to check whether they are set or not
- The data-fetching service will crash (and cause a 500) if `classroom_ids` is empty (note that this won't break teacher experience since this data can't be used on the front-end if there are no classrooms, but the errors still show up in Sentry)
## HOW
- Claude-guided refactor process to extract a common parent
- Claude-guided refactor process to extract a common parent
- Add an early return in the cache method if concept_id is not set
- Remove `to_i` calls from param loads in the controller, update the spec file to handle expectations of receiving strings instead of ints
- Add an early return to render an empty array (the value that comes back naturally if there are no practice assignments) if there are no available classrooms to fetch data for

- Run the snapshot generating logic to ensure that the refactored data services produce the same payloads before and after changes
- Trigger pre-caching workers, confirm that when the user loads the Skills Report, new database queries are not made and the data loads successfully from cache
- Confirm that when a teacher loads the Skills report for all skills, there is not a database call to cache junk data for the Student view with no concept_id
- Load various reports with various params to confirm that nothing is preventing params from being passed through

PR Checklist | Your Answer
------------ | -------------


services/QuillLMS/app/workers/skills_report/cache_api_data_worker_base.rb
```diff
@@ -0,0 +1,87 @@
...
+      filtered_rows = filter_rows_by_params(rows, concept_id:, all_history:, ell:)
+
+      key = user.model_cache_key(classroom,
+        key: cache_key_name,
+        groups: cache_key_groups(concept_id:, all_history:, ell:))
+
+      data[key] = compose_scores(filtered_rows, concept_id:, ell:)
+    end
+
+    private def cache_key_name
```
dandrabik: Small suggestion. Move all of the subclass required methods towards the top of the file so the protocol is clear (also, I don't think you need the messages):
```ruby 
def perform
...
# Subclasses must implement these 5 methods
private def generate_cache_data = (raise NotImplementedError)
private def cache_key_name = (raise NotImplementedError)
private def compose_scores(filtered_rows, concept_id:, ell:) = (raise NotImplementedError)
private def all_rows = (raise NotImplementedError)
private def concepts= (raise NotImplementedError)

Comment (dandrabik) [reply services/QuillLMS/app/workers/skills_report/cache_api_data_worker_base.rb]: This is a take or leave (you can ship as is), but if subclasses have to overwrite 5 methods and some of the subclasses overwrite several more, I wonder if something like mixin(s)/concern(s) would be a better fit for the truly shared pieces of functionality, even if it leaves some duplication.

File: services/QuillLMS/app/services/skills_report/question_responses_base.rb [hunk 2]
```diff
@@ -0,0 +1,104 @@
...
+                WHERE#{' '}
+                    /* The earliest concept_results with `question_uid` is from July 2022, so this condition limits the number of partitions we look at */
+                    cr.activity_session_completed_at BETWEEN '2022-07-01' AND NOW() + INTERVAL '1 hour'
+                    #{relevant_concept_results_where_clause}
+                GROUP BY cr.activity_session_id, cr.question_number
+        )
+      SQL
+    end
+
+    private def select_sql
```
dandrabik: I mention this below in another base class, but I'd put these all below the `run` method to make the protocol clear.

dandrabik (approved):
LGTM! Thanks for putting this together

happythenewsad (approved):

PR #13333, opened 2025-10-03
Add cron jobs to check llm models

brendanshean:
## WHAT
Add cron jobs that check for models that
* are not returning responses
* are deprecated within the next week
* are not returning embeddings 

## WHY
* the model might be down
* we should transition the model soon
* we should look at new embeddings model

## HOW
* add cron job and worker to check
* add cron job and worker to check
* add cron job and worker to check

### Screenshots

### Notion Card Links

I checked that these hooks work on staging.

PR Checklist | Your Answer
------------ | -------------


anathomical (approved):
Looks good.  I think it makes a lot of sense to have automated tasks to check for this kind of data health.  I wonder if there are other slices of our code where this sort of thing would be valuable.

services/QuillLMS/engines/evidence/app/models/evidence/research/gen_ai/llm.rb
```diff
@@ -78,6 +79,8 @@ def self.g_eval = find_by(vendor: GOOGLE, version: Evidence::Gemini::Chat::SMALL
         def self.form_options = production.where(reasoning: false) + non_production.active.where(reasoning: false).ordered
         def self.all_ordered = production.where(reasoning: false) + non_production.where(reasoning: false).ordered
 
+        def self.in_use = joins(trials: { llm_prompt: :activated_llm_prompts }).distinct
```
dandrabik: I think this needs a scope for `active` activated_llm_prompts

services/QuillLMS/app/workers/check_deprecated_llms_worker.rb
```diff
@@ -0,0 +1,15 @@
+# frozen_string_literal: true
+
+class CheckDeprecatedLlmsWorker
+  include Sidekiq::Worker
+
+  DEPRECATION_MESSAGE = ':calendar: LLMs deprecated already or within 1 week:'
+
+  def perform
+    deprecated_llms = Evidence::Research::GenAi::Llm.in_use.deprecated(as_of: 1.week.from_now).map(&:to_s)
```
dandrabik: Take or leave, it might be worth piggy-backing off of my query to get the `prompts` that are being deprecated. That might be easier for Curriculum to take action on.

dandrabik (approved):
Health check looks good. Small note about an 'active' scope that I think you need for the deprecation check, and a possible suggestion to show deprecated prompts instead of llms.

Also, you might need to filter out non-live activities, otherwise this will pick up archived and alpha activities (not all of which get upgraded if they aren't in use).

services/QuillLMS/engines/evidence/app/models/evidence/research/gen_ai/llm.rb
dandrabik: This also probably needs to be scoped to the Activity status to not pick up archived and alpha activities.

brendanshean: Good point.  Definitely don't want noise on those.

PR #12270, opened 2024-08-22
rails 7.1

happythenewsad:
## WHAT
- bump Rails to 7.1

## WHY
- keep the app up to date 

## HOW
- pin various gems to reduce the net code change
- address various breaking changes in our codebase 

## NOTE
`grpc` compilation (a gem we depend on) is longer than the CircleCI build timeout (which is not extensible unless we buy the scale CircleCI plan). This is an open issue (https://github.com/grpc/grpc/issues/35874)

I solved this by temporarily increasing the resource_class to xlarge, then reverting that change. So we'll need to do that again if we ever need to expire the cached grpc, assuming the issue is not addressed upstream.

### Screenshots
(If applicable. Also, please censor any sensitive data)

### Notion Card Links
(Please provide links to any relevant Notion card(s) relevant to this PR.)

### What have you done to QA this feature?
- [x] As a staffer, click into a Connect title card, ensure no rendering errors
- [x] Login with Gmail
- [x]  As premium admin, check that admin snapshot displays correctly
- [x] As a staffer, complete an Evidence question
- [x] As a student, complete a Connect activity
- [x] Login as teacher, assign an activity
- [x] monitor Sidekiq for unusual errors 
- [x] monitor Newrelic for unusual errors

PR Checklist | Your Answer
------------ | -------------
Have you added and/or updated tests? |  (The answer should mostly be 'YES'. If you answer 'NO', please justify.)
Have you deployed to Staging? | (Possible answers: YES, Not yet - deploying now!, NO - non-app change, NO - tiny change)
Self-Review: Have you done an initial self-review of the code below on Github? |

PR #13338, opened 2025-10-07
Ruby34 2

happythenewsad:
## WHAT
- upgrade to Ruby 3.4

## WHY
- planned maintenance 

## HOW
- touch CCI, Gemfiles, etc 

### NOTES
even though this was presented as a warning, it was actually the cause of evidence spec failures:

```
pk@Peters-MacBook-Pro-2 evidence % bin/rspec ./spec/models/evidence/research/gen_ai/trial_spec.rb
/Users/pk/.rbenv/versions/3.4.1/lib/ruby/gems/3.4.0/gems/pry-0.14.2/lib/pry/command_state.rb:3: warning: ostruct was loaded from the standard library, but will no longer be part of the default gems starting from Ruby 3.5.0.
You can add ostruct to your Gemfile or gemspec to silence this warning.
```

### Notion Card Links
https://www.notion.so/quill/Product-Board-30f97e4eb01246dbb8264253fb385073?p=237d42e6f94180deb377dccccf4ffe1b&pm=s

- [x] As a staffer, click into a Connect title card, ensure no rendering errors
- [x] Login with Gmail
- [x]  As premium admin, check that admin snapshot displays correctly
- [x] As a staffer, play an Evidence question
- [x] As a student, complete a Connect activity
- [x] Login as teacher, assign an activity
- [x] monitor Sidekiq for unusual errors 
- [x] monitor Newrelic for unusual errors

PR Checklist | Your Answer
------------ | -------------


happythenewsad:
Note: similar to #12270 
I temporarily bumped the CCI compute size, so that grpc compilation would not time out

services/QuillLMS/Gemfile
```diff
@@ -63,7 +63,8 @@ gem 'google-apis-classroom_v1', '~> 0.27.0'
 gem 'google-cloud-ai_platform-v1', '0.32.0'
 gem 'google-cloud-bigquery', '~> 1.43.0'
 gem 'google-cloud-storage', '~> 1.52.0'
-gem 'grover'
+gem 'grover', '~> 1.2.2'
+gem 'grpc', '~> 1.75.0'
```
brendanshean: I think you can do `bundle update grpc` without specifying the requirement in the Gemfile which would be preferable since `grpc` is really just a dependency on various google-cloud libraries and not a required Gem by itself.

services/QuillLMS/engines/evidence/evidence.gemspec
```diff
@@ -15,14 +15,15 @@ Gem::Specification.new do |s|
...
-  s.required_ruby_version = '3.3.7'
+  s.required_ruby_version = '3.4.1'
 
   s.files = Dir['{app,config,db,lib}/**/*', 'MIT-LICENSE', 'Rakefile', 'README.rdoc']
 
   s.add_dependency 'active_model_serializers', '0.10.14'
   s.add_dependency 'diff-lcs', '~> 1.5.0'
   s.add_dependency 'google-cloud-ai_platform', '1.0.0'
   s.add_dependency 'google-cloud-storage', '~> 1.52.0'
+  s.add_dependency 'grpc', '~> 1.75.0'
```
brendanshean: I think you can do `bundle update grpc` without specifying the requirement in the Gemfile which would be preferable since `grpc` is really just a dependency on various google-cloud libraries and not a required Gem by itself.

services/QuillLMS/spec/controllers/invitations_controller_spec.rb
```diff
@@ -121,7 +121,7 @@
       it 'should respond with the error' do
         delete :destroy_pending_invitations_to_specific_invitee, params: { invitation_type: 'anything', invitee_email: 'some@test.com' }
         parsed_response = JSON.parse(response.body)
-        expect(parsed_response['error']).to include("undefined method `destroy' for nil")
+        expect(parsed_response['error']).to include("undefined method 'destroy' for nil")
```
anathomical, line 124: This is an interesting change in the upgrade.

anathomical (approved):
Thanks for pulling this together.  Always nice to keep our code up to date.

services/QuillLMS/engines/evidence/evidence.gemspec
```diff
@@ -34,6 +35,7 @@ Gem::Specification.new do |s|
   s.add_development_dependency 'dotenv-rails', '2.7.6'
   s.add_development_dependency 'factory_bot_rails'
   s.add_development_dependency 'faker', '3.2.0'
+  s.add_development_dependency 'observer'
```
brendanshean: What is `observer` for? If it's required,  do we need it in main app?

brendanshean (approved):

eadams17 (approved):

services/QuillLMS/engines/evidence/evidence.gemspec
happythenewsad: We don't - the main app is on factory-bot 6, and evidence is on factory-bot 4.

happythenewsad: See the 'notes' section of this PR. We're not supposed to need the observer gem, but i found through trial and error that it was the cause issues with loading factory bot

services/QuillLMS/engines/evidence/evidence.gemspec
happythenewsad: good call - updated

services/QuillLMS/engines/evidence/evidence.gemspec
brendanshean: Maybe put a comment that we should look to remove when we upgrade factory bot on evidence?

PR #13339, opened 2025-10-07
fix-sls-completion-issue

anathomical:
## WHAT
- Replace `perform_async` with `perform_in(5.seconds)`
## WHY
There appears to be some sort of database-based race condition going on that involves `perform_async` running with a `classroom_unit.id` that can't be found if run immediately.  This shouldn't be happening since the `perform_async` call happens only after `classroom_unit.update` returns `true`, but maybe there's something going on with threads and transaction delays or something?  Anyway, this is the fix that we used in the code that generates data for the demo account, so it seems reasonable to try for data more broadly so that we can turn off our clean-up Cron jobs that go back later to backfill this data.
## HOW
- Straightforward replacement of `perform_async` with `perform_in(5.seconds)`.  We'll want to monitor Sidekiq to make sure that a large volume of `perform_in` calls doesn't cause performance issues.  Documentation for Sidekiq suggests that this shouldn't be a problem, but we'll want to make sure.

- On staging confirm that jobs get enqueued and run successfully

PR Checklist | Your Answer
------------ | -------------


dandrabik (approved):
This looks good for testing. Hoping this works!

happythenewsad (approved):

PR #13343, opened 2025-10-07
diagnostic-insights-tweak-alarm-code

anathomical:
## WHAT
Tweak the data alarm to exclude timing integrity errors for users who joined the classroom after the classroom_units were created
## WHY
Some data discrepancies in the Diagnostic Insights reports are caused by a bug with `assign_on_join` which would assign students to do practice before they completed the diagnostic.  Sometimes we can tell this happened because the `classroom_unit` is still marked `assign_on_join`, but some are not, so we've decided to fudge this by checking to make sure the student was already in the classroom when the assignment was created.

This will catch students who have data that is corrupted because the of the `assign_on_join` bug even if the `assign_on_join` flag is no longer set.  This may also exclude some number of students who should be flagged with bad data anyway, but we've decided that reducing reporting noise is a higher priority here.
## HOW
Add an additional check to compare the time the student joined the classroom to the earliest classroom units, and if the student joined after the classroom unit was created and may be subject to the `assign_on_join` bug we don't alarm for that student even if they have practice before diagnostic completions.

### Notion Card Links
https://www.notion.so/quill/Possible-assign_on_join-bugs-with-no-conclusive-proof-25ed42e6f94180d0960effea926094e9

PR Checklist | Your Answer
------------ | -------------


services/QuillLMS/app/services/diagnostic_insights/data_integrity/verify_diagnostic_before_practice.rb
```diff
@@ -46,6 +47,18 @@ def run
           .where(student_learning_sequence_activities: { classroom_units: { assign_on_join: true } })
           .exists?
       end
+
+      private def student_joined_classroom_after_classroom_unit(student_id)
+        # A classroom_unit may have been assign_on_join when a student joined the classroom, but no is no longer
```
brendanshean: but **now** is no longer?

brendanshean (approved):
This

services/QuillLMS/app/services/diagnostic_insights/data_integrity/verify_diagnostic_before_practice.rb
anathomical: Yeah, sorry, that is missing context.  We discovered a bug last spring which sometimes set the `assign_on_join` flag to `true` on Diagnostic Practice Recommendations despite that code trying to explicitly set it to `false`.  (There was an `after_save` callback that set the flag.)  That bug has been fixed, so this no longer happens, which means we only need to resolve existing data discrepancies.

brendanshean: Sorry, the context was clear.  I was just pointing out a typo in the comment.

anathomical: Haha.  Whoops!  Thanks, I'll update.

PR #13353, opened 2025-10-14
lessons tool page animation

emilia-friedberg:
## WHAT
Add the animation to the lessons tool page, with changes to the HTML/CSS to make sure it renders correctly at all breakpoints (with the seamless background gradient).

## WHY
We want this animation in place, like the other tool pages.

## HOW
Rather than having that one background section that wraps behind the image, like we do on the other tool pages, I updated this so the header has its own background and then there's an additional one in the `tool-section`, so that it looks like one background that matches the background in the animation, which is necessary due to the screen-switching effect. You can see how it scales in the attached video.

### Screenshots
https://share.zight.com/z8ugeR9z

### Notion Card Links
https://www.notion.so/quill/Quill-Lessons-Tool-Landing-Page-Animation-288d42e6f94180f89449f3012e83ef75?source=copy_link

Tested locally and had Jack QA on staging.

PR Checklist | Your Answer
------------ | -------------
Have you added and/or updated tests? | NO - erb


anathomical (approved):
I really appreciate the video to show off the dynamic behavior that this is intended to support.

eadams17 (approved):
nice solution!

PR #13357, opened 2025-10-14
update version of Activity Scores: Student Overview that shows in the admin panel to not render links to more granular reports

emilia-friedberg:
## WHAT
Update the version of Activity Scores: Student Overview that shows in the admin panel to not render links to more granular reports.

## WHY
These links weren't working. I suspect that had been the case for several years, as I couldn't find anything in particular that seemed to be the culprit, but since the same component was being used in both the teacher and admin version of this report, they were linking to the teacher reports. Those pages don't load, because the queries behind them are pulling data for the `current_user`, which in this case was the admin. I discussed this with Nattalie and we decided the easiest fix was just to not render links here for admins. They still have access to the more granular data if they want to log in to a teacher's account from the Account Management page.

## HOW
Pass down a prop and don't render links if it's true.

### Screenshots

### Notion Card Links
update version of Activity Scores: Student Overview that shows in the admin panel to not render links to more granular reports

Tested locally.

PR Checklist | Your Answer
------------ | -------------


services/QuillLMS/client/app/bundles/PremiumHub/components/activity_scores_student_overview.tsx
```diff
@@ -1,15 +1,15 @@
...
+import StudentOverview from '../../Teacher/components/progress_reports/student_overview';
 
 export const ActivityScoresStudentOverview = ({ location }) => {
   const previous = '/teachers/premium_hub/district_activity_scores';
   const imageSrc = 'https://assets.quill.org/images/icons/chevron-dark-green.svg';
 
   return (
     <div className="container gray-background-accommodate-footer">
-      <StudentOverview location={location} >
+      <StudentOverview inAdminView={true} location={location} >
```
happythenewsad, line 12: This can be shortened to `inAdminView`

happythenewsad (approved):

eadams17 (approved):

services/QuillLMS/client/app/bundles/PremiumHub/components/activity_scores_student_overview.tsx
eadams17: it can be, but our preference is to be explicit with prop syntax

PR #13362, opened 2025-10-16
consolidate fill in blank form

emilia-friedberg:
## WHAT
Consolidate the various fill in the blank forms into one file.

## WHY
This is high-priority consolidation work since we'll be moving fill in the blanks over to Grammar soon.

## HOW
The files themselves were pretty similar already, so I just extracted the code into one place.

### Screenshots

### Notion Card Links

Tested creation and editing of fill in the blank questions in Diagnostic and Connect.

PR Checklist | Your Answer
------------ | -------------


services/QuillLMS/client/app/bundles/Shared/components/fillInBlank/fillInBlankForm.tsx
```diff
@@ -0,0 +1,340 @@
...
+      blankAllowed: false,
+      caseInsensitive: false,
+      prompt: '',
+      newQuestionOptimalResponse: '',
+      instructions: '',
+      conceptID: '',
+      flag: 'alpha',
+      cuesLabel: '',
+      cues: '',
+      showDefaultInstructions: false
```
anathomical: It looks to me like this might be exactly the same values as those that are set above in `defaultState`.  If that's true, should we split this off into a shared constant that is referenced both there and here?

services/QuillLMS/client/app/bundles/Shared/components/fillInBlank/fillInBlankForm.tsx
```diff
@@ -0,0 +1,340 @@
...
+          {defaultInstructionsButtons}
+        </div>
+      </div>
+    );
+  };
+
+  submit = (e?: React.FormEvent) => {
+    if (e) {
+      e.preventDefault();
+    }
```
anathomical, line 145: I don't think I've seen conditional `preventDefault` calls like this before.  This feels like a good way of handling the fact that this is done in different ways in different call sites.

services/QuillLMS/client/app/bundles/Shared/components/fillInBlank/fillInBlankForm.tsx
```diff
@@ -0,0 +1,340 @@
...
+
+    const data = {
+      prompt,
+      blankAllowed: blankAllowed || false,
+      caseInsensitive: caseInsensitive || false,
+      cues: cues.split(','),
+      instructions,
+      conceptID,
+      flag: flag || 'alpha',
+      cuesLabel
```
anathomical, line 168: It looks like there's another potential parallel to the default values I flagged at the top of the file.  This one's less clear, but I wonder if there's a clean way of using `Object.assign` or something similar to override default keys if there's a non-empty value here so that the right sides of the `||` calls are clearly marked as the shared default that they are?

anathomical (approved):
Some thoughts on how to reference shared "default" or fallback values, but overall I think this could go out as-is without issue.

eadams17 (approved):

PR #13367, opened 2025-10-17
drop-unused-indexes

anathomical:
## WHAT
Drop unused indexes
## WHY
This will save both disk space in Postgres, and reduce the RAM footprint of our current indexes to improve overall performance
## HOW
- Use `heroku pg:diagnose` to get potential unused indexes
- Review by hand, digging into potential edge case uses of those indexes that might justify them
- Backstop with an analysis from LLMs
- Get review from other engineers to flag any additional edge cases we might have missed

### Notion Card Links
https://www.notion.so/quill/Drop-Unused-indexes-from-larger-tables-28cd42e6f94180d3b71ce81332460c3a#28cd42e6f9418023a0f8f74231098bbe

- Run locally, confirm all specs pass
- Deploy to staging and run
- Test teacher reports and activity playthroughs

PR Checklist | Your Answer
------------ | -------------
Have you added and/or updated tests? | No, no expected behavior changes


happythenewsad (approved):

services/QuillLMS/db/migrate/20251017153523_drop_unused_indexes.rb
```diff
@@ -0,0 +1,15 @@
+# frozen_string_literal: true
+
+class DropUnusedIndexes < ActiveRecord::Migration[7.1]
+  def change
+    remove_index :users, :time_zone
```
dandrabik: `users` is a large table (as are some of the other ones) so I believe we need to do something like  this to make sure Postgres doesn't lock the whole table (worth googling 'Rails Postgres concurrent drop index no table lock' or something like that.
```ruby    
class RemoveIndexFromTableName < ActiveRecord::Migration[7.1]
  disable_ddl_transaction! # Required for concurrent index operations

  def change
    remove_index :table_name, name: :index_name, algorithm: :concurrently
   end
 end
```

dandrabik (approved):
So I am approving this because I don't need to re-review it, but there is required code to make sure this doesn't lock tables during the migration.

PR #13377, opened 2025-10-21
bump-sls-delay

anathomical:
## WHAT
Add additional delay to our SLS-processing `perform` calls, going from 5 seconds to 15 for `HandleCompletion` and 0 to 15 seconds for `HandleCompletion`
## WHY
Going from 0 to 5 seconds on `HandleCompletion` seems to have reduced the number of incidents of records that encountered an unknown race condition and needed later clean-up by about 90%, so bumping this up to 15 seconds should be safe and will hopefully get us closer to 100% reduction.
## HOW
- Change the delay on `HandleCompletionWorker` from `5.seconds` to `15.seconds`
- Convert the `HandleAssignmentWorker.perform_async` to `HandleAssignmentWorker.perform_in(15.seconds, ...)`

- Deploy to staging
- Assign a pre-diagnostic, confirm that new `StudentLearningSequence` and associated `StudentLearningSequenceActivity` records get created
- Complete the diagnostic and confirm that the associated `StudentLearningSequenceActivity` is updated with completion data

PR Checklist | Your Answer
------------ | -------------


services/QuillLMS/app/controllers/api/v1/activity_sessions_controller.rb
brendanshean, line 1: You could consider making this an ENV variable.  Then no more PRs needed to tune.

brendanshean (approved):

services/QuillLMS/app/controllers/api/v1/activity_sessions_controller.rb
dandrabik: Yeah, as we chatted about, I think an ENV var here makes sense, e.g.
```ruby
COMPLETION_DELAY = ENV.fetch('LEARNING_SEQUENCES_COMPLETION_DELAY', '15').to_i

File: services/QuillLMS/app/controllers/api/v1/activity_sessions_controller.rb [hunk 1]
```diff
@@ -84,7 +84,7 @@ def destroy
   private def handle_student_activity_sequence_activity_completion
     return unless @activity_session.completed_at
 
-    StudentLearningSequences::HandleCompletionWorker.perform_in(5.seconds, @activity_session.id)
+    StudentLearningSequences::HandleCompletionWorker.perform_in(15.seconds, @activity_session.id)
```
dandrabik: Note, `perform_in(15, param)` and `perform_in(15.seconds, param)` are equivalent since the first var is converted to an `int` of seconds. But you may find the later more readable.

dandrabik (approved):
Overall looks good. I left some small notes.

PR #13382, opened 2025-10-21
update-diagnostic-insights-alarm

anathomical:
## WHAT
Tweak service for consistent user matching
## WHY
Sometimes teacher fixes will update data in such a way that the `StudentLearningSequence` points at a user that doesn't match the `ActivitySession`.  We want the two practice services to match, so we should filter on the same user_id column in both
## HOW
Update the query for skill practice lookups to match `user_id` on `StudentLearningSequence` rather than `ActivitySession` so that it matches the other practice related queries we do.

### Notion Card Links
https://www.notion.so/quill/Possible-issue-with-incomplete-merge_students-Teacher-Fix-25ed42e6f94180068499d637a9792360

- Run the snapshot generator and confirm data remains consistent pre and post change
- Deploy to staging
- Spot check a handful of teachers to confirm that counts between different pages remain consistent
- Additionally spot check known cases where this issue arose in data alarms (teacher id 13020429)

PR Checklist | Your Answer
------------ | -------------
Have you added and/or updated tests? | No, we do not expect changes in behavior except in edge cases where data has gotten into a weird shape


services/QuillLMS/lib/snapshots/diagnostic_insights/history_by_classroom.json
anathomical, line 1: All changes in this file are about `activity_session.visible` changes caused by a historical classroom being archived since the last time we ran this snapshot generator, but the rest of the data is identical.

services/QuillLMS/lib/snapshots/diagnostic_insights/practice_by_classroom.json
anathomical, line 1: The changes in this file appear to be a result of non-deterministic ordering, not actual changes to results.

services/QuillLMS/lib/snapshots/diagnostic_insights/skill_practice_by_classroom.json
anathomical, line 1: As with `practice_by_classroom`, this appears to be an issue of non-deterministic ordering in the data, not a change in the actual substance.

brendanshean (approved):

PR #13386, opened 2025-10-22
remove question class and lots of associated lib files

emilia-friedberg:
## WHAT
We had these `lib/question.js` files in Diagnostic, Connect, and Lessons (entirely unused in Lessons), which were older versions of code that now lives in `quill-marking-logic` -- they weren't being used when students actually play the activity, but were in use on the admin side for rematching a single question. I replaced them with the previously-nested `checkAnswer` files (same pattern as Grammar) that just call the actual marking logic, and that let me delete a ton of other files that were being called there and test files only used for those files. 

## WHY
We want to always be using the same grading logic regardless of whether the question is being submitted from the student view or the admin view.

## HOW
See "What".

### Screenshots

### Notion Card Links

Tested rematching a single question (the only place the old logic was being used) for Connect and Diagnostic and confirmed that it works correctly.

PR Checklist | Your Answer
------------ | -------------


brendanshean (approved):
I really appreciate this Fall Cleaning session.

happythenewsad (approved):

PR #13387, opened 2025-10-22
No yjit for workers

happythenewsad:
## WHAT
Disable YJIT for worker dynos 

## WHY
we notice potentially dangerous memory overages in some workers after enabling YJIT. web dynos are fine though.

## HOW
pass in a Procfile flag, to override the system-wide YJIT flag. 
Note:  With a 24 hour lookback, only `worker` and `reportworker` have excessive memory, so we really only need to apply the flag to those two. However, "all workers don't have YJIT" is an easier concept to track, and avoids issues with future queues.

### Screenshots

### Notion Card Links

I used a dummy class to confirm that the flag works as intended:

```ruby
+class DeletemeWorker
+  include Sidekiq::Worker
+
+  def perform
+    if defined?(RubyVM::YJIT) && RubyVM::YJIT.enabled?
+      puts "YJIT is enabled"
+    else
+      puts "YJIT is not enabled"
+    end
+  end
+end
```

PR Checklist | Your Answer
------------ | -------------


dandrabik (approved):
Looks good, let's try this and monitor.

PR #13391, opened 2025-10-22
STI Example refactor

brendanshean:
## WHAT
This is a major refactor of the dataset editor / trial runner involving `Evidence::Research::GenAi::{Prompt, Test, Draft}Example}` into one model `Evidence::Research::GenAi::Example` with subclasses for each of the three types.

## WHY
This will simplify the frontend and backend logic with respect to changes that are made in the dataset editor.  For an example, not the diff in payload shape with `client/app/bundles/Staff/components/evidence/datasetEditor/hooks/useLocalChanges.ts`

## HOW
1. Add migration for new Example table with type field and then all fields in prompt example
2. Update all frontend code to handle manipulation of local changes in terms of examples rather than promptExamples, testExamples and draftExamples
3. Update backend code to handle this new payload from the frontend in addition to the the creation/updating/deletion of these records

Note:  One major change in keys is the addition of `usage` and `prev_usage` keys as a proxy for `type` for the STI.

### Screenshots

### Notion Card Links

Tested extensively but also waiting for QA with curriculum next week.

PR Checklist | Your Answer
------------ | -------------


services/QuillLMS/client/app/bundles/Staff/components/evidence/datasetEditor/__tests__/useDatasetEditor.test.tsx
```diff
@@ -81,32 +81,22 @@ describe('useDatasetEditor', () => {
...
-      promptExamples: {
-        created: [],
-        updated: {},
-        deleted: []
-      },
-      draftExamples: {
+      examples: {
         created: [],
         updated: {},
         deleted: []
```
emilia-friedberg, line 87: Even just this little piece of cleanup feels soothing to my brain.

services/QuillLMS/client/app/bundles/Staff/components/evidence/datasetEditor/__tests__/useDatasetEditor.test.tsx
```diff
@@ -81,32 +81,22 @@ describe('useDatasetEditor', () => {
...
       }
     });
 
-    if (mockApiResponse.test_examples && mockApiResponse.test_examples.length > 0) {
-      const updatedExample = { ...mockApiResponse.test_examples[0], student_response: 'Updated text' };
+    const testExamples = mockDatasetEditorData.examples.filter(ex => ex.usage === 'test');
```
emilia-friedberg: NBD but we probably want to use the same constants in the test file that we do in implementation (ie, `USAGE_TEST`).

services/QuillLMS/client/app/bundles/Staff/components/evidence/datasetEditor/components/clusterTable/ClusterRow.tsx
```diff
@@ -175,7 +175,7 @@ export const ClusterRow: React.FC<ClusterRowProps> = ({
 
     // Only consider TEST examples for trial results (not prompt or draft examples)
     const testExampleIds = new Set(
-      examples.filter(ex => ex.usage === 'test').map(ex => ex.id)
+      examples.filter(ex => ex.usage === USAGE_TEST).map(ex => ex.id)
```
emilia-friedberg, line 195: Love seeing these constants!

services/QuillLMS/client/app/bundles/Staff/components/evidence/datasetEditor/hooks/helpers/exampleHelpers.ts
```diff
@@ -172,94 +22,43 @@ export const handleUsageChange = (
...
-      draftExamples: {
-        ...prev.draftExamples,
-        created: fromDraft
-          ? prev.draftExamples.created.filter(ex => ex.id !== exampleId)
-          : toDraft
-            ? [...prev.draftExamples.created, newExample]
-            : prev.draftExamples.created,
-        deleted: shouldAddToDeleted && fromDraft && deletedExample?.replacesId
-          ? Array.from(new Set([...prev.draftExamples.deleted, deletedExample.replacesId]))
-          : prev.draftExamples.deleted
```
emilia-friedberg, line 236: :raised_hands::raised_hands::raised_hands:

services/QuillLMS/client/app/bundles/Staff/components/evidence/datasetEditor/hooks/useDatasetEditor.ts
```diff
@@ -297,21 +299,41 @@ export const useDatasetEditor = (datasetId: string): UseDatasetEditorReturn => {
...
-    const originalPromptExamples = data.prompt_examples.filter(ex =>
-      ex.usage === 'prompt' && ex.curriculum_assigned_status === curriculum_assigned_status
-    );
+    const updatedToPromptExampleIds = Object.keys(localChanges.examples.updated)
+      .filter(id => {
+        const updates = localChanges.examples.updated[id];
+        const originalExample = data.examples.find(ex => String(ex.id) === id);
+        if (!originalExample) return false;
+
+        const finalUsage = updates.usage !== undefined ? updates.usage : originalExample.usage;
```
emilia-friedberg, line 357: Can usage ever be null? Just wondering why we're specifically checking `!== undefined` here.

emilia-friedberg (approved):
This cleanup makes so much of the code that was previously difficult to parse easier to read and more flexible if we wanted to add additional types in the future -- feels really good to see!

services/QuillLMS/client/app/bundles/Staff/components/evidence/datasetEditor/hooks/useDatasetEditor.ts
brendanshean: That's a good question.  The originalExample is always `UsageType` so never nullish.  However, updates.usage is `UsageType | undefined` since updates is `Partial<Example>` (e.g. an update for certain keys but not `usage`)

services/QuillLMS/client/app/bundles/Staff/components/evidence/datasetEditor/__tests__/useDatasetEditor.test.tsx
brendanshean: Thanks for this call out.  I definitely missed a bunch in here.

services/QuillLMS/db/migrate/20251023174711_convert_examples_to_sti.evidence.rb
```diff
@@ -0,0 +1,180 @@
...
+    end
+
+    llm_fk_count = 0
+    test_example_id_mapping.each do |old_id, new_id|
+      llm_fk_count += LlmExample.where(test_example_id: old_id).update_all(test_example_id: new_id)
+    end
+
+    Rails.logger.debug "Updated #{prompt_fk_count} llm_prompt_prompt_examples, #{llm_fk_count} llm_examples\n\n"
+
+    drop_table :evidence_research_gen_ai_prompt_examples
```
dandrabik: You might consider making the drops a different migration. That would give you the opportunity to spot check the data before dropping the old tables.

services/QuillLMS/db/migrate/20251023174711_convert_examples_to_sti.evidence.rb
```diff
@@ -0,0 +1,180 @@
...
+  class LlmPromptPromptExample < ApplicationRecord
+    self.table_name = 'evidence_research_gen_ai_llm_prompt_prompt_examples'
+  end
+
+  class LlmExample < ApplicationRecord
+    self.table_name = 'evidence_research_gen_ai_llm_examples'
+  end
+
+  def up
+    create_table :evidence_research_gen_ai_examples do |t|
```
dandrabik: Should you add indexes to this table?

services/QuillLMS/db/migrate/20251023174711_convert_examples_to_sti.evidence.rb
```diff
@@ -0,0 +1,180 @@
...
+      t.text :student_response, null: false
+      t.string :type, null: false
+      t.integer :cluster_id
+      t.integer :dataset_id, null: false
+      t.timestamps
+    end
+
+    prompt_example_id_mapping = {}
+    PromptExample.find_each do |prompt_example|
+      Example.create!(
```
dandrabik: Take or leave, but I'm not sure why you need to pick the id here instead of having it auto assigned, e.g.
```ruby
new_example = Evidence::Research::GenAi::PromptExample.create(prompt_example.attributes.except('id'))
prompt_example_id_mapping[prompt_example.id] = new_example.id

File: services/QuillLMS/db/migrate/20251023174711_convert_examples_to_sti.evidence.rb [hunk 9]
```diff
@@ -0,0 +1,180 @@
...
+      t.text :curriculum_proposed_feedback
+      t.text :highlight
+      t.text :notes
+      t.text :student_response, null: false
+      t.string :type, null: false
+      t.integer :cluster_id
+      t.integer :dataset_id, null: false
+      t.timestamps
+    end
+
```
dandrabik: Note, for this data migration, I'd strongly recommend a few SQL queries to check totals and maybe some group_by counts. Then I'd spot check a few records. I'd also follow this order:

1. Run migration
2. Do SQL checks and spot checks to confirm everything
3. Drop old tables.

services/QuillLMS/db/migrate/20251023174711_convert_examples_to_sti.evidence.rb
```diff
@@ -0,0 +1,180 @@
...
+    Rails.logger.debug "Migrated #{prompt_example_id_mapping.size} PromptExamples (IDs unchanged)"
+
+    max_id = Example.maximum(:id) || 0
+
+    test_example_id_mapping = {}
+    TestExample.find_each do |test_example|
+      new_id = test_example.id + max_id
+      Example.create!(
```
dandrabik: Just noting that this job is not idempotent. If one of these `create!` errors raises, you'd have to do a bunch of data clean, then restart it (otherwise, it would create duplicates for the ones that were already created, and you'd lose the mapping).

services/QuillLMS/db/migrate/20251023174711_convert_examples_to_sti.evidence.rb
```diff
@@ -0,0 +1,180 @@
...
+      test_example_id_mapping[test_example.id] = new_id
+    end
+
+    Rails.logger.debug "Migrated #{test_example_id_mapping.size} TestExamples with ID offset +#{max_id}"
+    Rails.logger.debug "TestExample ID mappings: #{test_example_id_mapping.inspect}\n\n"
+
+    max_id = Example.maximum(:id) || 0
+
+    draft_count = 0
+    DraftExample.find_each do |draft_example|
```
dandrabik: Take or leave, having done a migration like this in the past (albeit on customer-facing data so I had to be a little more deliberate), you might consider a different structure for this job, i.e. adding a reference the `example` table that you will later drop like `original_type` `original_id` and adding a relation to the original models
```ruby
# PromptExample` 
has_one :evidence_research_gen_ai_prompt_example ...

# in migration:
prompt_example.evidence_research_gen_ai_prompt_example.create(prompt_example.attributes.except('id'))

# You can even make the job idempotent
PromptExample.without_new_prompt_example.find_each 
```
The upside is the migration is idempotent, the migration becomes a little simpler, the mapping logic is permanent (until you drop it) instead of being ephemeral, and the auditing / checks are a little easier.

Maybe this is overkill for the internal staff data, and if your script runs and doesn't hit issues, it should be fine. Happy to chat through this if you'd like.

dandrabik (commented):
Leaving this as a comment since I suggested a good amount of changes, but it's probably better to just chat through.

services/QuillLMS/db/migrate/20251023174711_convert_examples_to_sti.evidence.rb
brendanshean: Excellent point

services/QuillLMS/db/migrate/20251023174711_convert_examples_to_sti.evidence.rb
brendanshean: I should probably add one for `type`.  I was going to wait on adding others.

services/QuillLMS/db/migrate/20251023174711_convert_examples_to_sti.evidence.rb
brendanshean: I really like this idea.  

The mapping is kind of clunky since we have to somehow import the data in if we need to fix something.

services/QuillLMS/db/migrate/20251029210337_convert_examples_to_sti.evidence.rb
```diff
@@ -0,0 +1,370 @@
...
+  class DraftExample < ApplicationRecord
+    self.table_name = 'evidence_research_gen_ai_draft_examples'
+
+    has_one :example,
+      -> { where(original_type: 'DraftExample') },
+      foreign_key: :original_id,
+      primary_key: :id,
+      class_name: 'ConvertExamplesToSti::Example'
+
+    scope :without_new_example, -> { where.not(id: Example.where(original_type: 'DraftExample').pluck(:original_id)) }
```
dandrabik, line 66: Take or leave, I think you can do this slightly simpler (I'm not sure if you'll need to adjust the pluralization for a has_one):
```ruby
scope :without_new_example, -> { left_joins(:examples).where(examples: { id: nil }) }

File: services/QuillLMS/db/migrate/20251029210337_convert_examples_to_sti.evidence.rb [hunk 13]
```diff
@@ -0,0 +1,370 @@
...
+
+    error_msg = "Found unexpected foreign key references that need to be migrated:\n  - #{unexpected_references.join("\n  - ")}"
+    Rails.logger.error error_msg
+    raise error_msg
+  end
+
+  private def migrate_prompt_examples
+    count = 0
+    PromptExample.without_new_example.find_each do |prompt_example|
+      Example.create!(
```
dandrabik, line 174: Take or leave, but I believe this would work:
```ruby
prompt_example.example.create!(prompt_example.attributes.except('id'))

File: services/QuillLMS/db/migrate/20251029210337_convert_examples_to_sti.evidence.rb [hunk 14]
```diff
@@ -0,0 +1,370 @@
...
+    Rails.logger.debug 'Updating foreign keys...'
+
+    prompt_fk_count = update_prompt_example_foreign_keys
+    llm_example_fk_count = update_test_example_foreign_keys
+
+    Rails.logger.debug "Updated #{prompt_fk_count} llm_prompt_prompt_examples, #{llm_example_fk_count} llm_examples"
+  end
+
+  private def update_prompt_example_foreign_keys
+    referenced_ids = LlmPromptPromptExample.distinct.pluck(:prompt_example_id).compact
```
dandrabik, line 233: Won't this be like 200K ids put into a SQL statement? Do we need this? Couldn't the update just do every one (if the record was already updated, it just wouldn't change anything)

services/QuillLMS/db/migrate/20251029210337_convert_examples_to_sti.evidence.rb
```diff
@@ -0,0 +1,370 @@
...
+    verify_spot_checks
+    verify_foreign_keys
+
+    Rails.logger.debug "\n#{'=' * 80}"
+    Rails.logger.debug 'MIGRATION SUCCESSFUL - SAFE TO DROP OLD TABLES'
+    Rails.logger.debug '=' * 80
+  end
+
+  private def verify_spot_checks
+    Rails.logger.debug "\nSpot checking first record of each type..."
```
dandrabik: Spot checking the first record of a set (and only the first record) isn't a particularly good spot check. I'd consider doing this manually, or pulling a sample of records throughout the set, plus the last record.

dandrabik: Also, a more comprehensive check is if you join the two tables, return any records in which all of the fields don't match. This plus counts is a pretty thorough check (could take the place of spot checking)

```sql
SELECT count(id)
FROM join_tables. ...
WHERE 
  old.automl_label != new.automl_label
  AND old.automl_primary_feedback != new.automl_primary_feedback
  ...

File: services/QuillLMS/db/migrate/20251029210337_convert_examples_to_sti.evidence.rb [hunk 16]
```diff
@@ -0,0 +1,370 @@
...
+
+  private def raise_if_unexpected_references_found(unexpected_references)
+    return if unexpected_references.empty?
+
+    error_msg = "Found unexpected foreign key references that need to be migrated:\n  - #{unexpected_references.join("\n  - ")}"
+    Rails.logger.error error_msg
+    raise error_msg
+  end
+
+  private def migrate_prompt_examples
```
dandrabik, line 171: Also note, I think looping through the records is preferred, but if you still end up hitting performance issues, as a secondary option we could do this as a mass insert (or upsert) into the new tables.

dandrabik (approved):
Overall looks good. Some suggestions about how to improve the spot checks and some take or leave. But looks good to ship if looks good on staging!

Nice work on this!

services/QuillLMS/db/migrate/20251029210337_convert_examples_to_sti.evidence.rb
brendanshean: I did some manual sampling but I like the idea of checking the end too.

services/QuillLMS/db/migrate/20251029210337_convert_examples_to_sti.evidence.rb
brendanshean: Much cleaner.  I will probably leave as-is since QA is testing the current implementation.

But good to keep in mind for the future.

services/QuillLMS/db/migrate/20251029210337_convert_examples_to_sti.evidence.rb
brendanshean: Doing it one-by-one was one of the bottlenecks I encountered.  This ended up being pretty fast on staging.

services/QuillLMS/db/migrate/20251029210337_convert_examples_to_sti.evidence.rb
brendanshean: I also like comprehensive check.

services/QuillLMS/.gitignore
```diff
@@ -8,6 +8,7 @@
 
 # AI tools
 CLAUDE.md
+AGENTS.md
```
brendanshean, line 11: For codex

PR #13393, opened 2025-10-22
postmark-inactive-recipient-errors

anathomical:
## WHAT
- Add a general `rescue_from Postmark::InactiveRecipientError` to the `ApplicationMailer` so that these failures don't cause unhandled exceptions
- Send those exceptions to `ErrorNotifier.report` and then swallow them
- Tweak the most common source of this error, the `SendRollupEmailWorker` so that if it encounters this issue it flags the user not to get future Rollup emails to avoid this coming up for that user again
- Note that this tweak re-raises so that the general rescue defined at the top here is still used
## WHY
- This error is new to our new email provider and triggers when we try to send an email to an address that Postmark has flagged as undeliverable.  We don't want Sidekiq jobs that encounter this error to end up in infinite retries
- We do want to know if this is happening so we can take appropriate actions in special case actions for specific emails
- We know that retries won't be successful, and there's no point in trying to send this user another rollup on the next day or week or whatever they're scheduled for
- We want a small special case handler, but want the general `rescue_from` to run in case it ever ends up doing more than just running `ErrorNotifier.report`
## HOW
- Add a `rescue_from` directive to the `ApplicationMailer` that all other mailers inherit from
- Rescue the error in the worker, and if it is found, set `TeacherInfo.notification_frequency` to 'never' before re-raising

- Manually set an email address (`thomas+inactive@quill.org`) to Inactive in the Postmark staging environment
- Trigger a coteacher invite email via worker to that email address
- Confirm that with current code it goes into a Sidekiq retry loop
- Deploy new code with `rescue_from`
- Re-send invitation email
- Confirm that
- - No email is received
- - Sidekiq does not error and retry
- - Sentry gets a notification of the error in the staging environment
- Manually remove the email address from the Postmark inactive list, resend invitation, and confirm that it is delivered

PR Checklist | Your Answer
------------ | -------------


dandrabik (approved):
Nice catch on this!

happythenewsad (approved):

brendanshean (approved):
Nice solution here.

PR #13396, opened 2025-10-24
fix bug where spelling errors were getting incorrectly added to following attempts

emilia-friedberg:
## WHAT
Fix bug where spelling error concept results were incorrectly getting added to later attempts.

## WHY
We were seeing spelling error concept results associated with optimal responses, which shouldn't be the case, and results in teachers seeing inaccurate data.

## HOW
It turned out to be the case that our grading logic was mutating the `responses` array passed in, so if a student had a spelling error but matched an existing response otherwise, that response was then being passed into the next round of grading logic with the spelling error still attached. We never want to mutate the responses array from the grading logic, so I just added some logic to avoid that. This function was effectively duplicated across matchers, even though the error was only reported for Connect sentence combining questions, so I extracted into a shared file so all grader files will have this applied.

### Screenshots

### Notion Card Links
https://www.notion.so/quill/Teacher-facing-grading-display-error-for-Question-6-and-7-of-Quill-Connect-Open-Sentence-Combining--296d42e6f94180a3b03af3ee33a09495

Used the same responses as example student reports where we had this bug and confirmed that we no longer see the issue.

PR Checklist | Your Answer
------------ | -------------


anathomical (approved):
Good catch on this.  I also like your approach of using a shared function that ensures data is copied, not modifiable.

PR #13398, opened 2025-10-24
fix clever account overwrite path

emilia-friedberg:
## WHAT
Prevent the following scenario:

- a user is logged into Quill through Clever
- they log out of their account on Clever, and someone else logs in, on the same device
- that user accesses Quill through Clever
- the logged-in account's auth information is overwritten with the new user's data

## WHY
This is not a thing we want to have happen. At all.

## HOW
Add a check to see if there is a current user with a clever id, and if there is and the attempted auth is from a user with a different Clever id, flash an error.

### Screenshots
Screenshot 2025-10-24 at 4 07 17 PM

### Notion Card Links
https://www.notion.so/quill/Teacher-losing-Clever-synced-classroom-and-assignment-data-28ed42e6f94180bbb440e3f432cf5745?v=77f294a5b0df466e98633244de7ebfef&source=copy_link

Tested on staging that the above flow works as expected, and also that new and existing Clever accounts can still log in to Quill from Clever if there isn't already a different user logged in.

PR Checklist | Your Answer
------------ | -------------


brendanshean (approved):
Nice work on figuring this one out!

happythenewsad (commented):
I'm not yet convinced this is the best approach. I believe the issue is on [this line](https://github.com/empirical-org/Empirical-Core/blob/develop/services/QuillLMS/app/controllers/auth/clever_controller.rb#L13), where current_user is not being properly clobbered with the user data coming from the clever callback. If this is the case, we should address with a mixin, after_action, or some other shared method, so that all SSO strategies avoid existing user pollution during sign-in attempts. 

Put another way: we need a single place where all SSO strategies clobber any existing `current_user` upon success.

Also: isn't it a better UX to allow User B to clobber User A's session on a shared machine, rather than throw an error? If we do that approach, we don't ever need to care about matching current_user.id and incoming_user.id. The matching logic is different across each SSO strategy, so that's a lot of unnecessary code.

services/QuillLMS/spec/controllers/auth/clever_controller_spec.rb
```diff
@@ -68,4 +68,86 @@ def setup_conflicting_user_response(user)
...
+            id: 'matching_clever_id',
+            email: user.email,
+            user_type: 'teacher'
+          }
+        })
+        allow(CleverIntegration::TeacherIntegration).to receive(:run).and_return({ type: 'user_success', data: user })
+
+        get :clever
+        # Should not redirect due to clever_id mismatch (clever_ids match)
+        expect(flash[:error]).not_to eq(I18n.t('clever.clever_id_mismatch'))
```
happythenewsad: Cleaner to expect that no error should be flashed at all

emilia-friedberg:
> I'm not yet convinced this is the best approach. I believe the issue is on [this line](https://github.com/empirical-org/Empirical-Core/blob/develop/services/QuillLMS/app/controllers/auth/clever_controller.rb#L13), where current_user is not being properly clobbered with the user data coming from the clever callback. If this is the case, we should address with a mixin, after_action, or some other shared method, so that all SSO strategies avoid existing user pollution during sign-in attempts.
> 
> Put another way: we need a single place where all SSO strategies clobber any existing `current_user` upon success.
> 
> Also: isn't it a better UX to allow User B to clobber User A's session on a shared machine, rather than throw an error? If we do that approach, we don't ever need to care about matching current_user.id and incoming_user.id. The matching logic is different across each SSO strategy, so that's a lot of unnecessary code.

@happythenewsad I think there are some instances where we don't want to clobber that use the same pathways -- like when an existing user links their account to an SSO via the My Account page, which also just goes through the #clever method touched here. I'm hesitant to make broader changes without thinking them through, but I want to get this fix out because it is a significant vulnerability. What do you think about me just changing the logic here for now so that we log out the current user if the clever ids don't match, allowing the rest of the sign-in to proceed as expected (because I agree with you that that's better UX), and adding this as an agenda item to the next eng meeting so we can discuss it more thoroughly?

anathomical (approved):
I think I'm in agreement with both @happythenewsad and @emilia-friedberg here: there's probably a cleaner solution to this that's more generalized for multiple providers so that this bug doesn't pop up in other contexts, but also that we should plug this bug in the short term while we figure out that broader solution.

PR #13400, opened 2025-10-27
add button loading spinner to lessons select/resume edition button

emilia-friedberg:
## WHAT
Disable the button and add a loading spinner to it when a user clicks the "Select/Resume" button on the Select an Edition view in Lessons.

## WHY
This button currently gives no indicator that it's been clicked, and allows users to submit an action (which seems to be expensive, given that it's where we often see slowdowns in Lessons) multiple times. This fix has better UX and will also hopefully decrease some load.

## HOW
Add a state variable to track the button state and use it to determine conditional rendering.

### Screenshots

### Notion Card Links
https://www.notion.so/quill/Quill-Lessons-Start-Add-Button-Spinner-and-disable-299d42e6f9418033807cf81f429ff53d

Tested locally that the spinner shows up when the button's been clicked.

PR Checklist | Your Answer
------------ | -------------
Have you added and/or updated tests? | NO. - tested locally


anathomical (approved):
At first I was confused as to why we were never seeming to unset `isLoading`, but then I remembered that when this process completes the user gets redirected.

eadams17 (approved):

PR #13406, opened 2025-10-27
Add unique jobs to user pack sequence jobs

brendanshean:
## WHAT
* Add sidekiq enterprise configuration
* Add unique jobs for `SaveUserPackSequenceItemsWorker` and `SaveUserPackSequenceItemWorker`

## WHY
* This is needed to utilize the `unique_for` configuration option for sidekiq
* Both of these jobs get provisioned with duplicates.

## HOW
* Update the gemfile
* Add `unique_for: 2.minutes` to both jobs to prevent duplicates within a two minute window.  We could probably go lower here given that we want critical to have a latency of under 1.minute but we'll start with 2 here for now.  The [docs](https://github.com/sidekiq/sidekiq/wiki/Ent-Unique-Jobs#use) suggest not going higher than a few minutes.

### Screenshots

### Notion Card Links

I have tested this configuration on staging both with the two jobs and then other jobs that do not have the configuration.  Those without can have duplicates provisioned whereas the two aforementioned jobs show:

`2 TID-8iy INFO: Skipping enqueue for SaveUserPackSequenceItemsWorker, not unique`

PR Checklist | Your Answer
------------ | -------------
Have you added and/or updated tests? |  No.  Manual tests.  This particular feature is turned off for testing due to docs recommendation.


anathomical (approved):
Seems like a straightforward update to me.  As long as we do enough testing in staging to be confident that there aren't any weird gotchas, I think this should be good.

services/QuillLMS/Gemfile.lock
```diff
@@ -227,6 +231,8 @@ GEM
       railties (>= 3.2)
     drb (2.2.3)
     dry-cli (1.1.0)
+    einhorn (1.0.1)
```
dandrabik, line 234: finkle

dandrabik (commented):
Looks good to me. Interested to see the impact!

services/QuillLMS/Gemfile.lock
brendanshean: 🐬

happythenewsad (approved):

PR #13409, opened 2025-10-28
move-skills-report-pre-cache-to-default

anathomical:
## WHAT
- Turn Skills Report pre-caching workers back on
- Move the jobs from the `critical` queue to the `default` queue
## WHY
- We still want these to pre-cache when there are available resources to do the work
- These jobs are not so critical that if they don't run the user's experience fully breaks since the report pages can just make uncached queries, which take a few seconds longer
## HOW
- Revert the changes that disabled these jobs
- Update the queue these jobs are configured to be in

- On staging
- Clear cache
- Load a premium teacher's dashboard
- Confirm that pre-caching happens
- Confirm that the job runs on the default queue
- - Delete all queues in staging Sidekiq
- - Confirm that the job creates the default queue and not another one (when a job is added to a queue that "doesn't exist" in Sidekiq, Sidekiq creates the queue)

PR Checklist | Your Answer
------------ | -------------


dandrabik (approved):
:+1: Let's just monitor when this goes out.

happythenewsad (approved):

PR #13411, opened 2025-10-29
Skills report seed data

brendanshean:
## WHAT
Add a thor script that 
* exports teacher and associated data to a yaml file
* imports the data in that same yaml file into new records

## WHY
* The information was created on staging and will be lost when update staging db with a copy of production.
* It would be useful to be able to import that saved data back into staging after the update

## HOW
* iterate through the various tables and store the data 
* go through the yaml file and create necessary records, additional an id mapping for the new records and use that mapping for the foreign keys

### Screenshots

### Notion Card Links

I had Peter Sharkey confirmed on staging that the export / import worked for the two users in question.

PR Checklist | Your Answer
------------ | -------------


services/QuillLMS/lib/tasks/teacher_seed.thor
```diff
@@ -0,0 +1,382 @@
...
+      )
+      puts "Importing data from #{input_path}...\n"
+
+      data = data.transform_keys(&:to_s)
+
+      puts 'Data loaded:'
+      data.each do |key, records|
+        puts "  #{key}: #{records&.count || 0} records"
+      end
+      puts
```
anathomical, line 275: Is this just to output an empty line?

services/QuillLMS/lib/tasks/teacher_seed.thor
```diff
@@ -0,0 +1,382 @@
...
+      return puts 'Warning: No teacher found in user data' unless teacher
+
+      old_email = teacher[:email] || teacher['email']
+      puts "Overriding teacher email: #{old_email} -> #{options[:teacher_email]}"
+      update_teacher_email(teacher)
+    end
+
+    def find_teacher_in_data(data)
+      users = data['users']
+      users&.find { |u| (u[:role] || u['role']) == 'teacher' }
```
anathomical: This is stylistic, but I find `u.symbolize_keys[:role]` is clearer than `(u[:role] || u['role'])`.

services/QuillLMS/lib/tasks/teacher_seed.thor
```diff
@@ -0,0 +1,382 @@
...
+      puts
+
+      data
+    end
+
+    def override_teacher_email(data)
+      teacher = find_teacher_in_data(data)
+      return puts 'Warning: No teacher found in user data' unless teacher
+
+      old_email = teacher[:email] || teacher['email']
```
anathomical: I find `teacher.symbolize_keys[:email]` more readable in cases like this.  Also, this pattern (and the `if` block in `update_teacher_email` feels weird.  Is there not a way to standardize whether our hash is using strings or symbols for keys given that the data is being generated in this script from a YAML file?

services/QuillLMS/lib/tasks/teacher_seed.thor
```diff
@@ -0,0 +1,382 @@
...
+
+require_relative '../../config/environment'
+
+class TeacherSeed < Thor
+  desc 'export EMAILS', 'Export teacher classroom data for teachers with given emails (comma-separated)'
+  option :output, type: :string, default: 'db/seeds/teacher_seed_data.yml', desc: 'Output file path'
+  def export(emails)
+    return unless check_environment
+
+    email_list = emails.split(',').map(&:strip)
```
anathomical, line 11: It appears the the export logic allows for multiple emails to be exported at once.  On the flip side, the import logic allows for teacher email to be overridden.  However, if I'm reading it right, overriding the email when importing a file with multiple teachers in it, you'll get the unexpected behavior of overriding whichever teacher is arbitrarily first in the file.  I'd recommend either disallowing multiple-teacher export, disallowing email overrides on import, or adding a check on import that confirms that if email override is specified there is only one teacher in the dataset.

anathomical (approved):
I made a couple of small style notes, but I think that this could be merged as-is given its scope of use.

services/QuillLMS/lib/tasks/teacher_seed.thor
brendanshean: Oh, good catch.  I've only been using it for one email in testing.

services/QuillLMS/lib/tasks/teacher_seed.thor
brendanshean: Correct.

PR #13412, opened 2025-10-29
Student dashboard/ active & inactive classes

eadams17:
## WHAT
add two different sections: active and inactive classrooms on the student dashboard (rather than just one big classes section)

## WHY
we wanted to make it less confusing for students to know which are active classrooms

## HOW
use two different queries when returning active and inactive classrooms, update UI and add caching logic where relevant

### Screenshots

### Notion Card Links
https://www.notion.so/quill/Student-active-inactive-classes-feature-292d42e6f941805ebdd2f02053879b4e?source=copy_link

tested on staging with various users to ensure that active/inactive classrooms are sorted as expected

PR Checklist | Your Answer
------------ | -------------


services/QuillLMS/client/app/bundles/Student/components/selectAClassroom.tsx
```diff
@@ -39,10 +43,38 @@ const SelectAClassroom: React.FC<SelectAClassroomProps> = ({
...
+    setShowInactiveClassrooms(!showInactiveClassrooms);
+  };
+
+  const renderClassroomCard = (classroom: Classroom) => {
+    const handleClick = () => {
+      onClickCard(classroom.id);
+    };
+    return (
+      <button
+        aria-label=""
```
emilia-friedberg: Oof -- I see that this was already in the code, but we should remove this.

services/QuillLMS/app/models/activity_session.rb
```diff
@@ -587,4 +588,8 @@ def save_user_pack_sequence_items
   private def record_teacher_activity_feed
     teachers.each { |teacher| TeacherActivityFeed.add(teacher.id, id) }
   end
+
+  private def invalidate_student_classrooms_cache
+    Rails.cache.delete("student_classrooms:#{user_id}") if user_id
+  end
```
emilia-friedberg: Should this actually be invalidating the cache for every student associated with the classroom that this activity session is associated with?

services/QuillLMS/app/models/classroom_unit.rb
```diff
@@ -169,4 +170,11 @@ def save_user_pack_sequence_items
       StudentLearningSequences::HandleAssignmentWorker.perform_in(StudentLearningSequences::LEARNING_SEQUENCES_ENQUEUE_DELAY, id, student_id)
     end
   end
+
+  private def invalidate_student_classrooms_cache
+    # Invalidate cache for all students assigned to this classroom unit
+    assigned_student_ids.each do |student_id|
+      Rails.cache.delete("student_classrooms:#{student_id}")
+    end
```
emilia-friedberg: Like in activity session -- I think this should actually be every student in the class, not just ones that have been assigned?

emilia-friedberg (commented):
I'm curious about how much these queries are actually slowing down the load for this page. It might be worth messing with the queries themselves to see if we can optimize performance so that we don't need the caching layer at all. For instance, it might be faster to gather all the classrooms, along with the dates of last assignment and last session completion, and then just group them either on the front- or backend by activity status, rather than running two separate queries. I also think that would make it easier to rewrite these as active record queries, which might be preferable for maintenance. If the queries are really slow, this approach makes sense to me, but it just seems like we're going to end up invalidating the caches a lot anyway so maybe it would be better to just not have to maintain that logic.

eadams17:
> I'm curious about how much these queries are actually slowing down the load for this page. It might be worth messing with the queries themselves to see if we can optimize performance so that we don't need the caching layer at all. For instance, it might be faster to gather all the classrooms, along with the dates of last assignment and last session completion, and then just group them either on the front- or backend by activity status, rather than running two separate queries. I also think that would make it easier to rewrite these as active record queries, which might be preferable for maintenance. If the queries are really slow, this approach makes sense to me, but it just seems like we're going to end up invalidating the caches a lot anyway so maybe it would be better to just not have to maintain that logic.

I think you're right-- the caching is overkill. I added it because of [this note](https://www.notion.so/quill/Student-active-inactive-classes-feature-292d42e6f941805ebdd2f02053879b4e?source=copy_link#292d42e6f941806c83e5c97d4cba47d3) in the spec, but when I checked a few users with high classroom and activity counts, it didn't actually take long on initial load. I removed the caching logic and updated the queries to use ActiveRecord

emilia-friedberg (approved):
LGTM! Thanks for doing that rewrite -- I think this'll be easier to maintain.

services/QuillLMS/app/controllers/profiles_controller.rb
```diff
@@ -167,6 +170,106 @@ def staff
...
+      .where(classroom_units: { classroom_id: student_classroom_ids, visible: true })
+      .where(activity_sessions: { visible: true, state: 'finished' })
+      .where.not(activity_sessions: { completed_at: nil })
+      .group('classroom_units.classroom_id')
+      .select('classroom_units.classroom_id, MAX(activity_sessions.completed_at) as last_completion_date')
+
+    # Main query - get student_classroom join date in subquery for proper ordering
+    join_date_subquery = StudentsClassrooms
+      .where(student_id: current_user.id)
+      .select('classroom_id, created_at as join_date')
```
emilia-friedberg, line 218: One additional note is that I think these are shared across methods, so you could break them out into their own functions and just call them in both -- you could probably just ask Claude to consolidate shared code here (there's more in the main query itself that's identical). Then we don't need to update code in two separate places if these queries require changes later.

emilia-friedberg (approved):

anathomical (approved):
I'm a little concerned about the complexity of the multiple queries being chained together, but nothing here seems like a huge issue.  I took a stab at a version of the query using `HAVING` clauses that is more readable, but running with a couple of test users shows that the version of this that I have is not 1-to-1 in results.  I don't think that it's worth spending a ton of time trying to massage my updated query to get exactly what we want, but I'm going to leave it here for posterity in case anyone ever comes back to refactor this code.

```ruby
def base_query
  Classroom
    .select('classrooms.id, classrooms.name, users.name AS teacher, students_classrooms.created_at AS join_date')
    .joins(classroom_units: :activity_sessions)
    .joins(classrooms_teachers: :user)
    .joins(:students_classrooms)
    .where(classroom_units: { visible: true, activity_sessions: { visible: true } })
    .where(classrooms_teachers: { role: 'owner' })
    .where(students_classrooms: { student_id: current_user.id })
    .group('classrooms.id, classrooms.name, users.name, students_classrooms.created_at')
    .order('students_classrooms.created_at DESC')
end

def active_students_classrooms
  having_clause = <<-SQL
    MAX(classrooms.created_at) > ? OR
    MAX(classroom_units.created_at) > ? OR
    MAX(activity_sessions.completed_at) > ?
  SQL

  base_query.having(having_clause, one_year_ago, six_months_ago, six_months_ago)
end

def inactive_students_classrooms
  having_clause = <<-SQL
    MAX(classroom_units.created_at) IS NULL OR
    (MAX(classroom_units.created_at) <= ? AND
    (MAX(activity_sessions.completed_at) IS NULL OR MAX(activity_sessions.completed_at) <= ?))
  SQL

  base_query.having(having_clause, six_months_ago, six_months_ago)
end
```

PR #13420, opened 2025-11-04
skills-report-caching-performance-enhancement

anathomical:
## WHAT
Use a smaller set of skills when pre-caching By Student data
## WHY
We were originally pulling all historical data for all skills in the report, but we no longer cache all of those, so we no longer need to pull all the data during pre-caching.  So we can speed up the worker by reducing data we're processing.
## HOW
Update the way we determine which skills to pull data for before doing cache processing

- Deploy to staging
- Use a rails console to manually run the caching jobs and determine their run times, confirm that they are lower than existing production run times

PR Checklist | Your Answer
------------ | -------------
Have you added and/or updated tests? | No, this is primarily a configuration change


dandrabik (approved):
:+1: Looks good to me. Good idea to start with a smaller set for the initial page load.

PR #13425, opened 2025-11-05
response component consolidation

emilia-friedberg:
## WHAT
Consolidate the following components:

- ResponseRouteWrapper
- ResponseList
- Response
- ResponseComponent (renamed to ResponsesManager to better reflect its function)

## WHY
These are the components that render that big response table on all question pages in Connect/Diagnostic/Grammar. It's a big chunk of identical code, and we'll be using it in the new FITB work in Grammar.

## HOW
Consolidate from components with fewest dependencies to most. It made the most sense to keep wrapper components here since we have to pass so many props through, but I tried to clean up as much shared code as possible. I also deleted some code that wasn't actually in use (showing child and linked responses) in both the components and the reducers/action files.

### Screenshots

### Notion Card Links

Tested responses page functionality in each of the affected apps.

PR Checklist | Your Answer
------------ | -------------


services/QuillLMS/client/app/bundles/Connect/components/questions/responsesManager.jsx
```diff
@@ -0,0 +1,39 @@
+import * as React from 'react'
```
happythenewsad, line 1: Are you planning to consolidate the three responseManagers at a later date?

happythenewsad (approved):
LGTM

services/QuillLMS/client/app/bundles/Connect/components/questions/responsesManager.jsx
emilia-friedberg: In theory, but I think we'd want to do that after moving actions files into a Shared bundle, and we're not there yet. Since the ResponseManagers are each used in several files in their respective bundles, we'd need to couple a lot of state to the parent components in order to just pass this stuff in as props there, and that feels like an antipattern to me.

anathomical (approved):
I feel like I've said this on a lot of your recent PRs, but it makes me smile every time we delete unused old code.

eadams17 (approved):

PR #13426, opened 2025-11-06
skills-report-pre-caching-queue-config

anathomical:
## WHAT
Allow pre-cache worker queues to be changed via ENV
## WHY
So that we can run tests putting this back in the `critical` queue while still allowing us to drop it into `default` without having to do a new deploy
## HOW
Use `ENV.fetch` for determining the queue to run these workers on while ensuring that the default value is the lowest risk option

- Deploy to staging
- Ensure that pre-caching jobs run as intended
- Use the new queue pause feature in Sidekiq to ensure that config changes change the queue workers end up on without letting them get processed instantly

PR Checklist | Your Answer
------------ | -------------
Have you added and/or updated tests? | No, it's a config change


dandrabik (approved):
Thanks for putting this in place. This will make it easier to test and to roll back to a different queue if there are issues.

PR #13428, opened 2025-11-06
fishtank-partner-integration

anathomical:
## WHAT
- Add a migration to create a `partner_activity_session` table
- Add a new `PartnerActivitySession` model to link anonymous `ActivitySession` records with specific Partners
- Add a new `PartnerController` that partners can use to send a student to start an activity with an activity_session UID of their choosing
- - Partners will use a URL in the form of `/partner/<PARTNER_NAME>/activity_session/<ACTIVITY_ID>/<UID_FOR_SESSION>`
- Add a worker that can process designated Partner callbacks for `ActivitySession` completions
- Hook into the `ActivitySession` completion controller so that it can trigger partner callbacks
## WHY
- We want to store records linked to `ActivitySessions` when a partner asks us to so that we can report back to the partners when appropriate
- Right now this only supports the Fishtank partnership, but Peter G has asked that we keep it relatively generic to support other potential partners in the future
- We want to validate the incoming request from partners, and then redirect the student directly into playing the activity
- We don't want to block on our existing controllers for an external callback, so do it through Sidekiq
- We want to enqueue these callbacks only for Partner activity sessions

### Notion Card Links
https://www.notion.so/quill/Data-Sharing-for-Partners-Passing-Data-to-Partner-Platform-Via-Webhook-26cd42e6f941806cbeecf281ea9704ba#271d42e6f941800a8d8cda6905eff716

- On staging
- Confirm 404 when using an invalid partner in the URL
- Confirm 404 when using an invalid activity id
- Confirm 404 when using an empty UID
- Confirm that when all params are valid, the page redirects to play the activity
- Confirm that if all params are valid, but the UID has already been used, the page redirects to an in-progress version of the activity
- Confirm that when an activity is completed, the `ReportPartnerSessionCompletionWorker` gets enqueued (another test that was much easier with the ability to pause Sidekiq queues)

PR Checklist | Your Answer
------------ | -------------


services/QuillLMS/app/models/partner_activity_session.rb
```diff
@@ -0,0 +1,33 @@
...
+# Table name: partner_activity_sessions
+#
+#  id                  :bigint           not null, primary key
+#  partner             :text             not null
+#  sent_to_partner     :datetime
+#  created_at          :datetime         not null
+#  updated_at          :datetime         not null
+#  activity_session_id :integer          not null
+#
+class PartnerActivitySession < ApplicationRecord
```
happythenewsad: This class name should include the terms 'partner' and 'activity session', but 'PartnerActivitySession' implies that this table is a type of ActivitySession, which it is not. This table is metadata for ActivitySession. I suggest something like 

`ActivitySessionPartnerMetadata`

services/QuillLMS/app/controllers/api/v1/activity_sessions_controller.rb
```diff
@@ -86,12 +88,19 @@ def destroy
...
   end
+
   private def handle_concept_results
     return if !@concept_results
 
     BatchSaveActivitySessionConceptResultsWorker.perform_async(@concept_results, @activity_session.id, @activity_session.uid)
   end
 
+  private def handle_partner_session_callbacks
+    return if @activity_session.partner_activity_session.nil?
```
happythenewsad: `return unless @activity_session.partner_activity_session` is cleaner

services/QuillLMS/app/controllers/api/v1/activity_sessions_controller.rb
```diff
@@ -45,6 +46,7 @@ def create
     if @activity_session.save
       handle_student_activity_sequence_activity_completion
       handle_concept_results if @concept_results
+      handle_partner_session_callbacks
```
happythenewsad: These two callbacks should be migrated to the ActiveRecord model, to keep the controller thin.

services/QuillLMS/app/controllers/partner_controller.rb
```diff
@@ -0,0 +1,41 @@
+# frozen_string_literal: true
+
+class PartnerController < ApplicationController
+  include ForceDbWriterRole
+
+  around_action :force_writer_db_role, only: [:activity_session]
+  before_action :confirm_valid_partner, only: [:activity_session]
+  before_action :validate_required_params, only: [:activity_session]
+
+  def activity_session
```
happythenewsad: - routes.rb declares this to be a GET, but it should be a POST, since it's performing a write.
- it's semantically better for a POST controller method to include a verb, since noun method names indicate a GET. Thus:  `create_partner_activity_session`

services/QuillLMS/app/controllers/partner_controller.rb
```diff
@@ -0,0 +1,41 @@
...
+      PartnerActivitySession.create!(activity_session:, partner: PartnerActivitySession::FISHTANK) unless activity_session.partner_activity_session
+
+      activity_session
+    end
+
+    play_params = {
+      student: activity_session.uid
+    }
+
+    redirect_to activity.send(:module_url_helper, play_params)
```
happythenewsad: Can avoid `send` by making this method public

happythenewsad (changes requested):
solid architecture

happythenewsad (approved):
LGTM

services/QuillLMS/app/models/activity_session_partner_metadata.rb
```diff
@@ -0,0 +1,39 @@
...
+#  updated_at          :datetime         not null
+#  activity_session_id :integer          not null
+#  external_id         :text             not null
+#
+# Indexes
+#
+#  index_activity_session_partner_metadata_on_activity_session_id  (activity_session_id) UNIQUE
+#  index_activity_session_partner_metadata_on_external_id          (external_id) UNIQUE
+#
+class ActivitySessionPartnerMetadata < ApplicationRecord
```
dandrabik: You might consider something shorter, e.g. `PartnerSession` or `PartnerActivitySession`. I'm not sure metadata is needed here, since all tables store data/metadata.

services/QuillLMS/app/controllers/partner_controller.rb
```diff
@@ -0,0 +1,36 @@
...
+    }
+
+    redirect_to activity_session.activity.module_url_helper(play_params)
+  end
+
+  private def activity_session_partner_metadata
+    ActivitySessionPartnerMetadata.create_or_find_by!(external_id:, partner:) do |partner_data|
+      partner_data.activity_session = ActivitySession.create!(activity_id:, user_id: nil)
+    end
+  end
```
dandrabik: You might consider moving this into the model like so:

```ruby
  def self.create_or_find_with_session!(external_id:, partner:, activity_id:)
    create_or_find_by!(external_id: external_id, partner: partner) do |partner_data|
      # Build, not create!, so Rails can save it as part of the same transaction.
      partner_data.build_activity_session(activity_id: activity_id, user_id: nil)
    end
  end

File: services/QuillLMS/app/models/activity_session.rb [hunk 8]
```diff
@@ -587,4 +595,8 @@ def save_user_pack_sequence_items
   private def record_teacher_activity_feed
     teachers.each { |teacher| TeacherActivityFeed.add(teacher.id, id) }
   end
+
+  private def marking_as_complete?
+    completed_at && completed_at_previously_was.nil? && saved_change_to_completed_at?
```
dandrabik: I think only the first two are needed (but I could be wrong)

services/QuillLMS/app/models/activity_session.rb
```diff
@@ -93,6 +94,7 @@ class StudentNotAssignedActivityError < StandardError; end
   before_save :set_completed_at, :set_activity_id
 
   after_save :determine_if_final_score, :update_milestones, :increment_counts
+  after_save :handle_partner_session_callbacks, if: -> { marking_as_complete? }
```
dandrabik: Take or leave, you might consider a `if: -> { marking_as_complete? && has_partner_session? }`

services/QuillLMS/app/models/activity_session.rb
```diff
@@ -537,6 +539,12 @@ def self.search_sort_sql(sort)
   end
   # rubocop:enable Metrics/CyclomaticComplexity
 
+  def handle_partner_session_callbacks
+    return unless activity_session_partner_metadata
+
+    ReportPartnerSessionCompletionWorker.perform_async(id)
```
dandrabik: Take or leave, I was surprised to see a worker with this name be passed an activity_session_id vs. activity_session_partner_metadata.id?

services/QuillLMS/app/models/partner_activity_session.rb
```diff
@@ -0,0 +1,39 @@
...
+class ActivitySessionPartnerMetadata < ApplicationRecord
+  PARTNERS = [
+    FISHTANK = 'fishtank'
+  ]
+
+  belongs_to :activity_session
+
+  validates :partner, presence: true, inclusion: { in: PARTNERS }
+
+  def send_to_partner
```
dandrabik, line 35: Should there be a check here for `sent_to_partner.nil?` here?

services/QuillLMS/app/models/activity_session_partner_metadata.rb
```diff
@@ -0,0 +1,39 @@
+# frozen_string_literal: true
+
+# == Schema Information
+#
+# Table name: activity_session_partner_metadata
+#
+#  id                  :bigint           not null, primary key
+#  partner             :text             not null
+#  sent_to_partner     :datetime
```
dandrabik: Take or leave, I wonder if this should be named something like `completion_notification_sent_at` or something else explicitly saying what was sent.

services/QuillLMS/app/models/partner_activity_session.rb
```diff
@@ -0,0 +1,39 @@
...
+    send_to_fishtank if partner == FISHTANK
+
+    update!(sent_to_partner: DateTime.current)
+  end
+
+  private def send_to_fishtank
```
dandrabik, line 43: Maybe we should make this a `raise NotImplementedError`. I get we don't know the logic, but if this runs, 'sent_to_partner' will be set.

services/QuillLMS/app/workers/report_partner_session_completion_worker.rb
```diff
@@ -0,0 +1,15 @@
+# frozen_string_literal: true
+
+class ReportPartnerSessionCompletionWorker
+  include Sidekiq::Worker
+
+  sidekiq_options queue: SidekiqQueue::DEFAULT
+
+  def perform(activity_session_id)
```
dandrabik: I'd say use the PartnerSession id as the param, but even without that change, this could be
```ruby
ActivitySessionPartnerMetadata.find_by(activity_session_id:)&.send_to_partner

File: services/QuillLMS/app/controllers/partner_controller.rb [hunk 15]
```diff
@@ -0,0 +1,36 @@
...
+
+  before_action :confirm_valid_partner, only: [:create_partner_activity_session]
+  before_action :validate_required_params, only: [:create_partner_activity_session]
+
+  def create_partner_activity_session
+    play_params = {
+      student: activity_session.uid
+    }
+
+    redirect_to activity_session.activity.module_url_helper(play_params)
```
dandrabik: Consider:
```ruby
activity_session.activity.module_url_helper(student: activity_session.uid)

File: services/QuillLMS/app/models/partner_activity_session.rb [hunk 16]
```diff
@@ -0,0 +1,39 @@
...
+  PARTNERS = [
+    FISHTANK = 'fishtank'
+  ]
+
+  belongs_to :activity_session
+
+  validates :partner, presence: true, inclusion: { in: PARTNERS }
+
+  def send_to_partner
+    send_to_fishtank if partner == FISHTANK
```
dandrabik, line 38: This is sorta early and could be a later follow-on / refactor, but it might be nice to extract this into a config, e.g. maybe each partner has a service defined.

```ruby
COMPLETION_SERVICES = {
  FISHTANK => PartnerSessions::FishtankCompletion
}

def completion_service = COMPLETION_SERVICES.fetch(partner)
def complete_session
  completion_service.run(self)
  update!(sent_to_partner: DateTime.current)
end

dandrabik (approved):
Overall looks good. I left a bunch of comments about naming and small refactors, but nothing is blocking. I think you should take another pass at naming the model. You are already using `PartnerSession` in a few places, and naming the model that (or something like it) would make the code a little more readable. The rest of the comments are more take or leave.

PR #13434, opened 2025-11-10
skills report premium level updates

emilia-friedberg:
## WHAT
- Teacher Premium teachers can no longer view All Practice or assign work from the Skills Report
- Instead, when on the All Practice page, they will see content that instructs them to reach out to an admin to upgrade
- When attempting to assign work, there will be a lock icon on the relevant buttons, which when clicked open a modal with similar content instructing them to reach out to an admin

## WHY
We want the Skills Report to help facilitate more school- and district-sales, while still offering valuable content to our Teacher Premium users. 

## HOW
Pass up `hasSchoolOrDistrictPremium` from the backend and use that (in a Context hook! first use in the codebase!) to determine what to show in different sections of the report, plus some other more minor frontend tweaks.

### Screenshots
Screenshot 2025-11-10 at 11 38 34 AM
Screenshot 2025-11-10 at 11 38 56 AM
Screenshot 2025-11-10 at 11 39 09 AM
Screenshot 2025-11-10 at 11 39 21 AM

### Notion Card Links
https://www.notion.so/quill/Skills-Report-Changes-to-Skills-Report-to-Facilitate-Sales-29bd42e6f94180c9a5dedf8bac89c0be?source=copy_link

Tested locally, Peter Sharkey and Jack are QAing on staging

PR Checklist | Your Answer
------------ | -------------


services/QuillLMS/app/models/subscription.rb
```diff
@@ -182,6 +182,10 @@ def is_trial?
     account_type && TRIAL_TYPES.include?(account_type)
   end
 
+  def is_school_or_district_premium?
```
happythenewsad: Note that a similar method is defined in User.rb, with subtly different business logic:

  ```
def school_premium?
    school&.subscription&.present?
  end

  def district_premium?
    school&.district&.subscription&.present?
  end

  def school_or_district_premium?
    school_premium? || district_premium?
  end
```

Reading the spec, I believe the existing method should be used.

File: services/QuillLMS/spec/controllers/teachers/progress_reports_controller_spec.rb [hunk 2]
```diff
@@ -287,29 +287,78 @@
...
+
+      context 'when user has a teacher subscription' do
+        let!(:subscription) { create(:subscription, account_type: Subscription::TEACHER_PAID) }
+        let!(:user_subscription) { create(:user_subscription, user: teacher, subscription:) }
+
+        it 'should assign has_school_or_district_premium as false' do
+          get :skills_report
+          expect(assigns(:has_school_or_district_premium)).to be false
+        end
+      end
```
anathomical: This is purely stylistic, so take or leave it, but you could slightly simplify this by splitting out which value is changing per test:
```ruby
context 'when the user has a subscription' do
  let(:subscription) { create(:subscription, account_type:) }
  let!(:user_subscription) { create(:user_subscription, user: teacher, subscription:) }

  context 'when it is a school subscription' do
    let(:account_type) { Subscription::SCHOOL_PAID }

    it 'should assign has_school_or_district_premium as true' do
      get :skills_report
      expect(assigns(:has_school_or_district_premium)).to be true
    end
  end

  context 'when it is a district subscription' do
    let(:account_type) { Subscription::SCHOOL_DISTRICT_PAID }

    it 'should assign has_school_or_district_premium as true' do
      get :skills_report
      expect(assigns(:has_school_or_district_premium)).to be true
    end
  end

  ... etc.
end
```
This reads slightly clearer to me because it makes it very clear which single value is being changed for each of these cases.

services/QuillLMS/app/models/subscription.rb
emilia-friedberg: Great catch, thank you!

services/QuillLMS/spec/models/subscription_spec.rb
```diff
@@ -96,6 +96,36 @@
     end
   end
 
+  describe '#is_school_or_district_premium?' do
+    let!(:subscription) { create(:subscription) }
+
+    it 'returns true if the subscription is a school subscription type' do
+      Subscription::OFFICIAL_SCHOOL_TYPES.each do |school_type|
+        subscription.update(account_type: school_type)
```
anathomical: Similar to the above comment on style, take or leave this, but I prefer `let` calls to `update` calls in tests.  This one's a bit less obviously "better" to me, so I think these could be slightly more readable as:
```ruby
  describe '#is_school_or_district_premium?' do
    let!(:subscription) { create(:subscription, account_type:) }

    context 'school subscriptions' do
      Subscription::OFFICIAL_SCHOOL_TYPES.each do |school_type|
        context "of type #{school_type}" do
          let(:account_type) { school_type } 

          it 'returns true' do
            expect(subscription.is_school_or_district_premium?).to be true
          end   
        end   
      end   
    end

    ...
  end
```

anathomical (approved):
Oh, I'm quite into this context hook thing.  It feels way cleaner than trying to pass attributes through ten layers of components.

happythenewsad (approved):

PR #13448, opened 2025-11-13
add new classroom and data to staff demo account

emilia-friedberg:
## WHAT
- Add a new classroom and set of activity pack templates (based on real data) to the demo account generation script for the quill.org/quill_staff_demo account
- Revise logic within create_demo to account for multiple classrooms, and some other optimizations like only enqueuing workers once other creation has happened (removing the need for the one minute delay, and also making sure we don't batch a bunch of workers for sessions that get rolled back at the end of the ActiveRecordTransaction, which was happening sometimes)
- Revise the session_data script slightly to pull unit_template info as well because we were ending up with duplicate sessions when the same activity was present in multiple packs

## WHY
The partnerships team wanted more robust data in this account to show off the new Skills Report.

## HOW
See "What".

Note: we were originally going to add this new classroom to the public-facing demo as well, but I expressed reservations because the code to regenerate the account when it's changed is taking 20+ minutes to fully run on staging -- it would be faster in prod with more workers, but how much faster is hard to say. I was working on speeding this up, but then PG messaged with the following:

"Hi Emilia, Shannon, and Christina - Emilia as a quick FYI, I flagged in the partnerships team channel, with you tagged, that I would not like to put the cloned demo into the public facing demo account for now. Instead, I’d prefer that we recreate that data ourselves at a later date so that it is data we originated. For the staff demo account this is okay for now, but once we have more of our own data we can switchover."

Then later:

"Thanks Emilia. Yes, so I will keep you in the loop here for sure. One bigger picture question is whether we want to lock down the demo account so that we don’t have to keep running this script. It has always not been the best solution, but if we want to add a lot more data to it, it gets difficult. I do wonder how hard it would be to simple disable all of the buttons that we don’t want people to click on. It’s a bit of a pain, but there are probably only like 100 buttons or something total, and so we could have some type of generic “Can’t edit demo account” pop up that triggers on all of these buttons, and if we had that, then the demo account wouldn’t need to regenerate. It might not actually be that much work to inventory these buttons and then create those blocks."

### Screenshots

### Notion Card Links
https://www.notion.so/quill/Duplicating-and-de-identifying-real-student-data-for-the-demo-data-account-2a8d42e6f94180e2a655e8cb1061b88b?source=copy_link

Tested on staging and having Nattalie and Shannon QA as well.

PR Checklist | Your Answer
------------ | -------------


happythenewsad (approved):

anathomical (approved):
Looks good.  Thanks for cleaning up a lot of the demo logic while you were at it with the data provisioning.

eadams17 (approved):

PR #13450, opened 2025-11-14
Add delay to save user pack sequence item

brendanshean:
## WHAT
Add a delay in the enqueueing of the `SaveUserPackSequenceItemWorker`

## WHY
Support surfaced that one of their highest-volume issues is failure of Staggered Release locks to update properly.  We suspect that race conditions are the likely culprit, so adding in a delay might help to mitigate this.

## HOW
Use `perform_in` 15 seconds for the job.

### Screenshots

### Notion Card Links

I tested this job in the Rails console on staging and confirmed that 15 jobs were sent to the scheduled queue.  Also when I reran the same `SaveUserPackSequenceItemsWorker.perform_async(classroom.id, user.id)` multiple times in a few seconds, I noticed that the retry queue did not grow (i.e. unique job was working as expected).

PR Checklist | Your Answer
------------ | -------------


services/QuillLMS/app/services/user_pack_sequence_item_saver.rb
```diff
@@ -5,6 +5,7 @@ class UserPackSequenceItemSaver < ApplicationService
 
   class UserPackSequenceItemSaverError < StandardError; end
 
+  ENQUEUE_DELAY = Rails.configuration.x.user_pack_sequence_items.enqueue_delay
```
brendanshean, line 8: New pattern that I'm using for [custom configuration](https://joyofrails.com/articles/mastering-custom-configuration-in-rails)

Rails docs [here](https://guides.rubyonrails.org/v7.1/configuring.html#custom-configuration)

happythenewsad (approved):

services/QuillLMS/app/services/user_pack_sequence_item_saver.rb
dandrabik: Interesting new pattern!

dandrabik (approved):
Thanks for pushing this through. I'm not sure how much is gained from using the custom config pattern for a single ENV var, but I could see it being helpful to put ALL of the config vars in a single file at some point, to make it easier to audit.

brendanshean:
> Thanks for pushing this through. I'm not sure how much is gained from using the custom config pattern for a single ENV var, but I could see it being helpful to put ALL of the config vars in a single file at some point, to make it easier to audit.

I was debating adding a generic file for this for future reuse, but wanted to keep this PR simple.  It would be very easy to rename this file to something more generic as we add more constants to it.  I actually have more of these in my clustering PR, so I will do that there.

PR #13454, opened 2025-11-17
optimize-skills-report-pre-cache

anathomical:
## WHAT
Only pre-cache "All History" report if teacher is school/district premium
## WHY
Only those teachers can even access that version of the report
## HOW
- Have the enqueue job check the user's premium status and pass it into the underlying cache jobs
- Have the cache jobs read the new attribute and not cache for all_history unless the user is premium

### Notion Card Links
https://www.notion.so/quill/Optimize-Skills-Report-pre-caching-workers-to-not-pre-cache-All-History-for-non-school-district-prem-2aed42e6f941803faa70c10c3b0d710c?source=copy_link

- Deploy to staging
- Log in with a teacher-premium teacher
- Confirm pre-caching takes place, but does not include all-history data

PR Checklist | Your Answer
------------ | -------------


services/QuillLMS/app/workers/skills_report/cache_api_data_worker_base.rb
```diff
@@ -10,9 +10,10 @@ class CacheApiDataWorkerBase
 
     attr_reader :user_id, :classroom_id, :data
 
-    def perform(user_id, classroom_id)
+    def perform(user_id, classroom_id, all_history)
```
brendanshean, line 13: I feel like the `all_history` name is a little surprising (it sounds like the scope rather than “whether All History is allowed”). I think something like `allow_all_history` or `precache_all_history` would be more self-documenting.

brendanshean (approved):

dandrabik (approved):
:+1: LGTM, thanks for putting this optimization in place.

services/QuillLMS/app/workers/skills_report/cache_api_data_worker_base.rb
anathomical: After a short discussion via Slack, we've decided to leave this as-is for now.

PR #13457, opened 2025-11-18
Evidence health stats part 1

happythenewsad:
## WHAT
- remove unused columns in Activity Health
- update some display texts in the table
- style the table header differently 

## WHY
- renovating the page 

### Screenshots
with patch (left), without patch (right)
Screenshot 2025-11-19 at 9 26 48 AM

### Notion Card Links
https://www.notion.so/quill/Evidence-Acivity-Health-Stats-Refresh-296d42e6f9418068ba06fa2091c9db23

- [x] cross check that table data looks correct, before and after patch

PR Checklist | Your Answer
------------ | -------------


anathomical (approved):
Given that this is an internal tool and is presumably performing well enough, this is probably good to go as-is, but I wonder if it's worth also looking at the code that generates this data for the front-end and stop pulling the numbers that we're no longer reporting.  That may be overkill for now, but I figured I'd raise it as a possibility.

eadams17 (approved):

PR #13459, opened 2025-11-19
Evidence LLM check: Update messaging and add retry on failure

dandrabik:
## WHAT
Update the LLM API test check with a more detailed message and add shorter retries when we find a potentially unhealthy LLM API.
## WHY
This job is run hourly, but if there's possible downtime, we probably want to keep track of it on a shorter timeline. Open to different ideas on how to do this.
## HOW
Queue another check 5 minutes later if we find an unhealthy LLM. I considered a `raise Error` here, but sidekiq's backoff is probably too fast for our needs.

I don't have a great way to QA this. Relying on test coverage here.

PR Checklist | Your Answer
------------ | -------------


services/QuillLMS/app/workers/check_unhealthy_llms_worker.rb
```diff
@@ -3,11 +3,20 @@
 class CheckUnhealthyLlmsWorker
   include Sidekiq::Worker
 
+  # This job is run hourly via, but can queue retries if there are failures.
+  # Adding a unique constraint to avoid multiple alarm threads for downtime over 1 hour
+  sidekiq_options queue: SidekiqQueue::DEFAULT, unique_for: 10.minutes
+
+  RETRY_TIME_MINUTES = 5
```
brendanshean: I would probably make this an environment variable to simplify tuning later.

brendanshean (approved):
I like this idea.  Also, smart use of unique job

PR #13463, opened 2025-11-19
add a migration for fill_in_the_blank attribute and a rake task to duplicate existing fitb data

emilia-friedberg:
## WHAT
Add a migration for `fill_in_the_blank` attribute for activities (will eventually be used for filtering in the Activity Library, as well as in the Grammar admin panel) and a rake task that will copy all Connect FITB activities, questions, responses, and concept feedbacks to Grammar, as well as archive the Connect versions and replace them with the Grammar versions in existing activity packs (the first part can be run anytime, but we won't want to run the second til we launch).

## WHY
We're moving all the Connect FITB content to Grammar. The migration will happen on production later, but getting it set up now makes it easier for me to test as I build out the admin infrastructure for the new tool.

## HOW
Used Claude to help me generate the script, and I ran it locally against staging data (pulling the responses from production, though updating them in staging, because we are missing a lot of responses in the staging database) and verified that the data looks as expected.

### Screenshots

### Notion Card Links

Tested locally.

PR Checklist | Your Answer
------------ | -------------


services/QuillLMS/lib/tasks/fitb_migration.rake
```diff
@@ -0,0 +1,433 @@
+# frozen_string_literal: true
+
+require 'csv'
+require 'httparty'
+
+namespace :fitb_migration do
+  desc 'Duplicate Connect FITB questions and activities to Grammar'
+  task duplicate_to_grammar: :environment do |_task, args|
+    dry_run = ENV['DRY_RUN'] == 'true'
```
happythenewsad: nit: since these flag names are so common, and could be inadvertently used to manage an unrelated feature in the future, it's safer to namespace: `FITB_DRY_RUN`

services/QuillLMS/lib/tasks/fitb_migration.rake
```diff
@@ -0,0 +1,433 @@
+# frozen_string_literal: true
+
+require 'csv'
+require 'httparty'
+
+namespace :fitb_migration do
+  desc 'Duplicate Connect FITB questions and activities to Grammar'
+  task duplicate_to_grammar: :environment do |_task, args|
```
happythenewsad, line 9: I get that we're writing in a 'script' paradigm here, but a 300-line method is a lot harder to grok and validate, vs 10 30-line methods.

services/QuillLMS/lib/tasks/fitb_migration.rake
```diff
@@ -0,0 +1,433 @@
...
+      puts ''
+
+      # Step 3: Duplicate Responses via CMS API
+      puts 'Step 3: Duplicating responses via CMS API...'
+      cms_url = ENV['QUILL_CMS'] || 'https://cms.quill.org'
+      puts "Using CMS URL: #{cms_url}"
+
+      question_mapping.each do |old_question_uid, new_question_uid|
+        # Fetch responses from CMS
+        response = HTTParty.get("https://cms.quill.org/questions/#{old_question_uid}/responses")
```
happythenewsad: Not sure how long this loop will take on staging/prod data

services/QuillLMS/lib/tasks/fitb_migration.rake
```diff
@@ -0,0 +1,433 @@
...
+          connect_q_uid = all_question_uids[i]
+          grammar_q_uid = connect_q_uid ? question_mapping[connect_q_uid] : nil
+          connect_a_uid = all_activity_uids[i]
+          grammar_a_uid = connect_a_uid ? activity_mapping[connect_a_uid] : nil
+
+          csv << [connect_q_uid, grammar_q_uid, connect_a_uid, grammar_a_uid]
+        end
+      end
+
+      puts "Mapping file saved to: #{csv_path}"
```
happythenewsad: Nice way to bake in validation of the script

services/QuillLMS/lib/tasks/fitb_migration.rake
```diff
@@ -0,0 +1,433 @@
...
+          if verbose
+            puts "  WARNING: Activity '#{activity.name}' (#{old_uid}) has #{non_fitb_questions.length} non-FITB questions that will be excluded"
+            puts "    Total questions: #{original_question_count}, FITB: #{fitb_question_count}, Non-FITB: #{non_fitb_questions.length}"
+          end
+        end
+
+        if dry_run
+          puts "  [DRY RUN] Would create Grammar activity #{new_uid} from #{old_uid}" if verbose
+        else
+          new_activity.save!
```
happythenewsad: This is going to make the script not idempotent (I think), since SecureRandom.uuid will create new UIDs for each Activity on each run

services/QuillLMS/lib/tasks/fitb_migration.rake
```diff
@@ -0,0 +1,433 @@
...
+        response = HTTParty.get("https://cms.quill.org/questions/#{old_question_uid}/responses")
+
+        if response.success?
+          responses = JSON.parse(response.body)
+          puts "  Found #{responses.count} responses for question #{old_question_uid}" if verbose
+
+          responses.each do |resp|
+            new_response_data = {
+              uid: resp['uid'],
+              question_uid: new_question_uid,
```
happythenewsad: There's a uniqueness constraint in CMS on (text, question_uid), so that's good for idempotency

services/QuillLMS/lib/tasks/fitb_migration.rake
emilia-friedberg: @happythenewsad it's slow -- like an hour. I'm not sure how much that matters for a one-off task -- do you think it would be better to move this into a worker?

services/QuillLMS/lib/tasks/fitb_migration.rake
emilia-friedberg: Yeah, good call, I'm updating this to just do `"fitb-#{activity.uid}"`, which will also make it easier to track down the original if we need to.

services/QuillLMS/lib/tasks/fitb_migration.rake
happythenewsad: Presumably you got the hour estimate from staging? If so, I think this is fine; I mentioned it because I wanted to avoid waiting 6 hours, then having the entire transaction fail due a single issue....

emilia-friedberg: @happythenewsad yeah, from staging, but copying from prod data (since we have many fewer responses in staging) so I'm optimistic it'll be fairly accurate.

happythenewsad:
Thanks for doing the method chunking refactor! I'll review again at EOD, once I finish some things off

happythenewsad (approved):

eadams17 (approved):

PR #13466, opened 2025-11-20
restore data to Quill Classroom demo classroom

emilia-friedberg:
## WHAT
When updating the demo script for the new Quill Skills classroom, I inadvertently made it so that we no longer had data in place for some of the activities from the Quill Classroom classroom. This fixes that by adding a fallback if there was no unit template associated with the original session, which is the case for some of our demo data, and rebuilding the ymls on that basis.

Sidenote: there is an additional issue where some of the data that should be in the Angie ELL account is missing (in production, so there's nothing to copy to the YML files). I'm trying to track that down now, but it appears to have been missing since at least this summer. The id associated with her ELL sessions appears to belong to a real student, not one we created to base demo data off of, so it may have something to do with that.

Also, make some changes to the test suite, because it was taking 5+ minutes to run with all the new data. We sacrifice some value in the tests themselves, but I don't think it's worth having them take that long to run.

## WHY
We want the Quill Classroom to still have all the associated data.

## HOW
See "What"

### Screenshots

### Notion Card Links
Shannon just messaged me about this this morning.

Tested the demo generation script locally and confirmed that all the data we should see is there for both classrooms.

PR Checklist | Your Answer
------------ | -------------


services/QuillLMS/app/services/demo/report_demo_creator.rb
```diff
@@ -1831,7 +1831,7 @@ def self.clone_activity_sessions(student_id, classroom_unit_id, clone_user_id, c
     matching_sessions = session_data.activity_sessions.filter do |session|
       session.activity_id == clone_activity_id &&
         session.user_id == clone_user_id &&
-        session_data.activity_session_unit_templates[session.id] == unit_template_id
+        (session_data.activity_session_unit_templates[session.id] ? session_data.activity_session_unit_templates[session.id] == unit_template_id : true)
```
brendanshean, line 1835: I think this line is kind of hard to parse.  You could put it as a guard clause at the top of the loop instead for readability.

brendanshean (approved):

services/QuillLMS/app/services/demo/report_demo_creator.rb
brendanshean: But I'm fine with looser standards in the demo script.

emilia-friedberg: I'll just add a comment explaining what it does!

services/QuillLMS/spec/services/demo/report_demo_creator_spec.rb
```diff
@@ -38,6 +36,11 @@
 
   context 'full data set' do
     before do
+      # Stub Concept lookup instead of creating thousands of records
+      allow(Concept).to receive(:find_by_id) do |concept_id|
+        Concept.unscoped.find_by(id: concept_id) || double('Concept', uid: "concept-#{concept_id}")
+      end
```
anathomical, line 42: This fallback logic is pretty nice.

anathomical (approved):

PR #13473, opened 2025-11-21
grammar fitb: general infrastructure, concepts, concept feedbacks pages

emilia-friedberg:
## WHAT
Start to build out the infrastructure for having FITB questions in Grammar -- updates to navbar, facelifts and clean up of dead and duplicated code for Concepts and ConceptsFeedbacks so we can use those for the new questions.

## WHY
This is part one of the admin updates we need to make to Grammar to support the new tooling! It's not ready for use yet, but the diff was getting big.

## HOW
Make sure we're passing all the data we need from the backend, then start setting up the new pages we'll need for Grammar FITB, starting with Concept and ConceptFeedbacks because they were mostly just getting a facelift and some minor new functionality from the old versions.

### Screenshots
Screenshot 2025-11-21 at 2 35 07 PM
Screenshot 2025-11-21 at 2 35 14 PM
Screenshot 2025-11-21 at 2 35 21 PM

### Notion Card Links

Tested Concept and ConceptFeedback pages locally on each app where they're used. Other changes will be tested as the functionality is finished.

PR Checklist | Your Answer
------------ | -------------
Have you deployed to Staging? | Will deploy when the rest of the functionality is in place


eadams17 (approved):
Looks great! Love the cosmetic updates too

services/QuillLMS/spec/models/activity_spec.rb
```diff
@@ -96,6 +96,25 @@
...
+      activity.fill_in_the_blank = true
+      activity.save!
+      expect(activity.reload.fill_in_the_blank).to be true
+    end
+
+    it 'should be settable to false' do
+      activity.fill_in_the_blank = false
+      activity.save!
+      expect(activity.reload.fill_in_the_blank).to be false
+    end
```
anathomical, line 113: I think that this test probably stands just fine as-is, but I do want to point out a weird thing going on here: if `fill_in_the_blank` is already `false`, this particular test doesn't actually confirm that this command changes things.  I think this is still probably good enough to cover most of what we want, but I wanted to flag just in case.

anathomical (approved):
Thanks for splitting this out before it got bigger.  I think you've been doing a really good job of keeping your PR sizes big enough to be impactful without being so big that they're overwhelming to review.

PR #13476, opened 2025-11-21
Fix badge for expired school subscription

brendanshean:
## WHAT
Fix badges for expired subscription

## WHY
Users with expired district and school subscriptions are showing a 'teacher expired' badge instead

## HOW
Check for last expired subscription subscription type and use that text.

### Screenshots

### Notion Card Links
https://www.notion.so/quill/Teachers-administrators-see-a-Teacher-Premium-expired-tag-when-their-School-Premium-or-District-Prem-2b2d42e6f941809aa923c8d555da58b9?source=copy_link

I have checked for both users in the support ticket have the correct badges on staging.

PR Checklist | Your Answer
------------ | -------------


services/QuillLMS/app/helpers/navigation_helper.rb
```diff
@@ -438,7 +438,15 @@ def show_social_studies_dashboard_tab?(current_user)
...
   private def locked_badge(current_user)
-    badge_html('TEACHER PREMIUM', 'yellow') + badge_counter("#{current_user.last_expired_subscription&.is_trial? ? 'trial ' : ''}expired")
+    last_subscription = current_user.last_expired_subscription
+    badge_text = if last_subscription&.account_type == Subscription::SCHOOL_DISTRICT_PAID
+                   'DISTRICT PREMIUM'
+                 elsif last_subscription&.account_type == Subscription::SCHOOL_PAID
+                   'SCHOOL PREMIUM'
+                 else
+                   'TEACHER PREMIUM'
+                 end
```
anathomical: This is fine as-is, so no need to make a change here unless you want to, but from a style perspective, as soon as I'm looking at more than two logic branches, I start thinking about using hashes for this kind of thing:
```ruby
BADGE_TEXTS = {
  Subscription::SCHOOL_DISTRICT_PAID => 'DISTRICT PREMIUM',
  Subscription::SCHOOL_PAID => 'SCHOOL PREMIUM'
}.freeze
badge_text = BADGE_TEXTS.fetch(last_subscription&.account_type, 'TEACHER PREMIUM')
```

anathomical (approved):
Small note on style, but it's purely preference, so this could go out as-is.  This does generally seem like the correct way to handle this: checking the account_type of the expired subscription.

services/QuillLMS/app/helpers/navigation_helper.rb
brendanshean: I find my branching conditional very ugly, so I will change to this.

PR #13477, opened 2025-11-21
Evidence: Use Structured Output for Newer Models

dandrabik:
## WHAT
We want to use Structured Output for the newer models from OpenAI and Gemini
## WHY
The rate of mis-formed responses when we don't use the structured output for these models is too high. Also, weirdly, using structured output for our main LLM (google-gemini-2.0) lowers it's accuracy for some weird reason, so we want to scope the use of it to the newer models.
## HOW
Pass in `schema` for the feedback output
Deployed to my staging and run multiple trials.

PR Checklist | Your Answer
------------ | -------------


brendanshean (approved):
Thanks for adding this schema stuff in

PR #13479, opened 2025-11-25
anonymous-partner-landing-tweaks

anathomical:
## WHAT
- Add the ability to force teacher preview mode on an activity playthrough by adding `teacher_preview=true` to the URL params
- Update `create_partner_activity_session` action to allow for JSON calls for unlinked partners
- Add Fishtank configuration for Anonymous Partner Landing page
- Move Fishtank from `LINKED_PARTNERS` to `UNLINKED_PARTNERS` until we build a full integration with them (which may be six months out)
- Update the Student "Start Activity" link to be a button that triggers code to set up a `PartnerActivitySession` for UNLINKED partners
- Change `PartnerActivitySession` to allow for null `external_id` so that we can use this table to flag sessions when they belong to an UNLINKED partner
## WHY
- We want teachers to be able to view a preview of the activity from the Anonymous Partner Landing page, but they often won't be logged in as a teacher which would prevent them from seeing the preview otherwise
- The original version of this action was designed to be called by partners to redirect students to linked sessions, but now we want to direct our own users there via AJAX call, so we need to get the URL back to set `window.location.href` instead of triggering a 302
- We want to allow Fishtank to temporarily use the UNLINKED workflow since we haven't gotten specifications from them about building a fully LINKED workflow
- Peter G asked that we flag anonymous sessions triggered by UNLINKED partners for reporting purposes
- In order to flag anonymous sessions for UNLINKED partners, we need to be able to record records with no `external_id`
## HOW
- Add a new utility to check for `teacher_preview=true` in the URL, and update all activity play code to enable teacher preview if it is true
- Fork the behavior of the `create_partner_activity_session` so that it allows for AJAX usage when the Partner is UNLINKED and uses the old logic when the Partner is LINKED
- Move the Fishtank partner listing in `PartnerActivitySession` from `LINKED_PARTNERS` to `UNLINKED_PARTNERS`
- Replace the `<a>` tag with a `<button>` that uses a newly-defined function to figure out where to redirect the student to
- Add a migration

### Notion Card Links
https://www.notion.so/quill/Add-Fishtank-as-Anonymous-Partner-for-this-year-2b2d42e6f94180f8904bdf768f0e6cb1

- Test teacher preview URL param for Grammar, Connect, Diagnostic, and Evidence when not logged in
- Launch a student activity with an UNLINKED partner and confirm a `PartnerActivitySession` gets created
- - Confirm that the `partner` value of the `PartnerActivitySession` is set correctly
- Launch a student activity with an unknown partner and confirm that no `PartnerActivitySession` gets created
- Confirm that using an anonymous partner link for Fishtank shows the Fishtank logo in the header

PR Checklist | Your Answer
------------ | -------------


services/QuillLMS/app/controllers/activity_sessions_controller.rb
```diff
@@ -98,7 +98,9 @@ def activity_session_from_classroom_unit_and_activity
     if @activity.classification.key == 'lessons'
       "#{ENV['DEFAULT_URL']}/preview_lesson/#{@activity.uid}"
     else
-      @activity.anonymous_module_url.to_s
+      url = @activity.anonymous_module_url.to_s
+      url += '&teacher_preview=true' if params.permit(:teacher_preview).fetch(:teacher_preview, false)
```
emilia-friedberg, line 102: Didn't know about the second argument for `.fetch`! Cool.

services/QuillLMS/client/app/bundles/Shared/libs/hasTeacherPreview.ts
```diff
@@ -0,0 +1,11 @@
...
+
+/**
+ * Checks if the teacher_preview parameter is present in the current URL
+ * @param url - Optional URL to check (defaults to window.location.href)
+ * @returns true if teacher_preview parameter exists, false otherwise
+ */
+export function hasTeacherPreview(url?: string): boolean {
+  const value = getParameterByName('teacher_preview', url, null);
+  return value !== null;
+}
```
emilia-friedberg, line 11: I don't necessarily think this is worth changing, since I can't really imagine a use case where we'd explicitly want to pass a `false` argument, but it does seem a little odd to me that we just check for the existence of any non-null value rather than an explicitly truthy value.

emilia-friedberg (approved):

PR #13480, opened 2025-11-26
TEKS/ Individual grade pages

eadams17:
## WHAT
add the individual grade pages for TEKS

the 3rd and final PR will add logic for report routing and Premium pricing page updates

## WHY
another component of the larger TEKS project

## HOW
React/HTML/css

### Screenshots

### Notion Card Links
- https://www.notion.so/quill/3rd-Grade-TEKS-Activities-29dd42e6f94180ffac2acd6d9d492414
- https://www.notion.so/quill/4th-Grade-TEKS-Activities-2a9d42e6f94180bdbf3ae22eb90113ed
- http://notion.so/quill/5th-Grade-TEKS-Activities-2a9d42e6f94180a491bac6cc2b15c30d

tested on staging with Jamie/Jack to check that all things look and function as expected

PR Checklist | Your Answer
------------ | -------------


emilia-friedberg (approved):
LGTM! I kind of wonder if it's worth consolidating the views/controller actions/main container files to just pass a grade param rather than having three separate ones, since it seems like we might be building a bunch of others in the future? But that's something we could do as a follow-up task if necessary.

services/QuillLMS/app/controllers/pages_controller.rb
```diff
@@ -234,6 +234,25 @@ def research_and_efficacy
...
+    set_school_admins_and_subscription_type
+  end
+
+  def quill_for_texas_fourth_grade
+    set_school_admins_and_subscription_type
+  end
+
+  def quill_for_texas_fifth_grade
+    set_school_admins_and_subscription_type
+  end
```
anathomical, line 248: This isn't a huge deal, so take or leave this suggestion: I think this is a little cleaner if you use a `before_action` here:
```ruby
before_action :set_school_admins_and_subscription_type, only: [:quill_for_texas_third_grade, :quill_for_texas_fourth_grade, :quill_for_texas_fifth_grade]
```
This is a little weird since it then means that these actions don't have any code since they're just here to load templates, but the convention I've seen in Rails is to just have a comment in them.  Something like:
```ruby
def quill_for_texas_third_grade
  # Display the associated view
end
```

anathomical (approved):
One small comment you can take or leave, but this looks good to me.

services/QuillLMS/app/controllers/pages_controller.rb
eadams17: thanks for the suggestion! I'll update

PR #13481, opened 2025-11-26
fill in the blank activities index and activity page

emilia-friedberg:
## WHAT
Build out the Fill in the Blank Activities index and Activity page, using Shared components (and making more in some instances) where possible.

## WHY
This is the next step of the Grammar Fill in the Blank project!

## HOW
We want to make this flow more like Connect's, so I consolidated the existing lessonForm code in Connect and Diagnostic for reuse here, and built out the new Fill in the Blank Activities index to use it, as well as the actions/reducer/backend support to actually create new activities and FITB questions.

Note: I originally had the reducer here named `fillInBlankQuestions`, which I think is clearer, but since some of the reused components expect that key to be just `fillInBlank` -- because that's what it is in Connect and Diagnostic -- I decided to just follow the existing standard. I'd love to clean that up at some point so they all use `fillInBlankQuestions`, but that seemed like a big undertaking in the middle of feature work.

### Screenshots
Screenshot 2025-11-26 at 3 46 57 PM
Screenshot 2025-11-26 at 3 47 03 PM

### Notion Card Links
https://www.notion.so/quill/Fill-in-the-Blank-spec-v2-257d42e6f9418011931aeafaa8a96d92?d=2b6d42e6f941805990d3001cd133e875#2afd42e6f94180c8b8d0ec9c7d16e049

Tested flows on new pages and double-checked lessonForm functionality on Connect and Diagnostic to make sure I didn't break it.

PR Checklist | Your Answer
------------ | -------------
Have you deployed to Staging? | NO - will deploy to staging when whole project is ready for review


services/QuillLMS/app/controllers/api/v1/lessons_controller.rb
```diff
@@ -23,13 +23,15 @@ def show
   def create
     uid = SecureRandom.uuid
     name = valid_params[:name] || valid_params[:title]
-    @lesson = Activity.create!(uid:, classification: @classification, data: valid_params, name:, flag: valid_params[:flag])
+    fill_in_the_blank = valid_params[:fill_in_the_blank] || false
```
anathomical, line 26: This is purely stylistic, and I don't know if it's actually better or clearer than what you have here, but you could also do this with `fetch` with a default argument: `valid_params.fetch(:fill_in_the_blank, false)`

services/QuillLMS/client/app/bundles/Grammar/actions/__tests__/fillInBlankQuestions.test.ts
```diff
@@ -0,0 +1,191 @@
...
+
+  describe('submitNewQuestion', () => {
+    const mockQuestion = {
+      prompt: 'Test question',
+      conceptID: 'concept-123',
+      flag: 'production'
+    };
+    const activityID = 'activity-456';
+
+    it('should dispatch success actions when question creation and addition succeed', async () => {
```
anathomical, line 43: I think having test coverage of dispatch chains is really useful.  I'm glad these exist.

services/QuillLMS/client/app/bundles/Grammar/components/fillInBlankActivities/fillInBlankActivity.tsx
```diff
@@ -0,0 +1,197 @@
...
+
+  const renderQuestionsAsList = (questionsForLesson: Question[]) => {
+    return questionsForLesson.map((question: Question, index: number) => {
+      const { cues, key, prompt } = question;
+      return (
+        <li className="question-box" key={key}>
+          <Link to={`/admin/fill-in-the-blanks/${key}/responses`}>
+            <p>
+              <span>{index + 1}.&nbsp;</span>
+              <span dangerouslySetInnerHTML={{ __html: prompt }} />
```
eadams17: I think we could just import 'html-react-parser' and do `{parse(prompt)}` here rather than use dangerouslySetInnerHTML

services/QuillLMS/client/app/bundles/Shared/components/lessons/lessonForm.tsx
```diff
@@ -0,0 +1,279 @@
...
+    }
+  };
+
+  const renderSearchBox = () => {
+    const { questionType } = formState;
+    const dataSource = { questions, fillInBlank, titleCards };
+    let options = hashToCollection(dataSource[questionType].data);
+    const conceptsData = concepts.data[0];
+    if (options.length > 0) {
+      if (questionType !== 'titleCards') {
```
eadams17: should we just combine this to `if(options.length > 0 && questionType !== 'titleCards')`?

eadams17 (approved):
LGTM! Just two small comments

anathomical (approved):

services/QuillLMS/client/app/bundles/Shared/components/lessons/lessonForm.tsx
emilia-friedberg: The `QuestionSearch` gets rendered even if `titleCards` is the questionType, the difference is some parsing that's contained within that included block, but you flagging this made me realize we can just do an early return here to avoid the confusion!

services/QuillLMS/app/controllers/api/v1/lessons_controller.rb
emilia-friedberg: Yeah, I kind of softly prefer what I have here, just because otherwise someone unfamiliar (me, before last week) has to go see what fetch's second param is to see what that's doing, but I'd be happy to change it if we want to.

PR #13487, opened 2025-12-03
Evidence: Update Repeated Feedback checker to Gemini 2.5

dandrabik:
## WHAT
Update the RepeatedFeedback Checker to Gemini 2.5.

Also, I started adding README.md files for how to update these APIs and run the baseline tests.
## WHY
So it is no longer on Gemini 2.0 which deprecates soon.
## HOW
Ran baseline tests, updated the prompt until I got comparable results. 

The results percentage is 0.4 below Gemini 2.0, but the breakdown is actually better: more real repeats are caught, and a few more non-repeats will be given secondary feedback (which isn't the worst thing).

### Screenshots
Screenshot 2025-12-03 at 12 54 44 PM

After with Gemini 2.5
Screenshot 2025-12-03 at 11 46 17 AM

### Notion Card Links
https://www.notion.so/quill/Repeated-Feedback-to-Gemini-2-5-2bed42e6f9418068ad7bcc4be35acfb5

Deployed to staging and tested that Secondary feedback still works

PR Checklist | Your Answer
------------ | -------------


brendanshean (approved):
Thank you for updating this and great idea on the READMEs, which will save headaches down the road.

PR #13497, opened 2025-12-08
grammar fitb pt 4: blank by blank grading logic addition, duplication script updates, generally getting everything working

emilia-friedberg:
## WHAT
- Updates to duplication script to include TranslationMappings, ActivityCategories, Topics, and ContentPartners (this made it much slower, so I broke the scripts into two parts -- one for questions and then, when that's done, for activities -- and also moved the response duplication into jobs)
- Build out Blank by Blank grading page and update logic for fill in blank questions to take blank by blank grading into account
- Build out new version of Responses manager with filtering by blank
- Other updates to make sure concept results get saved properly

## WHY
These are the final pieces we need to get Grammar Fill in the Blank activities working with both the old grading system and the new one!

## HOW
Little bit of everything.

### Screenshots
Screenshot 2025-12-08 at 2 45 22 PM
Screenshot 2025-12-08 at 2 45 26 PM
Screenshot 2025-12-08 at 2 46 09 PM

### Notion Card Links

Testing on staging, having Nattalie and Rachel QA as well.

PR Checklist | Your Answer
------------ | -------------


services/QuillLMS/client/app/bundles/Grammar/helpers/conceptResultsGenerator.ts
```diff
@@ -18,11 +18,14 @@ export function getConceptResultsForQuestion(question: Question): FormattedConce
   }
 }
 
-function getConceptResultsForAttempt(attempt: ResponseAttempt, question: Question, prompt: string, index: Number):FormattedConceptResult[] {
-  const answer = attempt.text;
+function getConceptResultsForAttempt(attempt: ResponseAttempt, question: Question, prompt: string, index: number):FormattedConceptResult[] {
+  // Handle both direct response format (regular Grammar) and wrapped format (fillInBlank)
```
emilia-friedberg, line 22: This is really not ideal to have to do, but existing Grammar attempts work differently than other apps in that they are stored without the wrapping `{ response: {...} }`, and changing that would mean potentially breaking active activity sessions.

services/QuillLMS/app/workers/fitb_response_duplication_worker.rb
```diff
@@ -0,0 +1,76 @@
...
+  rescue => e
+    Rails.logger.error("Exception while processing responses for #{old_question_uid}: #{e.message}")
+    raise
+  end
+
+  private def create_duplicated_response(resp, new_question_uid, cms_url)
+    translation_mapping_for_response = TranslationMapping.find_by(field_name: "cms_responses.#{resp['id']}")
+
+    new_response_data = {
+      uid: resp['uid'],
```
anathomical: I don't think this is necessarily wrong, but this implies to me that we'll end up with multiple CMS `Response` records with the same `uid` value.  In this case it's because one's a duplicate of the other, but that still feels a little weird to me.

Unrelated, but this might also be cleaner if it's explicit about what's being copied/replaced:
```ruby
new_response_data = resp.merge({
  'question_uid' => new_question_uid,
  'count' => resp['count'] || 1,
  'first_attempt_count' => resp['first_attempt_count'] || 0
})
```

services/QuillLMS/client/app/bundles/Grammar/helpers/conceptResultsGenerator.ts
anathomical: That feels like one of those "we should fix that at some point" things that we'll probably never get around to.  I think your solution here is the right approach.

services/QuillLMS/client/app/bundles/Grammar/libs/filterResponsesByBlank.ts
```diff
@@ -0,0 +1,141 @@
...
+ * @param searchValues - Array of search values (one per blank, empty string = any cue)
+ * @returns Filtered array of responses matching the search criteria
+ *
+ * @example
+ * filterByBlankCues(responses, question, ['a', 'good'])
+ * // Returns only responses where blank1='a' AND blank2='good'
+ *
+ * filterByBlankCues(responses, question, ['a', ''])
+ * // Returns only responses where blank1='a' (blank2 can be anything)
+ */
```
anathomical, line 97: Just want to flag that these big explanations of the functions are pretty useful for these highly specialized definitions.  I'm glad you put/kept them in.

anathomical (approved):

services/QuillLMS/client/app/bundles/Grammar/helpers/conceptResultsGenerator.ts
emilia-friedberg: Yeah. Thinking about this more, I think it makes sense to add a follow-up card that wraps regular Grammar responses in `{response: ...}` for this state, which would in theory eventually let us get rid of the code that expects it not to be wrapped (maybe next summer or something like that we could decide that we don't need to keep around ActiveActivitySessions for Grammar activities prior to when that change goes out).

eadams17: agreed-- this seems like the best path forward

eadams17 (approved):
LGTM!

PR #13499, opened 2025-12-08
cms update to allow a perPage param in responses search

emilia-friedberg:
## WHAT
Update the CMS to accept a `perPage` param in the responses search.

## WHY
FITB questions require cue-specific filtering on properties that the CMS has no model for, since it doesn't store question-level data. I played around with different ways to handle that on the backend, but it would require passing a lot of additional data through the controller action and still not actually play nicely with ElasticSearch unless we added new columns to the table, which I'd like to avoid since this table is so large and so important. Instead, the simplest option was just to have the CMS return a larger set of responses (currently set on the frontend to 1000, but most FITB questions have far fewer) and handle pagination on the frontend, which just required adding a `perPage` param.

## HOW
See "What".

### Screenshots

### Notion Card Links

Tested on staging.

PR Checklist | Your Answer
------------ | -------------


services/QuillCMS/app/controllers/responses_controller.rb
```diff
@@ -2,6 +2,7 @@
 require 'modules/response_aggregator'
 require 'modules/incorrect_sequence_calculator'
 
+# rubocop:disable Metrics/ClassLength
```
happythenewsad: I vote to  remove this rubocop, rather than add the disable directive. There are legitimate reasons for long classes.

anathomical (approved):

services/QuillCMS/lib/modules/response_search.rb
```diff
@@ -8,12 +8,13 @@ def search_responses(question_uid, query_filters)
       query: query_values
     }
     results = Response.__elasticsearch__.search(**search_payload)
-    page_number = query_filters[:pageNumber]
+    page_number = query_filters[:pageNumber] || 1
+    per_page = query_filters[:perPage] || 20
```
happythenewsad: Was 20 the previous default? How about using a constant called ELASTIC_PER_PAGE_DEFAULT or something

services/QuillCMS/spec/controllers/lib/response_search_spec.rb
```diff
@@ -53,4 +56,93 @@
       expect(ResponseSearch.build_query_string('000', test_query_filters, false)).to eq(expected_result)
     end
   end
+
+  describe '#search_responses' do
+    # rubocop:disable RSpec/VerifiedDoubles
+    let(:mock_page) { double('page', total_pages: 5, size: 20, records: []) }
```
happythenewsad, line 62: suggestion: add a 'minimize mocking' directive to your Rspec LLM prompt. This is the one I use:

`When writing rspec or jest tests, focus on minimalist tests that test very specific scenarios. Avoid mocking or stubbing, unless it's necessary (for API calls, for example).`

happythenewsad (approved):

PR #13511, opened 2025-12-11
Teacher premium pricing update (merge on Jan. 5)

brendanshean:
## WHAT
* Create new plan record for Teacher Premium pricing for $115/year
* Adjust UI copy and backend plan/Stripe mapping to use this new plan as the default.
* Keep existing users on the $80/year plan grandfathered in.

## WHY
* The new plan record ties into the price object in stripe's backend

## HOW
* Add a “Teacher Paid Legacy” plan name to the model and Stripe price ID mapping. Also, adjust renewal Stripe price selection to use plan’s `stripe_price_id` for teacher paid / legacy accounts. 
* Migrate old “Teacher Paid” ($80) records to legacy name, create a new “Teacher Paid” ($115) plan, and update Teacher Premium page show $115 per teacher per year.

For environment variables:
`STRIPE_TEACHER_PLAN_PRICE_ID` value gets assigned to `STRIPE_TEACHER_PLAN_LEGACY_PRICE_ID`
and `STRIPE_TEACHER_PLAN_PRICE_ID` get assigned the $115 price id from stripe.

### Screenshots

### Notion Card Links
https://www.notion.so/quill/Update-Teacher-Premium-pricing-page-and-update-Stripe-logic-for-Teacher-Premium-subscription-transac-2b1d42e6f9418012be58e3668e3f6ce5?source=copy_link

I tested a few users on staging:
1) a new user (displayed $115 fee for subscription)
2) an existing user with an active subscription (this displayed $80 / fee would be charged at renewal date)
3) an existing users with an expired subscription (this correctly showed $115 when trying to renew)

PR Checklist | Your Answer
------------ | -------------


services/QuillLMS/db/migrate/20251211142000_update_teacher_premium_plan_pricing.rb
```diff
@@ -0,0 +1,37 @@
...
+      .where(name: NAME, display_name: DISPLAY_NAME, price: LEGACY_PRICE, audience: AUDIENCE, interval: INTERVAL)
+      .update_all(name: LEGACY_NAME)
+
+    Plan.create!(
+      name: NAME,
+      display_name: DISPLAY_NAME,
+      price: NEW_PRICE,
+      audience: AUDIENCE,
+      interval: INTERVAL,
+      interval_count: INTERVAL_COUNT
```
anathomical: I suspect this doesn't matter for our purposes, but I figured I'd flag: there is a theoretical data state (one we shouldn't be in, hence it not mattering), but the `where` and `create!` don't directly overlap (`create` has `interval_count` as an argument), so you could, in theory, fail to update the legacy value and create the new one.  In practice, I don't think this will ever be an issue for us.

anathomical (approved):
This seems fine as-is.  I do wonder if we want to be using a migration instead of a rake or thor task of some sort to update the `Plan` records.  You could imagine someone running the migrations from a fresh install, and since we didn't create the original `Plan` records in a migration, will break the chain of migrations.

brendanshean:
> This seems fine as-is. I do wonder if we want to be using a migration instead of a rake or thor task of some sort to update the `Plan` records. You could imagine someone running the migrations from a fresh install, and since we didn't create the original `Plan` records in a migration, will break the chain of migrations.

Thanks for the callout.  I debated making this a rake task, but opted for a migration since this is kind of a configuration record that is loosely coupled to the UI. However, you reminded me that we didn't use a migration for original plan records but instead a [rake task](lib/tasks/temporary/plans.rake), so I will make it one.

services/QuillLMS/db/structure.sql
```diff
@@ -145,7 +145,7 @@ CREATE FUNCTION public.timespent_question(act_sess integer, question character v
           item timestamp;
```
dandrabik, line 145: Lots of structure changes unrelated to PR. Worth removing.

services/QuillLMS/lib/tasks/teacher_premium_pricing.thor
```diff
@@ -0,0 +1,31 @@
+# frozen_string_literal: true
+
+require_relative '../../config/environment'
+
+class TeacherPremiumPricing < Thor
+  INTERVAL = 'yearly'
+  AUDIENCE = 'teacher'
+  DISPLAY_NAME = 'Teacher Premium'
```
dandrabik, line 8: Probably better to use the constants from the models.

dandrabik (approved):
Looks good, thanks for setting this all up! Some small cleanup comments, but overall :+1:

PR #13524, opened 2025-12-16
Add upload page for dataset recovery json

brendanshean:
## WHAT
The dataset editor is a very complex page that keeps track of lots of state while attempting to label minor and major versions as a result of specific changes.  There are potentially some edge cases left to be uncovered.  When these do happen, the user sees a 
image

At this point, I attempt to remedy the bug based on the JSON structure up and until the point.  

Once the bug is resolved, though, the original snapshot has not been persisted outside of the JSON snapshot.  This PR adds a mechanism to upload that file for saving into a new dataset.

## WHY
A user might have lots of pending edits when attempting a save and without something like this has to manually add them all back in.

## HOW
1. User navigates to `/cms/evidence#/dataset-recovery`
2. User uploads a dataset-recovery-*.json file
3. System validates the file structure and extracts datasetId from `operationContext.url`
4. System stores failedPayload.changes in sessionStorage
5. System redirects to the dataset editor URL from the recovery file
6. Dataset editor detects pending recovery data, applies it as local changes, then clears sessionStorage

### Screenshots
image

### Notion Card Links
https://www.notion.so/quill/Dataset-Editor-Failed-to-Save-2cad42e6f94180038e4ffbb5ad28b4ba?source=copy_link

PR Checklist | Your Answer
------------ | -------------


services/QuillLMS/client/app/bundles/Staff/components/evidence/datasetRecovery/DatasetRecoveryUpload.tsx
```diff
@@ -0,0 +1,224 @@
...
+            <p className="dataset-id">
+              Target Dataset ID: <strong>{parsedData.validation.datasetId}</strong>
+            </p>
+
+            <h4>Changes to apply:</h4>
+            {renderSummary()}
+
+            <div className="actions">
+              <button
+                className="quill-button-archived contained primary medium"
```
emilia-friedberg, line 203: NBD but ideally we wouldn't be adding new instances of `quill-button-archived` (this should be roughly equivalent to `quill-button contained medium green focus-on-light`.

emilia-friedberg (approved):
This seems like a really useful piece of functionality!

dandrabik (approved):
This looks good code wise and seems reasonable to ship. 

This is outside the scope of this, but I am a little worried about the overall product pattern (e.g. might encourage making a lot of edits without saving, it's a seldom used path we'll have to support, doesn't solve the real problem, etc.) and wonder if there are other product solutions would mitigate the issue.

This is the issue stated: `"A user might have lots of pending edits when attempting a save and without something like this has to manually add them all back in."`

So shrinking "might have lots of pending edits" could help make the issue small enough so a backup option isn't needed. Maybe encouraging or forcing saves more often is a path worth investigating, e.g. auto-saving at some point, blinking the save button when there are multiple edits, adding a FAB save button when there are edits, etc.

brendanshean:
> This looks good code wise and seems reasonable to ship.
> 
> This is outside the scope of this, but I am a little worried about the overall product pattern (e.g. might encourage making a lot of edits without saving, it's a seldom used path we'll have to support, doesn't solve the real problem, etc.) and wonder if there are other product solutions would mitigate the issue.
> 
> This is the issue stated: `"A user might have lots of pending edits when attempting a save and without something like this has to manually add them all back in."`
> 
> So shrinking "might have lots of pending edits" could help make the issue small enough so a backup option isn't needed. Maybe encouraging or forcing saves more often is a path worth investigating, e.g. auto-saving at some point, blinking the save button when there are multiple edits, adding a FAB save button when there are edits, etc.

Just to add full context here.  This tool is meant to be a developer only one.  Workflow is a follows:
1) User reports bug and attaches dataset-recovery.json
2) Developer fixes bug using dataset-recovery.json as guide
3) After fix is deployed, developer uploads the dataset-recovery.json tool to attempt save.

PR #13527, opened 2025-12-17
fix-daily-cron-sidekiq-enqueues

anathomical:
## WHAT
- Use `as_json` on date args to meet strict_args requirements
- Add new spec cases that run the enqueue methods without any mocks while swallowing the actual enqueue calls
## WHY
- `strict_args` does not allow passing raw `Time` objects into workers.  This was causing `interval_1_day` to crash when trying to enqueue `DailyStatsEmailJob` and short-circuiting enqueueing the rest of the jobs defined to run daily.  Because `Cron` is run through the Heroku Scheduler Add-on which we have no instrumentation on, so when it crashes in this way we get no error reporting anywhere.
- ensure that no syntax or argument errors occur here since we don't have any actual instrumentation that would surface these in production
## HOW
- Make sure to use `as_json` to convert the `Time` to a valid JSON type when enqueuing the jobs so that it passes the `strict_args` setting and the jobs all enqueue successfully without crashing.
- Use `Sidekiq::Testing.fake!` to eat the queue, then run each Cron method and confirm it doesn't error

### Notion Card Links
https://www.notion.so/quill/Vitally-API-Sync-Has-Not-Run-Overnight-Since-12-11-25-2cad42e6f9418048bca2f2024833ad65?d=578d42e6f94183e4b3bd83cf0bc1bb8f

- Deploy to staging
- Turn off all workers
- Run `heroku run rake cron:interval_1_day`
- Confirm that the `heroku run` completes without raising an error
- Confirm that all the expected jobs are enqueued to Sidekiq on staging

PR Checklist | Your Answer
------------ | -------------
Have you added and/or updated tests? | Yes, added test cases to make sure that the methods run without crashing


services/QuillLMS/app/models/cron.rb
```diff
@@ -38,7 +38,7 @@ def self.interval_1_day
     date = Time.current.getlocal('-05:00').yesterday
 
     # internal analytics and security
-    DailyStatsEmailJob.perform_async(date) if production?
+    DailyStatsEmailJob.perform_async(date.as_json) if production?
```
brendanshean, line 41: Should we update the method signature:    `def perform(date)`

services/QuillLMS/app/models/cron.rb
```diff
@@ -60,7 +60,7 @@ def self.interval_1_day
 
     # caching
     MaterializedViewRefreshWorker.perform_async
-    RematchUpdatedQuestionsWorker.perform_async(date.beginning_of_day, date.end_of_day)
+    RematchUpdatedQuestionsWorker.perform_async(date.beginning_of_day.as_json, date.end_of_day.as_json)
```
brendanshean, line 63: Should we update the method signature:    `def perform(start_time = 1.day.ago, end_time = Time.current, delay = DELAY_PER_QUESTION)`

happythenewsad (approved):
Nice fix. I also think BS's suggestions are good.

brendanshean (approved):
Thank you for figuring this out

PR #13548, opened 2026-01-08
Attempts to speed up ActiveActivitySessions#update

dandrabik:
## WHAT
Trying a few optimizations on our most used endpoint.
## WHY
Any gain on a high traffic endpoint, is a big scalability win
## HOW
1. Remove NR user lookup (it should be noted that EVERY api call does a user lookup for NR, could probably get rid of more of these, although it's only like 4-5ms, so for most endpoints, it's trivial).
2. Remove some old API header editing from all API calls
3. Move route up in route.rb file (unclear if this helps, conflicting reports online, but improved speed on staging)
4. Consolidate OmniAuth calls

Tested this on staging and it seems to work fine. I also tested logging in with Google (OmniAuth) and that works as expected.

PR Checklist | Your Answer
------------ | -------------
Have you added and/or updated tests? |  Pure refactor, so no test changes needed


services/QuillLMS/config/database.yml
```diff
@@ -30,11 +30,11 @@ development:
   primary:
     <<: *local_shared
     database: emp_gr_development
-    url: <%= ENV['DATABASE_URL']  || 'postgres://127.0.0.1:5432/emp_gr_development' %>
+    url: <%= ENV['DEVELOPMENT_DATABASE_URL']  || 'postgres://127.0.0.1:5432/emp_gr_development' %>
```
dandrabik, line 33: Note, I'm changing this var name since I think it's a little more foolproof to not use the magic ENV var `DATABASE_URL`. I was reminded of that on my new laptop that lost my `.env.test` and that I almost ran tests on the staging DB.

If anyone has a better suggestion, I'm all ears, but I think this is a little cleaner.

happythenewsad (approved):

PR #13559, opened 2026-01-13
Wrap data table chips tightly

eadams17:
## WHAT
get data table chips to wrap tightly

## WHY
this was a requested design update

## HOW
tweak css

### Screenshots
Screenshot 2026-01-15 at 3 47 11 PM

### Notion Card Links
https://www.notion.so/quill/Figure-out-how-to-get-data-table-chips-to-wrap-tightly-25ed42e6f9418025b121d30a31cc5686?source=copy_link

tested on staging to check that `DataTableChip` displays as expected in all places where it's used

PR Checklist | Your Answer
------------ | -------------


emilia-friedberg (approved):
I think the idea behind this card was to figure out how to get rid of, for example in your screenshot, the extra space on the right after the "&" in "Modifying Phrases & Classes" or after "Confused" in "Commonly Confused Words" -- ideally, we could find a way for the chip to only extend as far as it needs to even when the text is wrapped (like it does for single-line chips). If this solution works for Jack, it works for me, but I just wanted to double check.

eadams17:
> I think the idea behind this card was to figure out how to get rid of, for example in your screenshot, the extra space on the right after the "&" in "Modifying Phrases & Classes" or after "Confused" in "Commonly Confused Words" -- ideally, we could find a way for the chip to only extend as far as it needs to even when the text is wrapped (like it does for single-line chips). If this solution works for Jack, it works for me, but I just wanted to double check.

Sorry @emilia-friedberg , I added the wrong screenshot-- just updated the description with the correct screenshot. It looks less awkward with this screenshot because it removes the extra padding for that cell that was causing it to add even more open space. I chatted with Jack and he's good with this change as he feels it still looks better than the truncated option. 

There is a way to get them to wrap tightly (you probably experimented with this, using `width: min-content`), but it causes the chips to break words too early without using the space available (see below screenshot) which Jack wasn't a fan of either.
 
Screenshot 2026-01-12 at 6 40 17 PM

I tried messing with this for a few hours, but I don't think there's a way that we can get exactly what Jack wants without maybe adding like custom max-widths on a cell by cell basis?

That said, Jack gave this approval and said to go ahead and ship it.

PR #13569, opened 2026-01-20
DO NOT MERGE - For temporary reference only

anathomical:
This is not intended for merging in its current state, but should serve as a reference as to the hacky stuff I did to get the prototype up and running on the front-end.


anathomical:
We no longer need this for reference

PR #13572, opened 2026-01-20
new-standards-report-prototype-back-end

anathomical:
## WHAT
- Create underlying models to support State Standards in the database
- Create a thor task to take a CSV as an input for TEKS standards to seed the database with that standard
- Create a service to fill in "implicit" State Standard links to grades that don't have an explicit definition in the CSV (this is to handle cases where, say, a 3rd grade and a 6th grade Standard both link to the same concept: for a given grade only one of those standards is considered "relevant")
- Use the `SkillsReports` services as a baseline to create new `StandardsReports` services.  Most of these are either direct copies of slightly modified with the exception of the `QuestionResponses*` services which are significantly overhauled to handle new queries
- Create API controller to provided access to the services from the front-end 

NOTE: There is a known performance issue for teachers with large datasets requesting data on the "By Standard" view with "All History" and "All Skills" selected.
## WHY
- We need these records to manage new standards
- Curriculum has already done the work of calculating which state standards correspond with which Level 1 Concepts in the Quill database, a thor script is faster to build and use than building a full-on management tool
- The spec requires each individual grade to have a specific link to a standard for it to be available, but there are some grades in some standards that don't have explicit standards (no 12th grade standards in TEKS, for instance), so we need a way to figure out which standard is "most appropriate" for a grade
- We want queries to work similar to the way they do in the SkillsReport, so no need to reinvent the wheel
- We want to access this data from the frontend
## HOW
See what and why

PR Checklist | Your Answer
------------ | -------------


services/QuillLMS/app/models/state_standard_framework.rb
```diff
@@ -0,0 +1,26 @@
...
+# Table name: state_standard_frameworks
+#
+#  id           :bigint           not null, primary key
+#  abbreviation :text             not null
+#  name         :text             not null
+#  status       :text             not null
+#  created_at   :datetime         not null
+#  updated_at   :datetime         not null
+#
+class StateStandardFramework < ApplicationRecord
```
brendanshean: I think `StateStandardsFramework` or `StandardsFramework` would be more appropriate for the since the framework applies to a collection of State Standards as opposed to one.  Contrast this with `StateStandardGrade` which is a record that belongs_to one State Standard.

services/QuillLMS/app/models/state_standard_concept.rb
```diff
@@ -0,0 +1,36 @@
...
+#  state_standard_id           :integer          not null
+#
+# Indexes
+#
+#  idx_on_state_standard_framework_id_grade_concept_id_53b6c94066  (state_standard_framework_id,grade,concept_id) UNIQUE
+#
+class StateStandardConcept < ApplicationRecord
+  belongs_to :concept
+  belongs_to :state_standard
+  belongs_to :state_standard_framework
```
brendanshean: Unless I'm missing something, it feels redundant to have `state_standard_framework_id` as foreign key on this model since we can derive from state_standard.

anathomical: Agreed that this is redundant, and this may be worth changing, but I went for an intentional denormalization here in order to reduce the complexity of some of the queries (avoiding an admittedly relatively small, <100k row table, join).

services/QuillLMS/app/models/classroom.rb
```diff
@@ -28,7 +28,8 @@ class Classroom < ApplicationRecord
 
   GRADES = %w[1 2 3 4 5 6 7 8 9 10 11 12 University]
   UNIVERSITY = 'University'
-  GRADE_INTEGERS = { Kindergarten: 0, University: 13, PostGraduate: 14 }
+  GRADE_INTEGERS = { PreKindergarten: 0, Kindergarten: 0, University: 13, PostGraduate: 14, Other: 6 }
```
emilia-friedberg, line 34: Other being `6` seems weird to me, since that's an existing grade -- is there a rationale for that?

services/QuillLMS/app/services/standards_report/question_response_scorer.rb
```diff
@@ -0,0 +1,52 @@
...
+    def run
+      return fitb_scoring if use_fitb_scoring?
+
+      default_scoring
+    end
+
+    private def attempts = response.attempts
+    private def cue_count = question.data['cues'].length
+    private def fewer_attempts_than_cues? = attempts.present? && attempts < cue_count
+    private def fill_in_the_blank? = response.question_type == 'connect_fill_in_blanks'
```
emilia-friedberg: We'll want to update this to `[Question::TYPE_CONNECT_FILL_IN_BLANKS, Question::TYPE_GRAMMAR_FILL_IN_BLANKS]`, I think.

emilia-friedberg: Which, actually, is making me realize we need to do the same thing for the skills report. I'll make a card to do that.

emilia-friedberg (approved):
A couple of minor things I think we might want to look at, but overall LGTM! Thanks so much for getting this together.

services/QuillLMS/app/models/classroom.rb
anathomical: This is a good question, and I should probably put a comment in the code to make this clear: "Other" is `6` because when the spec calls for "Other" being treated as the "middle" of the grade range when trying to calculate the "nearest" grade which is used when defining implicit `StateStandardConcept` records.  It's possible that we actually want to define this in a more separated map of records, but at the very least we want an explanation where this is defined.

services/QuillLMS/app/services/standards_report/question_response_scorer.rb
anathomical: Good call-out.  Probably worth updating the Skills Report value in a separate PR just to get it into production.  I'll look at that this afternoon.

services/QuillLMS/app/services/standards_report/classroom_scores_by_user.rb
```diff
@@ -0,0 +1,29 @@
...
+    def run
+      ComposeScoresByUser.run(question_responses:, user_hash:, questions_hash:)
+    end
+
+    private def questions_hash = @questions_hash ||= Question.where(uid: question_uids).group_by(&:uid)
+    private def question_responses = @question_responses ||= QuestionResponsesForClassroom.run(classroom_id:, state_standard_id:, state_standard_framework_id:, all_history:)
+    private def question_uids = question_responses.map(&:question_uid).uniq
+
+    private def user_hash
+      @user_hash ||= User.joins(:students_classrooms)
```
brendanshean, line 23: Does this need a student_classrooms: { visible: true } condition to match that in QuestionResponseBase?  Otherwise there might be a mismatch in counts.

services/QuillLMS/app/services/standards_report/question_responses_for_classroom_by_standard.rb
```diff
@@ -0,0 +1,47 @@
...
+
+    Result = Data.define(:user_id, :activity_session_id, :question_number, :question_uid, :attempts, :optimal, :question_type, :activity_session_completed_at, :timespent, :activity_id, :classroom_id, :state_standard_id, :state_standard_code)
+
+    def initialize(classroom_id:, state_standard_framework_id:, concept_id: nil, user_id: nil, all_history: false)
+      super(classroom_id:, concept_id:, state_standard_framework_id:, all_history:)
+      @user_id = user_id
+    end
+
+    private def relevant_concept_results_where_clause
+      return "AND act_s.user_id = #{user_id}" if user_id.present?
```
brendanshean: I think there's a SQL injection risk here.

You could use a [placeholder condition](https://guides.rubyonrails.org/v7.2/active_record_querying.html#placeholder-conditions) for `user_id`
```
  private def relevant_concept_results_where_clause
    return 'AND act_s.user_id = :user_id' if user_id.present?

    ''
  end
```

Then add a conditional bind for it in the question_responses_base:
```
  private def sql
    binds = { classroom_id:, grade:, state_standard_ids: }
    binds[:user_id] = user_id if respond_to?(:user_id) && user_id.present?

    ActiveRecord::Base.sanitize_sql([query_sql, binds])
  end
```

services/QuillLMS/app/models/concept.rb
```diff
@@ -22,10 +22,12 @@ class Concept < ApplicationRecord
   belongs_to :parent, class_name: 'Concept', foreign_key: :parent_id
   belongs_to :replacement, class_name: 'Concept', foreign_key: :replacement_id
   validates :name, presence: true
+  has_many :children, class_name: 'Concept', foreign_key: :parent_id
```
happythenewsad: `children` is a heavily loaded term. `subconcepts` would be more specific. Although that brushes up against our `l0, l1, l2` terminology...

services/QuillLMS/app/controllers/teachers/progress_reports_controller.rb
```diff
@@ -276,4 +289,21 @@ def track_teacher_assigned_targeted_practice
   private def authorize_teacher!
     classroom_teacher!(params[:classroom_id])
   end
+
+  private def fetch_state_standards
```
brendanshean: nit: why not name this `all_state_standards` for consistency with other method.

brendanshean: Maybe a `.distinct` call before the map in case there are duplicates from the joins?

services/QuillLMS/app/controllers/teachers/progress_reports_controller.rb
```diff
@@ -276,4 +289,21 @@ def track_teacher_assigned_targeted_practice
...
+    StateStandard
+      .joins(state_standard_concepts: { concept: :parent })
+      .includes(state_standard_concepts: { concept: :parent })
+      .where(state_standard_framework: @state_standard_framework)
+      .merge(StateStandardConcept.explicit)
+      .map { |state_standard| state_standard.as_json.merge({ grades: state_standard.state_standard_concepts.map(&:grade).uniq, l2_concepts: state_standard.state_standard_concepts.map(&:concept).map(&:parent).map(&:name).uniq }) }
+  end
+
+  private def all_skill_categories
```
brendanshean, line 303: Maybe a `.distinct` call before the map in case there are duplicates from the joins?

brendanshean (approved):
A long PR, but super easy to follow with all of the service objects and thorough testing.   Great job on this!

services/QuillLMS/app/services/standards_report/compose_scores_by_user.rb
```diff
@@ -0,0 +1,22 @@
+# frozen_string_literal: true
+
+module StandardsReport
+  class ComposeScoresByUser < ApplicationService
```
happythenewsad, line 4: +1 for the `Compose*` naming schema

happythenewsad (approved):
Nice work!

PR #13575, opened 2026-01-21
Make sure to score the new Grammar FITB questions properly

anathomical:
## WHAT
In Skills Report use special FITB scoring when a question is of the new Grammar FITB type, not just the old Connect FITB types.
## WHY
We want to score these new questions with the same logic rather than treating them all as having 5 attempts for scoring
## HOW
Use the FITB scoring logic for both values of `question_type`, not just the one

PR Checklist | Your Answer
------------ | -------------


emilia-friedberg (approved):
That was fast! Thank you!

eadams17 (approved):

PR #13576, opened 2026-01-22
add code to ensure grammar fitb activities aren't added to existing packs that already contain connect fitb activities

emilia-friedberg:
## WHAT
Add code to ensure that, if a unit already contains the Connect version of a FITB activity, updating it (normal path is by assigning more students) will not add the Grammar version of the activity. Also, add a script to clean up duplicates that have already been assigned (but not completed).

## WHY
We don't want duplicate activities in these packs.

## HOW
All duplicated Grammar activities have uids that are the same as the Connect activities, but with `fitb-` appended to the front, so we're just adding a check to see if the Connect version of the activity already exists before adding more activities to it.

### Screenshots

### Notion Card Links
https://www.notion.so/quill/FITB-migration-resulting-in-duplicate-assignments-2efd42e6f941801ab49acb509fc4d02f

Replicated the bug on staging, then implemented this fix and confirmed that I could no longer replicate it for existing packs, but assigning new packs still works as expected (contains Grammar FITB activities).

PR Checklist | Your Answer
------------ | -------------


happythenewsad (approved):
Nice work! 

An additional QA check to check for overzealous deletions would be helpful, since this is a destructive procedure. Although I do note that you check for not-started activities on line 820

PR #13589, opened 2026-01-27
new-standards-only-include-practice

anathomical:
## WHAT
- Add a service to find standards with practice by user
- And use that service when populating the drop-down in the report
## WHY
The spec calls for the Standards drop-down to only include options that have some amount of practice data in order to help us avoid having a bunch of options that lead to blank reports.
## HOW
- Build a service that can take a `classroom_id` or teacher `user_id` and return a list of `StateStandard.id`s, as well as a list of which classrooms for that teacher have practice associated so the front-end can further filter based on selected classroom
- Use the results from the service to further filter the controller that generates data for the drop-down
- Include the associated classrooom_ids the service identifies in the payloads passed to the front-end for additional filtering

### Notion Card Links
https://www.notion.so/quill/Standards-Report-2-0-Teacher-Views-2c3d42e6f9418176be79db0424e2a648?source=copy_link#2e2d42e6f9418086b30be1273fee96d4

- Deployed to staging
- Confirm that every listed Skill in the "By Student" reports has data associated with it

PR Checklist | Your Answer
------------ | -------------


emilia-friedberg (approved):
LGTM!

services/QuillLMS/app/services/state_standards/standards_with_practice.rb
```diff
@@ -0,0 +1,56 @@
...
+
+      ClassroomsTeacher.where(user_id:).pluck(:classroom_id)
+    end
+
+    private def standard_ids_sql
+      <<-SQL
+        WITH filtered_sessions AS (
+          SELECT DISTINCT activity_sessions.activity_id, classroom_units.classroom_id
+            FROM classroom_units
+            JOIN activity_sessions ON classroom_units.id = activity_sessions.classroom_unit_id
```
happythenewsad, line 40: This might need to be cached at some point, if it becomes too slow for a synchronous UX

happythenewsad (approved):
LGTM

PR #13595, opened 2026-01-28
new-standards-maximal-query-performance

anathomical:
## WHAT
Use LLM (Claude) to optimize the query for maximal data sets
## WHY
While most query cases had acceptable performance, the maximal data set (all history for all concepts) was choking the staging database
## HOW
Feed the query into Claude and implement its suggestions:
1. Declare `concept_uids` as `MATERIALIZED` in order to force the query planner to pre-calculate it rather than risk it being pulled in as a less efficient subquery
2. Use `CROSS JOIN LATERAL` to join `concept_results` to `relevant_sessions`.  This is another way to force the query planner into optimal behavior: specifically pulling `concept_results` filtered by its `activity_session_id` index

- Deploy to staging
- Confirm that maximal queries with users that were timing out run in less than 5 seconds

PR Checklist | Your Answer
------------ | -------------
Have you added and/or updated tests? | No.  We explicitly expect the behavior to stay the same


brendanshean (approved):
Nice lateral thinking.

anathomical:
> Nice lateral thinking.

It turns out that the LLMs are a lot better than I am at all the tricks that can be used to force the query planner to do the right thing.

happythenewsad (approved):

PR #13599, opened 2026-01-29
standards-report-new-smart-group-chips-backend

anathomical:
## WHAT
Add associated description, grade and skill lists to By Standards query
## WHY
So that the front-end can display this data to teachers as context for standards
## HOW
- Update the `RawSqlRunner` class to handle JSON values gracefully so that the results of `ARRAY_AGG` columns get parsed into Ruby arrays (an excellent LLM suggestion)
- Aggregate grades and "skills" (L2 concepts) associated with a given State Standard and surface that data to the API
- Tweak mocks in the `ClassroomScoresByStandard` spec to include the new data attributes

### Notion Card Links
https://www.notion.so/quill/Standards-Report-2-0-Teacher-Views-2c3d42e6f9418176be79db0424e2a648?source=copy_link#2e0d42e6f94180399387d0b41161ea69

- On staging confirm that API responses come back with Skill and Grade contexts for Standards
- Confirm that performance is not negatively impacted
- Confirm that the `description` is being populated for both Standards with data, and placeholder Standards with no data (which are sourced differently in the database)

PR Checklist | Your Answer
------------ | -------------


services/QuillLMS/app/services/standards_report/compose_scores_by_standard.rb
```diff
@@ -18,23 +18,32 @@ def run
...
           id = standard_data[:id]
-          StandardResponsesProcessor.run(user_responses:, name:, id:, questions_hash:, total_classroom_students:)
+          description = standard_data[:description]
+          standard_scope = { grades: standard_data[:grades], skills: standard_data[:skills] }
+          StandardResponsesProcessor.run(user_responses:, name:, id:, questions_hash:, total_classroom_students:, description:, standard_scope:)
         end
     end
 
-    private def standard_placeholders = state_standards.to_h { |standard| [standard.code, { id: standard.id, responses: [] }] }
+    private def standard_placeholders = state_standards.includes(:state_standard_concepts).to_h { |standard| [standard.code, { id: standard.id, description: standard.description, grades: standard.state_standard_grades.map(&:grade), responses: [] }] }
```
dandrabik: You can only use `includes` here if you are passing a finder. I suspect you are probably passing an array of results, so you can us `includes` here (of if you are passing a finder, you need to adjust your tests).

services/QuillLMS/app/controllers/teachers/units_controller.rb
```diff
@@ -365,7 +365,7 @@ def score_info
     units.map do |unit|
       classroom_student_ids = Classroom.find(unit['classroom_id']).students.ids
       if unit['assigned_student_ids'] && classroom_student_ids
-        active_assigned_student_ids = JSON.parse(unit['assigned_student_ids']) & classroom_student_ids
+        active_assigned_student_ids = unit['assigned_student_ids'] & classroom_student_ids
```
dandrabik, line 368: Just a note, this will error if one of these is nil (not sure if that is possible here). Wrapping in Array(thing) & Array(thing2) is a nil check if you need it.

emilia-friedberg (approved):
LGTM! Thank you for taking care of this so quickly!

dandrabik (approved):
Overall looks ok, some small comments, but nothing blocking.

services/QuillLMS/app/controllers/teachers/units_controller.rb
anathomical: I believe that the preceding `if` condition above nil-checks this.

services/QuillLMS/app/services/standards_report/compose_scores_by_standard.rb
anathomical: Good call-out.  This was a bad concession to the specs not passing a finder while the code itself does.  I've removed this check and updated the spec to pass a finder object.

PR #13608, opened 2026-02-02
Fix recommendation activity count discrepancy

brendanshean:
## WHAT
Activity counts on recommended practice page are off.

## WHY
They were including archived activities

## HOW
~~Remove archived activities from the count~~

Add `include_archived_activities` flag with default being true so that progress reports still show archived activities but recommendations panel does not.  

### Screenshots

### Notion Card Links
https://www.notion.so/quill/Discrepancy-between-displayed-vs-actual-number-of-recommended-practice-Pre-AP-Writing-Skills-Survey-2f8d42e6f941807eb4c3f1821118649b?source=copy_link

I confirmed that the counts are correct now.

PR Checklist | Your Answer
------------ | -------------


eadams17 (approved):

anathomical (approved):
I like the way you set up the flag on this one.  The early return on the query construction was cleaner than my own instinct of making the `where.not` itself conditional.

brendanshean:
> I like the way you set up the flag on this one. The early return on the query construction was cleaner than my own instinct of making the `where.not` itself conditional.

Thanks.  I tend to find the inline conditionals add too much cognitive overhead for me when reading it later.

PR #13613, opened 2026-02-03
teacher-standards-framework-selection

anathomical:
## WHAT
- Modify `StateStandardConcept` to allow for a given `StateStandardFramework` to have two different standards assigned to the same concept for a given grade (a thing that the index prevented before)
- Update the `StateStandards::FillImplicitRecords` service to handle the new recognition that a framework might assign multiple standards to the same concept within a given grade
- Add a `state_standards_framework_id` to `TeacherInfo`
- Update the `TeacherinfoController` to allow API calls to update the new framework value
- Update Standards Report controller to fetch the `StateStandardFrameworkId` from the `current_user` if it is available, and only to fall back to TEKS if it's not (note that we'll change this fallback behavior in the future)
- Overhaul the thor task that imports CSVs into `StateStandard` related records to properly parse the other defined states in our current CSV beyond Texas
## WHY
- This is intended behavior, and one of our other supported standards needs to do this
- We want to be able to know which state standard is appropriate for a given teacher
- The front-end will need to be able to set this value for a teacher using a modal selector
- Since we're storing this on the user now, we want to actually use our stored value when it's available
- We want to go ahead and get the other standards into our database to both test the robustness of the service and to allow for testing future framework selection functionality
## HOW
- Replace the old unique index with a new one that is slightly less restrictive
- Tweak the logic to no longer simply assume that there is one valid standard with the closest grade to the one being processed, instead find the closest grade value and then identify all relevant standards within the framework
- Create a migration to add an optional column to the `teacher_infos` table
- Tweak the `update` action in the `TeacherInfoController` to also read `state_standards_framework_id` from `params`, and use that value to update the appropriate column when it is provided
- Tweak the logic that sets `@default_state_standards_framework` to try to pull it from `current_user&.teacher_info` before using the default fallback
- Refactor the task to handle inputs and individual parsing more generically, with special case methods for handling unique parts of different states (like grade ranges in NY standards)

### Notion Card Links
https://www.notion.so/quill/Standards-Report-2-0-Teacher-Views-2c3d42e6f9418176be79db0424e2a648?d=2e7d42e6f9418011a766001cc237364d

- Ran the thor task on staging
- Ran the `StateStandards::FillImplicitRecords` service for each new state standards framework
- Confirmed that the the appropriate `StateStandardConcept` records are created, both explicit and implicit

PR Checklist | Your Answer
------------ | -------------


services/QuillLMS/app/services/state_standards/fill_implicit_records.rb
```diff
@@ -13,46 +13,69 @@ def initialize(state_standards_framework_id:)
...
+      @explicit_state_standard_concepts ||= StateStandardConcept.explicit.where(state_standards_framework_id:)
+    end
+
+    private def grouped_state_standard_concepts
+      explicit_state_standard_concepts.group_by(&:concept_id)
+    end
 
-    private def create_implicit_standards_for_concept(concept_id, explicit_standards)
-      explicit_grades = explicit_standards.map(&:grade).uniq
+    private def create_implicit_records_for_concept(concept_id, explicit_records)
```
happythenewsad: can use keyword args here and on line 65

happythenewsad (approved):

services/QuillLMS/app/controllers/teachers/progress_reports_controller.rb
```diff
@@ -69,8 +69,8 @@ def skills_report
   def standards_report
     @classrooms = classrooms.includes(:students).map { |classroom| classroom.with_students }
 
-    # TODO: pull this from current_user, but use default TEKS for now
-    @default_state_standards_framework = StateStandardsFramework.find_by(abbreviation: 'TEKS')
+    # TODO: pull this from current_user, use default fallback TEKS for now
+    @default_state_standards_framework = current_user&.teacher_info&.state_standards_framework || StateStandardsFramework.find_by(abbreviation: 'TEKS')
```
dandrabik: I think it would be good to move this default into the model as e.g. `StateStandardsFramework.default` and have a `DEFAULT_ABBREVIATION = 'TEKS'`

Also, TODO comment on line above can be removed.

dandrabik (commented):
Looks good, thanks for putting this together.

emilia-friedberg (approved):

PR #13616, opened 2026-02-04
Rubocop update Part 2: safe cops with no autocorrection

brendanshean:
## WHAT
Add cops that are safe (i.e. no false positives) along with no autocorrection.

Let's have discussion [here](https://www.notion.so/quill/Rubocop-Updates-Part-2-2fdd42e6f9418089a6e2e944522511e3?source=copy_link) about removal of any of the cops.

Most of these cops are straightforward but I would flag this [Rails/WhereNotWithMultipleConditions
](https://docs.rubocop.org/rubocop-rails/cops_rails.html#railswherenotwithmultipleconditions) in particular as a surprising one.

## WHY

## HOW

### Screenshots

### Notion Card Links
https://www.notion.so/quill/Rubocop-Updates-Part-2-2fdd42e6f9418089a6e2e944522511e3?source=copy_link

I have confirmed the flash messages work on staging.

PR Checklist | Your Answer
------------ | -------------


happythenewsad:
Noting here, but also applies to some other rubocop PRs: Although the Capybara gem is installed and referenced in a few places, we don’t actually use it (and shouldn’t, now that we have playwright). 

So the capybara cop directives are unnecessary.

services/QuillCMS/lib/tasks/import_from_firebase.rake
```diff
@@ -281,7 +281,7 @@ namespace :responses do
   end
 
   def unstringify_concept_results
-    Response.where.not(concept_results: nil, optimal: nil).in_batches do |batch|;
+    Response.where.not(concept_results: nil, optimal: nil).in_batches do |batch|; # rubocop:disable Rails/WhereNotWithMultipleConditions
```
anathomical, line 284: Why disable here instead of updating the query?

services/QuillLMS/app/models/subscription.rb
```diff
@@ -144,7 +144,14 @@ class RenewalNilStripeCustomer < StandardError; end
   scope :not_recurring, -> { where(recurring: false) }
   scope :not_stripe, -> { where(stripe_invoice_id: nil, stripe_subscription_id: nil) }
   scope :started, -> { where('start_date <= ?', Date.current) }
-  scope :paid_with_card, -> { where.not(stripe_invoice_id: nil, stripe_subscription_id: nil).or(where(payment_method: 'Credit Card')) }
+
+  scope :paid_with_card, lambda {
+    where(
+      'NOT (stripe_invoice_id IS NULL AND stripe_subscription_id IS NULL) ' \
+      "OR payment_method = 'Credit Card'"
```
anathomical: Not really a big deal, but I tend to not like `\` for line breaks.  Thoughts on doing this with `<<-`?
```
where(
  <<-SQL
    NOT (stripe_invoice_id IS NULL AND stripe_subscription_id IS NULL)
    OR payment_method = 'Credit Card'
  SQL
)
```

services/QuillLMS/db/migrate/20170809151404_add_primary_key_to_schools_users.rb
```diff
@@ -2,6 +2,6 @@
 
 class AddPrimaryKeyToSchoolsUsers < ActiveRecord::Migration[4.2]
   def change
-    add_column :schools_users, :id, :primary_key
+    add_column :schools_users, :id, :primary_key # rubocop:disable Rails/DangerousColumnNames
```
anathomical, line 5: I kind of love this rule.

services/QuillLMS/spec/services/vitally_integration/serialize_vitally_sales_organization_spec.rb
```diff
@@ -333,11 +333,11 @@
     end
   end
 
+  # rubocop:disable FactoryBot/ExcessiveCreateList
```
anathomical: Do we need this wrapper even though we're removing the `let!(:schools)` call because of the loop below with multiple `create_list` calls?

anathomical (approved):
This looks good to me.  A couple of notes about style and stuff, but this seems totally reasonable as-is.

brendanshean:
> Noting here, but also applies to some other rubocop PRs: Although the Capybara gem is installed and referenced in a few places, we don’t actually use it (and shouldn’t, now that we have playwright).
> 
> So the capybara cop directives are unnecessary.

That's a good point.  I'll remove in a later PR where I update all these gems.

services/QuillLMS/spec/services/vitally_integration/serialize_vitally_sales_organization_spec.rb
brendanshean: No, we definitely don't.  Thanks for catching that.

services/QuillLMS/app/models/subscription.rb
brendanshean: I also don't like `\` for line breaks.  I will fix this.

services/QuillCMS/lib/tasks/import_from_firebase.rake
brendanshean: I'll add a comment here.  Rails 6.1 changed its SQL [semantics](https://docs.rubocop.org/rubocop-rails/cops_rails.html#railswherenotwithmultipleconditions) with how these queries are handled.  However, this code was written before then and I'm guessing it's code we don't use anymore so I didn't want to spend anymore time on it.

happythenewsad (approved):

services/QuillCMS/app/workers/rematch_responses_for_question_worker.rb
```diff
@@ -47,7 +47,7 @@ def ungraded_responses(question_uid)
 
   def machine_graded_responses(question_uid)
     Response.where(question_uid:)
-      .where.not(parent_id: nil, parent_uid: nil)
+      .where("NOT (parent_id IS NULL AND parent_uid IS NULL)")
```
dandrabik: Knowing the logic of this, I believe we need to update this to `where.not(parent_id: nil).where.not(parent_uid: nil)`, e.g. the pre-6.1 logic

dandrabik (approved):
Overall, looks good. Thanks for adding these.

I noted one change which we want to change the query here (the double not change in 6.1 essentially introduced a bug there)

services/QuillCMS/app/workers/rematch_responses_for_question_worker.rb
brendanshean: Thank you.  I will update.

PR #13622, opened 2026-02-04
Add Distinct to queries in StudentLearningSequences::FillMissingWorker

dandrabik:
## WHAT
Add a `DISTINCT` to all of the queries in `StudentLearningSequences::FillMissingWorker`
## WHY
These queries are queuing a lot of jobs (and backing up our queues), and many of them appear to be duplicates. 
## HOW
I believe by adding a DISTINCT here it shouldn't change the logic of the query, but will remove duplicates.
### Screenshots
When exploring the data, I uploaded the SQL result to Sheets and removed the duplicates. So 150K jobs reduced to 37K
Screenshot 2026-02-04 at 4 59 19 PM

I've been running these queries in metabase and comparing the results in sheets and in the console. There seems to be more refactoring to do here, but this would be such a win I'd like to get it out. 

PR Checklist | Your Answer
------------ | -------------
Have you added and/or updated tests? |  No


anathomical (approved):
`DISTINCT` makes a ton of sense here.  If these params end up in the queue a second time, they'll no-op since the workers are idempotent, but still take up resources determining that the job has already run.

PR #13623, opened 2026-02-04
auto-assign-standards-frameworks-to-users

anathomical:
## WHAT
- Create a service that examines a user's schools to determine what their default `StateStandardsFramework` should be
- Add a worker to wrap the new service
- Add an `after_save` callback to the `SchoolsUsers` model in order to enqueue the worker when a school is attached to a user if that user is a Teacher/Admin and doesn't already have a framework assigned
- Add a Thor task to backfill this data for all existing Teachers/Admins
## WHY
We want to pre-populate selected StateStandardsFrameworks for users with our best guess of what they should have assigned so that they don't have to do work to select the correct one most of the time
## HOW
- Create a service that sets a `teacher_info.state_standards_framework_id` for a User based on what state their associated school is located in
- Create a simple Sidekiq worker wrapper on the `default` queue to let us run this service async
- Hook the worker up to an `after_save` hook that confirms that 1) the user is a teacher/admin, and 2) doesn't already have this value set so that new users will onboard with this set
- Create a Thor task that queries for all relevant users with missing data and enqueues them for this new service

### Notion Card Links
https://www.notion.so/quill/Standards-Report-2-0-Teacher-Views-2c3d42e6f9418176be79db0424e2a648?source=copy_link#2e2d42e6f941805c9c2eea2c778f70ea

- Deploy to staging
- Run the thor task
- Confirm all teachers and admins have this data set
- Confirm that no students have this data set
- Create a new teacher user
- Ensure that the new teacher gets this data set

PR Checklist | Your Answer
------------ | -------------


dandrabik (approved):
Overall, the code looks good. The only questions I have are really a product ones: 

Do we want to assign a state standard to a teacher if we don't have that state functionality yet (e.g. a teacher in Connecticut)? 

I could see another version of this in which a teacher's state standard stays blank unless we have that state standard, and every time we bring a state online, we run a script to populate teachers in that state. That would keep the data a little fresher, e.g. teachers may move in the time you run this script now and have these standards online.

There may be other reasons to populate it now, e.g. maybe there will be 'Connecticut is coming soon' functionality, I just figured it was worth asking the question.

happythenewsad (approved):

PR #13625, opened 2026-02-05
Rubocop updates Part 4: safe cop and unsafe autocorrect

brendanshean:
## WHAT
These are safe cops with unsafe autocorrection.

Most of these are straightforward but there are two in particular that deserve special attention:
[Rails/WhereEquals](https://github.com/empirical-org/Empirical-Core/pull/13625/commits/fe2051d484921b0b688eab22e7a4dc8f3dfcb629)
[Rails/WhereRange](https://github.com/empirical-org/Empirical-Core/pull/13625/commits/03111166c18e12f86f10423b9ee43d2e39088a05)
I feel like the active record shorthand reads cleaner, but you do have to guard against `nil` more carefully than with Raw SQL.

Let's have discussion [here](https://www.notion.so/quill/Rubocop-Updates-Part-4-2fdd42e6f94180bf81efdde18e60361f?v=2f6b6bedb9e74f55a581c015c3df582b&source=copy_link) about removal of any of the cops.

## WHY

## HOW

### Screenshots

### Notion Card Links
https://www.notion.so/quill/Rubocop-Updates-Part-4-2fed42e6f94180549ccbd52e598de5b9?source=copy_link

PR Checklist | Your Answer
------------ | -------------


services/QuillCMS/spec/controllers/questions_controller_spec.rb
```diff
@@ -203,7 +203,7 @@
       it 'should return graded responses, 1 from MultipleChoiceResponse and 1 from ungraded responses since there are no more graded responses' do
         get :multiple_choice_options, params: {question_uid: '123'}
 
-        expect(response.status).to eq 200
+        expect(response).to have_http_status 200
```
anathomical: This probably doesn't really matter, but I note that in this file the syntax is without parens (`have_http_status 200`), but in the following file it's with parens (`have_http_status(200)`).  Kind of weird that Rubocop doesn't enforce this one way or the other.

services/QuillLMS/app/controllers/cms/rosters_controller.rb
```diff
@@ -37,7 +37,12 @@ def upload_teachers_and_students
         student ||= User.create!(name: s[:name], email: student_email, password:, password_confirmation: password, role: 'student')
 
         teacher = User.find_by(email: teacher_email)
-        classroom = Classroom.joins(:classrooms_teachers).where('classrooms_teachers.user_id = ?', teacher.id).where(name: s[:classroom]).first
+        next unless teacher
+
+        classroom = Classroom.joins(:classrooms_teachers)
+          .where(classrooms_teachers: { user_id: teacher.id })
+          .where(name: s[:classroom])
+          .first
```
anathomical, line 45: If there was not a valid `User` for `teacher_email`, was the old code creating the classroom and then crashing when it tried to call `ClassroomsTeacher.create!` with a `nil` `teacher` object with the column in the database being configured as `NOT NULL`?

services/QuillLMS/app/controllers/admins_controller.rb
```diff
@@ -342,11 +342,12 @@ def create_and_link_accounts
   private def handle_bulk_archive_classes(teacher_id, filter_date)
     teacher = User.find_by(id: teacher_id)
     classroom_ids = teacher&.classrooms_i_own&.pluck(:id) || []
+    return if filter_date.blank?
 
     Classroom.joins(:classrooms_teachers)
       .where(id: classroom_ids)
       .where(classrooms_teachers: { user_id: teacher.id, role: 'owner' })
-      .where('classrooms.created_at < ?', filter_date)
+      .where(classrooms: { created_at: ...filter_date })
```
anathomical: I will say I'm so glad that this syntax got added to `ActiveRecord`

anathomical (approved):

services/QuillLMS/app/controllers/cms/rosters_controller.rb
brendanshean: That would be my read on this

services/QuillLMS/app/controllers/admins_controller.rb
brendanshean: I didn't know about the `...` vs `..` distinction until I turned this Cop on.

happythenewsad (approved):
Most of these are value-add. One isn't, but is not a big deal to leave in.

services/QuillLMS/app/controllers/admins_controller.rb
dandrabik: Yeah, I commented in Notion, but I personally find this corrected version harder to read. I can adjust if others like this one.

dandrabik (approved):
Code looks good. I don't love the Rails/WhereRange one, but will defer to the group.

dandrabik (approved):
original review didn't record for some reason.

services/QuillCMS/spec/controllers/questions_controller_spec.rb
brendanshean: There's a cop for this:
```
 Style/MethodCallWithArgsParentheses:
    Enabled: true                                                                                                                    
    EnforcedStyle: require_parentheses
```                                                                                                                                     
  Though note — with require_parentheses, it would flag a lot of common Ruby idioms like puts "hello", require 'foo', raise "error", etc. You'd likely need a hefty AllowedMethods list. Most projects that enable this cop end up with something like:
```
  Style/MethodCallWithArgsParentheses:
    Enabled: true
    EnforcedStyle: require_parentheses
    AllowedMethods:
      - require
      - require_relative
      - puts
      - print
      - raise
      - include
      - extend
      - prepend
      # ... many more

  It's one of the more controversial/noisy cops, which is probably why RuboCop doesn't enable it by default and why most projects
  leave it off.
```
we could probably figure out a configuration, but will leave that for some other time.

services/QuillLMS/app/controllers/admins_controller.rb
brendanshean: disabling this one.

PR #13626, opened 2026-02-05
Optimize the StudentLearningSequences::HandleAssignment service to cut down on finds.

dandrabik:
## WHAT
Optimize the `StudentLearningSequences::HandleAssignment` services to reduce SQL finds. Find all the objects upfront, then use those objects, instead of re-finding them.
## WHY
These jobs are backing up the default queue, so making them a little faster is a low-lift way to improve that. 

My PR yesterday(#13622) reduced the number of jobs loaded, this makes each job a little faster. Chipping away.
## HOW
Pass the `activity` and `classroom_unit` objects instead of the `id`s. Essentially we were finding all of the activities at once, then within a loop, finding each individual `activity` from its `id`.
### Screenshots
NR transaction traces aren't working right now, but long-running jobs had hundreds of `Activity.find` which were taking up 5 seconds of the 10 second run time.
Screenshot 2026-02-05 at 11 35 44 AM

### Notion Card Links
https://www.notion.so/quill/Improve-default-Delayed-Job-queue-back-ups-2fed42e6f941800980a3e2fe1fe05c3f

Ran the job locally.

PR Checklist | Your Answer
------------ | -------------


brendanshean (approved):
I like this low-lift approach.  Maybe worth doing elsewhere in the codebase

anathomical (approved):
I like this clean-up as well.  We should probably find patterns that let us pass more full objects around between related services instead of always using ids.

PR #13629, opened 2026-02-05
Rubocop updates Part 5: unsafe cop and unsafe autocorrect 

brendanshean:
## WHAT
These are unsafe cops (false positives possible) with unsafe autocorrection.

Let's have discussion [here](https://www.notion.so/quill/Rubocop-Updates-Part-5-2fdd42e6f9418089b38ad30536511c9c?v=2f6b6bedb9e74f55a581c015c3df582b&source=copy_link) about removal of any of the cops.

## WHY

## HOW

### Screenshots

### Notion Card Links
https://www.notion.so/quill/Rubocop-Updates-Part-5-2fdd42e6f9418089b38ad30536511c9c?v=2f6b6bedb9e74f55a581c015c3df582b&source=copy_link

PR Checklist | Your Answer
------------ | -------------


happythenewsad (approved):
Note: a lot of these are specs-only

dandrabik (approved):
I didn't review the code, but these rules seem good to me!

services/QuillLMS/app/controllers/activities_controller.rb
```diff
@@ -133,7 +133,7 @@ def calculate_selected_activities
   end
 
   protected def user_completed_view_lessons_tutorial?
-    !!Milestone.find_by(name: 'View Lessons Tutorial').users.include?(current_user)
+    !Milestone.find_by(name: 'View Lessons Tutorial').users.exclude?(current_user)
```
anathomical: Rubocop probably shouldn't recognize that this probably needs a complete refactor. This should probably be `Milestone.joins(:user_milestones).exists?(name: 'View Lessons Tutorial', user_milestones: { user_id: current_user.id })` or something like that

services/QuillLMS/app/controllers/blog_posts_controller.rb
```diff
@@ -139,11 +139,9 @@ def show_topic
   end
 
   private def topics(topic_names)
-    topics = []
-    topic_names.each do |name|
-      topics.push({ name:, slug: CGI::escape(name.downcase.gsub(' ', '-')) })
+    topics = topic_names.map do |name|
```
anathomical, line 142: Again, Rubocop wouldn't notice this, but do we even need to assign this value to `topics` since we could just let the `topic_names.map` return implicitly

services/QuillLMS/engines/evidence/app/models/evidence/labeled_entry.rb
```diff
@@ -62,7 +62,7 @@ def self.closest_prompt_texts(entry:, prompt_id:, label:, limit: DEFAULT_RAG_LIM
       if (1 - results.first.last) < label0_threshold
         [[results.first[0], LABEL0]]
       else
-        results.map { |result| result[0..1] }
+        results.pluck(0..1)
```
anathomical, line 65: Oh, weird.  I didn't know you could pass a range into `[]`

services/QuillLMS/engines/evidence/spec/factories/evidence/gen_ai/highlight_groups.rb
```diff
@@ -16,7 +16,7 @@
 #
 FactoryBot.define do
   factory :evidence_gen_ai_highlight_group, class: 'Evidence::GenAi::HighlightGroup' do
-    association :prompt, factory: :evidence_prompt
+    prompt factory: %i[evidence_prompt]
```
anathomical: This feels like some sort of mistake on Rubocop's part.  Shouldn't this just be `factory: :evidence_prompt`?  We don't need to pass an array of symbols to `factory:` I don't think.

services/QuillLMS/spec/controllers/api/v1/evidence_controller_spec.rb
```diff
@@ -3,7 +3,7 @@
 # Test to verify engine mounted properly
 require 'rails_helper'
 
-RSpec.describe Evidence::ActivitiesController, type: :controller do # rubocop:disable RSpec/SpecFilePathFormat
+RSpec.describe Evidence::ActivitiesController do # rubocop:disable RSpec/SpecFilePathFormat
```
anathomical, line 6: Beyond the scope of this PR, but any idea if this disable comment is still necessary on this block?  I wonder if we want to finish the new Rubocop settings series by doing an audit of still-disabled blocks and inline.

anathomical (approved):
I like the `pluck` rule.  Way cleaner to read than those `map` calls to extract a specific key.

services/QuillLMS/spec/controllers/api/v1/evidence_controller_spec.rb
brendanshean: This disable comment is necessary to the mismatch between the directory structure and the describe block class.
RSpec would expect this to be `RSpec.describe Api::V1::EvidenceController` based on the directory structure.

services/QuillLMS/app/controllers/activities_controller.rb
brendanshean: Good idea.  This will change it from 2 queries to 1.

services/QuillLMS/app/controllers/blog_posts_controller.rb
brendanshean: Rubocop should notice this.  I'm not sure why https://www.rubydoc.info/gems/rubocop/RuboCop/Cop/Lint/UselessAssignment
isn't enabled but I will look into it.

PR #13633, opened 2026-02-06
Rubocop updates Part 6: Misc

brendanshean:
## WHAT
- Upgrade rubocop and rubocop plugins
- Remove rubocop-capybara
- Add rubocop-graphql
- Cleanup failures as a result of the upgrade

## WHY
- I want everything upgraded before trying `NewCops: enabled`
- Capybara system specs are no longer used
- There is legacy graphql code, but this is so that we don't get suggestions for it
- To satisfy Rubocop

## HOW
- bundle install
- remove the gem from the Gemfile
- add plugin to `.rubocop.yml`
- use autocorrect where possible and then update other violations manually

### Screenshots

### Notion Card Links

PR Checklist | Your Answer
------------ | -------------


services/.rubocop_base.yml
```diff
@@ -1,8 +1,9 @@
-require:
+plugins:
   - rubocop-rails
   - rubocop-rspec
   - rubocop-rspec_rails
   - rubocop-factory_bot
+  - rubocop-graphql
```
brendanshean: In the previous [commit](https://github.com/empirical-org/Empirical-Core/pull/13633/commits/70b7995ade66723d6aa1e013e7ce7555d9466ab4), I added a useful configuration:  `SuggestExtensions: true` which suggests new plugins.

 I'm then adding in `rubocop-graphql` but then putting violations into `.rubocop_todo.yml` since I don't think it's worth fixing those at this point.

This way if newer rubocop plugins become available, we will see them without getting the noise of `rubocop-graphql` each time.

dandrabik (approved):
Looks good.

Note, for some reason, I thought we weren't using the <<~ for SQL queries because it mangled some of the dynamic queries. Reading it now, I can't see how that would happen (tests should also cover this). So looks good. But I would check the error logs post deploy to check for any mangled sql

happythenewsad (approved):

services/QuillLMS/app/controllers/api/v1/users_controller.rb
```diff
@@ -27,7 +27,7 @@ def student_and_teacher_ids_for_session
     return [] unless current_user
 
     RawSqlRunner.execute(
-      <<-SQL.squish
+      <<~SQL.squish
```
anathomical, line 30: This is a good rule.  Though I don't know that it actually matters when we're combining it with `squish`.  Anyway, I think it's a good rule to have on for cases where we don't call `squish`

anathomical (approved):

PR #13634, opened 2026-02-06
Rubocop updates part 7: NewCops enable

brendanshean:
## WHAT
After turning on the setting NewCops enable, these are the cops that get turned on.

I'm going to do the two following cops in separate PRs for clarity

1. ~~Rails/FindByOrAssignmentMemoization: The find_by here is not memoized if it finds nil.~~ 
2. ~~Style/FetchEnvVar: Some of these ENV variables should be refactored into constants for `config/initializers` but that's out of scope here.  Also, some of the fetch should probably not have a fallback (i.e. we want an error).  Or maybe we want that to be the default?~~

Feel free to add discussion [here](https://www.notion.so/quill/Rubocop-Updates-Part-7-2ffd42e6f94180febb3cc660c8be8916?source=copy_link)

## WHY

## HOW
Add `.rubocop_base.yml` and then address violations manually.

### Screenshots

### Notion Card Links
https://www.notion.so/quill/Rubocop-Updates-Part-7-2ffd42e6f94180febb3cc660c8be8916?source=copy_link

PR Checklist | Your Answer
------------ | -------------


dandrabik (approved):
Looks good. I'm on the fence about `Rails/FindByOrAssignmentMemoization` since the simpler syntax might be worth the potential extra DB calls, but I don't feel that strongly either way.

happythenewsad (approved):

services/QuillCMS/app/controllers/fragments_controller.rb
```diff
@@ -2,7 +2,7 @@ class FragmentsController < ApplicationController
 
   def is_sentence
     response = HTTParty.post(
-      "#{ENV['QUILL_NLP_URL']}/sentence_or_not",
+      "#{ENV.fetch('QUILL_NLP_URL', nil)}/sentence_or_not",
```
anathomical: I'll flag this in Notion, but I'm not sure I'm comfortable with changing this behavior.  Having this code raise an error if the value isn't set may be preferable to letting this code pass and then something break in hard-to-track-down ways.  I don't dislike moving to `fetch` calls, but putting in a default `nil` fallback seems like something we want to make explicit decisions about, not something we want Rubocop to yell at us about.

services/QuillLMS/app/models/feedback_history.rb
```diff
@@ -251,14 +251,14 @@ def self.save_feedback(
 
   def rule_violation_repititions?
     histories_from_same_session.where(rule_uid:)
-      .where('attempt < ?', attempt)
-      .count > 0
+      .where(attempt: ...attempt)
```
anathomical: I thought we decided to avoid this syntax?

anathomical (approved):
We discussed some of this in Slack, but with the exception of potential risks of adding `nil` defaults to `ENV.fetch` calls where we didn't have them before, all of this looks reasonable to me.

services/QuillLMS/app/models/feedback_history.rb
brendanshean: Thanks for catching that.  It's probably an artifact of some rebasing after I merged in parts 1-6.

PR #13637, opened 2026-02-10
Salesmate - part 2/2

happythenewsad:
## WHAT
- adds SchoolUploader, in the same pattern as DistrictUploader
- Adds Sidekiq wrapper and orchestrator classes 
- adds cron + API smearing, to stay under Salesmate rate limiting 

## WHY
- part of spec
- we want the data upload to happen async 
- self explanatory 

### Screenshots

### Notion Card Links
https://www.notion.so/quill/Sync-orgs-to-the-Salesmate-API-2eed42e6f94180578a22fa15e000bbbf

- [x] Run the SchoolUploader, then crosscheck the updated data with a historical snapshot
```
school = School.find 150600
SalesmateIntegration::SchoolUploader.run(record_id: school.id)
```
- [x] Run the SyncOrchestrator, check Sentry for errors afterward

PR Checklist | Your Answer
------------ | -------------


services/QuillLMS/app/services/salesmate_integration/adapter.rb
```diff
@@ -0,0 +1,74 @@
...
+      field_value
+    end
+
+    private def special_field?(field_name) = special_fields.key?(field_name)
+
+    private def handle_special_field(field_name) = special_fields[field_name].call(self)
+
+    private def date_field?(value) = value[:salesmate_internal_name]&.start_with?('dateCustomField')
+
+    # Abstract methods to be implemented by subclasses
```
brendanshean: I don't think this comment is needed.

services/QuillLMS/app/services/salesmate_integration/adapter.rb
```diff
@@ -0,0 +1,74 @@
...
+    private def record_exists_in_salesmate? = @record_exists_in_salesmate
+
+    private def resolve_field_value(salesmate_field_name, value)
+      return handle_special_field(salesmate_field_name) if special_field?(salesmate_field_name)
+
+      vitally_field = value[:vitally_field]
+      return nil unless vitally_field
+
+      field_value = vitally_data[vitally_field.to_sym]
+      return nil if date_field?(value) && field_value == 'N/A'
```
brendanshean: Let's make 'N/A' a constant

services/QuillLMS/app/services/salesmate_integration/adapter.rb
```diff
@@ -0,0 +1,74 @@
...
+
+      transform_record
+    end
+
+    private def transform_record
+      payload = {}
+      field_mappings.each do |key, value|
+        payload[value[:salesmate_internal_name]] = resolve_field_value(key, value)
+      end
+      return payload.merge(new_record_properties) unless record_exists_in_salesmate?
```
brendanshean: I think `if` construction reads more naturally here:

```
return payload if record_exists_in_salesmate?

payload.merge(new_record_properties)
```

brendanshean (commented):
Nice work on putting this all together.

services/QuillLMS/app/services/salesmate_integration/sync_orchestrator.rb
```diff
@@ -0,0 +1,52 @@
+# frozen_string_literal: true
+
+module SalesmateIntegration
+  class SyncOrchestrator < ApplicationService
+    SMEAR_RATE_IN_SECONDS = 5
+    DISTRICT_RECORD_LIMIT = 200
+    SCHOOL_RECORD_LIMIT = 1000
```
anathomical: Am I understanding the idea here being that we basically don't expect more than these limits of each type to have been updated in the last 24 hours (between syncs)?  So if, somehow, 201 Districts got updated in a 24 hour period, the last District wouldn't get queued for an update?  Would it make more sense to use `where('updated_at >= ?', DateTime.current - 24.hours)` or something like that in the `districts_to_sync` and `schools_to_sync` queries, or are these constants so much higher than actual updated-at counts that we don't really risk missing an actual update?

services/QuillLMS/spec/services/salesmate_integration/client_spec.rb
```diff
@@ -68,43 +68,97 @@
...
-        error_response = double(code: 404, message: 'Not Found', body: '{}')
+        error_response = double(
+          code: 404,
+          message: 'Not Found',
+          body: '{}',
+          parsed_response: { 'Error' => { 'message' => 'Resource not found' } }
+        )
+        allow(error_response).to receive(:dig).with('Error', 'message').and_return('Resource not found')
         allow(HTTParty).to receive(:post).and_return(error_response)
         allow(ErrorNotifier).to receive(:report)
```
anathomical: It might be overkill, but we could in theory extract this out into something at a higher context level since we're doing multiple error messaging tests:
```
let(:error_message) { 'Resource not found' }
let(:code) { 404 }
let(:message) { 'Not Found' }
let(:error_response) { double(code:, message:, body: {}, parsed_response: { 'Error' => { 'message' => error_message } }) }

before do
  allow(error_response).to receive(:dig).with('Error', 'message').and_return(error_message)
  ... # other allows
end
```

services/QuillLMS/spec/services/salesmate_integration/sync_orchestrator_spec.rb
```diff
@@ -0,0 +1,186 @@
...
+          .to receive(:perform_in).with(10.seconds, school_with_admin.id)
+
+        subject
+      end
+
+      it 'orders schools by updated_at desc' do
+        allow(SalesmateIntegration::DistrictUploaderWorker).to receive(:perform_in)
+
+        expect(SalesmateIntegration::SchoolUploaderWorker)
+          .to receive(:perform_in).with(5.seconds, school_with_user.id).ordered
```
anathomical: Maybe overkill, but I wonder if it's worth referencing `SyncOrchestrator::SMEAR_RATE_IN_SECONDS` instead of hard-coding the numbers here.

services/QuillLMS/spec/services/salesmate_integration/sync_orchestrator_spec.rb
```diff
@@ -0,0 +1,186 @@
...
+        expect(SalesmateIntegration::SchoolUploaderWorker)
+          .to receive(:perform_in).with(10.seconds, school_with_admin.id).ordered
+
+        subject
+      end
+    end
+
+    context 'with proper delays' do
+      let!(:districts) { create_list(:district, 3) }
+      let!(:schools) { districts.map { |d| create(:school, district: d) } }
```
anathomical, line 90: Personal style preference, so feel free to skip, but I'd be inclined to rename the mapped variable to `district` so you can have it passed to `create` implicitly: `districts.map { |district| create(:school, district:) }`

services/QuillLMS/app/models/cron.rb
```diff
@@ -54,6 +54,8 @@ def self.interval_1_day
     LearnWorldsIntegration::SyncOrchestratorWorker.perform_async if production?
     CalculateAndCacheSchoolsDataForSegmentWorker.perform_async if production?
     SendSegmentIdentifyCallForAllAdminsWorker.perform_async if production?
+    # TODO: uncomment below line after backfill is complete
+    # SalesmateIntegration::SyncOrchestratorWorker.perform_async if production?
```
anathomical, line 58: I'm sure this is already the plan, but I just wanted to flag that this will need to be done shortly after the backfill to make sure we don't accidentally generate a gap between the backfill and our update start.

anathomical (approved):
Looks good to me.  A couple of style thoughts, and some notes on how to handle generating our window for which items to update, but the fundamental structure here looks reasonable to me.

services/QuillLMS/app/services/salesmate_integration/sync_orchestrator.rb
happythenewsad: Yes, that is what this is doing. I like your idea better (since it scales infinitely with usage), so i'm going to go with that.

brendanshean (approved):

PR #13640, opened 2026-02-10
standards-report-performance-tweak

anathomical:
## WHAT
Tweak how data is mapped to take advantage of includes directive
## WHY
The old approach wasn't taking advantage of the `includes` directive and was doing hundreds of individual database fetches to extract `Concept` data
## HOW
Don't rely on the `has_many ... through` relationship in the `StateStandard` model.  Instead be explicit about the attributes being referenced

- Run service calls through the console in staging mode and watch the queries stacked
- Confirm that the `concepts` table is being queried way too often during service runs
- Confirm that after the change, there is only one query run to extract data

PR Checklist | Your Answer
------------ | -------------
Have you added and/or updated tests? | No, there is no expected behavioral change


dandrabik (approved):
Nice find.

PR #13645, opened 2026-02-11
Add source_example_id to Example

brendanshean:
## WHAT                                                                                    
* Added a `source_example_id` column to evidence_research_gen_ai_examples that tracks which 
original example a copied example originated from. 
* The serializer exposes an `original_created_at` field derived from the source example, and the frontend uses it for 
the Date column. 
* a thor task backfills existing data by matching examples across dataset versions by student_response.                                                     

## WHY
* When a new dataset version is created (via the dataset editor or activity review import), `ExamplesCopier` creates new Example records with fresh created_at timestamps. The Date column in the dataset editor displays created_at, so the date drifts with every version — making it look like synthetic data was generated on the date of the new dataset rather than when it was actually generated. The date should remain stable to reflect the original generation date.

## HOW
* Add migration and update Examples Copier to sets `source_example_id on every copy`, always pointing to the root
original (propagates existing source_example_id, or uses the source's own id if it's the first copy)
* ExampleSerializer adds original_created_at — returns the source example's created_at, falling back to own created_at for originals
* Thor task wraps `SourceExampleBackfiller` service walks datasets in version order per stem_vault, matches examples across versions by student_response (type-agnostic, since type can change from draft → test → prompt), and sets source_example_id to the root

### Screenshots

### Notion Card Links
Date for Synthetic responses should remain stable

PR Checklist | Your Answer
------------ | -------------


dandrabik (approved):
Looks good, this will be useful to keep track of related data.

PR #13647, opened 2026-02-11
standards-report-memory-use-enhancements

anathomical:
## WHAT
- refactor `RawSqlRunner` to pre-parse results into an array and ensure memory clean-up immediately after
- - Update a number of places in the code that use `RawSqlRunner` to use `flat_map` instead of `flatten.map` to process the new return value more cleanly
- refactor `all_state_standards_for_framework` to do one `map` call instead of a chain of `map` calls taking up more memory
## WHY
- `RawSqlRunner` uses some C libraries that do get garbage collected, but not by Ruby, and thus may linger outside of their immediate scope.  This fix should push the memory load into pure Ruby-land so garbage collection is more timely
- This is a lot less significant an enhancement, but should be notably more memory efficient
## HOW
- Convert the result to an array and return that value instead of the `pg_result` object, then ensure that the `pg_result` gets purged
- Replace `.map.map.map` chain

### Notion Card Links
https://trello.com/c/uwWX7U9M/51-fix-memory-leak-issue-thats-causing-the-server-to-intermittently-crash

PR Checklist | Your Answer
------------ | -------------
Have you added and/or updated tests? | No expected behavior changes


emilia-friedberg (approved):
LGTM! Thank you for doing this.

brendanshean (commented):
There are a bunch of other RawSqlRunner calls that have the form

`RawSqlRunner.execute(...).to_a`

I think we should update those to `RawSqlRunner.execute(...)`

services/QuillLMS/app/services/raw_sql_runner.rb
```diff
@@ -19,8 +19,10 @@ class RawSqlRunner
     end
 
   def self.execute(sql, name = nil)
-    ActiveRecord::Base.connection.execute(sql, name).tap do |pg_result|
-      pg_result.type_map = DEFAULT_TYPE_MAP_BY_OID
-    end
+    pg_result = ActiveRecord::Base.connection.execute(sql, name)
```
happythenewsad, line 22: Is there a writeup somewhere on the  internals of this solution?

happythenewsad (approved):

services/QuillLMS/app/services/raw_sql_runner.rb
brendanshean: Here's the original [PR](https://github.com/empirical-org/Empirical-Core/pull/7711)

PR #13651, opened 2026-02-12
standards-report-update-skill-nesting

anathomical:
## WHAT
Refactor query fetching to shape data to `{ l2_name: [l1_name1, l1_name2] }` for use on the front-end
## WHY
This format is necessary to show this data the way we want to on the front end
## HOW
Because the `Concept` tables are relatively small, doing this in SQL should be pretty fast, so do the roll-up there instead of messing around with the data in Ruby post-query

- Confirm new payload shape comes through on staging
- Confirm that performance is not negatively impacted

PR Checklist | Your Answer
------------ | -------------


emilia-friedberg (approved):
Thank you for making this change so quickly!

PR #13664, opened 2026-02-17
standards-report-stop-double-counting-questions

anathomical:
## WHAT
Tweak JOIN logic since we moved standards into a pre-query
## WHY
Joining `concept_uids` in both parts of the query was causing questions to be double-counted in cases where multiple standards were relevant to same question
## HOW
Remove the `concept_uids` JOIN from the last segment of the query since we had already moved that JOIN into the earlier part of the query

- Confirm known case of double counting questions properly single-counted them after the fix

PR Checklist | Your Answer
------------ | -------------


dandrabik (approved):
Nice find!

happythenewsad (approved):

PR #13665, opened 2026-02-18
tweak to z-index for tooltips attached to dropdowns

emilia-friedberg:
## WHAT
Tweak the z-index for tooltips attached to dropdowns so it only gets set to its high value when the tooltip is showing.

## WHY
I noticed that this change had been making it so that the hitbox for tooltips was too broad: it extended as wide as the tooltip itself would, so interacting with elements near to it, instead of just the trigger itself, would cause the tooltip to show, which was annoying. This fixes that so that the z-index is only set to this value when it is already open (necessary to prevent the tooltip from being obscured by the dropdown).

## HOW
Slightly narrow the rule.

### Screenshots

### Notion Card Links

Tested locally

PR Checklist | Your Answer
------------ | -------------
Have you added and/or updated tests? |  NO - CSS


happythenewsad (approved):

anathomical (approved):
Good catch!

PR #13674, opened 2026-02-24
add timestamps to activities unit template

emilia-friedberg:
## WHAT
Add timestamps to ActivitiesUnitTemplates table.

## WHY
Currently, we don't have them. This is a problem because, when these are changed, it is hard to tell when the change occurred -- see Notion card below for example of when this is problematic.

## HOW
Just add `created_at` and `updated_at` columns to this table.

### Screenshots

### Notion Card Links
https://www.notion.so/quill/FITB-migration-resulting-in-duplicate-assignments-2efd42e6f941801ab49acb509fc4d02f?d=7aad42e6f94182e39da7034d7e5f445a

Ran it locally, confirmed in console that updating and creating a record populate with timestamps as expected.

PR Checklist | Your Answer
------------ | -------------
Have you added and/or updated tests? | N/A - migration


brendanshean (approved):

happythenewsad (approved):
< 10,000 of these records in prod, so the migration should be fine

anathomical (approved):

